PR #2811, opened 2015-06-05
[ResourceBundle][TranslationBundle] Fix sorting on translated properties

cdaguerre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none
| License       | MIT
| Doc PR        | none

Check if properties are translated when sorting (as it's already done when filtering).

### DIFF
--- a/src/Sylius/Bundle/TranslationBundle/Doctrine/ODM/MongoDB/TranslatableResourceRepository.php
+++ b/src/Sylius/Bundle/TranslationBundle/Doctrine/ODM/MongoDB/TranslatableResourceRepository.php
@@ -69,9 +69,7 @@ class TranslatableResourceRepository extends DocumentRepository implements Trans
     }
 
     /**
-     * @param QueryBuilder $queryBuilder
-     *
-     * @param array $criteria
+     * {inheritdoc}.
      */
     protected function applyCriteria(QueryBuilder $queryBuilder, array $criteria = null)
     {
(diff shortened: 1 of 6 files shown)


pjedrzejewski:
@cdaguerre Nice, could you have a look at the failing build?

src/Sylius/Bundle/TranslationBundle/Doctrine/ODM/MongoDB/TranslatableResourceRepository.php
```diff
@@ -80,19 +78,43 @@ protected function applyCriteria(QueryBuilder $queryBuilder, array $criteria = n
...
+                    ->field($this->getPropertyName($property))->equals($value)
                 ;
             }
         }
     }
+
+    /**
+     * {inheritdoc}.
+     */
+    protected function applySorting(QueryBuilder $queryBuilder, array $sorting = null)
```
michalmarcinkowski: Why not simply initialize ```$sorting``` as an empty array?

```php
protected function applySorting(QueryBuilder $queryBuilder, array $sorting = array())
```

We could now simplify method by deleting check for null variable.

pjedrzejewski: This is convention from Doctrine, they do the same.

pjedrzejewski:
Thank you Christian! Nice work! :+1:

PR #2816, opened 2015-06-06
Update messages.es.yml

ysramirez:


pjedrzejewski:
Translation adjustments should be done on translate.sylius.org. :) Thanks!

PR #2835, opened 2015-06-11
[WebBundle] Change $.fn.datepicker format to yyyy-mm-dd.

jimbocoder:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no (I think)
| Deprecations? | no
| Fixed tickets | [comma separated list of tickets fixed by the PR]
| License       | MIT
| Doc PR        | [The reference to the documentation PR if any]

The motivation here is to allow the datetime comparison operators in the
database layer to was done so that native database comparisons can do
their job.


pjedrzejewski:
Sounds good to me! :+1: Thank you Aleksey!

aRn0D:
:-1: the format depends on the locale...

PR #2838, opened 2015-06-11
[ResourceBundle] check if requested paginate on allowed paginate array

Strontium-90:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | improvement |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets |  |
| License | MIT |
| Doc PR |  |

If we enable custom pagination size by setting `paginate: $paginate` user can request exstra large data by `paginate=10000`. But we can chech if `paginate` is in array `allowed_paginate`.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/Configuration.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/Configuration.php
@@ -273,7 +273,12 @@ class Configuration
 
     public function getPaginationMaxPerPage()
     {
-        return (int) $this->parameters->get('paginate', $this->settings['default_page_size']);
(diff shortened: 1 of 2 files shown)


michalmarcinkowski:
I agree that we should not allow for extra large data sets to be displayed in some cases and we should support some mechanism to prevent it, but I suppose that ``` allowed_paginate ``` parameter was aimed at frontend (to allow users change their preference about the number of results by page) and it is not the best idea to use it as validation.

Do you have some better idea for that? We could use ``` max_paginate ``` parameter, but I don't have idea for its default value...

PR #2840, opened 2015-06-11
Move first&last names in separate validation group

winzou:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

First & last names can be null in the model, so we can't force them in the form.

Plus, it's super easy to make them compulsory if they are optional (just add the validation group). The other way is not straight forward at all given the validation system!


pjedrzejewski:
Hmmm... How about we name this group s``ylius_name_required`` or is it a bad idea? I think ``name`` is to vague.

michalmarcinkowski:
We should think about convention, because right now we use e.g. 'guest', 'registration', 'user_create'. Should we prefix all that with 'sylius'? sylius_guest, sylius_registration, sylius_user_create, sylius_customer_name?

winzou:
Agreed with everything.

pjedrzejewski:
@michalmarcinkowski What do you think about this naming?

winzou:
no news is good news? :)

pjedrzejewski:
I think we should set some convention before we merge this and follow it everywhere from now on. I was thinking that "sylius" is default group, but for extra stuff we have more expressive group names, like "sylius_name_required". What do you think about this idea?

winzou:
We will never succeed in making groups convenient for everyone. So I would suggest to have just the `sylius` one for really mandatory fields, and then a few semantic others, like here `profile` and `registration`.

`name_required` is not semantic but field-related, I'm not sure it's a good way to go, otherwise we should have `birthday_required` and so on...

The ideal solution would be to have a flexible validation system. Overriding the validation groups would be amazing, but is not possible given the current validation system. Maybe a sylius extension adding this feature?

michalmarcinkowski:
I agree with @winzou, we should keep it simple for now.

Nice idea about the validation system, but I'm not sure we will have time for this feature in the nearest future :)

One more thing, I would change groups to more specific e.g. ``` sylius_customer_profile ```, ``` sylius_user_registration ```, ``` sylius_customer_guest ``` etc.  to explicitly indicate the resource WDYT?

winzou:
Done and green :)

michalmarcinkowski:
Great! Thanks Alexandre :+1:

pjedrzejewski:
Thanks a lot Alexandre! :)

PR #2844, opened 2015-06-12
[ThemeBundle] Theme bundle

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | none known
| Deprecations? | no
| License       | MIT
| Doc PR        | [Sylius/Sylius-Docs#306](https://github.com/Sylius/Sylius-Docs/pull/306)


src/Sylius/Bundle/ThemeBundle/.gitignore
```diff
@@ -0,0 +1,5 @@
+vendor/
+bin/
+
+composer.phar
+composer.lock
```
lchrusciel: composer.lock should not be ignored

src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstaller.php
```diff
@@ -0,0 +1,224 @@
...
+use Sylius\Bundle\ThemeBundle\Asset\PathResolverInterface;
+use Sylius\Bundle\ThemeBundle\Repository\ThemeRepositoryInterface;
+use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\SplFileInfo;
+use Symfony\Component\HttpKernel\Bundle\BundleInterface;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+class AssetsInstaller implements AssetsInstallerInterface
```
Zales0123: Missing author block.

src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstallerInterface.php
```diff
@@ -0,0 +1,53 @@
+<?php
+namespace Sylius\Bundle\ThemeBundle\Asset\Installer;
```
lchrusciel: No blankline before namespace

aRn0D:
Quick review on commit messages: "WIP" is useless and you should git commit --amend to avoid two commits with the same message ;)

src/Sylius/Bundle/ThemeBundle/.gitignore
pamil: The same is in every Sylius bundle and component, need to discuss this issue globally.

pamil:
@aRn0D thanks, commits will be squashed and made readable as soon as I get back from my holiday :)

aRn0D:
@pamil no problem, good holidays !

src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstallerInterface.php
```diff
@@ -0,0 +1,53 @@
...
+
+use Sylius\Bundle\ThemeBundle\Command\Model\SymlinkMask;
+use Symfony\Component\HttpKernel\Bundle\BundleInterface;
+
+
```
Zales0123: Extra blank line

steffenbrem:
What if you want your themes to be defined in a database? For example, in my case I can't use the filesystem to store my templates, so alternative ways should be easily implementable. Is it possible to implement something like this with the current state of this ThemeBundle?

aramalipoor:
@pamil Good job. Finally we're talking about theming :)

Let me add my two cents here hope it will help:

1. First question is why not treat Themes **as Bundles**? If we do so we make use of the world of bundles. E.g. A theme can register a **configuration FormType** so that use can change some preferences. Or if we ever created a Plugins bundle, downloading, installing, enabling of themes can easily be done with that plugins bundle. Or in a more fancy world (:D) a theme can extend or add some specific functionalities to the system. TL/DR When themes are bundles they are unstoppable.

2. We can add **automatic cache refreshing** (probably on `Kernel::onTerminate` event) which helps the user to not need to clear the cache manually. (Actually I've implemented a PoC and would be happy to contribute when this PR is merged)

3. If we have configurable themes (described in 1) so we can easily have **presets**! (Halloween, Summer, etc.) and a PresetContext can be there to choose what preset to be loaded for an specific request. We can also have different presets for different channels.

4. Again if they are bundles (described in 1) themes can easily register themselves in a central registry so that we can have a management **admin UI** for themes.

P.S To make a powerful ThemeBundle we need a powerful PluginsBundle.

aramalipoor:
@pamil Why do you need to change the default behavior of `assets:install`?

pjedrzejewski:
Thanks for your input @aramalipoor!

We want to do exactly what you said, except we do not want to make themes unstoppable, to discourage people doing nasty things there. Theme = templates, css, js, images and that's it. We will probably use Settings bundle to make them configurable as well, but that's it.

Plugins, this is another story. :) We definitely want to make them bundles. Actually, my very very old bundles (2011, crazy!) was implementing that concept. Making Sylius plugin awesome is our priority. :) I will be very happy to hear more ideas from you and everyone in the community. We will open a plugin RFC to discuss this. :)

I'd be also VERY interested in seeing your solution for the cache problem. :) The issue that https://github.com/Sylius/SyliusPluginsBundle had was that if the plugin has wrong services definitions, missing parameter or generally even the smallest bug that breaks the container compilation - app becomes broken and you cannot remove/uninstall the plugin easily...

pamil:
@steffenbrem I would like to see your implementation of database-based templates to see how it works and form my own opinion about it. I think that it should be possible, but provided by standalone bundle.

pamil:
@armalipoor useful comments, made me think it over once again :) I agree with @pjedrzejewski that we should have some rules and themes should change the way the application looks, not works. 

I'd also be interested in your solution of cache refreshing. I'm not into that SyliusPluginsBundle issue, but I think that's OK to do it automatically in  dev mode.

pamil:
@aramalipoor got typo in your name and can't see the availability to change comment above on mobile :) 

In case of `assets:install` command, assets from themes are renamed not to overwrite base asset  and they come from more than one directory, so I can't link whole directory as it have been done video. 

Moreover, I can't safely put theme assets in different directory because in case of CSS files it might break the links inside them.

aramalipoor:
@pjedrzejewski you're absolutely right ;) by unstoppable I mainly mean in theming features and scenarios, compared to other theming systems out there :palm_tree:

About cache refreshing, in simple and good scenarios @pamil, as you'd guess one would easily mark container for invalidation and then on `kernel.terminate` call remove() on `cache_clearer` service and remove all files under `%kernel.cache_dir%`. I've put some [snippets here](https://gist.github.com/aramalipoor/6e67bc6c3d936428661e) hope it's useful.

But @pjedrzejewski in harder scenarios I was considering using [this library](https://github.com/matthiasnoback/symfony-service-definition-validator) to check if there's any problems with a plugin and some kind a rollback if there's any! Since I'm not a symfony expert I'm not sure if its possible and is it going to cover all cases?!

aramalipoor:
@pamil If a theme's CSS files are only referencing to their own assets they're OK to be in their own directories. Do you want to cover a case where a theme is referencing an asset from base assets/another theme?

src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstaller.php
```diff
@@ -0,0 +1,224 @@
...
+        foreach ($this->themeRepository->findAll() as $theme) {
+            $sourceDir = $theme->getPath() . '/' . $bundle->getName() . '/public';
+            if (is_dir($sourceDir)) {
+                $sources[] = $sourceDir;
+            }
+        }
+
+        return $sources;
+    }
+}
```
vladapopster: Add new line.

pamil:
@aRn0D Commits fixed :)

@aramalipoor I will try to apply that cache invalidation idea you have mentioned.

In case of assets, their behaviour will change a lot in next commits, there will be support for extending themes by other themes, but there won't be any possibility to support `ThemeContext` builded from many themes on runtime.

src/Sylius/Bundle/ThemeBundle/.gitignore
pjedrzejewski: In case of Sylius/Sylius, it should not be ignored. :)

src/Sylius/Bundle/ThemeBundle/composer.json
```diff
@@ -0,0 +1,46 @@
...
+        },
+        {
+            "name":     "Community contributions",
+            "homepage": "http://github.com/Sylius/Sylius/contributors"
+        }
+    ],
+    "require": {
+        "php":                      ">=5.4",
+
+        "sylius/resource-bundle":   "0.14.*@dev",
```
sstok: I can be wrong but the resource-bundle doesn't seem to be used by any of the commits.
Is it possible to use the ThemeBundle without having to install the ResourceBundle?

src/Sylius/Bundle/ThemeBundle/.gitignore
pamil: And it isn't, it's `Sylius/ThemeBundle`'s `composer.lock` file.

lchrusciel: ok, my bad. Nevermind.

src/Sylius/Bundle/ThemeBundle/composer.json
pamil: You're right, thanks :+1:

src/Sylius/Bundle/ThemeBundle/.gitignore
pamil: Better safe than sorry :)

pamil:
Finally, ThemeBundle out of development phase, ready for review and deep testing :) @Sylius

aramalipoor:
@pamil Nice job ;)

.travis.yml
```diff
@@ -16,11 +15,6 @@ env:
 matrix:
     allow_failures:
         - php: 7.0
-    include:
-        - php: 5.3
-          env: TEST_SUITE=1 COMPOSER_FLAGS="--prefer-lowest"
-        - php: 5.3
```
michalmarcinkowski: Should we remove it or just change to php 5.4?

composer.json
```diff
@@ -28,7 +28,7 @@
         "doctrine/doctrine-fixtures-bundle":    "~2.2",
         "doctrine/doctrine-migrations-bundle":  "~1.0",
         "doctrine/migrations":                  "~1.0@dev",
-        "doctrine/orm":                         "~2.4",
+        "doctrine/orm":                         "~2.4.0",
```
michalmarcinkowski: Why we need a ``` .0 ``` added?

pamil: While using Doctrine 2.5 I run into some errors:
- It changed some method used in many specs, so they should be changed too
- It made some Behat suite not passing (I think it was country management in backend)

.travis.yml
pamil: It may be good idea, but test are passing anyway. I think that we need another PR to discuss version bump related issues. This operation shouldn't be hidden in ThemeBundle :)

composer.json
michalmarcinkowski: Ok :+1:

.travis.yml
michalmarcinkowski: It will be discussed separately, but I think changing it to 5.4 is a must :wink: Anyway you can leave it for now.

src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstaller.php
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * @param string $origin
+     * @param string $target
+     * @param boolean $symlink
+     *
+     * @throws IOException When failed to make symbolic link, if requested.
+     */
+    protected function doInstallAsset($origin, $target, $symlink)
+    {
+        if ($symlink) {
```
lchrusciel: Consider reversing condition, to eliminate nested if and else statements. 
```php 
if ($symlink) {
    $this->filesystem->symlink($origin, $target);
    
    if (!file_exists($target)) {
        throw new IOException('Symbolic link is broken');
    }
    
    return;
}

if (is_dir($origin)) {
    $this->filesystem->mkdir($target, 0777);
    $this->filesystem->mirror($origin, $target, Finder::create()->ignoreDotFiles(false)->in($origin));
    
    return;
}

$this->filesystem->copy($origin, $target);
```
or even move if($symlink) body to private method, to avoid nested if.

src/Sylius/Bundle/ThemeBundle/Asset/Installer/OutputAwareAssetsInstaller.php
```diff
@@ -0,0 +1,103 @@
...
+        $this->output = $output;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function installAssets($targetDir, $symlinkMask)
+    {
+        if ($this->hasOutput()) {
+            if (AssetsInstallerInterface::HARD_COPY !== $symlinkMask) {
```
lchrusciel: IMO it should go to the private method. In private method we can also avoid else statement,

Zales0123: Sure. You could extract method that returns only proper message, which is dependent on ``$symlinkMask`` value ;)

pamil: :+1:

src/Sylius/Bundle/ThemeBundle/Asset/Package/PathPackage.php
```diff
@@ -0,0 +1,55 @@
...
+     * @var ThemeContextInterface
+     */
+    protected $themeContext;
+
+    /**
+     * @var PathResolverInterface
+     */
+    protected $pathResolver;
+
+    public function __construct(
```
Zales0123: Missing docblock for constructor.

src/Sylius/Bundle/ThemeBundle/Command/AssetsInstallCommand.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        $this
+            ->setName('assets:install')
+            ->setDefinition([
+                new InputArgument('target', InputArgument::OPTIONAL, 'The target directory', 'web'),
+            ])
+            ->addOption('symlink', null, InputOption::VALUE_NONE, 'Symlinks the assets instead of copying it')
+            ->addOption('relative', null, InputOption::VALUE_NONE, 'Make relative symlinks')
+            ->setDescription('Installs bundles web assets under a public web directory')
+            ->setHelp(<<<EOT
```
Zales0123: Maybe return it from some private function, it's not needed to interfere readability in configuration method.

src/Sylius/Bundle/ThemeBundle/Command/SyliusThemeDebugCommand.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    protected function configure()
+    {
+        $this
+            ->setName('sylius:theme:debug')
+            ->setDescription('Shows list of detected themes.')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
```
Zales0123: {@inheritdoc} missing.

src/Sylius/Bundle/ThemeBundle/Command/SyliusThemeDebugCommand.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $this->showSuccessfullyLoadedThemes($output);
+    }
+
+    private function showSuccessfullyLoadedThemes(OutputInterface $output)
```
Zales0123: Docblock missing.

src/Sylius/Bundle/ThemeBundle/DependencyInjection/Compiler/ThemeCompilerPass.php
```diff
@@ -0,0 +1,36 @@
...
+
+        $serializedThemes = [];
+        try {
+            $themeFiles = $container->get('sylius.theme.locator')->locate('theme.json', null, false);
+            foreach ($themeFiles as $themeFile) {
+                $serializedThemes[] = serialize($loader->load($themeFile));
+                $container->addResource(new FileResource($themeFile));
+            }
+
+            $definition->addArgument($serializedThemes);
```
Zales0123: Maybe extract this to some well named method, again, to improve readability.

src/Sylius/Bundle/ThemeBundle/Command/AssetsInstallCommand.php
pamil: :+1:

src/Sylius/Bundle/ThemeBundle/Locator/RecursiveFileLocator.php
```diff
@@ -0,0 +1,99 @@
+<?php
+
+namespace Sylius\Bundle\ThemeBundle\Locator;
+use Symfony\Component\Config\FileLocatorInterface;
```
Zales0123: Blank line.

src/Sylius/Bundle/ThemeBundle/Locator/RecursiveFileLocator.php
```diff
@@ -0,0 +1,99 @@
...
+
+        if (!$filepaths) {
+            throw new \InvalidArgumentException(sprintf('The file "%s" does not exist (in: %s).', $name, implode(', ', $directories)));
+        }
+
+        return array_values(array_unique($filepaths));
+    }
+
+    /**
+     * Returns whether the file path is an absolute path.
```
Zales0123: Isn't that obvious (method name)?

src/Sylius/Bundle/ThemeBundle/Collector/ThemeDataCollector.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public function getAllThemes()
+    {
+        return $this->allThemes;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function collect(Request $request, Response $response, \Exception $exception = null)
```
lchrusciel: What do you want to achieve with $exception argument? YAGNI?

src/Sylius/Bundle/ThemeBundle/Locator/RecursiveFileLocator.php
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @param string $file A file path
```
Zales0123: Maybe $filePath, which allows us to avoid description

src/Sylius/Bundle/ThemeBundle/Locator/RecursiveFileLocator.php
pamil: Copypasted it from Symfony2 :)

src/Sylius/Bundle/ThemeBundle/Locator/RecursiveFileLocator.php
```diff
@@ -0,0 +1,99 @@
...
+        $finder
+            ->files()
+            ->name($name)
+            ->ignoreUnreadableDirs()
+            ->in($directories)
+        ;
+
+        /** @var SplFileInfo $file */
+        if (true === $first) {
+            if (null !== $file = $finder->getIterator()->current()) {
```
Zales0123: Maybe some private methods? I know, I'm boring ;)

src/Sylius/Bundle/ThemeBundle/Locator/RecursiveFileLocator.php
pamil: :+1:

src/Sylius/Bundle/ThemeBundle/Locator/RecursiveFileLocator.php
pamil: I think it's clear enough, we don't want to end up with thousand private methods :)

src/Sylius/Bundle/ThemeBundle/DependencyInjection/Compiler/ThemeCompilerPass.php
pamil: How would you do that? I don't think it'll make it cleaner in this case.

src/Sylius/Bundle/ThemeBundle/Factory/ThemeFactory.php
```diff
@@ -0,0 +1,87 @@
...
+        $themeData = $this->optionsResolver->resolve($themeData);
+
+        foreach ($themeData as $attributeKey => $attributeValue)
+        {
+            $attributeKey = $this->normalizeAttributeKey($attributeKey);
+
+            try {
+                $this->propertyAccessor->setValue($theme, $attributeKey, $attributeValue);
+            } catch (NoSuchPropertyException $exception) {
+                // Ignore properties that does not exist in given theme model.
```
lchrusciel: I'm not sure about ignoring exceptions. I understand what you wanted to achieve here, but ignoring and commenting exceptions is not the best solution. It happened not only here.

src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstaller.php
pamil: :+1:, done :)

src/Sylius/Bundle/ThemeBundle/Locator/RecursiveFileLocator.php
```diff
@@ -0,0 +1,99 @@
...
+        $finder = new Finder();
+        $finder
+            ->files()
+            ->name($name)
+            ->ignoreUnreadableDirs()
+            ->in($directories)
+        ;
+
+        /** @var SplFileInfo $file */
+        if (true === $first) {
```
lchrusciel:
```php
if ($first) {
```
Would be enough.

src/Sylius/Bundle/ThemeBundle/Locator/RecursiveFileLocator.php
lchrusciel: Again nested if. IMHO more readable condision will be
```php
if ($first && null !== ($file = $finder->getIterator()->current())) {
```

src/Sylius/Bundle/ThemeBundle/Locator/RecursiveFileLocator.php
```diff
@@ -0,0 +1,99 @@
...
+            if (null !== $file = $finder->getIterator()->current()) {
+                return $file->getPathname();
+            }
+        } else {
```
lchrusciel: Avoid else.

src/Sylius/Bundle/ThemeBundle/DependencyInjection/Compiler/ThemeCompilerPass.php
Zales0123: I would create method ``addSerializedThemesToDefinition`` or something like this, and throw whole "try" block content inside. I know it required some parameters ($definition and $loader, for example) but it is easier to read for someone who does not **want** to care about this adding themes logic. IMO it should be transparent, but of course I am a little bit pedantic ;)

pamil: That's a good idea, but overall I think that it should be included in `process()` method comment instead of extracting private method that holds entire other method :)

Zales0123: Ok, for me method ``process`` is quite general and does not say almost anything about functionality, so i would precise it in private methods - but i know it can be controversial for some, so I'm not persuading. However i still think i would be nicer ;)

src/Sylius/Bundle/ThemeBundle/Templating/Locator/TemplateLocator.php
```diff
@@ -0,0 +1,270 @@
...
+        $resourceBundle = null;
+        $bundles = $this->kernel->getBundle($bundleName, false);
+
+        $parameters = [
+            '%app_path%' => $this->appDir,
+            '%override_path%' => substr($templatePath, strlen('Resources/')),
+        ];
+
+        foreach ($this->paths['bundle_template'] as $path) {
+            foreach ($bundles as $bundle) {
```
Zales0123: Ok, but this for sure can be simplified :smile:

src/Sylius/Bundle/ThemeBundle/Templating/Locator/TemplateLocator.php
```diff
@@ -0,0 +1,270 @@
...
+        foreach (array_merge($themes, [null]) as $theme) {
+            $result = $this->getCache($template, $theme);
+            if (null !== $result) {
+                return $result;
+            }
+        }
+
+        if (0 === strpos($template->getPath(), '@')) {
+            return $this->locateBundleTemplateUsingThemes($template, $themes);
+        } else {
```
lchrusciel: Unnecessary else.

src/Sylius/Bundle/ThemeBundle/Templating/Locator/TemplateLocator.php
```diff
@@ -0,0 +1,270 @@
...
+        ];
+
+        foreach ($this->paths['bundle_template'] as $path) {
+            foreach ($bundles as $bundle) {
+                $parameters = array_merge($parameters, [
+                    '%bundle_name%' => $bundle->getName(),
+                    '%bundle_path%' => $bundle->getPath(),
+                ]);
+
+                if (false === strpos($path, '%theme_path%')) {
```
Zales0123: Maybe join this and nested condition in one? We could also (surprise :)) extract it to private method, named simple and nicely, and this allows us to avoid filthy ``else`` statement.

src/Sylius/Bundle/ThemeBundle/Templating/Locator/TemplateLocator.php
```diff
@@ -0,0 +1,270 @@
...
+     */
+    protected function locateAppTemplateUsingThemes(TemplateReferenceInterface $template, array $themes = [])
+    {
+        $parameters = [
+            '%app_path%' => $this->appDir,
+            '%override_path%' => $template->getPath(),
+        ];
+
+        foreach ($this->paths['app_template'] as $path) {
+            if (false === strpos($path, '%theme_path%')) {
```
Zales0123: Same here - it surely can be simplified.

src/Sylius/Bundle/ThemeBundle/Templating/Locator/TemplateLocator.php
```diff
@@ -0,0 +1,270 @@
...
+        $path = strtr($path, $parameters);
+
+        if (file_exists($path)) {
+            return $this->cache[$key] = $path;
+        }
+
+        return null;
+    }
+
+    private function getCacheKey(TemplateReferenceInterface $template, ThemeInterface $theme = null)
```
Zales0123: Missing docblocks for this and next method.

src/Sylius/Bundle/ThemeBundle/Templating/Locator/TemplateLocator.php
```diff
@@ -0,0 +1,270 @@
...
+
+        return null;
+    }
+
+    /**
+     * @param string $path
+     * @param array $parameters
+     * @param TemplateReferenceInterface $template
+     * @param ThemeInterface $theme
+     * @return null|string
```
Zales0123: Missing blank line before ``@return``.

src/Sylius/Bundle/ThemeBundle/Templating/Locator/TemplateLocator.php
lchrusciel: What do you think about this condition?
```php
if (false !== strpos($path, '%theme_path%')) {
    foreach ($themes as $theme) {
        $themeParameters = array_merge($parameters, [
            '%theme_path%' => $theme->getPath(),
        ]);
        if (null !== $checkedPath = $this->checkPath($path, $themeParameters, $template, $theme)) {
            return $checkedPath;
        }
    }
}
if (null !== $checkedPath = $this->checkPath($path, $parameters, $template)) {
    return $checkedPath;
}
```
Futhermore we can extract foreach loop into private method.

src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstaller.php
```diff
@@ -0,0 +1,272 @@
...
+     * @param $target
+     */
+    private function doCopyAsset($origin, $target)
+    {
+        if (is_dir($origin)) {
+            $this->filesystem->mkdir($target, 0777);
+            $this->filesystem->mirror($origin, $target, Finder::create()->ignoreDotFiles(false)->in($origin));
+
+            return;
+        } else {
```
lchrusciel: No need for else here ;)

src/Sylius/Bundle/ThemeBundle/DependencyInjection/Compiler/ThemeCompilerPass.php
```diff
@@ -0,0 +1,38 @@
...
+        } catch (\InvalidArgumentException $e) {
+            // No files found.
```
lchrusciel: Do you have any idea how to refator this one to? Ignoring exceptions is not a good solution.

src/Sylius/Bundle/ThemeBundle/Tests/Functional/ThemeBundleTestCase.php
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * @return Client
+     */
+    protected function getClient()
+    {
+        $client = $this->createClient(['test_case' => self::TEST_CASE, 'root_config' => 'config.yml']);
+        try {
+            $client->insulate();
+        } catch (\RuntimeException $e) {
```
lchrusciel: Another exception silencer.

src/Sylius/Bundle/ThemeBundle/DependencyInjection/Compiler/ThemeCompilerPass.php
pamil: `sylius.theme.locator` implements `Symfony\Component\Config\FileLocatorInterface` which has `@throws \InvalidArgumentException When file is not found` in the docs, so I think it's the only solution available.

lchrusciel: Ok, i get it. But im not convinced about catching exceptions without any logic. Mayby we should leave it here and handle it in some upper class?

Symfomany:
All this changes in 0.16-dev tag?

Symfomany:
In composer.json , added             "Symfony\\Cmf\\Bundle\\CreateBundle\\Composer\\ScriptHandler::downloadCreate"
in "post-install-cmd"

 "Symfony\\Cmf\\Bundle\\CreateBundle\\Composer\\ScriptHandler::downloadCreate"
in "post-update-cmd"

pamil:
@Symfomany what does it have in common with `ThemeBundle`?

src/Sylius/Bundle/ThemeBundle/Locator/ResourceLocatorInterface.php
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+interface ResourceLocatorInterface
+{
+    /**
+     * @param string $resourceName
+     * @param ThemeInterface[] $themes
+     */
```
Zales0123: Missing ``@return`` statement.

pamil: :+1:

composer.json
```diff
@@ -58,18 +58,18 @@
...
-        "symfony/assetic-bundle":               "~2.3",
-        "symfony/event-dispatcher":             "~2.3",
-        "symfony/expression-language":          "~2.4",
-        "symfony/form":                         "~2.3",
-        "symfony/framework-bundle":             "~2.3",
-        "symfony/intl":                         "~2.3",
-        "symfony/monolog-bundle":               "~2.3",
-        "symfony/options-resolver":             "~2.3",
-        "symfony/property-access":              "~2.3",
+        "symfony/assetic-bundle":               "~2.6",
```
gperdomor: Why not update all symphony stack to 2.7?

composer.json
```diff
@@ -58,18 +58,18 @@
...
+        "symfony/event-dispatcher":             "~2.7",
+        "symfony/expression-language":          "~2.7",
+        "symfony/form":                         "~2.7",
+        "symfony/framework-bundle":             "~2.7",
+        "symfony/intl":                         "~2.7",
+        "symfony/monolog-bundle":               "~2.7",
+        "symfony/options-resolver":             "~2.7",
+        "symfony/property-access":              "~2.7",
         "symfony/swiftmailer-bundle":           "~2.3",
```
gperdomor: Why not update all symphony stack to 2.7?

Symfomany:
:+1:  @gperdomor  I have migrate to Symfony 2.7 ;)

composer.json
pamil: It's just a temporary solution, this PR should not make the bump to PHP5.4 / Sf2.7 alone, it should be made in another PR and then rebased (after removing 6aafbb1592f83730521412843323872d0b100cfb).

steffenbrem:
@pamil Do you use the Twig_Loader_Chain btw? Since that is what I am using to load theme files from the database. I will check if I am able to supply you the code we use to accomplish that.

pamil:
@steffenbrem it has been a long time since I was working with ThemeBundle, but I tried to make the template locator Twig-agnostic, so probably I don't use it. It would help me greatly to see how your template loading logic looks like.

I'm planning to divide the template locator into three parts - the first for providing paths (identifiers) of themes, the second for loading these paths (it could be in filesystem, database, anywhere) and the third one for gluing these two.

codingideas:
What i suggest to the themes is that we are able to have two features:

Child themes.
Change theme in the URL.

The benifit of changing the theme in the URL is for a/b testing. I can see myself testing conversion rate between entire themes - a vs b in google analytics content experiments. However if you're testing a full theme while the visitors browse it needs to be set in a cookie.

Now child themes are a mirror of the parent with overrides for the parent. This way CRO guys and test different home pages such as with products vs with categories or maybe moving positing on the add to cart to test goal flow. Also so that theme developers can upgrade the theme for new features and compatibility. 

This is just a conversation I wanted have some some of you guys. I feel adding this type of functionality would bring some good marketers to pitch this software to their clients.

aramalipoor:
@pamil You've done a great job! We need to have ThemeBundle in our current Sprint so we've extracted you code in our private fork.

We are working on some features like:
* **Theme Settings**-support in `theme.json`.
* A global `theming` setting which admin user can set currently active theme.
* A **theme downloader** which works with git repositories and zip files.
* And some random improvements to the code base.

What is the best strategy here to contribute these back here? Should we wait to see ThemeBundle merged? Or create some PRs to your `pamil` fork?

Anyway is there anything important that we need to know when thinking about these features?

pjedrzejewski:
@aramalipoor That's great! I suggest:

* Theme Settings - could you clarify what you mean? My idea was to use Settings bundle for that, so that every theme can have configuration editable in the backend and easily defined in the theme package with single settings schema class. (requires improvements to Settings bundle)

* Global ``theming`` setting - could you clarify?

* Theme downloader sounds great, but I'd leave that after this is merged. So best would be if you implement PoC on your side and then possibly open a PR into main repo.

* This is very welcome as a PR to ``pamil`` branch.

Let's hear Kamil too, cause I know he had some ideas too. We will try to get this merged as soon as possible, because this is without integration with Sylius yet.

One big problem I have with current implementation - it requires clearing cache when changing/adding theme. The good thing is that if we figure this out - we could do the same with plugins. :>

pamil:
Thanks @aramalipoor! :) 

As for theme downloader, I think we may use Composer and `composer.json` to deal with it and resolve theme dependencies, we would only need to add `vendor` directory to theme locator (don't remember how I named it :), something like that).

@pjedrzejewski, that is quite tough part of ThemeBundle. As it interrupts vanilla templates, translations and assets system, each one of them is another problem. 
 - Templates - are cached on `cache:warmup`, paths in `app/cache/env/templates.php`, templates translated to PHP in `app/cache/env/twig`
 - Translations - are cached in container and `cache:warmup`, paths to them are injected directly to container to `Translator` creation, contents translated to PHP can be found in `app/cache/env/translations`
 - Assets - are dumped on `assets:install`

We will have to find some way to populate these data without rebuilding cache in the filesystem and for invalidating these caches.

Anyway, we should consider performance as an important factor, ThemeBundle can't slow the whole application significantly. It would be nice to stop "memory leaks" in Behat tests and then use Blackfire on Travis to have real-time feedback on resource usage and introduced bottlenecks.

pamil:
My another idea is to allow templates / translations / assets loading also from other sources than filesystem, so we could even inject new themes if we cannot make changes to the filesystem.

aramalipoor:
@pjedrzejewski 
* For **Theme Settings** we've decided to create a simple `ThemeSettingsFormFactory` service in ThemeBundle (or maybe SettingsFormFactory in SettingsBundle/CoreBundle) which populates a `FormBuilder` instance based on a simple array/yaml, like below.   

 This form factory can become very complicated in time, but IMO its seed idea is powerful for many use-cases such as Theme settings, Dynamic user-desfined custom forms, etc. Right now we've only added text, taxonomy, cmf menu, and image types which is what we mostly need in themes.

 In this way there's no need to have any PHP files in themes files, also to support an `image` field you need to *access some services* which is hard to do if theme only includes a simple PHP class.
```yml
   settings:
      logo:
         type: image
         validators:
            - required
      copyright:
         type: text
         default: Copyright 2015
      left_taxonomy:
         type: taxonomy
      special_offer:
         type: product
      contact_email:
         type: text
         validators:
            - email
```
* By a global `theme` settings I mean a simple way to integrate it with Sylius core. Introduce an `active_theme` field in `sylius_general` settings which saves currently active theme set by admin.

 From what I've seen in ThemeBundle it seems that **changing** themes does not need a clear cache! Am I missing something?!

* About **ThemeDownloader** and using composer.json, the problem is that we're using Docker images to ship our production-ready app to our **many** instances. So actually everything in our code-base except `app/Resources`, `app/config/parameters.yml` and `app/cache(logs)` is included in our Docker image and is deployed by every release. Hence on every deploy, composer.json/lock **changes to the latest** release.

  So we thought that maybe offering a non-composer solution would help Docker-based deployments a more decoupled way of handling themes. But I'm pretty aware that this may not be the use-case Sylius wants to cover, due to it's complexity (or hopefully not :D).

src/Sylius/Bundle/ThemeBundle/Templating/Locator/TemplateLocator.php
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * @param TemplateReferenceInterface $template
+     * @param ThemeInterface[] $themes
+     *
+     * @return null|string
+     */
+    protected function locateTemplateUsingThemes(TemplateReferenceInterface $template, array $themes)
+    {
+        foreach (array_merge($themes, [null]) as $theme) {
```
aramalipoor: Why is `[null]` here? It seems it's causing bugs.

When doing a cache warmup, first you're looking for the original template path by clearing the context ([line 77 of cache warmer](https://github.com/Sylius/Sylius/pull/2844/files#diff-f865413c726b6bacc90538256bae651bR77)) this causes locator to save a cache entry to the original template. The problem occurs on [line 82 of cache warmer](https://github.com/Sylius/Sylius/pull/2844/files#diff-f865413c726b6bacc90538256bae651bR82)  where you want to locate theme's template path BUT because of `[null]` here it will find the original template from cache therefore instead of **locating** [on line 125 below](https://github.com/Sylius/Sylius/pull/2844/files#diff-70e90ffda6ff082018c7d364aa60322dR125) it will just return original template path from cache.

Two solutions here, forget about the enhancement (of not having duplicate template paths for all themes [by an IF here](https://github.com/Sylius/Sylius/pull/2844/files#diff-f865413c726b6bacc90538256bae651bR84)) and put `$templates[$template->getLogicalName()] = $this->locator->locate($template);` after the for loop, so it doesn't create cache for original tempalte before reading themes templates.

Or just remove `[null]` from here if it is not needed!
Thanks @pamil :D

aramalipoor:
As a side (but a pain-is-in-the-a** solver) idea I have to tell you guys, we're using this bundle and big problem that we're having now besides Theme settings is that themes usually need fixtures (pages, menus, images, etc) so that they can work properly out-of-the-box.

Right now I have no solution to this, but it would awesome if we could have a way to load fixture data and images for every theme which is being installed.

It's not really related to this PR and ThemeBundle as a low-level library but it's a theme-related matter to consider for future maybe ;)

src/Sylius/Bundle/ThemeBundle/Templating/Locator/TemplateLocator.php
pamil: To be honest, I have no idea (I haven't used it yet since I had written it). In the next weeks I'll be doing a project based on themes, so I'll dive into ThemeBundle and try to make it better :)

The first ideas are to allow themes to be loaded at the runtime, remove (or extract) unnecessary state in services (like in `Locator`) and to make templates and assets storage-independent.

pamil:
Sorry for the delay, I might have accidentally set this notification as readed. I'm not sure about modifying application data through theme installing. Images should be already loaded by `assets:install` command. Also, I might have not understand the idea :)

Anyway, we can allow that and any other behaviour by dispatching theme-related events.

pjedrzejewski:
@pamil I think the use-case is that a theme may have specific blocks, menus and pages, which need to be loaded to our PHPCR. :) Same for some images and so on.

aramalipoor:
@pjedrzejewski is right, PHPCR content. We've done a quick PoC around this issue, adding an ability to ThemeBundle to look for any `DataFixtures` classes in theme directory and loads them via DoctrineFixtures as usual (when admin is requesting to `activate` a theme). I'd be happy to share it when ThemeBundle is merged ;)

Now everything is working fail-safe (create menus, images only if not exists; thanks to PHPCR route-based identifiers)

Anyway I think we are going to need to have some container-aware *PHP classes* OR a *YAML parser* for theme's **settings** and their data *fixtures*. We have a PoC for both. 

Other than clearing cache problem everything seems to be working fine with ThemeBundle.

pjedrzejewski:
@aramalipoor :+1: Great, thank you for your feedback and would be great if you can share your solutions. Next week we will brainstorm around the cache bundle and try to push forward this PR in order to get ThemeBundle merged.

src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstaller.php
```diff
@@ -0,0 +1,278 @@
...
+        return $effectiveSymlinkMask;
+    }
+
+    /**
+     * @param ThemeInterface $theme
+     * @param string $originDir
+     * @param string $targetDir
+     * @param integer $symlinkMask
+     *
+     * @return integer
```
Zales0123: Should be ``int``.

src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstaller.php
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+        return $effectiveSymlinkMask;
+    }
+
+    /**
+     * @param ThemeInterface $theme
+     * @param string $originDir
+     * @param string $targetDir
+     * @param integer $symlinkMask
```
Zales0123: Same here.

src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstaller.php
```diff
@@ -0,0 +1,278 @@
...
+        if (is_dir($sourceDir)) {
+            $sources[] = $sourceDir;
+        }
+
+        return $sources;
+    }
+
+    /**
+     * @param string $origin
+     * @param string $target
```
Zales0123: What about adding ``@throws`` into docblock?

src/Sylius/Bundle/ThemeBundle/Asset/Installer/OutputAwareAssetsInstaller.php
```diff
@@ -0,0 +1,122 @@
...
+    private function provideResultComment($symlinkMask, $effectiveSymlinkMask)
+    {
+        if ($effectiveSymlinkMask === $symlinkMask) {
+            switch ($symlinkMask) {
+                case AssetsInstallerInterface::HARD_COPY:
+                    return 'The assets were copied.';
+                case AssetsInstallerInterface::SYMLINK:
+                    return 'The assets were installed using symbolic links.';
+                case AssetsInstallerInterface::RELATIVE_SYMLINK:
+                    return 'The assets were installed using relative symbolic links.';
```
Zales0123: I would rather use properly written array with ``$symlinkMask`` as a key and message as a value, but maybe it's only me :smile:

src/Sylius/Bundle/ThemeBundle/Collector/ThemeDataCollector.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * @return ThemeInterface[]
+     */
+    public function getUsedTheme()
+    {
+        return $this->usedTheme;
+    }
+
+    public function getThemeHierarchy()
```
Zales0123: Missing docblock.

src/Sylius/Bundle/ThemeBundle/Command/AssetsInstallCommand.php
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        if ($input->getOption('relative')) {
+            $symlinkMask = max($symlinkMask, AssetsInstallerInterface::RELATIVE_SYMLINK);
+        }
+
+        $assetsInstaller->installAssets($input->getArgument('target'), $symlinkMask);
+    }
+
+    private function getHelpMessage()
```
Zales0123: Missing docblock.

src/Sylius/Bundle/ThemeBundle/HierarchyProvider/ThemeHierarchyProvider.php
```diff
@@ -0,0 +1,56 @@
...
+        $parents = [];
+        $parentsSlugs = $theme->getParentsSlugs();
+        foreach ($parentsSlugs as $parentName) {
+            $parent = $this->themeRepository->findOneBySlug($parentName);
+
+            if (null === $parent) {
+                throw new \InvalidArgumentException(sprintf(
+                    'Theme "%s" not found (required by theme "%s")!', $parentName, $theme->getSlug()
+                ), 0);
+            }
```
Zales0123: I would extract this to private method (``validateParent``? ``checkParentExistance``? anything? :smile:) to avoid nesting and improve readability.

src/Sylius/Bundle/ThemeBundle/Locator/ResourceLocatorInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ */
+interface ResourceLocatorInterface
+{
+    /**
+     * @param string $resourceName
+     * @param ThemeInterface $theme
+     *
+     * @return string
+     *
+     * @throw ResourceNotFoundException
```
Zales0123: ``@throws``

src/Sylius/Bundle/ThemeBundle/Locator/ResourceNotFoundException.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Sylius\Bundle\ThemeBundle\Locator;
```
Zales0123: Missing license block.

src/Sylius/Bundle/ThemeBundle/Tests/Functional/Bundle/TestBundle/Resources/config/routing.yml
```diff
@@ -0,0 +1,5 @@
+template:
+    path:     /template/{templateName}
+    defaults: { _controller: TestBundle:Templating:renderTemplate }
+    requirements:
+        templateName: .+
```
Zales0123: Missing blank line :smile:

Zales0123: Here and in many yml/twig files below.

src/Sylius/Bundle/ThemeBundle/Tests/Functional/Bundle/TestBundle/TestBundle.php
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+namespace Sylius\Bundle\ThemeBundle\Tests\Functional\Bundle\TestBundle;
+
+use Symfony\Component\HttpKernel\Bundle\Bundle;
+
+class TestBundle extends Bundle
+{
+
+}
```
Zales0123: And here.

src/Sylius/Bundle/ThemeBundle/Tests/Functional/Bundle/TestBundle/TestBundle.php
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\ThemeBundle\Tests\Functional\Bundle\TestBundle;
+
+use Symfony\Component\HttpKernel\Bundle\Bundle;
+
+class TestBundle extends Bundle
```
Zales0123: Who is the author of this class? :(

src/Sylius/Bundle/ThemeBundle/Tests/Functional/ThemeBundleTestCase.php
```diff
@@ -0,0 +1,51 @@
...
+            // Don't insulate requests if not possible to do so.
+        }
+
+        return $client;
+    }
+}
```
Zales0123: Blank line wanted.

src/Sylius/Bundle/ThemeBundle/Asset/PathResolver.php
```diff
@@ -0,0 +1,28 @@
...
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+final class PathResolver implements PathResolverInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function resolve($path, ThemeInterface $theme)
+    {
+        return str_replace('bundles/', 'bundles/_theme/' . $theme->getSlug() . '/', $path);
```
pjedrzejewski: Are we sure about using ``Slug`? Perhaps we should use ``Code``?

src/Sylius/Bundle/ThemeBundle/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+CHANGELOG
+=========
+
+### v0.11.0
```
pjedrzejewski: Trolololol...

src/Sylius/Bundle/ThemeBundle/Command/SyliusThemeDebugCommand.php
```diff
@@ -0,0 +1,61 @@
...
+ */
+class SyliusThemeDebugCommand extends ContainerAwareCommand
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('sylius:theme:debug')
```
pjedrzejewski: :+1:

src/Sylius/Bundle/ThemeBundle/Filesystem/FilesystemInterface.php
```diff
@@ -0,0 +1,189 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\ThemeBundle\Filesystem;
+
+use Symfony\Component\Filesystem\Exception\FileNotFoundException;
+use Symfony\Component\Filesystem\Exception\IOException;
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
```
pjedrzejewski: I believe there is some copy paste here from Symfony, let's add original authors too. :)

src/Sylius/Bundle/ThemeBundle/Filesystem/FilesystemInterface.php
```diff
@@ -0,0 +1,189 @@
...
+ */
+interface FilesystemInterface
```
pjedrzejewski: Maybe we should consider extending the filesystem interface here and only add methods which are needed in our case?

pamil: The thing is that Symfony does not provide any `FilesystemInterface`, but just bare class.

src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstaller.php
pamil: Done.

phpunit.xml.dist
```diff
@@ -22,10 +22,8 @@
         <whitelist>
             <directory>src</directory>
             <exclude>
-                <directory>src/*/*Bundle/Resources</directory>
-                <directory>src/*/*Bundle/Tests</directory>
-                <directory>src/*/Bundle/*Bundle/Resources</directory>
-                <directory>src/*/Bundle/*Bundle/Tests</directory>
+                <directory>src/Sylius/Bundle/*Bundle/Resources</directory>
```
lchrusciel: This changes shouldn't be part of this PR

src/Sylius/Bundle/ThemeBundle/.travis.yml
```diff
@@ -0,0 +1,12 @@
+language: php
+
+php:
+  - 5.5
+  - 5.6
+  - 7.0
+
+before_script: composer install --prefer-source --no-interaction
```
pjedrzejewski: I think we should adjust it to match other .travis.yml files - https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ApiBundle/.travis.yml#L8

(we can skip the memory if not needed)

src/Sylius/Bundle/ThemeBundle/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+CHANGELOG
+=========
+
+### v0.16.0
```
pjedrzejewski: 17

src/Sylius/Bundle/ThemeBundle/Command/SyliusThemeDebugCommand.php
```diff
@@ -0,0 +1,61 @@
...
+use Sylius\Bundle\ThemeBundle\Model\ThemeInterface;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Helper\Table;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+class SyliusThemeDebugCommand extends ContainerAwareCommand
```
pjedrzejewski: We never prefix commands with Sylius.

src/Sylius/Bundle/ThemeBundle/Command/SyliusThemeDebugCommand.php
```diff
@@ -0,0 +1,61 @@
...
+        $output->writeln("<question>Succesfully loaded themes:</question>");
+
+        $table = new Table($output);
+        $table->setHeaders(['Name', 'Slug', 'Path']);
+
+        foreach ($themes as $theme) {
+            $table->addRow([$theme->getName(), $theme->getSlug(), $theme->getPath()]);
+        }
+
+        $table->setStyle("borderless");
```
pjedrzejewski: Use single quotes.

src/Sylius/Bundle/ThemeBundle/Command/SyliusThemeDebugCommand.php
pamil: It was prefixed to distinguish it from `AssetsInstallCommand` which overrides the default `assets:install` command.

src/Sylius/Bundle/ThemeBundle/.travis.yml
pamil: Sure thing. In fact, we need to refactor our `.travis.yml` files as installing Composer dependencies or disabling XDebug should be included in `install` / `before_install` steps.

src/Sylius/Bundle/ThemeBundle/Command/SyliusThemeDebugCommand.php
pjedrzejewski: I'd say we don't need to do this. It just has to be well documented in the docs that we override standard symfony command. As you can see, I could not figure it out anyway. :P

src/Sylius/Bundle/ThemeBundle/Resources/meta/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2011-2015 Paweł Jędrzejewski
```
pjedrzejewski: 2016

src/Sylius/Bundle/ThemeBundle/Command/SyliusThemeDebugCommand.php
pamil: Anyway, I think there is an issue with installing assets and we shouldn't override the default command. Our command should only install assets for themes, leaving regular assets installing to the corresponding Symfony command.

pjedrzejewski: Okay, I  think it makes sense.

pjedrzejewski:
WOOO! Awesome work Kamil and thank you everyone for the feedback! :+1: :clap: 

Notes:

* This is a bit experimental, it is well tested in works but we are curious to see how it works out for other use cases. We want to improve it a lot with support for Gulp and NPM deps for Themes etc. but that is for further iterations.
* If you get any trouble with theming, just remove it from the Kernel and everything should be back to normal - yes, this is totally decoupled from Sylius. We will definitely make a blog post because other Symfony projects may find it useful.

Issue #2846, opened 2015-06-12
[ResourceBundle] The paginate parameter can not be overwritten in query string


cdaguerre:
Controller `Configuration` checks the `params` set by the `onKernelController` listener and general settings but never checks the query.

I'm not sure this is expected behaviour, but if it is, I think we could introduce a `paginatable` setting...

aRn0D:
I don't understand can you explain us a bit more, plz ?

pjedrzejewski:
Will be fixed in #2255 as well.

PR #2848, opened 2015-06-12
[ResourceBundle] Check query for paginate parameter

cdaguerre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2846
| License       | MIT

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/Configuration.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/Configuration.php
@@ -201,7 +201,7 @@ class Configuration
         $redirect = $this->parameters->get('redirect');
 
         if (!is_array($redirect) || empty($redirect['hash'])) {
-            return null;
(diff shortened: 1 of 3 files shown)


cdaguerre:
However this only works, for integer paginate query parameters because any string will be cast to `true` (even "false").

To avoid this, the following code would work:

```php
    public function isPaginated()
    {
        return (bool) null === filter_var(
            $this->getPaginationMaxPerPage(),
            FILTER_VALIDATE_BOOLEAN,
            FILTER_NULL_ON_FAILURE
        );
    }
```

2 problems with this:
First of all, it's horrible unintelligible code. 
And, request parameter conversion should not be a concern to the controller configuration.

I'll open an other ticket with respect to this issue and reference it here.

cdaguerre:
Closed

PR #2859, opened 2015-06-12
Update FrontendMenuBuilder.php

### LINKED ISSUE #2858: Add "current" class for currently selected menu item in FrontendMenuBuilder
A simple `$menu->setCurrentUri($this->request->getRequestUri());` should do the trick. The same method is also used inside the `BackendMenuBuilder` class. This will add a "current" class to the currently selected menu item, so that some additional CSS styling can be applied.

liviucmg:
Fixes #2858


pjedrzejewski:
Thank you Liviu! :+1:

PR #2861, opened 2015-06-13
[MailerComponent] Removing unused check, Adding more specs to Email model

mgonyan:
| Q | A |
---- | -----
| Bug fix?	| no |
| New feature?	| refactoring |
| BC breaks?	| no |
| Deprecations? | no |
| Fixed tickets	| no |
| License	| MIT |
| Doc | PR |

- Adding specs for Email model
- Removing unused check for email provider

### DIFF
--- a/src/Sylius/Component/Mailer/Provider/EmailProvider.php
+++ b/src/Sylius/Component/Mailer/Provider/EmailProvider.php
@@ -59,10 +59,6 @@ class EmailProvider implements EmailProviderInterface
             $email = $this->getEmailFromConfiguration($code);
         }
 
-        if (null === $email) {
-            throw new \InvalidArgumentException(sprintf('Email with code "%s" does not exist!', $code));
-        }
-
         return $email;
     }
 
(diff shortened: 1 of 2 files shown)


src/Sylius/Component/Mailer/Provider/EmailProvider.php
```diff
@@ -59,10 +59,6 @@ public function getEmail($code)
             $email = $this->getEmailFromConfiguration($code);
         }
 
-        if (null === $email) {
```
aRn0D: Why ?

mgonyan: Because at the method getEmailFromConfiguration is already doing that check. If it could not find the template code, it would throw an exception so i recon this checking is redundant. am i wrong?

michalmarcinkowski: That is correct.

michalmarcinkowski:
We should avoid fluent interfaces unless there is a really good reason for that. Related article http://ocramius.github.io/blog/fluent-interfaces-are-evil/.

mgonyan:
@michalmarcinkowski Totally agree with you after i reading the post, many thanks. I've amended the code.

michalmarcinkowski:
:+1:

pjedrzejewski:
Thank you Manuel! :+1:

PR #2886, opened 2015-06-17
[WIP][CoreBundle] Add missing data_class to shipping method choice type

aramalipoor:
Q  | A
------------- | -------------
Bug fix?  | yes
New feature?  | no
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

The case is when user abandons the checkout process (I think after a `Shipment` entity is created for the `Order` and is stored on its `shipments` collection) and then that user tries to complete the checkout, `$order` contains a shipment in its `shipments` and its `$method` is a real object of selected `ShippingMethod` so when form tries to set the data it does not expect an object! Because it's a `choice`.

I don't know how can I cover this case in tests and also I don't know what other places are affected by this change.


aramalipoor:
I think that was a impossible case because of some development environment mess! So I'm just closing for now.

PR #2899, opened 2015-06-19
fixes API routes for product_variant actions

Amunak:
ProductVariant has no field "product" - it has "object". The "id"
criteria was useless since it is ignored by the API route controller
which takes the parameter directly.

Fixes this error:

    [Semantical Error] line 0, col 112 near 'product = :product': Error: Class ...\ProductVariant has no field or association named product

It appeared when accessing `/api/products/<productId>/variants/<variantId>`.


pjedrzejewski:
Thank you @Amunak! :+1:

PR #2908, opened 2015-06-22
Use submit on forms instead of handleRequest

ahmadrabie:
Q  | A
------------- | -------------
Bug fix?  | yes
New feature?  | no
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

Sometimes when you use `handleRequest` form will not be valid. (`isValid()` returns false)


michalmarcinkowski:
From Symfony docs (http://symfony.com/doc/current/cookbook/form/direct_submit.html#cookbook-form-submit-request):
```
Passing the Request directly to submit() still works, but is deprecated and will be removed in Symfony 3.0. You should use the method handleRequest() instead.
```

``` isValid() ``` method validates the object and if it's not valid, it will return ``` false ```.

PR #2255, opened 2014-12-16
[WIP] SyliusResourceBundle Reloaded

pjedrzejewski:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #1764 #2232 #2223
| License       | MIT
| Doc PR        | -

**Do NOT review the code! It is a quick POC.**

Hey folks!

These changes are partially inspired by another bundle (inspired by Sylius resource, lol), but I could not find it on github/packagist... Can be that it is gone, I have looked at it long time ago and it inspired me to work on this.

I wanted to do it for a long time and seems like it is a last moment to introduce a BC break of this size in Sylius fundamentals. It is an alternative approach to @stloyd's #1764 DomainManagers and effectively decouples Sylius components from the Doctrine Persistence interfaces, while staying very close to them, so that any Symfony developer will be used to the same API. Overview of my proposed changes:

* Introducting **ResourceManagerInterface**.

Every resource gets it's own dedicated resource manager. Additionally, it has this EvenfulResourceManager, which dispatches our beloved events.

Also, I have made it a bit smarter, and it has only save() and delete() methods. Save will know if it is a new or existing resource and dispatch an appropriate event. (update or create)

* Decouple the "createNew" from Repository interface and move it to dedicated **ResourceFactoryInterface**. The reason is that in Sylius we often want several ways to create a resource, for example create variant for product, create taxon for taxonomy and I found factories really useful when working on multi-stock inventory (top secret, yet). I believe dedicated factory is  the right place, as opposed to putting it on the "DomainManager" in original PR from Joseph. (great work, just saying)

* Introduce ResourceEventDispatcherInterface. We have our resource events and I was tired of passing the strings around, to dispatch appropriate event name, now every resource gets a dedicated event dispatcher which can be used more easily.

Do not look at the resource controller yet, it needs some work and better flashes handling and so on. I will move them to a listener and generally, the controller will have all the previous functionality.

**IMPORTANT NOTE**
This change requires us to move to PHP 5.4 (ResourceInterface::getId). With recent decisions about PHP version for Symfony 3.0, I'm more than convinced to move to 5.4, especially, that we could do a good use of traits in many places. (in few we could hide design flaws too... by making our anemic models smaller in terms of LOC... joke)

Let me know what do you think.

PR #2910, opened 2015-06-22
[SettingsBundle] Use find() in ObjectToIdentifierTransformer

ahmadrabie:
Q  | A
------------- | -------------
Bug fix?  | yes
New feature?  | no
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

In Phpcr document repository `findOneBy` will look for properties and not document `id` or `path`. So if `ObjectToIdentifierTransformer`'s job is find the object by real document/entity (not user provided) identifier it's more universal to use `find()` method.

Anyway right now `ObjectToIdentifierTransformer` does not work with [Phpcr document repository](https://github.com/doctrine/phpcr-odm/blob/master/lib/Doctrine/ODM/PHPCR/DocumentRepository.php).

### DIFF
--- a/src/Sylius/Bundle/SettingsBundle/Transformer/ObjectToIdentifierTransformer.php
+++ b/src/Sylius/Bundle/SettingsBundle/Transformer/ObjectToIdentifierTransformer.php
@@ -42,6 +42,10 @@ class ObjectToIdentifierTransformer extends BaseTransformer implements Parameter
             return;
         }
 
+        if ('id' === $this->identifier) {
+            return $this->repository->find($value);
+        }
+
         return $this->repository->findOneBy([$this->identifier => $value]);
     }
 }


ahmadrabie:
@pjedrzejewski ping!

ahmadrabie:
@pjedrzejewski ping!

stloyd:
:-1: this should be adjusted for this specific case like Phpcr but not reverted to the `find()` method.

pjedrzejewski:
Yeah, I think we need to find a good and clean solution for PHPCR. With this PR, the behavior would change. I'd leave that after #2255 is merged, so for v0.16.0.

ahmadrabie:
@pjedrzejewski Now that you refactored `SyliusResourceBundle` and closed #2255 can you think about this PR?

pjedrzejewski:
We cannot assume that identifier is ``id``, we should add an if there.

ahmadrabie:
@pjedrzejewski Sorry for the late response.

pjedrzejewski:
Thanks Ahmad! 👍  No worries!

PR #2945, opened 2015-06-30
[CoreBundle] Decouple original locale provider and channel-aware locale provider

### LINKED ISSUE #2944: Use LocaleProvider instead of ChannelAwareLocaleProvider in admin panel
My problem is that when I want to edit my product I just see available locales in current channel in translations section, which is not the expected behavior IMO, how do you think it's possible to use LocaleProvider in `a2lix_translationsForms` only when in admin panel?

If any of you has any idea I can happily contribute :)

aramalipoor:
Q  | A
------------- | -------------
Bug fix?  | no
New feature?  | no
BC breaks? | no
Deprecations? | no
Fixed tickets | Fixes #2944
License | MIT
Doc PR | -

This change allows the admin to see all enabled locales (not only current channel available ones) when editing translations. 

Translations form type is using [LocaleBundle provider](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Resources/config/app/main.yml#L160) so it selects all enabled locales.

### DIFF
--- a/src/Sylius/Bundle/LocaleBundle/Controller/LocaleController.php
+++ b/src/Sylius/Bundle/LocaleBundle/Controller/LocaleController.php
@@ -11,12 +11,12 @@
 
 namespace Sylius\Bundle\LocaleBundle\Controller;
 
-use Sylius\Bundle\CoreBundle\Locale\ChannelAwareLocaleProvider;
-use Sylius\Bundle\ResourceBundle\Controller\ResourceController;
-use Sylius\Component\Locale\Context\LocaleContext;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Sylius\Component\Locale\Provider\LocaleProviderInterface;
+use Sylius\Component\Locale\Context\LocaleContextInterface;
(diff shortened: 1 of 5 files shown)


michalmarcinkowski:
:+1:

aramalipoor:
@pjedrzejewski ping!

src/Sylius/Bundle/LocaleBundle/Controller/LocaleController.php
```diff
@@ -65,6 +65,6 @@ protected function getLocaleContext()
      */
     protected function getLocaleProvider()
     {
-        return $this->get('sylius.locale_provider');
+        return $this->get('sylius.channel_aware_locale_provider');
```
pjedrzejewski: Locale bundle should not know about channels. This should be core specific customization.

pjedrzejewski:
@aramalipoor See my comment. :)

src/Sylius/Bundle/LocaleBundle/Controller/LocaleController.php
aramalipoor: You're right. I'm on it ;)

aramalipoor:
@pjedrzejewski Done. ;)

pjedrzejewski:
We should make it configurable perhaps in bundle cfg later. Thank you Aram! :+1:

PR #2225, opened 2014-12-06
Create Archetype Component and Bundle to replace Product Prototype and allow for inheritance

adamelso:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2001
| License       | MIT
| Doc PR        | pending

The new Archetype component and bundle is designed to replace the existing Product Prototypes, introducing inheritance between archetypes, and to be abstract, so any model other than a product can be the subject of archetyping.

image

I'll probably need some assistance with resolving the interfaces to the correct Doctrine mapping, or whether my implement is correct or not.

- [x] Create Archetype component
- [x] Pass specification tests
- [x] Create Archetype bundle
- [x] Replace Product Prototype with Archetype
    - [x] Update services
    - [x] Update Doctrine mapping
    - [x] Update validation
    - [x] Update translation fields
    - [x] Map parent and child via LoadMetadataSubscriber instead of XML to make mapping valid
- [x] Pass feature tests

PR #2948, opened 2015-06-30
[ArchetypeBundle] Add ON DELETE CASCADE constraint on archetype

ahmadrabie:
Q | A
------------- | -------------
Bug fix? | yes
New feature? | no
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

When I want to delete a product option, because of the foreign key constraint of `CASCADE` all related records will also get deleted from tables below:
- sylius_product_option
- sylius_product_option_translation
- sylius_product_option_value

but product option relation to product archetype (`sylius_product_archetype_option`) does not have a `CASCADE` constraint (It's currently `RESTRICT`) so a constraint violation occurs.

Is it correct to add `CASCADE ON DELETE` relation constraint?

### DIFF
--- a/src/Sylius/Bundle/ArchetypeBundle/EventListener/LoadMetadataSubscriber.php
+++ b/src/Sylius/Bundle/ArchetypeBundle/EventListener/LoadMetadataSubscriber.php
@@ -60,7 +60,7 @@ class LoadMetadataSubscriber implements EventSubscriber
                 continue;
             }
 
-            $this->mapVariations($metadata, $class, $subject);
+            $this->mapOptions($metadata, $class, $subject);
             $this->mapAttributes($metadata, $class, $subject);
             $this->mapParent($metadata, $class);
         }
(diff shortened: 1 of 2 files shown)


adamelso:
Thanks Ahmad :+1: 

Interestingly I had considered this when the feature was in development: https://github.com/Sylius/Sylius/pull/2225#discussion-diff-21792776R95

Also as a result, [disabling foreign key checks](https://github.com/Sylius/Sylius/blob/5ab0638d2917ad333ff12a1f68115d7c3d963e1a/src/Sylius/Bundle/CoreBundle/Behat/HookContext.php#L49) can also be removed.

pjedrzejewski:
You are right Ahmad! :+1: Thank you!

pjedrzejewski:
@ahmadrabie @adamelso We will have a look at removing these checks in a separate PR.

PR #2949, opened 2015-06-30
Make choice form types API friendly

aramalipoor:
Q  | A
------------- | -------------
Bug fix?  | no
New feature?  | refactor
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Form/Type/ProvinceChoiceType.php
+++ b/src/Sylius/Bundle/AddressingBundle/Form/Type/ProvinceChoiceType.php
@@ -50,10 +50,10 @@ class ProvinceChoiceType extends AbstractType
 
         $choiceList = function (Options $options) use ($repository) {
             if (null === $options['country']) {
-                return new ObjectChoiceList($repository->findAll());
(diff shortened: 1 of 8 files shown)


pjedrzejewski:
Maybe we can use ``sylius_*_choice`` types instead of entities? Choice type knows about the class already. What do you think?

aramalipoor:
@pjedrzejewski the problem is when the developer tries to create a resource by API he only knows about the ID (e.g. Option ID when creating an Archetype) but if we use choice types, symfony choice type expects a numeric index based on ChoiceList or inline `choices` which developer doesn't have any clue about it when POSTing.

I don't if it's possible to tweak choice types in a way so that the value expecting from the POST would be the ID of the entity, then we can use choice types both for API and regular forms. What do you think?

aramalipoor:
Oh my god, I didn't notice that ResourceChoiceType is actually extending `entity` type. I think you're right, we can use them. Let me check ;)

aramalipoor:
@pjedrzejewski I've fixed some of the choices now all my API calls work. Hope all entities are API friendly now.

pjedrzejewski:
Thank you Aram! :+1:

PR #2951, opened 2015-06-30
[ResourceBundle] Handle foreign-key constraint fails when deleting an object

### LINKED ISSUE #2950: Foreign key constraint violation when deleting some entities (e.g. Zone)
When the user tries to delete a Zone, Country, Province or Address, because of the foreign key constraints he's going to see a ForeignKeyConstraintViolation exception. Which in many cases confuses regular users.

What do you think is the best way to handle this situation?

Some suggestions are as follows:
* Show a user friendly message which warns user that this entity is being used.
* Allow him to provide a `force` or `yes-to-delete-all` kind of option so we will remove all related entities.
* Use `SET NULL` constraint in some situations, `CASCADE DELETE` is some others.

If we want to show a human-readable message (instead of the scary exception) where is the correct place to catch the exception? I think catching `PDOException`s with [error code `1451`](https://dev.mysql.com/doc/refman/5.5/en/error-messages-server.html#error_er_row_is_referenced_2) (which is for foreign key constraints on update and delete) in resource controller `deleteAction`, can be a good idea.

What do you guys think? :)

ahmadrabie:
Q | A
------------- | -------------
Bug fix? | yes
New feature? | no
BC breaks? | no
Deprecations? | no
Fixed tickets | Fixes #2950
License | MIT
Doc PR | -

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/DomainManager.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/DomainManager.php
@@ -12,9 +12,11 @@
 namespace Sylius\Bundle\ResourceBundle\Controller;
 
 use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException;
 use Sylius\Component\Resource\Event\ResourceEvent;
 use Symfony\Component\EventDispatcher\Event;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\HttpKernel\Exception\HttpException;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 
 /**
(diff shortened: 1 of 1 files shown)


michalmarcinkowski:
It is not the proper way to do it. If we don't want to allow for deleting object with associations we should add listener for that, instead of depending on DBAL exception.

aramalipoor:
@michalmarcinkowski you're right that would be better. Which events are a good fit here? Domain model events (`sylius.zone.pre_delete`) or Doctrine and ORM events?

michalmarcinkowski:
@aramalipoor model events (``` sylius.zone.pre_delete ```) are only dispatched in the controller, so if we want to be sure, that the object will never be deleted (anywhere in the code) without this check, we should depend on Doctrine events.

ahmadrabie:
@michalmarcinkowski I'm not a Doctrine expert but from what I get from you, you say that we listen to Doctrine `preRemove` event and somehow read the entity's metadata and its associations then throw an appropriate exception if that association exists. Am I right?

pjedrzejewski:
See my comment in #2950, we need to come up with a different solution.

Issue #2915, opened 2015-06-23
Order confirmation email - possibly on the wrong event


inssein:
I was just reading the docs for emails (http://docs.sylius.org/en/latest/book/emails.html), and it mentions that order confirmation email is sent after order has been paid for.

However, it clearly seems like the order confirmation email is sent in the finalize step, and not after the purchase step.

Is the documentation incorrect, or is this a bug? And if it's a bug, is the fix as easy as changing the event name inside the mailer xml? (from ```<tag name="kernel.event_listener" event="sylius.checkout.finalize.complete" method="sendOrderConfirmationEmail" />``` to ```<tag name="kernel.event_listener" event="sylius.checkout.purchase.complete" method="sendOrderConfirmationEmail" />```

If it is a bug, and my solution above is correct, I could send a PR right away.

kayue:
Shouldn't it use state machine to send email?

inssein:
@kayue do you have an example of that already happening somewhere else? Right now its just https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/EventListener/MailerListener.php, wired up by https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Resources/config/mailer.xml.

inssein:
So I just replaced the event with a compiler pass for now, until we figure it out that its a bug and it gets fixed in Sylius. The only thing to note is that ```sylius.checkout.purchase.complete``` does not work because it sends a PurchaseCompleteEvent with a Payment. I resorted to using the ```sylius.checkout.purchase.pre_complete``` for now, which works.

kayue:
@inssein Sorry for the late reply.

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Resources/config/state-machine.yml#L110-L121

To send order confirmation email is sent after order has been **paid** for, it should be listen to the `confirm` transition. 

Personally I would avoid hooking into steps event, but I think we have been doing so for some time.

inssein:
@kayue ah interesting, didn't know about that, still got lots to learn. I guess the order confirmation email functionality must have been written before addition of a state machine. Perhaps next weekend I could put in an RFC to upgrade the email event listeners to use the state machine instead.

pjedrzejewski:
@inssein That would be great! We should move everything to state machine.

kayue:
:+1:

michalmarcinkowski:
[...]

Issue #2514, opened 2015-02-24
Confirmation order email not sending


dslipcev:
Hi,

I have problems with sending confirmation order emails. Swiftmailer is configured and working normally. Sylius version is 0.14.

Sylius.yml (in SyliusCoreBundle) contains the configuration for emails;
  ```
sylius_mailer:
    templates:
        sylius.email.order_confirmation.name: SyliusWebBundle:Email:orderConfirmation.html.twig
        sylius.email.order_comment.name: SyliusWebBundle:Email:orderComment.html.twig
        sylius.email.user_confirmation.name: SyliusWebBundle:Email:userConfirmation.html.twig
    sender:
        name: Potato.com
        address: no-reply@potato.com
    emails:
        enabled: true
        order_confirmation:
            enabled: true
            subject: sylius.emails.order_confirmation.subject
            template: SyliusWebBundle:Email:orderConfirmation.html.twig
        order_comment:
            subject: sylius.emails.order_comment.subject
            template: SyliusWebBundle:Email:orderComment.html.twig
        user_confirmation:
            subject: sylius.emails.user_confirmation.subject
            template: SyliusWebBundle:Email:userConfirmation.html.twig
```

What am I doing wrong?

Thanx.

ARoddis:
It seems that the mailer.xml is not registered anymore...i did it in my bundle and it works as expected.

PR #2959, opened 2015-07-04
[CoreBundle] Use state machine for order and shipment email confirmation

inssein:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | https://github.com/Sylius/Sylius/issues/2915, https://github.com/Sylius/Sylius/issues/2514
| License       | MIT
| Doc PR        | 

Pretty simple PR, just moves the order and shipment confirmation email listeners to use the state machine instead of the event listener. In that process, it also fixes a bug. (Order email confirmation listener was bound to the wrong event).

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/DependencyInjection/SyliusCoreExtension.php
+++ b/src/Sylius/Bundle/CoreBundle/DependencyInjection/SyliusCoreExtension.php
@@ -63,7 +63,7 @@ class SyliusCoreExtension extends AbstractResourceExtension implements PrependEx
         'templating.xml',
         'twig.xml',
         'reports.xml',
-        'mailer.xml',
+        'email.xml',
     );
 
     /**
--- a/src/Sylius/Bundle/CoreBundle/EmailManager/OrderEmailManager.php
+++ b/src/Sylius/Bundle/CoreBundle/EmailManager/OrderEmailManager.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\CoreBundle\EmailManager;
+
+use Sylius\Bundle\CoreBundle\Mailer\Emails;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Mailer\Sender\SenderInterface;
+
+/**
+ * Sends the order confirmation email.
+ *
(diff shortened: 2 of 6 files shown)


inssein:
ping @kayue @pjedrzejewski - sent the pr according to what I said in https://github.com/Sylius/Sylius/issues/2915.

kayue:
Looks good to me :+1:

inssein:
Actually, the shipment confirmation email is not working, and I don't know why. @kayue would you be able to tell by looking at the commit? It's exactly the same as the callback above it, but doesn't get fired.

The only way I was able to get a callback, was to put a callback on the "ship" state of the order. I didn't want to go with this method because it add's event more backwards incompatible change as I can't provide the shipment to the email.

pjedrzejewski:
I'm thinking that we should avoid naming the services "Callbacks". Let's give it a more meaningful name, cause it can be used not only as a callback. I think SM should use appropriate services as callbacks and that's it. What do you think about that guys? I want to move pretty much everything to SM soon, so this is important. :) Thanks for your input!

inssein:
@pjedrzejewski that is what I was thinking earlier as well.

I initially wrote a "MailerManager" that sends out all the emails, just like the current MailerListener does, but it didn't make sense as this file could get big pretty fast, and it would be sending a lot of unrelated emails.

Instead, I broke it up the "MailerManager" per component, and currently have a ShipmentEmailCallback and OrderEmailCallback. Would it suffice to rename the suffix to "Manager" from "Callback", and put them somewhere more appropriate?

inssein:
@pjedrzejewski you get a chance to look at the issue I posted about before your comment?

inssein:
@kayue nevermind, figured it out. silly mistake.

pjedrzejewski:
@inssein I think this is a good idea! Just use EmailManager namespace for these, to clarify that it is related to mailing. :+1:

inssein:
@pjedrzejewski done. The only other thing I did is move the services definition from "state_machine.xml" to "mailer.xml", and then renamed "mailer.xml" to "email.xml".

inssein:
@pjedrzejewski not sure why tests failed.. seems very unrelated to what I have committed.

inssein:
ping @pjedrzejewski - is there something I am missing or is this good to go?

pjedrzejewski:
@inssein Could you amend your last commit and add a bit more informative message, please? Then it will be perfect, thanks!

inssein:
@pjedrzejewski rebased.

michalmarcinkowski:
Thank you Hussein! :+1: 
I think it resolves also https://github.com/Sylius/Sylius/issues/2514

pjedrzejewski:
Thank you Hussein!

PR #2963, opened 2015-07-05
[TranslationBundle] TranslatableResourceRepository->find() yield incorrect counts on multilang setups

Im0rtality:
| Q | A |
| --- | --- |
| Bug fix? | yes (partial fix) |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | n/a |
| License | MIT |
| Doc PR | n/a |

`TranslatableResourceRepository` selects all translations instead of one, thus reducing offer counts. See attached screenshots for front page "productRepo->findLatest(9)" call (cleaned up doctrine):

before fix:
screenshot 2015-07-05 20 55 31

after fix:
screenshot 2015-07-05 21 05 53

Problem - does not respect `fallback` locale.


pamil:
`TranslatableResourceRepository` got removed.

PR #2976, opened 2015-07-07
Enable header redirection in RedirectHandler

winzou:
Q | A
------------- | -------------
Bug fix? | no
New feature? | kind of
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

Add the ability to intercept a redirection. Use case: an xhr request. In javascript it is not possible to catch a redirection, so this way enable the JS to get this HTTP 200 response, and read the defined `X-SYLIUS-LOCATION` header, so that it can decide to redirect or not.


pjedrzejewski:
Really cool, thank you Alexandre! :+1:

PR #2987, opened 2015-07-10
Fix country sorting

winzou:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Fix the exception on `sylius_backend_country_index` page:

    RuntimeException: Cannot select distinct identifiers from query with LIMIT and ORDER BY
    on a column from a fetch joined to-many association. Use output walkers.

cf. http://demo.sylius.org/administration/countries/


michalmarcinkowski:
:+1:

pjedrzejewski:
Thank you Alexandre!

Issue #2985, opened 2015-07-09
PromotionType fields are required updating


rvanlaak:
I'm not using the full Sylius stack, but only make use of the `PromotionBundle` and it's dependencies (`Registry`, `MoneyBundle`, `Resource`, `Storage`).

Currently we use `0.14.0` but since that update the form fields of all types (eg. `PromotionType`, `ActionType`, `RuleType`) are set to required again.

The `PromotionType` itself does not seem to be changed, so why are the fields set to required again?
https://github.com/Sylius/Sylius/commits/master/src/Sylius/Bundle/PromotionBundle/Form/Type/PromotionType.php

Because I override the `PromotionType` I can solve this by getting al formFields manually, but this looks like a bug to me:

```php

public function buildForm(FormBuilderInterface $builder, array $options)
{
	parent::buildForm($builder, $options);

	$startsAt = $builder->get('startsAt');
	$options = array_merge($startsAt->getOptions(), array('required' => false));
	$builder->add('startsAt', 'datetime', $options);
    // ...

```

Tried to narrow this down to a specific commit, but unfortunately was not able to do this. Did already exclude the upgrade from symfony 2.6.4 to 2.7.1 as the cause of the problem. The problem was introduced somewhere between the following an `0.14.0`:

    "sylius/resource-bundle": "dev-master#ada589b",

Extending the type and globally setting required to false was enough before:

```php
    public function configureOptions(OptionsResolver $resolver)
    {
        parent::configureOptions($resolver);

        $resolver->setDefaults(array(
	        'required' => false,
        ));
    }
```

Rvanlaak:
The problem lies in the new `configureOptions` that was introduced in `2.7`. Renaming it back to `setDefaultOptions` solves this.

pjedrzejewski:
Sylius now requires 2.7.

rvanlaak:
Great! Why don't support the 2.8 LTS version, which API should be the same?

pjedrzejewski:
2.7 is minimum, should work with 2.8. :)

PR #2989, opened 2015-07-10
Fix FormType options not being passed when using symfony 2.7

Rvanlaak:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | yes
| Fixed tickets | https://github.com/Sylius/Sylius/issues/2985
| License       | MIT
| Doc PR        | 

Fixes the options after upgrading to Symfony 2.7. Currently the options can not be used.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Form/Type/AbstractResourceType.php
+++ b/src/Sylius/Bundle/ResourceBundle/Form/Type/AbstractResourceType.php
@@ -12,7 +12,7 @@
 namespace Sylius\Bundle\ResourceBundle\Form\Type;
 
 use Symfony\Component\Form\AbstractType;
-use Symfony\Component\OptionsResolver\OptionsResolverInterface;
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
We are still compatible with Symfony <2.7. We will do the upgrade soon and require at least Symfony 2.7 so this will be solved. Your solution won't work on lower versions. We will fix it in all forms at once.

Rvanlaak:
:+1: any idea when you're going to support the new LTS version?

Symfomany:
In form Types, setDefaultOptions is deprecated, use configureOptions to be compatible with Symfony 3

Rvanlaak:
@Symfomany see yesterday's comment

Symfomany:
https://github.com/Sylius/Sylius/issues/3002 Sorry... :)

PR #2991, opened 2015-07-11
[AttributeBundle] Product attribute value is not API friendly

ahmadrabie:
Q | A
------------- | -------------
Bug fix? | yes
New feature? | no
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

In product attribute value form, there an event listener which tries to remove `attribute` field. But when we're using this form in APIs removing this field is going to raise some issues. (Form serialization, extra fields, etc)

So if removing this field is only for a better view, IMO it's better to keep than make the form unsable in some situations.

What do you think?


pjedrzejewski:
I agree! For backend UI we will use different stuff so this makes sense. :+1: Thank you Ahmad!

PR #2992, opened 2015-07-11
[AttribtueBundle] Set correct value data on sylius_*_attribute_value form

ahmadrabie:
Q | A
------------- | -------------
Bug fix? | yes
New feature? | no
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

Because of [this portion](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Attribute/Model/AttributeValue.php#L110-L115) of the code, `value` field is going to be filled with choice's presentation (not index) which is the desired behavior for a lot of UI situations, but not when setting the form's data, which expects the real data. (Current choice's index) 

I don't like adding this new method, but is there any better way I'm missing?


michalmarcinkowski:
Will be fixed by https://github.com/Sylius/Sylius/pull/3608

PR #2993, opened 2015-07-13
Add extension point to add template data

winzou:
Q | A
------------- | -------------
Bug fix? | no
New feature? | no
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

Update FOSRestBundle to have the setTemplateData method, and add an extension point to add template data without overriding the whole action in ResourceController.

Side modification: in create/update action, do not set the form template data as it is automatically added by the ViewHandler cf [ViewHandler.php#L384-385](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/View/ViewHandler.php#L384-385)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1,10 +1,10 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at http://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "308a982c7cd2939d2ae9d778a0e87a21",
+    "hash": "b6d17dfe5edbd8dd93885622a0c0af64",
     "packages": [
         {
             "name": "a2lix/translation-form-bundle",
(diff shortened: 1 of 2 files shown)


winzou:
OK, I removed my side modification and build is green. Ready to merge.

Not sure what you want to do, but it seems that we don't use FOSRestBundle the right way:
  * When using a form, we should not pass data this way: [ResourceController.php#L209-L212](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php#L209-L212)
  * Instead, we should just pass the form as the data, the VewHandler does the rest: [ViewHandler.php#L384-L385](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/View/ViewHandler.php#L384-L385)
  * When submitted form is invalid, the http code is 400 by default cf [ViewHandler.php#L200-L201](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/View/ViewHandler.php#L200-L201) but in our behat scenario we expect 200 (hence the failing build with my modification)

winzou:
Ready to merge :)

winzou:
ping @pjedrzejewski PR waiting for more than a month

pjedrzejewski:
@winzou I love the use case and would like to have something for that, but it is a bit hacky don't you think? Overriding protected method just to add some data to template sounds a bit too much. I will think about something for this case! Thanks!

PR #2994, opened 2015-07-13
Add xhr option to header redirection

winzou:
Q | A
------------- | -------------
Bug fix? | no
New feature? | no
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

So that with the same route, you can choose to have an header redirection only when the request is done by XmlHttpRequest. Very useful!

I was wondering if this should be the default behavior, but at the end with these values (true, false or 'xhr') we can let the developer choose.


pjedrzejewski:
Makes sense, very useful! Thank you Alexandre!

PR #2996, opened 2015-07-13
[Taxonomy] Set root taxon locale based on current taxonomy locale

ahmadrabie:
Q | A
------------- | -------------
Bug fix? | yes
New feature? | no
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

Sometimes (e.g. submitting forms) entity's `Translation`s current locale is not set so we need to set it manually. 

Our current issue is that after creating `Taxonomy` from API `$resource` is going to be serialized for the response, because `$currentLocale` on root Taxon is not set we're going to have a `No current locale` exception.


pjedrzejewski:
Thanks Ahmad! :+1:

PR #2997, opened 2015-07-13
Ship by shipment.

kayue:
BC Break: No
New Feature: Yes

Allow ship shipment (ie: change inventory state to `sold`) by shipment. Useful when doing multi-shipment.

Sylius support multi-shipment, it makes more sense if admin can ship them individually.


michalmarcinkowski:
Implemented 🎉

PR #2998, opened 2015-07-13
Escaping slashes in search queries because of elasticsearch behavior

Dukecz:
Escaping slashes in search queries because of elasticsearch behavior, more here https://github.com/elastic/elasticsearch/issues/2980


pjedrzejewski:
Thank you @Dukecz!

PR #3001, opened 2015-07-14
Added GD extension to required

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
     ],
     "require": {
         "php":                                  ">=5.3.9",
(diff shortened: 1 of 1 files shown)


kayue:
Is it really required?

pamil:
@kayue You're right, it's `Imagine` library that requires `GD2`, `Imagick` or `Gmagick`, so we can't require GD.

PR #3010, opened 2015-07-15
Replace dot with underscore in parameter name.

kayue:
Bug fix: yes
BC break: no

Dots are replaced by underscores in PHP, so we should avoid using dot in parameter name. (ie: cookie name)

Without this fix, `CookieStorage` will not be able to retrieve what it set.

See more discussion in Symfony: https://github.com/symfony/symfony/issues/9009


pjedrzejewski:
Thank you Ka Yue! :+1:

Issue #3011, opened 2015-07-15
500 Internal Server Error - NonUniqueResultException


Symfomany:
I have a NonUniqueResultException when i go in product page in FrontOffice

at AbstractQuery ->getOneOrNullResult () 
in src/Sylius/Bundle/ResourceBundle/Doctrine/ORM/EntityRepository.php at line 78   + 
at EntityRepository ->findOneBy (array('slug' => '')) 
in src/Sylius/Bundle/CoreBundle/Routing/RouteProvider.php at line 137   + 
at RouteProvider ->getRouteCollectionForRequest (object(Request)) 
in vendor/symfony-cmf/routing/NestedMatcher/NestedMatcher.php at line 141   + 
at NestedMatcher ->matchRequest (object(Request)) 
in vendor/symfony-cmf/routing/DynamicRouter.php at line 258   + 

Url: http://localhost:8000/p/

pjedrzejewski:
Fixed in #3015.

PR #3015, opened 2015-07-15
Fixed empty product slug bug

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3011 
| License       | MIT
| Doc PR        |

After passing empty string as product slug, NonUniqueResultException was thrown, due to "findOneBy" method. Now it simply does not generate route, as for invalid slug.

However, it should be considered to generate some "invalidate product" page, rather than "No route found" exception :)

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Routing/RouteProvider.php
+++ b/src/Sylius/Bundle/CoreBundle/Routing/RouteProvider.php
@@ -134,9 +134,14 @@ class RouteProvider extends DoctrineProvider implements RouteProviderInterface
             ) {
                 $value = substr($path, strlen($this->routeConfigs[$className]['prefix']));
                 $value = trim($value, '/');
+
+                if (empty($value)) {
+                    continue;
+                }
+
                 $entity = $repository->findOneBy(array($this->routeConfigs[$className]['field'] => $value));
 
-                if (!$entity) {
+                if (null === $entity) {
                     continue;
                 }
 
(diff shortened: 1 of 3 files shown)


michalmarcinkowski:
@Zales0123 the proper fix will be to add one if statement before calling repository method:
```php
...
$value = trim($value, '/');
if (empty($value)) {
    continue;
}
$entity = $repository->findOneBy(array($this->routeConfigs[$className]['field'] => $value));
...
```

and please add Behat scenario for that :wink:

src/Sylius/Bundle/CoreBundle/Routing/RouteProvider.php
```diff
@@ -134,13 +134,13 @@ public function getRouteCollectionForRequest(Request $request)
             ) {
                 $value = substr($path, strlen($this->routeConfigs[$className]['prefix']));
                 $value = trim($value, '/');
-                $entity = $repository->findOneBy(array($this->routeConfigs[$className]['field'] => $value));
+                $entity = $repository->findBy(array($this->routeConfigs[$className]['field'] => $value));
 
-                if (!$entity) {
```
michalmarcinkowski: You can also change it to:
```php
if (null === $entity) {
```

Zales0123:
Sure, you're right! I have sometimes problem with complicating things to much ;)

pjedrzejewski:
@Zales0123 This scenario should be about 404, we do not want to show exception to the user...

pjedrzejewski:
Thank you Mateusz!

PR #3022, opened 2015-07-15
Fixed Bad Refactor in transliterate

patrick-mcdougle:
`empty()` is a safer check in the case that somehow the `$this->parentSlug` got set to NULL.

### DIFF
--- a/src/Sylius/Bundle/TranslationBundle/GedmoHandler/TranslationSlugHandler.php
+++ b/src/Sylius/Bundle/TranslationBundle/GedmoHandler/TranslationSlugHandler.php
@@ -190,7 +190,7 @@ class TranslationSlugHandler implements SlugHandlerInterface
      */
     public function transliterate($text, $separator, $object)
     {
-        if ('' !== $this->parentSlug) {
+        if (!empty($this->parentSlug)) {
             $text = $this->parentSlug.$this->usedPathSeparator.$text.$this->suffix;
         } else {
             // if no parentSlug, apply our prefix


Symfomany:
Add this fix in all slug of Model classes!

sarabkohli:
:+1: :+1:

patrick-mcdougle:
@Symfomany This is the only place I see this issue, please let me know if there's somewhere that you think I'm missing.

Symfomany:
Sorry Patrick but all __toString() thrown an exception when there is not the good language in database...:(

patrick-mcdougle:
@Symfomany I think that's out of the scope of what I'm fixing at this point. If I had to guess though, you probably need to install the php-intl extension and/or set your default locale. It's my understanding that [Stack Overflow](http://stackoverflow.com/questions/tagged/sylius) is the official support channel for Sylius if you need more help.

michalmarcinkowski:
:+1:

src/Sylius/Bundle/TranslationBundle/GedmoHandler/TranslationSlugHandler.php
```diff
@@ -190,7 +190,7 @@ public function onSlugCompletion(SluggableAdapter $ea, array &$config, $object,
      */
     public function transliterate($text, $separator, $object)
     {
-        if ('' !== $this->parentSlug) {
+        if (!empty($this->parentSlug)) {
```
stloyd: This should be:
```php
if (null !== $this->parentSlug && '' !== $this->parentSlug) {}

pjedrzejewski:
Thank you Patrick! :+1:

src/Sylius/Bundle/TranslationBundle/GedmoHandler/TranslationSlugHandler.php
patrick-mcdougle: What advantage does that have over `!empty()`?

PR #3040, opened 2015-07-16
[Rbac] Clear permissions cache on postUpdate fix

peteward:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Add postUpdate event to CacheListener to ensure that the cache is cleared when permissions are added to and removed from roles through the admin.

### DIFF
--- a/src/Sylius/Bundle/RbacBundle/EventListener/CacheListener.php
+++ b/src/Sylius/Bundle/RbacBundle/EventListener/CacheListener.php
@@ -53,6 +53,14 @@ class CacheListener
     /**
      * @param LifecycleEventArgs $args
      */
+    public function postUpdate(LifecycleEventArgs $args)
+    {
+        $this->clearCache($args);
+    }
+
+    /**
+     * @param LifecycleEventArgs $args
+     */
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
Thank you Peter! :+1:

jimbocoder:
Awesome, this goes right along with #3106 -- does anybody think this makes a higher TTL more reasonable now?  Most updates to RBAC data should, in a perfect world, occur through this class and automatically invalidate the cache.  I personally am always a proponent of the lowest TTL possible until a higher TTL is proven to be necessary.

peteward:
Yes, i was thinking that when I saw your PR merged that 60 seconds is a bit low if it's cleared on demand. Maybe 300 / 600 seconds?

jimbocoder:
300 makes sense to me.  What's the protocol for this sort of thing, brand new PR or commit to the old one or what?

pjedrzejewski:
@jimbocoder Brand new PR is the right way. :)

PR #3042, opened 2015-07-16
Compatibility with 3.0 and new theme for backend

Symfomany:
Several change about new theme & compatibility with 3.0


.gitignore
```diff
@@ -1,3 +1,5 @@
+.idea/
```
Strontium-90: Add this to global .gitignore

michalmarcinkowski:
Such PRs should be followed by a discussion. Please open RFC if this is still relevant.

PR #3048, opened 2015-07-17
Stop registering PHPCR ORM annotations in autoload

inssein:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

I tried running Sylius with PHP7 last weekend, and I had issues getting it to run. After reading https://github.com/doctrine/phpcr-odm/pull/412, it seems like this line is no longer necessary, and it will stop including the String class.

This should probably stop the PHP7 travis tests from breaking where it does now.

A separate pull request is also required for the Sylis/Standard, as it's autoload also has the same line.

### DIFF
--- a/app/autoload.php
+++ b/app/autoload.php
@@ -18,7 +18,6 @@ if (!function_exists('intl_get_error_code')) {
     require_once __DIR__.'/../vendor/symfony/symfony/src/Symfony/Component/Locale/Resources/stubs/functions.php';
 }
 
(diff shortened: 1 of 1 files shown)


michalmarcinkowski:
It fixed the issue with annotation on PHP7. Thank you Hussein! :+1: 

Unfortunately other thing is breaking the build on PHP7 now…

inssein:
I didn't see this error when I deployed my own site, but perhaps I could take a quick look over the weekend.

PR #3059, opened 2015-07-17
Update Search Repository to retain ordering

peter-fritchley:
The order that the results are returned from the search engine are important. Don't throw the order away when getting the entities from the database.

This assumes that every entity that is searchable will also have a getId() method - which is correct for products, I'm not sure about other entities.


michalmarcinkowski:
SearchBundle is already removed.

PR #3061, opened 2015-07-18
[PricingBundle] Consider standard pricing calculator in priceable form types

KasraKhosravi:
Q | A
------------- | -------------
Bug fix? | yes
New feature? | no
BC breaks? | no
Deprecations? | no
Fixed tickets | no
License | MIT
Doc PR | -

When using API, I had to post `pricingCalculator` in order to save product. Now with this change, it is not necessary to post it and it will use `standard` as default calculator.

### DIFF
--- a/src/Sylius/Bundle/PricingBundle/Form/EventListener/BuildPriceableFormSubscriber.php
+++ b/src/Sylius/Bundle/PricingBundle/Form/EventListener/BuildPriceableFormSubscriber.php
@@ -11,6 +11,7 @@
 
 namespace Sylius\Bundle\PricingBundle\Form\EventListener;
 
+use Sylius\Component\Pricing\Calculator\Calculators;
 use Sylius\Component\Pricing\Model\PriceableInterface;
 use Sylius\Component\Registry\ServiceRegistryInterface;
 use Sylius\Component\Resource\Exception\UnexpectedTypeException;
@@ -38,20 +39,31 @@ class BuildPriceableFormSubscriber implements EventSubscriberInterface
      */
     private $factory;
 
+    /**
+     * @param ServiceRegistryInterface $calculatorRegistry
+     * @param FormFactoryInterface $factory
+     */
     public function __construct(ServiceRegistryInterface $calculatorRegistry, FormFactoryInterface $factory)
     {
         $this->calculatorRegistry = $calculatorRegistry;
         $this->factory = $factory;
     }
 
+    /**
+     * {@inheritdoc}
+     */
     public static function getSubscribedEvents()
     {
         return array(
             FormEvents::PRE_SET_DATA => 'preSetData',
-            FormEvents::PRE_SUBMIT   => 'preBind'
+            FormEvents::PRE_SUBMIT   => 'preSubmit',
+            FormEvents::POST_SUBMIT  => 'postSubmit'
         );
     }
 
+    /**
+     * @param FormEvent $event
+     */
     public function preSetData(FormEvent $event)
     {
         $priceable = $event->getData();
@@ -61,13 +73,16 @@ class BuildPriceableFormSubscriber implements EventSubscriberInterface
         }
 
         if (!$priceable instanceof PriceableInterface) {
-            throw new UnexpectedTypeException($priceable, 'Sylius\Component\Pricing\Model\PriceableInterface');
+            throw new UnexpectedTypeException($priceable, PriceableInterface::class);
         }
 
         $this->addConfigurationFields($event->getForm(), $priceable->getPricingCalculator(), $priceable->getPricingConfiguration());
     }
 
-    public function preBind(FormEvent $event)
+    /**
+     * @param FormEvent $event
+     */
+    public function preSubmit(FormEvent $event)
     {
         $data = $event->getData();
 
@@ -76,6 +91,22 @@ class BuildPriceableFormSubscriber implements EventSubscriberInterface
         }
 
         $this->addConfigurationFields($event->getForm(), $data['pricingCalculator']);
+    }
+
+    /**
+     * @param FormEvent $event
+     */
+    public function postSubmit(FormEvent $event)
+    {
+        $priceable = $event->getData();
+
+        if (!$priceable instanceof PriceableInterface) {
+            throw new UnexpectedTypeException($priceable, PriceableInterface::class);
+        }
+
+        if (null === $priceable->getPricingCalculator()) {
+            $priceable->setPricingCalculator(Calculators::STANDARD);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


aramalipoor:
:+1:

aramalipoor:
@michalmarcinkowski ping!

src/Sylius/Bundle/PricingBundle/Form/EventListener/BuildPriceableFormSubscriber.php
```diff
@@ -85,8 +86,13 @@ public function preBind(FormEvent $event)
      * @param string        $calculatorType
      * @param array         $data
      */
-    protected function addConfigurationFields(FormInterface $form, $calculatorType, array $data = array())
+    protected function addConfigurationFields(FormInterface $form, $calculatorType = null, array $data = array())
```
stloyd: Why not put this: `Calculators::STANDARD` directly here?

pjedrzejewski: :+1:

michalmarcinkowski: :+1:

aramalipoor: This is a good thing to add but `$caculatorType` sometimes will passed as null (e.g. on [line 79](https://github.com/Sylius/Sylius/pull/3061#discussion-diff-38894690R79)), so that will override argument default value behavior.

stloyd: Hmmm, IMO `pricingCalculator` should never be `null`, if there is any case that set it to `null` it should be replaced with that `Calculators::STANDARD` value.

@pjedrzejewski @michalmarcinkowski Agree?

michalmarcinkowski: If we could avoid passing `null` that would be great.

aramalipoor: Our situation is we're posting data to product's API route, right now we only want to support `Standard` calculator so our developers expect that if they don't pass a calculator it becomes `standard` by default.

The way that resource bundle is handling the the API's content is using Forms so it means that product form will be created with an empty (null) pricingCalculator, hence on [this line](https://github.com/Sylius/Sylius/pull/3061/files#diff-1676b06edd6dab16504a09f9c110dc84L78) `$data['pricingCalculator']` is null which is breaking things.

If you think it is OK I can check for null value on that line!?

pjedrzejewski: No I get the idea. I think it makes sense to do it like this for now. :+1:

src/Sylius/Bundle/PricingBundle/Form/EventListener/BuildPriceableFormSubscriber.php
```diff
@@ -85,8 +86,12 @@ public function preBind(FormEvent $event)
      * @param string        $calculatorType
      * @param array         $data
      */
-    protected function addConfigurationFields(FormInterface $form, $calculatorType, array $data = array())
+    protected function addConfigurationFields(FormInterface $form, $calculatorType = Calculators::STANDARD, array $data = array())
     {
+        if (!$calculatorType) {
+            $calculatorType = Calculators::STANDARD;
+        }
```
stloyd:
```php
$calculatorType = $calculatorType ?: Calculators::STANDARD

Comment (aramalipoor) [inline src/Sylius/Bundle/PricingBundle/Form/EventListener/BuildPriceableFormSubscriber.php]: :+1:

File: src/Sylius/Bundle/PricingBundle/Form/EventListener/BuildPriceableFormSubscriber.php (hunk 1 above)
Comment (stloyd) [inline src/Sylius/Bundle/PricingBundle/Form/EventListener/BuildPriceableFormSubscriber.php]: @aramalipoor What if we replace mentioned line with:
```php
$this->addConfigurationFields($event->getForm(), $data['pricingCalculator'] ?: Calculators::STANDARD);

**Comment by aramalipoor** [comment]
After some investigations I figured out that the problem is that the process of `submitting` the form will erase `pricingCalculator` field on priceable subject. 

So the only solution I came up with was to set standard as default on `POST_SUBMIT` which I'm not sure if it's the right way to do it!

But anyway previous solution that we've been talking about would not work in some situations!

**Comment by aramalipoor** [comment]
@pjedrzejewski Please review and re-run the test :)

**Comment by michalmarcinkowski** [comment]
In order to make tests green you have to rebase with master (PHP 5.3 support has been dropped) :wink:

File: src/Sylius/Bundle/PricingBundle/Form/EventListener/BuildPriceableFormSubscriber.php [hunk 3]
```diff
@@ -79,6 +94,22 @@ public function preBind(FormEvent $event)
...
 
     /**
+     * @param FormEvent $event
+     */
+    public function postSubmit(FormEvent $event)
+    {
+        $priceable = $event->getData();
+
+        if (!$priceable instanceof PriceableInterface) {
+            throw new UnexpectedTypeException($priceable, 'Sylius\Component\Pricing\Model\PriceableInterface');
```
michalmarcinkowski: From PHP 5.5 we can use `PriceableInterface::class` instead of hardcoded string

src/Sylius/Bundle/PricingBundle/spec/Form/EventListener/BuildPriceableFormSubscriberSpec.php
```diff
@@ -67,4 +68,15 @@ function it_should_add_configuration_fields_in_pre_set_data(
 
         $this->preSetData($event);
     }
+
+    function it_should_set_default_pricing_calculator_to_standard_if_null_on_post_submit(
+        FormEvent $event,
+        PriceableInterface $priceable
+    ) {
+        $event->getData()->shouldBeCalled()->willReturn($priceable);
```
michalmarcinkowski: `shouldBeCalled()` is not necessary here :)

aramalipoor: I've fixed others as well, if it's right :p

stloyd:
:+1:

michalmarcinkowski:
:+1:

pjedrzejewski:
Thank you Kasra! :+1:

PR #3063, opened 2015-07-19
[WebBundle] Fixed showing ProductAttribute choices

zergu:
Fixed setting choices for product attribute by removing bootstrap "hide" class which does not work with jQuery show(). Visibility is now handled by JS only.

Also .hide class is deprecated.


pjedrzejewski:
Thank you Michał! :+1:

PR #3067, opened 2015-07-20
[ResourceBundle] Turning on back softdeleteable filter just after necessary operations

pamil:
Removed side effect.

| Q             | A
| ------------- | ---
| Bug fix?      | yes (a little bit)
| New feature?  | no
| BC breaks?    | yes (event listeners will get Doctrine with enabled  `softdeleteable` filter)
| Deprecations? | no
| License       | MIT


pjedrzejewski:
Thank you Kamil!

PR #3070, opened 2015-07-21
Fixes found while running JMSTranslationBundle

qkdreyer:
I had to fixes those in order to let JMSTranslationBundle run successfully


michalmarcinkowski:
This class doesn't exist any more.

PR #3035, opened 2015-07-16
Introducing Intl component to translate country names instead of Translatable

pamil:
PR based on #2720 (thanks @gperdomor!).

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes (there is no `USA` country no more, use `United States`)
| Deprecations? | no
| Fixed tickets | #2710, #2720 
| License       | MIT

PR #3074, opened 2015-07-22
[Addressing] Countries management - enabling & disabling

pamil:
What was done, it's a massive PR:

  - [x] Country translation based on ISO country code and `Intl` Symfony2 component
  - [x] Country management improved, ability to add new countries (from list) and enable or disable existing ones, removed ability to delete countries due to unspecified behaviour of present and previous orders based on address with that country etc.
  - [x] Removed redunant comments and fluent interface in Country

Based on #3098 and #3035.


pjedrzejewski:
@pamil Awesome work Kamil! :+1: Thank you and @gperdomor! You rock guys.

Issue #2979, opened 2015-07-08
Shipping Error when revisiting shipping method page


carlowens:
Here is the error:

The form's view data is expected to be of type scalar, array or an instance of \ArrayAccess, but is an instance of class Proxies\__CG__\Sylius\Component\Core\Model\ShippingMethod. You can avoid this error by setting the "data_class" option to "Proxies\__CG__\Sylius\Component\Core\Model\ShippingMethod" or by adding a view transformer that transforms an instance of class Proxies\__CG__\Sylius\Component\Core\Model\ShippingMethod to scalar, array or an instance of \ArrayAccess.

How to re-create the error is as follows:

Go through the checkout as usual, however when you get to the payment details page (last page), click on the basket in the header (back to /cart) and go through the checkout steps again.  
Once you hit the shipping method page you will see the error above.

This happens currently on my local (v0.14) and on the current live demo.

coudenysj:
@carlowens Can you supply some more details? The content of the cart, etc...? I've just tested this with a fresh install of 6c7d32d, and I can visit the page as much as I want without errors.

carlowens:
I only found this issue when updating from v0.11 to v0.14 which we have now decided to wait as there are a couple of issues which we don't have time to iron out right now.

But as I said I can get this error on the Sylius demo site (http://demo.sylius.org/).

Simply by adding a product to the card (any as far as i can tell), proceeding to the payment page (last page) and then clicking back on the basket and going through the steps again.  Once you reach the shipping method page you get the error.

Sorry I don't have time at the moment to provide more info or a behat example.

msumme:
Running into the same thing after an upgrade to 0.14

msumme:
I tracked this down to an issue with the ShippingMethodChoiceType in the ShippingBundle.  It works perfectly when a shipping method isn't selected, but breaks when a method is selected for the order.

This may have something to do with upgrading symfony versions.  The fix is to simply add a data_class to the OptionsResolver defaults.

msumme:
Also - as a workaround before 0.15 is released

```php
namespace AppBundle\Form;

use Sylius\Bundle\CoreBundle\Form\Type\Shipping\ShippingMethodChoiceType as BaseType;
use Symfony\Component\OptionsResolver\OptionsResolverInterface;

class ShippingMethodChoiceType extends BaseType
{
    public function setDefaultOptions(OptionsResolverInterface $resolver) 
    {
        parent::setDefaultOptions($resolver);
[...]

PR #3085, opened 2015-07-25
Fixing Type Error when Order has ShippingMethod

msumme:
This fixes the error "The form's view data is expected to be of type
scalar, array or an instance of \ArrayAccess, but is an instance of
class Proxies\__CG__\Sylius\Component\Core\Model\ShippingMethod" that
occurs when viewing the shipping form with either only one option
available, or when going back from checkout.

Fixes issue #2979

### DIFF
--- a/src/Sylius/Bundle/ShippingBundle/Form/Transformer/ObjectToIdentifierTransformer.php
+++ b/src/Sylius/Bundle/ShippingBundle/Form/Transformer/ObjectToIdentifierTransformer.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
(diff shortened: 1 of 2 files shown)


msumme:
Adding a data_class is insufficient since the form is sometimes used with "multiple" and sometimes not.

stloyd:

pjedrzejewski:
Perhaps we could use generic transformer from ResourceBundle or if it does not exist - add it there?

msumme:
That's a great idea.  There's already a transformer in another bundle, but
it seemed like it was duplicated elsewhere.  I figured (because of the
duplication) that there were some efforts to keep bundles isolated.   That
was just my best guess at current project architecture.

I won't be able to get to this for at least a month, unfortunately, so
someone else should probably just grab this one and make a new PR.

PR #3091, opened 2015-07-29
Faster feedback from Travis

pamil:
- [x] First feedback after 19 minutes compared to 32 minutes before
- [x] Changed order to see most errors after 4 jobs, newest PHP 5.6 is first, then PHP 5.3 with oldest possible libraries is tested
- [x] For now, PHP7 tests are executed at the end and build is passing just after last required test passed
- [x] More verbose output, shows at what time and what command was executed, no more counting commands order
- [x] Command suites are written clearer and don't contain any Travis logic
- [x] Removed commented out lines to make Travis files clearer


pjedrzejewski:
Thank you Kamil, awesome work!

PR #3094, opened 2015-07-30
[Locale] Translations based on Intl component, can be enabled or disabled

pamil:
Based on #3098.


features/localization/locale_selection.feature
```diff
@@ -5,29 +5,29 @@ Feature: Locale selection
     I want to to select my language in the storefront
 
     Background:
-        And there are following locales configured:
+          And there are following locales configured:
```
michalmarcinkowski: ``` And ``` -> ``` Given ```

src/Sylius/Bundle/CoreBundle/Behat/CoreContext.php
```diff
@@ -430,7 +430,16 @@ public function thereAreLocales(TableNode $table)
 
         foreach ($table->getHash() as $data) {
             $locale = $repository->createNew();
-            $locale->setCode($data['code']);
+
+            if (isset($data['code'])) {
+                $locale->setCode($data['code']);
+            } elseif (isset($data['name'])) {
+                $code = $this->getLocaleCodeByEnglishLocaleName($data['name']);
+
```
michalmarcinkowski: Unnecessary blank line

src/Sylius/Bundle/CoreBundle/Form/Type/LocaleType.php
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * {@inheritdoc}
+     *
+     * @param RepositoryInterface $localeRepository
+     */
+    public function __construct($dataClass, array $validationGroups = array(), RepositoryInterface $localeRepository)
+    {
+        $this->localeRepository = $localeRepository;
+
+        parent::__construct($dataClass, $validationGroups);
```
michalmarcinkowski: Parent constructor should be called at first.

src/Sylius/Bundle/WebBundle/Resources/views/Backend/Locale/macros.html.twig
```diff
@@ -8,24 +8,27 @@
...
                 <th>{{ sylius_resource_sort('enabled', 'sylius.locale.enabled'|trans) }}</th>
-                <th>{{ sylius_resource_sort('updatedAt', 'sylius.locale.updated_at'|trans) }}</th>
                 <th></th>
             </tr>
         </thead>
         <tbody>
         {% for locale in locales %}
-            <tr id="{{ locale.id }}">
+            <tr>
+                <td>{{ locale.id }}</td>
```
michalmarcinkowski: Do we need to display ID?

src/Sylius/Component/Locale/spec/Model/LocaleSpec.php
```diff
@@ -41,15 +61,56 @@ function its_code_is_mutable()
...
+    function it_can_set_enabled_value()
+    {
+        $this->setEnabled(false);
+        $this->isEnabled()->shouldReturn(false);
+
+        $this->setEnabled(true);
+        $this->isEnabled()->shouldReturn(true);
+
         $this->setEnabled(false);
-        $this->shouldNotBeEnabled();
```
michalmarcinkowski: Why?

pamil: I set that locale as disabled, so it should not be enabled :)

src/Sylius/Bundle/WebBundle/Resources/views/Backend/Locale/macros.html.twig
pamil: It was displayed on countries index too, but probably it's not needed by anyone :)

src/Sylius/Bundle/CoreBundle/Behat/CoreContext.php
pamil: I think it makes the code more readable.

src/Sylius/Component/Locale/spec/Model/LocaleSpec.php
michalmarcinkowski: My question was why you changed ``` $this->shouldNotBeEnabled() ``` to ``` $this->isEnabled()->shouldReturn(false) ```?

pamil: Aaaah, I know, I wrote new assertions on `isEnabled` and I used `isEnabled()->shouldReturn(bool)`, so this line was the only one with `shouldNotBeEnabled()`.

src/Sylius/Bundle/WebBundle/Resources/views/Backend/Locale/macros.html.twig
michalmarcinkowski: We have unique code, so no need for ID. IMO the table should be the following:

| Code        | Name           | Enabled  | Buttons |
| ------------- |:-------------:| -----:| -------:|

pamil: It's a good idea, I'll make a PR for countries too.

src/Sylius/Bundle/CoreBundle/Behat/CoreContext.php
pjedrzejewski: I'd actually avoid the variable, it is used only once. Just set the result of method directly.

pamil: It may do the job, but it will be quite a long line.

src/Sylius/Component/Locale/spec/Model/LocaleSpec.php
michalmarcinkowski: IMO all assertions should be written in this manner, because it's more readable.

michalmarcinkowski:
Thank you Kamil! :+1:

PR #3095, opened 2015-07-31
[UserBundle] Fix invalid serialization type 'int' to 'integer'.

jimbocoder:
The title nearly says it all.  `int` simply isn't a valid JMS serialization type, it should be `integer`.


pjedrzejewski:
Thank you Jim! :+1:

PR #3097, opened 2015-08-03
Status field not set on offline gateway

nicolasricci:
Status field is not set which results in a cancelled order when processing an order with the offline gateway


makasim:
:+1:

pjedrzejewski:
Thank you Nicolas! :+1:

PR #3098, opened 2015-08-03
[ResourceBundle] Toggleable interface & useful Behat steps and methods

pamil:
- [x] ResourceBundle got new `ToggleableInterface` for resources you can enable or disable, `Enabled` and `Disabled` constraints to validate your entities and custom enable and disable actions in `ResourceController` (*`Country` is the first one to implement `ToggleableInterface`*)
  - [x] ResourceBundle got new custom buttons for enabling/disabling
  - [x] ResourceBundle got new steps for manipulating tables and select inputs and checking if given resource is enabled/disabled

### DIFF
--- a/features/currencies/currency_switching.feature
+++ b/features/currencies/currency_switching.feature
@@ -57,4 +57,4 @@ Feature: Currency selection
         And I change the currency to "GBP"
         And I added product "PHP Top" to cart, with quantity "1"
         When I go to the cart summary page
-        Then I should see item with unit price "£6.83" in the list
+        Then I should see item with price "£6.83" in the list
--- a/src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
+++ b/src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
@@ -12,6 +12,7 @@
 namespace Sylius\Bundle\ResourceBundle\Behat;
 
 use Behat\Behat\Context\Context;
+use Behat\Mink\Element\NodeElement;
 use Behat\Mink\Exception\ElementNotFoundException;
 use Behat\MinkExtension\Context\RawMinkContext;
 use Behat\Symfony2Extension\Context\KernelAwareContext;
@@ -33,15 +34,11 @@ abstract class DefaultContext extends RawMinkContext implements Context, KernelA
     protected $applicationName = 'sylius';
 
     /**
-     * Faker.
-     *
      * @var Generator
      */
     protected $faker;
 
     /**
-     * Actions.
-     *
      * @var array
      */
     protected $actions = array(
@@ -74,8 +71,6 @@ abstract class DefaultContext extends RawMinkContext implements Context, KernelA
     }
 
     /**
-     * Find one resource by name.
-     *
      * @param string $type
      * @param string $name
      *
@@ -87,8 +82,6 @@ abstract class DefaultContext extends RawMinkContext implements Context, KernelA
     }
 
     /**
-     * Find one resource by criteria.
-     *
      * @param string $type
      * @param array  $criteria
      *
@@ -113,20 +106,16 @@ abstract class DefaultContext extends RawMinkContext implements Context, KernelA
     }
 
     /**
-     * Get repository by resource name.
-     *
-     * @param string $resource
+     * @param string $resourceName
      *
      * @return RepositoryInterface
      */
-    protected function getRepository($resource)
+    protected function getRepository($resourceName)
     {
-        return $this->getService($this->applicationName.'.repository.'.$resource);
+        return $this->getService($this->applicationName.'.repository.'.$resourceName);
     }
 
     /**
-     * Get entity manager.
-     *
      * @return ObjectManager
      */
     protected function getEntityManager()
@@ -135,8 +124,6 @@ abstract class DefaultContext extends RawMinkContext implements Context, KernelA
     }
 
     /**
-     * Returns Container instance.
-     *
      * @return ContainerInterface
      */
     protected function getContainer()
@@ -145,8 +132,6 @@ abstract class DefaultContext extends RawMinkContext implements Context, KernelA
     }
 
     /**
-     * Get service by id.
-     *
      * @param string $id
      *
      * @return object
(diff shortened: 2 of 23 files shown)


features/currencies/currency_switching.feature
```diff
@@ -57,4 +57,4 @@ Feature: Currency selection
         And I change the currency to "GBP"
         And I added product "PHP Top" to cart, with quantity "1"
         When I go to the cart summary page
-        Then I should see item with unit price "£6.83" in the list
+        Then I should see item with price "£6.83" in the list
```
michalmarcinkowski: Why?

src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
```diff
@@ -340,4 +321,109 @@ protected function fixStepArgument($argument)
...
+     * @param NodeElement $table
+     * @param array $fields
+     *
+     * @return NodeElement|null
+     *
+     * @throws \Exception If column was not found
+     */
+    protected function getRowWithFields(NodeElement $table, array $fields)
+    {
+        $foundRows = $this->getRowsWithFields($table, $fields, true);
```
michalmarcinkowski: The method returns array of rows. According to docblock we should return ``` NodeElement ``` instead we return this array.

src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
```diff
@@ -340,4 +321,109 @@ protected function fixStepArgument($argument)
...
+
+        return $foundRows;
+    }
+
+    /**
+     * @param NodeElement $table
+     * @param array $fields
+     * @param boolean $first
+     *
+     * @return NodeElement[]|NodeElement
```
michalmarcinkowski: From what I see, it always return array.

src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
```diff
@@ -340,4 +321,109 @@ protected function fixStepArgument($argument)
...
+        if (!isset($rows[0])) {
+            throw new \Exception("There are no rows!");
+        }
+
+        foreach ($fields as $columnName => $expectedValue) {
+            $columnIndex = $this->getColumnIndex($table, $columnName);
+
+            $fields[$columnIndex] = $expectedValue;
+            unset($fields[$columnName]);
+        }
```
michalmarcinkowski: This loop could be extracted to a private method.

src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
```diff
@@ -340,4 +321,109 @@ protected function fixStepArgument($argument)
...
+                    $found = false;
+
+                    break;
+                }
+            }
+
+            if ($found) {
+                $foundRows[] = $row;
+
+                if ($first) {
```
michalmarcinkowski: Why we need this?

features/currencies/currency_switching.feature
pamil: I made Behat contexts a little bit more strict. Earlier, "price" or "unit price" was just a name, that didn't change nothing. Moreover, test was passing if given value only existed on the page. Now it uses that name to find suitable column, and then checks if there's given value in given column.

src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
pamil: :+1:

src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
pamil: Sometimes we want to find only one row, not all.

features/currencies/currency_switching.feature
michalmarcinkowski: Ok, we've got "price" - for unit price and "total" - for all units overall price I see the point.

src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
michalmarcinkowski: IMO we should use different method then as this one (get**Rows**WithFields) suggest we get multiple :wink:

pamil: There's already `getRowWithFields` using `getRowsWithFields`, I don't want to duplicate code.

michalmarcinkowski: Ok then this two steps (https://github.com/Sylius/Sylius/pull/3098/files#diff-444c7166df771289a30354d3c2e236acR359, https://github.com/Sylius/Sylius/pull/3098/files#diff-444c7166df771289a30354d3c2e236acR373) looks only for the first row?

src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
pamil: Fixed, in fact, this function result was always compared with null. We should start writing tests for contexts too! :)

src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
pamil: I think it's like a query with `LIMIT 1`, it also returns an array with one element. We cannot guarantee that there will be multiple results.

pamil: It looks through all rows, but it cares only about first existence. If there's 1000 rows and the searched one is 10th, then it doesn't have to parse the rest.

michalmarcinkowski: Ok, so I misunderstood, the variable should be named ``` $onlyFirstOccurence ``` or something like that to avoid misunderstandings :)

pamil: Good idea :+1:

pjedrzejewski:
Thank you Kamil! :+1:

PR #3050, opened 2015-07-17
[Taxonomy] Update root taxon after taxonomy translation is added

pamil:
Continued PR #2786, originally written by @ahmadrabie, thanks! :)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2734, #2740, #2786
| License       | MIT

PR #3099, opened 2015-08-03
[TaxonomyBundle] Set root taxon's name same as taxonomy when creating

kasrakhosravi:
Q | A
------------- | -------------
Bug fix? | yes
New feature? | no
BC breaks? | no
Deprecations? | no
Fixed tickets | no
License | MIT
Doc PR | -


pamil:
I think it's the same as #3050, is it?

aramalipoor:
@pamil I'm not sure, I think this is a different scenario, Form component tries to call `setName()` of `TaxonomyTranslation` entity and nothing else. So we have to pass that name to `setName()` of root taxon of the `Taxonomy` being created.

We can test this after you've merged #3050 ;)

michalmarcinkowski:
Already fixed and will not be relevant after #3773.

PR #3104, opened 2015-08-04
[Locale] Refactored LocaleProvider and LocaleContext, changed architecture a little bit

pamil:


src/Sylius/Bundle/CoreBundle/Resources/config/app/main.yml
```diff
@@ -157,6 +157,6 @@ jms_serializer:
                 path: "@SyliusVariationBundle/Resources/config/serializer"
 
 a2lix_translation_form:
-    locale_provider: sylius.locale_provider
+    locale_provider: sylius.a2lix_locale_provider
```
aramalipoor: A problem that we were facing was that in administration panel we expect to see all enabled (not current channel's) locales, because admin users want to add translations to all languages. But because a2lix locale provider was channel-aware it would only show current channel's locales.

Can you verify that this change is not a regression on this matter? 
https://github.com/Sylius/Sylius/pull/2945

pamil: It's all right, `sylius.a2lix_locale_provider` uses `sylius.locale_provider` to get available locales. I divided previous `LocaleProvider` that implemented Sylius `LocaleProviderInterface` and A2lix `LocaleProviderInterface` into two classes.

aramalipoor: Ok great, thanks :+1:

pjedrzejewski:
Thank you Kamil! :+1:

gperdomor:
@pamil @pjedrzejewski i found a new problem with LocaleChoiceType, in LocaleBundle Configuration.php file, the default value of choice form must be Sylius\Bundle\LocaleBundle\Form\Type\LocaleChoiceType, but for some reason after change this value, i get a error when tried to edit or create new channel, because constructor not receive the correct parameter.

pamil:
@gperdomor thanks, will be fixed by #3147.

PR #3105, opened 2015-08-04
[RFC][Rbac] Use a finite default ttl in CachedPermissionMap.

jimbocoder:
Caching infinitely requires invalidating the cache entries whenever
roles or permissions are modified.  We could use a "sane" default TTL,
or make it configurable, or figure out an invalidation strategy, or...
But I think infinite by default is pretty unexpected behavior.


pjedrzejewski:
Definitely. Thank you Jim!

PR #3108, opened 2015-08-05
[FlowBundle] Remove unused events in FlowBundle

krizon:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no? (this events were never dispatched and are not documented but one might have implemented a listener which would never been called but could trigger some error with the absent of the event constants and classes)
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -


pjedrzejewski:
Thank you Kristian, the less code the better. :)

PR #3111, opened 2015-08-05
[Behat] Fixed undefined step in features/user/users.feature:123

pamil:


pjedrzejewski:
Thanks Kamil! Feels awkward, cause I could just high five you in the office. :+1:

PR #3117, opened 2015-08-08
[UserBundle] Modify CustomerRepository filter logic to support multi-term queries.

jimbocoder:
For example, searching for "John Smith" would previously query for `LIKE
'%John Smith%'` against fields including first name, last name, email, etc.  Even with a customer named John Smith, this would find no matches since none of the fields alone contain his full name.

The new suggested logic tests 'John' against each field, AND tests 'Smith'
 against each field, AND tests further tokens against each field, and requires at least one match from each token to pass.

### DIFF
--- a/src/Sylius/Bundle/UserBundle/Doctrine/ORM/CustomerRepository.php
+++ b/src/Sylius/Bundle/UserBundle/Doctrine/ORM/CustomerRepository.php
@@ -63,14 +63,22 @@ class CustomerRepository extends EntityRepository
         }
 
         if (isset($criteria['query'])) {
-            $queryBuilder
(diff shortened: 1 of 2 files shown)


TomasVotruba:
Could you fix and provide tests?

jimbocoder:
I'll try to find some time to do that.

I have no experience with any of the testing tools Sylius uses, so I'll be starting from scratch.  If anybody would like to provide the necessary tests, please feel free! I'd definitely learn from them as examples to start including tests in future PRs.

jimbocoder:
Ok, after a little reading, I managed to update the spec such that it should pass.  No new behat test though, and I don't think I can do that myself without a big time investment to get set up with Behat.

michalmarcinkowski:
We use Grid filters now.

PR #3119, opened 2015-08-08
[Settings] Add API routes for SettingsController and some improvements

aramalipoor:
Q  | A
------------- | -------------
Bug fix?  | yes
New feature?  | yes
BC breaks? | no
Deprecations? | no
Fixed tickets | no
License | MIT
Doc PR | -

* Add showAction for Settings schema's for API consumers. `/api/settings/{namespace}`
* `showAction` and `updateAction` now will not throw `MissingOptions` exception which is annoying when it's the first to read/update a Settings schema through an API call. (If there's a better way than ignoring this exception that'd be nice to guide me on that)


pjedrzejewski:
Nice, thank you Aram! :+1:

Issue #3122, opened 2015-08-11
sylius:cart:purge -> findExpired orders


poolerMF:
hello

in Sylius\Bundle\CoreBundle\Doctrine\ORM\OrderRepository -> findExpired() there is comparing row updatedAt ... 

this should be expiresAt ?

poolerMF:
there should by update of expiresAt, when user successfully made payment ... 

becouse ... if user is shopping 2,5 hours ... then complete order ... 15 minutes later made payment ... CRON automatically delete this order 15 minutes after payment

michalmarcinkowski:
@poolerMF the query takes into account the order state and doesn't clear completed orders.

poolerMF:
I'm sorry, I made mistake ...I thought command sylius:ORDER:purge ...
but as I see, commands sylius:order:purge and sylius:cart:purge calls uses expiresAt < now

in both commands, situation is same as I described abowe ... on cartChange (product remove, add, change) or on order complete ... row expiresAt must be updated

jimbocoder:
I think that would be a business decision to be made.  In some industries, it makes perfect sense to extend the cart's expiration time whenever it is modified in any way.  In other industries (such as event ticketing, with limited inventory in high demand) the last thing you want to do is allow someone to hoard all your inventory just by juggling items in and out of their cart.  Not sure of the correct approach to satisfy all the cases, but I can see @poolerMF's point.

poolerMF:
ok then

PR #3126, opened 2015-08-11
[CoreBundle] Fix findExpired orders method query

jimbocoder:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes, if users are currently anticipating the bug and changing behavior to work around
| Deprecations? | no
| Fixed tickets | #3122 
| License       | MIT
| Doc PR        | -

It was using the updatedAt column instead of the expiresAt column.  Big difference!


pjedrzejewski:
Awesome work Jim, thank you very much! :+1:

PR #3127, opened 2015-08-12
SEO & Metadata

pamil:


peteward:
Hi @pamil,

How are you getting on with this? @pjedrzejewski  and I discussed the need for SEO content that can be written explicitly at a hierarchical level, but more importantly derived from a formula at a hierarchical level for any generated content.

For generated content the obvious first candidates are taxons and products.

* Taxons can have SEO content derived from their parents (in the case of a large structure of taxons).
* Products can have their content derived from their product type (archetype) or possibly taxons hierarchy.

The idea being to make it as easy as possible to generate a content-rich site for a large product catalogue with the minimal effort.

For formulas we were looking at something like the following as a rough example as a page title:
```
Where product.archetype == 'T-Shirt':
"The {{ product.name }} T-Shirt in {{ product.colour.name }} plays its part in our iconic Tops collection and is available to buy online at Reiss."
```

I think that for the first iteration it makes sense to write the formulas in a pseudo-twig format rather than trying to build a fully-functional editor knowing all entity attributes...

There would need to be a configuration of entities that can have SEO content and the levels that formulas can be applied to. Maybe it's just a blanket formula for all entities, or maybe it's based on an attribute / relation of that entity (as mentioned for products and taxons).

Anyway, let me know how you're getting on and if you want any help or advice.

peteward:
and as with Associated products I'm looking on the wrong PR........

PR #3130, opened 2015-08-13
[WebBundle] Remove default icon for enable and disable buttons

michalmarcinkowski:
IMO this icons are unnecessary and we should remove them at default.


aRn0D:
:+1: I think we should keep them :dancer:

pamil:
Every button has its own icon, why do we want to make an exception? It will save some space on small screens too. Anyway, I think that we need enable and disable buttons to have the same width, it will look nice while rendering them in column.

pjedrzejewski:
Let's keep them for now, otherwise on small screen you won't see anything.

michalmarcinkowski:
I don't think that current icons are the most approriate, but agree with you guys :+1:

PR #3132, opened 2015-08-14
[WebBundle] Removed ID column from the list

pamil:
As it is in locales (#3094).


aRn0D:
Indeed it is useless !

michalmarcinkowski:
Thank you Kamil! :+1:

PR #3133, opened 2015-08-14
[RFC][ProductBundle] When finding products in a given taxon, consider the whole tree.

jimbocoder:
When asking for items in a group, I believe it makes sense to consider the group's hierarchy at the same time and return a list that includes implicitly-included items.

Would be nice if we could extend the [NestedTreeRepository](https://github.com/Atlantic18/DoctrineExtensions/blob/master/lib/Gedmo/Tree/Entity/Repository/NestedTreeRepository.php) as it has some methods already to deal with this, but I don't see a way to hack that in  since it already must extend the ResourceRepository.

As a less flexible but simpler compromise, I'm interested to hear any thoughts on whether we could do something like the example in this PR.


pjedrzejewski:
I think this is good change, but would like to hear some more opinions before merging this!

pjedrzejewski:
Ping @Sylius/core-team!

pjedrzejewski:
Thank you Jim! We will look further into this when reworking the frontend.

PR #3128, opened 2015-08-13
[Metadata] Extensible metadata component and bundle for Symfony2

pamil:
## Metadata

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2307, #2358
| License       | MIT
| Doc PR       | Sylius/Sylius-Docs#322

This PR is aiming to provide an extensible metadata component for PHP and a bundle to integrate it with Symfony2.

For integration with Sylius see #3134.

### Possible improvements and extensions
  - Caching to speed it up
  - More models (Open Graph, Schema.org)
  - More advanced rendering abilities
  - Dynamic expressions (eg. "Products" scope will define`"%PRODUCT_NAME% - %PRODUCT_PRICE%` as title and it will be used for its children, unless they defined their own title)
  - Ability to change page URL (eg. product's url)

### TODO
 - [ ] Documentation

PR #3134, opened 2015-08-14
[Metadata] Sylius integration with metadata bundle (#3128)

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3128 (base metadata PR)
| License       | MIT
| Doc PR       | Sylius/Sylius-Docs#322


src/Sylius/Component/Metadata/Model/RootMetadata.php
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Metadata\Model;
+
+
```
peteward: blank line

src/Sylius/Component/Metadata/Provider/MetadataProvider.php
```diff
@@ -0,0 +1,107 @@
...
+        $this->metadataHierarchyProvider = $metadataHierarchyProvider;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getMetadataBySubject(MetadataSubjectInterface $metadataSubject)
+    {
+        $identifiers = $this->getHierarchyByMetadataSubject($metadataSubject);
+
```
peteward: blank line

src/Sylius/Component/Metadata/Provider/MetadataProviderInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
```
peteward: blank line

peteward:
Hi Kamil, so comments on the correct PR now! :)

All looking good once I used `sylius:rbac:initialize`, I'm surprised that this isn't called on initial install although I understand why it might not be.

I definitely agree that building a feature of this size progressively makes sense, it's already an enormous PR (and I've read 3 versions!).

src/Sylius/Bundle/MetadataBundle/DynamicForm/DynamicFormBuilder.php
```diff
@@ -0,0 +1,116 @@
...
+     * @param FormBuilderInterface $builder
+     * @param string $name
+     * @param string $group
+     */
+    private function addModelDataListener(FormBuilderInterface $builder, $name, $group)
+    {
+        $builder->addEventListener(FormEvents::PRE_SET_DATA, function (FormEvent $event) use ($builder, $name, $group) {
+            try {
+                $data = $this->propertyAccessor->getValue($event->getData(), $name);
+            } catch (AccessException $exception) {
```
peteward: Should catch `Symfony\Component\PropertyAccess\Exception\RuntimeException` as PropertyAccessor can throw both:
```
        } catch (AccessException $e) {
            return false;
        } catch (UnexpectedTypeException $e) {
            return false;
        }
```
and they both extend the RuntimeException (I got an UnexpectedTypeException here).

peteward:
From what I can see of the form the current metadata is explicitly page-based which is interesting if you consider image alt tags and titles. I'm wondering how you are thinking of structuring the interface for the wider solution?

I see the Metadata management section as a place for re-usable formula construction. The formula's would be dependent on the Model type that is in use (Content Page, Product, Taxon, etc) as to what attributes you can use.

The assigning of these formulas at the top-level could then be applied within this section, but where the metadata is going to be related to a specific page it makes to embed that form where the 'page' is managed. So if it's a content-managed page, I would imagine it as a tab within that section (and page hierarchy could dictate the use of metadata). For a Taxon it could be another tab on that page too, again following the hierarchy rules.

Back to the main management interface,  splitting by Model makes sense to me and in some extent this would be hierarchical too:

```
TOP-LEVEL:
    CONTENT PAGE
    TAXON
        *Mens Taxons*
        *Womens Taxons*
    PRODUCT
        *Products of Type T-Shirt*
    CONTENT IMAGE
```

In the above structure if nothing was defined for taxons at all then the system default would be used. If the top-level taxon content is used then this will default for all taxons, etc. For the next level of definition where you derive lower levels of the hierarchy from the set of taxons it is very useful to represent it here, probably more so than within the taxon pages? Although in a future PR it might be good to include it in the Taxon pages, either as a preview of what the metadata will look like or to edit it directly within that page.

Content Page, Taxon and Product can all inherit Top-Level content because they can all be treated as pages but Content Image wouldn't necessary be it's own page but would have it's own Alt and Title tags.
Traits feel like they could work here as Content Image is really a trait for Products and Taxons but can be used in it's own right as a metadata type.

Related to each explicit model type you then have your 'bank' of formulas and formula construction that can only be used within that model implementation of the metadata (as product attributes won't apply outside of that scope).

src/Sylius/Bundle/MetadataBundle/DynamicForm/DynamicFormBuilder.php
```diff
@@ -0,0 +1,116 @@
...
+    private function addSubmittedDataListener(FormBuilderInterface $builder, $name)
+    {
+        $builder->get($name)->addEventListener(FormEvents::PRE_SUBMIT, function (FormEvent $event) use ($builder, $name) {
+            $formName = $event->getData()['_form'];
+            if (empty($formName)) {
+                return;
+            }
+
+            $this->addEmbeddedFormField($builder->get($name), $event->getForm()->get($name), $formName);
+        });
```
peteward: If I try to save a metadata without choosing a form I get 2x CSRF errors. 

If I try to choose a twitter form option and then 'deselect' it by choosing the blank dropdown option the last form is not removed.

If I submit the form I get the following exception, am I missing some configuration?

```
[1] Symfony\Component\Form\Exception\OutOfBoundsException: Child "twitter" does not exist.
    at n/a
        in /var/www/petewardSylius/vendor/symfony/symfony/src/Symfony/Component/Form/Form.php line 979

    at Symfony\Component\Form\Form->get('twitter')
        in /var/www/petewardSylius/src/Sylius/Bundle/MetadataBundle/DynamicForm/DynamicFormBuilder.php line 114

    at Sylius\Bundle\MetadataBundle\DynamicForm\DynamicFormBuilder->Sylius\Bundle\MetadataBundle\DynamicForm\{closure}(object(FormEvent), 'form.pre_bind', object(EventDispatcher))
        in  line 

    at call_user_func(object(Closure), object(FormEvent), 'form.pre_bind', object(EventDispatcher))
        in /var/www/petewardSylius/app/cache/dev/classes.php line 4832

    at Symfony\Component\EventDispatcher\EventDispatcher->doDispatch(array(array(object(BindRequestListener), 'preBind'), array(object(TrimListener), 'preSubmit'), array(object(CsrfValidationListener), 'preSubmit'), object(Closure)), 'form.pre_bind', object(FormEvent))
        in /var/www/petewardSylius/app/cache/dev/classes.php line 4765

    at Symfony\Component\EventDispatcher\EventDispatcher->dispatch('form.pre_bind', object(FormEvent))
        in /var/www/petewardSylius/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/ImmutableEventDispatcher.php line 43

    at Symfony\Component\EventDispatcher\ImmutableEventDispatcher->dispatch('form.pre_bind', object(FormEvent))
        in /var/www/petewardSylius/vendor/symfony/symfony/src/Symfony/Component/Form/Form.php line 556
```

pamil:
Closing in favour of #3350.

PR #3146, opened 2015-08-17
generate function to return generated coupons

agounaris:
Hello there,

I had to use coupon generator and afterwards send out an email with the generated codes.

Doesn't it make more sense to return a structure containing the coupons instead of having to query back the
db for those?

This change works great for me but maybe it affects sylius in a way I cannot see at the moment. Please let me 
know in case this can be achieved in any other way.

Cheers,
Argiris

### DIFF
--- a/src/Sylius/Component/Promotion/Generator/CouponGenerator.php
+++ b/src/Sylius/Component/Promotion/Generator/CouponGenerator.php
@@ -43,16 +43,21 @@ class CouponGenerator implements CouponGeneratorInterface
      */
     public function generate(PromotionInterface $promotion, Instruction $instruction)
     {
+        $generatedCoupons = array();
         for ($i = 0, $amount = $instruction->getAmount(); $i < $amount; $i++) {
             $coupon = $this->repository->createNew();
             $coupon->setPromotion($promotion);
             $coupon->setCode($this->generateUniqueCode());
             $coupon->setUsageLimit($instruction->getUsageLimit());
+            
+            $generatedCoupons[] = $coupon;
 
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
I am in favor of this change. What do you guys think?

michalmarcinkowski:
This could be useful, but also can cause memory problems when generating big number of coupons... Not sure if we should modify this class, maybe create other class/method?

peteward:
I don't think there will be a memory issue any worse than there is already.

Since the flush happens outside of the loop, the objects will all be in memory until flushed anyway, and the array will be pointing to the same object references as are already in the Doctrine entity manager.

In many cases the returned $generatedCoupons will be ignored and the memory will be freed.  In our case, we're generating a coupon then we want to email it's code to a customer immediately, so there won't be large quantities. 

so :+1:

src/Sylius/Component/Promotion/Generator/CouponGenerator.php
```diff
@@ -43,16 +43,21 @@ public function __construct(RepositoryInterface $repository, EntityManagerInterf
      */
```
peteward: You'll need to change the doc either here or in the Interface file for the @return param

michalmarcinkowski:
@peteward my bad, you're right! :+1:

pjedrzejewski:
@agounaris Please update the docblock according to @peteward's comment and it will be ready to merge!

agounaris:
@pjedrzejewski updated

pjedrzejewski:
Great, thank you Argi and guys for the discussion! :+1:

PR #3151, opened 2015-08-18
[WebBundle] Dispatch events after building frontend menus.

jimbocoder:
The [docs state](http://docs.sylius.org/en/latest/cookbook/extending-menu.html) that extending the menus via events is "available only for the backend at the moment."  This PR hopes to fix that for everyone.  I'll update the docs accordingly if we get that far.

### DIFF
--- a/src/Sylius/Bundle/WebBundle/Event/MenuBuilderEvent.php
+++ b/src/Sylius/Bundle/WebBundle/Event/MenuBuilderEvent.php
@@ -22,8 +22,13 @@ use Symfony\Component\EventDispatcher\Event;
  */
 class MenuBuilderEvent extends Event
 {
-    const BACKEND_MAIN = 'sylius.menu_builder.backend.main';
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
We will be reworking all these menus together with new theme, but this is a nice improvement for current storefront. Thanks a lot Jim! :+1:

jimbocoder:
The least favorite question in software: got an approximate (rough, ballpark, totally made up, ..) ETA for that new stuff?  I'm curious how much effort I should expend getting us up and running on the current state of SyliusWebBundle vs focusing elsewhere on our massive migration.

pjedrzejewski:
I think we should see the new theme within 2-3 months... But many templates will be reusable, so I think you can safely build your theme on top of SyliusWebBundle. Regarding menus - you'll probably just need to change event names, because we will make them more decoupled and separated.

PR #3157, opened 2015-08-19
Associated products

lchrusciel:


lchrusciel:
Sorry, my bad. Not yet :)

peteward:
Hi @lchrusciel, 

I've just read through this, it's looking really good, certainly in terms of structure. There's a few code style issues and perhaps some translations that shouldn't be there? Otherwise this seems very close to a review.

I see from your templates that you can currently only manage associations from the product page - there's no dedicated page for mapping a large number of products to one-another - is this correct? 

If you haven't done this then it's something that we could contribute after your PR has been merged.

Let me know when you're ready for further feedback by opening up this PR and I will review code style for you (but it's already very good).

Thanks!

peteward:
Ah... Now I see the real PR, checking it !

PR #3169, opened 2015-08-21
Update routing.yml

agounaris:
Hello there,

This is a small proposal for routing name change. We've realised that using /login for both normal registered users and also social login buttons can be a bit confusing. 

This also takes into consideration that a user might add a trailing slash at the end, which currently is a "white page" on demo.sylius.org.

What do you think?

Cheers,
Argiris

### DIFF
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -13,7 +13,7 @@ sylius_web:
 
 create:
     resource: "@CmfCreateBundle/Resources/config/routing/rest.xml"
-    
+
 create_image:
     resource: "@CmfCreateBundle/Resources/config/routing/image.xml"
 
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
Definitely in favor of changing the prefix. I know I am picky, but is it always a social login? :P This is basically OAuth. Any ideas for better name? If nobody comes up with something better, then we keep social, which is quite clear. :) Good idea Argi! :+1:

peteward:
connect-login/  ?

peteward:
or connect/login/ since the other section for the routing is already under connect/

peteward:
@pjedrzejewski what do you think?

agounaris:
@pjedrzejewski @peteward updated

pjedrzejewski:
Sounds good. Thank you guys! :+1:

Issue #1708, opened 2014-07-09
Error cmf generating routes


maiksuaco:
After add a cmf block or page and visit main page sylius raise an error: it's

CRITICAL - Uncaught PHP Exception Twig_Error_Runtime: "An exception has been thrown during the rendering of a template ("None of the chained routers were able to generate route: Route 'Route aware content Symfony\Cmf\Bundle\ContentBundle\Doctrine\Phpcr\StaticContent' not found") in "SyliusWebBundle:Frontend:layout.html.twig" at line 73." at /home/maiksuaco/workspace/Sylius/vendor/twig/twig/lib/Twig/Template.php line 148 

Loading fixtures does not raise an error because new route is added and is stored in DB...I was thinking if is possible, that can be fixed making a listener and when a new page or block from cmf is inserted then insert a new route too with the new element data, following decouple principle...

psyray:
Hi @pjedrzejewski 

I think you could close this bug, it is resolved ;)

jeffsacco:
I am getting this error when I create a new page

Wondering if this issue was cleared up?

jeffsacco:
Is there some kind of work around we can do?

psyray:
see #2696

nmarmon:
Put the "closed" text if this is solved, then

Issue #2895, opened 2015-06-18
Validation not working add new slideshow


amoraru:
I've tried to set up a slideshow. Instead of get a nice error, to tell me what field should I complete, I have the following error.
I've just select image to upload and enable it. All the other fields for add slide where empty 

NodeName property "name" may not be empty in document of class "Symfony\Cmf\Bundle\BlockBundle\Doctrine\Phpcr\ImagineBlock

psyray:
@amoraru 
The internal name is mandatory
![selection_071](https://cloud.githubusercontent.com/assets/1230954/9412980/628214a8-482f-11e5-89e9-7b9a56a4601d.png)

I will fix the cascade validation of the collection form to prevent raising an exception

psyray:
Fix done, waiting for remarks or merge ;)

PR #3172, opened 2015-08-21
[ContentBundle] Fix cascade validation of collection form (issue #2895)

psyray:
Hi,

Fix for this issue https://github.com/Sylius/Sylius/issues/2895

Tested and approved ;)

### DIFF
--- a/src/Sylius/Bundle/ContentBundle/Form/Type/MenuNodeType.php
+++ b/src/Sylius/Bundle/ContentBundle/Form/Type/MenuNodeType.php
@@ -69,6 +69,7 @@ class MenuNodeType extends AbstractResourceType
             ))
             ->add('content', 'sylius_static_content_choice', array(
                 'label' => 'sylius.form.menu_node_block.content',
+                'property' => 'title',
              ))
             ->add('uri', null, array(
                 'label' => 'sylius.form.menu_node_block.uri'
--- a/src/Sylius/Bundle/ContentBundle/Form/Type/MenuType.php
+++ b/src/Sylius/Bundle/ContentBundle/Form/Type/MenuType.php
@@ -47,6 +47,7 @@ class MenuType extends AbstractResourceType
                 'by_reference'     => false,
                 'label'            => false,
                 'button_add_label' => 'sylius.form.menu.add_menu_node',
+                'cascade_validation' => true,
              ))
             ->add('uri', null, array(
                 'required' => false,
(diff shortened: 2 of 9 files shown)


pjedrzejewski:
@psyray Thank you for the patch Raynald! Could you please have a look at the failing build? Thanks!

psyray:
Validation complains about the id that I removed from the form generation on this line
https://github.com/Sylius/Sylius/pull/3172/files#diff-db52331bf73c6ccbca0722ad103e1738L33

The purpose of this remove was this error
selection_072

That occurs when a user update an existing static content, click the add route button, and save without filling the route fields.
Like this
selection_073

I remove the id field from the form generation and validation works fine
selection_074

But while doing some tests, I've seen that the embedded route form need more validation, I explain :
- While you create a static content, the parent must be filled (don't know why but the embedded form does not use the read-only field like in the classic route form)
selection_075

- While you create a static content and specifying a route, the dropdown list with the static content is displayed. This is not good because the static content linked to this route must be the current one and not another one, which cause conflict while saving form

So we need to decide, 
- Keeping the embedded form ?
- Exclusively set route in the route section ?
- Polish route validation in the static content form ?

It's up to you @pjedrzejewski ;)

pjedrzejewski:
@psyray I'd go with the first option and remove the embedded form. At least we will make it work again and then we will try to come up with something more user friendly. Sounds like a plan?

psyray:
Done.
Embedded form removed
selection_077

I took the opportunity to **correct some bugs** and **add some more informations to the user** to make it more friendly and prevent mistakes.
I don't create a PR for each, that's good for you ?

**Corrected bugs:**
- Prevent static content element to display in frontend home page if route is not set (should fix #1708 and don't break demo website anymore with exception *"None of the chained routers were able to generate route: Route 'Route aware content ...."*)
- Display title in form content choice instead of ugly id (menu and route)
selection_078

**Informations to user:**
- Display routes attached to a page in Pages list
selection_076

**Question about content choice list !**
Do you know how to sort the content choice list ?
I don't know how to do with ResourceBundle (as choice list is generated by it)
Should I use a classic 'query_builder' parameter with anonymous function or is there an easier way ?
Maybe @aRn0D could help ?

Enjoy ;)

psyray:
up !

pjedrzejewski:
Thank you Raynald! :+1: Great work. Could you drop me an e-mail to pawel@sylius.org. I wanted to discuss something real quick, thanks!

psyray:
email dropped yesterday, did you receive it ?

pjedrzejewski:
@psyray Yes, I will get back to you soon. :) Thank you!

Issue #3167, opened 2015-08-21
Account password change doesn't work


gonzalovilaseca:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Password change in account doesn't work. 
For what I've found the problem seems to be in the User controller:
```
    protected function handleChangePassword(UserInterface $user, $newPassword)
    {
        $user->setPlainPassword($newPassword);

        $this->domainManager->update($user);

        $dispatcher = $this->get('event_dispatcher');
        $dispatcher->dispatch(UserEvents::PASSWORD_RESET_SUCCESS, new GenericEvent($user));

        if ($this->config->isApiRequest()) {
            return $this->handleView($this->view($user, 204));
        }
        $this->addFlash('success', 'sylius.user.password.change.success');

        return new RedirectResponse($this->generateUrl('sylius_account_homepage'));
    }
```
The plain password is encoded by a listener of the ``PASSWORD_RESET_SUCCESS`` event, but the entity is persisted before with `` $this->domainManager->update($user);``. So at the point where the entity is persisted, the encoded password hasn't been updated.

If this sounds reasonable I'll open a PR to fix it and also check that the reset works properly.

michalmarcinkowski:
@gonzalovilaseca actually we should have two events there (pre and post) as it is in resource. If you want to make a PR, please split this event into:

```
PRE_PASSWORD_RESET
POST_PASSWORD_RESET
```
The first one should be dispatched before flush, the second one after (replaces SUCCESS). Then attach the password updater to PRE event and everything should work fine.

If you have more time you could do the same for password change events (they are also wrong...).

Thank you @gonzalovilaseca for reporting that! :+1:

michalmarcinkowski:
BTW.
In method `handleChangePassword` we should dispatch `PRE_PASSWORD_CHANGE` and `POST_PASSWORD_CHANGE`. I've already corrected it in my UserBundle improvement (WIP - https://github.com/michalmarcinkowski/Sylius/pull/6/files#diff-4f4c5540de8e56b05f4e9184911081bdL244).

gonzalovilaseca:
@michalmarcinkowski Ok, will do a PR soon.

michalmarcinkowski:
Great!

jimbocoder:
[...]

PR #3173, opened 2015-08-22
[UserBundle] Dispatch user event *before* persisting user, so side effects persist too.

jimbocoder:
The immediate fix for #3167 while more complete fix is baked.


pjedrzejewski:
I have restarted the failing build.

pjedrzejewski:
Thank you Jim! :+1:

PR #3176, opened 2015-08-24
[Behat] Steps for checking whether row contains given value (not starts with)

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

**TL;DR**: You can add `%` at the beginning and the end of searched string and it will work just like SQL `LIKE` statement (no support for one-side `%` yet).

- `I should see resource with name containing "value"` is equal to `I should see resource with name "%value%"` (same comes with negation)
- `I should see the following rows: | name | %value% |` (same comes with negation)

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Behat/WebContext.php
+++ b/src/Sylius/Bundle/ResourceBundle/Behat/WebContext.php
@@ -42,7 +42,6 @@ class WebContext extends DefaultContext
         }
     }
 
-
     /**
      * @Given /^I am on the page of ([^""]*) with ([^""]*) "([^""]*)"$/
      * @Given /^I go to the page of ([^""]*) with ([^""]*) "([^""]*)"$/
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ResourceBundle/Behat/WebContext.php
```diff
@@ -277,6 +277,36 @@ public function iShouldNotSeeResourceWithValueInThatList($columnName, $value)
     }
 
     /**
+     * For example: I should see product with name containing "Wine X" in that list.
+     *
+     * @Then /^I should see (?:(?!enabled|disabled)[\w\s]+) with ([\w\s]+) containing "([^""]*)" in (?:that|the) list$/
```
pjedrzejewski: Poryte regexpy.

pamil: There's worse one for that **not** contaning :)

pjedrzejewski:
Thank you Kamil! :+1:

PR #3180, opened 2015-08-25
[Support] Ticket state machine

GSadee:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | no
| License       | MIT
| Doc PR        | no

Added states with state machine to Support Ticket


UPGRADE.md
```diff
@@ -3,6 +3,44 @@ UPGRADE
...
 
+### Support
+
+Contact component has been renamed to Support.
+ContactBundle has been renamed to SupportBundle.
+
+Model classes and interfaces have been renamed.
+
+* Request becomes Ticket.
+* Topic becomes Category.
```
aramalipoor: Maybe a more specific name like `SupportCategory` would be better. Category is very generic and can be confused.

pjedrzejewski: Yeah, we were debating that. Decision was to keep things simple and name it Category. Now I start to think that maybe these few more keystrokes is worth making things clear. :D

PR #3195, opened 2015-09-01
Updating translations from Crowdin

paceto256:
Translations I previously translated in Crowdin, later send to me by Paweł Jędrzejewski.


pjedrzejewski:
I think these won't work. @jjanvier Can we do something about crowdin sync? Does it work right now if I want to run it?

PR #3055, opened 2015-07-17
[Travis] Running JavaScript tests

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3241
| License       | MIT

Main changes:
 - [x] Javascript based scenarios are executed since now
 - [x] Repaired a few old and rusty Behat scenarios
 - [x] Fixed #3241

Changes made by-the-way:
 - [x] MongoDB won't install if not needed
 - [x] PHP7 tests are running now (and passing, PhpSpec omitted) :tada: 
 - [x] On failure, logs and screenshots will be provided (based on [Lakion/MinkDebugExtension](https://github.com/Lakion/MinkDebugExtension))
 - [x] Added `--strict` toggle to every Behat execution (it won't pass if there are pending or missing steps)
 - [x] Introduced caching (speeds each job up for ~2-3 minutes )
 - [x] Cleaned up Travis a little bit :)

PR #3216, opened 2015-09-03
Adjust .travis.yml file to use cache + more stable mongodb-odm versions

stloyd:


pjedrzejewski:
@stloyd IIRC we use --prefer-source to avoid github API limits. Or am I wrong? I do not remember. :P

pamil:
Cache was added by the way in #3055, `doctrine/mongodb-odm` version is set as stable in #3287, I guess it can be closed.

PR #3217, opened 2015-09-03
[Search] Refactored PropertyAccessor usage

### LINKED ISSUE #3211: Sylius\Bundle\SearchBundle\Accessor\PropertyAccessor doesn't do anything :)
`Sylius\Bundle\SearchBundle\Accessor\PropertyAccessor` is badly written (it creates property accessors based on FQCN, then iterates through only the first one (returns its value or returns null if there's no such property) and then has dead code block.

It is only used by `sylius.search.transformers.model.product` service (`FOS\ElasticaBundle\Transformer\ModelToElasticaAutoTransformer`) as a proxy for `Sylius\Component\Product\Accessor\ProductAccessor`.

To sum it up, we can delete it and replace with `ProductAccessor` and application behaviour won't change, but less code is better code :)

What do you think?

pamil:
Fixes #3211.


pjedrzejewski:
Thanks Kamil! :+1:

PR #3226, opened 2015-09-06
[Channel] Implement toggleable interface on channels.

gperdomor:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

Implement toggleable interface on channels
Fix phpspec directory on channel component.
Following the behaviour of countries, the channels cannot be deleted


src/Sylius/Bundle/WebBundle/Resources/config/routing/backend/channel.yml
```diff
@@ -30,11 +30,18 @@ sylius_backend_channel_update:
...
-            template: SyliusWebBundle:Backend/Misc:delete.html.twig
             redirect: sylius_backend_channel_index
+
+sylius_backend_channel_disable:
+    path: /{id}/disable
+    methods: [PATCH]
+    defaults:
+        _controller: sylius.controller.channel:disableAction
+        _sylius:
+            redirect: sylius_backend_channel_index
```
michalmarcinkowski: Missing blank line at the end of a file.

src/Sylius/Component/Channel/Model/Channel.php
```diff
@@ -187,7 +187,23 @@ public function isEnabled()
      */
     public function setEnabled($enabled)
     {
-        $this->enabled = (Boolean) $enabled;
+        $this->enabled = (boolean) $enabled;
```
michalmarcinkowski:
```php
(bool) $enabled
```

src/Sylius/Component/Channel/spec/Model/ChannelSpec.php
```diff
@@ -66,13 +71,34 @@ function its_color_is_mutable()
 
     function it_is_enabled_by_default()
     {
-        $this->shouldBeEnabled();
+        $this->isEnabled()->shouldReturn(true);
```
michalmarcinkowski: I think `$this->shouldBeEnabled()` and `$this->shouldNotBeEnabled()` is more readable.

src/Sylius/Component/Channel/Model/Channel.php
```diff
@@ -64,7 +64,7 @@ class Channel implements ChannelInterface
     protected $updatedAt;
 
     /**
-     * Constructor
+     * Constructor.
```
pamil: It's redundant.

src/Sylius/Bundle/ChannelBundle/Behat/ChannelContext.php
```diff
@@ -144,15 +145,35 @@ public function theDefaultChannelHasFollowingConfiguration(TableNode $table)
...
+    {
+        $this->thereIsChannel($name, $name, null, false);
+    }
+
+    /**
+     * @Given /^there is an enabled channel "([^""]*)"$/
+     */
+    public function thereIsEnabledChannel($name)
+    {
+        $this->thereIsChannel($name, $name, null, true);
```
stloyd: `true` is the default value so it's redundant here.

src/Sylius/Bundle/ChannelBundle/Behat/ChannelContext.php
```diff
@@ -144,15 +145,35 @@ public function theDefaultChannelHasFollowingConfiguration(TableNode $table)
...
      */
-    public function thereIsChannel($code, $name, $url, $locales = null, $currencies = "EUR", $shippingMethods = null, $paymentMethos = null, $flush = true)
+    public function thereIsChannel($code, $name, $url, $enabled = true, $locales = null, $currencies = "EUR", $shippingMethods = null, $paymentMethos = null, $flush = true)
     {
         /* @var $channel ChannelInterface */
-        $channel = $this->getRepository('channel')->createNew();
-        $channel->setCode($code);
-        $channel->setName($name);
-        $channel->setUrl($url);
+        if (null === $channel = $this->getRepository('channel')->findOneBy(array('code' => $code))) {
```
stloyd: `null ===` can be removed.

src/Sylius/Component/Channel/phpspec.yml
```diff
@@ -0,0 +1,5 @@
+suites:
+    main:
+        namespace: Sylius\Component\Channel
+        psr4_prefix: Sylius\Component\Channel
```
stloyd: :+1:

src/Sylius/Bundle/ChannelBundle/Behat/ChannelContext.php
michalmarcinkowski: I wouldn't rely on model default value and set it explicitly as it is now.

src/Sylius/Bundle/ChannelBundle/Behat/ChannelContext.php
michalmarcinkowski: But shouldn't. We should be as explicit as possible to make code more direct less guestwork (maybe it returns `false`?).

stloyd: @michalmarcinkowski Then that check would be still totally correct... Thing I missed to mention is `null === $channel` should be replaced with `!$channel`... And such comparsion is widely used in Sylius (IIRC) as well as in Symfony.

pamil: :+1: for explicit comparisons, while using PHP you can never be sure what converting to boolean returns :)

michalmarcinkowski: Casting object to boolean is not a good way to go... The PHP magic and sometimes illogical convertions allows for that, but it's not something we want to do any more. 

The question: What is a more readable check for zero integer?
```php
if (0 === $value)
```
or
```php
if (!$value)
```

stloyd: Using strict comparsion just for sake of using is wrong anyway, if that method will return `false`, `0`, empty string/array, that code will die with fatal error instead of correctly creating new channel...

@michalmarcinkowski Going further with your thinking, what's more redeable:
```php
if (!$value) {}
// vs
if (null === $value || 0 === $value || false === $value || '' === $value/* etc. */) {}
```
:stuck_out_tongue_winking_eye:

michalmarcinkowski: If that method will return any other value than `ChannelInterface` or `null` that means it's wrong method! The interface should allow only for one return type (like in every strong typehinted language).

The most readable and the only correct check is:
```php
if (0 === $value)
```

pamil: @stloyd if I want to check against everything, I would use `empty()` language construct. In this case we can only expect that it will return `null` if no channel found and something else if it was found. 

If it returns `false`, then the fatal error should appear, because in fact - there's an error :)

src/Sylius/Component/Channel/Model/Channel.php
gperdomor: Yes but it's for passing the <code>missing doc block</code> rule of php code sniffer

src/Sylius/Bundle/ChannelBundle/Behat/ChannelContext.php
gperdomor: This validation is used in other classes of Sylius project. If we wanna more readable comparisons of null values maybe we should use <code>is_null</code> function, and define as an standard inside project changing it in all classes.

gperdomor:
ping @pjedrzejewski

michalmarcinkowski:
It's already implemented :)

PR #2224, opened 2014-12-05
[WIP ]Adding thank you page in order process

antonioperic:
I talked with @pjedrzejewski about that we need to have really order complete/thankyou page and not redirect to homepage with flash message.

I started to adding this step, do you think this is good way, or maybe we should pull in some other way?

In order page will be message about status and some order information

Thanks!

PR #3230, opened 2015-09-07
[Core] Redirect customer to order tracking page after checkout completion

aramalipoor:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | no |
| License | MIT |
| Doc PR | - |

I wanted to redirect customer to a thank you page after order is placed successfully. With a quick overview I thought maybe we can redirect him to an Order Tracking (`order/track/{number}`) page.

Note that for now I had to [set buying _customer_ into current context](https://github.com/Sylius/Sylius/pull/3230/files#diff-79357f2667866d43b41932fa092a9746R43), so that I could check their authentication when doing _guest checkout_.

BTW If we generate order numbers using [hash generator](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Sequence/Number/HashGenerator.php) we can offer **order tracking without authentication** feature.

Related to #2224

### DIFF
--- a/features/frontend/checkout_finalize.feature
+++ b/features/frontend/checkout_finalize.feature
@@ -42,12 +42,8 @@ Feature: Checkout finalization
           And I select the "Dummy" radio button
           And I press "Continue"
          When I click "Place order"
-         Then I should be on the store homepage
+         Then I should be on the order tracking page for 000000001
           And I should see "Thank you for your order!"
-          And I am on my account orders page
-          And I should see "All your orders"
-          And I should see 1 orders in the list
-          And I should see "000000001"
 
     Scenario: Placing the order as Guest without email address
         Given I am not logged in
@@ -85,5 +81,5 @@ Feature: Checkout finalization
           And I select the "Dummy" radio button
           And I press "Continue"
          When I click "Place order"
-         Then I should be on the store homepage
+         Then I should be on the order tracking page for 000000001
           And I should see "Thank you for your order!"
--- a/src/Sylius/Bundle/CoreBundle/Checkout/CheckoutProcessScenario.php
+++ b/src/Sylius/Bundle/CoreBundle/Checkout/CheckoutProcessScenario.php
@@ -59,7 +59,10 @@ class CheckoutProcessScenario implements ProcessScenarioInterface
         $builder
             ->setDisplayRoute('sylius_checkout_display')
             ->setForwardRoute('sylius_checkout_forward')
-            ->setRedirect('sylius_homepage')
+            ->setRedirect('sylius_order_track')
+            ->setRedirectParams(array(
+                'number' => $cart->getNumber()
+            ))
             ->validate(function () use ($cart) {
                 return !$cart->isEmpty();
             })
--- a/src/Sylius/Bundle/CoreBundle/spec/Checkout/CheckoutProcessScenarioSpec.php
+++ b/src/Sylius/Bundle/CoreBundle/spec/Checkout/CheckoutProcessScenarioSpec.php
@@ -39,20 +39,25 @@ class CheckoutProcessScenarioSpec extends ObjectBehavior
         $this->shouldImplement('Sylius\Bundle\FlowBundle\Process\Scenario\ProcessScenarioInterface');
     }
 
-    function it_builds_checkout_process_with_proper_steps(ProcessBuilderInterface $builder)
+    function it_builds_checkout_process_with_proper_steps(ProcessBuilderInterface $builder, CartInterface $cart)
     {
-        $builder->add('security', 'sylius_checkout_security')->willReturn($builder)->shouldBeCalled();
-        $builder->add('addressing', 'sylius_checkout_addressing')->willReturn($builder)->shouldBeCalled();
-        $builder->add('shipping', 'sylius_checkout_shipping')->willReturn($builder)->shouldBeCalled();
-        $builder->add('payment', 'sylius_checkout_payment')->willReturn($builder)->shouldBeCalled();
-        $builder->add('finalize', 'sylius_checkout_finalize')->willReturn($builder)->shouldBeCalled();
-        $builder->add("purchase", "sylius_checkout_purchase")->willReturn($builder)->shouldBeCalled();
+        $cart->getNumber()->willReturn('000000001');
 
-        $builder->setDisplayRoute('sylius_checkout_display')->willReturn($builder)->shouldBeCalled();
-        $builder->setForwardRoute('sylius_checkout_forward')->willReturn($builder)->shouldBeCalled();
+        $builder->add('security', 'sylius_checkout_security')->willReturn($builder);
+        $builder->add('addressing', 'sylius_checkout_addressing')->willReturn($builder);
+        $builder->add('shipping', 'sylius_checkout_shipping')->willReturn($builder);
+        $builder->add('payment', 'sylius_checkout_payment')->willReturn($builder);
+        $builder->add('finalize', 'sylius_checkout_finalize')->willReturn($builder);
+        $builder->add('purchase', 'sylius_checkout_purchase')->willReturn($builder);
 
-        $builder->setRedirect('sylius_homepage')->willReturn($builder)->shouldBeCalled();
-        $builder->validate(Argument::any())->willReturn($builder)->shouldBeCalled();
+        $builder->setDisplayRoute('sylius_checkout_display')->willReturn($builder);
+        $builder->setForwardRoute('sylius_checkout_forward')->willReturn($builder);
+
+        $builder->setRedirect('sylius_order_track')->willReturn($builder);
+        $builder->setRedirectParams(array(
+            'number' => '000000001'
+        ))->willReturn($builder);
+        $builder->validate(Argument::any())->willReturn($builder);
 
         $this->build($builder);
     }
--- a/src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
+++ b/src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
@@ -14,6 +14,10 @@ namespace Sylius\Bundle\UserBundle\Context;
 use Sylius\Component\User\Context\CustomerContextInterface;
 use Sylius\Component\User\Model\CustomerInterface;
 use Sylius\Component\User\Model\UserInterface;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\Security\Core\Authentication\AuthenticationManagerInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 use Symfony\Component\Security\Core\SecurityContextInterface;
 
 /**
@@ -22,29 +26,70 @@ use Symfony\Component\Security\Core\SecurityContextInterface;
 class CustomerContext implements CustomerContextInterface
 {
     /**
-     * @var SecurityContextInterface
+     * @var TokenStorageInterface
      */
-    private $securityContext;
+    protected $tokenStorage;
 
     /**
-     * @param SecurityContextInterface $securityContext
+     * @var AuthorizationCheckerInterface
      */
-    public function __construct(SecurityContextInterface $securityContext)
-    {
-        $this->securityContext = $securityContext;
+    protected $authorizationChecker;
+
+    /**
+     * @var SessionInterface
+     */
+    protected $session;
+
+    /**
+     * @param TokenStorageInterface         $tokenStorage
+     * @param AuthorizationCheckerInterface $authorizationChecker
+     * @param SessionInterface              $session
+     */
+    public function __construct(
+        TokenStorageInterface $tokenStorage,
+        AuthorizationCheckerInterface $authorizationChecker,
+        SessionInterface $session
+    ) {
+        $this->tokenStorage = $tokenStorage;
+        $this->authorizationChecker = $authorizationChecker;
+        $this->session = $session;
     }
 
     /**
-     * Gets customer based on currently logged user.
+     * Gets customer based on currently logged user or from session (if user is not authenticated).
      *
      * @return CustomerInterface|null
      */
     public function getCustomer()
     {
-        if ($this->securityContext->getToken() && $this->securityContext->isGranted('IS_AUTHENTICATED_REMEMBERED')
-            && $this->securityContext->getToken()->getUser() instanceof UserInterface
-        ) {
-            return $this->securityContext->getToken()->getUser()->getCustomer();
+        $user = $this->getUser();
+
+        if (null !== $user && $user instanceof UserInterface) {
+            return $user->getCustomer();
+        }
+
+        return $this->session->get('customer');
+    }
+
+    /**
+     * Stores customer in session.
+     *
+     * @param CustomerInterface $customer
+     */
+    public function setCustomer(CustomerInterface $customer)
+    {
+        $this->session->set('customer', $customer);
+    }
+
+    /**
+     * @return UserInterface|null
+     */
+    protected function getUser()
+    {
+        if ($this->authorizationChecker->isGranted('IS_AUTHENTICATED_REMEMBERED')) {
+            if ($token = $this->tokenStorage->getToken()) {
+                return $token->getUser();
+            }
         }
 
         return null;
--- a/src/Sylius/Bundle/UserBundle/spec/EventListener/CustomerContextListenerSpec.php
+++ b/src/Sylius/Bundle/UserBundle/spec/EventListener/CustomerContextListenerSpec.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace spec\Sylius\Bundle\UserBundle\EventListener;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\Bundle\UserBundle\Context\CustomerContext;
+use Sylius\Component\User\Model\CustomerAwareInterface;
+use Sylius\Component\User\Model\CustomerInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+class CustomerContextListenerSpec extends ObjectBehavior
+{
+    function let(CustomerContext $customerContext)
+    {
+        $this->beConstructedWith($customerContext);
+    }
+
+    function it_is_initializable()
+    {
+        $this->shouldHaveType('Sylius\Bundle\UserBundle\EventListener\CustomerContextListener');
+    }
+
+    function it_sets_subjects_customer_in_context(
+        $customerContext,
+        GenericEvent $event,
+        CustomerAwareInterface $subject,
+        CustomerInterface $customer
+    ) {
+        $event->getSubject()->willReturn($subject);
+        $subject->getCustomer()->willReturn($customer);
+        $customerContext->setCustomer($customer)->shouldBeCalled();
+
+        $this->setCustomerContextFromSubject($event);
+    }
+
+    function it_throws_exception_if_event_subject_is_not_a_customer_aware(GenericEvent $event)
+    {
+        $event->getSubject()->willReturn(new \stdClass());
+
+        $this
+            ->shouldThrow('Sylius\Component\Resource\Exception\UnexpectedTypeException')
+            ->duringSetCustomerContextFromSubject($event)
+        ;
+    }
+}
--- a/src/Sylius/Bundle/WebBundle/Behat/WebContext.php
+++ b/src/Sylius/Bundle/WebBundle/Behat/WebContext.php
@@ -462,7 +462,13 @@ class WebContext extends BaseWebContext implements SnippetAcceptingContext
     {
         $order = $this->findOneBy('order', array('number' => $number));
 
-        $this->getSession()->visit($this->generatePageUrl('sylius_account_order_'.$action, array('number' => $order->getNumber())));
+        if ('tracking' === $action) {
+            $route = 'sylius_order_track';
+        } else {
+            $route = 'sylius_account_order_'.$action;
+        }
+
+        $this->getSession()->visit($this->generatePageUrl($route, array('number' => $order->getNumber())));
     }
 
     /**
@@ -471,7 +477,13 @@ class WebContext extends BaseWebContext implements SnippetAcceptingContext
      */
     public function iShouldBeOnTheOrderPage($action, $number)
     {
-        $this->assertSession()->addressEquals($this->generatePageUrl('sylius_account_order_'.$action, array('number' => $number)));
+        if ('tracking' === $action) {
+            $route = 'sylius_order_track';
+        } else {
+            $route = 'sylius_account_order_'.$action;
+        }
+
+        $this->assertSession()->addressEquals($this->generatePageUrl($route, array('number' => $number)));
     }
 
     /**
--- a/src/Sylius/Bundle/WebBundle/Resources/translations/messages.en.yml
+++ b/src/Sylius/Bundle/WebBundle/Resources/translations/messages.en.yml
@@ -2,6 +2,7 @@
 # (c) Paweł Jędrzejewski
 
 sylius:
+    back_to_shop: Back to shop
     emails:
         order_confirmation:
             name: Order confirmation
--- a/src/Sylius/Bundle/WebBundle/Resources/views/Frontend/Account/layout.html.twig
+++ b/src/Sylius/Bundle/WebBundle/Resources/views/Frontend/Account/layout.html.twig
@@ -12,5 +12,12 @@
 {% endblock %}
 
 {% block sidebar %}
-{{ knp_menu_render('sylius.frontend.account', {'template': 'SyliusWebBundle:Frontend:menu.html.twig'}) }}
+    {% if is_granted('IS_AUTHENTICATED_REMEMBERED') %}
+        {{ knp_menu_render('sylius.frontend.account', {'template': 'SyliusWebBundle:Frontend:menu.html.twig'}) }}
+    {% else %}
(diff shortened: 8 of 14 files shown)


src/Sylius/Bundle/CoreBundle/spec/Checkout/CheckoutProcessScenarioSpec.php
```diff
@@ -51,7 +52,10 @@ function it_builds_checkout_process_with_proper_steps(ProcessBuilderInterface $b
         $builder->setDisplayRoute('sylius_checkout_display')->willReturn($builder)->shouldBeCalled();
         $builder->setForwardRoute('sylius_checkout_forward')->willReturn($builder)->shouldBeCalled();
 
-        $builder->setRedirect('sylius_homepage')->willReturn($builder)->shouldBeCalled();
+        $builder->setRedirect('sylius_account_order_show')->willReturn($builder)->shouldBeCalled();
+        $builder->setRedirectParams(array(
+            'number' => $order->getNumber()
+        ))->willReturn($builder)->shouldBeCalled();
```
stloyd: All `shouldBeCalled()` calls should be removed as you already cover cases of calls with promises done by stubs (`willReturn()`). 

Calling those two in same place is waste of resources & _just wrong_, cause you first test with promises & later do almost same with predictions...

src/Sylius/Bundle/CoreBundle/Checkout/CheckoutProcessScenario.php
```diff
@@ -59,7 +59,10 @@ public function build(ProcessBuilderInterface $builder)
         $builder
             ->setDisplayRoute('sylius_checkout_display')
             ->setForwardRoute('sylius_checkout_forward')
-            ->setRedirect('sylius_homepage')
+            ->setRedirect('sylius_account_order_show')
+            ->setRedirectParams(array(
+                'number' => $cart->getNumber()
+            ))
```
stloyd: Please remove all fluent interface usage as in other PRs it's being done.

pjedrzejewski: This is a builder...

stloyd: Fluent everywhere is evil :smiling_imp:

src/Sylius/Bundle/CoreBundle/spec/Checkout/CheckoutProcessScenarioSpec.php
michalmarcinkowski: Actually we do want to check whether the method was called with this arguments or not, so the `shouldBeCalled()` is necessary here, because `willReturn` doesn't validate that method was called.

src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
```diff
@@ -40,10 +51,34 @@ public function __construct(SecurityContextInterface $securityContext)
...
+
+    /**
+     * Stores customer in session (if user is not authenticated) or on currently logged in user.
+     *
+     * @param CustomerInterface $customer
+     */
+    public function setCustomer(CustomerInterface $customer)
+    {
+        if ($user = $this->getUser()) {
+            $user->setCustomer($customer);
```
michalmarcinkowski: The `User` object shouldn't be able to exist without `Customer` and the user's customer should never be changed (setter is only used during creation and should never be used after the user was created).

src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
```diff
@@ -40,10 +51,34 @@ public function __construct(SecurityContextInterface $securityContext)
      */
     public function getCustomer()
     {
-        if ($this->securityContext->getToken() && $this->securityContext->isGranted('IS_AUTHENTICATED_REMEMBERED')
-            && $this->securityContext->getToken()->getUser()
-        ) {
-            return $this->securityContext->getToken()->getUser()->getCustomer();
+        if ($user = $this->getUser()) {
```
michalmarcinkowski: That is assignment, not check :wink:

features/frontend/checkout_finalize.feature
```diff
@@ -85,5 +81,5 @@ Feature: Checkout finalization
           And I select the "Dummy" radio button
           And I press "Continue"
          When I click "Place order"
-         Then I should be on the store homepage
+         Then I should be on the order show page for 000000001
           And I should see "Thank you for your order!"
```
michalmarcinkowski: Missing blank line.

src/Sylius/Bundle/CoreBundle/spec/Checkout/CheckoutProcessScenarioSpec.php
stloyd: This is wrong, cause [mock (aka predictions)](http://www.phpspec.net/en/latest/manual/prophet-objects.html#mocks) should be called __only__ when you don't care about what is being returned by the method, if you care about it you should call [stubs (aka promises)](http://www.phpspec.net/en/latest/manual/prophet-objects.html#stubs).

As said calling both is wrong & mixing behavior of two different things.

michalmarcinkowski:
I think it can be useful to store customer in a session. Adding `setCustomer` method on context is fine, but it should only set customer in a session, never assign it to an existing user.

The only issue I see here is that the order details page will be only accessible for guest customers just after the checkout when the customer is in session. We should think about possibility to display order details for guest customers, but I think it's a topic for a different PR :)

src/Sylius/Bundle/CoreBundle/spec/Checkout/CheckoutProcessScenarioSpec.php
aramalipoor: I'm not a phpspec expert but I think @stloyd is right, I've removed `shouldBeCalled`s and still all calls are expected as normal.

src/Sylius/Bundle/CoreBundle/Checkout/CheckoutProcessScenario.php
aramalipoor: Agreed with @pjedrzejewski they can be nice when used in builders... :D

src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
aramalipoor: You're right :+1:

aramalipoor:
Thanks everyone for useful comments. 

@michalmarcinkowski I had this in mind too, to make it possible to track an order with a tracking code and without the need for authentication. First I tried to change order number generator to `hash` so that it would be hard to guess by a hacker then just remove security check on order details controller; but then I realized that could be a bad thing.

What do you think? We are going to need this in our project so if it's OK I can come up with another PR for that.

features/frontend/checkout_finalize.feature
aramalipoor: Where exactly?!

michalmarcinkowski: There was a change or I saw something that wasn't here :smile:

aramalipoor: Holy moly :smile: OK.

src/Sylius/Bundle/CoreBundle/spec/Checkout/CheckoutProcessScenarioSpec.php
michalmarcinkowski: Without `shouldBeCalled` there is no assertion, so the spec will pass everytime, even when the `build` method will be empty. Why we want to have spec that doesn't validate anything?

src/Sylius/Bundle/CoreBundle/EventListener/CheckoutSecurityListener.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * @param GenericEvent $event
+     */
+    public function setCustomerInContext(GenericEvent $event)
+    {
+        $order = $event->getSubject();
+
+        if (!$order instanceof OrderInterface) {
```
stloyd: Use `CustomerAwareInterface` instead.

src/Sylius/Bundle/CoreBundle/EventListener/CheckoutSecurityListener.php
```diff
@@ -0,0 +1,61 @@
...
+namespace Sylius\Bundle\CoreBundle\EventListener;
+
+use Sylius\Bundle\UserBundle\Context\CustomerContext;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Model\CustomerInterface;
+use Sylius\Component\Resource\Exception\UnexpectedTypeException;
+use Symfony\Component\EventDispatcher\GenericEvent;
+
+/**
+ * Set customer in context, used when un-authenticated users are in checkout process.
```
stloyd: This is not really true, as this is __listener__ and you can set it on many other events than checkout.

src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
```diff
@@ -23,27 +25,58 @@ class CustomerContext implements CustomerContextInterface
...
-            && $this->securityContext->getToken()->getUser()
-        ) {
-            return $this->securityContext->getToken()->getUser()->getCustomer();
+        $user = $this->getUser();
+
+        if (null !== $user) {
+            return $user->getCustomer();
+        }
+
+        return $this->session->get('customer', null);
```
stloyd: IIRC `null` is default value when not finding data in bags.

aramalipoor:
Thanks @stloyd for your feedback.

src/Sylius/Bundle/CoreBundle/EventListener/CustomerContextListener.php
```diff
@@ -0,0 +1,61 @@
...
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\CoreBundle\EventListener;
```
stloyd: Why this is in `CoreBundle` and not in `UserBundle`? It doesn't really depend on anything from `Core` now.

src/Sylius/Bundle/CoreBundle/spec/Checkout/CheckoutProcessScenarioSpec.php
```diff
@@ -39,20 +39,25 @@ function it_implements_Sylius_process_scenario_interface()
...
-        $builder->setRedirect('sylius_homepage')->willReturn($builder)->shouldBeCalled();
-        $builder->validate(Argument::any())->willReturn($builder)->shouldBeCalled();
+        $cart->getNumber()->willReturn('000000001');
+
+        $builder->add('security', 'sylius_checkout_security')->willReturn($builder);
+        $builder->add('addressing', 'sylius_checkout_addressing')->willReturn($builder);
+        $builder->add('shipping', 'sylius_checkout_shipping')->willReturn($builder);
+        $builder->add('payment', 'sylius_checkout_payment')->willReturn($builder);
+        $builder->add('finalize', 'sylius_checkout_finalize')->willReturn($builder);
+        $builder->add("purchase", "sylius_checkout_purchase")->willReturn($builder);
```
stloyd: `"` vs `'` :wink:

src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
```diff
@@ -23,27 +25,58 @@ class CustomerContext implements CustomerContextInterface
...
-            && $this->securityContext->getToken()->getUser()
-        ) {
-            return $this->securityContext->getToken()->getUser()->getCustomer();
+        $user = $this->getUser();
+
+        if (null !== $user) {
+            return $user->getCustomer();
+        }
+
+        return $this->session->get('customer');
```
stloyd: Specs should be extended to cover this case.

src/Sylius/Bundle/CoreBundle/spec/Checkout/CheckoutProcessScenarioSpec.php
```diff
@@ -39,20 +39,25 @@ function it_implements_Sylius_process_scenario_interface()
...
+
+        $builder->setDisplayRoute('sylius_checkout_display')->willReturn($builder);
+        $builder->setForwardRoute('sylius_checkout_forward')->willReturn($builder);
+
+        $builder->setRedirect('sylius_account_order_show')->willReturn($builder);
+        $builder->setRedirectParams(array(
+            'number' => '000000001'
+        ))->willReturn($builder);
+        $builder->validate(Argument::any())->willReturn($builder);
```
michalmarcinkowski: I agree that we should build specs in such a way that we use either `mock` or `stub`, but with the builder's fluent interface it's not possible. Without `shouldBeCalled` there is no assertion, so the spec will pass everytime, even when the `build` method will be empty. Why we want to have spec that doesn't validate anything?

src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
michalmarcinkowski: :+1:

app/config/security.yml
```diff
@@ -79,6 +79,7 @@ security:
         - { path: ^/administration/login-check, role: IS_AUTHENTICATED_ANONYMOUSLY }
         - { path: "/administration.*", role: ROLE_ADMINISTRATION_ACCESS }
 
+        - { path: "/account/orders/[0-9-]*", role: IS_AUTHENTICATED_ANONYMOUSLY }
```
michalmarcinkowski: I don't think we want to do that... Maybe add other route for that case? (we could reuse the controller action)

src/Sylius/Bundle/CoreBundle/spec/Checkout/CheckoutProcessScenarioSpec.php
stloyd: Problem here is not that it's not using mock, problem is that test allows to pass anything while it should be argument of type or even just an interface/closure, this should be checked IMO.

michalmarcinkowski: The problem here is that there is nothing checked at all :wink:

src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
aramalipoor: Done ;)

src/Sylius/Bundle/CoreBundle/spec/Checkout/CheckoutProcessScenarioSpec.php
aramalipoor: So should I revert `shouldBeCalled`s?

src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
```diff
@@ -23,27 +25,58 @@ class CustomerContext implements CustomerContextInterface
...
+    }
+
+    /**
+     * Stores customer in session.
+     *
+     * @param CustomerInterface $customer
+     */
+    public function setCustomer(CustomerInterface $customer)
+    {
+        $this->session->set('customer', $customer);
```
stloyd: What if someone logs out and then logs in as different user on same browser*? :)

\* I know it's not really possible but still curious :wink:

src/Sylius/Bundle/UserBundle/spec/EventListener/CustomerContextListenerSpec.php
```diff
@@ -0,0 +1,56 @@
...
+        $customerContext->setCustomer($customer)->shouldBeCalled();
+
+        $this->setCustomerContextFromSubject($event);
+    }
+
+    function it_throws_exception_if_event_subject_is_not_a_customer_aware(
+        GenericEvent $event,
+        $product
+    ) {
+        $event->getSubject()->willReturn($product);
```
stloyd: Instead of such variable you could use `new \stdClass()`.

src/Sylius/Component/User/Context/CustomerContextInterface.php
```diff
@@ -22,4 +22,9 @@
      * @return CustomerInterface|null
      */
     public function getCustomer();
+
+    /**
+     * @param CustomerInterface $subject
+     */
+    public function setCustomer(CustomerInterface $subject);
```
stloyd: TBH. for me this interface is useless as it really doesn't mean anything...

src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
aramalipoor: If they login using their user account, customer will be read from User model not session. If their at a public place they should not do sensitive things at all or use incognito :D

src/Sylius/Component/User/Context/CustomerContextInterface.php
aramalipoor: I needed this when writing [specs here](https://github.com/Sylius/Sylius/pull/3230/files#diff-09bf6881a66fdabc66e869b85477d269R22), it feels weird to me too. Is it ok to use `CustomerContext` there?

aramalipoor:
@pjedrzejewski Can you please review and rerun the tests?

aramalipoor:
@pjedrzejewski Ping :)

src/Sylius/Bundle/UserBundle/spec/EventListener/CustomerContextListenerSpec.php
```diff
@@ -0,0 +1,53 @@
...
+        CustomerInterface $customer
+    ) {
+        $event->getSubject()->willReturn($subject);
+        $subject->getCustomer()->willReturn($customer);
+        $customerContext->setCustomer($customer)->shouldBeCalled();
+
+        $this->setCustomerContextFromSubject($event);
+    }
+
+    function it_throws_exception_if_event_subject_is_not_a_customer_aware(GenericEvent $event) {
```
stloyd: `{` should be at next line.

src/Sylius/Bundle/WebBundle/Behat/WebContext.php
```diff
@@ -459,7 +459,11 @@ public function iAmOnTheOrderPage($action, $number)
     {
         $order = $this->findOneBy('order', array('number' => $number));
 
-        $this->getSession()->visit($this->generatePageUrl('sylius_account_order_'.$action, array('number' => $order->getNumber())));
+        if ('tracking' === $action) {
+            $this->getSession()->visit($this->generatePageUrl('sylius_order_track', array('number' => $order->getNumber())));
```
stloyd: Instead of duplicating whole row, replace just route name?

src/Sylius/Bundle/WebBundle/Behat/WebContext.php
```diff
@@ -468,7 +472,11 @@ public function iAmOnTheOrderPage($action, $number)
      */
     public function iShouldBeOnTheOrderPage($action, $number)
     {
-        $this->assertSession()->addressEquals($this->generatePageUrl('sylius_account_order_'.$action, array('number' => $number)));
+        if ('tracking' === $action) {
+            $this->assertSession()->addressEquals($this->generatePageUrl('sylius_order_track', array('number' => $number)));
```
stloyd: Same here.

src/Sylius/Bundle/WebBundle/Resources/translations/messages.en.yml
```diff
@@ -2,6 +2,7 @@
 # (c) Paweł Jędrzejewski
 
 sylius:
+    back_to_shop: Go back to shop
```
stloyd: `Go back to the shop`?

src/Sylius/Bundle/WebBundle/Resources/views/Frontend/Account/layout.html.twig
```diff
@@ -12,5 +12,12 @@
 {% endblock %}
 
 {% block sidebar %}
-{{ knp_menu_render('sylius.frontend.account', {'template': 'SyliusWebBundle:Frontend:menu.html.twig'}) }}
+    {% if app.user %}
```
stloyd: IMO more safe would be call to the `is_granted()`.

src/Sylius/Bundle/WebBundle/Resources/translations/messages.en.yml
aramalipoor: IMO `Back to shop`  is better since it's shorter! WDYT?

src/Sylius/Bundle/WebBundle/Resources/views/Frontend/Account/layout.html.twig
aramalipoor: I only want to check user is logged in, so not any special authorization needed, will `is_granted('IS_AUTHENTICATED_REMEMBERED')` cover all logged in cases!?

michalmarcinkowski:
Looks good to me :+1:

winzou:
The common rule in the industry is to have a thank you page. You may have different needs of course, but I am not sure this `account/order` redirection will be useful for many stores. I would prefer to have something more generic in Sylius core.

kayue:
Agree with @winzou .

okwinza:
:+1: @winzou

aramalipoor:
@winzou My initial thought was `account/order` redirection, but that was very specific to our need. So I've changed the concept to an order tracking page `order/track`.

In order tracking page we can offer a way (using order tracking-code/hashed-id) for customers (which do not have a user account) to check their order's status and details *after* they've left the site, without the need to login.

Anyway if Sylius doesn't want to have **order tracking** feature in the core I'll happily change everything back to a Thank You page :)

aramalipoor:
@pjedrzejewski What do you think? We need this feature for our current sprint and your decision on this is important for us :)

src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
```diff
@@ -24,27 +25,58 @@ class CustomerContext implements CustomerContextInterface
...
+    protected $session;
 
     /**
      * @param SecurityContextInterface $securityContext
+     * @param SessionInterface $session
      */
-    public function __construct(SecurityContextInterface $securityContext)
-    {
+    public function __construct(
+        SecurityContextInterface $securityContext,
```
stloyd: Can you replace this deprecated functionality with proper ones (token storage & authentication manager)?

src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
```diff
@@ -24,27 +25,58 @@ class CustomerContext implements CustomerContextInterface
...
+    {
+        $this->session->set('customer', $customer);
+    }
+
+    /**
+     * @return UserInterface|null
+     */
+    protected function getUser()
+    {
+        if ($this->securityContext->getToken() && $this->securityContext->isGranted('IS_AUTHENTICATED_REMEMBERED')) {
```
stloyd: After replace of deprecated code as mentioned above this should look like:
```php
if ($this->authenticationManager->isGranted('IS_AUTHENTICATED_REMEMBERED')) {
    if ($token = $this->tokenStorage->getToken()) {
        return $token->getUser();
    }
}

return null;

src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
aramalipoor: Of course, good idea :+1:

aramalipoor:
@pjedrzejewski I think it's a false fail, can you re-run the test please? :)

aramalipoor:
@pjedrzejewski What do you think? We need this feature for our current sprint and your decision on this is important for us :)

Will you merge this or should I change it to a Thank You page with order summary? (Please re-read my updated PR comment)

aramalipoor:
@pamil Do you have any idea why javascript build fails?

michalmarcinkowski:
@aramalipoor I restarted the javascript build, it's some travis error that appears from time to time.

Regarding merging this PR. We have busy days right now, so I can't answer you when we will have time to merge/discuss it, sorry.... :disappointed: using fork will be safe option for you now as you don't have to wait for us.

pjedrzejewski:
@aramalipoor I am under a pile of stuff waiting to be discussed etc. but I will do my best to prepare my idea for the thank you page and hopefully we can agree on one universal solution. Actually, it is not that far from what you propose here, but just more generic and with more clear naming.

Workflow I have imagined:

* After you confirm the order with "Confirm" button on the finalize step, you actually leave the checkout flow;
* You are redirected to some Payment:pay action with given payment identifier, which then pretty much uses same logic as purchase step from checkout right now;
* PurchaseStep from checkout would be removed;
* If payment is successful, you are redirected to a thank you page - simple controller - it renders the template with order available as variable. This way every store can customize this page easily and include any tracking. Later we can come up with generic way to add more tracking pixels etc. cause I know it is pain in the ass to do it manually. Also, this page can include the order id in the url. We would need to verify if it belongs to currently logged in user and that's it.
* If payment is refused, you are redirected to Retry page.

Why is it good?

Redirecting the user to a "Pay" action (which redirects you to paypal/displays form/whatever, just like current purchase step) gives us a lot of flexibility. You can generate this link in E-mail payment reminders, put it in account panel, send it via support messages and do tons of things. Also it works great for multiple payments, much easier to handle. Currently, handling failed payment scenarios is super-annoying, cause you need to stay in the checkout flow, which brings a lot of problems.

Now, why it is not that simple:

To effectively implement thank you page, we need this "retry" logic and implement standalone purchase action. (almost copy paste from purchase step) Retry logic includes creating new payment if existing payments are failed or there is some amount due, not included in existing payments. Or update payment amount with status "new".

I think these are all my thoughts and that would make the whole workflow and flexibility much better in Sylius. If any brave soul is willing to implement this I will be happy to do my best to provide quite fast feedback as it is important feature and high on my list. Still not high enough to tackle it myself in the coming weeks. :/

michalmarcinkowski:
Implemented.

PR #3110, opened 2015-08-05
[PHP 5.5] [RFC] Future version bump to PHP 5.5 and Symfony 2.7

pamil:
Sylius requirements will bump in near future, so it's initial PR to do it. Will be helpful for Sylius developers, who code with the latest dependencies in mind.

PR #3236, opened 2015-09-08
[Behat] Changed the way of how the "I am on the page" assertion works

pamil:
Using PHP 5.5 generators (based on #3110), "I am on the (.+)? (page|step)" assertion compares current route with expected one, instead of comparing current address with expected address (generated from expected route). What it means - routes with parameters will work like a charm :)

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
+++ b/src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
@@ -525,9 +525,9 @@ abstract class DefaultContext extends RawMinkContext implements Context, KernelA
     {
         $router = $this->getService('router');
 
-        $url = $this->getSession()->getCurrentUrl();
-        $path = str_replace(rtrim($this->getMinkParameter('base_url'), '/'), '', $url);
-        $match = $router->match($path);
+        $request = $this->getSession()->getDriver()->getClient()->getRequest();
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -19,16 +19,16 @@
         }
     ],
     "require": {
-        "php":                                  ">=5.3.9",
+        "php":                                  "~5.5",
         "a2lix/translation-form-bundle":        "~2.0",
         "doctrine/collections":                 "~1.2",
-        "doctrine/common":                      "~2.3",
+        "doctrine/common":                      "~2.3,<2.5",
```
stloyd: Why?

pjedrzejewski: There is a bug in 2.5.1 IIRC. @pamil Right?

pamil: Version `2.5.0` has security issue and version `2.5.1` fixes that issue but introduces some bugs that affects our `@products` suite.

I think we may set `2.5` as default in `0.16`, but it'll need more work.

pamil: By the way, it is code from another PR, #3110 :)

pamil:
@pjedrzejewski ping, want to clean dependencies of #3134, as it depends on 3 pull requests now :)

pamil:
Closing as it's impossible to get page route while using Selenium2 driver and now we got javascript suite running.

Issue #3253, opened 2015-09-10
Frontend is not respecting the channel_based pricing 


paceto256:
1. I install 3 Instances of Sylius  (admin.com, chan1.com, chan2.com)
2. I created a channels for  chan1.com and chan2.com (the channels are defined fine because I am able to hide products just for one channel)
3. In admin.com I define a product "A" that is associated with  chan1.com and chan2.com
When defining the product "A" I am creating channel_based prices and I type 2 different prices for each channel.

Current behaviour:
The price in both channels chan1.com and chan2.com fallback to the default price.

Expected behaviour:
Each channel to respect the channel_based price associated to it.

pjedrzejewski:
Note to self/anyone who fixes it: Current channel is not passed in template to the pricing twig function.

michalmarcinkowski:
Should be fixed now.

PR #3255, opened 2015-09-10
Twig pricing extension should be aware of current channel

stloyd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| Fixed tickets | #3253


pjedrzejewski:
This is a hot fix, not proper solution. Channel is not only case where we miss the pricing context. We should figure out something more generic...

stloyd:
@pjedrzejewski What we do with this? IMO we should merge this as this fixes known issue, and later create more generic solution that will probably inject contexts automatically in proper places.

pjedrzejewski:
Fixed in #3409. See comments as it is not fixed properly yet. Ref. #3458.

Issue #2582, opened 2015-03-14
[RFC] Drop support for other templating engines than Twig?


stloyd:
Same way as Symfony 3 is making Twig the only possible default candidate for templating engine, I guess we should consider same for Sylius.

aRn0D:
yep, I am for!

patkar:
And how? Since Sylius uses only Twig, the templating engine of the enduser (the devs who are using Sylius), should decide on their own..

kayue:
If you are not using Twig, you probably are not using Sylius Standard Edition.

I am for aligning with Symfony 3.

patkar:
On my own, I'm using Twig ;) It was a general statement.

I'm interested how you want to drop the support :)

pjedrzejewski:
See my comment in #3258.

PR #3258, opened 2015-09-10
[RFC] Drop templating helpers in favor of direct using Twig extensions

stloyd:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC break | yes
| Fixed tickets | https://github.com/Sylius/Sylius/issues/2582


pjedrzejewski:
@stloyd RFC without a single word is not an RFC. Any reasoning? It used to be this way and we have changed to Helper + TwigExtension architecture.

stloyd:
@pjedrzejewski There was an RFC (without code changes suggestion): https://github.com/Sylius/Sylius/issues/2582, so this is more like proposal with code changes (forgot to mention that issue in description).

pjedrzejewski:
Twig is the default choice for Sylius as well, but I don't see why we should drop templating support when it is achieved with minimal effort and actually makes our code more decoupled. Let's not bother with that right now.

PR #3262, opened 2015-09-10
add BG translation ?

paceto256:


paceto256:
Translation are only imported via crowdin?

pjedrzejewski:
@paceto256 Yes, translations are only imported from Crowdin. I hope to get my head around this as quickly as humanly possible... We need to make the auto-sync bot work. :/

Issue #2749, opened 2015-05-19
[Translations] Traits?


patrick-mcdougle:
Pretty new to the project, but is there a reason that we're not using traits instead of class inheritance? By using traits, we should be able to decouple the bundles a lot easier. Instead of inheriting doctrine fields through extending parent classes, we could just add a translatable trait that adds the required fields.

Thoughts?

jimbocoder:
I haven't seen this specifically stated, but I believe Sylius intends to support PHP 5.3 for the foreseeable future, thus no traits.  The composer file specifies `php >= 5.3.9`.

michalmarcinkowski:
True, the traits weren't used because of the PHP 5.3 support, BUT actually in the short term we are going to move to PHP 5.4 and stop supporting PHP 5.3. 

@patrick-mcdougle if you have some free time, we will be more than happy to see the PR with Translations changed to traits :)

@gonzalovilaseca @pjedrzejewski do you see any drawbacks/barriers?

gonzalovilaseca:
No, in fact I would have done it with traits if it had been supported. Way better IMO.

aRn0D:
@pjedrzejewski when do you plan to move to php5.4? 0.17 ?

bicpi:
Is it worth to consider the stable library [DoctrineBehaviors](https://github.com/KnpLabs/DoctrineBehaviors) for this (and more like Timestampable, Softdeletable)?

steffenbrem:
This would make Sylius much more flexible. I currently have the issue that I cant extend the TreeRepository without copying Sylius code. So having this, would have fixed this easily.

michalmarcinkowski:
The traits are now supported! @patrick-mcdougle are you willing to work on this?

patrick-mcdougle:
I'm willing but I don't think I'll be able to get to it in a reasonable amount of time. I've got a big project at work that's eating all my time and when it's over I've got a backlog of a couple personal projects that I need to work on.

michalmarcinkowski:
Roger that. Thanks for quick response, so we will probably take care of that earlier :wink:

gonzalovilaseca:
@michalmarcinkowski Let me find some time during the following week to do it.

michalmarcinkowski:
Done

Issue #3248, opened 2015-09-10
[RFC] Use Traits to handle common entities


gperdomor:
With the #3110 merged,Sylius now require PHP 5.5, so can we use Traits, so i propose use some traits to handle common properties and methods which are repeated on many entities, like: <code>Gedmo\Timestampable\Traits\TimestampableEntity</code>, <code>Gedmo\SoftDeleteable\Traits\SoftDeleteableEntity</code>, and also create a new trait for toggleable entities

stloyd:
Yes, this is one of reason Sylius is migrating to newer PHP version, to be able to use advantages of the language :wink:

pjedrzejewski:
Yes, but before we introduce traits, let's discuss them here - what and where we want to use to avoid misusing them.

My quick list:
* Timestampable
* Softdeleteable
* Translatable
* Sortable

michalmarcinkowski:
I would add also `Toggleable` :wink:

gperdomor:
what about entities with images? I would add also <code>ImageTrait</code>

patkar:
Traits for all the XyzAwareInterface stuff?

stloyd:
@patkar Mostly, yes. That would allow to greatly reduce amount of duplicated code between components.

pjedrzejewski:
@patkar @stloyd This is wrong usage of traits. It only hides complexity. XyzAware is not a trait.

stloyd:
@pjedrzejewski Ah right, I misread that @patkar mentioned "aware" :wink:

patkar:
I don't get it, why not. For example the `OrderAwareInterface`. Most models implementing this interface have to repeat the same implementation. A `OrderAwareTrait` would supply a default code for all models sharing the same standard implementation.

A more complex model with other implementation of `OrderAwareInterface` must indeed code it's own implementation.

pjedrzejewski:
The thing is, with this approach, you can have a monstrum class, which uses 53445 traits and it only hides its complexity, which is wrong. We could do that for every common property, like Name, Id, Description etc. But this is wrong usage of trait, it hurts the quality of code.

patkar:
It's a valid and good argument.

I've always used traits to extract common getters and setters for relations which are used in multiple entities. For example address. But yes, it hides the implementation.

aRn0D:
Agreed with @patkar, we just need to limit the number of trait we want to use by classes and we need to be vigilant when we reviews PRs!

[...]

PR #3264, opened 2015-09-10
Introduce 3 new traits to reduce code duplication

stloyd:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC break | yes
| Fixed tickets | #2749 (partially)

### DIFF
--- a/src/Sylius/Component/Addressing/Model/Address.php
+++ b/src/Sylius/Component/Addressing/Model/Address.php
@@ -11,11 +11,15 @@
 
 namespace Sylius\Component\Addressing\Model;
 
+use Sylius\Component\Resource\Model\TimestampableTrait;
+
 /**
  * @author Paweł Jędrzejewski <pjedrzejewski@sylius.pl>
  */
 class Address implements AddressInterface
 {
+    use TimestampableTrait;
+
     /**
      * @var mixed
      */
@@ -65,16 +69,6 @@ class Address implements AddressInterface
      * @var string
      */
     protected $postcode;
-
-    /**
-     * @var \DateTime
-     */
-    protected $createdAt;
-
-    /**
-     * @var \DateTime
-     */
-    protected $updatedAt;
 
     public function __construct()
     {
@@ -247,37 +241,5 @@ class Address implements AddressInterface
     public function setPostcode($postcode)
     {
         $this->postcode = $postcode;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function getCreatedAt()
-    {
-        return $this->createdAt;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function setCreatedAt(\DateTime $createdAt)
-    {
-        $this->createdAt = $createdAt;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function getUpdatedAt()
-    {
-        return $this->updatedAt;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function setUpdatedAt(\DateTime $updatedAt)
-    {
-        $this->updatedAt = $updatedAt;
     }
 }
--- a/src/Sylius/Component/Order/Model/Adjustment.php
+++ b/src/Sylius/Component/Order/Model/Adjustment.php
@@ -11,11 +11,15 @@
 
 namespace Sylius\Component\Order\Model;
 
+use Sylius\Component\Resource\Model\TimestampableTrait;
+
 /**
  * @author Paweł Jędrzejewski <pawel@sylius.org>
  */
 class Adjustment implements AdjustmentInterface
 {
+    use TimestampableTrait;
+
     /**
      * @var mixed
      */
@@ -73,16 +77,6 @@ class Adjustment implements AdjustmentInterface
      * @var string
      */
     protected $originType;
-
-    /**
-     * @var \DateTime
-     */
-    protected $createdAt;
-
-    /**
-     * @var \DateTime
-     */
-    protected $updatedAt;
 
     public function __construct()
     {
@@ -288,38 +282,6 @@ class Adjustment implements AdjustmentInterface
     public function setOriginType($originType)
     {
         $this->originType = $originType;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function getCreatedAt()
-    {
-        return $this->createdAt;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function setCreatedAt(\DateTime $createdAt)
-    {
-        $this->createdAt = $createdAt;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function getUpdatedAt()
-    {
-        return $this->updatedAt;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function setUpdatedAt(\DateTime $updatedAt)
-    {
-        $this->updatedAt = $updatedAt;
     }
 
     private function recalculateAdjustable()
--- a/src/Sylius/Component/Product/Model/Product.php
+++ b/src/Sylius/Component/Product/Model/Product.php
@@ -15,6 +15,9 @@ use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Sylius\Component\Archetype\Model\ArchetypeInterface as BaseArchetypeInterface;
 use Sylius\Component\Attribute\Model\AttributeValueInterface as BaseAttributeValueInterface;
+use Sylius\Component\Resource\Model\SoftDeletableTrait;
+use Sylius\Component\Resource\Model\TimestampableTrait;
+use Sylius\Component\Resource\Model\ToggleableTrait;
 use Sylius\Component\Translation\Model\AbstractTranslatable;
 use Sylius\Component\Variation\Model\OptionInterface as BaseOptionInterface;
 use Sylius\Component\Variation\Model\VariantInterface as BaseVariantInterface;
@@ -25,13 +28,15 @@ use Sylius\Component\Variation\Model\VariantInterface as BaseVariantInterface;
  */
 class Product extends AbstractTranslatable implements ProductInterface
 {
+    use SoftDeletableTrait, TimestampableTrait, ToggleableTrait;
+
     /**
      * @var mixed
      */
     protected $id;
 
     /**
-     * @var null|ArchetypeInterface
+     * @var null|BaseArchetypeInterface
      */
     protected $archetype;
 
@@ -60,29 +65,10 @@ class Product extends AbstractTranslatable implements ProductInterface
      */
     protected $options;
 
-    /**
-     * @var \DateTime
-     */
-    protected $createdAt;
-
-    /**
-     * @var \DateTime
-     */
-    protected $updatedAt;
-
-    /**
-     * @var \DateTime
-     */
-    protected $deletedAt;
-
-    /**
-     * @var bool
-     */
-    protected $enabled = true;
-
     public function __construct()
     {
         parent::__construct();
+
         $this->availableOn = new \DateTime();
         $this->attributes = new ArrayCollection();
         $this->variants = new ArrayCollection();
@@ -107,7 +93,7 @@ class Product extends AbstractTranslatable implements ProductInterface
     }
 
     /**
-     * @param null|BaseArchetypeInterface $archetype
+     * {@inheritdoc}
      */
     public function setArchetype(BaseArchetypeInterface $archetype = null)
     {
@@ -302,7 +288,7 @@ class Product extends AbstractTranslatable implements ProductInterface
     public function getAttributeByCode($attributeCode)
     {
         foreach ($this->attributes as $attribute) {
-            if ($attribute->getAttribute()->getCode() === $attributeCode) {
+            if ($attributeCode === $attribute->getAttribute()->getCode()) {
                 return $attribute;
             }
         }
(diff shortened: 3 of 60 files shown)


src/Sylius/Component/Addressing/Model/Address.php
```diff
@@ -129,8 +119,6 @@ public function getFirstName()
     public function setFirstName($firstName)
     {
         $this->firstName = $firstName;
-
-        return $this;
```
patrick-mcdougle: So no more method chaining? Why?

src/Sylius/Component/Resource/Model/ToggleableTrait.php
```diff
@@ -0,0 +1,49 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Resource\Model;
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+trait ToggleableTrait
```
patrick-mcdougle: Enableable Trait?

src/Sylius/Component/Addressing/Model/Address.php
stloyd: http://ocramius.github.io/blog/fluent-interfaces-are-evil/

src/Sylius/Component/Resource/Model/ToggleableTrait.php
stloyd: IMO `toggable` is good enough.

src/Sylius/Component/Product/Model/Product.php
```diff
@@ -491,68 +443,14 @@ public function hasOption(BaseOptionInterface $option)
...
-        return $this->deletedAt;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function setDeletedAt(\DateTime $deletedAt = null)
     {
-        $this->deletedAt = $deletedAt;
+        $this->setDeletedAt($deletedAt);
```
michalmarcinkowski: I suppose it will cause never ending loop.

michalmarcinkowski:
All model interfaces should extend corresponding trait interface (`TimestampableInterface`, `SoftdeleteableInterface` etc.)

src/Sylius/Component/Addressing/Model/Address.php
```diff
@@ -11,6 +11,8 @@
 
 namespace Sylius\Component\Addressing\Model;
 
+use Sylius\Component\Resource\Model\TimestampableTrait;
+
 /**
  * Default address model.
```
michalmarcinkowski: Should be removed.

src/Sylius/Component/Addressing/Model/Address.php
```diff
@@ -18,6 +20,8 @@
  */
 class Address implements AddressInterface
 {
+    use TimestampableTrait;
+
     /**
      * Address id.
```
michalmarcinkowski: Should be removed.

src/Sylius/Component/Addressing/Model/Address.php
```diff
@@ -88,20 +92,6 @@ class Address implements AddressInterface
...
 
-    /**
-     * Creation time.
-     *
-     * @var \DateTime
-     */
-    protected $createdAt;
-
-    /**
-     * Last update time.
```
michalmarcinkowski: Quite obvious

src/Sylius/Component/Resource/Model/SoftDeletableTrait.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+trait SoftDeletableTrait
+{
+    /**
+     * @var \DateTime|null
+     */
+    protected $deletedAt;
+
+    /**
+     * Is item deleted?
```
michalmarcinkowski: Quite obvious comment

src/Sylius/Component/Resource/Model/SoftDeletableTrait.php
```diff
@@ -0,0 +1,53 @@
...
+     *
+     * @return bool
+     */
+    public function isDeleted()
+    {
+        return null !== $this->deletedAt && new \DateTime() >= $this->deletedAt;
+    }
+
+    /**
+     * Get the time of deletion.
```
michalmarcinkowski: Same here

src/Sylius/Component/Resource/Model/SoftDeletableTrait.php
```diff
@@ -0,0 +1,53 @@
...
+     *
+     * @return \DateTime
+     */
+    public function getDeletedAt()
+    {
+        return $this->deletedAt;
+    }
+
+    /**
+     * Set deletion time.
```
michalmarcinkowski: And here

src/Sylius/Component/Shipping/Model/Rule.php
```diff
@@ -37,18 +37,13 @@ class Rule implements RuleInterface
      *
      * @var array
      */
-    protected $configuration;
+    protected $configuration = array();
```
michalmarcinkowski: :+1:

src/Sylius/Component/Shipping/Model/Shipment.php
```diff
@@ -22,6 +23,8 @@
  */
 class Shipment implements ShipmentInterface, TimestampableInterface
```
michalmarcinkowski: `ShipmentInterface` should extend `TimestampableInterface`

src/Sylius/Component/Resource/Model/SoftDeletableTrait.php
stloyd: It comes from old methods.

src/Sylius/Component/Shipping/Model/Shipment.php
stloyd: This would be bigger BC break for no real reason, any advantage for doing this?

michalmarcinkowski: The major thing is that we display `createdAt` field in the admin backend. When someone will implement it's own `Shipment` class and doesn't define `createdAt`, we will have an exception during page render.

stloyd: But backend uses Core, and Core extends that model so your argument here is invalid :)

I guess such change should go to __different__ PR to not create monsterous one (once again :wink: ), @pjedrzejewski thoughts?

pamil: I don't think it will be a major BC break because `Shipment` would still implement `TimestampableInterface`, but it'll be more consistent when typehinting objects as `ShipmentInterface` as it would include the created at and updated at dates. Now you can't access those dates on `ShipmentInterface`.

One and only BC would be if someone defined these methods on interface extending `ShipmentInterface`, but that's the time to break the compatability.

stloyd:
@pjedrzejewski @umpirsky @aRn0D Feedback?

src/Sylius/Component/Addressing/Model/Address.php
patrick-mcdougle: But these aren't interfaces, they're classes.

stloyd: Then you didn't understand that post fully.

gperdomor:
Why create a new TimestampableTrait and SoftDeleteableTrait when can we use <code>Gedmo\Timestampable\Traits\TimestampableEntity</code> and <code> Gedmo\SoftDeleteable\Traits\SoftDeleteableEntity</code>?

The idea is reduce code duplication, but this new Traits are duplicate of the original class, and IMO, it's not necessary doing that

patrick-mcdougle:
^ also, with these, we get the doctrine behaviors for cheap/free.

stloyd:
This was first step for code reduction, next step would be testing those traits, or we could consider replacing Gedmo with KnpDoctrineBehaviors as mentioned in the issue in description.

gperdomor:
@stloyd What advantages have <code>KnpDoctrineBehaviors<code> over  <code>StofDoctrineExtensions</code>?

Regardless of which bundle use, both of them provide traits for <code>Timestampable</code> and <code>SoftDeleteable</code> entities, so i don't understand why we need the code duplication on this new traits.

Please can you explain why not use the Gedmo traits?, or even, in a extreme case, make this traits extends the Gedmo traits if want use a Sylius namespace on the Sylius components?

pjedrzejewski:
Gedmo is often handling simple things in really complex or even strange ways. Having our own traits is little duplication, but keeps us more decoupled, because in case we drop Gedmo someday, we won't need to change our code. :)

gperdomor:
@pjedrzejewski drop Gedmo support, or StofDoctrineExtensions implies the development of own custom doctrine behaviors, right?, so change code still needed. 

I only say that this traits are the same that Gedmo traits and are unnecesary if we can use the original traits.

Can we provide a wrapper to the original traits, we can make this traits extends the original traits, so if StofDoctrineExtensionBundle and Gedmo support is dropped in favor of KnpDoctrineBehavior, only need to change the traits to extend the traits provided by Knp and the mapping files.

PD: Sorry if this sounds offensive xD, is not my intention :+1:

pjedrzejewski:
@gperdomor Not offensive at all, I am really happy we have a discussion here. :) Your idea sounds reasonable, but I will have a look at all these traits and see if they are compatible. Still it is very little code in them, so I do not see big harm in having these few lines in our repo. :)

peteward:
I suggest you only try to achieve one thing with a PR, the addition of traits is really good (we're using very similar ones in our project), but removal of fluent interfaces should be a separate topic for debate, and a separate PR. It's not something that should be done as a part-refactor.

stloyd:
@pjedrzejewski @michalmarcinkowski @umpirsky @aRn0D ping

umpirsky:
I find this idea great. It is an improvement if we put BC break aside, which are still permitted.

If we can provide a wrapper to the original traits as @gperdomor suggested, it would be even better I guess. But having it as it is is fine imo.

stloyd:
@pjedrzejewski What should be done with this PR?

pjedrzejewski:
If you make the build green, I think we can merge it into 0.17.x unless someone has something against. I do not want to use Gedmo traits because that would couple models to Gedmo, we definitely do not want that. Thanks! :)

stloyd:
@pjedrzejewski Fixed unrelated change that was broking build.

src/Sylius/Component/Order/Model/Adjustment.php
```diff
@@ -104,8 +89,6 @@ public function getAdjustable()
         if (null !== $this->orderItem) {
             return $this->orderItem;
         }
-
-        return null;
```
michalmarcinkowski: This should not be removed. I know it's default behavior, but let's be explicit here.

src/Sylius/Component/Product/Model/Product.php
```diff
@@ -306,8 +292,6 @@ public function getAttributeByCode($attributeCode)
                 return $attribute;
             }
         }
-
-        return null;
```
michalmarcinkowski: This should be reverted. Same below.

src/Sylius/Component/Rbac/Model/Permission.php
```diff
@@ -68,15 +72,8 @@ class Permission implements PermissionInterface
     protected $level;
 
     /**
-     * @var \DateTime
+     * Constructor
```
michalmarcinkowski: This is redundant, can be removed.

src/Sylius/Component/User/Model/UserInterface.php
```diff
@@ -84,14 +86,9 @@ public function getUsernameCanonical();
...
-     * @param boolean $enabled
-     */
-    public function setEnabled($enabled);
-
-    /**
-     * @param boolean $locked
+     * @param bool $boolean
      */
-    public function setLocked($locked);
+    public function setLocked($boolean);
```
michalmarcinkowski: Should be reverted.

src/Sylius/Component/Order/Model/Adjustment.php
stloyd: This is done same to follow approach from Symfony.

michalmarcinkowski: In Sylius we are always explicit, so the `return null` should always end a method which returns something. I think we had this discussion in the past.

pjedrzejewski: :+1:

src/Sylius/Component/Resource/Model/ToggleableTrait.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        return $this->enabled;
+    }
+
+    /**
+     * @param bool $enabled
+     */
+    public function setEnabled($enabled)
+    {
+        $this->enabled = (bool) $enabled;
```
rvanlaak: What about fluent interfacing? Have a policy for that? Otherwise returning the object would be great.

stloyd: IIRC it's forbidden same as in Symfony, except builders etc.

src/Sylius/Component/Originator/Model/OriginAwareTrait.php
```diff
@@ -0,0 +1,60 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Originator\Model;
+
+/**
+ * @author Joseph Bielawski <stloyd@gmail.com>
+ */
+trait OriginAwareTrait
```
michalmarcinkowski: We should remove this trait, I don't think we need it. It's not correct usage of trait, because going in this direction we would have `AdjustmentsAwareTrait`, `ParentAwareTrait`, `NameAwareTrait` and `EverythingAwareTrait`. What is more it's used in only one place.

michalmarcinkowski:
@stloyd I would say that after applying the remaining comments it's ready to merge :+1: 

/cc @pjedrzejewski

michalmarcinkowski:
@stloyd are you willing to finish this PR?

stloyd:
Yes, just couldn't find time last days, hoping to get it done today/tomorrow.

michalmarcinkowski:
Great! :+1: 

Referencing issue #3248

stloyd:

pjedrzejewski:
Thank you Joseph! :+1:

kzap:
why was object chaining removed, is that an implementation detail and not a library feature?
even for simple setters it would be very useful

fluent classes are not the root of all evil :)

lchrusciel:
Yes, they are evil. We had a couple conversations about that and decide to resign from them. Here you have some explanation [why they are evil](http://ocramius.github.io/blog/fluent-interfaces-are-evil/)

kzap:
@lchrusciel ok fair enough, the library will not be opinionated so anyone else can extend and make it all chain

PR #3269, opened 2015-09-13
[Settings] Missing parameters issues in twig templates and settings loading

aramalipoor:
Q  | A
------------- | -------------
Bug fix?  | yes
New feature?  | no
BC breaks? | no
Deprecations? | no
Fixed tickets | no
License | MIT
Doc PR | -

There are a lot of times and loading a settings schema fails because we had a previous deprecated value in the database which we didn't want anymore and since it's not easy to create migrations for settings schema's I thought maybe it's safe to ignore missing options **ONLY** when loading from database.

### DIFF
--- a/src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
+++ b/src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
@@ -101,7 +101,7 @@ class SettingsManager implements SettingsManagerInterface
     /**
      * {@inheritdoc}
      */
-    public function loadSettings($namespace)
+    public function loadSettings($namespace, $ignoreUnknown = true)
     {
         if (isset($this->resolvedSettings[$namespace])) {
             return $this->resolvedSettings[$namespace];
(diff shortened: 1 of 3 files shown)


pjedrzejewski:
@aramalipoor Why closing this? :)

aramalipoor:
@pjedrzejewski After some thoughts it seemed a little odd to me and also there were many more situations, in which OptionsResolver's restrictions (e.g. `validateOptionsExistence`) is going to be annoying.

This PR is only covering unknowns parameters situations where we can easily check if it is not known then remove it.
BUT the problem is that what about **new proposed parameters** (say in a new version of schema) that are marked as *required*, OptionsResolver does not provide us a way to get a list of required parameters or a way to resolve without check for required params.

All in all I thought that this PR is incomplete and let's look for a better solution.

aramalipoor:
What do you think this solution? Do not resolve at all when loading from database, to prevent any migration issues. I don't think resolve-ing is going to bring us any value here!

aramalipoor:
@pjedrzejewski This PR is going to solve half of the problem but it seems that we really needs this half! Is it possible for you to review and merge it? Thanks ;)

pjedrzejewski:
@aramalipoor Could you rebase it with master? I will merge it then. :+1:

pjedrzejewski:
Thank you Aram!

PR #3270, opened 2015-09-14
[User] Removed unnecessary method due to PHP version bump

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Removed plain PHP duplication of `hash_pbkdf2()`.

### DIFF
--- a/src/Sylius/Component/User/Security/UserPbkdf2PasswordEncoder.php
+++ b/src/Sylius/Component/User/Security/UserPbkdf2PasswordEncoder.php
@@ -93,11 +93,7 @@ class UserPbkdf2PasswordEncoder implements UserPasswordEncoderInterface
             throw new \LogicException(sprintf('The algorithm "%s" is not supported.', $this->algorithm));
         }
 
-        if (function_exists('hash_pbkdf2')) {
-            $digest = hash_pbkdf2($this->algorithm, $plainPassword, $salt, $this->iterations, $this->length, true);
-        } else {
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
:+1:

elliot:
This should be refactored to use [```password_hash```](https://secure.php.net/manual/en/function.password-hash.php) instead of ```hash_pbkdf2```. 
Secure password storage mechanisms need to be slow, which the underlying bcrypt implementation in password_hash handles. 

Even with using the key stretching mechanism in PBKDF - using a system based on fast hashes (i.e anything in the SHA family) is a brute force security risk due to the advances in GPU/ASIC/FPGA tech.

pjedrzejewski:
@elliot is right here! What is the implementation in Symfony 3?

pamil:
We can aslo use some secure string comparison function to prevent [timing attacks](https://en.wikipedia.org/wiki/Timing_attack).

pjedrzejewski:
@pamil Let's apply @elliot's idea in a separate PR. Thanks!

michalmarcinkowski:
This encoder is used as a default User component encoder. No need to implement something more there, because we use Symfony Security component encoders (https://github.com/symfony/symfony/tree/2.7/src/Symfony/Component/Security/Core/Encoder) anyway and if anyone is willing to use component and not use Symfony security can implement encoder on it's own.

PR #3271, opened 2015-09-14
[PHP 5.5] Refactored \DateTime to \DateTimeInterface in typehints, docblocks and specs

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT


stloyd:
:-1: as it gives no value and just makes any rebase/merge much more painful.

lchrusciel:
:+1: As we should work on abstracts rather then on objects implementation.

pamil:
It makes using `\DateTimeImmutable` possible in Sylius and improves overall code quality as we're not bound to specific implementation.

Zales0123:
:+1: Agree with @lchrusciel, refactoring and improvements in code quality are always good.
(We've just create some kind of questionnaire here :smile:)

pjedrzejewski:
:+1:, but after #2255. ;>

aramalipoor:
Note that `\DateTimeInterface` does not guarantee existence of some methods such as `setTimezone` that can maybe cause some problems!

stloyd:
@aramalipoor Yep, that's why we __shouldn't__ do this that way! If we want to use immutable dates, replace the code in proper places, doing grep replacing is always bad idea.

pamil:
@aramalipoor @stloyd you're right, that change makes no sense :)

Issue #3259, opened 2015-09-10
[UserBundle] Missing Customer serialization reference


lchrusciel:
https://github.com/Sylius/Sylius/tree/master/src/Sylius/Bundle/UserBundle/Resources/config/serializer

PR #3278, opened 2015-09-15
[User] Serialization update

lchrusciel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #3259
| License       | MIT

I have added Customer serialization definition and change slightly current User serialization. Email and emailCanonical fields removal from User, resulted in is small BC break.


pjedrzejewski:
Thank you Łukasz!

Issue #1810, opened 2014-08-20
[RFC] Remove `promotion.item_based` from parameter.


kayue:
At the moment we have `sylius.promotion.item_based` parameter turned off by default. 

In my opinion this option can be removed, since it should be configured in individual promotion checker. (See https://github.com/Sylius/Sylius/pull/1744)
If store admin doesn't want any item-based promotion, he should simply don't apply any item-based action. 

[OrderPromotionListener](https://github.com/Sylius/Sylius/blob/1e040afb4e31ee9eac836f697795d6f51f731d3c/src/Sylius/Bundle/CoreBundle/EventListener/OrderPromotionListener.php#L92-L96) is the only place using this parameter.

/cc @stloyd @umpirsky

peteward:
Indeed if you create a promotion  with a FixedPriceAction and have `sylius.promotion.item_based:true` it will apply the fixed price discount to each item AND again to the order as well, because the processor applies to both.

I think it's better not to have `OrderItem` implement `PromotionSubjectInterface` and instead have all Rules and Actions work from the `Order` level.  In the above case it would be more obvious both in coding terms and to customers to implement FixedPriceOrderAction separate from a FixedPriceOrderItemAction applying discounts (adjustments) to either the entire order or individual items respectively, where the latter iterates through it's items to apply the discount.

This does create a downside of blurring the lines between rules and actions sometimes, so maybe there's a better way...

kayue:
@peteward Right now we do class check `$subject instanceof OrderInterface` in our action / rule, which solve the problem.

peteward:
It would do if the Action checks identified one or the other to work with, but both `PercentageDiscountAction` and `FixedDiscountAction` do the following:

```
        if (!$subject instanceof OrderInterface && !$subject instanceof OrderItemInterface) {
            throw new UnexpectedTypeException(
                $subject,
                'Sylius\Component\Core\Model\OrderInterface or Sylius\Component\Core\Model\OrderItemInterface'
            );
        }

        $adjustment = $this->createAdjustment($promotion);
        ...
        ...
        $subject->addAdjustment($adjustment);
```

and when the `OrderPromotionListener` does `processOrderPromotion()`, it applies to both items and orders where possible:

```
        $this->promotionProcessor->process($order);

        if ($this->itemBased) {
            foreach ($order->getItems() as $item) {
                $this->promotionProcessor->process($item);
            }
        }
```

So this flag can only break discounts currently. 

[...]

PR #3282, opened 2015-09-15
[WIP][Core] Add promotional product should mark promotion as it origin

stloyd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC break | yes(?)

TODO:
- [ ] migration


pjedrzejewski:
We are removing these promotions anyway. See discussion in #1810.

stloyd:
@pjedrzejwski This is still correct, it should be just adjusted (as every other action/rule) to work with cart/order. The current approach is broken in this action so IMO this should be merged before changing the approach to be "container" based.

PR #3307, opened 2015-09-21
[Promotion] Apply promotion on the cheapest item

elodie-nazaret:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets |  |
| License | MIT |
| Doc PR |  |

Promotion should apply on the cheapest item of the cart.
- [x] Add the rule
- [x] Add Behat tests
- [x] Add PhpSpec tests

**This development was sponsorised by Store Factory**


tristanbes:
is there anything blocking this PR @michalmarcinkowski  ?

pjedrzejewski:
@tristanbes Few things to consider before we merge it:

1. We are looking at reworking promotions, so I wonder if it is good idea to expand this functionality before that rework. #3368 
2. I do not like having the empty form type. I'd rather improve our promo system to support rules without any configuration. This will be very convenient for custom rules, which often do not contain any options, but only complex logic inside.

Thoughts?

pjedrzejewski:
Also, with the promo rework I mentioned, this would be a filter. /cc @peteward @PWalkow

peteward:
Yes, it would be a Filter.

It might already be included in Piotr's bundled Filters, however this is taking a back-seat to the Adjustments work since that underlies it. 

I'm guessing 6 weeks before Promotions are ready for a merge to either provide the above functionality or make it much easier to provide it.

PR #3315, opened 2015-09-21
[Translation] Remove fluent interfaces

TheMadeleine:
| Q                    | A
| -------------       | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | no
| License       | MIT

### DIFF
--- a/src/Sylius/Component/Translation/Model/AbstractTranslatable.php
+++ b/src/Sylius/Component/Translation/Model/AbstractTranslatable.php
@@ -19,15 +19,11 @@ use Doctrine\Common\Collections\ArrayCollection;
 abstract class AbstractTranslatable implements TranslatableInterface
 {
     /**
-     * Translations.
-     *
      * @var TranslationInterface[]
      */
     protected $translations;
 
     /**
-     * Current locale.
-     *
      * @var string
      */
     protected $currentLocale;
@@ -40,15 +36,10 @@ abstract class AbstractTranslatable implements TranslatableInterface
     protected $currentTranslation;
 
     /**
-     * Fallback locale.
-     *
      * @var string
      */
     protected $fallbackLocale;
 
-    /**
-     * Constructor.
-     */
     public function __construct()
     {
         $this->translations = new ArrayCollection();
@@ -71,8 +62,6 @@ abstract class AbstractTranslatable implements TranslatableInterface
             $this->translations->set($translation->getLocale(), $translation);
             $translation->setTranslatable($this);
         }
-
-        return $this;
     }
 
     /**
@@ -83,8 +72,6 @@ abstract class AbstractTranslatable implements TranslatableInterface
         if ($this->translations->removeElement($translation)) {
             $translation->setTranslatable(null);
         }
-
-        return $this;
     }
 
     /**
@@ -103,8 +90,6 @@ abstract class AbstractTranslatable implements TranslatableInterface
     public function setCurrentLocale($currentLocale)
     {
         $this->currentLocale = $currentLocale;
-
-        return $this;
     }
 
     /**
@@ -121,8 +106,6 @@ abstract class AbstractTranslatable implements TranslatableInterface
     public function setFallbackLocale($fallbackLocale)
     {
         $this->fallbackLocale = $fallbackLocale;
-
-        return $this;
     }
 
     /**
--- a/src/Sylius/Component/Translation/Model/AbstractTranslation.php
+++ b/src/Sylius/Component/Translation/Model/AbstractTranslation.php
@@ -11,22 +11,17 @@
 
 namespace Sylius\Component\Translation\Model;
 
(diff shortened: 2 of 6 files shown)


src/Sylius/Component/Translation/Model/AbstractTranslation.php
```diff
@@ -11,22 +11,18 @@
...
-
 /**
  * @author Gonzalo Vilaseca <gvilaseca@reiss.co.uk>
  */
 class AbstractTranslation implements TranslationInterface
 {
     /**
-     * Locale.
-     *
+
```
michalmarcinkowski: This shouldn't be added :wink:

michalmarcinkowski:
https://github.com/TheMadeleine/Sylius/blob/translation-fluent-interfaces/src/Sylius/Component/Translation/Model/AbstractTranslatable.php#L82 - this method should be on the interface

src/Sylius/Component/Translation/Model/AbstractTranslation.php
```diff
@@ -44,10 +40,6 @@ public function getTranslatable()
      */
     public function setTranslatable(TranslatableInterface $translatable = null)
     {
-        if ($translatable === $this->translatable) {
```
stloyd: This should be reverted.

src/Sylius/Component/Translation/Model/AbstractTranslation.php
```diff
@@ -44,10 +40,6 @@ public function getTranslatable()
...
-            return $this;
```
stloyd: As above should be reverted, this line should be just `return;`.

TheMadeleine:
@michalmarcinkowski :+1: done.

src/Sylius/Component/Translation/Model/AbstractTranslatable.php
```diff
@@ -19,15 +19,11 @@
 abstract class AbstractTranslatable implements TranslatableInterface
 {
     /**
-     * Translations.
-     *
      * @var TranslationInterface[]
```
stloyd: `Collection|TranslationInterface[]`

src/Sylius/Component/Translation/Model/TranslatableInterface.php
```diff
@@ -28,45 +28,34 @@
...
      * @param string $locale
-     *
-     * @return self
      */
     public function setFallbackLocale($locale);
 
     /**
-     * Get all translations.
-     *
      * @return TranslationInterface[]
```
stloyd: `Collection|TranslationInterface[]`

src/Sylius/Component/Translation/Model/AbstractTranslatable.php
pamil: IMHO putting `Collection` here is a bad idea as we don't want to couple this interface with Doctrine Commons

src/Sylius/Component/Translation/Model/TranslatableInterface.php
pamil: Same as above.

src/Sylius/Component/Translation/Model/AbstractTranslatable.php
stloyd: But __it's__ already coupuled with them in [`__construct()`](https://github.com/Sylius/Sylius/pull/3315/files#diff-90b0a0fcc484de0d02533a37e0903387R45)...

pamil: Ok, construct in abstract class is not the thing I woule expect :) Anyway, that should apply to `TranslatableInterface`.

michalmarcinkowski: @pamil decoupling from Doctrine Common is a topic for a separate discussion and is not concern of this PR.

pamil: @michalmarcinkowski Sure, that's why I want to prevent coupling at the moment :)

michalmarcinkowski: Let's stick to current convention to avoid inconsistency and if we decide to decouple we will do it everywhere at once.

gonzalovilaseca:
FWIW I'm working on a PR that gets rid of the abstract translation classes in favour of traits.

michalmarcinkowski:
@gonzalovilaseca that is great! If I will merge this PR will it make your rebase harder?

gonzalovilaseca:
@michalmarcinkowski Indeed ... but I might just copy this code to the trait, so go ahead, merge if you want and I'll find my way :-)

michalmarcinkowski:
Thank you Magda! :+1:

Issue #3303, opened 2015-09-18
[RFC] Migrations


michalmarcinkowski:
Issue
-------
Right now we don't run migrations on CI, so we don't validate if migrations work or even exist. Let's be honest they are not handled properly right now.

Solution
-----------
I suggest removing all existing migrations and generate first migration that will build the v0.15 database schema and do `app/console do:mig:mig` instead of `app/console do:sc:cr` on travis. Thanks to that we can validate that migrations work and there is nothing missing.

Enhancement
------------------
We can create `sylius:migrate` command which will run all migrations from `/vendor/sylius/sylius/app/migrations/` to ease Sylius update on existing projects (e.g. using Sylius-Standard).

gperdomor:
:+1:

kayue:
:+1:

pamil:
:+1:, any ideas on testing reverting migrations?

pamil:
Made PR with that changes (#3326). Do you think that migrations should be per change or per version?

 - Creating them per change may result in similiar operations being executing more times than needed (ex. creating table X, changing X.Y columnt type, adding new field X.Z etc.) 
 - Creating them per version drops support for those, who uses current dev version (and they can be really big).

IMHO we should add new migrations per change during development phase and bind them into one while releasing new version.

michalmarcinkowski:
@pamil your last sentence is the way we should handle them :wink:

pamil:
Anyway, what with Sylius bundles that we want to keep standalone and use migrations as well? I guess we will have to store these doubled, although we can use the same "cleaning" policy as with the application ones.

vikey89:
:+1:

pamil:
Solved by #6417.

pamil:
Travis runs migrations and validates schema, so we can be sure that migrations are up-to-date.

PR #3326, opened 2015-09-22
[Travis] New migrations policy

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3303
| License       | MIT

Deleted all existing migrations and created one that builds Sylius v0.15.0 database structure from scratch.


michalmarcinkowski:
:+1: Please update the changelog and we can get this merged! :wink:

winzou:
Wow, you cannot just delete migrations and let us deal with it when upgrading our sylius!
Please keep these migrations and add a new one from scratch to the first already existing migration.

michalmarcinkowski:
@winzou all migrations will be available in v0.15, so you can use whatever you need. Then in order to jump to v0.16 you would need to run just one migration. We plan to make one migration file per release to ease upgrades (all PRs will introduce their migrations and just before release we will merged them all into one file).

Does it sound good for you or you have some better idea? (related discussion https://github.com/Sylius/Sylius/issues/3303)

pamil:
@michalmarcinkowski done! :)

winzou:
What is the point of merging migrations?

For this PR, I would checkout Sylius just before the first migration, then do a `do:mi:di`, then add the generated file here. It would keep history and ease migrations (what if I want to update from 0.14 to dev-master?). I just hope that every schema change has its migration...

pamil:
@winzou New migrations will allow to migrate 0.15 -> 0.16 in just one migration. Development branch `master` will contain unmerged migrations so migrating 0.15 -> master would be easy to. The only drawback I can see is that migrating master -> next stable version would not be easily available - @michalmarcinkowski?

PR #3330, opened 2015-09-22
Add check to customer context

gonzalovilaseca:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

In one part of our site we don't need sylius autentication, just basic http with a generic username and password (which by default uses a Symfony user), so the Customer context fails when calling `->getCustomer()`.
Adding this check allows us to use basic http autentication.

### DIFF
--- a/src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
+++ b/src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
@@ -13,6 +13,7 @@ namespace Sylius\Bundle\UserBundle\Context;
 
 use Sylius\Component\User\Context\CustomerContextInterface;
 use Sylius\Component\User\Model\CustomerInterface;
+use Sylius\Component\User\Model\UserInterface;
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/UserBundle/Context/CustomerContext.php
```diff
@@ -42,6 +43,7 @@ public function getCustomer()
     {
         if ($this->securityContext->getToken() && $this->securityContext->isGranted('IS_AUTHENTICATED_REMEMBERED')
             && $this->securityContext->getToken()->getUser()
```
pamil: It's redundant now.

michalmarcinkowski:
@gonzalovilaseca :+1: Could you squash your commits into one?

gonzalovilaseca:
done!

pjedrzejewski:
Thank you Gonzalo!

PR #3333, opened 2015-09-22
[SettingsBundle] Remove all fluent interfaces

tuka217:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | no
| License       | MIT
| Doc PR        | no

### DIFF
--- a/src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
+++ b/src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
@@ -24,36 +24,26 @@ use Symfony\Component\Validator\Exception\ValidatorException;
 use Symfony\Component\Validator\ValidatorInterface;
 
 /**
- * Settings manager.
- *
  * @author Paweł Jędrzejewski <pawel@sylius.org>
  */
 class SettingsManager implements SettingsManagerInterface
 {
     /**
-     * Schema registry.
-     *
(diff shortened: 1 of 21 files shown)


src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
```diff
@@ -134,7 +118,6 @@ public function loadSettings($namespace)
 
     /**
      * {@inheritdoc}
-     * @throws ValidatorException
```
stloyd: IIRC this is correct, cause this method throws exception that is not listed in interface.

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManagerInterface.php
```diff
@@ -30,10 +27,10 @@
     public function loadSettings($namespace);
 
     /**
-     * Save settings under given namespace.
-     *
      * @param string   $namespace
      * @param Settings $settings
+     *
+     * @throws ValidatorException
```
stloyd: IMO this is wrong, as this is an implementation which throws exception, not interface that decides how to deal with such errors.

michalmarcinkowski:
:+1:

PR #3344, opened 2015-09-23
[Behat] [Travis] Made Behat waiting for resources to appear

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Added controlled timeouts to "I am on specific page" and "resource should be created" steps.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
+++ b/src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
@@ -456,6 +456,32 @@ abstract class DefaultContext extends RawMinkContext implements Context, KernelA
     }
 
     /**
+     * Callback can return any value, to stop waiting use anonymous function provided as first argument.
+     *
+     * @param callable $callback
+     * @param int $limit
+     * @param int $delay In miliseconds
+     *
+     * @return mixed
+     *
+     * @throws \RuntimeException If timeout was reached
+     */
+    protected function waitFor(callable $callback, $limit = 10, $delay = 100)
+    {
+        for ($i = 0; $i < $limit; ++$i) {
+            $payload = $callback();
+
+            if (null !== $payload) {
+                return $payload;
+            }
+
+            usleep($delay * 1000);
+        }
+
+        throw new \RuntimeException(sprintf('Timeout reached (%f seconds)!', round($limit * $delay / 1000, 1)));
+    }
+
+    /**
      * @param string $name
      *
      * @return string
--- a/src/Sylius/Bundle/ResourceBundle/Behat/FixtureContext.php
+++ b/src/Sylius/Bundle/ResourceBundle/Behat/FixtureContext.php
@@ -77,7 +77,10 @@ class FixtureContext extends DefaultContext
         $data = $table->getRowsHash();
         $type = str_replace(' ', '_', trim($type));
 
-        $object = $this->findOneByName($type, $data['name']);
+        $object = $this->waitFor(function () use ($type, $data) {
+            return $this->findOneByName($type, $data['name']);
+        });
+
         foreach ($data as $property => $value) {
             $objectValue = $accessor->getValue($object, $property);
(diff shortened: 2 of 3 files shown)


src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
```diff
@@ -456,6 +456,37 @@ protected function replaceColumnNamesWithColumnIds(NodeElement $table, array $fi
...
+            $payload = $callback($emergencyExit);
+
+            if ($forceBreak) {
+                return $payload;
+            }
+
+            usleep($delay * 1000);
+        }
+
+        throw new \RuntimeException(sprintf('Timeout reached (%f seconds)!', round($limit * $delay / 1000, 1)));
```
michalmarcinkowski:
```php
    protected function waitFor(callable $callback, $limit = 10, $delay = 100)
    {
        for ($i = 0; $i < $limit; ++$i) {
            if ($callback()) {
                return true;
            }
            usleep($delay * 1000);
        }
        throw new \RuntimeException(sprintf('Timeout reached (%f seconds)!', round($limit * $delay / 1000, 1)));
    }
```

src/Sylius/Bundle/ResourceBundle/Behat/FixtureContext.php
```diff
@@ -77,7 +77,19 @@ public function objectWithFollowingDataShouldBeCreated($type, TableNode $table)
...
+            function ($emergencyExit) use ($type, $data) {
+                $object = $this->findOneByName($type, $data['name']);
+                if (null !== $object) {
+                    $emergencyExit();
+
+                    return $object;
+                }
+
+                return null;
+            }
```
michalmarcinkowski: Would be simpler:
```php
function () use ($type, $data) {
    $object = $this->findOneByName($type, $data['name']);

    return null !== $object;
}
```

pjedrzejewski:
@pamil Could you just quickly explain what is the point of this change? :)

src/Sylius/Bundle/ResourceBundle/Behat/DefaultContext.php
```diff
@@ -456,6 +456,37 @@ protected function replaceColumnNamesWithColumnIds(NodeElement $table, array $fi
...
+    {
+        $forceBreak = false;
+        $emergencyExit = function () use (&$forceBreak) {
+            $forceBreak = true;
+        };
+
+        for ($i = 0; $i < $limit; ++$i) {
+            $payload = $callback($emergencyExit);
+
+            if ($forceBreak) {
```
michalmarcinkowski: Maybe we could simply use `null !== $payload` and get rid of this `emergencyExit`. It will make code more readable and less 'magic' :wink: `

pamil: Yeah, I made it a little bit overcomplicated. Wanted to add ability to wait for `null` values, but it seems like there's no real use case for this :)

pamil:
@pjedrzejewski our test suites are likely to fail if there's no resource they're looking for. I can't find the real example on Travis (because we rerun these failed jobs), but it seems like we want to do something with resource we created earlier, but there is no *the* resource yet. 

Like we have just created shipping method "Foo" and the next step is to check whether we are on the page of shipping method "Foo" - it tries to find "Foo" in database but there's none.

I may be wrong, but it may fix some issues, if not - there are no real drawbacks.

pamil:
@pjedrzejewski found one - [https://travis-ci.org/Sylius/Sylius/jobs/82529613](here!) IMHO it's more likely to happen in Javascript suite, but I've also seen it in some non-JS suites. The JS-related issue may have something in common with the way Selenium2 and its driver works, it may not wait for the whole page to load.

michalmarcinkowski:
Much better now :smile: Thanks!

PR #3346, opened 2015-09-24
[UserBundle] Fix CustomerContext to not use deprecated code anymore

stloyd:


src/Sylius/Component/User/Context/CustomerContextInterface.php
```diff
@@ -19,7 +19,7 @@
 interface CustomerContextInterface
 {
     /**
-     * @return CustomerInterface|null
+     * @return null|CustomerInterface
```
pamil: That change makes no sense.

pamil:
:+1:

src/Sylius/Component/User/Context/CustomerContextInterface.php
stloyd: This whole interface makes no sense at all, same as discussing that phpdoc order is correct or no, even if code will first return null not the object. EOT.

Zales0123: So if there is no sense in this interface, why do you even bother about it? :) And more, why do you change senseless phpdoc order in senseless interface? 
BTW, closing a discussion in such a rude manner, you make reviewing your PRs really *senseless* ;)

pamil: Anyway, it returns `null` first (in code), because it's the way we handle *edge* cases, we use early returns. The main target of `getCustomer()` method is to get customer.

pjedrzejewski: There is no EOT in Sylius PR discussions. :) I do not think we should remove this interface.

pjedrzejewski:
I don't we should remove this interface. :( Rest of changes was OK.

pjedrzejewski:
Ping @stloyd.

PR #3355, opened 2015-09-24
[Translation] Introducing traits

gonzalovilaseca:
| Q             | A
| ------------- | ---
| Bug fix?      | [no]
| New feature?  | [no]
| BC breaks?    | [yes]
| Deprecations? | [no]
| Fixed tickets | []
| License       | MIT
| Doc PR        | []

This PR breaks BC, so I'm wondering if we should leave the abstract classes.


gonzalovilaseca:
I will have a look at failing tests and update PR

src/Sylius/Component/Archetype/Model/Archetype.php
```diff
@@ -74,8 +78,7 @@ public function __construct()
         $this->attributes = new ArrayCollection();
         $this->options = new ArrayCollection();
         $this->createdAt = new \DateTime();
-
-        parent::__construct();
+        $this->__ttConstruct();
```
axelvnk: Hmm, I'm not sure __ttConstruct is readable enough.. Why not name it translatableTraitConstruct? Would be much clearer imo.

michalmarcinkowski:
That is already done. Thanks for the effort!

PR #3337, opened 2015-09-23
[Travis] Running PhpSpec on PHP 7

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3312, #3287, #3257, #3260
| License       | MIT

Based on #3287 and #3257.

PR #3356, opened 2015-09-25
Add "akeneo/phpspec-skip-example-extension" dependency & fix broken specs

stloyd:
Before:
Before

After:
After

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -85,7 +85,8 @@
         "coduo/php-matcher":                    "~1.0",
         "phpspec/phpspec":                      "~2.1",
         "phpunit/phpunit":                      "~4.1",
-        "lakion/mink-debug-extension":          "^1.0.1"
+        "lakion/mink-debug-extension":          "^1.0.1",
+        "akeneo/phpspec-skip-example-extension": "~1.2"
     },
     "replace": {
         "sylius/addressing":         "self.version",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "9223f62cbb3516e3e019a3648c2cc4c1",
+    "hash": "f45642c2c8997291516f7c2ad78c9ac5",
     "packages": [
         {
             "name": "a2lix/translation-form-bundle",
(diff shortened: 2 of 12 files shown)


michalmarcinkowski:
Already fixed in https://github.com/Sylius/Sylius/pull/3337

stloyd:
@michalmarcinkowski Nope, this is not fully fixed, please compare PRs before closing...

michalmarcinkowski:
Do not make irrelevant changes with relevant ones. Like:

* What is wrong with this line https://github.com/Sylius/Sylius/compare/master...stloyd:bugfix/phpspec#diff-e02e8f73786f6373e93576a3f80290e0R23?
* CS fixes - https://github.com/Sylius/Sylius/compare/master...stloyd:bugfix/phpspec#diff-2f2865b3defb8a22fd68469ed579a576R43

stloyd:
@michalmarcinkowski First is totally correct as that phpspec extension __can't__ skip spec if PHP extension is __not__ loaded, and that spec will always fail if you don't have `intl` PHP extension loaded as default Symfony subs support __only__ english language!

Without that change you will get:
> Sylius/Bundle/MoneyBundle/Templating/Helper/MoneyHelper                              
  38  - it formats the integer amounts into string representation
      exception [exc:Symfony\Component\Intl\Exception\MethodArgumentValueNotImplementedException("The Symfony\Component\Intl\NumberFormatter\NumberFormatter::__construct() method's argument $locale value 'de_DE' behavior is not implemented. Only the locale "en" is supported.  Please install the "intl" extension for full localization capabilities.")] has been thrown.

> Sylius/Bundle/MoneyBundle/Templating/Helper/MoneyHelper                         
  46  - it allows to format money in different currencies
      exception [exc:Symfony\Component\Intl\Exception\MethodArgumentValueNotImplementedException("The Symfony\Component\Intl\NumberFormatter\NumberFormatter::__construct() method's argument $locale value 'de_DE' behavior is not implemented. Only the locale "en" is supported.  Please install the "intl" extension for full localization capabilities.")] has been thrown.

Second is just matter of approach, when I fix code, why should I not fix another issue that is in that file? With such approach you should "moan" in at least 50% of PRs being done :stuck_out_tongue_winking_eye:

michalmarcinkowski:
You're right with the `intl`, I didn't realize it can cause a problem.

The problem with second is that it makes code compare less readable and confuses what is a real fix and what is just a CS or esthetic. 

E.g. It is hard to see if you just reordered specs like here https://github.com/Sylius/Sylius/pull/3357/files#diff-bdc3b250d685513f2ff47620755d6405R8 or removed it totaly like here https://github.com/Sylius/Sylius/pull/3357/files#diff-bdc3b250d685513f2ff47620755d6405L24.

stloyd:
@michalmarcinkowski Yes, I know it makes a bit harder to read first time, but some time ago we done some work to sort it, and I see some PRs ignored that change and broke sorting, that was the reason for this change.

While removing was done for purpose as there is __no__ such components like `Settings`, `Payum` or `Installer` and this caused to create empty folders by phpspec which is wrong IMO.

michalmarcinkowski:
@stloyd I agree that both changes are correct, but it's worth to at least mention it in the PR description to make the review easier.

stloyd:
@michalmarcinkowski My bad, will try to give more details what's done if the purpose is not so obvious.

Can you check the new PR that fixes problems mentioned above?

PR #3376, opened 2015-09-30
[Installer] Added new administrator password confirmation

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        |

While creating new administrator account in Sylius installer, it is not required to confirm password, which should be obvious, in my opinion ;)

### DIFF
--- a/src/Sylius/Bundle/InstallerBundle/Command/SetupCommand.php
+++ b/src/Sylius/Bundle/InstallerBundle/Command/SetupCommand.php
@@ -108,7 +108,7 @@ EOT
             } while ($exists);
 
             $user->setEmail($email);
-            $user->setPlainPassword($this->askHidden($output, 'Choose password:', array(new NotBlank())));
+            $user->setPlainPassword($this->getAdministratorPassword($output));
         }
 
         $user->setEnabled(true);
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/InstallerBundle/Command/SetupCommand.php
```diff
@@ -108,7 +108,7 @@ protected function setupAdministratorUser(InputInterface $input, OutputInterface
             } while ($exists);
 
             $user->setEmail($email);
-            $user->setPlainPassword($this->askHidden($output, 'Choose password:', array(new NotBlank())));
+            $user->setPlainPassword($this->getAdministratorPassword($output));
```
stloyd: This should be `$dialog->askHiddenResponseAndValidate()`, where `$dialog` is default Symfony helper.

Zales0123: Sure, you're right. I have only one problem with using Symfony Dialog Helper - using it, we could only ask and validate password confirmation. What if **the first** password was typed wrongly? We could obviously nest one validator in another, but it smells really bad.

michalmarcinkowski:
Thank you Mateusz

PR #3383, opened 2015-10-02
[FixturesBundle] - added missing taxonomies associations for channels sample data

nakashu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

When executing doctrine:fixtures:load (or any superset like sylius:install), channels would be without any associated taxonomies.

This would result in missing taxonomies on frontend, and possibly other important places.

PR adds all taxonomies for all channels.


michalmarcinkowski:
Thank you @nakashu

PR #3384, opened 2015-10-02
[Sylius] - register DebugBundle by default in appKernel dev/test

nakashu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Enable debugBundle for dev/test environment by default
Many SF developers are used for the shorthand dump() function provided by debugBundle, and will register the bundle anyway.

This makes the DX better, as people will be able to use what they are used to.


pjedrzejewski:
Thank you @nakashu! :+1:

PR #3385, opened 2015-10-02
[InstallerBundle] - fixed recommended version of php lower then required

nakashu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Recommended version of was php lower then required,
required -> 5.5.9 
recommended -> 5.3.8

Also added constant for php recommended version RECOMMENDED_PHP_VERSION
set to 5.6.13, just because its current.

If anyone got suggestion what the recommended version should be, drop it here and I'll change it.


pjedrzejewski:
:+1:

michalmarcinkowski:
Thank you @nakashu

PR #3386, opened 2015-10-04
[User] Add user:create and user:delete commands

aramalipoor:
Q  | A
------------- | -------------
Bug fix?  | no
New feature?  | yes
BC breaks? | no
Deprecations? | no
Fixed tickets | no
License | MIT
Doc PR | -

### DIFF
--- a/src/Sylius/Bundle/UserBundle/Command/CreateUserCommand.php
+++ b/src/Sylius/Bundle/UserBundle/Command/CreateUserCommand.php
@@ -1,0 +1,208 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\UserBundle\Command;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Sylius\Bundle\ResourceBundle\Doctrine\ORM\EntityRepository;
+use Sylius\Component\Core\Model\UserInterface;
+use Sylius\Component\Core\Model\CustomerInterface;
+use Sylius\Component\Rbac\Model\RoleInterface;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Aram Alipoor <aram.alipoor@gmail.com>
+ */
+class CreateUserCommand extends ContainerAwareCommand
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName('sylius:user:create')
+            ->setDescription('Creates a new user account.')
+            ->setDefinition(array(
+                new InputArgument('email', InputArgument::REQUIRED, 'Email'),
+                new InputArgument('password', InputArgument::REQUIRED, 'Password'),
+                new InputArgument('roles', InputArgument::IS_ARRAY, 'RBAC roles'),
+                new InputOption('super-admin', null, InputOption::VALUE_NONE, 'Set the user as a super admin'),
+                new InputOption('disabled', null, InputOption::VALUE_NONE, 'Set the user as a disabled user'),
+            ))
+            ->setHelp(<<<EOT
+The <info>%command.name%</info> command creates a new user account.
+EOT
+            )
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/UserBundle/Command/CreateUserCommand.php
```diff
@@ -0,0 +1,252 @@
...
+     * @param string $firstName
+     * @param string $lastName
+     * @param bool $enabled
+     * @param string $currency
+     * @param array $roles
+     * @param array $securityRoles
+     *
+     * @return UserInterface
+     */
+    protected function createUser($email, $password, $firstName, $lastName, $enabled, $currency, array $roles, array $securityRoles = ['ROLE_USER'])
```
diimpp: array('ROLE_USER')

src/Sylius/Bundle/UserBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,90 @@
...
+
+        $manager->flush();
+    }
+
+    /**
+     * @return EntityManagerInterface
+     */
+    protected function getEntityManager()
+    {
+        return $this->getContainer()->get('doctrine.orm.entity_manager');
```
diimpp: You should be using sylius.manager.customer and sylius.manager.user

src/Sylius/Bundle/UserBundle/Command/CreateUserCommand.php
```diff
@@ -0,0 +1,252 @@
...
+    {
+        $this
+            ->setName('sylius:user:create')
+            ->setDescription('Creates a new user account.')
+            ->setDefinition(array(
+                new InputArgument('email', InputArgument::REQUIRED, 'Email'),
+                new InputArgument('password', InputArgument::REQUIRED, 'Password'),
+                new InputArgument('first-name', InputArgument::OPTIONAL, 'First name'),
+                new InputArgument('last-name', InputArgument::OPTIONAL, 'Last name'),
+                new InputArgument('currency', InputArgument::OPTIONAL, 'Currency'),
```
diimpp: UserBundle doesn't know about currencies.
CoreBundle know about them.
So it should be general command for UserBundle and extended version of CoreBundle, if it's required.

diimpp:
- I think it should be covered with phpspec
- Single command for user and customer doesn't sound good, but I'm unsure how it should be done.

src/Sylius/Bundle/UserBundle/Command/CreateUserCommand.php
aramalipoor: Good point! I'll work on it when I'm free!

michalmarcinkowski:
@aramalipoor when do you think this feature can be useful? We can create users in the backend, so why we need a console command for that?

pjedrzejewski:
@michalmarcinkowski Mostly for development purposes, much easier to run a command instead of going to the backend.

michalmarcinkowski:
:+1:

src/Sylius/Bundle/UserBundle/Command/CreateUserCommand.php
```diff
@@ -0,0 +1,198 @@
...
+    {
+        $canonicalizer = $this->getContainer()->get('sylius.user.canonicalizer');
+
+        /**
+         * @var $user UserInterface
+         * @var $customer CustomerInterface
+         */
+        $user = $this->getUserRepository()->createNew();
+        $user->setUsername($email);
+        $user->setEmail($email);
```
michalmarcinkowski: The `User` can't exist without `Customer` and the `email` field doesn't exist on user entity https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/User/Model/User.php#L414.

Take a look how is it handled in the Behat context https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Behat/CoreContext.php#L717 and repeat the logic here.

src/Sylius/Bundle/UserBundle/Command/DeleteUserCommand.php
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $email = $input->getArgument('email');
+        $manager = $this->getEntityManager();
+
+        /** @var UserInterface $user */
+        $user = $this->getUserRepository()->findBy(array('email' => $email));
```
michalmarcinkowski: This will not work... You should use `findOneByEmail` method https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/UserBundle/Doctrine/ORM/UserRepository.php#L163.

pjedrzejewski:
@aramalipoor Nice, but we are missing a database migration. Actually, this change would be best in a separate PR, just to keep things simple. :+1:

aramalipoor:
@pjedrzejewski Sorry it took so long to respond, I've reverted the change to the database in favor of #3484 ;)

pjedrzejewski:
Thanks Aram! No worries. :)

Issue #3218, opened 2015-09-03
[Installation] Error during assets removal.


lchrusciel:
The forth step of installation try to install assets. If you have already some, it will try to remove them. But not proper rights on web/bundles/ folder will result with error: 
```
 [Symfony\Component\Filesystem\Exception\IOException]                              
  Failed to remove file "web/bundles/syliusshipping/js/sylius-shipping-method.js".
```

pjedrzejewski:
We do check permission in the first step, don't we?

Zales0123:
Maybe, command doesn't check this permissions perfectly. Let's assume:
- firstly, checked directories were ``web/bundles/*`` and ``web/assets/*``
- secondly, they were changed by @pamil to ``web/bundles/`` and ``web/assets/`` (which is obviously correct)

But what about files/folders that are **inside** this directories, can they have improper permissions even though their parents have them set right? Function ``ensureDirectoryIsWritable($directory, OutputInterface $output)`` check and try to set proper permissions on said catalogs, but evidently command ``sylius:install:assets`` must have them set to already existing files in that folders as well, due to some deletion stuff. So maybe the best solution, will be setting permissions, somehow, recursively, on all already existing files and catalogs inside ``web/bundles/`` and ``web/assets/``?

Zales0123:
I think it can be closed, as this problem should not happen anymore.

PR #3387, opened 2015-10-05
[Installer] Fixed permissions checking during installation

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Sylius/Sylius/issues/3218
| License       | MIT
| Doc PR        |

There were some problem with permissions checking during running ``sylius:install:assets`` command. Now installer checks recursively permissions of ``web/bundles/`` and ``web/assets/`` directories and changes them if following directories are not writeable. If they cannot be changed, user gets message with specific directory that should have permissions changed manually.


src/Sylius/Bundle/InstallerBundle/Checker/CommandDirectoryCheckerInterface.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+namespace Sylius\Bundle\InstallerBundle\Checker;
+
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface CommandDirectoryCheckerInterface
```
stloyd: This interface is useless IMO.

src/Sylius/Bundle/InstallerBundle/Checker/CommandDirectoryChecker.php
```diff
@@ -0,0 +1,139 @@
...
+        $this->name = $name;
+    }
+
+    /**
+     * @param string          $directory
+     * @param OutputInterface $output
+     */
+    private function createDirectory($directory, OutputInterface $output)
+    {
+        if (!mkdir($directory, 0755, true)) {
```
stloyd: Why not using Filesystem component from Symfony?

michalmarcinkowski:
Thank you Mateusz

src/Sylius/Bundle/InstallerBundle/Checker/CommandDirectoryChecker.php
```diff
@@ -0,0 +1,158 @@
...
+use Symfony\Component\Debug\Exception\ContextErrorException;
+use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Iterator\RecursiveDirectoryIterator;
+use Symfony\Component\Security\Core\Exception\AccessDeniedException;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class CommandDirectoryChecker
```
pjedrzejewski: CommandDirectoryChecker is a weird name. I had no idea what it does initially. It not only checks directories, but creates them and changes permissions. I think we should come up with a better naming for that class.

src/Sylius/Bundle/InstallerBundle/Resources/config/services.xml
```diff
@@ -90,6 +91,10 @@
         <service id="sylius.installer.yaml_persister" class="%sylius.installer.yaml_persister.class%">
             <argument>%kernel.root_dir%/config/parameters.yml</argument>
         </service>
+
+        <service id="sylius.installer.checker.command_directory" class="%sylius.installer.checker.command_directory.class%">
```
pjedrzejewski: Not sure about the service name. Will we have more checkers? (see comment above about checker name)

src/Sylius/Bundle/InstallerBundle/Checker/CommandDirectoryChecker.php
Zales0123: You're bloody hell right! I just couldn't came out with more proper name, but I would definitely turn on my creativity in nearest future and push PR with some pretty good name for this service ;)

PR #3390, opened 2015-10-05
[Resource] Add filter manipulator trait

aramalipoor:
Q  | A
------------- | -------------
Bug fix?  | yes
New feature?  | no
BC breaks? | no
Deprecations? | no
Fixed tickets | no
License | MIT
Doc PR | -

Main reason to add this trait is that we wanted to completely turn off `softdeleteable` filter. We ran into many exceptions because this filter no longer existed.

**Still working on it, but comments are very welcome on the whole idea :)**

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
+++ b/src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
@@ -45,11 +45,10 @@ class OrderController extends ResourceController
         $paginator->setMaxPerPage($this->config->getPaginationMaxPerPage());
 
         // Fetch and cache deleted orders
-        $entityManager = $this->get('doctrine.orm.entity_manager');
-        $entityManager->getFilters()->disable('softdeleteable');
+        $this->disableFilter('softdeleteable');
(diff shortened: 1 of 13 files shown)


michalmarcinkowski:
For sure we need a debate about the softdeletable entities and filters, because we need a safe and generic way to handle all relations and logic related to it.

Closing this PR, because @pjedrzejewski has reworked the `ResourceController` and `EntityRepository` in the https://github.com/Sylius/Sylius/pull/2255 (which is close finishing), so your changes will not be valid in the new approach. Sorry for that, but we need to wait for this PR in order to apply enhancements to `ResourceBundle`.

aramalipoor:
@michalmarcinkowski It's good news you're going to merge #2255 ;)
 Anyway I really am looking for a debate over softdeletable as well, since it has brought us some bugs in various placed.

I've heard soft-deleteable is not a good approach because you're trying to do the job of the database layer in your application layer. And you have to use Audit Logs and things like that instead of soft-deleteable, but I don't have a personal experience on that matter.

PR #3397, opened 2015-10-06
Remove unnessary softdeleteable in CoreBundle User mappings

diimpp:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        |

While not exactly a bug, it have a great chance to become one. :)

Note that CoreBundle:Customer.orm.xml is correct and doesn't have softdeleteable duplication of mapping.


michalmarcinkowski:
You're right, the duplication is unnecessary :+1:

PR #3398, opened 2015-10-06
Replace twig code with labe_state macro at backend customer show template

diimpp:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Single difference, that label "yes" would rendered as "YES", which is default behaviour for `Backend/Customer:index.html.twig`. So I consider this as fix in consistency.


michalmarcinkowski:
Thank you Dmitri

Issue #1624, opened 2014-06-18
[BUG] Bug when edit taxonomies's permalink.


liverbool:
Before edit permalink: 
> category/books

After i try to edit to **category/booksx** and then save. Permalink have changed to:
> category/category-booksx

The result should be:
> category/booksx

![screenshot jun 19 00 04 09](https://cloud.githubusercontent.com/assets/206804/3317235/cb808ccc-f70a-11e3-94da-ce2ab15f2d5e.png)

familie-redlich:
I run into the same issue :(

it looks to me if an listener, listen on the flush method and overwrite the permalink every time.
but I cannot locate the exact line of code

workflow: 
create taxonomy -> setLocale and setName
create taxon -> setLocale and setName
add taxon to taxonomy // permalink is null
persist taxonomy // permalink is null
flush // permalink is taxon->parent->permalink . '/' . taxon->slug

change taxon permalink to 'abc' // permalink is abc
persist // permalink is abc
flush // permalink is taxon->parent->permalink . '/' . taxon->slug 

hope this help

familie-redlich:
i found the "problem" i TaxonTranslation.orm.xml -> Sylius\Bundle\TranslationBundle\GedmoHandler\TranslationSlugHandler

the method 'transliterate' build the permalink... I would say its "Works as Designed" :)
i subclass (override) TaxonTranslation and the orm mapping to use my own TranslationSlugHandler ... problem solved.

nmarmon:
"Closed", please

michalmarcinkowski:
Should be fixed now.

PR #3399, opened 2015-10-07
[Translation] Fixed bug while editing taxon permalink

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Sylius/Sylius/issues/1624
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Sylius/Bundle/TranslationBundle/GedmoHandler/TranslationSlugHandler.php
+++ b/src/Sylius/Bundle/TranslationBundle/GedmoHandler/TranslationSlugHandler.php
@@ -129,6 +129,8 @@ class TranslationSlugHandler implements SlugHandlerInterface
                 }
             }
         }
+
+        $slug = $this->deleteUnnecessaryParentSlug($slug);
     }
 
     /**
@@ -154,7 +156,7 @@ class TranslationSlugHandler implements SlugHandlerInterface
      */
     public function onSlugCompletion(SluggableAdapter $ea, array &$config, $object, &$slug)
     {
-        $slug = $this->transliterate($slug, $config['separator'], $object);
+        $slug = $this->transliterate($slug);
 
         if (!$this->isInsert) {
             $wrapped = AbstractWrapper::wrap($object, $this->om);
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/TranslationBundle/GedmoHandler/TranslationSlugHandler.php
```diff
@@ -128,6 +128,8 @@ public function postSlugBuild(SluggableAdapter $ea, array &$config, $object, &$s
                 }
             }
         }
+
+        $slug = $this->deleteUnnecessaryParentSlug($slug);
```
stloyd: This param is never used.

stloyd: I don't see point for placing new one-line method that runs only preg replace.

pjedrzejewski: Yeah, for this one function call I think it could stay here.

Zales0123: *Primo:* Unfortunately ``$slug`` is passed by reference, so it's in fact used somewhere else ;)
*Secundo:* The reason is simple - clean code principles. Let's assume we have that line in the code:
```php
$slug = preg_replace('/^' . preg_quote($this->parentSlug.$this->usedPathSeparator, '/') . '/', '', $slug);
```
Even if your fluent in regular expressions, it takes a to figure out what does this line do. I would probably need a comment to explain it. Seeing this:
```php
$slug = $this->deleteUnnecessaryParentSlug($slug);
```
we are 100% sure what the hell is going on. Just following Uncle Bob advices ;)

pjedrzejewski: Where is Secundo? :D I agree. Let's keep it this way.

pjedrzejewski: Ok I see now. Need to wake up. :)

nakashu:
@pjedrzejewski Why this wasn't merged ? It will also fix this patialy #3662, 3rd example.
@Zales0123 - can you rebase it or should I?

Zales0123:
@nakashu my pleasure with rebasing. I suppose they are too many PR's and such a small change could be easily lost. Nevertheless, @pjedrzejewski if you think it's good to merge - it's ready ;)

michalmarcinkowski:
Thank you Mateusz!

PR #3404, opened 2015-10-07
Removed unused use statement

umpirsky:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT


stloyd:
I guess that this should be done in every file that have such unused `use` (check Scritunizer fixes for details)

PR #3406, opened 2015-10-07
[Resource] Fix ObjectToIdentifierTransformer bad implementation

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Sylius/Sylius/issues/2979
| License       | MIT
| Doc PR        |

While fixing https://github.com/Sylius/Sylius/issues/2979 I realized, that ``ObjectToIdentifierTransformer`` has inversely implemented methods! It caused mentioned bug, but also some others. There is open PR for this issue (https://github.com/Sylius/Sylius/pull/3085) with implemented custom transformer, but I think it will be not necessary after accepting this one ;)

### DIFF
--- a/features/frontend/checkout_shipping.feature
+++ b/features/frontend/checkout_shipping.feature
@@ -100,3 +100,12 @@ Feature: Checkout shipping
          When I press "Continue"
          Then I should be on the checkout addressing step
           And "We're sorry" should appear on the page
+
+    Scenario: Returning to shipping method page
+        Given I go to the checkout start page
+          And I fill in the shipping address to United States
+          And I press "Continue"
+         When I select the "FedEx" radio button
+          And I press "Continue"
+          And I click "Back"
+         Then I should see "FedEx"
--- a/src/Sylius/Bundle/ResourceBundle/Form/DataTransformer/ObjectToIdentifierTransformer.php
+++ b/src/Sylius/Bundle/ResourceBundle/Form/DataTransformer/ObjectToIdentifierTransformer.php
@@ -18,29 +18,22 @@ use Symfony\Component\Form\Exception\UnexpectedTypeException;
 use Symfony\Component\PropertyAccess\PropertyAccess;
 
 /**
- * Object to id transformer.
- *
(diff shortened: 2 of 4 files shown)


michalmarcinkowski:
@Zales0123 what about Behat scenario for this bug?

Zales0123:
@michalmarcinkowski I can of course write scenario like this:
```behat
Scenario: Returning to shipping method page
    Given I go to the checkout start page
      And I fill in the shipping address to United States
      And I press "Continue"
     When I select the "FedEx" radio button
      And I press "Continue"
      And I click "Back"
     Then I should see "FedEx"
```
but it looks quite artificial. Do we need this only because some bug with some class has occurred? I'm really asking, not judging, sincerely :)

michalmarcinkowski:
Could we remove this https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/SettingsBundle/Transformer/ObjectToIdentifierTransformer.php transformer then?

michalmarcinkowski:
@Zales0123 answering to your question. I think we need a debate about the economy of tests and what we should test using Behat. The test suites are really big right now and I'm not sure what to do with them. On one hand we need tests for such basic, but crucial operations like moving back and forth on the checkout on the other hand we can't test every possible UI interaction with Behat, because it will lead to tests execution for days instead of minutes/hours :smile:

Zales0123:
It's more than satisfying answer :+1: Ok, so for now, I'll add this scenario, to be 100% sure that this nasty bug will not happen anymore.
If it's about deleting ``ObjectToIdentifierTransformer`` from *SettingsBundle*, I'm not certain. Yes, it provides identical logic as this fixed transformer from *ResourceBundle*, but it is not used as **form** transformer (notice, it's not even in ``Form``-around namespace), implements specific ``ParameterTransformerInterface``, which is used by ``SettingsBuilder`` and is generally totally different type of object.

pjedrzejewski:
Thank you Mateusz! :+1:

PR #3409, opened 2015-10-08
[Core] Fixed improper ChannelBasedCalculator implementation

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Sylius/Sylius/issues/3253
| License       | MIT
| Doc PR        |

There is another PR fixing this issue, but I think this fixes real source of the problem, which was bad ``ChannelBasedCalculator`` implementation. However, I know that @pjedrzejewski was thinking about some more generic solution, so I would appreciate some feedback ;)

### DIFF
--- a/src/Sylius/Bundle/ProductBundle/Behat/ProductContext.php
+++ b/src/Sylius/Bundle/ProductBundle/Behat/ProductContext.php
@@ -14,6 +14,7 @@ namespace Sylius\Bundle\ProductBundle\Behat;
 use Behat\Gherkin\Node\TableNode;
 use Doctrine\Common\Collections\ArrayCollection;
 use Sylius\Bundle\ResourceBundle\Behat\DefaultContext;
+use Sylius\Component\Core\Model\ProductInterface;
 
 class ProductContext extends DefaultContext
 {
@@ -86,6 +87,10 @@ class ProductContext extends DefaultContext
 
             if (isset($data['deleted']) && 'yes' === $data['deleted']) {
                 $product->setDeletedAt(new \DateTime());
+            }
+
+            if (isset($data['pricing calculator']) && '' !== $data['pricing calculator']) {
+                $this->configureProductPricingCalculator($product, $data);
             }
 
             $manager->persist($product);
@@ -288,5 +293,39 @@ class ProductContext extends DefaultContext
         }
 
         $manager->flush();
+    }
+
+    /**
+     * @param ProductInterface $product
+     * @param array            $data
+     */
+    private function configureProductPricingCalculator(ProductInterface $product, array $data)
+    {
+        $product->getMasterVariant()->setPricingCalculator($data['pricing calculator']);
+
+        if (!isset($data['calculator configuration']) || '' === $data['calculator configuration']) {
+            throw new \InvalidArgumentException('You must set chosen calculator configuration');
+        }
+
+        $product->getMasterVariant()->setPricingConfiguration($this->getPricingCalculatorConfiguration($data));
+    }
+
+    /**
+     * @param array $data
+     *
+     * @return array
+     */
+    private function getPricingCalculatorConfiguration(array $data)
+    {
+        $calculatorConfiguration = $this->getConfiguration($data['calculator configuration']);
+
+        $finalCalculatorConfiguration = array();
+        foreach ($calculatorConfiguration as $channelCode => $price) {
+            $channel = $this->getRepository('channel')->findOneBy(array('code' => $channelCode));
+
+            $finalCalculatorConfiguration[$channel->getId()] = $price * 100;
+        }
+
+        return $finalCalculatorConfiguration;
     }
 }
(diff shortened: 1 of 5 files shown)


src/Sylius/Bundle/ProductBundle/Behat/ProductContext.php
```diff
@@ -289,4 +294,38 @@ public function theFollowingOptionTranslationsExist(TableNode $table)
...
+     */
+    private function getPricingCalculatorConfiguration(array $data)
+    {
+        $calculatorConfiguration = $this->getConfiguration($data['calculator configuration']);
+
+        $finalCalculatorConfiguration = array();
+        foreach ($calculatorConfiguration as $channelCode => $price) {
+            $channel = $this->getRepository('channel')->findOneBy(array('code' => $channelCode));
+
+            $finalCalculatorConfiguration[$channel->getId()] = $price * 100;
```
stloyd: This `$price * 100` is wrong, you should use money helpers for that.

src/Sylius/Bundle/ProductBundle/Behat/ProductContext.php
```diff
@@ -289,4 +294,38 @@ public function theFollowingOptionTranslationsExist(TableNode $table)
...
+     *
+     * @return array
+     */
+    private function getPricingCalculatorConfiguration(array $data)
+    {
+        $calculatorConfiguration = $this->getConfiguration($data['calculator configuration']);
+
+        $finalCalculatorConfiguration = array();
+        foreach ($calculatorConfiguration as $channelCode => $price) {
+            $channel = $this->getRepository('channel')->findOneBy(array('code' => $channelCode));
```
stloyd: Don't fetch repository in loop.

src/Sylius/Component/Core/Pricing/ChannelBasedCalculator.php
```diff
@@ -11,17 +11,43 @@
...
+     * @param array              $configuration
+     * @param array              $context
+     *
+     * @return integer
+     */
+    public function calculate(PriceableInterface $subject, array $configuration, array $context = array())
+    {
+        $currentChannel = $this->channelContext->getChannel();
+
+        return $subject->getPricingConfiguration()[$currentChannel->getId()];
```
stloyd: If such array key not exisist this will produce error.

src/Sylius/Component/Core/spec/Pricing/ChannelBasedCalculatorSpec.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    function it_implements_calculator_interface()
+    {
+        $this->shouldImplement('Sylius\Component\Pricing\Calculator\CalculatorInterface');
+    }
+
+    function it_calculates_subject_price_based_on_current_channel(PriceableInterface $subject, ChannelInterface $channel, $channelContext)
+    {
+        $channelContext->getChannel()->willReturn($channel)->shouldBeCalled();
```
stloyd: Again I will repeat that this approach is wrong ;) If you call `willReturn()` __don't__ call `shouldBeCalled()`...

src/Sylius/Bundle/ProductBundle/Behat/ProductContext.php
```diff
@@ -289,4 +294,40 @@ public function theFollowingOptionTranslationsExist(TableNode $table)
...
+
+    /**
+     * @param ProductInterface $product
+     * @param array            $data
+     */
+    private function configureProductPricingCalculator(ProductInterface $product, array $data)
+    {
+        $product->getMasterVariant()->setPricingCalculator($data['pricing calculator']);
+
+        if (!isset($data['calculator configuration']) || '' === $data['calculator configuration']) {
```
michalmarcinkowski: How about changing `'' === ` to `empty`? I think that empty array is invalid configuration as well?

src/Sylius/Bundle/ProductBundle/Behat/ProductContext.php
```diff
@@ -289,4 +294,40 @@ public function theFollowingOptionTranslationsExist(TableNode $table)
...
+    {
+        $calculatorConfiguration = $this->getConfiguration($data['calculator configuration']);
+
+        $finalCalculatorConfiguration = array();
+        $channelRepository = $this->getRepository('channel');
+
+        foreach ($calculatorConfiguration as $channelCode => $price) {
+            $channel = $channelRepository->findOneBy(array('code' => $channelCode));
+
+            $finalCalculatorConfiguration[$channel->getId()] = (int) round($price * 100);
```
michalmarcinkowski: I don't think we need to `round`. There can't be any precision loss when multiplying by `100`.

src/Sylius/Bundle/ProductBundle/Behat/ProductContext.php
Zales0123: Yeah, but ``$data['caluculator configuration`]`` is not *de facto* an array - it's a string, that is parsed to array in ``$this->getPricingCalculatorConfiguration($data)`` method ;)

michalmarcinkowski: You're right! :+1:

pjedrzejewski:
Thank you Mateusz! This is a good fix for now, but we need to remember that in the long-run we should make the calculators state-less. Sometimes you want to run them on an existing order and you can be in a different channel context. For example when editing an order. We should generally pass info to options, but there is no easy way to do it in the template right now. Not a simple issue...

PR #3412, opened 2015-10-09
[Core] Fix problem authenticating with facebook

samura:
If you don't ask for the field "email" on facebook graph, sylius will crash trying to "INSERT INTO sylius_customer" with an empty e-mail where the field is required.


pjedrzejewski:
Thank you @samura! :+1:

PR #3422, opened 2015-10-13
Made attribute bundle configuration methods protected

mihaichitic:
The goal is to override them in extended classes.

I want to implement AttributeValueTranslation and avoid code duplication (continue to use the validation from parent, and just append 'translation' under 'attribute_value' key)

### DIFF
--- a/src/Sylius/Bundle/AttributeBundle/DependencyInjection/Configuration.php
+++ b/src/Sylius/Bundle/AttributeBundle/DependencyInjection/Configuration.php
@@ -52,7 +52,7 @@ class Configuration implements ConfigurationInterface
      *
      * @param ArrayNodeDefinition $node
      */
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
Hey @mihaichitic! Having AttributeValues translatable is something we would be interested to have in core. Have you considered working on this? We generally do not recommend extending these methods (and configuration classes in general), that's why they are private. :) Let me know!

Tetragramat:
I'm sorry, that referenced comits are my big mistake.

pjedrzejewski:
@mihaichitic Please reopen if still interested in this issue. Translatable attribute values is a separate topic, but generally I don't think we will ever recommend extending these classes thus they should stay private. Thanks!

a2xchip:
:open_mouth: Are not attribute values translatable? :scream: @pjedrzejewski

PR #3428, opened 2015-10-13
Flush only payments in PaymentProcessor

umpirsky:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| License | MIT |

While I find it wrong to flush in `PaymentProcessor`, I find it even worse to flush everything.
I know it is not optimal to flush them one by one, but at least I am preventing `PaymentProcessor` to flush what shouldn't be flushed.


michalmarcinkowski:
Outdated, we no longer flush in payment processor.

PR #3430, opened 2015-10-14
Fixed body padding when having the sticky toolbar

rpg600:


michalmarcinkowski:
Thank you Reinis! :+1:

PR #3431, opened 2015-10-14
Added quotes to fix strings ending with :

pfwd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Added quotes around English translation strings that ended in : as these were breaking yaml parsers and IDES such as netbeans

### DIFF
--- a/src/Sylius/Bundle/WebBundle/Resources/translations/messages.en.yml
+++ b/src/Sylius/Bundle/WebBundle/Resources/translations/messages.en.yml
@@ -36,8 +36,8 @@ sylius:
             list: All your addresses
             update: Edit my address
         home:
-            email: "Email :"
-            id: "Customer ID :"
+            email: Email 
+            id: Customer ID
             outline: Enjoy all the features that are available on your personal space to view, track and manage all your data.
             since: You are customer since
             welcome: Welcome to your space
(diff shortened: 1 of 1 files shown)


stloyd:
IMO all those `:` should be removed from translations.

pfwd:
No problem. I can remove them if you want?

pfwd:
@stloyd I've removed them

src/Sylius/Bundle/WebBundle/Resources/translations/messages.en.yml
```diff
@@ -575,7 +575,7 @@ sylius:
         state: State
         immutable_warning: Already shipped orders are immutable.
         promotion_coupons: Coupons
-        no_promotion: No coupons were used.order:
+        no_promotion: No coupons were used.order
```
stloyd: This one looks like wrong merge, IMO should be `No coupons were used.`

pjedrzejewski:
Thanks @pfwd Peter for raising this! We are currently working on keys transformer, which will simplify all keys in Sylius (so far 1393 -> 796), but preserve translations and we will remove all ":" automatically when running this transformation. :) I will leave this open so that we keep the track of this problem and if we discover any problem with the solution.

pfwd:
OK sounds good. Thanks for looking into it :+1:

michalmarcinkowski:
Fixed in #3908.

PR #3432, opened 2015-10-14
A Docker Compose configuration to develop in Docker containers.

jverdeyen:
With this Docker setup it's very easy to test drive all kind of services like Elasticsearch, Redis etc..

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | https://github.com/jverdeyen/Sylius/blob/feature-docker-development-environment/docker/README.md

### DIFF
--- a/etc/docker/php/php.ini
+++ b/etc/docker/php/php.ini
@@ -1,0 +1,4 @@
+date.timezone = Europe/Brussels
+upload_max_filesize = 40M
+post_max_size = 40M
+memory_limit = 2048M
--- a/etc/docker/README.md
+++ b/etc/docker/README.md
@@ -1,0 +1,69 @@
+# Description
+This Docker Compose development environment includes
+
+* PHP 7.0
+* MariaDB
+* Nginx 
+* Composer
+
+# Usage
+
+First you need to install Docker and Docker Compose.
+
+```bash
+cd docker
+docker-compose up
+```
+
+Now you have a few options to get started
+
+## Basic
+
+Get the ip of the Nginx container.
+
+```
+docker inspect $(docker-compose ps -q nginx) | grep IPAddress
+```
+
+## Advanced
+
+Run a `dnsdock` container before `docker-compose up`, more info: https://github.com/tonistiigi/dnsdock
+Access the containers from the dns records.
+
+# Troubleshooting
+
+## How to enter a container?
+
+Enter the php container to install composer vendors etc.
+
+```bash
+docker exec -it $(docker-compose ps -q php) bash
+```
+
+## The application is too slow.
+
+Install composer vendors in the container and symlink them to the application directory.
+Execute inside the php container:
+
+```bash
+mkdir /vendor && ln -sf /vendor ./vendor
+```
+
+Using Symfony2 inside Vagrant can be slow due to synchronisation delay incurred by NFS.
+You can write the app logs and cache to a folder on the php container.
+
+Enter the php container and create the directory:
+
+```bash
+docker exec -it $(docker-compose ps -q php) bash
+mkdir /dev/shm/sylius/
+setfacl -R -m u:"www-data":rwX -m u:`whoami`:rwX /dev/shm/sylius/
+setfacl -dR -m u:"www-data":rwX -m u:`whoami`:rwX /dev/shm/sylius/
+```
+
+To view the application logs, run the following commands:
+
+```bash
+tail -f /dev/shm/sylius/app/logs/prod.log
+tail -f
+```
--- a/etc/docker/docker-compose.yml
+++ b/etc/docker/docker-compose.yml
@@ -1,0 +1,45 @@
+data:
+    image: busybox
+    volumes:
+        - ../../:/var/www/sylius
+        - /vendor
+        - ~/.ssh/id_rsa:/root/.ssh/id_rsa:ro
+        - ~/.composer:/root/.composer
+        - /tmpfs
+
+nginx:
+    image: nginx
+    links:
+        - php
+    volumes_from:
+        - data
+    volumes:
+        - ./nginx/default.conf:/etc/nginx/conf.d/default.conf:ro
+
+mysql:
+    image: mariadb
+    environment:
+        MYSQL_ROOT_PASSWORD: root
+        MYSQL_DATABASE: sylius_dev
+        MYSQL_USER: sylius
+        MYSQL_PASSWORD: sylius
+
+php:
+    build: php
+    working_dir: /var/www/sylius
+    volumes_from:
+        - data
+    links:
+        - mysql
+        - elasticsearch
+        - memcached
+    mem_limit: 2000000000
+
+elasticsearch:
+    image: elasticsearch
+
+memcached:
+    image: memcached
+
+selenium:
+    image: selenium/standalone-chrome-debug
--- a/etc/docker/nginx/default.conf
+++ b/etc/docker/nginx/default.conf
@@ -1,0 +1,59 @@
+upstream php-upstream {
+  server php:9000;
+}
+
+server {
+  listen   80;
+
+  root /var/www/sylius/web;
+  server_tokens on;
+
+  access_log  /var/log/nginx/access.log;
+  error_log  /var/log/nginx/error.log;
+
+  client_max_body_size 200M;
+
+  index app_dev.php;
+
+  location / {
+    try_files $uri $uri/ /app_dev.php$is_args$args;
+  }
+
+  location ~ \.php$ {
+    fastcgi_pass              php-upstream;
+    fastcgi_param SERVER_NAME       $host;
+    fastcgi_split_path_info   ^(.+\.php)(/.*)$;
+
+    fastcgi_param QUERY_STRING      $query_string;
+    fastcgi_param REQUEST_METHOD    $request_method;
+    fastcgi_param CONTENT_TYPE      $content_type;
+    fastcgi_param CONTENT_LENGTH    $content_length;
+
+    fastcgi_param SCRIPT_FILENAME   $document_root$fastcgi_script_name;
+    fastcgi_param PATH_INFO         $fastcgi_script_name;
+
+    fastcgi_param SCRIPT_NAME       $fastcgi_script_name;
+    fastcgi_param REQUEST_URI       $request_uri;
+    fastcgi_param DOCUMENT_URI      $document_uri;
+    fastcgi_param /var/www/app/web     $document_root;
+    fastcgi_param SERVER_PROTOCOL   $server_protocol;
+
+    fastcgi_param GATEWAY_INTERFACE CGI/1.1;
+    fastcgi_param SERVER_SOFTWARE   nginx/$nginx_version;
+
+    fastcgi_param REMOTE_ADDR       $remote_addr;
+    fastcgi_param REMOTE_PORT       $remote_port;
+    fastcgi_param SERVER_ADDR       $server_addr;
+    fastcgi_param SERVER_PORT       $server_port;
+    fastcgi_param SERVER_NAME       $server_name;
+
+    fastcgi_param HTTPS             $https;
+
+    # PHP only, required if PHP was built with --enable-force-cgi-redirect
+    fastcgi_param REDIRECT_STATUS   200;
+
+    fastcgi_param             SCRIPT_FILENAME    $document_root$fastcgi_script_name;
+    fastcgi_param             HTTPS              off;
+    fastcgi_read_timeout      120;
+  }
+}
--- a/etc/docker/php/Dockerfile
+++ b/etc/docker/php/Dockerfile
@@ -1,0 +1,31 @@
+FROM php:7.0.2-fpm
+
+RUN \
+  apt-get update && \
+  apt-get install -y \
+  curl \
+  wget \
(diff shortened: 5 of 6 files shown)


docker/README.md
```diff
@@ -0,0 +1,86 @@
...
+```bash
+tail -f /dev/shm/sylius/app/logs/prod.log
+tail -f
+```
+
+## No frontcontroller access
+
+Remove the following lines from `app_dev.php`
+
+```
```
stloyd: Add `php` here to enable highlight.

docker/README.md
```diff
@@ -0,0 +1,86 @@
+# Description
+This Docker Compose development environment includes
+
+* PHP 5.6.*
+* MySQL 5.6
+* Nginx 1.6.2
```
stloyd: Please add ElasticSearch too.

stloyd: Also is Git available?

pjedrzejewski:
Awesome work Joeri! I am not a Docker expert, so I will appreciate any comments on this, as always. :) Thanks for your work man!

docker/README.md
amenophis: Very bad practice to remove this protection !! Will be better to change te code to check with an ENV Var (eg. SYLIUS_ENV=dev or SYLIUS_ENV=prod).
This will not allow to deploy this unsecured version of the dev controller.

docker/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
+data:
+    image: busybox
+    volumes:
+        - ../:/var/www/app
+
+nginx:
+    image: yappabe/nginx
```
amenophis: why not using official images here ?

docker/README.md
stloyd: Agree about that, and IIRC there other workarounds than this with env.

michaelperrin:
Really great to add a Docker Compose configuration!
I am wondering if there is a specific reason not to use the official PHP and MySQL Docker images?

docker/README.md
jverdeyen: Adding ElasticSearch is as easy as adding the following lines:
```
elasticsearch:
    image: elasticsearch
    environment:
        DNSDOCK_ALIAS: elasticsearch.sylius.docker
```
Should this be added in a default development environment?

Git is available in the `php` container, but I don't think it's good practice to use git inside a container while developing. It's required for `composer`.

docker/docker-compose.yml
jverdeyen: With these images you don't need extra configuration files nor custom Dockerfiles, it's plug and play.

l3l0: To use offical nginx image you do not need to define custom dockerfile, basically you need to use one of offical php-fpm images and define own default.conf.

Official nginx image is better cause you can control version of nginx then. This image does not have any tags as well, so I would not use in my own project ;)

amenophis: I think it is better to user official image through local Dockerfile, it will be better to allow user override the configuration

docker/README.md
blazarecki: I think you should create a `docker-compose.yml.dist`. 

When someone want to use docker with Sylius, he copy the dist to create the `docker-compose.yml`. Don't forget to add it in the gitignore.

Soon, docker-compose will use file "docker-compose.override.yml" to override configuration (https://github.com/docker/compose/blob/master/compose/config/config.py#L79).

docker/docker-compose.yml
blazarecki: :+1: for official nginx and php images. 

I also recommend to use `php:5.6-apache`. Like this you have only one container for the `web`. I think it's easier to maintain.

What do you think ?

docker/docker-compose.yml
```diff
@@ -0,0 +1,46 @@
...
+    mem_limit: 2000000000
+
+mailcatcher:
+    image: yappabe/mailcatcher
+    environment:
+        DNSDOCK_ALIAS: mailcatcher.sylius.docker
+
+# password: secret
+# vnc port 5999
+selenium:
```
blazarecki: Same here you can use official selenium images : `selenium/standalone-firefox`

docker/docker-compose.yml
jverdeyen: Yes, official images need some additional configuration and this setup should be straight-forward for a developer. I wanted to provide an easy way to develop Sylius within containers.

docker/docker-compose.yml
jverdeyen: This image provides a vnc-server to see the features running.

pamil: Using Firefox for our `javascript` Behat suite is not the best idea - I tried a lot of options while I was integrating JS tests with Travis and Chrome is the best available choice.

docker/docker-compose.yml
blazarecki: You can add all configuration files, Dockerfiles and other stuff in a `docker` dir at the root of the project. (same as `vagrant`).

For developers, it will be straight-forward since he doesn't have to do anything to make it work.

docker/docker-compose.yml
blazarecki: You can use `selenium/standalone-chrome`. And there is a VNC server in the official image. 

see https://github.com/SeleniumHQ/docker-selenium

docker/docker-compose.yml
pjedrzejewski: :+1: @blazarecki @l3l0 @amenophis

docker/docker-compose.yml
pjedrzejewski: :+1: for Chrome. :) Trust @pamil on that. :D

pjedrzejewski:
Hey @jverdeyen, let us know if you plan on continuing this PR. :)

jverdeyen:
I've updated this PR with the official images etc

pjedrzejewski:
@jverdeyen Awesome, I will have a look but I am by no means Docker expert, so any feedback here is appreciated as always, let's give the community few days to have a look. Great work, thank you!

cdaguerre:
I think `wkhtmltopdf` needs to be included aswell.
There is a [docker image](https://hub.docker.com/r/openlabs/docker-wkhtmltopdf-aas/) to use it as a service over TCP but it's not really satisfactory and it won't work with Snappy anyway, so I think it should be installed in the fpm service (just like composer).

```shell
WKHTMLTOPDF_VERSION=0.12
WKHTMLTOPDF_RELEASE=0.12.1

curl -so /tmp/wkhtmltopdf.deb "http://download.gna.org/wkhtmltopdf/${WKHTMLTOPDF_VERSION}/${WKHTMLTOPDF_RELEASE}/wkhtmltox-${WKHTMLTOPDF_RELEASE}_linux-trusty-amd64.deb"
dpkg -x /tmp/wkhtmltopdf.deb /
mv /usr/local/bin/wkhtmltopdf /usr/bin/wkhtmltopdf
mv /usr/local/bin/wkhtmltoimage /usr/bin/wkhtmltoimage
```

steffenbrem:
I can contribute some stuff here. I use docker too with sylius in a specialized vagrant setup. Getting 40-100ms response times and 100-200 on app_dev. It also supports both windows and automatically switches to smb and uses nfs2 on mac etc.

steffenbrem:
I will post some gists soon after I eat

pjedrzejewski:
@steffenbrem 100-200 on app_dev? Woah, I get 700-800 up to ~1.3s (form submits) on the built-in server. That sounds really cool, waiting for the gist! :D Bon appetit haha!

steffenbrem:
@pjedrzejewski On the api side, I get around 200ms, yes. But store front I get 500ms on app dev. Also very fast for vagrant setup + docker containers.

steffenbrem:
https://gist.github.com/steffenbrem/fa4a777dea2e882c859d

This gist contains a `Vagrantfile` that uses a seperate `vagrant.yml` file for the most important configurations. It will automatically detect if your using Windows and use `smb` for synced folders and `nfs2` on Mac. It's just a matter of running `vagrant up` and then `docker-compose up` and your good to go, Windows or Mac, doesn't matter.

I can also provide you the setup script I use to setup a mac machine with the needed software, think of Vagrant, Docker Compose, VirtualBox etc.

steffenbrem:
If you also like a good script that automatically setups up everything on an empty mac osx installation, check this gist: https://gist.github.com/steffenbrem/925fa2a28a0c73598124

steffenbrem:
@jverdeyen What do you think about including this kind of vagrant configuration?

@pjedrzejewski If we would include this kind of vagrant stuff, should we get rid of the existing one? Would be nice to force every one in the community to use the docker containers, then we would all work on the same system configuration.

jverdeyen:
@steffenbrem Looks good, I'm using this setup: http://tech.yappa.be/docker-php-development
We've been tweaking this for the best performance and stability.

roukmoute:
It's a bad idea to use `root` in php container.
Good thing is to connect with our same username in our bash.

Personally I use [this file](https://github.com/roukmoute/docker-symfony-bootstrap/blob/b0b9e063e93225b095207dcbb3957d9f69b464ee/php/docker-entrypoint.sh) to create a user.

Next to connect with our username:
`docker exec -it $(docker-compose ps -q php) /bin/login -p -f our_username`

xantrix:
@jverdeyen amazing work! 
I had a tremendous performance improvement compared to vagrant setup.
I've just one issue with Behat / Selenium configuration. 
I have linked the selenium container to the php one
[see here](https://github.com/xantrix/Sylius/blob/feature-docker-development-environment/etc/docker/docker-compose.yml#L33)
and set selenium2 **wd_host** correctly to selenium container (http://selenium:4444/wd/hub)
in my behat.yml configuration
```
imports: ["behat.yml.dist"]
--SNIP--
                chrome:
                    selenium2:
                        wd_host: http://selenium:4444/wd/hub
```

But I still get an 
**Failed to set the 'cookie' property on 'Document': Cookies are disabled inside 'data:' URLs.**

```
15:30:24.286 INFO - Executing: [get: http://localhost:8080/])
15:30:24.446 INFO - Done: [get: http://localhost:8080/]
15:30:24.463 INFO - Executing: [add cookie: MOCKSESSID=6e2d948bdb10f9f00053b4b5e133b1cd3705264e5f843beeefeadbd64eaf06b2; path=/])
15:30:24.629 WARN - Exception thrown
org.openqa.selenium.WebDriverException: <unknown>: **Failed to set the 'cookie' property on 'Document': Cookies are disabled inside 'data:' URLs.**
  (Session info: chrome=48.0.2564.109)
  (Driver info: chromedriver=2.20.353124 (035346203162d32c80f1dce587c8154a1efa0c3b),platform=Linux 3.13.0-44-generic x86_64) (WARNING: The server did not provide any stacktrace information)

```
with javascript scenarios.

I googled for a while without result. 
Any clues ?
Thanks in advance for any help.

jverdeyen:
@xantrix Thanks!

Did tried this with the `selenium/standalone-firefox-debug` image? Just to check if this is the image' behaviour or the setup.

pjedrzejewski:
Ok guys, if nobody has any comments, then I will be happy to merge this! The issue with Selenium can be resolved in separate PR from someone who figures it out, wdyt?

xantrix:
@pjedrzejewski sure, we can continue the discussion in a separate issue.

jverdeyen:
@pjedrzejewski okay, Docker is still moving fast. So there will always be better ways to solve problems etc.. or opinions ;))

xantrix:
@jverdeyen :+1: awesome!

pamil:
@xantrix can you give the example of scenarios that are failing? The `wd_host` should be also added at `sessions.firefox.selenium2`. Anyway, you may want to have a look at `Sylius\Behat\CookieSetter::setCookie` for current Behat and `Sylius\Bundle\CoreBundle\Behat\CoreContext::iAmLoggedInAsRole` / `Sylius\Bundle\CoreBundle\Behat\CoreContext::prepareSessionIfNeeded` for legacy Behat.

etc/docker/php/php.ini
```diff
@@ -0,0 +1,4 @@
+date.timezone = Europe/Brussels
```
steffenbrem: Isn't it better to use UTC here?

pjedrzejewski:
Thank you Jeori for your work! This is really cool, I will try it out and we should document it in docs.sylius.org as well. ;)

jverdeyen:
@pjedrzejewski okay, I'll try to help you out! 🎉

xantrix:
@pamil Thanks for your suggestion, for sure there was something misconfigured in my side before, now I'm able to run behat without problem (nor cookie error or timeout) also with selenium/standalone-chrome-debug.

I confirm this @jverdeyen PR is amazing I had just to fine tuning some params.
If anyone need any support on this I'm happy to share my config.
Thanks again!

jverdeyen:
@xantrix Thanks!
We've fast page loads in dev when using our custom vagrant setup, optimized for docker and faster nfs mounts.

https://github.com/yappabe/vagrant-docker

We wrote an article about our workflow:
http://tech.yappa.be/docker-php-development

PR #3433, opened 2015-10-14

SyliusBot:


pamil:
- Added support for dialects (`pt_BR` is no longer `pt`, `de_CH` is not merged into `de`)
- Added translations for new bundles like `UserBundle`

As dialects are up and running we shouldn't translate them whole, but just customize the differences from  main language (ex. `de_CH` should only customize `de` in such fields as changing `provincez translation from `Bundesländer` to `Kanton`).

pjedrzejewski:
Thanks @SyliusBot, great work @pamil! :+1:

Issue #2615, opened 2015-03-23
[Storage] CookieStorage provider not working properly


MichaelMackus:
Seems as though the cart cookie isn't being set properly (or even used - see: https://github.com/Sylius/Sylius/pull/2089#issuecomment-62068728).

I've tried setting invalidate_session: false and setting sylius_cart storage provider to the "cookie" storage. I then added a listener to the kernel response to set the cookie manually there. But the cart is still not working properly with the CookieStorage provider.

steffenbrem:
I had this issue too. See this link: https://github.com/Sylius/Sylius/pull/2089#issuecomment-62068728

pjedrzejewski:
Duplicate of #3009.

PR #3437, opened 2015-10-15
Fix cookie storage by attache cookie to response.

### LINKED ISSUE #3009: Does Cookie storage actually persist?
CookieStorage's set method only set data to the request's cookies header, I think the data won't persist until we set cookie in the HTTP respond. Am I missing anything?

https://github.com/Sylius/Sylius/blob/0c15f8353e5c70817d7d4d23c5ff53eccd4f8015/src/Sylius/Component/Storage/CookieStorage.php

/cc @stloyd

kayue:
Fix #3009, #2615.

### DIFF
--- a/src/Sylius/Component/Storage/CookieStorage.php
+++ b/src/Sylius/Component/Storage/CookieStorage.php
@@ -11,7 +11,10 @@
 
 namespace Sylius\Component\Storage;
 
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\HttpFoundation\Cookie;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
 
 /**
  * @author Joseph Bielawski <stloyd@gmail.com>
@@ -22,6 +25,12 @@ class CookieStorage implements StorageInterface
      * @var Request
      */
     protected $request;
+    protected $eventDispatcher;
+
+    public function __construct(EventDispatcherInterface $eventDispatcher)
+    {
+        $this->eventDispatcher = $eventDispatcher;
+    }
 
     /**
      * @param Request $request
@@ -53,6 +62,12 @@ class CookieStorage implements StorageInterface
     public function setData($key, $value)
     {
         $this->request->cookies->set($key, $value);
+
+        $this->eventDispatcher->addListener('kernel.response', function (FilterResponseEvent $event) use ($key, $value) {
+            $event->getResponse()->headers->setCookie(
+                new Cookie($key, $value, new \DateTime('+30 days'))
+            );
+        });
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Sylius/Component/Storage/CookieStorage.php
```diff
@@ -61,5 +76,11 @@ public function setData($key, $value)
     public function removeData($key)
     {
         $this->request->cookies->remove($key);
+
+        $this->eventDispatcher->addListener('kernel.response', function (FilterResponseEvent $event) use ($key) {
+            $event->getResponse()->headers->setCookie(
```
winzou: `clearCookie` instead ;)

src/Sylius/Component/Storage/CookieStorage.php
```diff
@@ -53,6 +62,12 @@ public function getData($key, $default = null)
     public function setData($key, $value)
     {
         $this->request->cookies->set($key, $value);
+
+        $this->eventDispatcher->addListener('kernel.response', function (FilterResponseEvent $event) use ($key, $value) {
+            $event->getResponse()->headers->setCookie(
+                new Cookie($key, $value, new \DateTime('+30 days'))
```
winzou: this could be a parameter

winzou:
Crazy fast, thanks!

src/Sylius/Component/Storage/CookieStorage.php
kayue: Should i change the interface?

winzou: No, I would inject it directly in the constructor

src/Sylius/Component/Storage/CookieStorage.php
```diff
@@ -53,6 +64,12 @@ public function getData($key, $default = null)
     public function setData($key, $value)
     {
         $this->request->cookies->set($key, $value);
+
+        $this->eventDispatcher->addListener('kernel.response', function (FilterResponseEvent $event) use ($key, $value) {
```
pjedrzejewski: This is coupling the component to Symfony. Is there any way to avoid that? I'd be fine with coupling it to PSR-7.

kayue: @pjedrzejewski Perhaps we should move this to CoreBundle?

Symfony 2.3 doesn't support PSR-7 anyway.

pjedrzejewski: We are bumping the requirement to 2.7.5, does it support it?

pjedrzejewski: If ^ is no, then yeah, let's move it to core bundle.

pjedrzejewski:
Ping @kayue.

asjustas:
Ping @kayue.

PR #3445, opened 2015-10-16
Fix wrong transformer in LocaleChoiceType

winzou:
The `choice` field has `multiple` false by default, but the `sylius_locale_choice` was adding a wrong `CollectionToArray transformer`


src/Sylius/Bundle/LocaleBundle/Form/Type/LocaleChoiceType.php
```diff
@@ -43,8 +43,6 @@ public function __construct(ObjectRepository $repository)
     public function buildForm(FormBuilderInterface $builder, array $options)
```
aramalipoor: Why do you need this at all!? ;)

pjedrzejewski: :+1:

PR #3449, opened 2015-10-16
Symfony update to 2.7.5

TheMadeleine:
| Q                    | A
| -------------       | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Update due to the 'case mismatch' bug detected while setting up Sylius locally. 
PR from Symfony: https://github.com/symfony/symfony/pull/15783

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,7 +70,7 @@
         "symfony/options-resolver":             "^2.7",
         "symfony/property-access":              "^2.7",
         "symfony/swiftmailer-bundle":           "^2.3",
-        "symfony/symfony":                      "^2.7",
+        "symfony/symfony":                      "^2.7.5",
         "symfony/twig-bundle":                  "^2.7",
         "twig/extensions":                      "~1.0",
         "twig/twig":                            "~1.11",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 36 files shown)


pamil:
We need to bump requirements in `composer.json` too.

TheMadeleine:
@pamil  :+1:

composer.json
```diff
@@ -67,7 +67,7 @@
         "symfony/options-resolver":             "^2.7",
         "symfony/property-access":              "^2.7",
         "symfony/swiftmailer-bundle":           "^2.3",
-        "symfony/symfony":                      "^2.7",
+        "symfony/symfony":                      "^2.7.5",
```
michalmarcinkowski: Proper criteria should be `>=2.7.5, <3.0.0`

pamil: `^2.7.5` means the same :)

michalmarcinkowski: I somehow saw tilde instead of caret :smile: my bad then :wink:

pjedrzejewski:
@pamil Do you mean just main ``composer.json``? (singular form) If yes, then I will add that we should adjust all component&bundle jsons.

TheMadeleine:
@pjedrzejewski I'll take care of it ASAP :)

TheMadeleine:
Okay @pjedrzejewski what if, we don't have any dependencies to symfony/symfony or symfony/debug (which was being fixed by that PR in symfony)? :) Soo all the component & bundle jsons are already adjusted, am I right?

pjedrzejewski:
Thank you Magda!

PR #3460, opened 2015-10-20
[ReportBundle] Remove wrong translation

psyray:
Hi,

Remove wrong translation from report bundle.
It's promotion bundle ones, but they already exists in this bundle.

So I remove them.


pjedrzejewski:
Thanks Raynald! :+1:

PR #3463, opened 2015-10-20
Removed data transformer call in EntityHiddenType

rpg600:
The entity_hidden form type was broken due to PR #3406.


rpg600:
Another thing came to my mind : i'm not 100% sure but i think that the `data_class` option should not be used in the EntityHiddenType : it need to stay null, otherwise the symfony Form class will compare the view data (which is a scalar due to the `ObjectToIdentifierTransformer`) and the data_class then will throw an exception because it expects the view data as an instance of data_class : https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Form/Form.php#L373

pjedrzejewski:
@rpg600 Could you rebase this PR, please? Thanks! Or let us know if you miss the time to work on this.

rpg600:
Rebased.

michalmarcinkowski:
Should be fixed in #3831

michalmarcinkowski:
There is no `Sylius/Bundle/ResourceBundle/Form/Type/EntityHiddenType` class anymore.

PR #3465, opened 2015-10-20
Fix taxonomy bundle configuration constraint

diimpp:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Tested to be working correctly. Besides other bundles uses same schema.


pjedrzejewski:
Nice, thank you very much Dmitri! :)

PR #3466, opened 2015-10-20
Introduce mainTaxon field to Product entity

tuka217:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | no
| License       | MIT
| Doc PR        | no

[CoreBundle] Add new fields: mainTaxon in ProductType,main_taxon_id in Product.orm.xml, add new spec for ProductType
[ProductBundle] Add method for step from background for products.feature
[TaxonomyBundle] Add method getNonRootTaxons in TaxonRepository, edit congifureOptions function in TaxonChoiceType, add spec for getNonRootTaxons, edit it_has_options in TaxonChoiceTypeSpec
[WebBundle] Add messages for new fields MainTaxon in messages.en.yml, in background for product: _categorization.html.twig - add new row MainTaxon, show.hmtl.twig - add new row in general info table
[Core] Add methods getMainTaxon, setMainTaxon in Product and ProductInerface, add specs for those methods


src/Sylius/Bundle/CoreBundle/Form/Type/ProductType.php
```diff
@@ -59,6 +58,11 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'empty_value' => '---',
                 'label'       => 'sylius.form.product.restricted_zone',
             ))
+            ->add('mainTaxon', 'sylius_taxon_choice', array(
+                'filter' => null,
```
pjedrzejewski: These options are default, I think. We do not need them, hm?

src/Sylius/Component/Core/Model/Product.php
```diff
@@ -359,4 +363,20 @@ public function setShortDescription($shortDescription)
...
+     */
+    public function getMainTaxon()
+    {
+       return $this->mainTaxon;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setMainTaxon(TaxonInterface $mainTaxon)
```
pjedrzejewski: We should allow passing ``null`` value here. :)

src/Sylius/Component/Core/Model/ProductInterface.php
```diff
@@ -146,4 +145,14 @@ public function getImages();
...
+
+    /**
+     * @return BaseTaxonInterface
+     */
+    public function getMainTaxon();
+
+    /**
+     * @param TaxonInterface $mainTaxon
+     */
+    public function setMainTaxon(TaxonInterface $mainTaxon);
```
pjedrzejewski: Same comment as above.

pjedrzejewski:
Nice, please fix the comments and have a look at the failing build. (check screenshots and log files at the end of build log for tips, probably)

features/backend/products.feature
```diff
@@ -190,3 +191,34 @@ Feature: Products
         Given I am on the product index page
         When I click "Mug"
         Then I should be on the page of product "Mug"
+
+    Scenario: Creating product with main taxon
+        Given I am on the product creation page
+        And I fill in the following:
+            |Name       | The best T-shirt        |
+            |Description| Interesting description |
+            |Price      | 2.99                    |
```
Zales0123: Missing spaces.

src/Sylius/Bundle/CoreBundle/Form/Type/ProductType.php
```diff
@@ -59,6 +58,11 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                'taxonomy' => null,
+                'label' => 'Main taxon'
```
Zales0123: I suppose, it should be translated.

src/Sylius/Bundle/CoreBundle/spec/Form/Type/ProductTypeSpec.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+namespace spec\Sylius\Bundle\CoreBundle\Form\Type;
```
Zales0123: Missing license block.

src/Sylius/Bundle/CoreBundle/spec/Form/Type/ProductTypeSpec.php
```diff
@@ -0,0 +1,85 @@
...
+
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Sylius\Component\User\Canonicalizer\CanonicalizerInterface;
+use Symfony\Component\Form\FormBuilderInterface;
+
+class ProductTypeSpec extends ObjectBehavior
```
Zales0123: Missing author docblock.

src/Sylius/Bundle/CoreBundle/spec/Form/Type/ProductTypeSpec.php
```diff
@@ -0,0 +1,85 @@
...
+            ->add('masterVariant', 'sylius_product_variant', Argument::cetera())
+            ->shouldBeCalled()
+            ->willReturn($builder)
+        ;
+        $builder
+            ->add('attributes', 'collection', Argument::cetera())
+            ->shouldBeCalled()
+            ->willReturn($builder)
+        ;
+        $builder->add('options', 'sylius_product_option_choice', Argument::cetera())
```
Zales0123: Be consistent and move ``->add...`` to next line ;)

src/Sylius/Bundle/ProductBundle/Behat/ProductContext.php
```diff
@@ -325,9 +325,22 @@ private function getPricingCalculatorConfiguration(array $data)
...
+     * @Given product :productName has main taxon :mainTaxonName
+     */
+    public function productHasMainTaxon($productName, $mainTaxonName)
+    {
+        $manager = $this->getEntityManager();
+
+        $product = $this->findOneByName('product', $productName);
+        $mainTaxon = $this->findOneByName('taxon', $mainTaxonName);
+        $product->setMainTaxon($mainTaxon);
+        $manager->flush();
```
Zales0123: I think we could flush only changed object: ``$manager->flush($product);``

src/Sylius/Bundle/TaxonomyBundle/Form/Type/TaxonChoiceType.php
```diff
@@ -74,13 +80,13 @@ public function configureOptions(OptionsResolver $resolver)
...
         $resolver
             ->setDefaults(array(
-                'choice_list' => $choiceList,
+                    'choice_list' => $choiceList,
             ))
             ->setRequired(array(
-                'taxonomy',
-                'filter',
+                    'taxonomy',
+                    'filter',
```
Zales0123: Is such a big indent required (here and above)?

src/Sylius/Bundle/TaxonomyBundle/spec/Doctrine/ORM/TaxonRepositorySpec.php
```diff
@@ -65,4 +66,20 @@ function it_finds_one_taxon_by_permalink($em, QueryBuilder $builder, AbstractQue
...
+
+    function it_finds_non_root_taxons($em, QueryBuilder $builder, AbstractQuery $query)
+    {
+        $em->createQueryBuilder()->shouldBeCalled()->willReturn($builder);
+        $builder->select('o')->shouldBeCalled()->willReturn($builder);
+        $builder->from(Argument::any(), 'o', Argument::cetera())->shouldBeCalled()->willReturn($builder);
+        $builder->addSelect('translation')->shouldBeCalled()->willReturn($builder);
+        $builder->leftJoin('o.translations', 'translation')->shouldBeCalled()->willReturn($builder);
+        $builder->where('o.parent IS NOT NULL')->shouldBeCalled()->willReturn($builder);
+        $builder->orderBy('o.left')->shouldBeCalled()->willReturn($builder);
```
Zales0123: I would add some blank lines and/or work on formatting to improve readability.

peteward:
What's the point of a main taxon?

okwinza:
@peteward as a quick example out of my head having such a relation will allow you to easy customize your products URL so you can have "{taxon_slug}/{product_slug}" which is more convenient for some projects than the current scheme IMO.

:+1: :+1: would like to see it merged since travis is green. 
Or there is some kind of blocker on this? @pjedrzejewski

peteward:
:-1: 

For the use-case you have suggested I think Archetypes covers this and adding a Main Taxon clashes with exactly what Archetype is.  It adds confusion for me.

If you can come up with a better use-case or a reason why Archetypes don't work here then fair enough.

pjedrzejewski:
@peteward Because I imagine Archetypes as Types of products, which are 1) templates for creating/maintaining similar products 2) having perhaps some configuration options specific for a type of products.

I've never seen Archetypes as a way to categorize products and definitely not for nice URL generation etc. :( Is having mainTaxon is a big obstacle for you guy? We could think how to merge best of these 2.

peteward:
@pjedrzejewski Archetypes by there very nature define groupings of things, just at a definition level rather than a container.

A product IS AN archetype
A product CAN BE IN A category

If a product is a something then it definitely makes sense for it to include that in it's URL when generating it, e.g. 'reiss-bless-t-shirt-in-red'.  It has the word T-Shirt in it's url because this is it's Archetype - it's what the product is. It doesn't necessarily mean that it will be in the t-shirts category. You might remove it if it's not selling well or old season. It might still be searchable or be in a 'Sale' category.

I'm just saying that the use-case given doesn't justify adding a Main Taxon to me, it bypasses the use of Archetypes which is pretty much perfect in this case - especially seeing as they are themselves hierarchical.

This causes me no problems whatsoever, I just think it's duplication.

peteward:
You shouldn't need to join product -> archetype -> category, Why would you be relating archetype to category?

We manage all of our products outside Sylius and I think that gives even more reason to use Archetypes when you have a larger set of provided products.  I was more thinking of smaller systems that DO need to configure all product attributes themselves manually then Archetypes would be an extra step of work and I would concede the point.

I'm all for flexibility in Sylius but I think it should promote a 'core approach' and because this starts to present 2 slightly different ways of doing the same thing, I would consider this more of a Marketplace feature than something core, but we're obviously nowhere near a Market Place yet...

nakashu:
@tuka217 can you please add db migration for these changes?

pjedrzejewski:
@tuka217 Yes, please do generate a migration.

PR #3471, opened 2015-10-21
[USER] Change password validation

lchrusciel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | 
| License       | MIT

Password validation was done inside a UserController. It wasn't translatable and it was hard to receive error message when you work with API, because error message was displayed in a flash message. I removed this part form a controller and move this logic to custom validatior. 
I have also added two more parameters in a service.xml file to keep all class paths in a one place. It results probably in a BC break.


features/user/account/account_password.feature
```diff
@@ -20,7 +20,7 @@ Feature: User account password change
           And I fill in "Confirmation" with "newpassword"
           And I press "Save changes"
          Then I should still be on my account password page
-          And I should see "Wrong current password"
+          And I should see "Wrong current password."
```
stloyd: IIRC we agreed that matching texts should not contain dots etc.

@aRn0D @pjedrzejewski @umpirsky ping

src/Sylius/Bundle/UserBundle/Validator/Constraints/WrongCurrentPasswordValidator.php
```diff
@@ -0,0 +1,62 @@
...
+        $this->tokenStorage = $tokenStorage;
+        $this->encoderFactory = $encoderFactory;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function validate($password, Constraint $constraint)
+    {
+        $user = $this->tokenStorage->getToken()->getUser();
```
stloyd: Token can be empty which will give fatal error.

src/Sylius/Bundle/UserBundle/Resources/translations/validators.en.yml
```diff
@@ -12,6 +12,7 @@ sylius:
         plainPassword:
             not_blank: Please enter your password.
             mismatch: The entered passwords don't match
+            wrong_current: Wrong current password.
```
stloyd: `Given password is different than current`?

features/user/account/account_password.feature
umpirsky: Not sure, I prefer to have them. :)

stloyd: @umpirsky I'm talking about matching not about translations :)

pjedrzejewski:
http://symfony.com/doc/current/reference/constraints/UserPassword.html ?

lchrusciel:
Thanks @pjedrzejewski! I will use symfony constraint instead of custom one.

features/user/account/account_password.feature
umpirsky: @stloyd Me too. :)

src/Sylius/Bundle/UserBundle/Resources/config/services.xml
```diff
@@ -179,15 +183,20 @@
...
             <tag name="form.type" alias="sylius_gender" />
         </service>
 
         <!-- Validators -->
-        <service id="validator.unique.registered_user" class="Sylius\Bundle\UserBundle\Validator\Constraints\RegisteredUserValidator">
+        <service id="validator.unique.registered_user" class="%sylius.user.validator.unique_registered_user%">
             <argument type="service" id="sylius.repository.customer" />
             <tag name="validator.constraint_validator" alias="registered_user_validator" />
         </service>
+        <service id="validator.wrong.current_password" class="%sylius.user.validator.wrong_current_password%">
```
michalmarcinkowski: I don't think we need this sevice.

michalmarcinkowski:
Thanks Łukasz!

src/Sylius/Bundle/UserBundle/Resources/config/services.xml
```diff
@@ -179,12 +182,12 @@
         <service id="sylius.form.type.user_change_password" class="%sylius.form.type.user_change_password.class%">
             <tag name="form.type" alias="sylius_user_change_password" />
         </service>
-        <service id="sylius.form.type.gender" class="Sylius\Bundle\UserBundle\Form\Type\GenderType">
+        <service id="sylius.form.type.gender" class="%sylius.form.type.gender.class%">
```
stloyd: I don't think this is correct, we should follow Symfony changes & not use class names as parameters.

lchrusciel: You can always open a RFC about this. But for now, we have convention to use class names as parameters and I found it more appropriate to stay consistent then to change practise without discussion, just to follow some changes.

pjedrzejewski: Yes, if we decide to change it, we do it everywhere. :+1: for the RFC.

features/user/account/account_password.feature
```diff
@@ -20,7 +20,7 @@ Feature: User account password change
           And I fill in "Confirmation" with "newpassword"
           And I press "Save changes"
          Then I should still be on my account password page
-          And I should see "Wrong current password"
+          And I should see "Provided password is different then the current one"
```
bigfoot90: Provided password is different than the current one

PR #3474, opened 2015-10-21
Update PromotionEligibilityChecker to work with coupons

lbaey:
I think we need to keep the eligibility, so coupons can be combined with rules.

### DIFF
--- a/src/Sylius/Component/Promotion/Checker/PromotionEligibilityChecker.php
+++ b/src/Sylius/Component/Promotion/Checker/PromotionEligibilityChecker.php
@@ -87,7 +87,7 @@ class PromotionEligibilityChecker implements PromotionEligibilityCheckerInterfac
             return $eligible;
         }
 
-        return $this->areCouponsEligibleForPromotion($subject, $promotion);
+        return $eligible && $this->areCouponsEligibleForPromotion($subject, $promotion);
     }
 
     /**


tristanbes:
@lbaey 
So that means that we can create coupons,
And then only make them availble if only conditions are met inside rule's promotion with your PR, that's right ?

lbaey:
Yes, exactly.
I had this case : with a COUPON, 10% off on first order delivered in store. I set up a coupon based promotion with 2 rules and 1 action. Even if the rules condition were not met, the coupon was OK, and the action was applied. I didn't want this behaviour, and I think when you set up a promotion with rules and coupon you want all the conditions to be met.

PWalkow:
@lbaey Seems legit. Do you mind updating the tests for check this behaviour? You have changed the logic and no tests were changed. I wonder how the tests are passing now.

tristanbes:
so that's a great idea, i'm :+1:

AlucardleVash:
:+1:

tristanbes:
@PWalkow I guess tests are passing because there are no fixtures that adds a coupon conditions on a promotion rules. :)

pjedrzejewski:
Yes, let's add scenario for that. :+1:

lbaey:
I'm working on it :)
will be added to this PR

pjedrzejewski:
Ping @lbaey. :)

pjedrzejewski:
Friendly ping @lbaey. :)

michalmarcinkowski:
Fixed.

Issue #3467, opened 2015-10-21
Cannot overload LoadTaxonomiesData fixture


pfwd:
I get the following error when when overloading LoadTaxonomiesData fixture and running 
    
    php app/console sylius:install

    [Doctrine\DBAL\Exception\UniqueConstraintViolationException]                                                                                                                                                                               
      An exception occurred while executing 'INSERT INTO sylius_taxon_translation (name, slug, permalink, description, locale, translatable_id) VALUES (?, ?, ?, ?, ?, ?)' with params ["Category", "category", "category", null, "en_EN", 11]:  
      SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'en_EN-category' for key 'permalink_uidx'

This is because the Sylius taxonomies fixture is loaded before my child fixture. This is shown when running 

    php app/console doctrine:fixtures:load

    Careful, database will be purged. Do you want to continue Y/N ?y
    > purging database
    > loading [10] AppBundle\DataFixtures\ORM\LoadLocalesData
    > loading [10] AppBundle\DataFixtures\ORM\LoadPaymentMethodsData
    > loading [10] Sylius\Bundle\FixturesBundle\DataFixtures\ORM\LoadRbacData
    > loading [10] AppBundle\DataFixtures\ORM\LoadRbacData
    > loading [10] Sylius\Bundle\FixturesBundle\DataFixtures\ORM\LoadTaxonomiesData
    > loading [10] AppBundle\DataFixtures\ORM\LoadImagesData
    > loading [10] AppBundle\DataFixtures\ORM\LoadCountriesData
    > loading [10] AppBundle\DataFixtures\ORM\LoadCurrencyData
    > loading [10] AppBundle\DataFixtures\ORM\LoadGroupsData
    > loading [10] AppBundle\DataFixtures\ORM\LoadContactTopicData
    > loading [10] AppBundle\DataFixtures\ORM\LoadTaxonomiesData

This is how the overloaded fixture starts:

    namespace AppBundle\DataFixtures\ORM;

    use Doctrine\Common\Persistence\ObjectManager;
    use Sylius\Bundle\FixturesBundle\DataFixtures\ORM\LoadTaxonomiesData as BaseFixture;
    use Symfony\Component\Yaml\Parser;

    class LoadTaxonomiesData extends BaseFixture
    {

pjedrzejewski:
[...]

PR #3476, opened 2015-10-22
Sample data config

pfwd:
| Q | A |
| --- | --- |
| Bug fix? | [yes] |
| New feature? | [yes] |
| BC breaks? | [no] |
| Deprecations? | [no] |
| Fixed tickets | [ #3467 ] |
| License | MIT |
| Doc PR | [ #3467 ] |

Possible solution to #3467 
It allows a parameter called sylius.installer.fixtures.argument to be defined.  
If the parameter has been supplied and its value is not 'EDITME' then it is used as the value for the
 --fixtures argument.

If the parameters value is 'EDITME' then the process will continue as normal and the Sylius fixtures will be used.

This means that we can now define our own fixtures bundle
EG:
sylius.installer.fixtures.argument: src/AcmeFixturesBundle

I'm not sure if this follows the correct naming conventions etc but it is a simple solution to the issue


pjedrzejewski:
Could you please rebase this PR cause there are some invalid commits and it's hard to review. Thanks!

pfwd:
@pjedrzejewski On its way

pamil:
No longer applicable.

PR #3479, opened 2015-10-23
very  basic multi store

codebakery-ch:
I started working on a multi Store functionality.
The way I figured it works:

- A User with the role store_owner gets created
- A Store Owner creates a product, which then gets the ID of the store
- When a Costumer orders, and there are Products with store_ids in the cart, the Order gets split up to each Store Owner
- Same with payment and shipping
- If the Store Owner logs into the administration, he can only see and edit his own Orders, Products and Shipments.

That way, there is no need for a default store front. It all just gets split up at the checkout FInalize Step.

I just wanted to see it working as a concept, will integrate it better of course.


steffenbrem:
@codebakery-ch It looks a lot like you use a `Store` as what is a `Channel` currently in Sylius. Multiple channels could technically be used to have a multi store on one installation. The `Core` component has for example integrated channels with orders. What is in your opinion different in your approach now?

pjedrzejewski:
Hello @codebakery-ch, sorry for getting back to you so late. There is something happening around Sylius multi-tenancy/store but we need to wait until we know how it works out. That being said, this is definitely out of scope for Sylius 1.0. I appreciate sharing your code with us, hopefully it will give us some ideas for the future but for now - we won't support this out-of-box. If this becomes supported in any way then most likely it will be an extension bundle. :+1:

PR #3484, opened 2015-10-24
[User] Added missing unique indexes on customer.email and emailCanonical

peteward:
Import a million users to your database and you'll really need this index...

user.username and usernameCanonical seem to be nullable? so I haven't added any to this, although customer.email will be where db queries are mainly run against anyway.

### DIFF
--- a/app/migrations/Version20151024153713.php
+++ b/app/migrations/Version20151024153713.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Sylius\Migrations;
+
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
@peteward Could you generate a database migration for that change? Thanks!

peteward:
updated

pjedrzejewski:
Thank you very much Peter! :+1:

okwinza:
@awerffeli this will be fixed in #4009

_At least here i was faster than @pamil :D_

PR #3286, opened 2015-09-17
[Archetype] Removed fluent interfaces

1XWP:
| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | no
| License       | MIT
| Doc PR        | -

PR #3485, opened 2015-10-24
Added instance return on all setters

pfwd:
Added fluent setters as this is handy when creating productArchetypes

### DIFF
--- a/src/Sylius/Component/Archetype/Model/Archetype.php
+++ b/src/Sylius/Component/Archetype/Model/Archetype.php
@@ -28,6 +28,7 @@ use Sylius\Component\Variation\Model\OptionInterface;
  */
 class Archetype extends AbstractTranslatable implements ArchetypeInterface
 {
+
     /**
      * @var mixed
      */
@@ -105,6 +106,7 @@ class Archetype extends AbstractTranslatable implements ArchetypeInterface
     public function setCode($code)
     {
         $this->code = $code;
+        return $this;
     }
 
     /**
@@ -125,6 +127,7 @@ class Archetype extends AbstractTranslatable implements ArchetypeInterface
     public function setAttributes(Collection $attributes)
     {
         $this->attributes = $attributes;
+        return $this;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


TheMadeleine:
About a month ago we've been removing fluent interfaces from Sylius - https://github.com/Sylius/Sylius/pull/3286/files see here.

pfwd:
@TheMadeleine Oh that sucks. Are there any coding standards I could read/use in future?

TheMadeleine:
@pfwd The reason why fluent interfaces are evil is that when a method's name is ``setCode($code)`` it should not do anything else than setting a code. :)

pfwd:
@TheMadeleine  Agreed but it does help when a setter variable has a type hint. It's possibly something to look at again once PHP7 is well adopted as this will have scalar type hints.  But I get your point.

Should this be the case for all setters in Sylius?  Perhaps we should write this down in a guide somewhere

adamelso:
Worth noting that it's also cumbersome to test, especially with PhpSpec where you have to continuously mock and stub, like this:

```php
function it_builds_a_product_from_an_archetype(
    Archetype $shoe, Product $prada, Attribute $material, Attribute $brand
) {
    $shoe->getAttrbutes()->willReturn([$material, $brand]);

    $prada->setName('PRADA Court')->shouldBeCalled()->willReturn($prada);
    $prada->setCode('prada_court')->shouldBeCalled()->willReturn($prada);
    $prada->setDescription('You know, for walking.')->shouldBeCalled()->willReturn($prada);
    $prada->addAttribute($material)->shouldBeCalled()->willReturn($prada);
    $prada->addAttribute($brand)->shouldBeCalled()->willReturn($prada);
}
```

They are however useful in cases where a change in the object's state determines how the next chained method call behaves like with the Doctrine Query Builder.

Marco Pivetta from Doctrine [has more on the subject](http://ocramius.github.io/blog/fluent-interfaces-are-evil/).

Indeed, a coding standards guide would be a worthwhile addition to the docs.

PR #3491, opened 2015-10-26
Fixed waitFor function

### LINKED ISSUE #3490: Bug in behat WebContext
In methods waitForModalToAppear and waitForModalToDisappear callback return value is not correct: waitFor expects return result to be null or not null, but callback always return bool, and waitFor will always return on first iteration.

pamil:
Fixes #3490.


michalmarcinkowski:
:+1:

PR #3496, opened 2015-10-27
[USER] Reset password request fixes

lchrusciel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | 
| License       | MIT

Implementation of missing not registered validator. It is used when requesting reset password email. To do that user has to exist. Before, validation was done inside a controller. I moved this logic to custom validator.

Small bc break. Parameter for unique_registered_user didn't have *.class* sufix.

**Update:**
After reconsidering [this comment](https://github.com/Sylius/Sylius/pull/3496#issuecomment-151654275) I decided to remove not registered user validator. Instead, I change behaviour slightly to present same message regardless if user exists or not, and avoid null pointer exception.
In #3471 I changed change-password form validation, to check if user provided proper current password. But reset password endpoint used same form type. It results in broken reset-password logic. So i split this two form models.  We don't have tests for reseting password after email receiving. I will create an issue for that.


src/Sylius/Bundle/UserBundle/Validator/Constraints/NotRegisteredUser.php
```diff
@@ -0,0 +1,38 @@
...
+namespace Sylius\Bundle\UserBundle\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+/**
+ * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
+ */
+class NotRegisteredUser extends Constraint
+{
+    public $message = 'This email is not registered. Please register first.';
```
Arminek: What do you think about adding here translation key instead of hard coded message ?

src/Sylius/Bundle/UserBundle/Validator/Constraints/NotRegisteredUserValidator.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $this->userRepository = $userRepository;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function validate($email, Constraint $constraint)
+    {
+        $existingCustomer = $this->userRepository->findOneByEmail($email);
```
michalmarcinkowski: `$registeredUser`?

src/Sylius/Bundle/UserBundle/Validator/Constraints/NotRegisteredUser.php
michalmarcinkowski: :+1:

src/Sylius/Bundle/UserBundle/spec/Validator/Constraints/NotRegisteredUserValidatorSpec.php
```diff
@@ -0,0 +1,60 @@
...
+
+    function it_extends_symfony_constraint_validator()
+    {
+        $this->shouldHaveType('Symfony\Component\Validator\ConstraintValidator');
+    }
+
+    function it_add_violation_if_user_with_given_mail_does_not_exist($userRepository, $context, NotRegisteredUser $notRegisteredUserConstraint)
+    {
+        $userRepository->findOneByEmail('not_registered@user.com')->willReturn(null);
+        $notRegisteredUserConstraint->message = 'This email is not registered. Please register first.';
```
michalmarcinkowski: Redundant line

rpg600:
IMO providing this information is a security flaw even if it's useful for the end user.
It is safer to have something like : 
```If there is an account associated with xxx@xxx.xx you will receive an email with a link to reset your password.```

lchrusciel:
:+1: For new message. Tommorow I will change it.

pjedrzejewski:
Thank you Łukasz!

Issue #3240, opened 2015-09-08
[RFC] Integration tests for bundles


adamelso:
So we have web acceptance tests in Behat that cover the whole of Sylius as an application. Then there are the unit tests in PhpSpec that cover using the individual classes.

Since the bundles aren't being developed individually, perhaps a set of integration tests would help prepare the bundles towards being stable.

Ideas can be taken from the [SensioLabs PageObjects extension for Behat:](https://github.com/sensiolabs/BehatPageObjectExtension/blob/master/features/using_page_objects_with_a_factory.feature#L19) - concept is the same as Symfony where the main container is extended with additional services.

Very basic test case:

```feature
Given I have a file "app/AppKernel.php" that contains:
  """
<?php

class AppKernel extends Kernel
{
    public function registerBundles()
    {
        new Sylius\Bundle\ProductBundle\SyliusProductBundle(),
        new Sylius\Bundle\ResourceBundle\SyliusResourceBundle(),
    }
}
  """
```

.. then there are two ways of testing ...

As Behavior:

```feature
When I run the command "container:debug sylius.builder.product"
Then it should run successfully
And I should see "Sylius\Component\Product\Builder\ProductBuilder" as the class name
```

```php
function iRunTheCommand($cmd)
{
    $phpBin = (new PhpExecutableFinder())->find();
    $this->process = new Process($phpBin . "app/console " . $cmd);
}

function itShouldRun()
{
    $this->process->mustRun();
}

function iShouldSeeTheClassName($fqcn)
{
    expect($this->process->getOutput())->toContain($fqcn);
}
```

As Integration:

```feature
When I get the service "sylius.builder.product" from the Kernel container
Then I should get an instance of "Sylius\Component\Product\Builder\ProductBuilder"
```

```php
function iGetTheService($service)
{
    $this->app = new AppKernel('test', true);
    $this->service = $this->app->getContainer()->get($service);
}

function serviceIsInstanceOf($fqcn)
{
    expect(get_class($this->service))->toBe($fqcn);
}
```

Inspired by the [ReportBundle fix](https://github.com/Sylius/Sylius/pull/3231)

PWalkow:
Has this been forgotten? 
Is there already a way of adding integration tests in your mind @pjedrzejewski ?
These would be very useful in Sylius considering the size and complexity of some bundles and relations between these. 

[...]

PR #3497, opened 2015-10-28
[Core] Introduction of integration tests

PWalkow:
| Q             | A
| ------------- | ---
| Bug fix?      | [no]
| New feature?  | [yes?]
| BC breaks?    | [no]
| Deprecations? | [no]
| Fixed tickets | [~]
| License       | MIT
| Doc PR        | [~]

This is introduction of integration tests that are checking if services are existing and listeners are called on given events. It's just a start and further work on it is needed to help Sylius to be stable.

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -397,6 +397,13 @@ default:
             filters:
                 tags: "@security"
 
+        integration:
+            contexts:
+                - Behat\MinkExtension\Context\MinkContext
+                - Sylius\Bundle\CoreBundle\Behat\CoreIntegrationContext
+            filters:
+                tags: "@integration"
+
     extensions:
         Lakion\Behat\MinkDebugExtension:
             directory: travis/build-output
--- a/features/integration/services.feature
+++ b/features/integration/services.feature
@@ -1,0 +1,26 @@
+@integration
+Feature: Integration
+    In order to have stable application
+    As a developer
+    I need to be able to check existence of services
+
+    Scenario Outline: Existence of the Services in CoreBundle
+         When I get the service "<serviceName>" from the Kernel container
+         Then I should get an instance of "<serviceClass>"
+
+    Examples:
+    | serviceName            | serviceClass                                          |
+    | sylius.route_provider  | Sylius\Bundle\CoreBundle\Routing\RouteProvider        |
+    | sylius.route_generator | Sylius\Bundle\CoreBundle\Routing\SyliusAwareGenerator |
+#        ... to be continued
+
+    Scenario Outline: Existence of the dispatchers in CoreBundle
+        When I dispatch an event with "<eventName>"
+        Then I should see "<eventDispatcher>" with "<methodName>" called
+
+    Examples:
+    | eventName                           | eventDispatcher                                                | methodName           |
+    | sylius.cart.initialize              | Sylius\Bundle\CoreBundle\EventListener\OrderCurrencyListener   | processOrderCurrency |
+    | sylius.user.security.implicit_login | Sylius\Bundle\CoreBundle\EventListener\CartBlamerListener      | blame                |
+    | sylius.user.security.implicit_login | Sylius\Bundle\UserBundle\EventListener\UserLastLoginSubscriber | onImplicitLogin      |
+#        ... to be continued
--- a/src/Sylius/Bundle/CoreBundle/Behat/CoreIntegrationContext.php
+++ b/src/Sylius/Bundle/CoreBundle/Behat/CoreIntegrationContext.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\CoreBundle\Behat;
+
+use Sylius\Bundle\ResourceBundle\Behat\DefaultContext;
+
+class CoreIntegrationContext extends DefaultContext
+{
+    /**
+     * @var array[]
+     */
+    private $listeners;
+
+    /**
+     * @var mixed
+     */
+    private $service;
+
+    /**
+     * @Given /^I get the service "([^""]*)" from the Kernel container$/
+     */
+    function iGetTheService($service)
+    {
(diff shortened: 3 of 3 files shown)


pamil:
I'd like to have Sylius tested even more, but I really think that it isn't the best way. All these events and event listeners should be covered by existing Behat scenarios, that tell us about *behaviour*.

For example, testing that on `sylius.user.security.implicit_login` event calls `CartBlamerListener::blame()` does not tell us much, only that something was called (and needs customization after renaming event / listener class / listener method). Instead, we should write scenarios (I suppose we already have them) that checks the flow, e.g. if I'm not logged and I have cart, after I log in, the cart should be assigned to me (which can be seen somehow by user).

peteward:
@pamil  these are Integration Tests, quite different from what you are talking about - both should exist for coverage from different perspectives. Not all testing needs to be through BDD

michalmarcinkowski:
I'm not sure if it's the best way to use Behat. IMO for integration tests 'raw' PHPUnit would be better, but it will be a new convention, so it needs discussion @Sylius/core-team

peteward:
It's a great way of using Behat.   It's far better than using it to run lots of UI tests which take over an hour to run.

The important point in the features as defined here is who is the intended audience of that feature through the top level description `as a developer`.

We have rewritten all of our Behat tests, moved them away from being primarily UI tests to cover core business logic and sometimes integrations like this.  Our full test suite including Behat runs in under 10 mins and is core to everything we do - that's not possible with an hours worth if UI tests - it just isn't practical for any team of community. 

This is a shift in approach which was discussed by Konstantin himself at a Behat meet-up earlier this year.

peteward:
However, if the core team really doesn't like the approach of Behat for integration tests then we'd be very happy to look at alternatives.

@PWalkow is working on a refactoring of the Adjustment model and needs a solid set of Integration tests to really achieve his work due to the complexity of the event-driven model and it's not feasible to try to cover it all through Behat UI tests.

michalmarcinkowski:
@peteward I totally agree that Behat shouldn't be used only for UI tests and current tests in Sylius are far away from ideal. Sooner or later (I hope sooner) we have to have a discussion about how we should approach testing and how to rework the existing tests, because they are now painfull to both maintain and work with. I would be happy to hear more about your approach and thoughts on that topic :)

PWalkow:
As @adamelso tried to explain here: https://github.com/Sylius/Sylius/issues/3240 , integration tests are very important for the stability of any app. I followed his approach to test it in the tool that Sylius is currently using. My personal thought would be using just simple webTestCase with bunch of useful methods rather than using Behat for that but I see the point for Behat if it's the main tool here.

@adamelso had insights about changing kernel and/or different bundles added to the app but I had no time to research that solution yet.

gonzalovilaseca:
Behat is not just a BDD tool, it can also be used as a testing tool. 

You can do Behat tests without UI and that's what my idea of an integration test is: test the core feature without relying on UI, and if a service is not present, or an event is not triggered, then the feature test should fail. I really don't care how that is implemented, I only want to be sure that when `A` happens I get `B` as a result.

The proposed testing would be a pain to maintain IMO.

adamelso:
@PWalkow  Good start, but it's not quite what I had I mind. This somewhat tests what has already been tested by the acceptance tests. What is missing is how the bundle works with other Symfony applications that are not Sylius.

That is, can I successfully register a new Bundle in the AppKernel without an exception being thrown? Or is there a service missing from that bundle that possibly exists in another bundle? And how does the bundle work when I change the config.

umpirsky:
On the one hand, this kind of tests are missing, and I think we should have them. I was also thinking about covering some API emthods, API OAuth authentication etc..

On the other hand testing if some service ID correspondents to some object instance of certain type or some event correspondents to some callback smells like copy-paste and testing container or event dispatcher, rather then Sylius.

So I think they should be more advanced and make more sense. 

Like for example, get product builder from container, create some product and make assertions on the result returned. 

While Behat can be used to do that, I don't think it is the best tool for the job. Because of Gherkin, it is not meant to be used for this kind of tests. I would prefer Codeception or PHPUnit for this kind of task.

My 2 cents.

pjedrzejewski:
I will try to give you more thoughts on this in the coming days, but I generally agree with everything @umpirsky said, except Codeception. :D I think we should use PHPUnit and test stuff working together (like taxes, shipping charges on order etc.), not service definitions. Behat could be used for that as well, simply skip the UI part. We could rework existing scenarios (for taxes for example) to skip the ui in scenarios themselves.

PWalkow:
@pjedrzejewski I'd rather move these to PHPUnit with the use of the WebTestCases but unfortunately we are not running any phpUnit tests for now by default. As I mentioned here https://github.com/Sylius/Sylius/issues/3240 there are some ApiTestCases that are not running. And will fail if you would try to run these.

PWalkow:
Does someone started to work with this ApiTestCases that are not running on travis? (And failing if tried to run)
@umpirsky @pjedrzejewski @michalmarcinkowski @pamil @adamelso 

I'd like to start PHPUnit integration tests again but I wonder if someone moved it a little forward already.

michalmarcinkowski:
@PWalkow we have done some work on that already in Lakion. Hope to share it in the comming days.

umpirsky:
@michalmarcinkowski I would like to review. :)

PWalkow:
@michalmarcinkowski any progress on that? if not I might move this one forward somehow. Maybe fixing these ApiTestCases would be good at start?

michalmarcinkowski:
@PWalkow we would like to extract ApiTestCase (as a standalone component/library/bundle(?) - not decided yet), because it's independent from Sylius and can be used by various applications. 

At Lakion we expanded and enhanced the ApiTestCase class and I opened a PR for initial review (https://github.com/Lakion/ApiTestCase/pull/1 - @umpirsky looking forward to your comments) with this class extracted for workshop purposes. There is more, but we need to find time to extract it :)

We would like to have more or less similar tests to these https://github.com/michalmarcinkowski/php-summercamp-2015-rapid-api/tree/master/src/AppBundle/Tests. Any suggestions are welcome :wink:

pjedrzejewski:
Closing this, because using Behat for this kind of tests does not make much sense. We should consider another approach for testing components in integration, I will try to propose an alternative soon.

Issue #2604, opened 2015-03-19
Failure during installation process


kenkataiwa:
There is a problem with the installation process it fails here:

Fatal error: Uncaught exception 
'Symfony\Component\Filesystem\Exception\IOException' with message 'Cannot rename "<path\to\>\sylius\app\cache\de_" to "<path\to\>\sylius\app\cache\dev".' in <path\to\>\sylius\vendor\symfony\symfony\src\Symfony\Component\Filesystem\Filesystem.php:262

Even when I do: php app/console cache:clear I get the same "Cannot rename" error

pjedrzejewski:
@kenkataiwa This is something wrong with your filesystem permissions. http://symfony.com/doc/current/book/installation.html#book-installation-permissions

kenkataiwa:
@pjedrzejewski it's not a permissions issue because I'm on a Windows system and have full access to the account I'm using so I am able to write to that directory. It might be something do with Symfony caching in general, or on Sylius maybe. But I think it's an issue that's why I raised it here.

Because if I do "php app/console cache:clear" it fails but when I do: "php app/console cache:clear --no-warmup" it works.

pjedrzejewski:
You are right, I recall this issue... I remember someone had it already with older version of Sylius. What version are you using?

kenkataiwa:
It's the very latest because I did "composer create-project sylius/sylius" today. 
I've looked at the composer.json it says: 0.14-dev
![capture](https://cloud.githubusercontent.com/assets/2436107/6734940/13ce645c-ce6c-11e4-82a2-555b3e277fd7.PNG)

AlexisBarat:
Hi, i've the same issue here, using 0.14.*@dev.
Anyone found the solution ?

hantsy:
I have encountered the same problem in 0.14-dev.

hantsy:
$ php app/console sylius:install
Installing Sylius...

Step 1 of 4. Checking system requirements.
+------------------------------------+------------------------------------+
| Issue                              | Recommendation                     |
+------------------------------------+------------------------------------+
| sylius.settings.short_open_tag     |                                    |
| sylius.settings.session.auto_start |                                    |
| sylius.extensions.accelerator      | sylius.extensions.accelerator.help |
+------------------------------------+------------------------------------+
Success! Your system can run Sylius properly.

Step 2 of 4. Setting up the database.
Creating Sylius database for environment dev.



[...]

PR #3501, opened 2015-10-29
CI AppVeyor

mhujer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none
| License       | MIT
| Doc PR        | none

This PR adds a config file for AppVeyor which is a CI similar to travis, except they run the builds on Windows machines. Therefore any possible incompatibilites can be easily discovered (ref #2604)

~~**It is still very WIP, I just want to:
1) prevent possibly duplicate work, if someone starts working on the same thing
2) get feedback as early as possible**~~

You can see the build output here https://ci.appveyor.com/project/mhujer/sylius

To integrate it with this project, it is necessary to login to https://ci.appveyor.com/ using github account and enable the build for this repo.

(Symfony has recently started to use AppVeyor - https://github.com/symfony/symfony/pull/15575, that's where most of the config comes from)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,63 @@
+build: off
+cache:
+    - c:\php -> appveyor.yml
+    - '%LOCALAPPDATA%\Composer\files -> appveyor.yml'
+clone_folder: c:\projects\sylius
+
+environment:
+  fast_finish: true
+  matrix:
+    - TEST_SUITE: phpspec
+    - TEST_SUITE: 1a
+    - TEST_SUITE: 1b
+    - TEST_SUITE: 2a
+    - TEST_SUITE: 2b
+
+services:
+  - mysql
+
+init:
+    - SET PATH=c:\php;%PATH%
+    - SET COMPOSER_NO_INTERACTION=1
+    - SET PHP=1
+    - SET ANSICON=121x90 (121x90)
+
+install:
+    - IF EXIST c:\php (SET PHP=0) ELSE (mkdir c:\php)
+    - IF %PHP%==1 cd c:\php
+    - IF %PHP%==1 appveyor DownloadFile http://windows.php.net/downloads/releases/php-5.6.15-Win32-VC11-x86.zip
+    - IF %PHP%==1 7z x php-5.6.15-Win32-VC11-x86.zip -y >nul
+    - IF %PHP%==1 echo [PHP] >> php.ini
+    - IF %PHP%==1 echo max_execution_time=3600 >> php.ini
+    - IF %PHP%==1 echo memory_limit=-1 >> php.ini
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -20,6 +20,8 @@
     ],
     "require": {
         "php":                                  ">=5.5.9",
+        "ext-exif":                             "*",
```
mhujer: Exif extension is not required by any package, but if it is missing, it causes error during `cache:warmup`

```
php app\console cache:warmup --env=test

Warming up the cache for the test environment with debug true
                                                                
  [Imagine\Exception\NotSupportedException]                     
  PHP exif extension is required to use the ExifMetadataReader  
                                                                
 
 ```

**Comment by pjedrzejewski** [comment]
This is awesome, thank you for your work Martin! :+1: I hope this is the first step in much better Windows support.

**Comment by adamelso** [comment]
aww beat me to it https://github.com/adamelso/Sylius/pull/6

but nice! :+1:  I'll let you continue

**Comment by pjedrzejewski** [comment]
@adamelso I like to have such problems in the community! :) Let's combine your work and get the best possible configuration. Would be great if you guys discuss your approaches.

File: appveyor.yml [hunk 2]
```diff
@@ -0,0 +1,63 @@
...
+    - php composer.phar install --prefer-dist --no-interaction --no-progress --ansi
+
+test_script:
+    - php app\console doctrine:database:create --env=test
+    - php app\console doctrine:schema:create --env=test
+    - php app\console doctrine:phpcr:repository:init --env=test
+    - php app\console cache:warmup --env=test
+
+    - IF %TEST_SUITE%==phpspec travis\appveyor\suite-phpspec.cmd
+    - IF %TEST_SUITE%==1a travis\appveyor\suite-1a.cmd
```
mhujer: I needed to split the existing suites into two parts to fit into AppVeyor build time limit

travis/appveyor/suite-phpspec.cmd
```diff
@@ -0,0 +1,3 @@
+call php app/console sylius:install --no-interaction -vvv
```
mhujer: This should verify, that Sylius is actually able to install itself

pjedrzejewski:
Great work @mhujer! What configuration should I setup in AppVeyor? I have registered and added Sylius project. I merge this to see if it works, but I guess I will need to do a bit more setup. I really appreciate the effort you have put into this, it will make our life so much easier and hopefully we see less bug reports in Windows. :) :+1:

mhujer:
@pjedrzejewski There shouldn't be any setup necesarry in the AppVeyor - you can either set it through the web UI or through the config file.

mhujer:
The only downside of the AppVeyor is that the build is quite slow :snail:

PR #3502, opened 2015-10-29
Configure adjustments api and permissions

pjedrzejewski:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Sylius/Bundle/OrderBundle/Controller/AdjustmentController.php
+++ b/src/Sylius/Bundle/OrderBundle/Controller/AdjustmentController.php
@@ -14,6 +14,7 @@ namespace Sylius\Bundle\OrderBundle\Controller;
 use Sylius\Bundle\ResourceBundle\Controller\ResourceController;
 use Sylius\Component\Order\Model\AdjustmentInterface;
 use Sylius\Component\Order\Model\OrderInterface;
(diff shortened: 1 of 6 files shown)


src/Sylius/Bundle/OrderBundle/Controller/AdjustmentController.php
```diff
@@ -53,4 +53,33 @@ public function lockAction(Request $request, $id)
...
         return $this->redirectHandler->redirectTo($order);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createNew()
+    {
+        if (null === $orderId = $this->getRequest()->get('orderId')) {
+            throw new NotFoundHttpException('No order id given.');
```
PWalkow: just for consistency, isn't it better to use 'createNotFoundException' method? In that case we don't need to have 'NotFoundHttpException' here.

michalmarcinkowski: Actually this should throw 400 bad request.

:+1: for createNotFound in condition below

michalmarcinkowski:
Thanks Paweł! :+1:

PR #3504, opened 2015-10-30
[User] Decoupled password encoders and updaters from UserInterface

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Introduced `CredentialingInterface` which implementation needs to be passed to password encoders and updaters instead of `UserInterface` with methods like `setCustomer()` etc.

### DIFF
--- a/src/Sylius/Bundle/UserBundle/Security/UserPasswordEncoder.php
+++ b/src/Sylius/Bundle/UserBundle/Security/UserPasswordEncoder.php
@@ -11,6 +11,7 @@
 
 namespace Sylius\Bundle\UserBundle\Security;
 
+use Sylius\Component\User\Model\CredentialingInterface;
 use Sylius\Component\User\Model\UserInterface;
 use Sylius\Component\User\Security\UserPasswordEncoderInterface;
 use Symfony\Component\Security\Core\Encoder\EncoderFactoryInterface;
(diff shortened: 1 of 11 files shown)


pjedrzejewski:
Is this name a result of fun discussion in the office? :D

TheMadeleine:
@pjedrzejewski, flawless observation. :+1:

pjedrzejewski:
@TheMadeleine @pamil Are we sure it makes sense? Credentialing means something a bit different as far as I know/googled. Perhaps CredentialsHolderInterface or let's go full-Javan SecurityCredentialsHolderInterface?

stloyd:
IMO this interface is useles... What's the real point for it? Are so commonly re-use this interface somehow?

pamil:
@stloyd right now I have to implements UserBundle's `UserInterface` which ships with `Customer` relation, `OAuthAccount`s relation and all the stuff I don't really need in my project. This interface will let me use UserBundle's `PasswordUpdater` and `PasswordEncoder`.

pamil:
@pjedrzejewski `CredentialsHolderInterface` looks promising and way more understandable, thanks! :)

pjedrzejewski:
Thank you Kamil!

PR #3507, opened 2015-10-30
[Addressing] Country fixes

peteward:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        | 

Provinces for a country ordered by name by default
Ensure __toString() won't break on Country. I may have an old DB, but some countries in my db couldn't retrieve a translation so blew up when used in a country dropdown.

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Resources/config/doctrine/model/Country.orm.xml
+++ b/src/Sylius/Bundle/AddressingBundle/Resources/config/doctrine/model/Country.orm.xml
@@ -30,6 +30,9 @@
             <cascade>
                 <cascade-all/>
             </cascade>
+            <order-by>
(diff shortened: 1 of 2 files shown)


pamil:
I'm just curious, what countries codes can't be translated to localised country name by `Intl`?

peteward:
('BV', 'QU', 'HM', 'AN', 'QO', 'CS')

They were in my Sylius DB from a long time ago

pjedrzejewski:
Interesting, we should investigate this more and provide better output if possible, but this is definitely a needed fix, thank you Peter! :+1:

PR #3511, opened 2015-10-31
[UserBundle] Fix customer gender layout on backend admin form

peteward:
From:

screen shot 2015-10-31 at 09 00 36

To:

screen shot 2015-10-31 at 09 01 18

NOTE: There's an underlying theme bug here which doesn't render 'default' labels properly. Ideally you shouldn't have to do this to make your widget render the same as every other, but it's still worth merging to make it consistent.

I will open a separate issue for this one, but don't have time to fix it right now.


pjedrzejewski:
@peteward Could you have a look at the failing build? (suite 1)

peteward:
fixed

pjedrzejewski:
Thank you Peter! :)

PR #3518, opened 2015-11-02
Fix filter form

rpg600:


src/Sylius/Bundle/WebBundle/Controller/Backend/FormController.php
```diff
@@ -45,8 +45,14 @@ public function showAction($type, $template)
      */
     public function filterAction($type, $template)
     {
+        $request = $this->get('request_stack')->getMasterRequest();
```
stloyd: :-1: 

This should be put in:
```php
public function filterAction(Request $request, $type, $template) {}

rpg600: No, because this controller is called inside a sub request and we need the master request.

pjedrzejewski:
We will have a better solution for filters soon, but this is a good fix, thank you very much Reinis! :+1:

PR #3525, opened 2015-11-04
Add required PHP extensions to composer.json

mhujer:
Some extensions (so far I've discoverd: `exif`, `fileinfo`, `gd`, `intl`) are necessary for tests to run, so they should be listed in the `require-dev` part of the `composer.json`

They gave me hard time during #3501 because every build takes a few minutes because of the Composer dependencies :clock1:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
     "require": {
         "php":                                  ">=5.5.9",
         "ext-exif":                             "*",
+        "ext-gd":                               "*",
         "a2lix/translation-form-bundle":        "~2.0",
         "doctrine/collections":                 "~1.2",
         "doctrine/common":                      "~2.5",
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
     ],
     "require": {
         "php":                                  ">=5.5.9",
+        "ext-exif":                             "*",
         "a2lix/translation-form-bundle":        "~2.0",
         "doctrine/collections":                 "~1.2",
         "doctrine/common":                      "~2.5",
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
     "require": {
         "php":                                  ">=5.5.9",
         "ext-exif":                             "*",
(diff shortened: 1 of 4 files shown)


patkar:
Are you sure, they are only needed in tests an not in general?

mhujer:
@patkar I can't say for sure. They may be necessary only for specific modules or so.

pjedrzejewski:
@Sylius/core-team

michalmarcinkowski:
Sylius will not run properly without `ext-intl`, so it should be in general require. Not sure about the rest...

composer.json
```diff
@@ -75,6 +75,10 @@
         "winzou/state-machine-bundle":          "~0.2"
     },
     "require-dev": {
+        "ext-exif":                             "*",
+        "ext-fileinfo":                         "*",
+        "ext-gd":                               "*",
+        "ext-intl":                              "*",
```
umpirsky: Indention. :tongue:

umpirsky:
I think this should match `Sylius\Bundle\InstallerBundle\Requirement\ExtensionsRequirements`.

mhujer:
Thanks for the feedback, I'll update the PR later today.

composer.lock
```diff
@@ -3264,61 +3265,6 @@
             "time": "2015-08-27 11:27:33"
         },
         {
-            "name": "mathiasverraes/money",
```
mhujer: I used only commands such as `composer require ext-fileinfo:*` so the unrelated changes in `composer.lock` are probably result of it being out of sync with `composer.lock`.

mhujer:
I've just updated the PR (the commits are completely new).

I disabled the extensions and tried to install Sylius with them disabled. Then, when I got an error message, I marked the missing extension as required, enabled it and tried again. See the commit messages for more details why each extension is necessary.

michalmarcinkowski:
Good job! :+1:

pjedrzejewski:
Great work Martin, thank you very much! :)

PR #3528, opened 2015-11-04
[Resource] EntityHiddenType fix

lchrusciel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT

#3406 PR fixed problem with reversed logic of `ObjectToIdentifierTransformer`. But unfortunately we forgot that *EntityHiddenType* use this transformer. I changed the name of method used by *EntityHiddenType* to avoid problems like this:
`Expected argument of type "AppBundle\Entity\ProductVariant", "string" given`


michalmarcinkowski:
We shouldn't use the transformer this way, but it's a blocking issue now, so ok for the fix :+1:

rpg600:
Actually these calls are not needed anymore, it's a dead code, see #3463

PR #3529, opened 2015-11-04
[Flow] Allow ProcessValidator to access ProcessContext

umpirsky:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

This can be useful when you want to implement some advanced process validation that need to be aware of the some process context data, such as current step.


pjedrzejewski:
Sounds like a good idea, thanks Sasha!

PR #2617, opened 2015-03-23
Adding default outgoing ip address for dev (default vm outgoing network)

vihuvac:

PR #3539, opened 2015-11-06
Allow vagrant users to access app_dev.php from host IP

stefandoorn:
After using the supplied vagrant box and trying to access 'http://sylius.dev/app_dev.php' I got the error that I was not allowed to access that file. This change adds the host IP to allow for that on vagrant usage.

### DIFF
--- a/web/app_dev.php
+++ b/web/app_dev.php
@@ -19,7 +19,7 @@ use Symfony\Component\HttpFoundation\Request;
 
 if (isset($_SERVER['HTTP_CLIENT_IP'])
     || isset($_SERVER['HTTP_X_FORWARDED_FOR'])
-    || !(in_array(@$_SERVER['REMOTE_ADDR'], array('127.0.0.1', 'fe80::1', '::1', '113.0.0.1'))
+    || !(in_array(@$_SERVER['REMOTE_ADDR'], array('127.0.0.1', 'fe80::1', '::1', '113.0.0.1', '10.0.0.1'))
         || php_sapi_name() === 'cli-server')
 ) {
     header('HTTP/1.0 403 Forbidden');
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
Sounds good to me, let's hear more thoughts. :)

PWalkow:
:+1: I have enough of commenting the app_dev file every time I change branches.

mhujer:
I just tried to run the Vagrant but as it is overriding the networks settings its IP 10.0.0.200 colides with my wifi network and the Vagrant does not start:

```
The specified host network collides with a non-hostonly network!
This will cause your specified IP to be inaccessible. Please change 
the IP or name of your host only network so that it no longer 
matches that of a bridged or non-hostonly network.
```

When I changed it to 10.1.0.200, it started properly, but my local address was different, so I still needed to change it in the `app_dev.php`.

If you look at a list of [common wifi router IP ranges](http://www.techspot.com/guides/287-default-router-ip-addresses/), it could probably be resolved by changing the default IP 10.11.0.200 which should not colide with those ranges.

stefandoorn:
It could be an option to change the IP range for the vagrant box + the whitelisted IP to e.g.: 192.168.133.x. Its still a non-outside localhost environments IP range and not used in standard router setups. I've never seen anyone using it either anyway. 

What do you think?

pjedrzejewski:
I recall we had a discussion about using Vagrant IP, but can't find the issue. Would be good to have a look at it before merging, so if anyone has a reference that would be great. I remember it was related to security concerns.

stefandoorn:
@pjedrzejewski Probably this thread: https://github.com/Sylius/Sylius/pull/2617. Might be better not to take this change based on arguments in that pull request.

pjedrzejewski:
Yes, thanks for finding this thread. I agree with the idea of customizing this per project - it is safer. Thanks Stefan!

PR #3548, opened 2015-11-07
[ResourceBundle] Search id or slug in request query strings

SebLours:
A POST, PUT or PATCH request like this:
```
{
    "country_code": "FR",
    "first_name": "First name",
    "last_name": "Last name",
    "slug": "my-slug"
}
```
throws a 404 response because the slug became a default search criteria.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
@@ -438,9 +438,9 @@ class ResourceController extends FOSRestController
      */
     public function findOr404(Request $request, array $criteria = array())
     {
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
The thing is, we may want display something by slug, which is in the request attributes. (eg. /foo/{slug})

This change works for Sylius, but in the case I mentioned above with custom resources it can be a problem. Perhaps we should look in the query AND attributes bag? What do you think? That would skip the "request" bag because it does not make sense to look there anyway.

SebLours:
Hi Pawel,

Of course, you're right.
Now request's attributes & query are tested.

pjedrzejewski:
Perfect, thank you very much Sébastien!

PR #3550, opened 2015-11-08
configure header comment fixer for phpcs

Ma27:


pjedrzejewski:
Thank you Maximilian! :)

PR #3553, opened 2015-11-09
Sylius cs fix fixes

PWalkow:
I wonder why we are having these kind of tools in the repo and these are not used. I guess it might be good to add it as a requirement to pass the tests to fix this kind of 'problem'.

Added again without PHPSpecs.

The new added 'fix' for php-cs-fix broke it:
https://github.com/Sylius/Sylius/pull/3550

This one is also fixing it.


okwinza:
if it is just about CS issues then you might consider using "[ci skip]" in your commit messages so travis wont be bothered building it.

pamil:
@okwinza IMHO there are too many changes to omit Travis, you can't check it manually and CS fixer can be erroneous too :)

PWalkow:
There is no rush with it, we can wait for travis. Cs-fixer also fixed some class naming so my initial cs-fix broke one of the specs... that was not used at all.

okwinza:
@pamil @PWalkow 
ok i see the point now :) :+1:

pjedrzejewski:
@PWalkow The only reason why I am hesitant to merge this now is that there is plenty of quite big PR in queue and that will make rebases a bit harder. Perhaps we could deal with the biggest PRs and then run cs fixer. After that I'd setup something like fabbot.io to prevent any CS violations getting into the repo. What do you think?

PWalkow:
@pjedrzejewski It's fine for me and I do understand that it may be very beneficial to reviews rather than pointing 'new-lines' etc. Automatic checker/fixer would be a good idea.

Anyway, I've added this fix as well to this branch:
https://github.com/PWalkow/Sylius/commit/e3c6ae6ffb14e03fed0ddc4f18f6dd0de4808de3
which is needed to actually run the sylius-cs-fixer, would you like me to cherry-pick it and add as separate PR?

pjedrzejewski:
@PWalkow Yes, please do! Thanks. :)

PWalkow:
@pjedrzejewski fooor you! https://github.com/Sylius/Sylius/pull/3555/files

PR #3554, opened 2015-11-09
Updated PhpSpec to 2.4@alpha, native PHP7 support

pamil:
PhpSpec does not need typehints transformer no more.
Updated `composer.lock`, fixed specifications due to Symfony's `Command` class changes.


pjedrzejewski:
Nice, thanks Kamil!

PR #3558, opened 2015-11-09
fixed puppet manifest

### LINKED ISSUE #2311: First vagrant up command ends with non 0 exit status due to PHPUnit deprecation via pear.
As PHPUnit installation is now deprecated via Pear (cf. https://github.com/sebastianbergmann/phpunit/wiki/End-of-Life-for-PEAR-Installation-Method ) we got an error when executing vagrant up command.

```
==> default: err: /Stage[main]//Php::Pear::Module[PHPUnit]/Exec[pear-PHPUnit]/returns: change from notrun to 0 failed: pear -d preferred_state=stable install  pear.phpunit.de/PHPUnit returned 1 instead of one of [0] at /etc/puppet/modules/php/manifests/pear/mod
ule.pp:115
```

Installation process continues but at the end we got 
```
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
```

We should now use the composer method installation (It seems it's already added in the composer.json) file. 
Maybe we just need to remove the phpunit from the puppet file in manifest folder.

Thanks!

### LINKED ISSUE #3544: Vagrant installs incompatible (old) version of PHP
It installs PHP 5.4.45, but the minimal version for Sylius is 5.5.9, therefore it is not possible to install Sylius in Vagrant.

Ma27:
- install php 5.5 on vagrant environment
- don't install phpunit from pear channel (phpunit is installed through composer)
- resolves #3544
- resolves #2311

### DIFF
--- a/vagrant/manifests/default.pp
+++ b/vagrant/manifests/default.pp
@@ -33,9 +33,12 @@ Class['::apt::update'] -> Package <|
 and title != 'software-properties-common'
 |>
 
-    apt::key { '4F4EA0AAE5267A6C': }
+apt::key { '4F4EA0AAE5267A6C': }
 
-apt::ppa { 'ppa:ondrej/php5-oldstable':
+# when bumping the php version again, please
+# have a look at the PPA documentary in order to provide a proper PPA alias:
+# https://launchpad.net/~ondrej/+archive/ubuntu/php5
+apt::ppa { 'ppa:ondrej/php5':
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
I restarted the build, good to go when it is green. :)

pjedrzejewski:
Thank you Maximilian!

PR #3531, opened 2015-11-05
[Fixtures] Load fakers with the available locales

skigun:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3530 
| License       | MIT
| Doc PR        | -

PR #3560, opened 2015-11-10
[FixturesBundle] Reverted loading fakers with the available locales

GSadee:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

I have discussed this problem with @pjedrzejewski and for now, I have reverted changes from #3531.

In future, we should consider a better solution for loading fakers locales in fixtures.


pjedrzejewski:
Thanks Grzesiu!

PR #3566, opened 2015-11-12
[WIP] [RFC] Ability not to show the product price with the option "for logged user"

thiphamyp:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | no |
| License | MIT |
| Doc PR | no |

This PR add the feature to choose not to display the price of product on the frontend. Similar to a « catalogue »  feature

What do you think of not showing the price if the option is inactive

​*_This pull request has been sponsored by Store Factory_​*


pjedrzejewski:
I am :+1:, just wondering which is better -> display price or hide price option? Thoughts?

Thank you very much Hang for the patch!

okwinza:
:+1: Since the price is displayed by default IMO "Hide Price" checkbox is better.

tristanbes:
We need your point of view on this feature.

- [ ] Hide price
    - [ ] Except if user is logged in
- [ ] Can't add to cart
    - [ ] Except if user is logged in

1/ Is this flexibility wanted, or we can reduce to 2 checkbox and write that it does not concerns the logged in user ? 
 
2/ Should those options live under 
1. product ? 
2. categories ? 
3. Global options ? 
4.All of the option above ?

@pjedrzejewski @okwinza

pjedrzejewski:
I think we would like to have this flexibility. Only problem I see, that there are tons of other small options like this and I don't want our Product entity become too big. Any ideas, how we could avoid that? If we don't figure this out now and keep adding this kind of options (which are great) we will have a mess in the product model.

Issue #3535, opened 2015-11-05
Cart Events complexity and structure


PWalkow:
My plan is to unify first three events into one that will call each of these. In my opinion it will unify the way how we deal and save the cart/order. This is the current structure:

![screen shot 2015-11-05 at 12 39 35](https://cloud.githubusercontent.com/assets/1130582/10970894/4442a8d8-83c9-11e5-817f-5bcb524d29aa.png)

@pjedrzejewski Do you think it could be somehow easier? Please let me know about your issues with these events, if you have any.

pjedrzejewski:
Definitely something to look at. I was thinking about making cart provider stateless, by introducing some CartOwner/HolderInterface, that could be passed to it and return the cart based on that - would allow getting the cart for any Customer in the shop. That being said, we still need to handle not logged in users - keep the id in session. Do you have any ideas? Regarding events - Yes, I'd actually like to remove the cart initialize event and introduce another way to handle things like:

* setting current channel on cart
* setting current locale on cart
* setting current currency
* setting current customer on cart

We need all this to happen when cart is created, but also when any of these changes. (switching locale etc.)

PWalkow:
![22acbbcc-96a8-487f-bb41-2301a63480e7](https://cloud.githubusercontent.com/assets/1130582/11145330/48337042-89fc-11e5-8b71-2476c3771fb2.jpg)
We worked hard in the office and we are considering changing current event model with similar to this one on the photo. Probably middle layer of events would not be added in the first stage, it might be too complex to understand for [new] developers.

PWalkow:
@pjedrzejewski Any ideas how to get rid of cart in CartProvider? I doubt that having cart in invalid state is a good idea at all. When I'm remodelling these events I end up with lots of failing tests due to  invalid cart state, e.g. cart without channel etc..

PR #3575, opened 2015-11-17
[WIP] Remodel cart events

PWalkow:
Issue:
https://github.com/Sylius/Sylius/issues/3535


pjedrzejewski:
@PWalkow I will do my best to give you some feedback ASAP, because we were discussing this recently too and definitely need to change all these events.

PR #3576, opened 2015-11-17
[Behat] [Travis] ~5 times faster tests suites

### LINKED ISSUE #1363: Doctrine metadata cache bug related to Product & Sluggable
Hi all

In our site, sluggable in `Product` is malfunctioning when doctrine metadata cache is on (I was testing with APC). This cause `null` product slug when creating new product.

I have posted an issue in Doctrine: https://github.com/doctrine/common/issues/320

Kindly let me know if you know a work around for this problem, I think this might be related to how we extend product class.

pamil:
Makes executing whole test suite on Travis approximately 5 times faster (30min compared to 2,5h earlier). It also makes our tests running in environment more similar to production one, which results in revealing bugs that does not happen with `kernel.debug` set to `true` or without cache (like #1363 and #3578).

 - [x] Fixed #1363 (rather naive implementation, will be rewritten in the near future to provide universal and elegant way to handle sluggable objects, instead of Gedmo)
 - [x] PHP7 support (two issues remaining, Travis PHP7 builds fails, may be connected with development version)
 - [x] Create two test environments, the first similar to previous one (like dev), the second with all that caching stuff (like prod, used for CI)


.travis.yml
```diff
@@ -54,6 +58,7 @@ before_install:
     - if [ $TEST_SUITE = "javascript" ]; then travis/prepare/prepare-javascript; fi
 
 before_script:
+
```
pjedrzejewski: Why? :P

src/Sylius/Bundle/CoreBundle/Behat/CoreContext.php
```diff
@@ -645,12 +647,20 @@ private function createShipment($string)
...
+
+        $session = $this->getService('session');
+        $session->set('_security_user', serialize($token));
+        $session->save();
+
+        if ($this->getSession()->getDriver() instanceof Selenium2Driver) {
+            $this->visitPath('/');
+        }
+
+        $this->getSession()->setCookie($session->getName(), $session->getId());
```
pjedrzejewski: Nice trick. :)

tristanbes:
:+1: great improvement

PWalkow:
:+1: Great one, I wish I would have this one already in my branch...

pjedrzejewski:
@PWalkow This should be merged quite soon (I think today). We will hold off with tweets and stuff until we have a blog post ready, probably for tomorrow. :)

composer.json
```diff
@@ -33,7 +33,7 @@
         "doctrine/doctrine-fixtures-bundle": "~2.2",
         "doctrine/doctrine-migrations-bundle": "~1.0",
         "doctrine/orm": ">=2.4.8,<2.5",
-        "doctrine/phpcr-odm": "^1.2.2",
+        "doctrine/phpcr-odm": "^1.3@rc",
```
pamil: For PHP7 support - `<1.3.0` contains documents annotated with `@String` which we don't really use. `cache:warmup` generates proxies for all Doctrine entities and documents if `doctrine.orm.auto_generate_proxy_classes` is `false`. Sylius sets it to `%kernel.debug%`.

When `kernel.debug` is:
 - `true`: proxies are lazy-generated, only when needed (`@String` annotation is never parsed)
 - `false`: proxies are generated only once at warmup

okwinza:
:+1: outstanding work @pamil !

pjedrzejewski:
Boom! Please wait with celebrations, we will announce this improvement with a proper blog post explaining what happened. :) Enjoy faster builds and thank you @pamil for this brilliant work. :+1: :fireworks:

PWalkow:
:+1: Thanks! Rebasing now!

pamil:
Nearly a hundred Travis builds on this branch, it was worth it! :tada:

pjedrzejewski:
@pamil :D

michalmarcinkowski:
The travis build time is no longer a pain, with such a fast feedback we can really push forward the development! Good job :+1:

PR #3583, opened 2015-11-18
WebBundle: using form variables that are available, not extra params

PWalkow:
| Q             | A
| ------------- | ---
| Bug fix?      | [yes, possibly]
| New feature?  | [no]
| BC breaks?    | [no]
| Deprecations? | [no]
| License       | MIT

We were using cart.item's values rather than values that were available in the form itself. Strange looking loop.index0 in loop has been removed. I found this one while trying to add 'recalculate' on summary page. Error was:
```Key "1" in object with ArrayAccess of class "Symfony\Component\Form\FormView" does not exist in SyliusWebBundle:Frontend/Cart:_item.html.twig at line 8```
because cart items are not reindexed while recalculating.


pjedrzejewski:
Nice catch, thank you Piotr! :+1:

PR #3584, opened 2015-11-18
Change payment method description field to text doctrine type

pjedrzejewski:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

No reason to limit it to 255 characters and I actually got an exception during fixtures loading, when lorem ipsum was too long for the database.

### DIFF
--- a/app/migrations/Version20151118223453.php
+++ b/app/migrations/Version20151118223453.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Sylius\Migrations;
+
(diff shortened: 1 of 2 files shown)


okwinza:
:+1:

michalmarcinkowski:
Good catch :)

Issue #3590, opened 2015-11-20
[User][RFC] User component requires mb_string extension


cdaguerre:
The default `Canonicalizer` requires `mb_string` extension, ref. https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/User/Canonicalizer/Canonicalizer.php#L23

I see three options:
- Add `ext-mbstring` to `composer.json`,
- Only use `mb_*` functions if the extension is loaded,
- Keep it as is and let the user implement his own `Canonicalizer` although I don't think this is a good idea ;)

I'd be glad to work on a PR.

PR #3591, opened 2015-11-20
[User] Fix canonicalizer if mb_string extension isnt loaded

cdaguerre:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3590 
| License       | MIT
| Doc PR        |

### DIFF
--- a/src/Sylius/Component/User/Canonicalizer/Canonicalizer.php
+++ b/src/Sylius/Component/User/Canonicalizer/Canonicalizer.php
@@ -20,6 +20,10 @@ class Canonicalizer implements CanonicalizerInterface
 {
     public function canonicalize($string)
     {
+        if (!extension_loaded('mbstring')) {
+            return strtolower($string);
+        }
+
         return null === $string ? null : mb_convert_case($string, MB_CASE_LOWER, mb_detect_encoding($string));
     }
 }


stloyd:
IMO better would be to use: https://packagist.org/packages/symfony/polyfill-mbstring

cdaguerre:
@stloyd So you would add that package as a hard dependency? Installing it even if the `mb_string` extension is available? 
Or add suggest `ext-mbstring` and `symfony/polyfill-mbstring` in `composer.json` and check if the `mb_convert_case` function exists in the `Canonicalizer`, ie. `function_exists('mb_convert_case')` rather than `extension_loaded('mb_string')`?

pjedrzejewski:
@carlowens :+1: for your second proposition. I think it is best solution right now.

stloyd:
IMO this should be done same as it's done for Symfony. Add `polyfill` to required & done, all checks (`function_exists()` etc.) will be handled already by this component.

pjedrzejewski:
@stloyd @cdaguerre Well, I am in favor of sticking to what Symfony follows, so I think we could actually add this one to deps.

@carlowens Sorry, I picked wrong mention autocomplete. :/

stloyd:
With `polyfill` we can also add usage of `intl` etc.

cdaguerre:
:+1:

PR #3596, opened 2015-11-20
[Travis] Composer-related performance optimzations

pamil:
- Travis now does not implicitly run `cache:clear` / `assets:install`  while installing packages by Composer
- Fixed composer.lock caching (while requiring `doctrine/mongodb-odm`)
- Removed `~/.composer` from Travis cache

Doing these things makes that every build, that does not change `composer.json` or `composer.lock`, will not have to upload new cache archive, which usually takes about 40-60s. Moreover, cache archives will be smaller so we will save another second or two here.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -157,6 +157,14 @@
         "sylius/web-bundle":         "self.version"
     },
     "scripts": {
+        "travis-no-javascript-build": [
+            "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
+            "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap"
+        ],
+        "travis-javascript-build": [
+            "@travis-no-javascript-build",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -77,7 +77,8 @@
         "twig/twig": "~1.11",
         "white-october/pagerfanta-bundle": "~1.0",
         "willdurand/hateoas-bundle": "~0.4",
-        "winzou/state-machine-bundle": "~0.2"
+        "winzou/state-machine-bundle": "~0.2",
+        "doctrine/mongodb-odm": "^1.0.2"
```
stloyd: This should be reverted, it's optional not required.

pamil: :+1:, haven't deleted it after testing composer cache script

pjedrzejewski:
Nice, 30 mins... what the hell! :) Thanks Kamil!

PR #3597, opened 2015-11-20
Add missing default value for PaymentMethod configuration

diimpp:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| License | MIT |

`sylius_payment_method_choice` uses method from this repository, which renders configuration defaults to be broken.

### DIFF
--- a/src/Sylius/Bundle/PaymentBundle/DependencyInjection/Configuration.php
+++ b/src/Sylius/Bundle/PaymentBundle/DependencyInjection/Configuration.php
@@ -68,7 +68,7 @@ class Configuration implements ConfigurationInterface
                             ->children()
                                 ->scalarNode('model')->defaultValue('Sylius\Component\Payment\Model\PaymentMethod')->end()
                                 ->scalarNode('controller')->defaultValue('Sylius\Bundle\ResourceBundle\Controller\ResourceController')->end()
-                                ->scalarNode('repository')->end()
+                                ->scalarNode('repository')->defaultValue('Sylius\Bundle\PaymentBundle\Doctrine\ORM\PaymentMethodRepository')->end()
                                 ->arrayNode('form')
                                     ->addDefaultsIfNotSet()
                                     ->children()


pjedrzejewski:
This is ORM specific repository. It should be configured in the <driver>.xml file in PaymentBundle. Repositories should not have default class in configuration tree. They are selected based on the driver.

diimpp:
I wasn't aware of this feature.


I've added repository parameter to `Sylius\Bundle\PaymentBundle\Resources\config\driver\orm.xml`
```xml
<parameter key="sylius.repository.payment_method.class">Sylius\Bundle\PaymentBundle\Doctrine\ORM\PaymentMethodRepository</parameter>
```
and it showed up in `debug:container`, but I'm getting original error, like before this parameter were set.
```
Catchable Fatal Error: Argument 1 passed to Sylius\Bundle\PaymentBundle\Form\Type\PaymentMethodChoiceType::Sylius\Bundle\PaymentBundle\Form\Type\{closure}() must be an instance of Sylius\Component\Payment\Repository\PaymentMethodRepositoryInterface, instance of Doctrine\ORM\EntityRepository given 
```

---

I would appreciate if you could say how the heck `$repository` variable is set. :)
```php
     /**
      * {@inheritdoc}
      */
     public function setDefaultOptions(OptionsResolverInterface $resolver)
     {
         parent::setDefaultOptions($resolver);
 
         $queryBuilder = function (Options $options) {
             $repositoryOptions = array(
                 'disabled' => $options['disabled'],
             );
 
             return function (PaymentMethodRepositoryInterface $repository) use ($repositoryOptions) {
                 return $repository->getQueryBuidlerForChoiceType($repositoryOptions);
             };
         };;
 
         $resolver
             ->setDefaults(array(
                 'query_builder' => $queryBuilder,
                 'disabled'      => false,
             ))
         ;
     }
```

Issue #3453, opened 2015-10-19
[QUESTION] Why ROLE_SYLIUS_ADMIN and ROLE_ADMINISTRATION_ACCESS?


peteward:
I've got these roles in my user.roles field in the DB. There seems to be code meaning attached to both.

Are both required? Should there be two (one seems to imply the other - it would be weird to be an admin without administration access)?

Is this needed at all with the RBAC setup? Is it outstanding legacy?

pjedrzejewski:
``ROLE_SYLIUS_ADMIN`` is legacy from "before-RBAC" era. Right now we use Symfony roles to restrict access to specific URLS (like API/admin), but RBAC decides who is the admin and who is not. Probably this is not included in the migration or you have not run it. Please reopen if anything is not clear. Of course all this will go to the docs to make things clear.

michalmarcinkowski:
So we should adjust the confusing fixtures https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/FixturesBundle/DataFixtures/ORM/LoadUsersData.php#L36. Should I open other issue for that?

peteward:
Yes there's quite a few references to both:

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/WebBundle/Resources/views/Frontend/layout.html.twig
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/WebBundle/Controller/Frontend/Account/OrderController.php
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/WebBundle/Controller/Backend/SecurityController.php

Definitely needs a clean-up here around the uses of both.

cdaguerre:
and here ;)
https://github.com/Sylius/Sylius/blob/master/app/config/security.yml#L35

pjedrzejewski:
Let's use this issue. :)

pjedrzejewski:
@cdaguerre Ouch, we should have a separate role for that to give it more flexibility, not every user with admin access should be able to impersonate. :) Good catch! :+1:

cdaguerre:
@pjedrzejewski Actually someone already had that in mind here: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius.yml#L669

pjedrzejewski:
[...]

PR #3601, opened 2015-11-20
Removed ROLE_SYLIUS_ADMIN

cdaguerre:


pjedrzejewski:
:+1:

pjedrzejewski:
These builds. That's just insane. :) Thank you Christian!

cdaguerre:
Blazing fast ;) I think this closes #3453 @pjedrzejewski

psyray:
Thanks for this but you forgot one :)
PR submitted above

psyray:
I think also we should set up a migration script to remove the ROLE_SYLIUS_ADMIN from db, no ?

michalmarcinkowski:
@psyray someone can use this role in his custom code, so I don't think we should remove it since this additional role doesn't break anything :)

psyray:
ok, you're right

PR #2574, opened 2015-03-12
[PoC] Refactor Sylius RBAC authorization to use Symfony Security Voter approach

stloyd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Fixed tickets | #2538
| License       | MIT

Alternative for #2542.

PR #3603, opened 2015-11-20
[RBAC][WIP] Use symfony security authorization_checker and voter 

cdaguerre:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2574
| License       | MIT
| Doc PR        | 

Alternative to #2574.

#### RBAC Component
It has to remain fully decoupled from Symfony's `Security` component.

##### Authorization checker
Based on the assumption that the `AuthorizationChecker` should handle both **roles** and **permissions**, the following decision tree appears when checking if the user should be granted access or not. This is basically what happens here: https://github.com/cdaguerre/Sylius/blob/voter-poc/src/Sylius/Component/Rbac/Authorization/Voter/RbacVoter.php#L62
The chart summarizes the suggested permission logic:
capture d ecran 2015-11-23 a 17 09 10

##### Voters
The RBAC component comes with two voters (not talking about Symfony's `VoterInterface` here):

- **Default voter**: The default voter will always return true if the user has the required role, regardless of the passed resource, ref. https://github.com/cdaguerre/Sylius/blob/voter-poc/src/Sylius/Component/Rbac/Authorization/Voter/RbacVoter.php#L113

- **Delegating voter**: 
The delegating voter will check all registered services tagged with `rbac.resource_voter` and return as soon as one of them claims to support the passed permission and resource. 
 - `ResourceVoter`s implement the `ResourceVoterInterface`, ref. here: https://github.com/cdaguerre/Sylius/blob/voter-poc/src/Sylius/Component/Rbac/Authorization/Voter/ResourceVoterInterface.php
 - An `AbstractResourceVoter` is provided for DX., ref. here: https://github.com/cdaguerre/Sylius/blob/voter-poc/src/Sylius/Component/Rbac/Authorization/Voter/AbstractResourceVoter.php

##### Permission map
In the current state of things, it would make sense to add the private `getPermission($code)` method to the public interface and add caching on said method in the `CachePermissionMap`. It would make the voters lighter and caching permission retrieval is mandatory anyway.
Any reasons not to do so?

### Bundle
In the `RBACBundle`, the voter is configurable through the Bundle's semantic configuration, and is wrapped by the `RbacVoterBridge` (which implements the Symfony `VoterInterface`), see.  https://github.com/cdaguerre/Sylius/blob/voter-poc/src/Sylius/Bundle/RbacBundle/Authorization/Voter/RbacVoterBridge.php

## ToDo
- [x] Remove `sylius.authorization_checker` in favor of Symfony's `security.authorization_checker`
- [x] Remove redundant Twig extension
- [x] Implement `security.voter` enabling checks on specific permissions, optionally for a provided a resource 
- [x] Update `ResourceController`'s 403 method and calls
- [ ] Decouple `UserBundle` from `RBACBundle`
- [ ] Fix Behat tests
- [ ] Complete specs
- [ ] Check whether the default `ROLE_*` based Symfony voter needs to be disabled in `RbacBundle`

Thoughts and feedback much appreciated ;)


src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
```diff
@@ -540,8 +540,8 @@
             <tag name="form.type" alias="sylius_data_fetcher_number_of_orders" />
         </service>
 
-        <service id="sylius.authorization_checker.toggleable" class="%sylius.authorization_checker.toggleable.class%">
-            <argument type="service" id="sylius.authorization_checker.default" />
+        <service id="sylius.rbac.voter" class="%sylius.rbac.voter.toggleable.class%" public="false">
```
pamil: We can use [Symfony decorator](http://symfony.com/doc/current/components/dependency_injection/advanced.html#decorating-services) here.

src/Sylius/Bundle/RbacBundle/Security/RbacVoter.php
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * @var int
+     */
+    protected $ttl;
+
+    public function __construct(
+        PermissionMapInterface $permissionMap,
+        PermissionProviderInterface $permissionProvider,
+        RolesResolverInterface $rolesResolver,
+        Cache $permissionsCache,
```
pamil: What do you think about extracting caching logic to another decorator?

src/Sylius/Bundle/RbacBundle/Security/RbacVoter.php
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * @var Cache
+     */
+    protected $permissionsCache;
+
+    /**
+     * @var int
+     */
+    protected $ttl;
+
```
pamil: Missing docblock :)

src/Sylius/Bundle/RbacBundle/Security/RbacVoter.php
```diff
@@ -0,0 +1,143 @@
...
+    public function __construct(
+        PermissionMapInterface $permissionMap,
+        PermissionProviderInterface $permissionProvider,
+        RolesResolverInterface $rolesResolver,
+        Cache $permissionsCache,
+        $ttl = self::DEFAULT_TTL
+    ) {
+        $this->permissionMap      = $permissionMap;
```
pamil: It's my personal feel, but if we use another, longer variable here, we will have to modify all the assignments.

src/Sylius/Bundle/RbacBundle/Security/RbacVoter.php
```diff
@@ -0,0 +1,143 @@
...
+            return false;
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supportsClass($class)
+    {
+        return in_array('Sylius\Component\Rbac\Model\IdentityInterface', class_implements($class));
```
pamil: `IdentityInterface::class` - to be consistent with `instanceof` a few lines below.

src/Sylius/Bundle/RbacBundle/Security/RbacVoter.php
```diff
@@ -0,0 +1,143 @@
...
+    {
+        $identity = $token->getUser();
+
+        if (!$identity instanceof IdentityInterface) {
+            return VoterInterface::ACCESS_ABSTAIN;
+        }
+
+        $deny = false;
+
+        foreach ($attributes as $permissionCode) {
```
pamil: We can use [simpler voters](http://symfony.com/blog/new-in-symfony-2-6-simpler-security-voters), that would simplify logic, we won't have to implement handling multiple arguments :)

src/Sylius/Bundle/RbacBundle/Security/RbacVoterInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+namespace Sylius\Bundle\RbacBundle\Security;
+
+use Symfony\Component\Security\Core\Authorization\Voter\VoterInterface;
+
+/**
+ * @author Christian Daguerre <christian@daguer.re>
+ */
+interface RbacVoterInterface extends VoterInterface
```
pamil: I'm not sure about this empty interface, do we really benefit from it? We ans Symfony use `VoterInterface` method only.

src/Sylius/Bundle/RbacBundle/spec/Security/RbacVoterSpec.php
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * @author Christian Daguerre <christian@daguer.re>
+ */
+class RbacVoterSpec extends ObjectBehavior
+{
+    function let(
+        PermissionMapInterface $permissionMap,
+        PermissionProviderInterface $permissionProvider,
+        RolesResolverInterface $rolesResolver,
+        Cache $cache
```
pamil: As I said before, extracting caching can simplify this voter, in case of specs it will be much easier to write specification from caching and voting logic standalone.

src/Sylius/Bundle/RbacBundle/DependencyInjection/Compiler/TagVoterPass.php
```diff
@@ -0,0 +1,31 @@
...
+ * @author Christian Daguerre <christian@daguer.re>
+ */
+class TagVoterPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if ($container->hasDefinition('sylius.rbac.voter')) {
```
pamil: I'm not sure, but if you use Symfony decorators, we can delete this compiler pass and tag the base voter instead.

pamil:
Good job :+1:

isometriks:
This would be great to have for projects that don't use RBAC so the default symfony authorization works properly :+1:

pjedrzejewski:
I am :+1: as long as we still can use Rbac component standalone without Symfony Security.

liverbool:
How about milestone of this? @pjedrzejewski :)

cdaguerre:
@pamil Thanks for your valuable feedback Kamil and props for the builds ;)
Believe it or not, in our case the ["simple" abstract voter](http://symfony.com/blog/new-in-symfony-2-6-simpler-security-voters) is harder to implement than implementing the interface, so sticking with our own...
Still working on the DI and I'm having a hard time figuring out how to use service decoration to make things cleaner. I pushed some stuff I did and updated the issue description with other points that need to be discussed, too. Once again, feedback appreciated!

/cc @sylius/core-team

michalmarcinkowski:
@cdaguerre in order to mention the core-team you just need to make the first letter upper case. Like this @Sylius/core-team :)

cdaguerre:
I'll come up with a much simpler and cleaner PR later today.

pjedrzejewski:
@cdaguerre Sounds great, I love the idea. :+1:

PR #3604, opened 2015-11-21
[Behat] [Travis] Investigating failing JS builds

pamil:
- Longer timeouts for accessing database (half of fails, timeout was set to 1s, increased to 3s)
- Shorter timeouts for loading page by Selenium (another half of fails, timeout was set to 10min [equal to Travis no-output timeout :/], decreased to 30s)

I hope the second change will give us insight about the issue and help to fix it.


pjedrzejewski:
Thank you Kamil!

PR #3605, opened 2015-11-21
[Behat] Missing security suite added, ordered alphabetically

pamil:
- Added `security` suite, which was missing
- Updated `security` scenarios (new table matching)
- Ordered Behat suites in `travis/suites/*` alphabetically


pjedrzejewski:
Thanks Kamil! :)

PR #3606, opened 2015-11-21
[Core] Removed unused file and parameter defined in it

### LINKED ISSUE #3598: Sylius\Component\Core\Model\VariantImage used, but not defined
`Sylius\Component\Core\Model\VariantImage` is used in [@CoreBundle/Resources/config/driver/doctrine/orm.xml](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Resources/config/driver/doctrine/orm.xml) but does not exist.

pamil:
Fixes #3598.


pjedrzejewski:
:+1:

PR #3607, opened 2015-11-23
[Report] Replace private variable declarations with protected ones

pamil:
I need to overwrite `unique` property of `Report::$code` in my project and I'm getting `Property MyComponent\Model\Report::$code does not exist` exception.


pjedrzejewski:
Thank you Kamil!

PR #3609, opened 2015-11-23
[Composer] Updated phpspec requirement from 2.4@alpha to 2.4@rc

pamil:
Doing so, packages dependent on `sylius/sylius` can use `"minimal-stability": "rc"` instead of `"minimal-stability": "alpha"`.


pjedrzejewski:
:+1:

PR #3480, opened 2015-10-23
[Taxonomy] [RFC] Constructor for the Taxonomy model

TheMadeleine:
| Q                    | A
| -------------       | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #2884 #2557 
| License       | MIT

While using component independently, if we wanted to use ``$taxonomy = new Taxonomy()`` we would get an error, because we are not setting the root taxon while creating a new Taxonomy. Therefore I suppose we need a constructor.

PR #3613, opened 2015-11-23
Introduce Factory for all resources

### LINKED ISSUE #480: Variant class hardcoded in Product
The CoreBundle Product constructor currently instantiates a concrete Variant. This makes it impossible to use a different Variant implementation without replacing the entire constructor.

I'm not familiar enough with Doctrine to know if the variant should be given to the Product constructor.. this would also create a dependency between the Product and Variant repositories.

Thoughts?

pjedrzejewski:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | https://github.com/Sylius/Sylius-Docs/pull/370

This is first step extracted from #2255. The original PR became too big and unmergeable and very hard to maintain for me. It was not the best move from my side to make it that big, but it gave me a good insight into how I want SyliusResourceBundle to look like.

This is the first step, which will effectively allow us to fix #480, #3480 and probably much more bugs. Also, it makes everything much cleaner and we can add more methods to our Factories. For now, I kept only createNew by default (except InventoryUnitFactory, which required changes) because I want to keep this PR as slim as possible.

Main thing here is that we can easily inject stuff into Factories, which was not possible/simple/clean with Repository taking care of object creation. I updated all tests, specs and also tracked down ``createNew`` calls in stuff that has no coverage of specs/tests/scenarios.

This is a BC break, pretty big one, but this PR took me maybe 8-12 hours to update whole Sylius, so it should be a pretty straightforward upgrade. :)

Review appreciated!

* [x] Update changelogs
* [x] Fix the build
* [x] Add a PR to documentation

QA:

* [x] ORM fixtures
* [x] PHPCR Fixtures
* [x] sylius:install


peteward:
:+1:  looks sensible to me, although difficult to see the bigger picture 

It took you 8 - 12 hours to do find/replace?

pjedrzejewski:
@peteward I wish it was as simple as find replace. You need to update specs etc. but sometimes it does not mean only replacing Repository with a Factory. Sometimes you need both etc. I think 8-12 hours is too much, was just a shot to give some idea how difficult upgrade is.

peteward:
I was joking ;) good work

pjedrzejewski:
Green! :tada:

pamil:
Good job :+1: Related PR: #3480 :)

src/Sylius/Component/Translation/Factory/TranslatableFactory.php
```diff
@@ -0,0 +1,56 @@
...
+
+use Sylius\Component\Resource\Exception\UnexpectedTypeException;
+use Sylius\Component\Resource\Factory\Factory;
+use Sylius\Component\Translation\Model\TranslatableInterface;
+use Sylius\Component\Translation\Provider\LocaleProviderInterface;
+
+/**
+ * @author Paweł Jędrzejewski <pawel@sylius.org>
+ */
+class TranslatableFactory extends Factory implements TranslatableFactoryInterface
```
pamil: Can we make this a decorator for `FactoryInterface`? Doing so, we can write specs for setting both current locale and fallback locale.

src/Sylius/Component/Translation/spec/Factory/TranslatableFactorySpec.php
```diff
@@ -0,0 +1,50 @@
...
+    }
+    
+    function it_implements_translatable_factory_interface()
+    {
+        $this->shouldImplement('Sylius\Component\Translation\Factory\TranslatableFactoryInterface');
+    }
+
+    function it_creates_translatable_and_sets_locales(LocaleProviderInterface $localeProvider)
+    {
+        $localeProvider->getCurrentLocale()->willReturn('pl_PL');
```
pamil: Like in a comment at `TranslatableFactory` definition, we don't check whether these locales are set on the resource.

src/Sylius/Component/Translation/Factory/TranslatableFactoryInterface.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Sylius\Component\Translation\Factory;
+
+use Sylius\Component\Resource\Factory\FactoryInterface;
+
+/**
+ * Interface for factories creating translatable resources.
+ *
+ * @author Paweł Jędrzejewski <pawel@sylius.org>
+ */
+interface TranslatableFactoryInterface extends FactoryInterface
```
pamil: What's the point with `TranslatableFactoryInterface`? IMHO it would have the same methods as `FactoryInterface`, so it will be empty, wouldn't it? If we only overwrite the phpdocs with `@return AbstractTranslatable` (not sure about the classname though :)), it will have more sense.

src/Sylius/Component/Translation/Factory/TranslatableFactory.php
pamil: Anyway, if we rewrite this class as decorator, it is possible to make some vendor class translatable even without changing its factory code.

src/Sylius/Component/Translation/Factory/TranslatableFactoryInterface.php
pjedrzejewski: This interface is a marker right now and I think we should keep it, cause I am thinking about introducing some methods like ``createForLocale`` etc.

As for PHPDoc - Yes, when we rework translations we should add it here, but not AbstractTranslatable but probably TranslatableInterface. :) WDYT? (about both points)

src/Sylius/Component/Translation/Factory/TranslatableFactory.php
pjedrzejewski: I know there was a reason why I have not used a decoration here, can't recall now. Will try to do this and see how it works out, should not be that hard to change.

src/Sylius/Component/Translation/Factory/TranslatableFactoryInterface.php
pamil: Okay, seems right. I'm not much into all the translation stuff, but will there be any problem, if someone calls `createForLocale` two times for the same base entity but different languages?

src/Sylius/Bundle/ArchetypeBundle/DependencyInjection/Configuration.php
```diff
@@ -12,10 +12,13 @@
 namespace Sylius\Bundle\ArchetypeBundle\DependencyInjection;
 
 use Sylius\Bundle\ResourceBundle\SyliusResourceBundle;
+use Sylius\Component\Resource\Factory\Factory;
+use Sylius\Component\Translation\Factory\TranslatableFactory;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
 
+
```
michalmarcinkowski: Why?

src/Sylius/Bundle/ShippingBundle/DependencyInjection/Configuration.php
```diff
@@ -186,21 +192,13 @@ private function addClassesSection(ArrayNodeDefinition $node)
...
                         ->end()
-//                        ->arrayNode('shipping_method_translation')
-//                            ->addDefaultsIfNotSet()
-//                            ->children()
-//                                ->scalarNode('model')->defaultValue('Sylius\Component\Shipping\Model\ShippingMethodTranslation')->end()
-//                                ->scalarNode('controller')->defaultValue('Sylius\Bundle\ResourceBundle\Controller\ResourceController')->end()
-//                                ->scalarNode('repository')->end()
-//                                ->scalarNode('form')->defaultValue('Sylius\Bundle\ShippingBundle\Form\Type\ShippingMethodTranslationType')->end()
-//                            ->end()
-//                        ->end()
```
michalmarcinkowski: :+1:

src/Sylius/Component/Core/Promotion/Action/AddProductAction.php
```diff
@@ -41,14 +40,12 @@ class AddProductAction implements PromotionActionInterface
...
-     *
-     * @param RepositoryInterface $itemRepository
+     * @param FactoryInterface $itemFactory
      * @param RepositoryInterface $variantRepository
      */
-    public function __construct(RepositoryInterface $itemRepository, RepositoryInterface $variantRepository)
+    public function __construct(FactoryInterface $itemFactory, RepositoryInterface $variantRepository)
     {
-        $this->itemRepository    = $itemRepository;
+        $this->itemFactory    = $itemFactory;
```
michalmarcinkowski: Unnecessary blank spaces before `=`

src/Sylius/Component/Core/spec/OrderProcessing/InventoryHandlerSpec.php
```diff
@@ -47,7 +50,7 @@ function it_implements_Sylius_inventory_handler_interface()
     }
 
     function it_creates_inventory_units_via_the_factory(
-        $inventoryUnitFactory,
+        InventoryUnitFactoryInterface $inventoryUnitFactory,
```
michalmarcinkowski: We should use the interface also here https://github.com/pjedrzejewski/Sylius/blob/introduce-resource-factories/src/Sylius/Component/Core/spec/OrderProcessing/InventoryHandlerSpec.php#L36

src/Sylius/Bundle/ArchetypeBundle/DependencyInjection/Configuration.php
pjedrzejewski: Misclick.

src/Sylius/Component/Core/spec/OrderProcessing/InventoryHandlerSpec.php
pjedrzejewski: :+1:

src/Sylius/Component/Inventory/Factory/InventoryUnitFactory.php
```diff
@@ -52,7 +35,7 @@ public function create(StockableInterface $stockable, $quantity, $state = Invent
         $units = new ArrayCollection();
 
         for ($i = 0; $i < $quantity; $i++) {
-            $inventoryUnit = $this->repository->createNew();
+            $inventoryUnit = parent::createNew();
```
michalmarcinkowski: The method is inherited, so we should use `$inventoryUnit = $this->createNew();`.

pjedrzejewski: Good catch, I think this is leftover from some experimenting. :)

src/Sylius/Component/Translation/Factory/TranslatableFactoryInterface.php
michalmarcinkowski: I will create two resources with different locales set, so it shouldn't break anything ;)

src/Sylius/Component/Translation/Factory/TranslatableFactory.php
michalmarcinkowski: If there is nothing against going for decorator patter, we should pick this direction, because it definitely makes sense here.

michalmarcinkowski:
@pjedrzejewski great job! :+1: Small step that starts the big improvement, looking forward to see more!

src/Sylius/Component/Translation/Factory/TranslatableFactoryInterface.php
pjedrzejewski: Yeah, but with Translatable, two different locales is still single resource but with 2 *Translations inside. :) We will need to see how it works out. Something for later. :) Out of scope for this PR. :dancer:

src/Sylius/Component/Core/Promotion/Action/AddProductAction.php
pamil: :+1:

pjedrzejewski: I need one more :+1: to be sure. :D

SyliusBot: :+1:  for not aligning assignments

pjedrzejewski:
Applied all changes. Good to go! :) Woohoo!

michalmarcinkowski:
@pjedrzejewski one more thing left, global changelog update :)

PR #3614, opened 2015-11-23
Throwing correct exception if step is unknown

tenitski:
At the moment Process.php throws ```\InvalidArgumentException``` however the code which uses this class (```Sylius\Bundle\FlowBundle\Controller\ProcessController```) catches ```Sylius\Bundle\FlowBundle\Process\Coordinator\InvalidArgumentException``` instead resulting in 5xx error where 404 is expected


pjedrzejewski:
Sounds good, thanks Alexei! We have to change this exception to new one with a more meaningful name. :+1:

PR #3624, opened 2015-11-25
[Payum] Integration with Paypal Pro Checkout

GSadee:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | no
| License       | MIT
| Doc PR        | no

### DIFF
--- a/app/config/parameters.yml.dist
+++ b/app/config/parameters.yml.dist
@@ -41,6 +41,12 @@ parameters:
     paypal.express_checkout.signature: EDITME
     paypal.express_checkout.sandbox: true
 
+    paypal.pro_checkout.username: EDITME
+    paypal.pro_checkout.password: EDITME
+    paypal.pro_checkout.partner: EDITME
+    paypal.pro_checkout.vendor: EDITME
+    paypal.pro_checkout.sandbox: true
+
     stripe.publishable_key: EDITME
     stripe.secret_key: EDITME
     stripe.test_mode: true
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/payum.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/payum.yml
@@ -19,6 +19,14 @@ payum:
                 signature: %paypal.express_checkout.signature%
                 sandbox:   %paypal.express_checkout.sandbox%
 
+        paypal_pro_checkout:
+            paypal_pro_checkout_nvp:
+                username: %paypal.pro_checkout.username%
+                password: %paypal.pro_checkout.password%
+                partner:  %paypal.pro_checkout.partner%
+                vendor:   %paypal.pro_checkout.vendor%
+                sandbox:  %paypal.pro_checkout.sandbox%
+
         be2bill_direct:
             be2bill_direct:
                 identifier: %be2bill.identifier%
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius.yml
@@ -167,6 +167,7 @@ sylius_payment:
         be2bill: Be2Bill
         stripe_checkout: Stripe Checkout
         paypal_express_checkout: Paypal Express Checkout
+        paypal_pro_checkout: Paypal Pro Checkout
 
 sylius_payum: ~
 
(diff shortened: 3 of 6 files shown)


src/Sylius/Bundle/PayumBundle/Action/ConvertPaymentToPaypalProAction.php
```diff
@@ -0,0 +1,57 @@
...
+use Payum\Core\Exception\RequestNotSupportedException;
+use Payum\Core\Request\Convert;
+use Sylius\Component\Core\Model\AdjustmentInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Model\PaymentInterface;
+
+/**
+ * @author Grzegorz Sadowski <grzegorz.sadowski@lakion.com>
+ */
+class ConvertPaymentToPaypalProAction implements ActionInterface
```
makasim: Not sure this class is needed, this one already does same thing: https://github.com/Payum/Payum/blob/master/src/Payum/Paypal/ProCheckout/Nvp/Action/ConvertPaymentAction.php#L12

Doesn't it work for you?

app/config/parameters.yml.dist
```diff
@@ -41,6 +41,12 @@ parameters:
     paypal.express_checkout.signature: EDITME
     paypal.express_checkout.sandbox: true
 
+    paypal.pro_checkout.username: EDITME
+    paypal.pro_checkout.password: EDITME
+    paypal.pro_checkout.partner: EDITME
+    paypal.pro_checkout.vendor: EDITME
+    paypal.pro_checkout.sandbox: true
```
pjedrzejewski: We should not add these parameters here. I know it sucks, but we don't want to have any of these in the parameters.yml.dist, that should be configurable in the backend and it is already possible, but there is no UI for that. Let's add it in our project, but not in Sylius itself. :+1:

src/Sylius/Bundle/CoreBundle/Resources/config/app/payum.yml
```diff
@@ -19,6 +19,14 @@ payum:
...
                 sandbox:   %paypal.express_checkout.sandbox%
 
+        paypal_pro_checkout:
+            paypal_pro_checkout_nvp:
+                username: %paypal.pro_checkout.username%
+                password: %paypal.pro_checkout.password%
+                partner:  %paypal.pro_checkout.partner%
+                vendor:   %paypal.pro_checkout.vendor%
+                sandbox:  %paypal.pro_checkout.sandbox%
+
```
pjedrzejewski: Same as for parameters.yml.dist comment.

src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius.yml
```diff
@@ -167,6 +167,7 @@ sylius_payment:
         be2bill: Be2Bill
         stripe_checkout: Stripe Checkout
         paypal_express_checkout: Paypal Express Checkout
+        paypal_pro_checkout: Paypal Pro Checkout
```
pjedrzejewski: Same as above, we will later figure a way to load these automatically. :)

src/Sylius/Bundle/FixturesBundle/DataFixtures/ORM/LoadPaymentMethodsData.php
```diff
@@ -29,6 +29,7 @@ public function load(ObjectManager $manager)
     {
         $manager->persist($this->createPaymentMethod('Dummy', 'dummy', 'fixed', array('amount' => 0)));
         $manager->persist($this->createPaymentMethod('PaypalExpressCheckout', 'paypal_express_checkout', 'fixed', array('amount' => 1000)));
+        $manager->persist($this->createPaymentMethod('PaypalProCheckout', 'paypal_pro_checkout', 'fixed', array('amount' => 0)));
```
pjedrzejewski: Let's remove this too.

app/config/parameters.yml.dist
makasim: @pjedrzejewski  :+1:  to make it configurable via backend, would be a great step further

src/Sylius/Bundle/FixturesBundle/DataFixtures/ORM/LoadPaymentMethodsData.php
makasim: :+1:

src/Sylius/Bundle/PayumBundle/Action/ConvertPaymentToPaypalProAction.php
GSadee: You are right of course, in my case, I don't need this custom class :+1:

GSadee:
So these changes are unnecessary at this moment

makasim:
It would be good to put such things to documentation, on how to configure different payment gateways, in this case Paypal Pro

michalmarcinkowski:
@makasim that's the plan, but we probably won't have time to do this in the nearest future...

PR #3164, opened 2015-08-20
[Support] Refactoring contact

GSadee:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #3545, #1951
| License       | MIT
| Doc PR        | no

Renamed Contact component to Support and ContactBundle to SupportBundle

PR #3625, opened 2015-11-25
[Behat] Drastically reduced memory usage

pamil:
Based on #3627.

Eg. `checkout` suite dropped form 1.6GB to 250MB, now we can run whole Behat in one command! :)

### DIFF
--- a/app/TestKernel.php
+++ b/app/TestKernel.php
@@ -1,0 +1,56 @@
+<?php
+
+use Sylius\Bundle\CoreBundle\Kernel\Kernel;
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+class TestKernel extends Kernel
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function shutdown()
+    {
+        if (false === $this->booted) {
+            return;
+        }
+
+        if (!in_array($this->environment, ['test', 'test_cached'])) {
+            parent::shutdown();
+            return;
+        }
+
+        $container = $this->container;
+        parent::shutdown();
+        $this->cleanupContainer($container);
+    }
+
+    /**
+     * Remove all container references from all loaded services
+     */
+    protected function cleanupContainer($container)
+    {
+        $object = new \ReflectionObject($container);
+        $property = $object->getProperty('services');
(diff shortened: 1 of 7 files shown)


app/TestKernel.php
```diff
@@ -0,0 +1,58 @@
...
+            $serviceObject = new \ReflectionObject($service);
+            foreach ($serviceObject->getProperties() as $prop) {
+                $prop->setAccessible(true);
+                if ($prop->isStatic()) {
+                    continue;
+                }
+
+                try {
+                    $prop->setValue($service, null);
+                } catch (\ReflectionException $e) {
```
PWalkow: does it mean that sometimes not everything is being cleared?

behat.yml.dist
```diff
@@ -400,16 +400,18 @@ default:
...
-        Behat\Symfony2Extension: ~
+        Behat\Symfony2Extension:
+            kernel:
+                class: TestKernel
+                path: app/TestKernel.php
 
     gherkin:
         filters:
-            tags: "~@javascript"
+            tags: "~@javascript && ~@contact"
```
PWalkow: why this contact is so bad?
isn't easier to fix the contact suite?

bin/sylius-cs-fixer
```diff
@@ -1,4 +0,0 @@
-#!/bin/bash
-
-php-cs-fixer fix . --config-file .php_cs
```
PWalkow: what happened to php-cs-fixer? does this change with behat really involves php-cs-fixing? if not maybe another PR would be better place for discussion?

behat.yml.dist
pamil: Contact is currently refactored in #3164.

app/TestKernel.php
pamil: Don't know exactly, its slightly modified version of snippet from symfony/symfony#11236.

bin/sylius-cs-fixer
pamil: :+1:, reverted

pjedrzejewski:
Boom! This is just insane. Thanks Kamil!

Issue #3621, opened 2015-11-25
Doctrine Metadata Cache


inssein:
I am having a ton of problems running Sylius in production, when metadata caching is enabled with APC. I don't think I am alone, as a few people (@kayue, @Richtermeister) reported the same issues in  https://github.com/Sylius/Sylius/issues/1363.

The issue is that all Gedmo extensions stop working when they are inherited. To give an example, my Order model doesn't set automatic timestamps for updatedAt and deletedAt, just because they are defined initially in OrderBundle, but then extended in CoreBundle.

What is everyone else doing to get around this problem? Unless you have a multi-server deployment, and relying on memcached, I don't really see another way, short of just not caching the metadata.

@pjedrzejewski does the sylius demo not use any metadata cache?

inssein:
Fixed by https://github.com/Sylius/Sylius/pull/3627.

Issue #1438, opened 2014-04-27
Sylius override product Integrity constraint violation


Bahlaouane-Hamza:
I'm trying to override the product bundle, i did it successfuly but getting an error when trying to delete a parent entity that have products as childs

here's the mapping file product.orm.xml

    <?xml version="1.0" encoding="UTF-8"?>
    <doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
                      xmlns:gedmo="http://gediminasm.org/schemas/orm/doctrine-extensions-mapping">

        <mapped-superclass name="Sylius\Component\Product\Model\Product" table="sylius_product">
            <id name="id" column="id" type="integer">
                <generator strategy="AUTO" />
            </id>

            <field name="name" column="name" type="string">
                <gedmo:versioned />
            </field>
            ...
        </mapped-superclass>

    </doctrine-mapping>

and here's my product.orm.xml file

    <?xml version="1.0" encoding="UTF-8"?>
    <doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
                      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                      xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
                                          http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
        <entity name="WeShop\Bundle\CoreBundle\Entity\Product" table="sylius_product">
            <many-to-one field="boutique" target-entity="WeShop\Bundle\BoutiqueBundle\Entity\Boutique" inversed-by="produits">
                <join-column name="boutique_id" referenced-column-name="id" nullable="true" onDelete="SET NULL" />
            </many-to-one>
        </entity>
    </doctrine-mapping>

in WeShop\Bundle\BoutiqueBundle\Entity\Boutique i do have

    <?php
    namespace WeShop\Bundle\BoutiqueBundle\Entity;
    use Sylius\Component\Core\Model\ProductInterface as ProductInterface;
    use Doctrine\ORM\Mapping as ORM;
    use Doctrine\Common\Collections\ArrayCollection;
    
    class Boutique
    {
        private $produits;
    
        public function __construct(){
            $this->produits = new ArrayCollection();
        }
    
        public function getProduits(){
            return $this->produits;
        }
    
        public function setProduits(ProductInterface $produit){
            $this->produits[] = $produit;
            return $this;
        }
    
    }

[...]

Issue #478, opened 2013-10-29
Taxonomy Issues


Richtermeister:
Hi all,

I noticed a few issues/areas for improvement for the Taxonomy Bundle.

 * When removing taxons, sometimes siblings get removed (not just children). This looks like a gedmo mapping issue.. Does the gedmo:tree-root parameter need to be mapped?

 * The parent/children properties are mapped in the CoreBundle, but it should be possible to keep this mapping in the TaxonomiesBundle (against Interfaces). Is this just a remnant from the pre-DRTE days?

 * There is a Taxonomy.orm.xml mapping file in the CoreBundle that adds nothing to the mapping other than the table name. Is this neccessary, or can the table name live in the Taxonomies bundle? It seems where the table name lives is inconsistent from bundle to bundle..

I'm happy to tackle these issues if I could get some feedback on these questions.

Thanks!

pjedrzejewski:
@beleneglorion Why would you delete root taxon? Root taxon has all other taxons under it, so it removes them as well.

beleneglorion:
@pjedrzejewski i think you misunderstood the problem,
 Root Taxon A
    -> Child A1
    -> Child A2
     .
    -> Child An
    Root Taxon B
    -> Child B1
    -> Child B2
     .
    -> Child Bn

When you delete Root A taxon () Taxon child taxons from Root B are deleted to ....

and i need to delete a root Taxon because some time my client remove a category from his catalogue of tools ;)

antipodes1010:
Any updates on this one?

I have the same issue like the first point.
To reproduce:

Taxon 1
--> child 1.1

Taxon 2
--> child 2.1

Delete Taxon 1 and then neither Taxon 1 and 2 would show up.
Check the sylius_taxon table, all four would have 'deleted_at' values.

And change 'cascade-all' to 'cascade-persist' does not work for me, complaint about 'slug' cannot be null.

michalmarcinkowski:
Will be fixed in #3773.

michalmarcinkowski:
Should be fixed already

PR #3627, opened 2015-11-25
[Resource] Fixed issue with mapping inheritance and mapped superclass (Gedmo)

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1363, #3621, #1438, #478
| License       | MIT

Probably fixes all bugs related to Gedmo DoctrineExtensions and mapping inheritance :)


pjedrzejewski:
Thanks Kamil!

inssein:
@pamil sweet! I will definitely be checking this out once I get home.

From taking a quick look, I can't seem to make sense of it easily. How does loading of the metadata when using cache get fixed by changing the priority of the event?

pamil:
@inssein SyliusResourceBundle does some really weird things with mapping (allowing mapped superclasses to have relationships and ids, etc.). As far as I know, it happened only while metadata cache was enabled and when the initial class metadata cache was created not in main application, but in some command (I'm not 100% sure here).

Anyway, it resulted in the `loadClassMetadata` listeners being in the order like that:
 - SluggableListener
 - TreeListener
 - **LoadORMResourcesListener**
 - TimestampableListener
 - SoftDeletableListener

This is the reason we had those problems with slugs and trees, but not with much more widely used timestampable and softdeletable behaviours.

If you take a look at Gedmo's `ExtensionMetadataFactory::getExtensionMetadata($meta)`, you'll see this code at the beginning:
```php
if ($meta->isMappedSuperclass) {
    return; // ignore mappedSuperclasses for now
}
```

So our mapped superclasses were transformed into entities just after the first two Gedmo listeners were called and their metadata was empty at that moment :)

inssein:
@pamil For me (https://github.com/Sylius/Sylius/issues/3621) Timestampable also wasn't working.

pamil:
@inssein It depends on how your application looked like, with ResourceBundle listeners guaranteed to be on top, it should work everywhere :)

PR #3628, opened 2015-11-25
[Product] Fixed ProductBuilderSpec on PHP7

pamil:
Doubled definition in `it_creates_attribute_if_it_does_not_exist` removed.
Also ordered the parameters in the same way as they are in `let` :)


pjedrzejewski:
Scrutinizer dude, you need to catch up! Thank you Kamil! :+1:

PR #3630, opened 2015-11-25
[Promotion] Made scenarios for product promotions more specific

pamil:
Sometimes, "Ubu" was matched from "Newest products" sidebar.


pjedrzejewski:
Thanks Kamil! Crazy how many weird issues we had on the way, but seems like we are almost there. :D

PR #3634, opened 2015-11-26
[Travis] OPCache optimizations

pamil:
This should prevent `zend_mm_heap corrupted` errors happening sometimes while running `phpspec` and `composer`.

### DIFF
--- a/travis/prepare/opcache.ini
+++ b/travis/prepare/opcache.ini
@@ -3,8 +3,9 @@
 opcache.enable = 1
 opcache.enable_cli = 1
 
-opcache.memory_consumption = 128
-opcache.max_accelerated_files = 4000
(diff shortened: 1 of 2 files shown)


travis/prepare/opcache.ini
```diff
@@ -3,8 +3,9 @@
 opcache.enable = 1
 opcache.enable_cli = 1
 
-opcache.memory_consumption = 128
-opcache.max_accelerated_files = 4000
-opcache.revalidate_freq = 60
+opcache.memory_consumption = 256
+opcache.max_accelerated_files = 32531
```
PWalkow: Couple of comment lines would not kill this file. Maybe it would be good to say why this value is 32531 ?

PR #3635, opened 2015-11-26
[FlowBundle] Fix version so we can use this bundle with Symfony 2.8 too

tristanbes:


src/Sylius/Bundle/FlowBundle/composer.json
```diff
@@ -22,7 +22,7 @@
     "require": {
         "php": ">=5.5.9",
 
-        "symfony/framework-bundle": "~2.7.5"
+        "symfony/framework-bundle": "~2.7,>=2.7.5"
```
michalmarcinkowski: We can simply use `^` instead of `~` and that should be corrected in all bundles...

pamil: Should be `^2.7.7`, versions from `2.7` to `2.7.6` have [some security issues](http://symfony.com/blog/symfony-2-7-7-released).

tristanbes:
done

pjedrzejewski:
Thank you Tristan!

PR #3637, opened 2015-11-26
Stateless cart provider

PWalkow:
Services should be stateless. CartProvider holding cart was just premature optimisation and just wrong.

### DIFF
--- a/src/Sylius/Bundle/CartBundle/Provider/CartProvider.php
+++ b/src/Sylius/Bundle/CartBundle/Provider/CartProvider.php
@@ -11,36 +11,26 @@
 
 namespace Sylius\Bundle\CartBundle\Provider;
 
-use Doctrine\Common\Persistence\ObjectManager;
 use Sylius\Component\Cart\Context\CartContextInterface;
-use Sylius\Component\Cart\Event\CartEvent;
 use Sylius\Component\Cart\Model\CartInterface;
 use Sylius\Component\Cart\Provider\CartProviderInterface;
 use Sylius\Component\Cart\SyliusCartEvents;
 use Sylius\Component\Resource\Factory\FactoryInterface;
 use Sylius\Component\Resource\Repository\RepositoryInterface;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
 
 /**
- * Default provider cart.
+ * Default cart provider.
  *
  * @author Paweł Jędrzejewski <pawel@sylius.org>
  */
 class CartProvider implements CartProviderInterface
 {
     /**
-     * Cart context.
-     *
      * @var CartContextInterface
      */
-    protected $context;
-
-    /**
-     * Cart manager.
-     *
-     * @var ObjectManager
-     */
-    protected $manager;
+    protected $cartContext;
 
     /**
      * @var FactoryInterface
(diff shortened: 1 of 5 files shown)


src/Sylius/Bundle/CartBundle/Provider/CartProvider.php
```diff
@@ -48,11 +48,11 @@ class CartProvider implements CartProviderInterface
     protected $cartFactory;
 
     /**
-     * Cart repository.
+     * Cart Repository.
```
lchrusciel: Quite obvious, don't you think?

src/Sylius/Bundle/CartBundle/Provider/CartProvider.php
```diff
@@ -62,27 +62,26 @@ class CartProvider implements CartProviderInterface
...
+    public function __construct(
+        CartContextInterface $context,
+        ObjectManager $manager,
+        FactoryInterface $cartFactory,
+        RepositoryInterface $cartRepository,
+        EventDispatcherInterface $eventDispatcher
+    )
     {
         $this->context = $context;
         $this->manager = $manager;
```
pamil: `$this->cartManager = $cartManager` - to be consistent with other cart* variables?

src/Sylius/Bundle/CartBundle/Provider/CartProvider.php
```diff
@@ -118,7 +113,6 @@ public function getCart()
      */
     public function setCart(CartInterface $cart)
```
pamil: Should we have `setCart` method on a service that is meant to be stateless?

src/Sylius/Bundle/CartBundle/Provider/CartProvider.php
```diff
@@ -127,36 +121,36 @@ public function setCart(CartInterface $cart)
...
-     * @return CartInterface|null
+     * @return CartInterface
      */
-    protected function getCartByIdentifier($identifier)
+    private function provideCart()
     {
-        return $this->repository->find($identifier);
-    }
+        $cartIdentifier = $this->context->getCurrentCartIdentifier();
+        if ($cartIdentifier) {
```
lchrusciel: I am not sure about that, but I suppose it is possible to have `$cartIdenetifier` with value 0. In this case we would skip this statement. Nevertheless I think it is better to compare it directly with null, or any other value.

src/Sylius/Bundle/CartBundle/Provider/CartProvider.php
PWalkow: This one is all about saving current cart into the cartContext.

src/Sylius/Bundle/CartBundle/spec/Provider/CartProviderSpec.php
```diff
@@ -30,79 +33,100 @@ function let(
...
         FactoryInterface $cartFactory,
-        $eventDispatcher,
         CartInterface $cart
     ) {
         $context->getCurrentCartIdentifier()->willReturn(null);
         $cartFactory->createNew()->willReturn($cart);
-        $eventDispatcher->dispatch(SyliusCartEvents::CART_INITIALIZE, Argument::any())->shouldBeCalled();
+        $cartRepository->find()->shouldNotBeCalled();
+
+        $this->initialize_cart_should_be_called_on($eventDispatcher);
```
lchrusciel: Spec's are not the classes, so it is not a good practice to remove duplications. I found it less readable with this private method. And instead of reading what it should do, I have to jump to private declaration. And it is used only twice. Remove this method.

src/Sylius/Bundle/CartBundle/spec/Provider/CartProviderSpec.php
```diff
@@ -30,79 +33,100 @@ function let(
...
+        EventDispatcherInterface $eventDispatcher,
+        CartInterface $cart,
         FactoryInterface $cartFactory,
-        $repository,
-        $eventDispatcher,
         CartInterface $cart
     ) {
         $context->getCurrentCartIdentifier()->willReturn(7);
-        $repository->find(7)->shouldBeCalled()->willReturn(null);
+        $cartRepository->find(7)->shouldBeCalled()->willReturn(null);
```
lchrusciel: In this spec it is enough to only mock response for `$cartRepository`. We don't have to ensure, that this method was called.

PWalkow: the createNew() on FactoryInterface should be called only if there is no cart in RepositoryInterface. It's good to know that in this case we are actually calling find() on RepositoryInterface. Behaviour of this method is to actually check in the RepositoryInterface if the cart is there

PWalkow:
Ready.

src/Sylius/Bundle/CartBundle/Provider/CartProvider.php
peteward: When is this used? That might help name it.

peteward:
Feedback please, this is underpinning sorting out finances.

pjedrzejewski:
@peteward @PWalkow :+1: Thank you Piotr and Peter. :)

Issue #2980, opened 2015-07-08
Wrong currency when using stripe checkout


carlowens:
When trying to checkout a basket using a currency like GBP, when I get to the final payment page, the stripe dialogue box has the correct amount but always uses USD ($ in front of the amount).  I have GBP as my default currency within config.yml and three currencies configured within my admin (GBP, USD and EUR).

Not had a chance to see what the issue is yet but thought id put it on here incase its just something simple I am missing.

This is with version 0.14

michalmarcinkowski:
Should be fixed now.

PR #3641, opened 2015-11-26
fix for issue #2980 - using wrong total for the payment

carlowens:
Please test this.

Basically the issue for me was that using three currencies with GBP as the default, if someone was to checkout (using Stripe checkout JS), everything would look fine for the user in terms of the price, but the amount sent to stripe via SyliusPayumBundle was using the GBP value but in Euros.

I tested this myself using version 0.15 but I saw the same code is here in 0.16.

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/Action/CapturePaymentAction.php
+++ b/src/Sylius/Bundle/PayumBundle/Action/CapturePaymentAction.php
@@ -19,9 +19,23 @@ use Payum\Core\Request\Capture;
 use Payum\Core\Request\Convert;
 use Sylius\Bundle\PayumBundle\Request\GetStatus;
 use Sylius\Component\Core\Model\PaymentInterface as SyliusPaymentInterface;
+use Sylius\Component\Currency\Converter\CurrencyConverterInterface;
 
 class CapturePaymentAction extends GatewayAwareAction
 {
+    /**
+     * @var CurrencyConverterInterface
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
Nice, thanks for the patch Carl! It looks good to me, although I am wondering if we should apply this change also to https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/PayumBundle/Action/ConvertPaymentToPaypalExpressAction.php#L28 and other actions. What do you think? @makasim how does it look to you? I think it is OK.

makasim:
Cannot say anything, I bit out of scope. What does converter? Does it converts money from EUR to GBP?

pjedrzejewski:
@makasim Yeah, basically it converts amount in base currency to target currency. So if your base currency is EUR, then it converts it to GBP. (if that's order currency) All payment apis need to get the amount it their desired currency, so yeah, I think this is a good change and we should apply it to other actions too.

pjedrzejewski:
@carlowens Would you find time to apply it to other actions too? I don't think Capture will work fine if Convert actions are not using right amounts.

makasim:
Sounds good!

carlowens:
Yes I will take a look when I get a chance.

michalmarcinkowski:
Thanks for reporting Carl! Fixed already 😉

carlowens:
👍

PR #3643, opened 2015-11-27
[Travis] Javascript suites on every PHP version, lazy cache:warmup

pamil:
No BC breaks :)

 - [x] Deleted PHP 5.6 Javascript job on Travis
 - [x] Added Javascript Behat suites on every PHP version
 - [x] Made `cache:warmup` command not to require database schema
 - [x] Solved issue with storing new cache on every build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,9 @@ before_script:
 script:
     - travis/run-tests default
 
+before_cache:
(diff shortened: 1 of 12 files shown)


pjedrzejewski:
Thank you Kamil! :+1:

PR #3649, opened 2015-11-29
[Rbac][RFC] Use symfony auth checker by tweaking role voter rather than implementing our own

cdaguerre:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | #2574 |
| License | MIT |
| Doc PR |  |

This PR suggests tweaking the Symfony `RoleVoter` rather than implementing our own.

By default, the Symfony `SecurityBundle` ships with two security voters to handle roles. Depending on the `security.roles_hierarchy` setting, it enables either the [`RoleVoter`](https://github.com/symfony/security/blob/master/Core/Authorization/Voter/RoleVoter.php) or the [`RoleHierarchyVoter`](https://github.com/symfony/security/blob/master/Core/Authorization/Voter/RoleHierarchyVoter.php).
This PR enables use of Sylius RBAC by slightly tweaking the `RoleHierarchyVoter` providing him with the roles hierarchy he needs either from `RbacBundle` configuration or from Database, by introducing [`RoleHierarchyProvider`s](https://github.com/cdaguerre/Sylius/tree/rbac-voter-poc/src/Sylius/Bundle/RbacBundle/Security/Role/Provider).

This PR includes the following providers:
- [**ArrayHierarchyProvider**](https://github.com/cdaguerre/Sylius/blob/rbac-voter-poc/src/Sylius/Bundle/RbacBundle/Security/Role/Provider/ArrayHierarchyProvider.php): The Rbac bundle configuration is transformed to a role hierarchy the voter understands at container compilation time, stored as a container parameter and injected to this provider that simply passes it on to the voter.
- [**RepositoryHierarchyProvider**](https://github.com/cdaguerre/Sylius/blob/rbac-voter-poc/src/Sylius/Bundle/RbacBundle/Security/Role/Provider/RepositoryHierarchyProvider.php): Generates the role hierarchy from the database using the `RoleRepository` and the `PermissionRepository`
- [**FallbackHierarchyProvider**](https://github.com/cdaguerre/Sylius/blob/rbac-voter-poc/src/Sylius/Bundle/RbacBundle/Security/Role/Provider/FallbackHierarchyProvider.php): Takes two providers: Asks first provider for the role hierarchy and falls back to the second one if the first one fails, (eg. the repository provider fails if rbac is not initialized)
- [**CachedHierarchyProvider**](https://github.com/cdaguerre/Sylius/blob/rbac-voter-poc/src/Sylius/Bundle/RbacBundle/Security/Role/Provider/CachedHierarchyProvider.php): Adds caching to any other provider

By default, the `RbacBundle` is configured to use database driven role hierarchy and falls back to the configured one if it is not initialized. The good thing about this approach is that you don't need to bother whether Rbac is initialized or not at runtime. It simply falls back to bundle configuration.

Note that the `RbacBundle` doesn't rely on any of the `Rbac` component classes so it could be used without it if you don't need database driven role hierarchy.

What do you think?


src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius.yml
```diff
@@ -295,442 +295,507 @@ sylius_rbac:
...
+            description: Manage contact requests
+            child_permissions: [sylius.contact_request.show, sylius.contact_request.index, sylius.contact_request.create, sylius.contact_request.update, sylius.contact_request.delete]
+        sylius.manage.contact_topic:
+            description: Manage contact topics
+            child_permissions: [sylius.contact_topic.show, sylius.contact_topic.index, sylius.contact_topic.create, sylius.contact_topic.update, sylius.contact_topic.delete]
+        sylius.manage.settings:
+            description: Manage settingss
+            child_permissions: [sylius.settings.general, sylius.settings.security, sylius.settings.taxation]
+
+        sylius.catalog:
```
peteward: Should these have the 'manage' prefix as well? or some other prefix if they're the level above?

peteward:
Looks really interesting. Going to ask @gonzalovilaseca to take a look at this tomorrow as this might help him out...

isometriks:
Nice. Should we also think about doing something with this line? https://github.com/cdaguerre/Sylius/blob/rbac-voter-poc/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php#L563

If you aren't using RBAC and want to use a regular broader role, say something like ROLE_ADMIN_BLOG to be able to use a blog controller as a whole, then this would turn your role into `application_name.blog.ROLE_ADMIN_BLOG`. It would obviously create much more verbose permissions in the routing files to just use the permission straight up, don't know if that's a big issue or not but that would create the most flexibility. I guess you could also create another protected method in the controller to return the permission string, that way in SettingsController you could just override it that one method instead of copying the whole grantor403. Although I guess you could check for existence of `ROLE_` at the beginning. The default symfony role voter defaults to `ROLE_` to be used, but does have an optional constructor. There is no configurable parameter but it would be possible to change with a compiler pass.

src/Sylius/Bundle/RbacBundle/Security/Role/Inflector.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    const DEFAULT_PREFIX = 'ROLE_';
+
+    /**
+     * @var string
+     */
+    protected $prefix;
+
+    /**
+     * Constructor.
```
pamil: I guess it's not needed to explain this fact :)

src/Sylius/Bundle/RbacBundle/Security/Role/Provider/ArrayHierarchyProvider.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+class ArrayHierarchyProvider implements HierarchyProviderInterface
+{
+    /**
+     * @var array
+     */
+    protected $map;
+
+    /**
+     * Constructor.
```
pamil: Same here :)

src/Sylius/Bundle/RbacBundle/Security/Role/Provider/CachedHierarchyProvider.php
```diff
@@ -0,0 +1,74 @@
...
+    protected $ttl;
+
+    /**
+     * Constructor.
+     *
+     * @param HierarchyProviderInterface $provider
+     * @param Cache                      $cache
+     * @param int                        $ttl
+     */
+    public function __construct(HierarchyProviderInterface $provider, Cache $cache, $ttl = self::TTL)
```
pamil: `$provider` -> `$hierarchyProvider`, `$decoratedHierarchyProvider`, WDYT?

src/Sylius/Bundle/RbacBundle/Security/Role/Provider/FallbackHierarchyProvider.php
```diff
@@ -0,0 +1,52 @@
...
+     */
+    protected $fallback;
+
+    /**
+     * Constructor.
+     *
+     * @param HierarchyProviderInterface $default
+     * @param HierarchyProviderInterface $fallback
+     */
+    public function __construct(HierarchyProviderInterface $default, HierarchyProviderInterface $fallback)
```
pamil: `$default` -> `$defaultProvider`, `$defaultHierarchyProvider`? Same for `$fallback`.

src/Sylius/Bundle/RbacBundle/Security/Role/Provider/RepositoryHierarchyProvider.php
```diff
@@ -0,0 +1,152 @@
...
+
+        foreach ($this->sort($roles) as $role) {
+            $this->addRole($role);
+        }
+
+        return $this->map;
+    }
+
+    /**
+     * Add permission to map.
```
pamil: We can remove this comment by renaming method name to `addPermissionToMap` :)

features/backend/shipments.feature
```diff
@@ -20,7 +20,7 @@ Feature: Shipments
             | customer          | address                                                | shipment |
             | klaus@example.com | Klaus Schmitt, Heine-Straße 12, 99734, Berlin, Germany | FedEx    |
             | lars@example.com  | Lars Meine, Fun-Straße 1, 90032, Vienna, Austria       | DHL      |
-          And I am logged in as administratorzs
+          And I am logged in as administrator
```
pamil: Ouch, random characters in the code are my superpower :)

aramalipoor:
:+1: for this. We're using User, Rbac and Resource in a multi-tenant application and we should work with `security.authorization_checker` but now we're tied to use Sylius's one.

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
```diff
@@ -562,7 +562,7 @@ protected function isGrantedOr403($permission)
         if ($permission) {
             $grant = sprintf('%s.%s.%s', $this->config->getBundlePrefix(), $this->config->getResourceName(), $permission);
 
-            if (!$this->get('sylius.authorization_checker')->isGranted($grant)) {
+            if (!$this->get('security.authorization_checker')->isGranted($grant)) {
```
aramalipoor: I think we should pass `$resource` here, if it's available, so that ACL voters work properly.

liverbool:
:+1:

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
cdaguerre: @aramalipoor I agree, but I was waiting for feedback from @pjedrzejewski. Pawel, what do you think? I'm afraid this work could become outdated quite quickly...

liverbool:
@pjedrzejewski, How about milestone of this PR?

/:

steffenbrem:
Really like this PR, would make handling security a lot easier :+1:  Currently in need to use my own voters, but cannot do that without extending the Resource Controller, right?

pjedrzejewski:
I can't promise but I will do my best to check this out properly after 0.17.0 is tagged very soon (tm).

oatto:
@pjedrzejewski, I'm waiting...

okwinza:
ping (tm)

steffenbrem:
@cdaguerre Could you rebase this? Hopefully it will get merged soon! 😄

peteward:
is this still intended for 0.18?

pjedrzejewski:
@peteward As much as I hate to say that, I doubt we get that in v0.18. I just briefly looked at the code and there are still some issues & BC breaks involved. One example is coupling between User & Rbac (see the command).

peteward:
ok np, just checking, thanks.

pamil:
Unfortunately, there's no Rbac bundle since v1.0.0-alpha.1

PR #3651, opened 2015-11-30
[Cart] Fixed and polished CartProviderSpec on PHP7

pamil:
Build was failing on PHP7 due to doubled `$cart` in methods definitions.
Fixed formatting, renamed `$context` to `$cartContext`, changed order in specification methods to be the same as in `let`.


pjedrzejewski:
Thank you Kamil!

I think we should bring PHP7 back to the required builds. Wdyt?

pamil:
We may go for it since PHP7 builds are passing :) I'll prepare a PR for that.

PR #3656, opened 2015-11-30
CartBundle: fix for missing identifiers in cartContext

PWalkow:
| Q             | A
| ------------- | ---
| Bug fix?      | [yes]
| New feature?  | [no]
| BC breaks?    | [no]
| Deprecations? | [no]
| License       | MIT

We were 'saving' carts even if there was no clear indication which cart has been saved because of non-persisted carts in that case we were creating plenty of instances of carts and overriding these by each other in CartContext. It is impossible now to save a cart in CartContext if there is no identifier.

### DIFF
--- a/src/Sylius/Bundle/CartBundle/Provider/CartProvider.php
+++ b/src/Sylius/Bundle/CartBundle/Provider/CartProvider.php
@@ -11,6 +11,7 @@
 
 namespace Sylius\Bundle\CartBundle\Provider;
 
+use Doctrine\Common\Persistence\ObjectManager;
 use Sylius\Component\Cart\Context\CartContextInterface;
 use Sylius\Component\Cart\Model\CartInterface;
 use Sylius\Component\Cart\Provider\CartProviderInterface;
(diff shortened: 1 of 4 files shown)


peteward:
What about the getCurrentCart() / ensureCurrentCart() distinction?

src/Sylius/Bundle/CartBundle/Resources/config/services.xml
```diff
@@ -88,6 +88,7 @@
             <argument type="service" id="sylius.factory.cart" />
             <argument type="service" id="sylius.repository.cart" />
             <argument type="service" id="event_dispatcher" />
+            <argument type="service" id="doctrine.orm.entity_manager" />
```
pamil: `sylius.manager.cart`?

src/Sylius/Bundle/CartBundle/Provider/CartProvider.php
```diff
@@ -57,13 +63,15 @@ public function __construct(
         CartContextInterface $cartContext,
         FactoryInterface $cartFactory,
         RepositoryInterface $cartRepository,
-        EventDispatcherInterface $eventDispatcher
+        EventDispatcherInterface $eventDispatcher,
+        EntityManagerInterface $entityManager
```
pamil: We can use `ObjectManager` instead.

src/Sylius/Bundle/CartBundle/Provider/CartProvider.php
```diff
@@ -130,6 +138,15 @@ private function provideCart()
...
 
         $cart = $this->cartFactory->createNew();
+
+        $event = new GenericEvent($cart);
+        $this->eventDispatcher->dispatch(SyliusCartEvents::CART_INITIALIZE, $event);
+
+        $cart = $event->getSubject();
+
+        $this->entityManager->persist($cart);
+        $this->entityManager->flush();
```
pamil: I know cart identifier, but why don't create one so we don't have to flush database here?

PWalkow: Identifier is... 
```php
    public function getIdentifier()
    {
        return parent::getId();
    }

Comment (peteward) [inline src/Sylius/Bundle/CartBundle/Provider/CartProvider.php]: We talked about using a separate sequence but that seemed like overkill. The way this all works kinda needs this id to exist, open to other ideas here, but unless there's a major upheaval of structure you really need to flush.

**Comment by PWalkow** [comment]
@pamil 
I know we need cart identifier, but why don't create one so we don't have to flush database here?
@pwalkow
Identifier is... 
```php
    public function getIdentifier()
    {
        return parent::getId();
    }
```
so without saving it to db we have no ID.

@peteward 
We talked about using a separate sequence but that seemed like overkill. The way this all works kinda needs this id to exist, open to other ideas here, but unless there's a major upheaval of structure you really need to flush.

Edit: I kept it here because it went missing after update of the code.

PWalkow:
@pjedrzejewski ping, ready

michalmarcinkowski:
Closing since this class no longer exist (https://github.com/Sylius/Sylius/issues/5565).

PR #3658, opened 2015-12-01
[WIP] Removal of calculateTotal on Order and OrderItem to getTotal

PWalkow:


src/Sylius/Component/Order/Model/OrderItem.php
```diff
@@ -245,6 +247,8 @@ public function getTotal()
      */
     public function setTotal($total)
     {
+        throw new \RuntimeException('this method should not exists');
```
pjedrzejewski: Haha, let's remove it. :+1:

pjedrzejewski:
We should add it to all methods which modify the total. Not the ``getTotal``. That could still give you invalid totals when someone does not call it.

We should add it to ``add/removeItem``, ``add/removeAdjustment`` etc.

PWalkow:
Unfortunately it's not that easy as it should be. If there is 'calculateTotal' anywhere in the order, even on the 'getTotal' this one is failing:
```
features/pricing/group_based_pricing.feature:69
```
Try it yourself, fellow developers!

Solution for that would be saving the cart... during the login(?) Any ideas are welcome...

peteward:
@pjedrzejewski Close this one, it should be looked at later, but needs more thought

PR #3660, opened 2015-12-01
[Composer] Added explicit dependency symfony/proxy-manager-bridge

pamil:
It looks like Symfony 2.8 does not include `symfony/proxy-manager-bridge` and changes made in #3643 related to lazy `cache:warmup` aren't working in that case.

Also moved `cache:warmup` above creating schema in Travis, to ensure that cache warming up does not require database schema.


pjedrzejewski:
Thanks Kamil!

PR #3663, opened 2015-12-02
[WIP] Rework promotion rule checkers to allow promotion repeating

stloyd:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | kinda* |
| Deprecations? | yes |
| License | MIT |

The only BC break in current state is that `isEliglibe()` will no longer return `bool` value.

### DIFF
--- a/src/Sylius/Bundle/PromotionBundle/Form/Type/Rule/ItemCountConfigurationType.php
+++ b/src/Sylius/Bundle/PromotionBundle/Form/Type/Rule/ItemCountConfigurationType.php
@@ -44,12 +44,16 @@ class ItemCountConfigurationType extends AbstractType
                     new Type(array('type' => 'numeric')),
                 ),
             ))
-            ->add('equal', 'checkbox', array(
+            ->add('equal', 'choice', [
                 'label' => 'sylius.form.rule.item_count_configuration.equal',
-                'constraints' => array(
-                    new Type(array('type' => 'bool')),
-                ),
-            ))
+                'choices' => [
+                    'sylius.form.rule.item_count_configuration.equal_or_more' => 'equal',
+                    'sylius.form.rule.item_count_configuration.more_than' => 'more_than',
+                    'sylius.form.rule.item_count_configuration.exactly' => 'exactly',
+                    'sylius.form.rule.item_count_configuration.repeatable' => 'modulo',
+                ],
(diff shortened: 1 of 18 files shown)


PWalkow:
Some additional behat testing would be great to check this 'repeat' promotion. I wonder how the 'revert' would work on these.

pamil:
Why does `isEligible` return `int` instead of `bool`?

stloyd:
Because it now will return how many times this promotion is eliglible, but in fact that name is quite wrong after this change (don't have better for now), cause it will return information from rules (check `modulo` option for `ItemCountRuleChecker`) how many times this promotion can be applied.

pjedrzejewski:
Could you please give us at least few words about why changing this?

stloyd:
Without this change promotion processor will only know that it eliglibe, but this is very limiting cause in real world you want to know how many times this one promotion can be applied.

Promotion example:
```feature
    Background:
        Given store has default configuration
          And the following products exist:
            | name       | price |
            | Coca-cola  | 10    |
            | Pepsi      | 9     |
          And the following promotions exist:
            | name                             | description                                         |
            | Buy 3 Coca-cola get 4th for free | If you buy 3 Coca-cola's, you will get 4th for free |
          And promotion "Buy 3 Coca-cola get 4th for free" has following rules defined:
            | type       | configuration           |
            | Item count | count: 3, equal: modulo |
          And promotion "Buy 3 Coca-cola get 4th for free" has following actions defined:
            | type        | configuration                        |
            | Add product | variant:Coca-cola,quantity:1,price:0 |
          And all products are assigned to the default channel
          And all promotions are assigned to the default channel

    Scenario: Free product is applied when the cart has the required count of items
        Given I am on the store homepage
         When I add product "Coca-cola" to cart, with quantity "3"
         Then I should be on the cart summary page
          And I should see 4 item in the list
          And "Grand total: €30.00" should appear on the page

    Scenario: Free product is applied when the cart has more than required count of items
        Given I am on the store homepage
         When I add product "Coca-cola" to cart, with quantity "6"
         Then I should be on the cart summary page
          And I should see 8 item in the list
          And "Grand total: €60.00" should appear on the page
```

Of course action should reflect few changes in logic to allow setting multiple items on cart etc.

PR #3677, opened 2015-12-04
Update action refreshes the resource when updated

### LINKED ISSUE #3676: [ResourceBundle] update action in resource controller do not check for stopped events
Hello everyone,

The problem is that I created an event listener for an event like sylius.order.pre_create, I stopped the event propagation but it didn't work.

I followed the trace and checked that `$this->domainManager->update($resource);` is not returning to updated resourcer or the event in case it was stopped.

I will shortly sent a PR with the required change, though I don't know if it will have any side effect.

aleherse:
Fixes the situation in which the domain manager returns an event because the propagation was stopped.

Closes #3676


pjedrzejewski:
Woah, good catch. Thank you very much Alejandro! :+1:

Issue #3671, opened 2015-12-03
[RFC] [ChannelBundle] Channel should have a defaultLocale field


axelvnk:
Now, when you open a storefront through its domain, it automatically uses the sylius_locale parameter. However, if that locale is not part of the enabled locales for that channel, it doesn't make any sense to load the storefront with that locale enabled... Even if the default locale is also enabled for that channel, a defaultlocale field would be nice, for more control.

I would be happy to do this myself and create a pull request for us to discuss.

pjedrzejewski:
Agree, I think we should also have default Currency on every channel.

PR #3681, opened 2015-12-04
#3671 add defaultLocale and defaultCurrency to Channel model

axelvnk:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3671 
| License       | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Behat/CoreContext.php
+++ b/src/Sylius/Bundle/CoreBundle/Behat/CoreContext.php
@@ -72,7 +72,9 @@ class CoreContext extends DefaultContext
         $channel->setName('Default');
         $channel->setUrl('http://example.com');
         $channel->addCurrency($currency);
+        $channel->setDefaultCurrency($currency);
         $channel->addLocale($locale);
+        $channel->setDefaultLocale($locale);
         $manager->persist($channel);
 
         $manager->flush();
--- a/src/Sylius/Bundle/CoreBundle/Form/Type/ChannelType.php
+++ b/src/Sylius/Bundle/CoreBundle/Form/Type/ChannelType.php
@@ -37,9 +37,15 @@ class ChannelType extends BaseChannelType
                 'label'    => 'sylius.form.channel.locales',
                 'multiple' => true
             ))
+            ->add('defaultLocale', 'sylius_locale_choice', array(
+                'label'    => 'sylius.form.channel.locale_default'
+            ))
             ->add('currencies', 'sylius_currency_choice', array(
                 'label'    => 'sylius.form.channel.currencies',
                 'multiple' => true
+            ))
+            ->add('defaultCurrency', 'sylius_currency_choice', array(
+                'label'    => 'sylius.form.channel.currency_default'
             ))
             ->add('shippingMethods', 'sylius_shipping_method_choice', array(
                 'label'    => 'sylius.form.channel.shipping_methods',
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/Channel.orm.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/Channel.orm.xml
@@ -17,6 +17,12 @@
                                       http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
 
     <mapped-superclass name="Sylius\Component\Core\Model\Channel" table="sylius_channel">
+        <many-to-one field="defaultLocale" target-entity="Sylius\Component\Locale\Model\LocaleInterface">
+            <join-column name="default_locale_id" referenced-column-name="id" nullable="true" on-delete="SET NULL"/>
+        </many-to-one>
+        <many-to-one field="defaultCurrency" target-entity="Sylius\Component\Currency\Model\CurrencyInterface">
+            <join-column name="default_currency_id" referenced-column-name="id" nullable="true" on-delete="SET NULL"/>
+        </many-to-one>
         <many-to-many field="currencies" target-entity="Sylius\Component\Currency\Model\CurrencyInterface">
             <join-table name="sylius_channel_currencies">
                 <join-columns>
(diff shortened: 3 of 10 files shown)


axelvnk:
@pjedrzejewski I think we should add a listener on kernel.request that sets locale and currency for the session. But not sure where I should put this. CoreBundle would make the most sense I guess.

axelvnk:
Testing in PHP7 fails, bin/behat --strict -f progress -p javascript_cached is not executed...

pjedrzejewski:
I restarted the build, if this does not help, please rebase with master, there have been some tweaks to these scripts.

axelvnk:
@pjedrzejewski Awesome, restarting the build seems to solve it.

src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/Channel.orm.xml
```diff
@@ -17,6 +17,12 @@
                                       http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
 
     <mapped-superclass name="Sylius\Component\Core\Model\Channel" table="sylius_channel">
+        <many-to-one field="defaultLocale" target-entity="Sylius\Component\Locale\Model\LocaleInterface">
+            <join-column name="default_locale_id" referenced-column-name="id" nullable="true" on-delete="SET NULL"/>
+        </many-to-one>
+        <many-to-one field="defaultCurrency" target-entity="Sylius\Component\Currency\Model\CurrencyInterface">
+            <join-column name="default_currency_id" referenced-column-name="id" nullable="true" on-delete="SET NULL"/>
+        </many-to-one>
```
axelvnk: I made the fields nullable, although I'm not sure this is the way to go. How currencies/locales are set up right now doesn't allow me to make these not nullable. 

But perhaps we can fallback on the first locale/currency from the available ones that are set for the channel when the defaults are null. And if no locales/currencies are added we can fallback on the applications default locale/currency.

src/Sylius/Bundle/CoreBundle/Form/Type/ChannelType.php
```diff
@@ -37,10 +37,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'label'    => 'sylius.form.channel.locales',
                 'multiple' => true
             ))
+            ->add('defaultLocale', 'sylius_locale_choice', array(
+                'label'    => 'sylius.form.channel.locale_default',
+                'multiple' => false
```
pjedrzejewski: This is default value, no need to specify it here.

src/Sylius/Bundle/CoreBundle/Form/Type/ChannelType.php
```diff
@@ -37,10 +37,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+            ))
             ->add('currencies', 'sylius_currency_choice', array(
                 'label'    => 'sylius.form.channel.currencies',
                 'multiple' => true
             ))
+            ->add('defaultCurrency', 'sylius_currency_choice', array(
+                'label'    => 'sylius.form.channel.currency_default',
+                'multiple' => false
```
pjedrzejewski: Same as above. ^

src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/Channel.orm.xml
pjedrzejewski: For now let's keep them nullable, but in the end, they should not be nullable. But that requires a bit more work, we will get there. :)

src/Sylius/Component/Core/Model/Channel.php
```diff
@@ -69,6 +79,38 @@ public function __construct()
...
      */
+    public function getDefaultCurrency()
+    {
+        return $this->defaultCurrency;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setDefaultCurrency($defaultCurrency)
```
pjedrzejewski: Missing typehints.

src/Sylius/Component/Core/Model/ChannelInterface.php
```diff
@@ -38,4 +40,23 @@
...
+
+    /**
+     * @return LocaleInterface
+     */
+    public function getDefaultLocale();
+
+    /**
+     * @param CurrencyInterface $currency
+     */
+    public function setDefaultCurrency($currency);
```
pjedrzejewski: Missing typehints.

pjedrzejewski:
Have a look at my comments, also, please squash commits, 11 is a bit too much for such simple addition. :) Thanks!

axelvnk:
@pjedrzejewski done

pjedrzejewski:
@axelvnk Sorry, I forgot to mention it earlier - please generate a migration for the database. Thanks for your time Axel!

axelvnk:
@pjedrzejewski Hey is it possible someone forgot to generate migrations for their changes? Because I only have 5 query statements that differ from the currrent schema but the diff is generating 8 more than expected. See screenshot below :

screen shot 2015-12-08 at 10 10 50

pjedrzejewski:
Can be but I see that you have associations which are not in core yet. :) Are you sure that these queries do not come from your custom code?

axelvnk:
Omfg, you are right :') haha. Now that was dumb. I'll get rid of it and commit. Do you want me to squash again?

pjedrzejewski:
@axelvnk Yes, please. :)

axelvnk:
Okay should be all good now..

pjedrzejewski:
Perfect, thanks again Axel! :+1:

src/Sylius/Bundle/CoreBundle/Behat/CoreContext.php
```diff
@@ -72,7 +72,9 @@ public function storeHasDefaultConfiguration()
         $channel->setName('Default');
         $channel->setUrl('http://example.com');
         $channel->addCurrency($currency);
```
pamil: Btw. shouldn't currency and locale be added automatically when setting them as default?

PR #3682, opened 2015-12-06
[RFC] Create the "etc" directory for environment specific configuration.

adamelso:
With the introduction of AppVeyor continuous integration and improved setup of Travis, the root directory may start to become disorganised or cluttered, especially if there are more tools to be integrated in the future.

Moving files for AppVeyor, Travis and Vagrant into an "etc" directory will help establish a new location for similar files. For example, I prefer to use Ansible instead of the other the other tools.

By the way, what happened to the Capistrano / Capifony files? I can't seem the find any issues or pull requests related to its removal.


pjedrzejewski:
I think we should have:

``ci/travis``
``ci/appveyor``
``vagrant``
``docker`` (#3432 )

In the main directory. I think it is pretty clean structure, wdyt?

pamil:
@pjedrzejewski I like the idea to group CI configs. We can do the same for `vagrant` and `docker` directories as well.

pjedrzejewski:
@pamil Definitely but I miss a proper name for these. :)

pamil:
@pjedrzejewski seems like `etc` is the way to go :tada:

pjedrzejewski:
@pamil :+1: @adamelso Do you have time to update the PR? That would be awesome, thank you!

adamelso:
@pjedrzejewski can do this evening . What should the final structure be?

Reason for putting them all in one directory is that the same tools used to provision a local machine or a production server might be used to provision a CI platform . "Etc" simply represents environment configuration.

Also are you guys familiar with the new Symfony 3 directory structure ? Just extending on it with this PR :)

adamelso:
Closing in favour of #3711

PR #3686, opened 2015-12-06
Fix password reset forms

inssein:
The reset password page was referring to 'newPassword' which was never defined anywhere.

This should probably be tested, but I am not very familiar with Behat / BDD yet. Perhaps if I get some time tonight I'll try taking a stab at it.


inssein:
@pjedrzejewski not sure what happened to the PHP 7 tests there. Is there a way I can re-run them?

pjedrzejewski:
Thank you Hussein!

Issue #3368, opened 2015-09-28
[RFC] Promotions Enhancements - Benefits & Filters


peteward:
This RFC describes a suggested approach to improving the promotions system within Sylius to cover a much wider set of use-cases than it currently does.

For a longer description of how these limitations were discovered see #1810, but to summarise - Currently there’s no way of filtering the target of your promotion logic beyond being applied to the whole order. You cannot apply it only on certain OrderItems that match a criteria.

We spend a lot of time last week researching possible permutations of promotions along with a few whiteboard sessions. Our resulting approach is based on these permutations and trying to model them from the admin users’ perspective whilst respecting the existing code as far as possible. No serious code has been written yet, we wanted to put it out for feedback first.

Firstly here are the scenarios that we looked at trying to model (all starred ones cannot be achieved with the current functionality):
- Get 20% off your first order
- Get 20% off if spend more than 200
- Get a free gift if you buy at least one dress
- Get a free gift for every dress you buy*
- Get a 10% off jacket if you buy trousers*
- Get free shipping if you spend more than 500
- Get a free gift card and 20% off dresses if you buy dress*
- Get 10% off any sale dresses*
- Get 10£ off suits if you spend more than 500 and get free shipping*
- Get 20% off suits if you spend more than 500 and get 50£ off order*
- Get 10% off bow tie and belts if you buy jackets or trousers*
- Get 10% off on dresses if these costs more than 300 and get 15 off jackets if these costs more less than 250 when you place first order*

The easiest way to explain the proposal is visually. These visuals are not final and after discussing with a UX designer they could definitely be simplified, but will hopefully make sense in the following format to those familiar with the existing engine:

![screen shot 2015-09-28 at 15 07 05](https://cloud.githubusercontent.com/assets/7068515/10139387/8c62dfda-65fa-11e5-9316-ab8ba1d09e12.png)

This mock-up screenshot models a very complex promotion.
### Explanation:

Rules still exist, they decide whether your basket qualifies for the promotion and will be run in the same way. Your order must pass all rules.

Actions have changed and now consist of 2 parts - Benefits and Filters.
Benefits are what we previously called Actions and reflect what the customer gets from this promotion. 
Filters decides which of the PromotionSubject components (OrderItems) are valid targets for that Benefit. It could be a very simple promotion where the target is the ‘Whole Order’ or a subset of the OrderItems defined by some logic.
It’s possible that a single Benefit could require multiple Filters that are applied together (AND).
[...]

PR #3689, opened 2015-12-08
[Promotion] Rework rule checkers for count & total to be more flexible

stloyd:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| License | MIT |

### DIFF
--- a/src/Sylius/Bundle/PromotionBundle/Form/Type/Rule/ItemCountConfigurationType.php
+++ b/src/Sylius/Bundle/PromotionBundle/Form/Type/Rule/ItemCountConfigurationType.php
@@ -34,11 +34,13 @@ class ItemCountConfigurationType extends AbstractType
                     new Type(array('type' => 'numeric')),
                 ),
             ))
-            ->add('equal', 'checkbox', array(
-                'label' => 'sylius.form.rule.item_count_configuration.equal',
-                'constraints' => array(
-                    new Type(array('type' => 'bool')),
-                ),
+            ->add('equal', 'choice', array(
+                'label' => 'sylius.form.rule.equal',
+                'choices' => [
+                    'equal' => 'sylius.form.rule.equal_or_more',
+                    'more_than' => 'sylius.form.rule.more_than',
+                    'exactly' => 'sylius.form.rule.exactly',
+                ],
             ))
         ;
     }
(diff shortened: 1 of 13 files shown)


stloyd:
Something missing @pjedrzejewski @michalmarcinkowski ?

pjedrzejewski:
Equal with values ``equal``, ``more_than`` etc. does not sound right ... We will need different comparators in many other rules, so would be good to standardize this right now.

stloyd:
I know naming is not best, but still better than `true/false` :wink:

stloyd:
Also I was thinking about simple trait for such comparsion, which would unify other places too.

stloyd:
@pjedrzejewski WDYT?

stloyd:
@pjedrzejewski @michalmarcinkowski ping pong poke... :)

src/Sylius/Bundle/PromotionBundle/Form/Type/Rule/ItemCountConfigurationType.php
```diff
@@ -34,11 +34,13 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                     new Type(array('type' => 'numeric')),
                 ),
             ))
-            ->add('equal', 'checkbox', array(
-                'label' => 'sylius.form.rule.item_count_configuration.equal',
-                'constraints' => array(
-                    new Type(array('type' => 'bool')),
-                ),
+            ->add('equal', 'choice', array(
```
lchrusciel: Please be consistent ;) Once you use `array(` (see below) once you put `[` :)

src/Sylius/Component/Promotion/Checker/CountComparisonTrait.php
```diff
@@ -0,0 +1,47 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Promotion\Checker;
+
+trait CountComparisonTrait
+{
+    /**
+     * @param int    $count
```
lchrusciel: We are not aligning params vertically any more.

pjedrzejewski:
@stloyd We do not want to touch these atm as I am not entirely sure about this solution and we will be doing some work based on #3368.

stloyd:
@pjedrzejewski Please wait a week or less with approach based on #3368, I have done something almost identical for company I work now & must talk about open-sourcing the code (we use only API so there is no frontend for it unfortunatelly ;)), and as well this code is one of the changes done for that approach (just extracted).

PR #3690, opened 2015-12-08
Responses generated by ProcessValidators now get processed.

fred-jan:
ProcessValidators return ActionResult instances by default, which needs to be transformed to regular responses by the Coordinator.


pjedrzejewski:
Sounds good, thank you @fred-jan!

PR #3691, opened 2015-12-08
Updated PHPDoc to allow ProcessValidatorInterface instances to be passed

fred-jan:
The ProcessBuilder:validate implementation supports ProcessValidatorInterface instances to be passed, but the PHPDoc @param declaration in it's interface said otherwise.


pjedrzejewski:
Nice, thank you @fred-jan! :+1:

PR #3692, opened 2015-12-08
[Promotion] Introduce code for Promotion

tuka217:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/app/migrations/Version20151210134607.php
+++ b/app/migrations/Version20151210134607.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Sylius\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20151210134607 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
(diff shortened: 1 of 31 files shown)


peteward:
Doesn't code become confusing with coupon codes? I assumed that's why there weren't already there.

pjedrzejewski:
@tuka217 Nice work, please add a database migration and add a PR to the docs.

@peteward Can be a bit confusing, but I think it is essential for all imports/exports to identify a promotion with a unique code. Also, helps if you have some custom logic in your code specific to a promotion (should not be done, although gives you a lot of flexibility if you have to do something like this in a short timeframe)

Unfortunately I do not see way for avoiding confusion other than documenting fields well in the docs. Saying that this code is a "machine name" for promotion and should not be confused with codes on the promotion coupon entity.

stloyd:
@pjedrzejewski Personally I get `code` quite confusing, IMO those all PRs were adding `identifiers` or `internal identifiers/codes`.

pjedrzejewski:
I don't see how ``code`` is confusing except for the Promotion example. I'd say if we document it well, there will be no problem. Identifier would be confused with the id and redundant.

peteward:
There is now a consistency with the term `code` within Sylius, with the recent PR's I think it's clear enough from a code architecture. Some though might need to be given to avoiding confusion to the admin user though so that they know what to expect.

pjedrzejewski:
@tuka217 Could you take a look at the failing build?

pjedrzejewski:
Thank you Ania! :+1:

PR #3704, opened 2015-12-10
CoreBundle: checker for promotion coupons added

PWalkow:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | ~
| License       | MIT

Allows to use form extension to modify 'sylius_cart' if promotion coupons are removed from the form itself. It is throwing 'A new entity was found through the relationship' because we are having detached entity from form.


pjedrzejewski:
Makes sense, thank you @PWalkow!

PR #3705, opened 2015-12-10
Missing menu translations and alphabetized

peteward:
There's another issue here of double-translations because translate() is called in the MenuBuilder (for separation of messages into menu domain) and also when it's rendered, so you get a lot of translation errors on the backend.

I might look at this again soon but here's the missing messages to start with.


stloyd:

pjedrzejewski:
Thanks Peter!

PR #3520, opened 2015-11-02
Adjustments on inventory unit

PWalkow:
Moving Adjustments from OrderItem to inventoryUnit level adds possibility to apply taxes on InventoryUnit, so far I moved taxed.

In the future it makes process of refund easier - calculation of taxes and promotions is easier to manage.

Work done:
- orderItem is no longer Ajdustable, InventoryUnit is Adjustable instead
- removed 'calculateOrder' from all of this usage and put it back into the model.
- added TaxationProcessorSpecs to actually check how this class is working now.
- simplified the cart form on summary page to use values from form not cart passed to the view.
- added refreshment of cart to 'cart-change' event to add also the InventoryUnits which were not loaded at this stage.

| Q             | A
| ------------- | ---
| Bug fix?      | [no]
| New feature?  | [yes]
| BC breaks?    | [yes, adjustments cannot be applied on OrderItem anymore]
| Deprecations? | [yes]
| Fixed tickets | [https://github.com/Sylius/Sylius/issues/3441]
| License       | MIT
| Doc PR        | [~]

Note: cs-fixer will be run later, when some other branches will be merged.

Todo:
- [x] Add adjustment via AdjustmentSubscriber and AdjustmentDTO (like in discountAction)
- [ ] Enforce not null values on adjustment in db

PR #3712, opened 2015-12-14
[RFC] Order Finance Rework - Adjustments onto InventoryUnits

peteward:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | 
| License       | MIT

replaces #3520

### Why?

For medium to large organizations, or anyone who wants to do refunds, we must know finances at the InventoryUnit level. 

Adjustments are great, giving a clear history of all financial changes to entities. However at the OrderItem level they don’t give enough definition to easily know how much an item was worth to the business.

Large stock systems in retail often have no knowledge of Orders or OrderItems, all they know about is single units of stock that get sold for an amount and could be refunded. To integrate with these systems (tax, finance, accounting, stock) you need to be able to send accurate unit values.

With the current model of Adjustments at the OrderItem level, if a customer buys qty = 4 of a product, then returns 2, how do you know the financial value of a specific shipped item of stock? You can just about work it out, but it’s not fun. If you want to do this in a report for all items sold over the past month… ewww! 

### The idea

* Move `Adjustments` from the `OrderItem` to the `InventoryUnit` to satisfy the above.
* Unify Adjustment creation so that it works at InventoryUnit-level for our common types (tax and promotion).
* Shipping and Payment type adjustments stay at the Order-level. Shipping is really a special case of a unit connected to an order but that's a different discussion… 
* Ensure that there is a clear and accurate process to how adjustments are applied to ensure nothing can be calculated incorrectly.

### Progress

- [x] Move adjustments - This was not as clean as I had hoped due to the separation of OrderBundle and InventoryBundle. I've tried to leave a consistent interface for Adjustments where Order calls OrderItem which wraps around InventoryUnit in Core.
- [x] Change the front-end cart summary page to demonstrate different uses of adjustments.
- [x] Removed `setTotal()` on Orders and OrderItems as they should always be calculated
- [x] Adjustment creation was refactored to work via a Data Transform Object and events. I'm unsure about this pattern as it's not used in Sylius. Now that we have Factories it might be better to work directly through the factory rather than events?
- [x] Separating revert and apply of taxes and promotions to ensure remaining taxes don't impact promotional discount amounts (and vice versa)
- [ ] Promotions - We have a very big rework of promotions pending these changes, but some smaller changes need to be made here to consider order-level adjustments when calculating tax at inventory-unit level. At the moment a discount will not be considered for tax.
- [ ] Tests - these are mainly written but I removed them to make the PR shorter and more readable.

### Future Changes

* Rationalise the cart events. There are too many imo and we have raised this topic before (#3525 and #3575). A possibility would be to nest CART_CHANGE so it's called inside other checkout stage events rather than attach the same listeners to everything.
* Ensure promotions record amounts at InventoryUnit-level 

I'm looking for feedback on this approach before I continue to modify promotions and clean up tests. These are the main areas:
* Order => OrderItem => InventoryUnit & Adjustment structure
* Adjustment creation via DTO.


peteward:
To emphasise how severe this situation is. Currently if you are using both promotions and tax calculation (which isn't exactly unusual), this is what you could see:

image

The subtotal of items here is **$390**.
The promotion here is %10 off  (**-$39**).
The taxes here are flat rate 8% off items. However this is coming out at **$31.20** when it should be **$28.08**.

The customer is being overcharged because the taxable amount is not considering the promotional discount. It is calculating 8% off **$390** instead of 8% off **$351**.

It does this because the promotional discount adjustment is at the order-level, and taxes are at the item level. Taxes therefore do not consider order-level adjustments. This is unsurprising because it's painful to do so, how would you know what or how to handle this if you have a promotional discount that needs to be split across multiple items? It's not the job of the tax processor to do this. It should be the promotion processor to ensure promotional discounts are recorded at the same level.

This is a very big issue that needs addressing fast. I've spent a long time thinking about this and based on past experience I really think the only long-term solution is to move adjustments down to the inventory unit level. 

We have a lot of the building blocks in place here, just need some guidance and feedback to get it acceptable for the community.

pjedrzejewski:
@peteward Quick question before I dive into this PR: Is there any reason why we can't keep OrderItem adjustable too? That would work for some simpler use cases. Sylius itself would use adjustments on order and inventory unit levels?

peteward:
I understand the reason behind the question, and it's probably possible, but I really recommend not.

The problem being that in the above situation where you have adjustments at different levels it is extremely difficult to calculate at another level.  Like in the example above if you have an Order-level promotion combined with an OrderItem-level promotion and InventoryUnit-level taxes, how do you work out how much tax should be due on an InventoryUnit?

You have to take the Order-level promotion adjustment (say $10 off), divide it up between all OrderItems and then again between each InventoryUnit (handling any rounding issues). You then have to do the same to the OrderItem-level promotion to split it between InventoryUnits. 

This is doable, but prohibitively complex, especially if you want to retrieve reports for hundreds or thousands of items. It's not the sort of logic you want to be running on-demand.

As Sylius stands now, I would say ditch Order-level promotion adjustments and make sure everything is at least at the same OrderItem-level to work correctly.  But it's the future of refunds where you need to split finances down to InventoryUnit-level that we need to build for. Even if it's not going to be supported immediately.

Having finances at the InventoryUnit-level gives you a solid grounding to go forwards.

I've checked lots of other platforms and none of them handle this well, it's a real opportunity to get it right.

peteward:
I think a way around the coupling issues would be to create an `OrderItemUnit` in the OrderBundle, implementing an `OrderItemUnitInterface` which is then also implemented by (and replaced by) the Core `InventoryUnit`. 

This would give a consistent structure at the 'unit-level' regardless of being a stand-alone bundle or the full stack application. 

It also means that OrderItem can consistently be an abstraction on top of Units for the "simpler" use-cases that you're referring to. It wouldn't have Adjustments directly against it, but would wrap them up with convenient methods for retrieval as required.

pjedrzejewski:
@peteward I like the idea. These could also be related to / implement ShipmentItemInterface etc. It kinda feels odd to apply adjustment to something that is also ShipmentItem or InventoryUnit.

michalmarcinkowski:
I believe this can be closed in favor of #3795, #3842, #3868.

PR #3714, opened 2015-12-14
[BugFix][PHPCR] Fix criteria "name" for DocumentRepository

steffenbrem:
Fix bug where you could not select phpcr documents with criteria "name" (it would throw: `Cannot use nodename property "name" of class "Xyz" as a dynamic operand use "localname()"` instead.)


pjedrzejewski:
Thanks Steffen!

PR #3718, opened 2015-12-14
Avoid flushing only a single resource when not required

peteward:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

If you only flush a single resource it's possible that you can lose other important entity flushes - especially as Sylius uses Sequences called on preFlush.

We just had an issue where a Customer Sequence was not being flushed in checkout because only the single Customer resource was being flushed. The sequence therefore didn't change and it was impossible to register because the sequence had not incremented and the value was not unique next time around...

Unless there's a really particular reason not to flush everything, I think we always should.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Checkout/Step/SecurityStep.php
+++ b/src/Sylius/Bundle/CoreBundle/Checkout/Step/SecurityStep.php
@@ -188,7 +188,7 @@ class SecurityStep extends CheckoutStep
     protected function saveResource($resource)
     {
         $this->getManager()->persist($resource);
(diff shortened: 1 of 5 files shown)


pjedrzejewski:
Definitely, this is causing only bugs. That being said, we need to do our best to have only a single flush for every request, which is totally possible for most of actions. Thank you very much Peter! :+1:

peteward:
Yes - the single flush approach will be an interesting refactor!!

PR #3719, opened 2015-12-14
Update TestKernel.php

gonzalovilaseca:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

So we had a problem with Symfony's ``MimeTypeGuesser`` as it implements the singleton pattern, and this was code was resetting ``$guessers``. So I guess it will break any other singleton instance that has properties.

### DIFF
--- a/app/TestKernel.php
+++ b/app/TestKernel.php
@@ -18,6 +18,7 @@ class TestKernel extends Kernel
 
         if (!in_array($this->environment, ['test', 'test_cached'])) {
             parent::shutdown();
+
             return;
         }
 
@@ -42,13 +43,19 @@ class TestKernel extends Kernel
             }
 
             $serviceObject = new \ReflectionObject($service);
-            foreach ($serviceObject->getProperties() as $prop) {
-                $prop->setAccessible(true);
-                if ($prop->isStatic()) {
-                    continue;
-                }
 
-                $prop->setValue($service, null);
+            $properties = $serviceObject->getProperties();
+            // If it has a static property it could be a singleton, therefore we should not
+            // reset it's properties
+            foreach ($properties as $property) {
+                if ($property->isStatic()) {
+                    continue 2;
+                }
+            }
+
+            foreach ($properties as $property) {
+                $property->setAccessible(true);
+                $property->setValue($service, null);
             }
         }
         $property->setValue($container, null);


app/TestKernel.php
```diff
@@ -42,13 +42,28 @@ protected function cleanupContainer($container)
...
+            foreach ($properties as $prop) {
                 $prop->setAccessible(true);
                 if ($prop->isStatic()) {
-                    continue;
+                    $isStatic = true;
+                    break;
                 }
+            }
+
+            if (!$isStatic) {
```
pamil: We can use early (~~return~~) `continue` here :)
```php
if ($isStatic) {
    continue;
}
```

pamil:
Looks good for me and memory usage is still acceptable. :)

app/TestKernel.php
```diff
@@ -42,13 +42,28 @@ protected function cleanupContainer($container)
             }
 
             $serviceObject = new \ReflectionObject($service);
-            foreach ($serviceObject->getProperties() as $prop) {
+            
+            $properties = $serviceObject->getProperties();
+            // If it has a static property it could be a singleton, therefore we should not
+            // reset it's properties
+            $isStatic = false;
+            foreach ($properties as $prop) {
```
pjedrzejewski: $property

app/TestKernel.php
```diff
@@ -42,13 +42,28 @@ protected function cleanupContainer($container)
...
                 if ($prop->isStatic()) {
-                    continue;
+                    $isStatic = true;
+                    break;
                 }
+            }
+
+            if (!$isStatic) {
+                foreach ($properties as $prop) {
+                    $prop->setAccessible(true);
```
pamil: Anyway, we don't need this line, as we set every variable accessible earlier.

app/TestKernel.php
```diff
@@ -42,13 +42,28 @@ protected function cleanupContainer($container)
...
+                    if ($prop->isStatic()) {
```
pamil: Same here for the `if`, there cannot be any static property as we skip these.

gonzalovilaseca:
It was a last minute PR, I'll recheck everything tomorrow morning.

gonzalovilaseca:
PR updated

app/TestKernel.php
```diff
@@ -31,7 +32,7 @@ public function shutdown()
      */
     protected function cleanupContainer($container)
     {
-        $object = new \ReflectionObject($container);
+        $object   = new \ReflectionObject($container);
```
pjedrzejewski: Let's not align variable assignments.

app/TestKernel.php
```diff
@@ -42,14 +43,27 @@ protected function cleanupContainer($container)
...
-                if ($prop->isStatic()) {
-                    continue;
+
+            $properties = $serviceObject->getProperties();
+            // If it has a static property it could be a singleton, therefore we should not
+            // reset it's properties
+            $isStatic = false;
+            foreach ($properties as $property) {
+                if ($property->isStatic()) {
+                    $isStatic = true;
```
pamil: That's a little bit of sorcery, but we can remove `$isStatic` and put `continue 2;` here and it will jump to the beginning of the next iteration of services loop :)

pjedrzejewski: :+1:

gonzalovilaseca: Cool, didn't know that was possible!

pamil: Me too till today :D

pjedrzejewski: :sunglasses:

gonzalovilaseca:
done!

app/TestKernel.php
```diff
@@ -42,14 +43,22 @@ protected function cleanupContainer($container)
...
-                if ($prop->isStatic()) {
-                    continue;
+
+            $properties = $serviceObject->getProperties();
+            // If it has a static property it could be a singleton, therefore we should not
+            // reset it's properties
+            foreach ($properties as $property) {
+                if ($property->isStatic()) {
+                    continue 2;
+                    break;
```
stloyd: This shoud be remove as it's dead code caused of `continue` call line above.

app/TestKernel.php
```diff
@@ -42,14 +43,22 @@ protected function cleanupContainer($container)
...
                 }
+            }
 
-                $prop->setValue($service, null);
+            foreach ($properties as $property) {
+                $property->setAccessible(true);
+                $property->setValue($service, null);
             }
+
```
stloyd: Should be reverted IMO. New line could be added after `}` line below.

app/TestKernel.php
pamil: :+1:

pjedrzejewski:
Thanks Gonzalo!

Issue #3066, opened 2015-07-20
[RFC] Renaming Country::isoName to more appropriate name


pamil:
Right now, `Country::isoName` contains country codes from [ISO 3166-1 alpha-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2).

It's rather not a fortunate name, so I'm thinking about refactoring `isoName` to `isoCode`, `countryCode` or simply `code`. What do you think about that? Which name do you prefer?

michalmarcinkowski:
We will introduce ``` code ``` field for most entities, in order to uniquely identify objects during import/export or API calls from other systems. IMO we could change this field to be simply ``` code ```. @Sylius/core-team ?

Symfomany:
:+1:  @michalmarcinkowski

kayue:
+1 for `code`

pamil:
:+1: `code`

gperdomor:
+1 for <code>code</code> too

patrick-mcdougle:
I personally like isoCode because it reminds us that the values we place in there should be taken from the ISO 3166-1 alpha-2 standard and shouldn't be just any old abbreviation.

pjedrzejewski:
``code`` it is then! :+1:

pamil:
Fixed by #3696.

PR #3616, opened 2015-11-24
[WIP] Configuration trees update

TheMadeleine:
| Q                    | A
| -------------       | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | no
| License       | MIT

We will be changing the structure of configuration for all resources, mainly introducing another level.
Before:
```yaml
sylius_currency: 
    classes: 
        currency: 
            model: Currency
```
After
```yaml
sylius_currency: 
    resources: 
        currency: 
            classes: 
                model: Currency
```

Issue #2821, opened 2015-06-08
Model class parameter for custom resources


umpirsky:
When configuring your resource, like documented in http://docs.sylius.org/en/latest/bundles/SyliusResourceBundle/configuration.html, would be nice to set model class parameter in container config.

aRn0D:
@umpirsky We should parse the configuration like the "advanced" way. For that, we need to refactor the extension and specify a common configuration structure.

Related PR/issue : #2520, #2651, #2622

pjedrzejewski:
I am actually thinking we should remove advanced configuration from bundles and simply use ``prepend`` to configure everything under ``sylius_resource`` configuration.

1. We want to have 1:1 functionality for both cases.
2. That will remove a lot of code and duplication from resource bundle.

Basically make Sylius use the simple configuration inside of bundles. What do you guys think? @Sylius/core-team @michalmarcinkowski

michalmarcinkowski:
@pjedrzejewski Good idea! :+1:

steffenbrem:
@pjedrzejewski Nice, like that idea too. Would simplify it a lot! :+1:

michalmarcinkowski:
Fixed in #3721

Issue #2622, opened 2015-03-25
[RFC][ResourceBundle] Loading "drivers"


aRn0D:
First, we need remove the `DatabaseDriverFactory` because it is not easily extendable. I think we should use a kind of chain of responsibility pattern because it allow us to configure several drivers (Doctrine and custom drivers). Every drivers must implement the same interface which ask to create the controller, repository and the manager as a service. Is the `DatabaseDriverInterface` is enough? I am sure because it ask to *load* the driver but it does not ask to *create* the right services.

Thus, what do we do ?

1 - Update the `DatabaseDriverInterface`, we should rename it in `ResourceDriverInterface`

```php
ResourceDriverInterface 
{
   // Create the controller definition
   public function createControllerDefinition($classe);

   // Create the object manager definition
   public function createManagerDefinition($classe);

   // Create the repository definition
   public function createRepositoryDefinition($classe);

   // Return the name of the name
   public function getSupportedDriver();
}
```

2 - Remove the `DatabaseDriverFactory`

3 - Declare every drivers as service (service definition file : driver.xml) with a tag `sylius_resource.driver_name`, *driver_name* is the name of the driver and it should match the name defined by `getSupportedDriver`.

```xml
        <!-- Resources/config/driver.xml -->
        <service id="id" class="%class%">
            <tag name="sylius_resource.custom_driver" />
        </service>
```

4 - Create a compiler pass `LoadResourceDriverPass`, for each resource it will get the right driver and create the right services
 
```php
namespace Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler;

class ObjectToIdentifierServicePass implements CompilerPassInterface
{
   public function process(ContainerBuilder $container)
   {
        $resources = container->getParameter('sylius.resource.config');

        foreach ($resources as $config) {
            // Get the driver
            $driver = $taggedServices = $container->findTaggedServiceIds('sylius_resource.driver');

            // Create the service that the resource bundle need to work
            $driver->addMethodCall('createControllerDefinition', array($config));
            $driver->addMethodCall('createManagerDefinition', array($config));
            $driver->addMethodCall('createRepositoryDefinition', array($config));
        }
    }
}
```

5 - We need to create `ResourceRepositoryInterface` `RecourceManagerInterface` which will be a generics interfaces for repositories and object managers. They will reduce the coupling to Doctrine and permit create custom driver easily.

@Sylius/core-team @stloyd What do you think ?

Related issue/PR #253, #2520

[...]

PR #3721, opened 2015-12-15
Rework SyliusResourceBundle drivers

pjedrzejewski:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | https://github.com/Sylius/Sylius/issues/2821 https://github.com/Sylius/Sylius/pull/2520 https://github.com/Sylius/Sylius/issues/2622 https://github.com/Sylius/Sylius/pull/3616
| License       | MIT
| Doc PR        | https://github.com/Sylius/Sylius-Docs/pull/383

Based on #2255  & #3616.

Hey folks!

This one is big one but I believe step in the right direction. The diff may be hard to follow because of tree changes but here is an overview:

* Reworked drivers system and **most importantly, we finally have a single place where all stuff for resource is generated** etc. This gives us a lot of power for future features and fixes, previously it was a complete mess;
* Introduced Metadata and Registry to properly manage individual resources, no longer array processing;
* Non-Sylius resources will have the same feature set;
* Removed a lot of inheritance and abstraction that was not needed (for loading files etc.)
* Changed (standardized) configuration tree structure for all resources:

Before (simplified):

```
sylius_shipping: 
    driver: doctrine/orm
    classes: 
        shipping_method: 
            model: ShippingMethod
```

After (full):

```
sylius_shipping: 
    resources: 
        shipping_method: 
            classes: 
                model: ShippingMethod
                interface: ShippingMethodInterface
                controller: ResourceController
                repository: ShippingMethodRepository
                factory: ShippingMethodFactory
                form:
                    default: ShippingMethodType
                    choice: ShippingMethodChoiceType
             translation:
                 <same structure>
```


peteward:
I think your example above should be:

```
sylius_shipping: 
    resources: 
        shipping_method: 
            classes: 
                   ....
```

src/Sylius/Bundle/ProductBundle/DependencyInjection/SyliusProductExtension.php
```diff
@@ -57,20 +65,28 @@ private function prependAttribute(ContainerBuilder $container, array $config)
...
                     'product' => array(
-                        'subject'         => $config['classes']['product']['model'],
+                        'subject'         => $config['resources']['product']['classes']['model'],
                         'attribute'       => array(
-                            'model'       => 'Sylius\Component\Product\Model\Attribute',
-                            'repository'  => 'Sylius\Bundle\TranslationBundle\Doctrine\ORM\TranslatableResourceRepository',
-                            'translation' => array(
-                                'model' => 'Sylius\Component\Product\Model\AttributeTranslation'
+                            'classes' => array(
+                                'model'       => 'Sylius\Component\Product\Model\Attribute',
```
peteward: You're not using `X::class` here compared to everywhere else, is there a reason?

peteward:
Looks pretty sensible to me :+1:  What power will these changes give in future?

TheMadeleine:
@peteward imo it is important to show the detailed structure with all the available nodes, since for example some nodes are newly introduced and applied everywhere on each resource. Take `interface` for instance, or the new, unified `form:`node structure.

@pjedrzejewski The `form` node has wrong indent level in the description ;)

app/config/config_dev.yml
```diff
@@ -23,7 +23,7 @@ monolog:
             level: info
 
 assetic:
-    use_controller: true
+    use_controller: false
```
pamil: Why?

src/Sylius/Bundle/AttributeBundle/DependencyInjection/SyliusAttributeExtension.php
```diff
@@ -12,10 +12,12 @@
 namespace Sylius\Bundle\AttributeBundle\DependencyInjection;
 
 use Sylius\Bundle\ResourceBundle\DependencyInjection\Extension\AbstractResourceExtension;
+use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 
 /**
- * Attribute extension.
+ * Archetype extension.
```
stloyd: Not really.

src/Sylius/Bundle/CartBundle/DependencyInjection/SyliusCartExtension.php
```diff
@@ -23,37 +24,48 @@
  * @author Saša Stamenković <umpirsky@gmail.com>
  * @author Jérémy Leherpeur <jeremy@leherpeur.net>
  */
-class SyliusCartExtension extends AbstractResourceExtension implements PrependExtensionInterface
```
stloyd: This should be reverted otherwise, `prepend()` method will not be called.

src/Sylius/Bundle/RbacBundle/DependencyInjection/SyliusRbacExtension.php
```diff
@@ -52,20 +57,4 @@ public function load(array $config, ContainerBuilder $container)
         $container->setParameter('sylius.rbac.default_permissions', $config['permissions']);
         $container->setParameter('sylius.rbac.default_permissions_hierarchy', $config['permissions_hierarchy']);
     }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function prepend(ContainerBuilder $container)
```
stloyd: What if someone want to use this bundle standalone?

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        if (!$metadata->hasParameter('validation_groups')) {
+            return;
+        }
+
+        $validationGroups = $metadata->getParameter('validation_groups');
+
+        foreach ($validationGroups as $formName => $groups) {
+            $suffix = ('default' === $formName ? '' : sprintf('_%s', $formName));
```
stloyd: `(` & `)` can be removed.

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
```diff
@@ -0,0 +1,193 @@
...
+        $container->setDefinition($metadata->getServiceId('controller'), $definition);
+    }
+
+    /**
+     * @param ContainerBuilder $container
+     * @param MetadataInterface $metadata
+     */
+    protected function addFactory(ContainerBuilder $container, MetadataInterface $metadata)
+    {
+        $translatableFactoryInterface = 'Sylius\Component\Translation\Factory\TranslatableFactoryInterface';
```
stloyd: `TranslatableFactoryInterface::class`?

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /**
+     * @param ContainerBuilder $container
+     * @param MetadataInterface $metadata
+     */
+    protected function addForms(ContainerBuilder $container, MetadataInterface $metadata)
+    {
+        foreach ($metadata->getClass('form') as $formName => $formClass) {
+            $suffix = ('default' === $formName ? '' : sprintf('_%s', $formName));
```
stloyd: Same about `(` & `)`.

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
```diff
@@ -0,0 +1,193 @@
...
+                case 'choice':
+                    $definition->setArguments(array(
+                        $metadata->getClass('model'),
+                        $metadata->getDriver(),
+                        $alias,
+                    ));
+                break;
+
+                default:
+                    $validationGroups = array('Default');
```
stloyd: Move this to `} else {` instead of rewriting variable?

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/DriverFactory.php
```diff
@@ -0,0 +1,57 @@
...
+        }
+
+        switch ($type = $metadata->getDriver()) {
+            case SyliusResourceBundle::DRIVER_DOCTRINE_ORM:
+                return self::$drivers[$type] = new DoctrineORMDriver();
+            case SyliusResourceBundle::DRIVER_DOCTRINE_MONGODB_ODM:
+                return self::$drivers[$type] = new DoctrineODMDriver();
+            case SyliusResourceBundle::DRIVER_DOCTRINE_PHPCR_ODM:
+                return self::$drivers[$type] = new DoctrinePHPCRDriver();
+            default:
```
stloyd: Instead you can move this `throw` after `switch`.

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/DriverFactory.php
```diff
@@ -0,0 +1,57 @@
...
+     */
+    public static function get(MetadataInterface $metadata)
+    {
+        $type = $metadata->getDriver();
+
+        if (isset(self::$drivers[$type])) {
+            return self::$drivers[$type];
+        }
+
+        switch ($type = $metadata->getDriver()) {
```
stloyd: Duplicated assignment.

src/Sylius/Bundle/TranslationBundle/EventListener/ORMTranslatableListener.php
```diff
@@ -67,15 +67,23 @@ public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
      */
     private function mapTranslatable(ClassMetadata $metadata)
     {
-        // In the case A -> B -> TranslatableInterface, B might not have mapping defined as it
-        // is probably defined in A, so in that case, we just return.
-        if (!isset($this->configs[$metadata->name])) {
+        $className = $metadata->name;
+
+        try {
+            $resourceMetadata = $this->registry->getbyClass($className);
```
stloyd: Typo? `getByClass()`

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
pjedrzejewski: Translation component is optional at this point, I will open an RFC about including it in the Resource, but for now we can't assume that the interface exists.

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/DriverFactory.php
pjedrzejewski: :+1:

app/config/config_dev.yml
pjedrzejewski: I was checking something, will revert. :)

pamil:
Also fixes #3066.

src/Sylius/Bundle/ProductBundle/DependencyInjection/SyliusProductExtension.php
pjedrzejewski: Nope, will change that! Good catch!

pjedrzejewski:
@pamil Are you sure about this issue? I did not touch Country model. This will be fixed in #3696.

pamil:
@pjedrzejewski looked at wrong issue number, I mean #3141 :)

pamil:
@pjedrzejewski looked at wrong issue number, I mean #3141 :)

src/Sylius/Bundle/RbacBundle/DependencyInjection/SyliusRbacExtension.php
pjedrzejewski: That was a copy paste mistake, I copied this class from my previous PR where there was no prepend. :) Reverting! Thanks.

aRn0D:
The new configuration tree is good but :
* The `DriverFactory` must be deleted because it avoid to add extra drivers (custom ones). Use case: you can use the resource bundle as a proxy to work with API.
* Tree levels for the Doctrine drivers, I think it is too much! Using trait could be allow to remove `AbstractDoctrineDriver` but I am not sure it is the best solution.
* You remove the chain DependencyInjection loader, I think it is not a great idea because lot of people who use the bundle standalone want to write those files in yaml.
`RepositoryInterface` should not extend the `ObjectRepository` because this bundle does not need every method from this interface (like getClassName) and it decouples this bundle from doctrine (yes a bit...).
* The new extensions (`DependencyInjection`) are cleaner but it miss some old features

A compiler pass could process every drivers which are declared as a service and had a custom tag ( already explained there #2622). This solution seems to more salable.

To be honest I am not a fan of this PR but you're the better to decide what it is the better for Sylius.

pjedrzejewski:
@aRn0D 

* Our current implementation does not allow to ``custom`` drivers too. I am perfectly aware that it would be great to have these and that's one of main points of this PR. We have now clean DriverInterface which can be implemented. I hate the fact that DriverFactory is static but I don't see how deleting it would improve the situation. :D How would you get a proper driver? I will definitely look into clean way of having all Drivers registered in a single place + ability to add your own.

* I don't like how Driver inheritance is handled right now: AbstractDriver -> AbstractDoctrineDriver -> DoctrineORMDriver but this is the best I could do given the existing code + time. Obviously these will be refactored but definitely not using Traits. Traits hide complexity just like previous implementation did. Yes, you had some fancy helper methods in all these classes but it was total mess and magic. Configure call with masks that was executing crazy amount of ifs + logic spread across AbstractResourceExtension and Drivers. Now everything is in a single place -> DriverInterface implementation. (not entirely, I'd like to move translation handling into 1 place as well but that's for later.

* There was a lot of overhead and code to do a simple thing like loading files. I don't like "configuring" anything via overriding protected methods or attributes. I know that doing ``protected $configFiles = array(...)`` looks simple but one needs to check that in parent class this property is used to load files and generally it was a nasty inheritance magic. I know that in some places we still use protected methods to handle some logic and I plan to refactor this too but we need to make small steps otherwise we end up with PRs like #2255. Lesson learned.

* RepositoryInterface is completely different story and out of scope for this PR. In #2255 I dropped the ``extends ObjectRepository`` but it is not that simple. I'd leave that for a separate RFC.

Regarding your idea of registering drivers via tags. Sounds great but is not easily doable. Tags would be processed (via compiler pass) after container is built via extensions and also you can tag services. How would you use a service on ContainerBuilder? This is not possible. I tried to approach this from a different angle and register resources in the registry during loading time and then process services in compiler passes but this makes whole thing more complicated - you don't have access to all Resource-specific services in DIC extensions.

I understand you are not a fan of this PR because it replaces a lot of work you have done on resource (in master and as POC) but I hope that this comment justifies why I think this (very first step of) new approach is cleaner.

There is plenty of work left to do and I will be opening further PRs.

aRn0D:
@pjedrzejewski don't worry, I understand! It was just feedback, I have not tested my solution too. :+1: for small PR ;)

pjedrzejewski:
@aRn0D <high five>

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
```diff
@@ -0,0 +1,193 @@
...
+    {
+        foreach ($metadata->getClass('form') as $formName => $formClass) {
+            $suffix = 'default' === $formName ? '' : sprintf('_%s', $formName);
+            $alias = sprintf('%s_%s%s', $metadata->getApplicationName(), $metadata->getName(), $suffix);
+
+            $definition = new Definition($formClass);
+
+            switch ($formName) {
+                case 'choice':
+                    $definition->setArguments(array(
```
pamil: Why don't we pass validation groups to `choice` type?

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/DriverFactory.php
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * @param MetadataInterface $metadata
+     *
+     * @return DriverInterface
+     *
+     * @throws UnknownDriverException
+     */
+    public static function get(MetadataInterface $metadata)
+    {
+        $type = $metadata->getDriver();
```
pamil: `$metadata->getDriverType()`?

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/DriverFactory.php
```diff
@@ -0,0 +1,57 @@
...
+
+        if (isset(self::$drivers[$type])) {
+            return self::$drivers[$type];
```
pamil: It's quite counterintuitive that `DriverFactory` does not always create new driver.

pjedrzejewski: DriverResolver? This class is crap anyway, I want to get rid of it soon, but for now it has to stay. WDYT?

pjedrzejewski: DriverProvider?

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
pjedrzejewski: This is existing behavior and I did not want to change it. Not sure if it makes sense, but let's leave that for a separate RFC.

src/Sylius/Bundle/ResourceBundle/EventListener/LoadORMMetadataSubscriber.php
```diff
@@ -65,28 +63,25 @@ public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
...
-                } else if (isset($class['translation']['model'])
-                    && $class['translation']['model'] === $metadata->getName()) {
-                    $metadata->isMappedSuperclass = false;
-
-                    if (isset($class['translation']['repository'])) {
-                        $metadata->setCustomRepositoryClass($class['translation']['repository']);
-                    }
-                }
+        foreach ($this->resourceRegistry->getAll() as $alias => $resourceMetadata) {
+            if ($resourceMetadata->hasClass('repository') && $resourceMetadata->getClass('model') === $metadata->getName()) {
```
pamil:
```php
if ($resourceMetadata->getClass('model') !== $metadata->getName()) {
  continue;
}

if ($resourceMetadata->hasClass('repository')) {
  $metadata->setCustomRepositoryClass($resourceMetadata->getClass('repository'));
}

$metadata->isMappedSuperclass = false;
```

src/Sylius/Bundle/TranslationBundle/DependencyInjection/Mapper.php
```diff
@@ -24,12 +24,15 @@ class Mapper
...
     {
         $mapping = array();
 
-        $mapping[$config['model']] = $config;
-        $mapping[$config['translation']['model']] = $mapping[$config['model']];
+
+        $mapping[$resource['classes']['model']] = $resource;
+        $mapping[$resource['translation']['classes']['model']] = $mapping[$resource['classes']['model']];
+
+
```
pamil: Three blank lines in a row! :)

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/DriverFactory.php
pamil: `DriverProvider` looks good.

src/Sylius/Bundle/ResourceBundle/EventListener/LoadORMMetadataSubscriber.php
pjedrzejewski: :+1:

src/Sylius/Bundle/TranslationBundle/DependencyInjection/Mapper.php
pjedrzejewski: Oh, this file should be removed actually.

src/Sylius/Bundle/ResourceBundle/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ CHANGELOG
 ### v0.16.0
 
 * Introduce Factory for all resources.
+* New ``DriverInterface`` for resources.
+* Reworked all listeners to use new Registry.
```
michalmarcinkowski: Please complete the changelog. Configuration tree has been changed, validation_group -> validation_groups etc.

UPGRADE.md
```diff
@@ -37,7 +38,9 @@ https://github.com/Sylius/Sylius/pull/2717
  * Email no longer exist on User. It is on Customer now
  * In the checkout we depend on Customer not User
  * In templates in many places we use Customer instead of User entity now
-
+ * Change your configuration for bundles all resources are configured under ``sylius_<bundle> -> resources -> <resource>`` node
+ * Rename validation groups parameter names if you used them anywhere
```
michalmarcinkowski: I belive this should go to 0.15 -> 0.16 upgrade :)

michalmarcinkowski:
Great job @pjedrzejewski! :+1: Looking forward to next PRs! :smile:

pjedrzejewski:
:fireworks: Thanks @TheMadeleine for help!

src/Sylius/Bundle/ResourceBundle/DependencyInjection/SyliusResourceExtension.php
```diff
@@ -31,84 +31,43 @@ class SyliusResourceExtension extends AbstractExtension
...
 
-        if ($translationsEnabled) {
-            $mapper = new Mapper();
-        }
+            DriverProvider::get($metadata)->load($container, $metadata);
 
-        foreach ($configs as $name => $config) {
-            list($prefix, $resourceName) = explode('.', $name);
-            $manager = isset($config['object_manager']) ? $config['object_manager'] : 'default';
+            if (isset($resourceConfig['translation'])) {
```
bendavies: there is an error here. 
you are no longer checking if the translation bundle exists.

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Configuration.php
```diff
@@ -58,28 +57,58 @@ private function addResourcesSection(ArrayNodeDefinition $node)
...
+                                ->addDefaultsIfNotSet()
+                                ->children()
+                                    ->arrayNode('default')
+                                        ->prototype('scalar')->end()
+                                        ->defaultValue(array())
+                                    ->end()
+                                ->end()
+                            ->end()
+                            ->arrayNode('translation')
+                                ->addDefaultsIfNotSet()
```
bendavies: this seems wrong to me. you cannot disable translations per resource with this. @pjedrzejewski

bendavies: please continue discussion here.

src/Sylius/Bundle/ResourceBundle/DependencyInjection/SyliusResourceExtension.php
bendavies: https://github.com/Sylius/Sylius/pull/3749

Issue #3672, opened 2015-12-03
[Rbac][Possible Bug] Permission issue with static content


steffenbrem:
If your not logged in you cannot view static content, which should be possible obviously. Currently Sylius configured CMF to use the `ResourceController` to show `StaticContent`, so this means it goes through Sylius authorization process. Because of the following line: https://github.com/Sylius/Rbac/blob/master/Authorization/AuthorizationChecker.php#L64 you can never pass this authorization (since it will always return `false`). 

Shouldn't it only try to authorize when the permission actually exists in the database, so we could leave out the `sylius.static_content.show` permission in the `rbac` configuration, so it will always be granted.

Am I missing something or is this a bug?

pjedrzejewski:
We should disable the security for this route then, in Sylius routing:

```
app_route_xyz:
    ...
    _sylius:
        permission: false
```

PR #3664, opened 2015-12-02
[WebBundle] Fixed showing static content by route sylius_static_content_show

pamil:
No BC breaks, bug fixed. Variable named `cmfMainContent` did not exist in that template.

PR #3722, opened 2015-12-15
[CMF Integration] StaticContent/show template can be used with both ContentController and ResourceController

pamil:
As @psyray noticed in #3664, we have two endpoints that can show static content. In that PR I wasn't aware of `ContentController` and it resulted in BC change, this one reverts it and adds compatibility layer. Possibly related to #3672.


pjedrzejewski:
Thank you guys!

PR #3726, opened 2015-12-15
Support for Doctrine ORM >=2.5

pamil:


pjedrzejewski:
Could you update other composer.json's?

https://github.com/Sylius/Sylius/search?l=json&q=doctrine%5C%2Form&type=Code&utf8=%E2%9C%93

michalmarcinkowski:
@pamil does the fix work on `v2.5.1`? If not we should exclude this version.

src/Sylius/Component/Channel/composer.json
```diff
@@ -22,7 +22,7 @@
     "require": {
         "php": ">=5.5.9",
 
-        "doctrine/orm": "2.4.*",
+        "doctrine/orm": "^2.4.8",
```
pjedrzejewski: Channel really requires Doctrine ORM? Should not be the case ...

pamil:
@michalmarcinkowski it should be fine on all `2.5.*` releases.

src/Sylius/Component/Channel/composer.json
pamil: Looks like ORM's exceptions are used in PhpDocs in `ChannelRepositoryInterface`.

`doctrine/common` is used too in `ChannelsAwareInterface` only.

michalmarcinkowski: ORM's exceptions should be removed from interface docs. The interface doesn't know anything about the implementation especially about the thrown exceptions.

Let's require `doctrine/common` then.

pjedrzejewski:
Thanks Kamil!

src/Sylius/Component/Channel/composer.json
pjedrzejewski: I opened a separate issue for that #3729.

michalmarcinkowski: :+1:

Issue #1295, opened 2014-03-27
Validation not taken into account in promotion configuration forms


bburnichon:
As validation_groups is set to ```sylius``` by default, Symfony's Form Component does not validate any of the Constraints set in Forms.

This is a big issue for configuration of promotions which have no underlying object to validate. Only an array.

On a fresh install of Sylius, I just added GreaterThan in ItemCountConfigurationType.

```php
->add('count', 'integer', array(
    'label' => 'sylius.form.rule.item_count_configuration.count',
    'constraints' => array(
        new NotBlank(),
        new Type(array('type' => 'numeric')),
        new GreaterThan(array('value' => 0)),
   )
))
```

I could saved a -1 value in the item_count_configuration of New Year Promotion

umpirsky:
@pjedrzejewski I am interested into fixing this issue. Any directions?

michalmarcinkowski:
@umpirsky According to docs (http://symfony.com/doc/current/book/forms.html#adding-validation), you can just add validation groups to constraint like this:
```php
new NotBlank(array('groups' => array('sylius'))
```

Tell me if it works :)

umpirsky:
@michalmarcinkowski This did the trick, thanks!

But, why it does not work with:

```php
class MyConfigurationType extends AbstractType
{
    public function setDefaultOptions(OptionsResolverInterface $resolver)
    {
        $resolver
            ->setDefaults(array(
                'validation_groups' => ['sylius'],
            ))
        ;
    }
}
```

michalmarcinkowski:
@umpirsky Not sure what are you asking.
 
If the form is embedded I suppose it may be connected with cascade validation (http://symfony.com/doc/current/reference/forms/types/form.html#cascade-validation)

Is that your case?

umpirsky:
@michalmarcinkowski Yes, but that is the case with all rule checker config types. They are embedded by default.

PR #3731, opened 2015-12-16
Remove configuration form validation groups

pjedrzejewski:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | https://github.com/Sylius/Sylius/issues/1295
| License       | MIT
| Doc PR        | -

Based on #3721.

Helloes.

This change basically forces all Sylius resource forms to use Default validation group. Result:

* On (promotion action/rule, calculator, report etc.) configuration forms we can safely use constraints without any groups, so can safely get rid of these ugly and long configuration nodes for these forms. Use-case where someone wants to override validation of FixedDiscountConfigurationType is super rare and we still allow this by overriding whole form. Supporting this rare use-case with current approach is not worth all the complexity that is introduced.
* DX is improved - developer does not need to bother/remember/read in documentation that he needs to add "sylius" validation group to overridden entity in their custom app. He can just add constraint and be done with it.
* All Sylius constraints use proper group "sylius" and ``validation_groups`` config node is available for every single resource, thus we maintain the ability to completely rework validation.

I did that for Promotion only right now, but will do this for everything else if I get support from you guys. :) Let me know what do you think!

Will be easier to review once #3721 is merged. Perhaps wait with reviewing the code.


aRn0D:
@pjedrzejewski I think you're right! This feature is an overkill, I guest it is use rarely...

michalmarcinkowski:
I like cleanups! Thanks Paweł! :+1:

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
```diff
@@ -154,16 +154,15 @@ protected function addForms(ContainerBuilder $container, MetadataInterface $meta
...
-                        $validationGroups = new Parameter($validationGroupsParameterName);
-                    } else {
+                    if (!$container->hasParameter($validationGroupsParameterName)) {
                         $validationGroups = array('Default');
                     }
 
                     $definition->setArguments(array(
                         $metadata->getClass('model'),
-                        $validationGroups,
+                        $validationGroups
```
stloyd: IMO this should be reverted ;)

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
```diff
@@ -154,16 +154,15 @@ protected function addForms(ContainerBuilder $container, MetadataInterface $meta
 
                 default:
                     $validationGroupsParameterName = sprintf('%s.validation_groups.%s%s', $metadata->getApplicationName(), $metadata->getName(), $suffix);
+                    $validationGroups = new Parameter($validationGroupsParameterName);
```
stloyd: This should be on `} else {` IMO.

michalmarcinkowski: I somehow missed that file... IMO this should be changed to:
```php
$validationGroups = array('Default');

if ($container->hasParameter($validationGroupsParameterName) {
   $validationGroups = new Parameter($validationGroupsParameterName);
}
```

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
michalmarcinkowski: :+1:

PR #3733, opened 2015-12-16
TestKernel restores default values on services' properties

pamil:
The best solution I came into, however the following issues are still present:

- stateful classes that are not registered as services are stateful between scenarios
- `Sylius\Bundle\CoreBundle\Tests\MySqlDriver::$connection` is nulled at each kernel shutdown, forcing to create over a thousand of connections


pjedrzejewski:
The connections issue is a bit nasty but I am sure we will figure something out. Thanks Kamil!

PR #3734, opened 2015-12-17
[API] Add customer show api path

gonzalovilaseca:
| Q             | A
| ------------- | ---
| Bug fix?      | [yes]
| New feature?  | [no]
| BC breaks?    | [no]
| Deprecations? | [no]
| Fixed tickets |  
| License       | MIT
| Doc PR        | 

`sylius_api_customer_show` route doesn't exist, should we create it or just use `sylius_api_user_show`?

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/routing/customer.yml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/routing/customer.yml
@@ -1,0 +1,8 @@
+# This file is part of the Sylius package.
+# (c) Paweł Jędrzejewski
+
+sylius_api_customer_show:
+    path: /{id}
+    methods: [GET]
+    defaults:
+        _controller: sylius.controller.customer:showAction
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/routing/main.yml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/routing/main.yml
@@ -103,6 +103,10 @@ sylius_api_user:
     resource: @SyliusApiBundle/Resources/config/routing/user.yml
     prefix: /users
 
+sylius_api_customer:
+    resource: @SyliusApiBundle/Resources/config/routing/customer.yml
+    prefix: /customers
+
 sylius_api_setting:
     resource: @SyliusApiBundle/Resources/config/routing/setting.yml
     prefix: /settings


src/Sylius/Bundle/UserBundle/Resources/config/serializer/Model.User.yml
```diff
@@ -20,7 +20,7 @@ Sylius\Component\User\Model\User:
     relations:
         - rel: user
```
michalmarcinkowski: I think that proper fix will be to change this `rel` to customer, since we still do `id: expr(object.getCustomer().getId())`.

pjedrzejewski: :+1:

michalmarcinkowski: Or if you want the User id change to:
```yml
- rel: self
          href:
              route: sylius_api_user_show
              parameters:
                  id: expr(object.getId())
```

gonzalovilaseca: so should I leave the route as it was before? but the it has to be created in some routing.yml file as it doesn't exist.

michalmarcinkowski: Ok, so we have a bug :smile: 

Proper fix - add route for api customer show
Quick fix - remove this `rel`

Are you willing to fix this?

gonzalovilaseca: Sure, I found it bc it was breaking my app! I'll fix it during the day.

gonzalovilaseca:
I've added the route in the ``user.yml`` file, unless someone thinks it's better to create a ``customer.yml`` file

michalmarcinkowski:
@gonzalovilaseca it's better to create a `customer.yml` file since now the routing looks like `/users/customer/{id}`. The proper routing will be `/customers/{id}`.

gonzalovilaseca:
@michalmarcinkowski updated!

michalmarcinkowski:
Thank you Gonzalo! :+1:

PR #3732, opened 2015-12-16
Fix some deprecations

bendavies:
As per the title!

PR #3736, opened 2015-12-17
replace FQCNs with class constants

bendavies:
Here is the change i mentioned in https://github.com/Sylius/Sylius/pull/3732


src/Sylius/Bundle/AddressingBundle/spec/Form/EventListener/BuildAddressFormSubscriberSpec.php
```diff
@@ -34,12 +36,12 @@ function let(ObjectRepository $countryRepository, FormFactoryInterface $factory)
 
     function it_is_initializable()
     {
-        $this->shouldHaveType('Sylius\Bundle\AddressingBundle\Form\EventListener\BuildAddressFormSubscriber');
+        $this->shouldHaveType(BuildAddressFormSubscriber::class);
```
michalmarcinkowski: This change should be reverted. In a moment we write spec we can't use `BuildAddressFormSubscriber::class` because this class does not exist yet. What is more, we want to verify the whole namespace together with possible typos. This change makes this spec useless because it will always be true. 

Same thing for `implements` and in every other spec.

bendavies: Hey,

I'm not sure what you mean by `BuildAddressFormSubscriber` not existing. It does?

I see what you mean about the check being useless. I'll revert all similar `it_is_initializable` tests.
Are you sure the `implements` checks are also useless?

pamil: We can use these language construct even when class does not exist. PHP does not try to load the class, it only returns the FQCN.

pamil: Also, it makes renaming the class name much simplier, as we don't have to look for the class as for FQCN string, but IDE does its job.

bendavies: I'll leave it for now until you lot come to a consensus!

michalmarcinkowski: @pamil how you want to get FQCN of class which does not exist? 
Writing this:
```php
MyClass::class
```

Will not return the full namespace, because the class does not exist and the namespace is unknow. IMO `shouldHaveType` should always be string.

Regarding `implements` we can discuss it separately.

bendavies: @michalmarcinkowski when are you ever testing for a class that does not exist?

peteward: @michalmarcinkowski :+1: 

When you are using TDD / BDD methodology.

pamil: @michalmarcinkowski `MyClass::class` will always refer to `Current\Namespace\MyClass` unless you add a `use` statement.

Anyway, it also helps you to see whether the class (or interface) really exists as IDE also highlights unexisting classes.

michalmarcinkowski: @pamil exactly, so it will return `spec\Sylius\...\MyClass` what is definitely not the correct class namespace.

But in the specification stage we do not care if the class exist or not, phpspec will tell you that. Anyway the class we're about to implement shouldn't exist before we write the first spec, so using `::class` will not work, before that and if you introduce it after the class is implemented it is equal to removing this spec, because it will always pass, so doesn't change anything and is useless.

michalmarcinkowski:
Concluding discussion:

1. In spec for initialization of the class we should use **string**.
```php
function it_is_initializable()
{
    $this->shouldHaveType('Sylius\...\MyClass');
```

2. For 'external' classes (`it_extends`, `it_implements` etc.) we should use `OtherClass::class`.
```php
function it_implements_some_interface()
{
    $this->shouldImplement(SomeInterface::class);
}
```

@Sylius/core-team, @pjedrzejewski, @pamil please elaborate on this topic, so we have a clear convention.

aRn0D:
:+1:

pamil:
I don't like the idea that we would use `::class` everywhere, but not in this one step. `it_is_initializable` is present in all examples and generated by default, but it isn't any PhpSpec special method and it does not define the class to be instantiated. 

In fact, the SUT isn't initialized until some expectation isn't called (`$this->should*()`), so I'm almost sure that using `shouldHaveType` is just a convention. I can't see no disadvantages of using `::class` everywhere if we require PHP5.5 anyway, but there are some pros like faster class renaming and consistency in our codebase.

bendavies:
@pjedrzejewski any thoughts on the above? It would be good to get this merged asap, as it touches a LOT of code so pretty much needs a rebase and conflict fixing every time it goes behind master.

bendavies:
@pamil did you reach a decision?

pamil:
@bendavies among with @pjedrzejewski and @michalmarcinkowski we agreed on https://github.com/Sylius/Sylius/pull/3736#issuecomment-165732890 :)

bendavies:
@pamil ok thanks, i'll revert those.

bendavies:
@pamil done!

pjedrzejewski:
Awesome, thanks for your work Ben! I think we should mention that somewhere in our Behat Contributing Guide.

bendavies:
:+1:

PR #3646, opened 2015-11-28
[POC] UiBundle and AdminBundle

pjedrzejewski:
Hello folks!

I'd like to start a discussion about the required changes to SyliusWebBundle. This PR shows the direction I want to move towards. Let me know what do you think. This is a result of plenty thoughts and discussion but I will try to summarize the most important reasons:

* We don't have time and don't want to maintain a very custom admin interface

Yeah. We need to be realistic about how much work it is to build a nice theme. I don't think we have that many frontend devs in the community and definitely not enough time. So I want to avoid as much as possible work here. **AdminLTE** sounds like a perfect solution here.

* AdminLTE is MIT licensed, based on Bootstrap and available via Bower.

MIT license - perfect. Based on Bootstrap - perfect, easy to use and everyone knows it, even for backend developers, who don't enjoy fighting with html/css too much. Available via Bower - awesome, we don't have to commit it to the repository.

* AdminLTE has plenty of components/widgets ready - My POC is super, super simple, but the package has plenty of examples and ready to use templates - just have a look here - https://almsaeedstudio.com/preview. It will be very easily customizable!

* Fully responsive backend out-of-the-box. Yeps, pretty much without any effort, the backend will be really responsive and mobile friendly. What we have now is far from perfect and would require great deal of work.

OK, this is about AdminLTE - I hope everyone is convinced that this is the right choice. I was considering Materialize, but tried it and it requires much more work, pretty much the same as our current custom Bootstrap backend. AdminLTE has more out-of-the-box than we could implement in next 3-6 months.

---

What's up with the AdminBundle and UiBundle? Goals:

* Decouple Administration panel from the shop frontend

There are projects already which use Sylius as API and backend. We don't want to force anyone to include whole bundle with the frontend if he is not using it. Also, it allows for a nice and clean separation of Admin panel and Shop frontend. (yes, next step is ShopBundle and removing WebBundle :D)

* Allow to reuse Sylius UI and simpler installation of standalone bundles

There is plenty of projects which use just selected bundles. Would be great if they could benefit from using Sylius UI. Also, some macros/buttons/widgets/menus/forms could be shared between bundles. Right now there is no easy way to do that, that's why I come up with the idea of SyliusUiBundle. It will contain all the AdminLTE integration stuff and common templating tools.

[...]

PR #3739, opened 2015-12-18
Remove useless default templates

pjedrzejewski:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

Maintaining default templates is too much work and really hard to get it right. We will provide UiBundle (see #3646), which will drastically reduce setup time for most of bundles. (templates for CRUD)


peteward:
:+1: 

The more templates, the longer it takes to warm-up cache, so getting rid of any unnecessary ones is a very good thing.

michalmarcinkowski:
Perfect! Thanks Paweł!

Issue #3698, opened 2015-12-09
SyliusFlowBundle deprecation notice for Symfony 2.8


andreopmaria:
I use Symfony 2.8 and try to prepare for 3.0, watching deprecated notes.

The Process\SyliusScenario class extends Symfony\Component\DependencyInjection\ContainerAware that is deprecated since version 2.8, to be removed in 3.0. Use the ContainerAwareTrait instead

gabiudrescu:
I have tried a pull request based on your issue. Hope its fine. 

Also, I have a question: can you tell me the link where I can see the deprecation notice for this?

michalmarcinkowski:
The FlowBundle was removed.

PR #3743, opened 2015-12-18
using ContainerAwareTrait

gabiudrescu:
| Q             | A
| ------------- | ---
| Bug fix?      | [no]
| New feature?  | [no]
| BC breaks?    | [no]
| Deprecations? | [yes]
| Fixed tickets | [#3698]
| License       | MIT
| Doc PR        | [n/a]

What bothers me is that even if I remove "implements ContainerAwareInterface" from Sylius\Bundle\FlowBundle\Process\Step\AbstractContainerAwareStep phpspec still tells me everything is fine.

Is this the expected behaviour?

### DIFF
--- a/src/Sylius/Bundle/FlowBundle/Process/Step/AbstractContainerAwareStep.php
+++ b/src/Sylius/Bundle/FlowBundle/Process/Step/AbstractContainerAwareStep.php
@@ -12,7 +12,7 @@
 namespace Sylius\Bundle\FlowBundle\Process\Step;
 
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 
 /**
  * Container aware step.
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
@gabiudrescu Yes. Symfony container does not check interfaces, it just calls setContainer and that's it but of course we should keep the interface. Thanks for the patch! @Sylius/core-team Good to go if green.

michalmarcinkowski:
Thank you Gabi! :+1:

gabiudrescu:
@pjedrzejewski that's good. I was thinking that maybe there's a test explicitly validating that AbstractContainerAwareStep is implementing ContainerAwareInterface or maybe there's a typehint somewhere for ContainerAwareInterface.

do you think I should keep the ContainerAwareInterface or remove it?

aRn0D:
:+1:

pjedrzejewski:
Definitely should be kept. We should not rely on traits but on interfaces. :)

gabiudrescu:
@pjedrzejewski I don't know if I am being more catholic than the Pope, but I added this: https://github.com/Sylius/Sylius/pull/3751

Issue #3748, opened 2015-12-18
[ResourceBundle] Can't ignore translation config.


liverbool:
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ResourceBundle/DependencyInjection/Configuration.php#L82-L113

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ResourceBundle/DependencyInjection/SyliusResourceExtension.php#L57-L68

Now `resource-bundle` enforce to config `translation` part. I think it's not comfortable using.

cc/ @pjedrzejewski

pjedrzejewski:
I will check it but I tried it with custom/standalone resource without translation and worked fine.

pjedrzejewski:
Do you get an error? It is optional. I actually want to create an RFC about merging Translation into Resource (same for bundles) because it is a difficult task to decouple them properly, so we don't have it right at the moment, yet we get all the overhead with maintenance of 4 packages instead of 2.

liverbool:
This code block will never `false`.
```php
if (isset($resourceConfig['translation'])) {
    ....
}
```

case of


```php
->arrayNode('translation')
    ->addDefaultsIfNotSet()
    ->children()
    ....
```

To fix it for now i just remove `->addDefaultsIfNotSet()`.

```php
->arrayNode('translation')
    ->children()
    ....
```

bendavies:
ref https://github.com/Sylius/Sylius/pull/3749

nakashu:
@liverbool @pjedrzejewski got the same issue, liverbool fix work for me also.
-

liverbool:
:+1: solved, Thanks you all ***Syliusers***.

michalmarcinkowski:
Thank you guys for reporting and fixing! :+1:

PR #3749, opened 2015-12-18
[ResourceBundle] check translations are possible before configuration them

bendavies:
I am not using translations in my application, and an error is thrown here as I've not configured translations for my resources.
This was introduced by https://github.com/Sylius/Sylius/pull/3721.

Also, it seems it is not not possible to disable translations per resource, see [here](https://github.com/Sylius/Sylius/pull/3721/files#r48045552).

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/DependencyInjection/Configuration.php
+++ b/src/Sylius/Bundle/ResourceBundle/DependencyInjection/Configuration.php
@@ -80,7 +80,6 @@ class Configuration implements ConfigurationInterface
                                 ->end()
                             ->end()
                             ->arrayNode('translation')
-                                ->addDefaultsIfNotSet()
                                 ->children()
                                     ->arrayNode('classes')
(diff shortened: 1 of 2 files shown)


bendavies:
just noticed this is related to https://github.com/Sylius/Sylius/issues/3748

liverbool:
I thinks it's not comprehensive fix as you said 

> **"it seems it is not not possible to disable translations per resource"**

It still can't disable translation per resource.

bendavies:
yes there are 2 issues here.
1. translations should not be mandatory (not fixed here)
2. don't try to configure translations at all if the Translation bundle is not available

pjedrzejewski:
Sounds good until we figure out what to do with coupling. :) Thank you very much Ben! :+1:

bendavies:
thanks @pjedrzejewski

pjedrzejewski:
@bendavies Let me know what do you think in #3750.

PR #3751, opened 2015-12-18
adding a test ...

gabiudrescu:
...to ensure AbstractContainerAwareStep is implementing ContainerAware interface

### DIFF
--- a/src/Sylius/Bundle/FlowBundle/Process/Step/AbstractContainerAwareStep.php
+++ b/src/Sylius/Bundle/FlowBundle/Process/Step/AbstractContainerAwareStep.php
@@ -12,7 +12,7 @@
 namespace Sylius\Bundle\FlowBundle\Process\Step;
 
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 
 /**
  * Container aware step.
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/FlowBundle/spec/Process/ProcessSpec.php
```diff
@@ -28,6 +28,11 @@ function it_is_a_process()
         $this->shouldImplement('Sylius\Bundle\FlowBundle\Process\ProcessInterface');
     }
 
+    function it_is_container_aware()
+    {
+        $this->shouldImplement('Symfony\Component\DependencyInjection\ContainerAwareInterface');
```
tristanbes: isn't that deprecated ? Isn't the new recommended thing is to use the ContainerAwareTrait ?

pamil: I've just added this to 2.8 upgrade list, only abstract class `ContainerAware` is deprecated, symfony/symfony#17061.

pjedrzejewski:
Could you take a look at the failing build? Also, you can use ``::class``.

gabiudrescu:
yeah, now I realise it failed because src/Sylius/Bundle/FlowBundle/spec/Process/ProcessSpec.php is a spec file for src/Sylius/Bundle/FlowBundle/Process/Process.php not for src/Sylius/Bundle/FlowBundle/Process/Step/AbstractStep.php or src/Sylius/Bundle/FlowBundle/Process/Step/AbstractContainerAwareStep.php

shouldn't there be a StepSpec.php?

PR #3755, opened 2015-12-21
[Product] added availableTo parameter

starspire:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Sometimes except availableOn product parameter we need also availableTo one. This is especially usefull when You selling something like concert or training tickets which events take place at particular day and it is useless to selling them after that day. When You leave this parameter blank nothing new happens but when You set a date, Sylius will validate it with availableOn in isAvailable method.

This is my first bigger commit to Sylius so any tips how to improve my code are highly appreciated. Hope it will be usefull for community.

### DIFF
--- a/src/Sylius/Component/Product/Model/DateRange.php
+++ b/src/Sylius/Component/Product/Model/DateRange.php
@@ -1,0 +1,86 @@
+<?php
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Product\Model;
+
+/**
+ * @author Krzysztof Wędrowicz <krzysztof@wedrowicz.me>
+ */
+class DateRange
+{
+    /**
+     * @var \DateTime
+     */
+    private $start;
+
+    /**
+     * @var \DateTime
+     */
+    private $end;
+
+    /**
+     * @param \DateTime $start
+     * @param \DateTime $end
+     */
+    public function __construct(\DateTime $start = null,\DateTime $end = null)
+    {
+        $this->start = $start;
+        $this->end = $end;
+    }
+
+    /**
+     * Checks whether current datetime meets two requirements:
+     * is greater than $start or equal,
+     * is smaller than $end or equal.
+     * Checks only these conditions which values ($start, $end) are not null.
+     *
+     * @return bool
+     */
+    public function isInRange()
+    {
+        if($this->start > new \DateTime() || ($this->end !== null && $this->end < new \DateTime())){
+            return false;
+        }
+        return true;
+    }
+
+    /**
+     * @param \DateTime $start
+     */
+    public function setStart($start)
+    {
+        $this->start = $start;
+    }
+
+    /**
+     * @param \DateTime $end
+     */
+    public function setEnd($end)
+    {
+        $this->end = $end;
+    }
+
+    /**
+     * @return \DateTime
+     */
+    public function getStart()
+    {
+        return $this->start;
+    }
+
+    /**
+     * @return \DateTime
+     */
+    public function getEnd()
+    {
+        return $this->end;
+    }
+}
+
--- a/src/Sylius/Component/Product/Model/DateRange.php
+++ b/src/Sylius/Component/Product/Model/DateRange.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * This file is part of the Sylius package.
  *
@@ -13,7 +14,7 @@ namespace Sylius\Component\Product\Model;
 /**
  * @author Krzysztof Wędrowicz <krzysztof@wedrowicz.me>
  */
-class DateRange
+final class DateRange
 {
     /**
      * @var \DateTime
@@ -29,7 +30,7 @@ class DateRange
      * @param \DateTime $start
      * @param \DateTime $end
      */
-    public function __construct(\DateTime $start = null,\DateTime $end = null)
+    public function __construct(\DateTime $start = null, \DateTime $end = null)
     {
         $this->start = $start;
         $this->end = $end;
@@ -54,7 +55,7 @@ class DateRange
     /**
      * @param \DateTime $start
      */
-    public function setStart($start)
+    public function setStart(\DateTime $start = null)
     {
         $this->start = $start;
     }
@@ -62,7 +63,7 @@ class DateRange
     /**
      * @param \DateTime $end
      */
-    public function setEnd($end)
+    public function setEnd(\DateTime $end = null)
     {
         $this->end = $end;
     }
--- a/src/Sylius/Component/Product/Model/Product.php
+++ b/src/Sylius/Component/Product/Model/Product.php
@@ -41,6 +41,11 @@ class Product extends AbstractTranslatable implements ProductInterface
     protected $availableOn;
 
     /**
+     * @var \DateTime
+     */
+    protected $availableUntil;
+
+    /**
      * @var Collection|BaseAttributeValueInterface[]
      */
     protected $attributes;
@@ -189,7 +194,7 @@ class Product extends AbstractTranslatable implements ProductInterface
      */
     public function isAvailable()
     {
-        return new \DateTime() >= $this->availableOn;
+        return (new DateRange($this->availableOn, $this->availableUntil))->isInRange();
     }
 
     /**
@@ -206,6 +211,22 @@ class Product extends AbstractTranslatable implements ProductInterface
     public function setAvailableOn(\DateTime $availableOn = null)
     {
         $this->availableOn = $availableOn;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getAvailableUntil()
+    {
+        return $this->availableUntil;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setAvailableUntil(\DateTime $availableUntil = null)
+    {
+        $this->availableUntil = $availableUntil;
     }
 
     /**
--- a/src/Sylius/Component/Product/Model/ProductInterface.php
+++ b/src/Sylius/Component/Product/Model/ProductInterface.php
@@ -51,6 +51,4 @@ interface ProductInterface extends
      * @param null|\DateTime $availableUntil
      */
     public function setAvailableUntil(\DateTime $availableUntil = null);
-
-
 }
--- a/src/Sylius/Component/Product/Model/ProductInterface.php
+++ b/src/Sylius/Component/Product/Model/ProductInterface.php
@@ -41,4 +41,16 @@ interface ProductInterface extends
      * @param null|\DateTime $availableOn
      */
     public function setAvailableOn(\DateTime $availableOn = null);
+
+    /**
+     * @return \DateTime
+     */
+    public function getAvailableUntil();
+
+    /**
+     * @param null|\DateTime $availableUntil
+     */
+    public function setAvailableUntil(\DateTime $availableUntil = null);
+
+
 }
--- a/src/Sylius/Component/Product/Model/Variant.php
+++ b/src/Sylius/Component/Product/Model/Variant.php
@@ -24,6 +24,11 @@ class Variant extends BaseVariant implements VariantInterface
      */
     protected $availableOn;
 
+    /**
+     * @var \DateTime
+     */
+    protected $availableUntil;
+
     public function __construct()
     {
         parent::__construct();
@@ -52,7 +57,7 @@ class Variant extends BaseVariant implements VariantInterface
      */
     public function isAvailable()
     {
-        return new \DateTime() >= $this->availableOn;
+        return (new DateRange($this->availableOn, $this->availableUntil))->isInRange();
     }
 
     /**
@@ -72,6 +77,26 @@ class Variant extends BaseVariant implements VariantInterface
 
         if ($this->isMaster() && null !== $this->object) {
             $this->getProduct()->setAvailableOn($availableOn);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getAvailableUntil()
+    {
+        return $this->availableUntil;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setAvailableUntil(\DateTime $availableUntil = null)
+    {
+        $this->availableUntil = $availableUntil;
+
+        if ($this->isMaster() && null !== $this->object) {
+            $this->getProduct()->setAvailableUntil($availableUntil);
         }
     }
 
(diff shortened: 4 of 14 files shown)


src/Sylius/Component/Product/Model/Product.php
```diff
@@ -189,7 +194,7 @@ public function setMetaDescription($metaDescription)
      */
     public function isAvailable()
     {
-        return new \DateTime() >= $this->availableOn;
+        return new \DateTime() >= $this->availableOn && (!$this->availableTo || $this->availableTo >= new \DateTime());
```
tristanbes: you should invert your comparaison order and make it like: 
```
new \DateTime() <= $this->availableTo;
```

to keep the same logic with the check before

src/Sylius/Component/Product/Model/Variant.php
```diff
@@ -52,7 +57,7 @@ public function setProduct(ProductInterface $product = null)
      */
     public function isAvailable()
     {
-        return new \DateTime() >= $this->availableOn;
+        return new \DateTime() >= $this->availableOn && (!$this->availableTo || $this->availableTo >= new \DateTime());
```
tristanbes: same remark as before

pamil:
It would be great to have some `DateRange` VO that can tell us if date is between start and end date.

starspire:
@pamil Did you mean a simple class called DateRange, with 2 parameters in constructor and boolean methods like isInRange() ?

pamil:
Something like that, another idea is to not require these dates, as no date would mean no restriction. So empty `DateRange` would match every other date, `DateRange` with start date only would match every date past the start date etc.

It's not essential to do it in this PR, but I think it would be a nice addition and it would make our codebase cleaner (especially when we can use Doctrine ORM 2.5 with embeddables support :)).

src/Sylius/Component/Product/Model/Variant.php
starspire: @tristanbes Fixed.

starspire:
@pamil I can write it Today, just tell me when I should to put this class, because I'm not sure now.

aramalipoor:
Maybe `availableUntil` is more convenient here, I thought you're PR is about restricting a product **to** some users or groups ;)

starspire:
@aramalipoor You are right, I'll change it.

pamil:
@starspire as we don't have any common package shared in all other Sylius packages, let's put it in Product component and bundle (if needed), it'd be easy to move it somewhere else if there is a need for it.

starspire:
@pamil DateRange class added. I think the whole work is done now :)

app/migrations/Version20151221151112.php
```diff
@@ -0,0 +1,40 @@
...
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('DROP INDEX fulltext_search_idx ON sylius_search_index');
+        $this->addSql('CREATE INDEX fulltext_search_idx ON sylius_search_index (item_id)');
+        $this->addSql('ALTER TABLE sylius_product_variant CHANGE available_to available_until DATETIME DEFAULT NULL');
```
aramalipoor: I think you don't need this migration when you have the other one. Just change `available_to` to `available_util` in previous migration.

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,41 @@
+<?php
+/**
+ * User: Krzysztof Wędrowicz krzysztof@wedrowicz.me
+ * Date: 21.12.15
+ * Time: 15:41
+ */
```
aramalipoor: Fix copyright, do it like other files ;)

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,41 @@
...
+     * @var \DateTime
+     */
+    protected $availableOn;
+
+    /**
+     * @var \DateTime
+     */
+    protected $availableUntil;
+
+
```
aramalipoor: Extra space.

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,41 @@
...
+    public function __construct($availableOn = null, $availableUntil = null)
+    {
+        $this->availableOn = $availableOn;
+        $this->availableUntil = $availableUntil;
+    }
+
+    public function isInRange()
+    {
+        $now = new \DateTime();
+        if($this->availableOn && $this->availableUntil)
```
aramalipoor: Use `{` and `}`

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    protected $availableUntil;
+
+
+    public function __construct($availableOn = null, $availableUntil = null)
+    {
+        $this->availableOn = $availableOn;
+        $this->availableUntil = $availableUntil;
+    }
+
```
aramalipoor: Missing doc here for `isInRange`

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+namespace Sylius\Component\Product\Model;
+
+class DateRange
+{
+    /**
+     * @var \DateTime
+     */
+    protected $availableOn;
```
aramalipoor: I think it's better to name this `$start`

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * @var \DateTime
+     */
+    protected $availableUntil;
```
aramalipoor: I think it's better to name this `$end`

src/Sylius/Component/Product/Model/VariantInterface.php
```diff
@@ -42,4 +42,14 @@ public function getAvailableOn();
...
+
+    /**
+     * @return \DateTime
+     */
+    public function getAvailableUntil();
+
+    /**
+     * @param null|\DateTime $availableUntil
+     */
+    public function setAvailableUntil(\DateTime $availableUntil = null);
```
aramalipoor: I think it's not right to have default value of null here!? When are you calling it with no arguments? If you want to set to null you just `set(null)`

aramalipoor:
@pamil Is it possible to have a `DateRange` kind of thing in only **one** field? That would be a useful thing if possbile ;)

src/Sylius/Component/Product/Model/VariantInterface.php
starspire: I just followed the convention above. There was already setAvailableOn with default null value.

starspire:
@aramalipoor Thanks for your tips, the code was corrected.

starspire:
@pjedrzejewski Could you take a look if this PR is ready to merged? I think it has been done.

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
```
pjedrzejewski: Missing license block.

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * @author Krzysztof Wędrowicz <krzysztof@wedrowicz.me>
+ */
+class DateRange
+{
+    /**
+     * @var \DateTime
+     */
+    protected $start;
```
pjedrzejewski: Should be private.

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * @var \DateTime
+     */
+    protected $end;
+    
+    public function __construct($start, $end)
```
pjedrzejewski: Missing docblock.

pjedrzejewski: And typehints.

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * Checks whether current datetime is between start and end.
+     *
+     * @return bool
+     */
+    public function isInRange()
+    {
+        $now = new \DateTime();
```
pjedrzejewski: CS and I believe this could be a bit cleaner.

pjedrzejewski:
Is there really no core php class that we could use? I don't think Product component is the right place for such generic class.

starspire:
@pjedrzejewski I think there isn't. At least php.net (http://php.net/manual/en/book.datetime.php) doesn't know anything about it. DatePeriod is also something different from what we want. Eventually I found http://period.thephpleague.com/ but also this isn't strictly what we needed (but it has boolean contains() and overlaps() methods). I agree that current location isn't the most fortune one, but was suggested by @pamil to me, as the best of weak choices. I'm with pleasure make this in the way you decide but must to know how :)

starspire:
@pjedrzejewski How about now? I followed your notices but nothings done in this PR for a week and because I already need this in my project I'd like to ask if I can still improve something or it is good enough to merge.

pjedrzejewski:
@starspire That's a bit late from my side, but I did not notice that there are no specs. Would be cool if you can add them, do you know how to write them?

starspire:
@pjedrzejewski I'm not mastering it, but think that can write some looking on examples of others. Will be in a few hours :)

starspire:
@pjedrzejewski Done. Could you see why Travis throw Segmentation fault in PHP7?

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * @var \DateTime
+     */
+    private $end;
+
+    /**
+     * @param \DateTime $start
+     * @param \DateTime $end
+     */
+    public function __construct($start, $end)
```
lchrusciel: Missing typehint.

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /**
+     * @return \DateTime
+     */
+    public function getEnd()
+    {
+        return $this->end;
+    }
+}
```
tristanbes: missing EOF new lien

starspire: Added.

starspire:
@pjedrzejewski @pamil @aramalipoor @tristanbes @lchrusciel All of your suggestions were used :) Is there anything else to improve?

pjedrzejewski:
@starspire Nice, it is good to go now but I have to ask you for last little thing. :) Could you please squash your commits? 14 is a bit too much for such small change. Please squash it to a single commit.

https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/

Thanks!

starspire:
@pjedrzejewski Done. Everything passed :)

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,86 @@
+<?php
+/*
```
stloyd: Missing new line before this comment.

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,86 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Product\Model;
+
+/**
+ * @author Krzysztof Wędrowicz <krzysztof@wedrowicz.me>
+ */
+class DateRange
```
stloyd: IMO this class should be `final`.

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * @var \DateTime
+     */
+    private $end;
+
+    /**
+     * @param \DateTime $start
+     * @param \DateTime $end
+     */
+    public function __construct(\DateTime $start = null,\DateTime $end = null)
```
stloyd: Missing space after `,`.

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,86 @@
...
+     * Checks whether current datetime meets two requirements:
+     * is greater than $start or equal,
+     * is smaller than $end or equal.
+     * Checks only these conditions which values ($start, $end) are not null.
+     *
+     * @return bool
+     */
+    public function isInRange()
+    {
+        if($this->start > new \DateTime() || ($this->end !== null && $this->end < new \DateTime())){
```
stloyd: This should be splitted, and this is wrong as `start` also can be `null`.

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,86 @@
...
+            return false;
+        }
+        return true;
+    }
+
+    /**
+     * @param \DateTime $start
+     */
+    public function setStart($start)
```
stloyd: Missing typehint.

src/Sylius/Component/Product/Model/DateRange.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $this->start = $start;
+    }
+
+    /**
+     * @param \DateTime $end
+     */
+    public function setEnd($end)
```
stloyd: Missing typehint.

src/Sylius/Component/Product/Model/ProductInterface.php
```diff
@@ -41,4 +41,16 @@ public function getAvailableOn();
...
+     * @return \DateTime
+     */
+    public function getAvailableUntil();
+
+    /**
+     * @param null|\DateTime $availableUntil
+     */
+    public function setAvailableUntil(\DateTime $availableUntil = null);
+
+
```
stloyd: Those two new lines should be removed.

src/Sylius/Component/Product/Model/DateRange.php
starspire: Nope, because null cant be greater than anything so this condition would be redundant :)

pjedrzejewski:
Thanks Krzysztof! :+1:

src/Sylius/Component/Product/Model/DateRange.php
lchrusciel: Anyway we used to use yoda style comparison: ``null !== $this->end``

src/Sylius/Component/Product/Model/DateRange.php
lchrusciel: Too many blank lines :P

src/Sylius/Component/Product/Model/DateRange.php
tristanbes: missing whitespace right after `ìf`

PR #3759, opened 2015-12-21
[Installer] Allow easier extend of SetupCommand for package users

aramalipoor:
Q  | A
------------- | -------------
Bug fix?  | no
New feature?  | no
BC breaks? | no
Deprecations? | no
Fixed tickets | no
License | MIT
Doc PR | -

### DIFF
--- a/src/Sylius/Bundle/InstallerBundle/Command/SetupCommand.php
+++ b/src/Sylius/Bundle/InstallerBundle/Command/SetupCommand.php
@@ -26,6 +26,26 @@ use Symfony\Component\Validator\Constraints\NotBlank;
 class SetupCommand extends AbstractInstallCommand
 {
     /**
+     * @var string
+     */
+    protected $defaultCountries = 'US';
+
+    /**
+     * @var string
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
We need to get rid of languages, countries etc. setup for installer command. Only admin user should be added there and rest should be configured in the backend. I will leave this open, but I think this is a bigger issue than just allowing to override these. Anyway, thanks for the PR, Aram! :+1: 

Ideal scenario:

* Run sylius:install - check requirements, create/update database, create admin user -> print link  to ``/admin/setup`` or just ``/admin``
* User goes to ``/admin`` and sees a notification "Hey, setup your first sales channel" etc.

I'd love to have ``/admin/setup`` but I think this is not a priorty right now.

In ``/admin/setup`` we would have presets, where you just fill to which countries you sell and everything is preconfigured.

michalmarcinkowski:
Closing as it is no longer relevant after https://github.com/Sylius/Sylius/pull/4865.

PR #3764, opened 2015-12-22
Created phpspec.yml.dist, improved .gitignore

pamil:
- [x] Created `phpspec.yml.dist` to allow local overwriting
- [x] Improved `.gitignore` to contain absolute paths instead of relative ones


pjedrzejewski:
Thank you Kamil! :+1:

PR #3768, opened 2015-12-22
fix incorrect import

bendavies:
Added a test to prevent regression for `ResourceChoiceType`.
`AbstractResourceBundle` isn't covered by a test so no test added there.


pjedrzejewski:
Thanks Ben!

Issue #3441, opened 2015-10-15
[RFC] Reworking Adjustments to work at the InventoryUnit level


peteward:
We have run into an issue with OrderItem-level Adjustments. They create a lot of complexity when trying to cope with the following:

* **Refunds** - This is not core Sylius yet, but when a customer returns an item for a refund, they are very likely to return less than the full quantity of the OrderItem.  As Adjustments are at the OrderItem level, if you have promotional adjustments these need to be 'reversed' only for the quantity that is being returned. While this is doable it gets complex quite quickly.
* **Financial Reporting at an InventoryUnit level**. This is a very common requirement for any enterprise system that you might want to integrate with (we have several). If you want to know how much a single sold unit is worth after adjustments you have to consider the quantity. If you add in refunds here this becomes incredibly difficult to quantify how much an item was worth.

I’ve done a review of several other Open Source commerce systems and none of them seem to deal with this particularly well – we think we can make Sylius stand out in this respect.

Our proposal is to handle Adjustments for OrderItems at the InventoryUnit level, but to keep the relationship with OrderItem in such a way that it still works without a huge BC break. We are suggesting quite a big fundamental change, but it shouldn’t break how Adjustments are used in existing reports.

We will add the following to the Adjustment entity:
* `inventoryUnitId`
* `isRefund`

To illustrate how this would look in a cut-down database table:
* There is an OrderItem with id 7, it has a quantity of 3 with a unitPrice of $50.
* It has a 20% discount promotion on this item which is applied.
* Tax is calculated and applied afterwards.

![screen shot 2015-10-15 at 17 36 44](https://cloud.githubusercontent.com/assets/7068515/10520674/58e22280-7363-11e5-8505-9f56c76790df.png)

* We record every adjustment (above shows promotion and tax) at the InventoryUnit level, but also keep the relation with the OrderItemId. This means that existing methods that just add up adjustments on OrderItem still work, such as:

```
/**
 * {@inheritdoc}
 */
public function calculateAdjustmentsTotal()
{
    $this->adjustmentsTotal = 0;

    foreach ($this->adjustments as $adjustment) {
        if (!$adjustment->isNeutral()) {
            $this->adjustmentsTotal += $adjustment->getAmount();
        }
    }

    return $this;
}
```

* It also means that listing adjustments still works, although there will be more of them.
* Order-level adjustments are unaffected (shipping).

[...]

Issue #605, opened 2013-11-17
[Bug] Taxes don't take into account Promotions


winzou:
Currently, TaxationProcessor applies its taxes independently from promotions that could be already applied.

Example: 
- Items total = 100€
- Promotion total = 10€ (a coupon offering 10% discount)
- => Tax (20% / included in price) total = 20€

So the guys would be paying 90€ with 20€ taxes instead of 18€.

This is because taxes are applied on each item and finaly summed. But the promotion total is an adjustment in addition to the order total, no link to each item total.

What would be the solution?

Apply the promotion on tax total? What if each item has a different tax applied?

It's a tricky issue!

psyray:
What's going on about this issue ?
I think, [as explained by] (https://github.com/Sylius/Sylius/issues/605#issuecomment-29746437) by @simshaun , promotions need an option to define what tax behavior apply to fulfill each customer use case

patrick-mcdougle:
Furthermore, as an added complexity, if you ever want to be able to ship to multiple destinations in a single order, tax needs to be applicable on a shipment level as well because in the US sales tax is calculated on shipping address.

michalmarcinkowski:
It is already fixed and covered with Behat scenarios (see https://github.com/Sylius/Sylius/pull/7187/files).

PR #3771, opened 2015-12-23
[RFC][Order] Introduce OrderItemUnit

Zales0123:
Hello everybody!

This is a Proof of Concept PR that illustrates idea how to resolve issues with how adjustments (thus taxes, shipping fees and discounts) are handled on the Order model. It is based on excellent work and research of Peter from REISS.

See corresponding PRs/Issues:
- https://github.com/Sylius/Sylius/pull/3712
- https://github.com/Sylius/Sylius/issues/3441
- https://github.com/Sylius/Sylius/issues/605

Overview of how Adjustments are handled currently:
- every adjustment can be set on Order or OrderItem
- each tax/promotion/etc. can be applied on these two levels

Full rationale for why this is wrong is available in the issues listed above.

**Solution:**
To fix this problem, we’ve created completely new ``OrderItemUnit``. In this highly preliminary PR, there is a proper model and interface introduced and used as ``InventoryUnit`` and ``ShipmentItem`` in ``sylius.yml``.

The logic of splitting OrderItem’s quantity into separate OrderItemUnits will be part of Order Component. In Core, to avoid duplication and for performance reasons, ``OrderItemUnit`` will implement ``ShipmentItemInterface`` and ``InventoryUnitInterface``, thus our current logic regarding shipping and inventory can remain untouched.

Having OrderItemUnits will allow us to apply adjustments more precisely and get the financials right. We can also provide two implementations for any service, like 2 TaxationProcessors, one which applies taxes on OrderItemUnit level (default) and optional for simpler use-cases that applies taxes to Order level.

There is plenty of work with adjustments to make them perfect and suitable for modern e-commerce system which Sylius obviously is, but it can be first, important step to do this. When we do this, we also want to finally clean up the way that totals are calculated and remove this non-sense with ``setTotal()`` method. :)

Let me hear your thoughts!

**UPDATE:**
Related docs PR: https://github.com/Sylius/Sylius-Docs/pull/394


src/Sylius/Bundle/InventoryBundle/Resources/config/doctrine/model/InventoryUnit.orm.xml
```diff
@@ -24,7 +24,7 @@
 
         <field name="inventoryState" column="inventory_state" type="string" />
 
-        <many-to-one field="stockable" target-entity="Sylius\Component\Inventory\Model\StockableInterface">
+        <many-to-one field="stockable" target-entity="Sylius\Component\Core\Model\ProductVariant">
```
pamil: `InventoryBundle` shouldn't know about `Core` :)

Zales0123: Sure it should not! I'm aware about some decoupling issues that can appear in this PR and I will be grateful for pointing them out - it will make my work much easier ;)

src/Sylius/Component/Core/Model/OrderItemInterface.php
```diff
@@ -56,4 +57,26 @@ public function removeInventoryUnit(InventoryUnitInterface $unit);
...
+     * @param OrderItemUnitInterface $itemUnit
+     */
+    public function removeItemUnit(OrderItemUnitInterface $itemUnit);
+
+    /**
+     * @param OrderItemUnitInterface $itemUnit
+     *
+     * @return bool
+     */
+    public function hasItemUnit(OrderItemUnitInterface $itemUnit);
```
pamil: I don't know if there is any convention about ordering methods that refer to the same resource (like `get` and `set`), but I prefer queries first (`get`, `has`, `is`) and then commands (`set`, `add`, `remove`). Of course this is just my opinion :)

src/Sylius/Component/Core/Model/OrderItemUnit.php
```diff
@@ -0,0 +1,214 @@
...
+use Sylius\Component\Shipping\Model\ShipmentInterface as BaseShipmentInterface;
+use Sylius\Component\Shipping\Model\ShippableInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class OrderItemUnit extends BaseOrderItemUnit implements OrderItemUnitInterface
+{
+    /**
+     * @var string
```
pamil: `@var string InventoryUnitInterface::STATE_*` like you did below?

src/Sylius/Component/Order/Model/Adjustment.php
```diff
@@ -105,6 +110,10 @@ public function getAdjustable()
             return $this->orderItem;
         }
 
+        if (null !== $this->orderItemUnit) {
+            return $this->orderItemUnit;
+        }
+
         return null;
```
pamil: Proposition of hacky hack:

```php
return $this->order || $this->orderItem || $this->orderItemUnit;
```

Zales0123: :+1:

src/Sylius/Component/Order/Model/OrderItemUnitInterface.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Sylius\Component\Order\Model;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface OrderItemUnitInterface extends AdjustableInterface
+{
+    /**
+     * @return integer
```
pamil: `int` or even `mixed` here

src/Sylius/Bundle/InventoryBundle/Resources/config/doctrine/model/InventoryUnit.orm.xml
pjedrzejewski: This relation should be just removed from here.

src/Sylius/Bundle/OrderBundle/Resources/config/doctrine/model/Adjustment.orm.xml
```diff
@@ -45,6 +45,10 @@
         <many-to-one field="orderItem" target-entity="Sylius\Component\Order\Model\OrderItemInterface" inversed-by="adjustments">
             <join-column name="order_item_id" referenced-column-name="id" nullable="true" on-delete="CASCADE"/>
         </many-to-one>
```
peteward: Is the plan to keep adjustments at the orderItem level? Or is this a later phase?

pjedrzejewski: I think we should discuss it, but in the first step we don't want to do too many changes so they will stay in this PR. We perhaps remove them in second step. I'd like to keep them for the sake of simpler use-cases where applying adjustments at order item level is just fine. I can imagine something like this:

OrderTaxesApplier uses TaxesStrategyInterface

UnitLevelTaxesStrategy (used by default in Sylius)
ItemLevelTaxesStrategy

Anyway, let's do small, quick steps.

src/Sylius/Component/Order/Model/OrderItemUnit.php
```diff
@@ -0,0 +1,161 @@
...
+{
+    /**
+     * @var int
+     */
+    protected $id;
+
+    /**
+     * @var int
+     */
+    protected $total = 0;
```
peteward: What is `total` at the `OrderItemUnit` level?

At the `OrderItem` level it was (`quantity` * `unit_price`) + `adjustments_total`, which was calculated, likewise at the order level.

At the unit level what do we need here? Is there a `price` or `value` field that gets set based on the `OrderItem.unitPrice`? `total` would then be calculated based on `price` + `adjustmentsTotal`?

peteward: This would then need a `calculateTotal()` method.  You would also want to remove the setter for total as it should only be calculated?

src/Sylius/Component/Order/Model/OrderItemUnit.php
```diff
@@ -0,0 +1,161 @@
...
+
+        return $this->adjustments->filter(function (AdjustmentInterface $adjustment) use ($type) {
+            return $type === $adjustment->getType();
+        });
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function addAdjustment(AdjustmentInterface $adjustment)
```
peteward: Considering it's in at least 3 places (at the moment anyway), what about an `Adjustable` trait for these methods?

src/Sylius/Component/Order/Model/OrderItemUnit.php
```diff
@@ -0,0 +1,161 @@
...
+     */
+    public function getAdjustmentsTotal($type = null)
+    {
+        if (null === $type) {
+            return $this->adjustmentsTotal;
+        }
+
+        $total = 0;
+        foreach ($this->getAdjustments($type) as $adjustment) {
+            $total += $adjustment->getAmount();
```
peteward: what about neutrality here? (could be a copy-paste) but this would add neutral adjustments and this might be something optional?  I think if you want a total you probably don't want neutral adjustments included by default.

src/Sylius/Bundle/OrderBundle/Resources/config/doctrine/model/Adjustment.orm.xml
peteward: sounds good to come back to it later.

Although I think you'll find it gets unnecessarily complicated to support both and if you can provide 'rolling up' of values at the display level it won't be needed. 

We'll see...

src/Sylius/Component/Order/Model/OrderItemUnit.php
pjedrzejewski: Sounds good to me!

src/Sylius/Component/Order/Model/OrderItemUnit.php
pjedrzejewski: Yes, I think unit should have price and total should be price + adjustmentsTotal. Setter should be removed too and everything calculated inside.

src/Sylius/Component/Core/Model/OrderItemInterface.php
Zales0123: This time it's pretty sensible :smile:

src/Sylius/Component/Order/Model/OrderItemUnit.php
Zales0123: Agreed :+1:

peteward: it is... just wondering for being explicit if it's worth having it here too? For easy reporting? I guess joining on OrderItem isn't exactly hard work... and there's no chance that price could differ by unit, so yeah!

src/Sylius/Component/Order/Model/OrderItemUnit.php
Zales0123: Ok, after some discussion in team, we decided not to use trait in this case. It's because in next iteration, we're going to rewrite total calculation logic on all levels (for now in ``Order``, ``OrderItem`` and ``OrderItemUnit``) and it would be necessary to use ``calculateTotal()`` method inside some of adjustments-related functions. Therefore, this methods cannot be implemented in trait and implementing some "adjustements" methods in trait and some in specific class is even worse than code duplication :/

src/Sylius/Component/Order/Model/OrderItemUnit.php
pjedrzejewski: :+1:

src/Sylius/Component/Order/Model/OrderItemUnit.php
peteward: :+1:  Makes sense.

So where are we at with this PR? Is it ready for merge? I have no further comments at this time - it looks good to start with.

src/Sylius/Component/Order/Model/Adjustment.php
Zales0123: Ok, sadly :-1: After looking at this it would return bool, but we expect not-null object or ``null`` ;)

src/Sylius/Bundle/CoreBundle/Resources/config/serializer/Model.InventoryUnit.yml
```diff
@@ -1,6 +1,6 @@
-Sylius\Component\Core\Model\InventoryUnit:
+Sylius\Component\Core\Model\OrderItemUnit:
```
aramalipoor: File name should be `Model.OrderItemUnit.yml`

Zales0123: :+1:

src/Sylius/Component/Core/Model/OrderItemUnitInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Core\Model;
+
+use Sylius\Component\Inventory\Model\InventoryUnitInterface as BaseInventoryUnitInterface;
```
michalmarcinkowski: Do we need to make an alias here?

src/Sylius/Component/Core/OrderProcessing/InventoryHandler.php
```diff
@@ -67,19 +62,13 @@ public function __construct(
...
         if ($item->getQuantity() > $nbUnits) {
             $this->createInventoryUnits($item, $item->getQuantity() - $nbUnits);
         } elseif ($item->getQuantity() < $nbUnits) {
-            foreach ($item->getInventoryUnits()->slice(0, $nbUnits - $item->getQuantity()) as $unit) {
-                $item->removeInventoryUnit($unit);
-            }
-        }
-
-        foreach ($item->getInventoryUnits() as $unit) {
-            if ($unit->getStockable() !== $item->getVariant()) {
```
michalmarcinkowski: Why we remove this?

src/Sylius/Component/Core/OrderProcessing/InventoryHandler.php
```diff
@@ -132,12 +121,22 @@ public function updateInventory(OrderInterface $order)
...
+    /**
+     * @param OrderItemInterface $item
+     * @param int $quantity
+     * @param string $state
+     */
     protected function createInventoryUnits(OrderItemInterface $item, $quantity, $state = InventoryUnitInterface::STATE_CHECKOUT)
     {
-        $units = $this->inventoryUnitFactory->createForStockable($item->getVariant(), $quantity, $state);
+        if ($quantity < 1) {
+            throw new \InvalidArgumentException('Quantity of units must be greater than 1.');
```
michalmarcinkowski: greater than **0**

src/Sylius/Component/Inventory/Factory/InventoryUnitFactory.php
```diff
@@ -26,7 +24,7 @@ class InventoryUnitFactory extends Factory implements InventoryUnitFactoryInterf
     /**
      * {@inheritdoc}
      */
-    public function createForStockable(StockableInterface $stockable, $quantity, $state = InventoryUnitInterface::STATE_SOLD)
+    public function createForStockable($quantity, $state = InventoryUnitInterface::STATE_SOLD)
     {
         if ($quantity < 1) {
             throw new \InvalidArgumentException('Quantity of units must be greater than 1.');
```
michalmarcinkowski: greater than **0**

src/Sylius/Component/Core/OrderProcessing/InventoryHandler.php
michalmarcinkowski: greater than **0**

src/Sylius/Component/Inventory/Factory/InventoryUnitFactory.php
```diff
@@ -17,8 +17,6 @@
 use Sylius\Component\Resource\Factory\Factory;
 
 /**
- * Default inventory operator.
- *
  * @author Paweł Jędrzejewski <pawel@sylius.org>
  */
 class InventoryUnitFactory extends Factory implements InventoryUnitFactoryInterface
```
michalmarcinkowski: Do we need this class?

src/Sylius/Component/Core/spec/Model/OrderItemUnitSpec.php
```diff
@@ -0,0 +1,114 @@
...
+        $this->getSku()->shouldReturn('test_sku');
+    }
+
+    function it_can_be_sold_or_backorderded()
+    {
+        $this->setInventoryState(InventoryUnitInterface::STATE_SOLD);
+        $this->isSold()->shouldReturn(true);
+
+        $this->setInventoryState(InventoryUnitInterface::STATE_BACKORDERED);
+        $this->isBackordered()->shouldReturn(true);
```
michalmarcinkowski: I would add also ```$this->shouldNotBeSold();```

src/Sylius/Component/Core/spec/Model/OrderItemUnitSpec.php
```diff
@@ -0,0 +1,114 @@
...
+        $orderItem->getVariant()->willReturn($variant);
+        $this->setOrderItem($orderItem);
+
+        $this->getSku()->shouldReturn('test_sku');
+    }
+
+    function it_can_be_sold_or_backorderded()
+    {
+        $this->setInventoryState(InventoryUnitInterface::STATE_SOLD);
+        $this->isSold()->shouldReturn(true);
```
michalmarcinkowski: `$this->shouldBeSold()`

src/Sylius/Component/Core/spec/Model/OrderSpec.php
```diff
@@ -252,11 +274,75 @@ function it_not_a_backorder_if_contains_no_backordered_units(
...
     }
+
+    function it_adds_and_removes_payments(PaymentInterface $payment)
+    {
+        $payment->getState()->willReturn(PaymentInterface::STATE_PENDING);
+        $payment->setOrder($this)->shouldBeCalled();
+        $payment->setOrder(null)->shouldBeCalled();
+
+        $this->addPayment($payment);
+        $this->hasPayment($payment)->shouldReturn(true);
```
michalmarcinkowski: `$this->shouldHavePayment`

src/Sylius/Component/Core/OrderProcessing/InventoryHandler.php
Zales0123: Now ``OrderItemUnit::getStockable()`` is implemented as ``return $this->orderItem->getVariant()``, so there is no need to check if ``OrderItem`` variant is the same as ``OrderItemUnit`` stockable, cause it it always true.

src/Sylius/Component/Inventory/Factory/InventoryUnitFactory.php
Zales0123: I thought we're using it somewhere else, not in *CoreBundle*, but after quick search we don't, so sure, can be removed ;)

src/Sylius/Component/Core/spec/Model/OrderItemUnitSpec.php
Zales0123: I'm not a big fun of these magically named methods possible in specs, but of course, no problem with changing this (and other similar cases) ;)

src/Sylius/Component/Core/OrderProcessing/InventoryHandler.php
michalmarcinkowski: :+1:

michalmarcinkowski:
I think it's ready to merge! @Zales0123 please squash your commits, at least the last two :)

peteward:
:+1:

UPGRADE.md
```diff
@@ -1,6 +1,14 @@
 UPGRADE
 =======
 
+## From 0.16 to 0.17.x
+
+### Order and SyliusOrderBundle
+
+* Introduced ``OrderItemUnit``, which represents every single unit in ``Order``
+* Removed ``InventoryUnit``, as it's replaced by ``OrderItemUnit``, which can be used both as ``InventoryUnit`` and ``ShipmentItem``
```
pjedrzejewski: Replace ``InventoryUnit`` with ``OrderItemUnit`` in the core. This entity will be used as InventoryUnit and ShipmentItem;

UPGRADE.md
```diff
@@ -1,6 +1,14 @@
...
+* Replace each usage of ``InventoryUnit`` with ``OrderItemUnit``
```
pjedrzejewski: Reduntant.

UPGRADE.md
Zales0123: :+1:

src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius.yml
```diff
@@ -171,10 +171,7 @@ sylius_inventory:
     resources:
         inventory_unit:
             classes:
-                model: Sylius\Component\Core\Model\InventoryUnit
-        stockable:
-            classes:
-                model: %sylius.model.product_variant.class%
+                model: Sylius\Component\Core\Model\OrderItemUnit
```
pjedrzejewski: We should use a parameter here, ``%sylius.model.order_item_unit.class%. Same for ShipmentItem.

src/Sylius/Component/Core/Model/OrderInterface.php
```diff
@@ -173,71 +135,51 @@ public function getCurrency();
...
      */
     public function setExchangeRate($exchangeRate);
 
     /**
-     * Adds promotion coupon.
-     *
      * @param BaseCouponInterface $coupon
      */
-    public function addPromotionCoupon($coupon);
+    public function addPromotionCoupon(BaseCouponInterface $coupon = null);
```
pjedrzejewski: This does not look like a change relevant to this PR, why?

pjedrzejewski: Same for changes below.

src/Sylius/Component/Core/Model/OrderItemUnitInterface.php
```diff
@@ -0,0 +1,23 @@
...
+namespace Sylius\Component\Core\Model;
+
+use Sylius\Component\Inventory\Model\InventoryUnitInterface;
+use Sylius\Component\Order\Model\OrderItemUnitInterface as BaseOrderItemUnitInterface;
+use Sylius\Component\Shipping\Model\ShipmentItemInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface OrderItemUnitInterface extends BaseOrderItemUnitInterface, InventoryUnitInterface, ShipmentItemInterface
```
pjedrzejewski: :+1:

src/Sylius/Component/Core/Model/OrderInterface.php
pjedrzejewski: Ok, ignore these comments. :)

Zales0123: No problem with ignoring :smile:

pjedrzejewski:
Thank you Mateusz and the 2 additional brains behind making sure that Sylius handles $ well - @peteward and @michalmarcinkowski. :+1: Happy rebasing of related PRs. :D

peteward:
:+1: :+1: :+1: 

Great work all.

PWalkow:
Great to see that this one is on such a good path to be finished :)

PR #3779, opened 2015-12-27
[Shipment] Fix join on ShipmentRepository for Shipment Admin page

peteward:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

It could be specific to our app that shipments are created before addresses sometimes, but there's no point in this query being a left-join anyway since the `Backend/Shipment/macros.html.twig` file requires an address (and an order) for displaying shipments:

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/WebBundle/Resources/views/Backend/Shipment/macros.html.twig#L33


pjedrzejewski:
Makes sense, thanks Peter!

PR #3781, opened 2015-12-28
[WIP][Order] Order calculation refactoring

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

This PR is based on https://github.com/Sylius/Sylius/pull/3771

It's continuation of work started in PR mentioned above. Here I'm trying to get rid of absurd ``setTotal()`` methods on ``Order`` and ``OrderItem``, by slightly changing order/orderItem total calculation logic. It's now set after every significant change in object, such as:
- adding/removing item (in ``Order``)
- adding/removing unit (in ``OrderItem``)
- setting unit price (for units in ``OrderItem``)
- setting order item (in ``OrderItemUnit``)
- adding/removing adjustment (for all of them)

In the nearest future there should some kind of ``OrderItemUnitsProcessor`` (?) implemented, that will set order item units for every created ``OrderItem`` according to it's quantity.


src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
```diff
@@ -68,7 +68,7 @@ public function findForDetailsPage($id)
         $queryBuilder
             ->leftJoin('o.adjustments', 'adjustment')
             ->leftJoin('o.customer', 'customer')
-            ->leftJoin('item.inventoryUnits', 'inventoryUnit')
+            ->leftJoin('item.itemUnits', 'itemUnit')
```
michalmarcinkowski: Maybe simply `units`?

peteward:
There's a flaw in this approach (we tried this), you can't cascade up calculation calls as a result of changes to a child that aren't directly visible on the parent.

* If you change the quantity on an OrderItem, this won't trigger a call to calculateTotal on the order (unless you change quantity by a complete remove/add item).
* If you add / remove an adjustment at a child level it won't call calculate on the parent either - itemUnit => item => order. Again, unless you get were to do a complete remove and add, which would get crazy.
* it also doesn't cope with changes on external objects such as the a customer logging in or changing channel or currency (although if on the order these are not a problem).

It's also a bit less efficient if you're doing a lot of recalculations unnecessary in loops, although this is not really a big deal.

You need to think about all of the use-cases on which you need to recalculate the order and this doesn't cover them unfortunately, it covers some but not in such a way that will solve the problems.

The best idea we came up with was around unifying cart events (there were a couple of PRs and issues discussing this) and making sure they were called in a sensible order such that the order will definitely be recalculated at the end (and all children will cascade and recalculate too). If you combine this with these changes it might work.

An immediate example would be if you calculateTaxes and add them at the Unit level your OrderItem total won't recalculate and neither will your Order.

michalmarcinkowski:
@peteward take a look here https://github.com/Sylius/Sylius/pull/3786.

Regarding the issues you mentioned:
* We have a bidirectional relation, so we can easily trigger `calculateTotal` (or some better named method) of the parent (e.g. Order) after the child (e.g. OrderItem) has changed in a way that infuences total (take a look at OrderItemUnit in the above PR)
* I would like to rework the quantity management as described in the PR, but we will brainstorm on that beforehand.
* The rule is simple, when my total is influenced I will communicate to all interested objects to recalculate themselves. Changing something on the order level will trigger children (OrderItems) etc. The approach is the same as with listeners, but we do this on the model level.

michalmarcinkowski:
Closing in favor of https://github.com/Sylius/Sylius/pull/3786

PR #3786, opened 2015-12-29
[PoC] Order recalculation refactor

michalmarcinkowski:
This PR is based on #3771 and replaces https://github.com/Sylius/Sylius/pull/3781.

Please do not review the code as it's just first version of PoC which is intendent to demonstrate that this approach is practicable. The PR only includes proper communication between OrderItem and OrderItemUnit, but when we finish up the final approach we can easily do almost the same for Order and OrderItem. 

Two main points:

### Quantity

My idea is to remove the setQuantity method and do not set quantity directly, but through add/removeItemUnit (or add/removeUnit) method. Get quantity could be calculated then by just `count($units)`. We could have a separate service which will be responsible for handling all the logic regarding the quantity change. We would pass the quantity and the OrderItem to the method that will accordingly add/remove Units and take care of doing it properly (e.g. when we have few units available and few units backordered then during quantity decrease we would like to remove the backordered ones instead of available etc.). It will probably require a custom controller action for handling quantity changes and items add/remove that in the frontend, but we will figure it out once we get there :)

### Performance

Do not worry about many repeated recalculations, it will be improved in the next iteration. I have an idea of recalculating only the things that has changed in a very controlled way, tailored to each specific method.

PS.
@Sylius/core-team we can think of introducing Behat domain tests. I added them for my personal convenience, but it would be good to have functional tests on the domain level for such use cases in Sylius. There is no UI for this yet, so that's the only way we can validate proper communication between classes :)


peteward:
oooooh @michalmarcinkowski!

So simple, yet so clever, why didn't I think of setting the parent on the child!

:+1: :+1: :+1: for the idea.

src/Sylius/Component/Order/Model/OrderItem.php
```diff
@@ -131,175 +130,214 @@ public function setUnitPrice($unitPrice)
...
+        if ($this->total < 0) {
+            $this->total = 0;
         }
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function removeAdjustment(AdjustmentInterface $adjustment)
+    protected function getItemUnitsTotal()
```
peteward: Is there a policy on getters being protected? I prefer all getters to be public.  It seems like a slight contradiction in having `calculateAdjustmentsTotal` compared to `getItemUnitsTotal`. I know that calculate stores the value in the object so it's a bit different, but I'm not convinced on the naming?

michalmarcinkowski: No policy :smile:  I think we will introduce the `itemUnitsTotal` filed on the class and the getter will be public for sure :wink: naming and behavior of recalculation methods will be change greatly (ref. **Performance** point :)).

peteward:
As you asked I won't review code closely, just a couple of usability thoughts.

One of the requirements that we implemented in our iteration was the ability to 'roll up' adjustment totals (possibly of a particular type) for displays or reports. 

An example would be how much tax is there on this order? We might want to show this on the cart page, checkout, order confirmation emails, or the admin screen. This would need a parent -> grandchild totalling of all adjustments of that [optional] type.

Maybe not something that is required for the initial pass here, but worth thinking about.

Great work!

michalmarcinkowski:
@peteward totalling of all adjustments is on the roadmap of this PoC and should be easy to implement so no worries! :)

michalmarcinkowski:
*OrderItem <--> OrderItemUnit* communication and calculations are handled correctly right now on the model level.

@peteward @pjedrzejewski @Zales0123 please review my commit :)

peteward:
:+1: 

Not the subject of the PR, but also related - what about the fetching adjustments recursively as you have started to do totals? So you can get all TAX adjustments directly from the Order object no matter what depth they are at? Useful for display.

michalmarcinkowski:
@peteward already implemented for adjustment total. Not sure about the method name, but we didn't figure out anything better, open for suggestions :) 

See corresponding code:
https://github.com/Sylius/Sylius/pull/3786/files#diff-920978c7c9397feba23dc56643bedc80R341
https://github.com/Sylius/Sylius/pull/3786/files#diff-80558d029a5d54880de294667e488179R426

peteward:
@michalmarcinkowski yes, I mean the actual adjustment objects. Because you might want to get the unique descriptions to list what they were (e.g. for Tax you might want to display "US Sales Tax - 12%").

michalmarcinkowski:
@peteward ok I see what you do mean now :)

I understand your use case, good point but I see a few solutions for that so we should probably discuss it in a separate issue/PR. We could add a few more recursive methods but let's follow YAGNI and implement them when we decide that they make sense and they are really useful.

peteward:
sounds good. 

So what's the status of this PR now?

michalmarcinkowski:
@Zales0123 will work on doing the same for *Order* <-> *OrderItem* communication and making the build green with changes from https://github.com/Sylius/Sylius/pull/3795.

michalmarcinkowski:
This can be closed as the changes are already applied in https://github.com/Sylius/Sylius/pull/3795.

PR #3788, opened 2015-12-29
Fixes #3787.

### LINKED ISSUE #3787: Cannot load provinces for the country
After upgrade to the newest master version loading of provinces for specific country doesn't work anymore.

When calling this URL /province/choice-form?countryId=175, I've got this response.
```
Controller "sylius.controller.province:choiceFormAction" for URI "/province/choice-form" is not callable.
```

This is our configuration of route, which is the same as you have in the sylius.

```
sylius_province_choice_form:
    path: /province/choice-form
    defaults:
        _controller: sylius.controller.province:choiceFormAction
        _sylius:
            template: SyliusAddressingBundle:Province:_provinceChoiceForm.html.twig
```

I have noticed, templates have been removed recently, but even if I add this specific template, there is same error.

And as it seems from the source code, templates are still in the use, even though they have been deleted.

```
$form = $this->createProvinceChoiceForm($country);

        $content = $this->renderView($this->getConfiguration()->getTemplate('_provinceChoiceForm.html'), array(
            'form' => $form->createView(),
        ));

        return new JsonResponse(array(
            'content' => $content,
        ));
```

ufik:


pjedrzejewski:
:+1: @Sylius/core-team good to go when green.

pjedrzejewski:
Thanks @ufik! :+1:

PR #3792, opened 2015-12-29
[RFC] [API] Rename response files

lchrusciel:
Just a thought. We can follow action naming convention with expected response naming. WDYT?


pjedrzejewski:
Looks good to me. Let's hear more opinions. :)

michalmarcinkowski:
Good idea, thanks Łukasz! :+1:

PR #3807, opened 2016-01-04
[Attribute] Update *AttributeByName => *AttributeByCode inline with updated model

peteward:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

`name` is now translated so this no longer makes sense

### DIFF
--- a/src/Sylius/Component/Product/Model/Product.php
+++ b/src/Sylius/Component/Product/Model/Product.php
@@ -280,10 +280,10 @@ class Product extends AbstractTranslatable implements ProductInterface
     /**
      * {@inheritdoc}
      */
-    public function hasAttributeByName($attributeName)
+    public function hasAttributeByCode($attributeCode)
     {
         foreach ($this->attributes as $attribute) {
-            if ($attribute->getName() === $attributeName) {
+            if ($attribute->getAttribute()->getCode() === $attributeCode) {
                 return true;
             }
         }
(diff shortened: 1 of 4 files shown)


pjedrzejewski:
Definitely, thanks Peter! :+1:

src/Sylius/Component/Product/Model/Product.php
```diff
@@ -280,10 +280,10 @@ public function hasAttribute(BaseAttributeValueInterface $attribute)
     /**
      * {@inheritdoc}
      */
-    public function hasAttributeByName($attributeName)
+    public function hasAttributeByCode($attributeCode)
     {
         foreach ($this->attributes as $attribute) {
-            if ($attribute->getName() === $attributeName) {
+            if ($attribute->getAttribute()->getCode() === $attributeCode) {
```
michalmarcinkowski: This `$attribute` should be renamed to `$attributeValue` because it looks weird :)

Btw.
Not sure why `$this->attributes` isn't `$this->attributesValues`...

peteward: definitely agreed, but this isn't the subject of this PR. one for the future.

pjedrzejewski: :+1:

pjedrzejewski:
Thanks Peter!

PR #3812, opened 2016-01-04
Mark some services as lazy to prevent db connection

pjedrzejewski:
This is so bad and annoying. It is obviously a workaround for now, but I marked some services as lazy so that it does not connect to the db during cache warmup.


michalmarcinkowski:
We should definitely fix this services properly later on.

Issue #3820, opened 2016-01-06
CartEvent isFresh() is deprecated


flajos:
Sylius\Component\Cart\Event::isFresh() method is deprecated.
In the past it was used to recalculate the cart totals: 

    if ($event->isFresh()) {		
        $this->refreshCart($cart);		
    }

The checking for the isFresh() values were removed in commit 93218a3206fea527cba1846f02485796cf572455. The refreshing of the cart is solved now with the RefreshCartListener which listens to SyliusCartEvents::CART_CHANGE event.

The Sylius\Component\Cart\Event::isFresh() method is not checked currently anywhere. It is only set in the CartController and CartItemController as a remnant of the old code. As it is not used it can be removed.

michalmarcinkowski:
Fixed in #3821

PR #3821, opened 2016-01-06
[Cart] remove deprecated isFresh() method from CartEvent

flajos:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | #3820
| License       | MIT

In older versions of Sylius the cart totals were recalculated (aka the cart was "refreshed") via setting the CartEvent to fresh and checking if it is fresh on the right places. If the event was fresh, the cart was recalculated. Now that is solved via triggering the SyliusCartEvents::CART_CHANGE event. The isFresh() method is removed (it was only set, resulting in a more convoluted code, but never checked for after the switch to event listeners).

### DIFF
--- a/src/Sylius/Bundle/CartBundle/Controller/CartController.php
+++ b/src/Sylius/Bundle/CartBundle/Controller/CartController.php
@@ -69,7 +69,6 @@ class CartController extends Controller
 
         if ($form->handleRequest($request)->isValid()) {
             $event = new CartEvent($cart);
-            $event->isFresh(true);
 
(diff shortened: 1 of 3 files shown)


michalmarcinkowski:
@flajos the build is failing.

FYI the cart recalculation process will be totally reworked in the comming weeks :)

flajos:
Yep, thx, it seems OK now.

I'm eager to see the rework :)

pjedrzejewski:
Thanks Lajos!

PR #3827, opened 2016-01-07
[Travis] Improved error reporting on Travis

pamil:
Logs and screenshots on `test` and `test_cached` environment should be easily viewed now on Travis.

PR #3833, opened 2016-01-08
[Travis] [Core] Improved showing exceptions in test environment

pamil:
Reverted #3827, as we don't use default Twig's exception handler but the FOSRestBundle one.

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -7,9 +7,6 @@ imports:
     - { resource: migrations.yml }    
     - { resource: @SyliusCoreBundle/Resources/config/app/main.yml }
 
-parameters:
-    sylius.twig.show_exceptions: %kernel.debug%
-
 framework:
     translator:      { fallback: %sylius.locale% }
     secret:          %sylius.secret%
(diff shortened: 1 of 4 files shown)


src/Sylius/Bundle/CoreBundle/Controller/ExceptionController.php
```diff
@@ -0,0 +1,34 @@
...
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+class ExceptionController extends BaseExceptionController
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function showAction(Request $request, $exception, DebugLoggerInterface $logger = null)
+    {
+        if ($this->container->hasParameter('sylius.show_exception') && $this->container->getParameter('sylius.show_exception')) {
```
stloyd: Instead of calling this always, maybe add listener in test env that will be run only there, not also in production etc...

pamil: :+1:

pjedrzejewski:
Thanks Kamil!

PR #3841, opened 2016-01-11
[Promotion rule] added items quantity

starspire:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Sometimes you want to give promotions when items in cart exceeds some quantity, doesn't matter if they are different or not (Item count rule counts only different items). This PR does actually this - add rule which calculate total items quantity.


starspire:
@pjedrzejewski up

michalmarcinkowski:
That was implemented in #4257.

PR #3795, opened 2015-12-30
[Order] Implementation of OrderItemQuantityModifier

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3524
| License       | MIT
| Doc PR        | https://github.com/Sylius/Sylius-Docs/pull/400

This PR is based on https://github.com/Sylius/Sylius/pull/3771 and correspond with @michalmarcinkowski work in https://github.com/Sylius/Sylius/pull/3786.

In this PR there is ``OrderItemQuantityModifier`` implemented, that handles item units management. The idea is, to change ``OrderItem`` quantity only by adding/removing units.

**Possible improvements/modifications**
- it will be better to remove *backordered* units first, then *sold* ones

**UPDATE**

I've applied commits from https://github.com/Sylius/Sylius/pull/3808, as they're necessary to make behat scenarios green.

**UPDATE vol.2**

Docs PR mentioned.

PR #3842, opened 2016-01-11
[Order] Order-related cleanups

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

In this PR, based on https://github.com/Sylius/Sylius/pull/3795, there are some cleanups made, including remove unnecessary listeners, methods, cs fixes and so on.

I'm sure any review will be much easier after merging https://github.com/Sylius/Sylius/pull/3795 or at least https://github.com/Sylius/Sylius/pull/3771 :smile:


src/Sylius/Bundle/OrderBundle/DependencyInjection/Configuration.php
```diff
@@ -144,6 +147,20 @@ private function addResourcesSection(ArrayNodeDefinition $node)
...
+                        ->arrayNode('order_item_unit')
+                            ->addDefaultsIfNotSet()
+                            ->children()
+                                ->arrayNode('classes')
+                                    ->addDefaultsIfNotSet()
+                                    ->children()
+                                        ->scalarNode('model')->defaultValue(OrderItemUnit::class)->cannotBeEmpty()->end()
+                                        ->scalarNode('interface')->defaultValue(OrderItemUnitInterface::class)->cannotBeEmpty()->end()
+                                        ->scalarNode('factory')->defaultValue(OrderItemUnitFactory::class)->end()
+                                        ->scalarNode('repository')->cannotBeEmpty()->end()
```
michalmarcinkowski: Swap `->scalarNode('repository')` with `->scalarNode('factory')` to stay consistent with the rest of configs.

src/Sylius/Component/Core/Model/OrderItemInterface.php
```diff
@@ -13,6 +13,7 @@
 
 use Doctrine\Common\Collections\Collection;
 use Sylius\Component\Cart\Model\CartItemInterface;
+use Sylius\Component\Inventory\Model\InventoryUnitInterface;
```
michalmarcinkowski: Redundant

src/Sylius/Component/Order/Model/Order.php
```diff
@@ -206,16 +206,12 @@ public function addItem(OrderItemInterface $item)
...
-            if ($item->equals($existingItem)) {
-                $existingItem->merge($item, false);
-
-                return;
-            }
-        }
-
         $item->setOrder($this);
+        $itemTotal = $item->getTotal();
+        $this->itemsTotal = $this->itemsTotal + $itemTotal;
```
michalmarcinkowski: Why not simply `$this->itemsTotal += $item->getTotal();`

src/Sylius/Component/Order/Model/Order.php
```diff
@@ -574,4 +448,126 @@ public function removeIdentity(IdentityInterface $identity)
...
+            $this->removeAdjustment($adjustment);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function clearAdjustments()
+    {
+        $this->adjustments->clear();
```
michalmarcinkowski: We should recalculate adjustments after `clear`

src/Sylius/Component/Order/Model/Order.php
```diff
@@ -206,16 +206,12 @@ public function addItem(OrderItemInterface $item)
...
-                return;
-            }
-        }
-
         $item->setOrder($this);
+        $itemTotal = $item->getTotal();
+        $this->itemsTotal = $this->itemsTotal + $itemTotal;
         $this->items->add($item);
+
+        $this->calculateTotal();
```
michalmarcinkowski: Let's be consistent and rename this to `recalculateTotal()`

src/Sylius/Component/Order/Model/Order.php
```diff
@@ -248,125 +247,14 @@ public function getItemsTotal()
...
-        if (!is_int($itemsTotal)) {
-            throw new \InvalidArgumentException('Items total must be an integer.');
-        }
-        $this->itemsTotal = $itemsTotal;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function calculateItemsTotal()
```
michalmarcinkowski: `recalculate...`

src/Sylius/Component/Order/Model/OrderItem.php
```diff
@@ -131,103 +124,133 @@ public function setUnitPrice($unitPrice)
...
+        return $this->total;
+    }
 
-        return $this->adjustments->filter(function (AdjustmentInterface $adjustment) use ($type) {
-            return $type === $adjustment->getType();
-        });
+    /**
+     *  Recalculates total after units total or adjustments total change.
+     */
+    public function recalculateTotal()
```
michalmarcinkowski: Should be `protected`

src/Sylius/Component/Order/Model/OrderItem.php
```diff
@@ -131,103 +124,133 @@ public function setUnitPrice($unitPrice)
...
 
-        foreach ($this->adjustments as $adjustment) {
-            if (!$adjustment->isNeutral()) {
-                $this->adjustmentsTotal += $adjustment->getAmount();
+            $this->quantity--;
+            $this->unitsTotal -= $unit->getTotal();
+            $this->recalculateTotal();
+
+            if (null !== $this->order) {
+                $this->order->calculateItemsTotal();
```
michalmarcinkowski: I would say this should be called each time `$this->recalculateTotal()` method is called, so we can move it there.

Zales0123: :+1: I came up with the same idea while working on next changes, just didn't thought about implementing this here :smile:

src/Sylius/Component/Order/Model/Order.php
Zales0123: Done :+1:

michalmarcinkowski:
General remarks:
**Order** model:
* `getTotalItems` should be removed
* `countItems` should be renamed to `getItemsNumber`

src/Sylius/Component/Order/Model/Order.php
```diff
@@ -558,12 +530,14 @@ public function removeAdjustments($type)
     public function clearAdjustments()
     {
         $this->adjustments->clear();
+
```
michalmarcinkowski: IMO this blank line is redundant

pjedrzejewski:
@michalmarcinkowski I think we should discuss the naming first. We usually need:

1. Total quantity of all items.
2. Number of unique items in the cart.

I'd like to propose the following naming:

1. ``getTotalQuantity()``
2. ``getNumberOfItems()``

WDYT guys?

michalmarcinkowski:
1. `getTotalQuantity()` is already there
2. `getNumberOfItems()` better than `getItemsNumber` :+1: 

Totally agree :)

app/migrations/Version20151231002659.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_order_item ADD units_total INT NOT NULL');
```
stloyd: Such migrations are often failing on database that contain data, this would fail that this field cannot be `null` but it will be, as every new field in existing data is set to `null`.

First this should contain `null` as default value, later there should be `update` sql that sets 0 (or other value) and after that there should be set `not null` on this new field.

src/Sylius/Bundle/CartBundle/spec/Form/Type/CartItemTypeSpec.php
```diff
@@ -37,12 +38,18 @@ function it_is_a_form_type()
         $this->shouldImplement(FormTypeInterface::class);
     }
 
-    function it_builds_form_with_quantity_field(FormBuilder $builder)
+    function it_builds_form_with_quantity_field($orderItemQuantityDataMapper, FormBuilder $builder)
     {
         $builder
             ->add('quantity', 'integer', Argument::any())
+            ->willReturn($builder)
             ->shouldBeCalled()
```
stloyd: `shouldBeCalled()` removed, here & below as you already call `willReturn()`.

src/Sylius/Bundle/FixturesBundle/DataFixtures/ORM/LoadOrdersData.php
```diff
@@ -53,7 +54,8 @@ public function load(ObjectManager $manager)
                 $item = $orderItemFactory->createNew();
                 $item->setVariant($variant);
                 $item->setUnitPrice($variant->getPrice());
-                $item->setQuantity(rand(1, 5));
+
+                $orderItemQuantityModifier->modify($item, rand(1, 5));
```
stloyd: Use `mt_rand()`.

src/Sylius/Bundle/OrderBundle/DependencyInjection/SyliusOrderExtension.php
```diff
@@ -41,5 +39,10 @@ public function load(array $config, ContainerBuilder $container)
         foreach ($configFiles as $configFile) {
             $loader->load($configFile);
         }
+
+        $orderItemQuantityDataMapper = $container->getDefinition('sylius.data_mapper.order_item_quantity');
+        $orderItemType = $container->getDefinition('sylius.form.type.order_item');
+
+        $orderItemType->addArgument($orderItemQuantityDataMapper);
```
stloyd: This should be `new Reference()` not direct definition setting.

src/Sylius/Component/Inventory/Model/InventoryUnitInterface.php
```diff
@@ -33,16 +33,6 @@
     public function getStockable();
 
     /**
-     * @param StockableInterface $stockable
-     */
-    public function setStockable(StockableInterface $stockable);
```
stloyd: Why removing this?

pjedrzejewski: Cause the Component never uses it most likely will never use, we don't care how stockable is defined for the InventoryUnit, we just need it to be there.

app/migrations/Version20151231002659.php
michalmarcinkowski: Adding `DEFAULT 0` at the end will not work?

src/Sylius/Bundle/OrderBundle/DependencyInjection/SyliusOrderExtension.php
pjedrzejewski: :+1:

src/Sylius/Bundle/CartBundle/spec/Form/Type/CartItemTypeSpec.php
pjedrzejewski: I think we discussed this many times, we agree that these are redundant except for form/query builders. Otherwise specs for these are pointless. Agree?

pamil: Anyway, isn't it more natural to call `willReturn()` / `willThrow()` at the end of method chain?

app/migrations/Version20151231002659.php
pamil: @michalmarcinkowski I guess we don't want to have default values on columns that shouldn't be  default. On the other hand, we may use `DEFAULT` in a query that creates the column, and then drop the default value in the next one.

src/Sylius/Bundle/CartBundle/spec/Form/Type/CartItemTypeSpec.php
michalmarcinkowski: :+1: I prefer that also :)

peteward:
@ quantity naming, fwiw, 

The important distinction here is in the word describing what you want the quantity of - `Unit` and `Item`.

I think that the surrounding adjectives can just cause the exact confusion that you're trying to avoid. So I would go for something like:

```
getItemQuantity()
getUnitQuantity()
```

or if you like number of then:

```
getNumberOfItems()
getNumberOfUnits()
```

Put them next to each other in the Interface and clearly explain the distinction (although I hope it's actually fairly obvious!).

src/Sylius/Bundle/CartBundle/spec/Form/Type/CartItemTypeSpec.php
Zales0123: Agree :+1:

Zales0123:
@peteward We've been talking a while about naming convention on today's standup and what we agreed on:
- ``getTotalQuantity()`` is most natural name of function - somebody who is familiar with **Sylius** will probably have no problem with catching what ``units`` in ``getNumberOfUnits()`` are, but *de facto* ``number of units`` is not as clear as ``total quantity`` for anybody else
- ``getNumberOfItems()`` is pretty obvious and clear, so we will go this way

Are you convinced and good with ``getTotalQuantity()``? Nevertheless, we're all conscious how hard naming is, as it's essence of programming :smile:

peteward:
sure thing, it was just my opinion! :)

pjedrzejewski:
Thank you Mateusz! :+1:

PR #3844, opened 2016-01-11
Fix deps in most of bundles

pjedrzejewski:


src/Sylius/Bundle/ResourceBundle/composer.json
```diff
@@ -37,9 +38,11 @@
     "require-dev": {
         "phpspec/phpspec": "^2.4",
         "akeneo/phpspec-skip-example-extension": "~1.2",
-        "doctrine/mongodb-odm": "1.0.*@dev",
+        "doctrine/mongodb-odm": "^1.0",
```
pamil: I guess it will fail because it requires `ext-mongo` which we don't have.

pjedrzejewski: Yes and we don't have it on 7.0 yet I think, so we will need to live with a failing build anyway because excluding mongodb examples for 7.0 only is work that requires more time than I have right now. :P We can look at it shorlty after release I guess.

pamil: It will affect 5.5 and 5.6 unless we add `echo "extension = mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini` before installing dependencies.

bendavies: you can use mongofill https://github.com/mongofill/mongofill

`- if [[ $TRAVIS_PHP_VERSION == '7.0' ]]; then composer require mongofill/mongofill=dev-master --no-update;`

michalmarcinkowski:
:+1:

bendavies:
resource bundle needs `symfony/validator`

PR #3847, opened 2016-01-11
Fix dependencies of individual bundles

pjedrzejewski:


src/Sylius/Bundle/ResourceBundle/composer.json
```diff
@@ -38,6 +39,7 @@
     "require-dev": {
         "phpspec/phpspec": "^2.4",
         "akeneo/phpspec-skip-example-extension": "~1.2",
+        "mongofill/mongofill": "dev-master",
```
bendavies: i know i recommended this in the other issue, but what is it for? the tests pass without this or the mongo extention

pjedrzejewski: Good point, I forgot we are not using any of "native" Mongo extension classes so we can safely remove it. Thanks for the recommendation, I did not know about this replacement. :+1:

michalmarcinkowski:
:+1:

PR #3852, opened 2016-01-12
[RFC] [POC] [Resource] Resource configuration builder

pamil:
Resource configuration builder allows us to define bundle's resources configuration without messing with endless `->end()`, tabs, method chains having hundreds of lines and no autocompletion after first `->end()` call :)

It also ensures that the most of future changes can be done seamlessly and without any trouble by modifying only a single class.

This PR is still in progress, I'm open for any thoughts about the architecture and the whole idea. There are two example configurations rewritten to use the resource configuration builder: 
  - [AttributeBundle](https://github.com/Sylius/Sylius/commit/da7e3e52415527aa4a854940ad031ae8c0f3a8b4) (which is a bit unusual configuration, because `sylius_attribute.resources` holds an array prototype)
  - [LocaleBundle](https://github.com/Sylius/Sylius/commit/fedc704065f0c355e8da5f4d0fd67fba0451ad40) (simple configuration made simplier :))

### DIFF
--- a/src/Sylius/Bundle/AttributeBundle/DependencyInjection/Configuration.php
+++ b/src/Sylius/Bundle/AttributeBundle/DependencyInjection/Configuration.php
@@ -15,7 +15,10 @@ use Sylius\Bundle\AttributeBundle\Controller\AttributeController;
 use Sylius\Bundle\AttributeBundle\Form\Type\AttributeTranslationType;
 use Sylius\Bundle\AttributeBundle\Form\Type\AttributeType;
 use Sylius\Bundle\AttributeBundle\Form\Type\AttributeValueType;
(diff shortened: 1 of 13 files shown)


pamil:
We can also provide documentation and examples for configuration using `->info()` and `->example()` on the nodes, just like in [Symfony's SecurityBundle configuration](https://github.com/symfony/symfony/blob/582f4753a343f230fbe18b4e9a0747d48351ddfb/src/Symfony/Bundle/SecurityBundle/DependencyInjection/MainConfiguration.php#L388).

bendavies:
very nice

pamil:
Closing as there's no further work incoming in the near future.

PR #3866, opened 2016-01-13
[Promotions] Orphan removal on actions and rules

steffenbrem:
When updating promotions through the API, it leaves orphans in the database. Right now it only disconnects the relationship, but when you leave out an action, you want it to be gone completely. With this PR they are correctly removed when they are disconnected from their promotion.


pjedrzejewski:
Makes sense, thank you Steffen! :+1:

stloyd:
Migration is required IIRC.

PR #3867, opened 2016-01-14
Fixed issue if variant is not selected

antonioperic:


pjedrzejewski:
Thank you Antonio! :+1:

Issue #2842, opened 2015-06-11
[RFC] Multiple Shipping Addresses


patrick-mcdougle:
Anyone have a clear idea on what it would take to ship order items to different addresses? For example if I am buying 5 shirts for 5 of my friends, I would need to place 5 orders instead of making one order with 5 shipping addresses.

There's a business case for this too, since there's usually a fixed cost for payment processing, so minimizing (by allowing 5 different ship-to addresses) total number of orders results in decreased costs and increased profits.

Thoughts?

pjedrzejewski:
This is currently not implemented and won't be possible in 1.0. But definitely interesting use-case. I could not find any related issues, so I will mark this as a feature request.

patrick-mcdougle:
I think the path forward for this would be adding address information to the shipment entity (and maybe the cart item)?

pjedrzejewski:
@patrick-mcdougle Yes, this makes sense!

vanclei:
Hi @patrick-mcdougle , did you find a way to implement this?

Cheers

patrick-mcdougle:
I did, but it starts to get complicated when dealing with taxes and coupons. We essentially had to highly modify some of the tax and promotion processing to make it work. In short, start storing the address as a value object on the shipments themselves as opposed to storing them on the order. Sorry, I can't really share any code from that, because it's proprietary. Sorry.

vanclei:
Hi Patrick, thank you for your reply.  I'll have a look, and do not worry about proprietary.

cheers

patrick-mcdougle:
I think this is still a compelling feature that should eventually get on a roadmap.

psren:
stale bot sucks...

tudorradubarbu:
Anyone got more news on this one?

PR #3868, opened 2016-01-14
[Order] Order taxes

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

This PR is based on previous order logic changes - https://github.com/Sylius/Sylius/pull/3842 and previous PR's.

To handle taxes in a new, better way, some new services and *Order* component's methods are introduced:
- reworked and refactored ``OrderTaxesApplicator`` apply taxes on each ``OrderItemUnit``, after proper distribution
- ``getAdjustmentsRecursively($type)`` methods on ``Order`` and ``OrderItem`` returns target and its children's flat adjustments' array
- ``AdjustmentsAggregator`` is used to aggregate taxes by ``description``
- custom twig extension with ``sylius_aggregate_adjustments`` function is used to display aggregated adjustments


peteward:
Why calculate tax at the `OrderItem` level and then require a distributor to split the value? Why not just calculate at the `OrderItemUnit` level? It would be a lot cleaner. Also you rule out the possibility (or at least make it much harder) for people to extend with scenarios where variants have different tax rates (it's possible).

I think there is definitely a place for a `Distributor`, but it makes more sense in promotions where you want to apply a value that is chosen by an administrator at a higher level (order/item) down to lower levels. For tax why not just run calculations at the `OrderItemUnit` level?

We need to split `apply()` out into `revert()` and `apply()` separately which listen at respective priorities to ensure calculations are correct.

The current process of applying adjustments is:
* Apply promotions (remove promotion adjustments, recalculate them)
* Apply taxes (remove tax adjustments, recalculate them)

If this process is rerun, you will have tax adjustments on your items when you calculate promotions for them, which will then be removed and reapplied on something completely wrong.

I think it needs to be:
* Revert taxes
* Revert promotions
* Calculate promotions
* Calculate taxes

There needs to be some very comprehensive behat scenarios with non-neutral tax and promotion adjustments to show this.

Zales0123:
@peteward I was wondering about your questions about calculating taxes... On first sight they look reasonable, on second also :smile: However, there is one big problem with calculating taxes on ``OrderItemUnit`` model, when total amount cannot be equally split between each unit.
 Let's say we have 4 units, 250 cents each. There are two possibilities of calculating 23% VAT for them:
1. Calculate 23% of 1000 cents (total) and split it using distributor on these items - 57 + 57 + 58 + 58 = 230
2. Calculate 23% of each units - 23% x 250 = 57,5

Sadly, on second approach we must somehow have knowledge about other units, which is inconvenient as it's calculated per **one** ``OrderItemUnit``. First way is not so bad, The only thing we do more contains only integers/floats operations, which is negligible for performance and which *de facto* should be done also for second method of calculating. So in my opinion it's not so bad to make some more arithmetic operations, as they make as certain about proper taxes on each unit ;)

Unfortunately, variants are not *taxable* right now (only ``Product`` implements ``TaxableInterface``), but we have fixing this issue in schedule. However, different variants are still different items in cart, so it won't be a problem to handle this (at least I think so, @pjedrzejewski @michalmarcinkowski am I right or am I wrong?)

Last but not the least, I'm :+1: with ``apply()`` - ``revert()`` approach, we should definitely refactored code this way ;)

peteward:
Hi @Zales0123, yes that Variant comment was just wrong... of course it's Variant at the OrderItem level...

As for tax calculation, this is a very good point you've raised! And the question is - which way is truly correct?

Scenario:
* I walk into our store upstairs and take 3 identical white t-shirts priced at  £25 each to the checkout till
* I'm quoted a final price of £75.01 because tax calculated at the item level means a rounding up of 1 pence across 3 items.
* What do you think I say? :)

If items are priced and quoted individually then I think the calculation should be done at that level. You will always get rounding errors when it comes to money, but I disagree with trying to round up tax in this way in an attempt to be *more accurate* because it can be confusing and inconsistent.

michalmarcinkowski:
Hi @peteward @Zales0123,

Regarding tax calculation:

Each `OrderItemUnit` have the same `Taxable` entity as it's `OrderItem` (right now it's `Product`, but we're going to make it a `Variant`), so there is no need to calculate tax on each `OrderItemUnit` separately. We should calculate the tax on `OrderItem` level and simply distribute this across `OrderItemUnit`s. If we will have a proper calculations and rounding there is no way the prices will not match, but calculating it separately can introduce some rounding issues (as in @Zales0123 example) :) @peteward how did you get this £75.01?

Regarding applying taxes and promotions:

In order to have a full understanding about the correct logic with promotions, we need to rethink and design the promotions system first. However thanks to @peteward I actually realized that we will have a logical issue when tax is included in price. Currently `OrderItem` unit price = variant price, so in that case it will be amount with included tax, so even after clearing all tax adjustments we will still calculate the promotion with tax included for this variant. So in order to be consistent we should always set real unit price (when tax included in price then unitPrice = variantPrice - tax) on `OrderItem` and then just take it into account when displaying.

There is a lot of things to discuss and think of when implementing promotions, but let's leave it until we start working on that :)

peteward:
Hi @michalmarcinkowski,

Using @Zales0123 example of 23% tax split across 4 items, you will end up with 4 Units with different tax values. Surely this is fundamentally wrong?

I know that in some (maybe even most) applications you will 'hide' this by rolling up/aggregating these on cart or checkout screens, but I can easily imagine some customers or applications where they do not want to visually aggregate in this way, possibly on the backend, or on invoices, email receipts, etc. It would be strange and wrong for these to display different tax values for the same items. When you come to refunds it will also be obvious as the tax amounts now must be split visually to show these differences.

Also, if a Customer orders 4 units individually they will pay a different amount to them buying 4 together - again surely this is wrong?  

You may argue that it's more accurate to calculate 23% of 1000 (230) than 23% of 250 (57.5) which needs to be rounded up or down. But what about there the item total doesn't divide correctly either? What if the item is 249? Then it's 996 and then your tax total is 229.08 so this isn't a clean integer and your tax isn't technically 'accurate' either, even at the item level.

My example about a real store was just trying to make the point that you wouldn't go into a bricks & mortar store and pay a different amount in tax for an item based on the quantity.


As for promotions, I disagree with your point about subtracting tax for calculating promotions. If a price is tax inclusive then usually the tax component is calculated at the end as a result of all other finances (same as tax exclusive). You inform the customer at the end of the process how much of the price that they have paid is calculated as tax.

e.g. if it's a 10% discount off a £50 jumper you would give the customer £5 discount, and tax would be calculated on the remaining £45 that the customer would pay. You wouldn't take 23% off the £50 then take 10% off, THEN add the tax back on.

michalmarcinkowski:
What is the correct tax for the following use case (I assumed tax rate equal to 23%):
```
Given there is product „T-Shirt” priced at £10.20
When I add 10 „T-Shirt” products to my cart
Then the total tax in my cart should be ?
```

1. Rounding tax on OrderItem level
    £102.00 * 0.23 = £23.46
2. Rounding tax on OrderItemUnit level
    £10.20 * 0.23 = 2.346 per unit = ~2.35 * 10 units = £23.50

Same example with rounding down:
```
Given there is product „T-Shirt” priced at £10.10
When I add 10 „T-Shirt” products to my cart
Then the total tax in my cart should be ?
```

1. £101.00 * 0.23 = £23.23
2. £10.10 * 0.23 = 2.323 per unit = ~2.32 * 10 units = £23.20

The  tax values diverges and will diverge more with units number increase.
```
Given there is product „T-Shirt” priced at 10.20
When I add 1000 „T-Shirt” products to my cart
Then the total tax in my cart should be ?
```

1. £10200.00 * 0.23 = £2346.00
2. £10.20 * 0.23 = 2.346 per unit = ~2.35 * 1000 units = £2350.00

I don't want to argue I just don’t know what are the accounting regulations/rules for that use cases. Calculating tax on the OrderItem level results in rounding the number only once, in comparison to X times when calculating it on the OrderItemUnit level.

I agree that having different tax amount on the same units is strange, but the fundamental question here is what are the correct tax values for these and similar use cases from accounting point of view. I don’t have any objections in calculating tax on OrderItemUnit level (maybe despite small additional overhead), but calculating it on OrderItem level seems better for me because we have a better accuracy when we round the number only once.

The refunds are a separate question, after we determine this issue.

@peteward regarding promotions I think you misunderstood me :) The change proposed by me is going to make the calculations more flexible because we can decide if we do want to calculate something based on amount tax inclusive or exclusive. Right now we don't have this flexibility and even worse, we can have few items in a cart when for some of them the tax will be included in `unitPrice` and for others the `unitPrice` will be a price without tax. Calculating something based on these values can lead to errors. IMO the `unitPrice` should be consistent and should always be equal to a unit price of an item without tax and the tax should always be applied by adjustment. This way we have a full control over the calculations order, whether we want to calculate taxes first and then apply promotions or do it the other way around. This change will not introduce any limitation.

Actually it doesn't matter when the promotion is percentage:
```
Given there is product „T-Shirt” priced at £50.00 with tax included in price
And there is promotion "10% off"
And tax rate is 20% for all products
When I add „T-Shirt” product to my cart
Then my grand total should be £45.00
And the total tax in my cart should be £7.50
```

1. Discount apply on item with tax
    £50 - (£50 * 0.1) = £45 (grand total) - apply discount
    £45 - (£45 / 1.2) = £37.50 (item total) - substract tax
    £45 - £37.5 = £7.50 (tax)

2. Discount apply on item without tax
    £50 - (£50 / 1.2) = £41.67 (item total without discount) - calculate value without tax
    £41.67 - (£41.67 * 0.1) = £37.50 (item total) - apply discount
    £37.50 * 0.2 = £7.50 (tax)
    £37.50 + (£7.50) = £45 (grand total) - add tax

The values are the same for both cases. The order makes difference when the promotion is fixed, but let's leave it to a separate discussion (again a long story).

peteward:
Hi @michalmarcinkowski,

Yes agreed, I think we both see the point here. It depends how you define accuracy. I can see from your perspective that it feels like the total amount of tax paid is closer to a 'true' 23% of tax on the Items, but from my experience in retail this isn't how it works.

I'm going to try and get some qualified knowledge on my end to back this up though. I'll try to get an opinion from our tax consultant then I'll get back to you. Unfortunately they can be a bit slow...


As for promotions I see your point. Just to make you aware that we support both prices inclusive of tax and exclusive of tax (US). So whatever scheme you decide on, please consider that international retailers could often have prices with both. It's been a while since I checked default Sylius but I think it only supports a single price which can be adjusted by exchange rates for different currencies. We have explicit prices for different channels via PriceLists and their prices can be with or without tax.

Agreed let's not bloat out this PR any further talking about promotions, I can annoy you further when you start on it ;)

peteward:
A quick google and I can find a couple of points of view:

Ryan Szarma of Commerce Guys (https://www.drupal.org/node/1137840):
```
That said, I think you're wrong. VAT is rounded at the unit price level, so you can't have a unit price that gets multiplied by an even number end up an odd number. Your "correct subtotal" of 44,63 € just doesn't make sense to me. It looks to me like the correct calculation is as rounded, which this issue was closed fixed as supporting.
```

There seems to be much more emphasis on the rounding method used rather than the level at which you calculate. I would just assume rounding to the nearest integer but it appears that some people/rules required specific rounding either UP, DOWN or NEAREST.

On the other hand this is an interesting article that shows an issue in the UK with a company repeatedly changing the way they pay and calculate tax, it's quite old but probably still relevant: http://old.tax.org.uk/attach.pl/5450/5752/029-030_TA_0407.pdf

This suggests that you could probably do either, at least in the UK, if you communicate the issue with HMRC.

I would still find it very strange to have units with different tax values, but it might be something that should be configurable...

michalmarcinkowski:
@peteward thanks for feedback! We had an internal discussion and we came to the same conclusion that the strategies probably differs from one country to other, from one company to other etc. We would like to support all cases and will build the architecture that allows for different strategies, so we can decide which strategy should be used.

Regarding prices, no worries we will support both inclusive of tax and exclusive of tax.

src/Sylius/Bundle/CoreBundle/Distributor/TaxesDistributor.php
```diff
@@ -0,0 +1,47 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\CoreBundle\Distributor;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class TaxesDistributor implements TaxesDistributorInterface
```
michalmarcinkowski: Is it really only taxes distributor? It is generic and can be used to distribute any number.

peteward: :+1:

src/Sylius/Bundle/CoreBundle/Distributor/TaxesDistributor.php
```diff
@@ -0,0 +1,47 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function distribute($taxItems, $taxAmount)
```
michalmarcinkowski: `($numberOfTargets, $amount)`?

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
```diff
@@ -0,0 +1,175 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function apply(OrderInterface $order)
+    {
+        if ($order->getItems()->isEmpty()) {
+            return;
+        }
+
+        $this->clearAdjustments($order);
```
michalmarcinkowski: This should be called on empty cart also, so we won't have a tax adjustments applied to an empty order.

I would rename the method to `clearTaxes`.

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
```diff
@@ -0,0 +1,175 @@
...
+        $this->taxRateResolver = $taxRateResolver;
+        $this->zoneMatcher = $zoneMatcher;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function apply(OrderInterface $order)
+    {
+        if ($order->getItems()->isEmpty()) {
```
michalmarcinkowski: `$order->isEmpty()`

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
```diff
@@ -0,0 +1,175 @@
...
+
+            if (null === $rate) {
+                continue;
+            }
+
+            $percentageAmount = $rate->getAmountAsPercentage();
+            $totalTaxAmount = $this->calculator->calculate($item->getTotal(), $rate);
+            $label = sprintf('%s (%s%%)', $rate->getName(), (float) $percentageAmount);
+
+            $splitTaxes = $this->orderUnitTaxesDistributor->distribute($item->getUnits()->count(), $totalTaxAmount);
```
michalmarcinkowski: `$item->getQuantity()`

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
```diff
@@ -0,0 +1,175 @@
...
+    /**
+     * @param CalculatorInterface $calculator
+     * @param DefaultTaxZoneProviderInterface $defaultTaxZoneProvider
+     * @param FactoryInterface $adjustmentFactory
+     * @param TaxesDistributorInterface $orderUnitTaxesDistributor
+     * @param TaxRateResolverInterface $taxRateResolver
+     * @param ZoneMatcherInterface $zoneMatcher
+     */
+    public function __construct(
+        CalculatorInterface $calculator,
```
michalmarcinkowski: I would call it `$taxCalculator`.

src/Sylius/Bundle/CartBundle/spec/Form/Type/CartItemTypeSpec.php
```diff
@@ -44,12 +44,14 @@ function it_builds_form_with_quantity_field($orderItemQuantityDataMapper, FormBu
             ->add('quantity', 'integer', Argument::any())
             ->shouldBeCalled()
             ->willReturn($builder)
+            ->shouldBeCalled()
```
pamil: Doubled `shouldBeCalled()` (and this one placed after `willReturn()` :()

src/Sylius/Bundle/CoreBundle/Distributor/TaxesDistributor.php
pamil: I find it weird, in real life you're distributing $1000 into 10 parts, the ordering of the arguments seems wrong.

pamil:
Ooops, I've just realized I'd already reviewed this code in #3907, though many of comments were hidden by some not-related changes :)

src/Sylius/Bundle/CoreBundle/Distributor/TaxesDistributor.php
michalmarcinkowski: :+1:

Zales0123: As @pamil have written [here](https://github.com/Sylius/Sylius/pull/3868#issuecomment-173246708), I've fix it in https://github.com/Sylius/Sylius/pull/3907 (I mean arguments order ;))

src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
```diff
@@ -183,6 +186,19 @@
             <tag name="sylius.process.step" alias="sylius_checkout_finalize" />
         </service>
 
+        <service id="sylius.provider.default_zone_provider" class="%sylius.provider.default_zone_provider.class%">
+            <argument type="service">
+                <service
+                        factory-service="sylius.settings.manager"
+                        factory-method="loadSettings"
+                        class="Sylius\Bundle\SettingsBundle\Model\Settings"
+                        >
```
michalmarcinkowski: Strange formating

src/Sylius/Bundle/CoreBundle/spec/Distributor/IntegerDistributorSpec.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $this->distribute(4, 1000)->shouldReturn(array(250, 250, 250, 250));;
+        $this->distribute(4, -1000)->shouldReturn(array(-250, -250, -250, -250));;
+    }
+
+    function it_distributes_taxes_that_cannot_be_split_equally()
+    {
+        $this->distribute(3, 1000)->shouldReturn(array(334, 333, 333));;
+        $this->distribute(3, -1000)->shouldReturn(array(-334, -333, -333));;
+    }
```
michalmarcinkowski: Add spec for distributing `0` amount.

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
```diff
@@ -0,0 +1,176 @@
...
+            }
+
+            $percentageAmount = $rate->getAmountAsPercentage();
+            $totalTaxAmount = $this->taxCalculator->calculate($item->getTotal(), $rate);
+            $label = sprintf('%s (%s%%)', $rate->getName(), (float) $percentageAmount);
+
+            $splitTaxes = $this->integerDistributor->distribute($item->getQuantity(), $totalTaxAmount);
+
+            foreach ($splitTaxes as $key => $tax) {
+                $this->addAdjustment($item->getUnits()->get($key), $tax, $label, $rate->isIncludedInPrice());
```
michalmarcinkowski: Should we add adjustments with value `0`? IMO:

```php
if (0 === $tax) {
    continue; // or break?
}
```

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
```diff
@@ -0,0 +1,176 @@
...
+        foreach ($order->getItems() as $item) {
+            $rate = $this->taxRateResolver->resolve($item->getProduct(), array('zone' => $zone));
+
+            if (null === $rate) {
+                continue;
+            }
+
+            $percentageAmount = $rate->getAmountAsPercentage();
+            $totalTaxAmount = $this->taxCalculator->calculate($item->getTotal(), $rate);
+            $label = sprintf('%s (%s%%)', $rate->getName(), (float) $percentageAmount);
```
michalmarcinkowski: Maybe we could add a method `getLabel` to the `TaxRate` model, WDYT?

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
```diff
@@ -0,0 +1,176 @@
...
+     * @param string $label
+     * @param bool $included
+     */
+    protected function addAdjustment(OrderItemUnitInterface $unit, $amount, $label, $included)
+    {
+        $adjustment = $this->adjustmentFactory->createNew();
+        $adjustment->setType(AdjustmentInterface::TAX_ADJUSTMENT);
+        $adjustment->setAmount($amount);
+        $adjustment->setDescription($label);
+        $adjustment->setNeutral($included);
```
michalmarcinkowski: We should extract this to a factory.

src/Sylius/Bundle/OrderBundle/Aggregator/AdjustmentsAggregator.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+namespace Sylius\Bundle\OrderBundle\Aggregator;
+
+use Sylius\Component\Order\Model\AdjustmentInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class AdjustmentsAggregator implements AdjustmentsAggregatorInterface
```
michalmarcinkowski: How about `AdjustmentsByLabelAggregator`?

src/Sylius/Bundle/OrderBundle/spec/Twig/AggregateAdjustmentsExtensionSpec.php
```diff
@@ -0,0 +1,91 @@
...
+
+                if (!$key instanceof \Twig_SimpleFunction) {
+                    throw new FailureException('Key of "hasFunction" matcher must be \Twig_SimpleFunction object');
+                }
+
+                $contain = false;
+                /** @var \Twig_SimpleFunction $subjectElement */
+                foreach ($subject as $subjectElement) {
+                    if ($subjectElement->getName() === $key->getName() && $subjectElement->getCallable()[1] === $key->getCallable()[1]) {
+                        $contain = true;
```
michalmarcinkowski: `return true;`

src/Sylius/Bundle/OrderBundle/spec/Twig/AggregateAdjustmentsExtensionSpec.php
```diff
@@ -0,0 +1,91 @@
...
+
+                        break;
+                    }
+                }
+
+                return $contain;
```
michalmarcinkowski: `return false;`

src/Sylius/Component/Order/Model/Order.php
```diff
@@ -491,7 +504,7 @@ public function getAdjustmentsTotal($type = null)
         }
 
         $total = 0;
-        foreach ($this->getAdjustments($type) as $adjustment) {
+        foreach ($this->getAdjustmentsRecursively($type) as $adjustment) {
```
michalmarcinkowski: Why? `getAdjustmentsTotal` should return only adjustments on current level, do we need to get it recursively anywhere?

src/Sylius/Bundle/OrderBundle/Aggregator/AdjustmentsAggregator.php
pjedrzejewski: :+1:

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
Zales0123: :+1:

src/Sylius/Component/Order/Model/Order.php
Zales0123: Yep, I was wondering why is it here and I've just realized that I changed this but in other PR :D Surely it should be there.

src/Sylius/Bundle/CoreBundle/Distributor/IntegerDistributor.php
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function distribute($taxItems, $taxAmount)
+    {
+        if (!is_int($taxItems) || 1 > $taxItems) {
+            throw new \InvalidArgumentException('Tax items number must be an integer, bigger than 0.');
+        }
+
+        $sign = ($taxAmount < 0) ? -1 : 1;
```
stloyd: `()` is useless.

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
```diff
@@ -0,0 +1,176 @@
...
+    private function provideTaxZone(AddressInterface $shippingAddress = null)
+    {
+        $zone = null;
+        if (null !== $shippingAddress) {
+            $zone = $this->zoneMatcher->match($shippingAddress);
+        }
+
+        if (null === $zone) {
+            $zone = $this->defaultTaxZoneProvider->provide();
+        }
```
stloyd:
```php
if (null !== $shippingAddress) {
    $zone = $this->zoneMatcher->match($shippingAddress);
} else {
    $zone = $this->defaultTaxZoneProvider->provide();
}

return $zone;

File: src/Sylius/Component/Order/Model/OrderInterface.php [hunk 19]
```diff
@@ -153,4 +153,11 @@ public function getAdditionalInformation();
      * @param string $information
      */
     public function setAdditionalInformation($information);
+
+    /**
+     * @param null $type
+     *
+     * @return Collection
```
stloyd: It's not `Collection`, it's just `array`.

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
Zales0123: Not exactly. ``ZoneMatcher::match()`` could still return ``null`` and if it would, we should try to provide it with ``$defautTaxZoneProvider``

src/Sylius/Component/Order/spec/Model/OrderSpec.php
```diff
@@ -11,6 +11,7 @@
 
 namespace spec\Sylius\Component\Order\Model;
 
+use Doctrine\Common\Collections\Collection;
```
stloyd: Never used.

src/Sylius/Component/Order/Model/OrderItemInterface.php
```diff
@@ -89,4 +89,16 @@ public function addUnit(OrderItemUnitInterface $itemUnit);
      * @param OrderItemUnitInterface $itemUnit
      */
     public function removeUnit(OrderItemUnitInterface $itemUnit);
+
+    /**
+     * @param string|null $type
+     *
+     * @return Collection
```
stloyd: This is just plain `array`.

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
stloyd: Here you go :smile: :
```php
if (null !== $shippingAddress) {
    $zone = $this->zoneMatcher->match($shippingAddress);
}

return $zone ?: $this->defaultTaxZoneProvider->provide();

File: src/Sylius/Bundle/CoreBundle/Distributor/IntegerDistributor.php [hunk 22]
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class IntegerDistributor implements IntegerDistributorInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function distribute($taxItems, $taxAmount)
```
pjedrzejewski: This is integer distributor, not specific to taxes, so this should be: $numberOfTargets, $amount or something like that.

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
```diff
@@ -0,0 +1,176 @@
...
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\CoreBundle\OrderProcessing;
+
+use Sylius\Bundle\CoreBundle\Distributor\IntegerDistributorInterface;
+use Sylius\Bundle\CoreBundle\Provider\DefaultTaxZoneProviderInterface;
```
pjedrzejewski: Maybe this should go to ``Taxation`` namespace but definitely in the component, not core bundle!

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
```diff
@@ -0,0 +1,176 @@
...
+     * {@inheritdoc}
+     */
+    public function apply(OrderInterface $order)
+    {
+        $this->clearTaxes($order);
+        if ($order->isEmpty()) {
+            return;
+        }
+
+        $zone = $this->provideTaxZone($order->getShippingAddress());
```
pjedrzejewski: ``getTaxZone``.

In next iterations, I think we should have another interface ``OrderTaxZoneResolver|Provider::getTaxZone($order)``, which would be a nice abstraction layer.

* It would use DefaultTaxZoneProviderInterface + ZoneMatcher and maybe in future it would also allow to use billing address instead of shipping as an option.

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicatorInterface.php
```diff
@@ -9,22 +9,17 @@
  * file that was distributed with this source code.
  */
 
-namespace Sylius\Component\Core\OrderProcessing;
+namespace Sylius\Bundle\CoreBundle\OrderProcessing;
```
pjedrzejewski: Why moving it to the bundle???

src/Sylius/Bundle/CoreBundle/Provider/DefaultTaxZoneProvider.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+namespace Sylius\Bundle\CoreBundle\Provider;
+
+use Sylius\Bundle\SettingsBundle\Model\Settings;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class DefaultTaxZoneProvider implements DefaultTaxZoneProviderInterface
```
pjedrzejewski: Interface should go to the component! Remember that everything not coupled to Symfony goes to components.

src/Sylius/Bundle/CoreBundle/Provider/DefaultTaxZoneProvider.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    public function __construct(Settings $settings)
+    {
+        $this->settings = $settings;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function provide()
```
pjedrzejewski: I am not sure about this Provider -> provide naming... Not saying that ``DefaultTaxZoneProvider::getZone/getTaxZone/getDefaultTaxZone`` is better but I don't like both actually. :D

pjedrzejewski: I'd rename it to ``getZone`` simply because there is no convention about provide() and getX is more common in Sylius, so let's not mix.

src/Sylius/Bundle/OrderBundle/Resources/config/services.xml
```diff
@@ -26,6 +26,9 @@
         <parameter key="sylius.order_item_quantity_modifier.class">Sylius\Bundle\OrderBundle\Modifier\OrderItemQuantityModifier</parameter>
         <parameter key="sylius.form.data_mapper.property_path.class">Symfony\Component\Form\Extension\Core\DataMapper\PropertyPathMapper</parameter>
         <parameter key="sylius.form.data_mapper.order_item_quantity.class">Sylius\Bundle\OrderBundle\Form\DataMapper\OrderItemQuantityDataMapper</parameter>
+
+        <parameter key="sylius.aggregator.adjustments.class">Sylius\Bundle\OrderBundle\Aggregator\AdjustmentsByLabelAggregator</parameter>
```
pjedrzejewski: We do not have any other aggregators and this is not a generic concept for all resources, etc. I'd say here ``sylius.adjustments_aggregator``.

src/Sylius/Bundle/OrderBundle/Resources/config/services.xml
```diff
@@ -49,5 +52,11 @@
             <argument type="service" id="sylius.order_item_quantity_modifier" />
             <argument type="service" id="sylius.form.data_mapper.property_path" />
         </service>
+
+        <service id="sylius.aggregator.adjustments" class="%sylius.aggregator.adjustments.class%" />
+        <service id="sylius.twig.extension.aggregate_adjustments" class="%sylius.twig.extension.aggregate_adjustments.class%" public="false">
```
pjedrzejewski: We put twig extensions in separate ``twig.xml`` files.

src/Sylius/Bundle/OrderBundle/Twig/AggregateAdjustmentsExtension.php
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+namespace Sylius\Bundle\OrderBundle\Twig;
+
+use Sylius\Bundle\OrderBundle\Aggregator\AdjustmentsAggregatorInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class AggregateAdjustmentsExtension extends \Twig_Extension
```
pjedrzejewski: We should create a templating helper and inject it here, like we do for rest of Twig Extensions https://github.com/Sylius/Sylius/tree/master/src/Sylius/Bundle/SettingsBundle/Templating/Helper.

src/Sylius/Bundle/WebBundle/Resources/views/Backend/Order/show.html.twig
```diff
@@ -130,10 +130,10 @@
         <tr class="taxes">
             <td colspan="2">
                 <h5>{{ 'sylius.order.taxes'|trans }}</h5>
-                {% if not order.adjustments(taxAdjustment).isEmpty() %}
+                {% if not order.getAdjustmentsRecursively(taxAdjustment) is empty %}
```
pjedrzejewski: We are getting it twice here and in line 135. Let's set it as variable.

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
Zales0123: :+1:

src/Sylius/Bundle/CoreBundle/OrderProcessing/OrderTaxesApplicator.php
michalmarcinkowski: Maybe we can start with simply `TaxZoneResolver::getTaxZone($order->getShippingAddress())`?

src/Sylius/Bundle/CoreBundle/Provider/DefaultTaxZoneProvider.php
michalmarcinkowski: :+1:

src/Sylius/Bundle/OrderBundle/Twig/AggregateAdjustmentsExtension.php
Zales0123: :+1:

src/Sylius/Bundle/CoreBundle/Provider/DefaultTaxZoneProvider.php
Zales0123: done :+1:

Zales0123:
@pjedrzejewski I think it's ready ;)

pjedrzejewski:
Awesome. Thanks Mateusz and everyone involved! :+1:

Zales0123:
Great! :tada: Thanks guys for help and comments, both on programming and business level ;) #rebaseTime

michalmarcinkowski:
:+1: Let's face the shipping taxes now!

patrick-mcdougle:
Where did we land on this? Taxes on the `orderitem` level or the `orderitemunit` level? To add more complexity to this conversation, here in the US taxes are often based on the destination of the shipment and if we want to deal with multi-address shipping at some point in the future #2842 I think pinning the taxes on the orderItemUnit is the best solution.

Issue #2739, opened 2015-05-15
[Promotion] Applied to orders after adding


rvanlaak:
Our custom implementation of the PromotionBundle separates adding the Rules and Actions from adding a Promotion. We first add a Promotion, and thereafter add the Rules and Actions. 

This can result in an empty promotion that will be applied to an order by the processor. The documentation already states that a `Promotion` without an `Action` is not effective, but I think this should be solved in the `PromotionProcessor` directly:

1. By counting the number of Actions and not process the Promotion any further if there are no actions
2. By introducing an 'active' flag, and let the `PromotionRepository` take that flag into account.

The second option has my :+1: because otherwise the same will occur between adding two actions.

pjedrzejewski:
Let's add ``enabled`` boolean flag to Promotions. But after #3098 is merged.

Rvanlaak:
:+1:

pjedrzejewski:
#3098 is merged, so if someone is willing to do a PR with this change, it will be awesome! :+1:

rvanlaak:
Will add this on my TODO list for the PromotionBundle, but will changing the types in there be enough to make this work in the WebBundle?

PR #3342, opened 2015-09-23
[Product] Implement Toggleable interface on Product

TheMadeleine:
| Q                    | A
| -------------       | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3213
| License       | MIT

PR #3871, opened 2016-01-14
[Promotion] make Toggleable

rvanlaak:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | #2739 , #3098 |
| License | MIT |
| Doc PR |  |
#3098 did implement the `ToggleableInterface`, which also should be applied to `Promotion` because others can use their own custom implementation and don't have `Rules` when a promotion is added.

Because `0.17` bumped php to `>5.4` the actual implementation that already was there at the `Country` model has been refactored to a trait.

This PR probably isn't ready to merge, because we don't use all the Sylius components. Do we need to make changes to the web bundle?

### DIFF
--- a/src/Sylius/Component/Addressing/Model/Country.php
+++ b/src/Sylius/Component/Addressing/Model/Country.php
@@ -13,6 +13,7 @@ namespace Sylius\Component\Addressing\Model;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
+use Sylius\Component\Resource\Model\ToggleableTrait;
 use Symfony\Component\Intl\Intl;
 
 /**
@@ -22,6 +23,8 @@ use Symfony\Component\Intl\Intl;
  */
 class Country implements CountryInterface
 {
+    use ToggleableTrait;
+
     /**
      * @var mixed
      */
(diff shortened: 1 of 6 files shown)


stloyd:
Please wait for #3264, I will try to fix remaining problems today.

rvanlaak:
Ah, did not find that one while doing a quick search for trait. Concerning the promotion, do I miss something related to the web bundle?

rvanlaak:
@stloyd #3264 was merged, you want me to do this over again? Does the current implementation miss something?

rvanlaak:
ping @stloyd

okwinza:
@rvanlaak i'd say you can start with rebase and making the build green. Also please use traits from #3264 and, if you want to somehow refactor them -- do it in different PR.

And behat scenario covering enabling/disabling is required too, i guess.
You can use #3342 as a reference.
Overall :+1: for the feature.

pjedrzejewski:
Hey @rvanlaak, would be great to have this in core. One change I'd suggest is to filter out not enabled promotions on the repository level, for performance reasons. IIRC there is a ``findActive`` method that checks the dates. So we should simply check for ``enabled`` field there too. Oh, and for WebBundle, checkbox on promotion form will be enough. :+1:

PR #3873, opened 2016-01-14
[Archetype] Fix ParentArchetypeListener

lchrusciel:
The previous implementations overrode `$parentOptions` defined above with *query_builder* option.


pjedrzejewski:
Thank you Łukasz!

PR #3814, opened 2016-01-05
[InstallerBundle] Simplification and testing of sylius:install:setup command

TheMadeleine:
| Q                    | A
| -------------       | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | 
| License       | MIT

Removed multiple currencies, languages, countries and channels configuration.

Now when the currency is chosen it is set as `base`.

Added Behat testing of the command.

PR #3887, opened 2016-01-15
[Currency] Add isBase flag on CurrencyInterface

TheMadeleine:
| Q                    | A
| -------------       | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | 
| License       | MIT

Added all methods required to set a Currency as a base one. This PR is connected with #3814, because now during the installation we get only one currency from the user, and we want to set it as the base one.

### DIFF
--- a/src/Sylius/Bundle/CurrencyBundle/Resources/config/doctrine/model/Currency.mongodb.xml
+++ b/src/Sylius/Bundle/CurrencyBundle/Resources/config/doctrine/model/Currency.mongodb.xml
@@ -23,6 +23,7 @@
         <field fieldName="code" name="code" type="string" length="3" />
         <field fieldName="exchangeRate" name="exchangeRate" type="float" />
         <field fieldName="enabled" name="enabled" type="boolean" />
+        <field name="base" column="base" type="boolean" />
 
         <field fieldName="createdAt" name="createdAt" type="date">
             <gedmo:timestampable on="create"/>
--- a/src/Sylius/Component/Currency/spec/Model/CurrencySpec.php
+++ b/src/Sylius/Component/Currency/spec/Model/CurrencySpec.php
@@ -91,6 +91,17 @@ class CurrencySpec extends ObjectBehavior
         $this->shouldNotBeEnabled();
     }
 
+    function it_is_not_base_currency_by_default()
+    {
+        $this->shouldNotBeBase();
+    }
+
+    function it_can_can_be_base_currency()
+    {
+        $this->setBase(true);
+        $this->shouldBeBase();
+    }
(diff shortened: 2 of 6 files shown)


pamil:
Somehow I find `isBase`counterintuitive, is there any reason why wouldn't we go for `isDefault`?
Also, we need some way to ensure that there is only one currency with that field enabled.

pjedrzejewski:
@pamil @TheMadeleine DefaultableInterface :D

src/Sylius/Bundle/CurrencyBundle/Resources/config/doctrine/model/Currency.mongodb.xml
```diff
@@ -23,6 +23,7 @@
         <field fieldName="code" name="code" type="string" length="3" />
         <field fieldName="exchangeRate" name="exchangeRate" type="float" />
         <field fieldName="enabled" name="enabled" type="boolean" />
+        <field name="isBase" column="is-base" type="boolean" />
```
stloyd: Hmmm, every other column is called in camelCase, why changing approach for this one?

src/Sylius/Component/Currency/spec/Model/CurrencySpec.php
```diff
@@ -91,6 +91,17 @@ function it_can_set_enabled_value()
...
 
+    function it_is_not_base_currency_by_default()
+    {
+        $this->isBase()->shouldReturn(false);
+    }
+
+    function it_can_can_be_base_currency()
+    {
+        $this->setBase(true);
+        $this->isBase()->shouldReturn(true);
```
stloyd:
```php
$this->shouldBeBase();

File: src/Sylius/Component/Currency/spec/Model/CurrencySpec.php [hunk 3]
```diff
@@ -91,6 +91,17 @@ function it_can_set_enabled_value()
         $this->shouldNotBeEnabled();
     }
 
+    function it_is_not_base_currency_by_default()
+    {
+        $this->isBase()->shouldReturn(false);
```
stloyd:
```php
$this->shouldNotBeBase();

src/Sylius/Component/Currency/spec/Model/CurrencySpec.php
pamil: IMHO the current code is more verbose and easier to understand than `shouldBeBase()`.

pjedrzejewski:
Thanks Magda!

PR #3891, opened 2016-01-15
[InstallerBundle] Removed web installer heritage

pamil:
Removed residues left after web installer had been removed.


pjedrzejewski:
:+1:

PR #3895, opened 2016-01-15
[Currency] Base currency's enabled status and exchange rate should not be editable

TheMadeleine:
| Q                    | A
| -------------       | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| related PRs | #3814 #3887 
| License       | MIT

### DIFF
--- a/src/Sylius/Component/Currency/Model/Currency.php
+++ b/src/Sylius/Component/Currency/Model/Currency.php
@@ -111,6 +111,10 @@ class Currency implements CurrencyInterface
      */
     public function setExchangeRate($rate)
     {
+        if ($this->isBase()) {
+            throw new \LogicException('You cannot change the exchange rate of the base currency!');
+        }
+
         $this->exchangeRate = $rate;
     }
 
@@ -127,6 +131,10 @@ class Currency implements CurrencyInterface
      */
     public function setEnabled($enabled)
     {
+        if ($this->isBase()) {
+            throw new \LogicException('You cannot change the enabled status of the base currency!');
+        }
+
         $this->enabled = (bool) $enabled;
     }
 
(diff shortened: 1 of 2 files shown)


src/Sylius/Component/Currency/Model/Currency.php
```diff
@@ -111,7 +111,10 @@ public function getExchangeRate()
      */
     public function setExchangeRate($rate)
     {
-        $this->exchangeRate = $rate;
+        if(!$this->isBase())
+        {
```
pjedrzejewski: CS

src/Sylius/Component/Currency/Model/Currency.php
```diff
@@ -175,7 +178,9 @@ public function enable()
      */
     public function disable()
     {
-        $this->enabled = false;
+        if(!$this->isBase()) {
```
pjedrzejewski: CS

pjedrzejewski:
I think it should throw an exception. Otherwise it is a bit hidden behavior.

michalmarcinkowski:
Definitely it should throw an exception. IMO `\LogicException` will be appropriate.

src/Sylius/Component/Currency/Model/Currency.php
pamil:
```php
if($this->isBase()) {
  // throw exception / return;
}

$this-exchangeRate = $rate;
```

This way the "happy" path goes at the base indentation level :)

TheMadeleine:
:+1:  :)

pjedrzejewski:
@TheMadeleine Don't you think it would be useful to add a message? :P

pjedrzejewski:
Thanks Magda!

Issue #3897, opened 2016-01-15
Error when creating a imagine block


pfwd:
Login to admin area
Click Blocks under Content
Click Imagine Block
Click Create Imagine Block
Notice error
When doing this locally I get this error on Dev:

Catchable Fatal Error: Argument 3 passed to Sylius\Bundle\ContentBundle\Form\Type\ImagineBlockType::__construct() must be an instance of Liip\ImagineBundle\Imagine\Filter\FilterConfiguration, none given, called in /app/cache/dev/appDevDebugProjectContainer.php on line 12323 and defined

This can be replicated on the demo site

lighta971:
Me too after updating 0.16-dev.
The Bundle/ContentBundle/Resources/config/services.xml file disapeared.

pjedrzejewski:
Fixed in #3900.

pjedrzejewski:
Thanks for the report!

PR #3900, opened 2016-01-16
[ContentBundle] Imagine block fix

okwinza:
The issue described in #3897 was introduced in #3721 
I didnt find a better way to fix it so i just put the missing lines back for now.

Here is latest Travis build job for ContentBundle to illustrate the problem: https://travis-ci.org/Sylius/SyliusContentBundle/jobs/102600357

Let me know what you guys think.

P.S.
Maybe it makes sence to tag a patch release after this fix?


pjedrzejewski:
@okwinza Unfortunately we can't do that easily because we do not maintain separate branches until 1.0.0. Thanks for the fix! :+1:

PR #3908, opened 2016-01-18
[WebBundle] Translations cleanup

pamil:
- [x] Translations ordered alphabetically
 - [x] Removed punctuations marks at the end of translation
 - [x] Every translation starts with a capital letter

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -400,6 +400,7 @@ default:
     extensions:
         Lakion\Behat\MinkDebugExtension:
             directory: etc/build
+            clean_start: false
         Behat\MinkExtension:
             sessions:
                 default:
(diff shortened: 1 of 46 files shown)


behat.yml.dist
```diff
@@ -400,6 +400,7 @@ default:
     extensions:
         Lakion\Behat\MinkDebugExtension:
             directory: etc/build
+            clean_start: false
```
pamil: I've smuggled a tiny change here to make logs of failing scenarios available (as `MinkDebugExtension.clean_start=true` by default and it cleared the `etc/build/` directory).

pjedrzejewski:
Thanks Kamil! :+1: Also, thanks for smuggling changes explicitly!

PR #3854, opened 2016-01-12
Repository clean up and TranslatableResourceRepository deletion

lchrusciel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #2614
| License       | MIT
| Doc PR        | 

I've made some clean ups in repositories and introduced some new repository interfaces. Moreover, I have also removed TranslatableResourceRepository and translation fields from configuration. It will reduce amount of magic in our repositories and will result in explicit calls to them. I am aware that this is a huge BC break, but it is better to do such changes earlier than later ;) 

*Update*
**NOTE:** In order to be as explicit as it is possible, I removed all magic methods and functions like *getQueryBuilder*. Repositories will be much more readable if we will call methods directly, where one has to declare every used join. Previous implementation was hard to maintain as some times we used some objects joined in a parent class of parent class of parent class of the class we used. This was misleading and generated enormous amount of bugs. I removed some spec also, cuz implementation of repository has been changed and it's logic is tested anyway in a behat scenarios. 

*Uodate2*
**NOTE:** Contains changes from #4271

PR #3910, opened 2016-01-18
Removed repositories specifications

pamil:
As we keep removing repositories specifications in #3854, we can also get rid of these ones in ResourceBundle & TranslationBundle. Repositories are located at the boundaries of the project and we do not benefit from having them - PhpSpec cannot be used here as design tool and also sucks when it comes to testing them (e.g. [wrong assumptions](https://github.com/Sylius/Sylius/blob/0b5866d2773788c41aa99e429bfb257993677f5d/src/Sylius/Bundle/ResourceBundle/spec/Doctrine/ORM/EntityRepositorySpec.php#L182-L185)).

Moreover, we can make our Travis build simplier by removing these, as we will no longer need support for cached `composer.json` and `composer.lock` files.


pjedrzejewski:
:+1:

PR #3918, opened 2016-01-19
[Core] Refactor order checkout state machine

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

Hello everybody! This PR is the result of recent brainstorm with @pjedrzejewski and @michalmarcinkowski about cart checkout flow. We've decided to refactor *Cart* state machine and, finally, use it to handle whole checkout logic to get rid of these nasty cart/checkout events and listeners.

And this is the first step. ``sylius_order_checkout`` state machine has ben reworked to have better named states and transitions. Whole checkout workflow now looks like this:
checkout

Besides, there are proper transitions applied on ``Cart`` object in checkout steps. In the nearest future, *cart* and *checkout* events and listeners will be removed and replaced with proper state machine callbacks.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Checkout/Step/FinalizeStep.php
+++ b/src/Sylius/Bundle/CoreBundle/Checkout/Step/FinalizeStep.php
@@ -18,8 +18,6 @@ use Sylius\Component\Core\SyliusOrderEvents;
 use Sylius\Component\Order\OrderTransitions;
 
 /**
- * Final checkout step.
- *
  * @author Paweł Jędrzejewski <pawel@sylius.org>
  * @author Fernando Caraballo Ortiz <caraballo.ortiz@gmail.com>
  */
@@ -49,6 +47,12 @@ class FinalizeStep extends CheckoutStep
         return $this->complete();
     }
 
+    /**
+     * @param ProcessContextInterface $context
+     * @param OrderInterface $order
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
     protected function renderStep(ProcessContextInterface $context, OrderInterface $order)
     {
         return $this->render($this->container->getParameter(sprintf('sylius.checkout.step.%s.template', $this->getName())), [
@@ -58,8 +62,6 @@ class FinalizeStep extends CheckoutStep
     }
 
     /**
-     * Mark the order as completed.
-     *
      * @param OrderInterface $order
      */
     protected function completeOrder(OrderInterface $order)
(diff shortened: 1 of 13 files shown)


src/Sylius/Bundle/CoreBundle/Checkout/Step/FinalizeStep.php
```diff
@@ -66,6 +68,7 @@ protected function completeOrder(OrderInterface $order)
         $this->dispatchCheckoutEvent(SyliusOrderEvents::PRE_CREATE, $order);
         $this->dispatchCheckoutEvent(SyliusCheckoutEvents::FINALIZE_PRE_COMPLETE, $order);
 
+        $this->applyTransition('complete', $order);
         $this->get('sm.factory')->get($order, OrderTransitions::GRAPH)->apply(OrderTransitions::SYLIUS_CREATE, true);
```
stloyd: Seems that this is some left-over here, or if not IMO should be refactored to new approach.

src/Sylius/Bundle/CoreBundle/Checkout/Step/PaymentStep.php
```diff
@@ -63,6 +67,13 @@ public function forwardAction(ProcessContextInterface $context)
         return $this->renderStep($context, $order, $form);
     }
 
+    /**
+     * @param ProcessContextInterface $context
+     * @param OrderInterface $order
+     * @param FormInterface $form
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
```
stloyd: Don't use FQCN.

src/Sylius/Bundle/CoreBundle/Checkout/Step/ShippingStep.php
```diff
@@ -61,9 +61,11 @@ public function forwardAction(ProcessContextInterface $context)
 
         $form = $this->createCheckoutShippingForm($order);
 
-        if ($form->handleRequest($request)->isValid()) {
+         if ($form->handleRequest($request)->isValid()) {
```
stloyd: Should be reverted.

src/Sylius/Bundle/CoreBundle/Checkout/Step/FinalizeStep.php
Zales0123: It's because there are two different state machines here - ``applyTransition`` function applies transitions from ``sylius_order_checkout`` state machine, this line however runs transition from ``sylius_order``.
I don't know should we create another function (or pass graph name as parameter) only for this single call.

pjedrzejewski: We want to maintain most of current behavior to avoid too many BC breaks in a single version and too many changes in a single PR.

peteward:
This sounds sensible and easy enough to customise through state-machine configuration (our flow is a bit different to this, but a bit of configuration and change of flow sounds doable).

It's also easy to read and understand.

Would you define constants for the transitions as well as the states?

pjedrzejewski:
@peteward Yeah, I think we should add them. /cc @Zales0123

pjedrzejewski:
Thank you Mateusz! :+1:

Issue #3914, opened 2016-01-19
AttributeValue __toString() invalid


peteward:
Since the value is typed this no longer works. You could cast with (string) but this probably won't have the desired affect on date fields so it should probably just be removed?

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Attribute/Model/AttributeValue.php#L75

peteward:
Although it looks like the new model for attributes is still expecting __toString() to work:

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/WebBundle/Resources/views/Frontend/Product/_attribute.html.twig#L1

Also this looks broken 'oldhref' so if someone is going to look at this: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/AttributeBundle/Resources/views/attributeChoice.html.twig#L13

I'm not sure how well tested this is with datetimes, but maybe a twig function is required to render attribute values, because casting like this won't work.

nakashu:
@peteward oldhref is doing nothing there, its probably some kind of typo.
I fixed it in enhancment PR https://github.com/Sylius/Sylius/pull/3797 but could not get the js test to pass (don't have time test it in nix where I can run the tests localy)

peteward:
OK cool, looks like some nice improvements there but doesn't solve the underlying issues around types.

nakashu:
Sure it was a bit off topic, just for the oldhref thingy

Zales0123:
@peteward about [this file](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/WebBundle/Resources/views/Frontend/Product/_attribute.html.twig#L1) I think it's confusing naming, because ``attribute`` here is really a ``AttributeValue`` object, so getting its ``value`` is 100% correct ;)

[...]

Issue #3913, opened 2016-01-19
DatetimeAttributeType - wrong casing?


peteward:
Shouldn't this be all lower case?

https://github.com/Sylius/Sylius/blob/50cfdb51f8c3617b106312dc3acea3c6cb44df8b/src/Sylius/Bundle/AttributeBundle/Resources/config/doctrine/model/AttributeValue.orm.xml#L28

Apologies I don't have time to check this right now but currently `dateTime` only works in DB queries but there's no other reference of it with this case anywhere, it's all lower case as `datetime` in the storage definitions:

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Attribute/Model/AttributeValueInterface.php#L24

It could be fine, it just seems like it could be wrong...

peteward:
This looks fixed in the above PR, I'll reopen if there are any further issues but should be fine.

pjedrzejewski:
@peteward Thanks for closing, that helps a lot! :+1:

PR #3921, opened 2016-01-19
[Attribute] Fixed some minor issues with attributes

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Sylius/Sylius/issues/3914, https://github.com/Sylius/Sylius/issues/3913
| License       | MIT
| Doc PR        |

- changed ``dateTime`` field to ``datetime``
- removed weird ``oldhref`` from attribute choice element
- removed broken ``__toString()`` from ``AttributeValue``

### DIFF
--- a/src/Sylius/Bundle/AttributeBundle/Resources/views/attributeChoice.html.twig
+++ b/src/Sylius/Bundle/AttributeBundle/Resources/views/attributeChoice.html.twig
@@ -10,9 +10,9 @@
                     <div class="list-group row">
                         {% for attributeField in form %}
                             {% if attributeField.vars.name != '_token' %}
-                            <a oldhref class="list-group-item col-xs-12 col-md-6">
+                            <div class="list-group-item col-xs-12 col-md-6" style="cursor: default;">
                                 {{ attributeField.vars.label }} {{ form_widget(attributeField, { 'id': attributeField.vars.label~' attribute' }) }}
-                            </a>
+                            </div>
                             {% endif %}
                         {% endfor %}
                     </div>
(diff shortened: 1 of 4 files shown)


src/Sylius/Bundle/AttributeBundle/Resources/views/attributeChoice.html.twig
```diff
@@ -10,9 +10,9 @@
                     <div class="list-group row">
```
peteward: Hi @Zales0123,

I'm just updating our backend for an entity to use this attribute UI - looking good! 

The only issue I'm having is that this line here is coupled to ProductAttributes (and I need to use it for a different model), as you can see from the routing:

```
sylius_backend_render_attribute_forms:
    path: /attribute-forms
    methods: [GET]
    defaults:
        _controller: sylius.controller.product_attribute:renderAttributeValueFormsAction
```

Is this something that you can look at?

pjedrzejewski: Good catch, I think we have one more coupling issue in the CompilerPass. It assumes there is ``product_attribute`` service and that's incorrect.

pjedrzejewski:
Thanks Mateusz!

src/Sylius/Bundle/AttributeBundle/Resources/views/attributeChoice.html.twig
Zales0123: Sure, these are things that have to be fixed :+1: I'm sure I can handle these issues in the nearest future.

PR #3927, opened 2016-01-20
Validate composer

bendavies:
The current `composer.lock` hash is out out date, so `composer install` shows a warning.
Fix that, and prevent from happening again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,8 @@ before_script:
     - php -i | grep -v GITHUB_OAUTH_TOKEN
 
 script:
+    - composer validate
+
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -66,6 +66,8 @@ before_script:
     - php -i | grep -v GITHUB_OAUTH_TOKEN
 
 script:
+    - composer validate
```
pamil: :+1:

pjedrzejewski:
Nice! :+1:

michalmarcinkowski:
Thanks Ben!

Issue #3912, opened 2016-01-18
[RFC][ApiBundle] Sparse Fieldsets 


umpirsky:
Sometimes with bigger numbers of items per page and serialization config which product lots of nested data structures, response time can be more then several seconds and response size can measure in MB.

What would be cool (not sure if possible), is to allow client to request that an endpoint return only specific fields in the response on a per-type basis by including a fields[TYPE] parameter - [sparse fieldsets](http://jsonapi.org/format/#fetching-sparse-fieldsets).

I would like to know if something like this is supported by REST libraries we use.

pjedrzejewski:
@umpirsky Most of resources have very rough serializer configuration, but we will be tweaking it one by one. We can use serialization groups for that, I am not sure if it would be easy to support sparse fields though.

umpirsky:
@pjedrzejewski Can serialization group be specified in the routing config?

nakashu:
@umpirsky @pjedrzejewski field exclusion should not be hard.

Working solution is here: http://jolicode.com/blog/how-to-implement-your-own-fields-inclusion-rules-with-jms-serializer
For the sake if the site would die, I'll this here also, it the same just formated to answer on stack.
http://stackoverflow.com/questions/33126812/symfony-fosrestbundle-show-selected-fields

pjedrzejewski:
@umpirsky Yes:

```
...
    _sylius:
        serialization_groups: [foo, bar]
```

pjedrzejewski:
@nakashu Nice, we could use that! :+1: Thanks for the link.

umpirsky:
@nakashu @pjedrzejewski Very cool, thanks for the info guys. Will close this now as it seems to be pretty flexible. Will try it out. :+1:

umpirsky:
@pjedrzejewski Maybe I can contribute this thing? It works really well.

pjedrzejewski:
@umpirsky That would be really cool!

umpirsky:
@pjedrzejewski Qick and dirty WIP pushed https://github.com/Sylius/Sylius/pull/3938.

PR #3762, opened 2015-12-21
[ResourceBundle] New ResourceController

pjedrzejewski:
Hello folks,

One of the most important classes in whole Sylius was without any specs and getting really ugly. I am perfectly aware that this can be done in a cleaner way but that's just first iteration. When we have specs, then we can refactor it further.

Nicer diff: https://github.com/pjedrzejewski/Sylius/blob/new-resource-controller/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php. :P

Major changes overview:

* RequestConfiguration is no longer magically set on the controller via listener - it is created inside action instead.
* Most of logic is extracted to ResourceController specific services so this gives us ability to easily change behavior of the controller and also you can inject specific implementation per resource.

Known issues:
* I have not spec enable/disable actions and not going to that for moveUp/Down etc. because we want to get rid of these. They update just one field on the model and this is messy. Instead we should use PATCH calls to updateAction, which allows us much more user-friendly UI and gets rid of a lot of code. That being said, this is a change for a separate PR.
* Not entirely API friendly as previous controller, I want to wait for #3701 as it will be much easier for me to implement proper behavior with tests in place. :)
* I am not touching any logic in other controllers, just making them compatible, so there will be ugly changes like ``$this->container->get('templating')`` etc. but I don't want to refactor whole Sylius in one run. :)
* Naming of service may be not ideal.
* Missing moveUp/moveDownActions & updateStateAction.

Appreciate any ideas, suggestions and of course review. :)

PR #3938, opened 2016-01-20
[WIP][ApiBundle] Sparse Fieldsets

umpirsky:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | #3912 |
| License | MIT |

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Serializer/Exclusion/SparseFieldsetsExclusionStrategy.php
+++ b/src/Sylius/Bundle/ApiBundle/Serializer/Exclusion/SparseFieldsetsExclusionStrategy.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\ApiBundle\Serializer\Exclusion;
+
+use JMS\Serializer\Exclusion\ExclusionStrategyInterface;
+use JMS\Serializer\Metadata\ClassMetadata;
+use JMS\Serializer\Metadata\PropertyMetadata;
+use JMS\Serializer\Context;
+
+class SparseFieldsetsExclusionStrategy implements ExclusionStrategyInterface
+{
+    private $fields;
+
+    public function __construct(array $fields)
+    {
+        $this->fields = $fields;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function shouldSkipClass(ClassMetadata $metadata, Context $navigatorContext)
+    {
+        return false;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function shouldSkipProperty(PropertyMetadata $property, Context $navigatorContext)
+    {
+        $name = $property->serializedName ?: $property->name;
+
+        if ($property->inline || null === $name) {
+            return false;
+        }
+
+        $fields = $this->getFields($property, $navigatorContext);
+        if (null === $fields) {
+            return false;
+        }
+
+        return !in_array($name, $fields);
+    }
+
+    private function getFields(PropertyMetadata $property, Context $navigatorContext)
+    {
+        $key = $navigatorContext->getMetadataFactory()->getMetadataForClass($property->class)->xmlRootName;
+        if (null === $key || !isset($this->fields[$key])) {
+            return null;
+        }
+
+        return $this->fields[$key];
+    }
+}
--- a/src/Sylius/Bundle/ApiBundle/spec/Serializer/Exclusion/SparseFieldsetsExclusionStrategySpec.php
+++ b/src/Sylius/Bundle/ApiBundle/spec/Serializer/Exclusion/SparseFieldsetsExclusionStrategySpec.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace spec\Sylius\Bundle\ApiBundle\Serializer\Exclusion;
+
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use JMS\Serializer\Exclusion\ExclusionStrategyInterface;
+use JMS\Serializer\Metadata\ClassMetadata;
+use JMS\Serializer\Metadata\PropertyMetadata;
+use JMS\Serializer\Context;
+
+
+class SparseFieldsetsExclusionStrategySpec extends ObjectBehavior
+{
+    function let()
+    {
+        $this->beConstructedWith(['foo', 'bar']);
+    }
+
+    function it_is_initializable()
+    {
+        $this->shouldHaveType('Sylius\Bundle\ApiBundle\Serializer\Exclusion\SparseFieldsetsExclusionStrategy');
+    }
+
+    function it_implements_exclusion_strategy_interface()
+    {
+        $this->shouldImplement(ExclusionStrategyInterface::class);
+    }
+
+    function it_should_never_skip_class(PropertyMetadata $property, Context $navigatorContext)
+    {
+         $this->callOnWrappedObject(
+            'shouldSkipProperty',
+            [$property->getWrappedObject(), $navigatorContext->getWrappedObject()]
+        )->shouldReturn(false);
+    }
+}
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
@@ -107,6 +107,28 @@
     </parameters>
 
     <services>
+        <!-- todo -->
+        <service id="sylius.view_handler.default" class="Sylius\Bundle\CoreBundle\View\ViewHandler" decorates="fos_rest.view_handler.default">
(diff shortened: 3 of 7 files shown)


src/Sylius/Bundle/ApiBundle/spec/Serializer/Exclusion/SparseFieldsetsExclusionStrategySpec.php
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    function it_implements_exclusion_strategy_interface()
+    {
+        $this->shouldImplement(ExclusionStrategyInterface::class);
+    }
+
+    function it_should_never_skip_class(PropertyMetadata $property, Context $navigatorContext)
+    {
+        $this->shouldSkipProperty($property, $navigatorContext)->shouldReturn(false);
```
umpirsky: This is silly:

```
./bin/phpspec run src/Sylius/Bundle/ApiBundle/spec/Serializer
Sylius/Bundle/ApiBundle/Serializer/Exclusion/SparseFieldsetsExclusionStrategy     
  30  - it should never skip class
      no skipProperty([array:2]) matcher found for
      [obj:Sylius\Bundle\ApiBundle\Serializer\Exclusion\SparseFieldsetsExclusionStrategy].

                         66%                                     33%             3
1 specs
3 examples (2 passed, 1 broken)
15ms
```

src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
```diff
@@ -107,6 +107,28 @@
     </parameters>
 
     <services>
+        <!-- todo -->
+        <service id="sylius.view_handler.default" class="Sylius\Bundle\CoreBundle\View\ViewHandler" decorates="fos_rest.view_handler.default">
```
umpirsky: @pjedrzejewski What is the best way to override it?

src/Sylius/Bundle/ApiBundle/Serializer/Exclusion/SparseFieldsetsExclusionStrategy.php
```diff
@@ -0,0 +1,64 @@
...
+        if (null === $fields) {
+            return false;
+        }
+
+        return !in_array($name, $fields);
+    }
+
+    private function getFields(PropertyMetadata $property, Context $navigatorContext)
+    {
+        $key = $navigatorContext->getMetadataFactory()->getMetadataForClass($property->class)->xmlRootName;
```
umpirsky: We include by class metadata here. My idea is to do it based on property metadata, but I couldn't find a way to get it.

pjedrzejewski:
I hope you like ApiTestCase. :D We should definitely give more thought about security here because I can imagine this would allow to serialize password field from users. ;) Would be cool to specify which fields are only permitted.

Also, the view handler may not be necessary after #3762 is merged today as it has custom view handler already implemented so we could handle it there. WDYT?

umpirsky:
@pjedrzejewski Yes, it's cool in deed. :)

Good point about security. But specifying which fields are permitted may be lots of work?
Maybe blacklist instead...but still can't think of something elegant.

Also https://github.com/Sylius/Sylius/pull/3938#discussion_r50320665 is something I would like to fix.

Good work on #3762, will wait for this one to get merged first. :+1:

nakashu:
From experience it is worth to specify explicitly which fields are allowed. 
You have all fields exlcuded by default, and only exclude the default set, which the user can make even sorter by defining which field he wants.

Why is it better? Because with the combination of excluding specific fields and serialization groups you will eliminate infinite nested serialization on associated resources.
For example User, Groups, Permission triangle
User belongs to groups and have permission.
Groups have permissions and users.
Permissions have users and groups.

I can provide an code example how it is solved by, exclusion and serialization groups, if needed.

umpirsky:
@nakashu Thanks for your feedback, any example can help so if you can share please do. :+1:

nakashu:
Quick Serialization groups explanation:

 * <resourceName>Detail serializes that property only for single resource retrieval. / getByID
 * <resourceName>List serializes that property only for listed/paginated resource retrieval.

```php
/**
* @ExclusionPolicy("all")
*/
class User implements UserInterface 
{
    /**
     * @Expose
     * @var integer
     */
    private $id;

    // note that password is NOT exposed, ever.
    /**
     * @var string
     */
    private $password;

    /**
     * @Expose
     * @Groups({"userDetail"})
     * @var Group[]|ArrayCollection
     */
    private $groups;

    /**
     * @Expose
     * @Groups({"userDetail", "groupDetail"})
     * @var Privilege[]|ArrayCollection
     */
    private $privileges;
    // ... other properties and methods
}
/**
* @ExclusionPolicy("all")
*/
class Group implements GroupInterface
{
    // ... other properties
    /**
     * @Expose
     * @Groups({"groupDetail", "userDetail"})
     *
     * @var Privilege[]|ArrayCollection
     */
    private $privileges;

    /**
     * @Expose
     * @Groups({"groupDetail"})
     *
     * @var User[]|ArrayCollection
     */
    private $users;
    // ... other properties and methods
}

 * @ExclusionPolicy("all")
 */
class Privilege implements PrivilegeInterface
{
    /**
     * @var User[]|ArrayCollection
     */
    private $users;

    /**
     * @var Group[]|ArrayCollection
     */
    private $groups;
}
```

In the controller it is important to serialize also with the Default groups.
Here it is with annotation, but serialization context can be used easily as well to build the groups dynamicly.
```php
    /**
     * @Annotations\View(serializerGroups={"Default","groupList"})
     */
    public function getGroupIndexAction()
    /**
     * @Annotations\View(serializerGroups={"Default","groupDetail"})
     */
    public function getGroupDetailAction($groupId)
    /**
     * @Annotations\View(serializerGroups={"Default","userDetail"})
     */
    public function getUserDetailAction($groupId)
```

With this setup, you'll eliminated the recursion, and have exact controll what will return where.
I'll describe it in word and json data are bellow

For Group Detail you'll get list of its privileges and users with its own special privileges if assigned.
- on the nested users you WONT get Groups again, as it would create recursion
- on the nested privileges you WONT get Groups and Users again as it would create recursion

For User Detail you'll get list of its privileges and groups with its own privileges.
- on the nested privileges you WONT get Groups and Users again as it would create recursion
- on the nested groups you WONT get Users again, as it would create recursion
- on the nested groups.privileges you WONT get Groups and Users again as it would create recursion

Groups Listing:
```json
{
  "limit": "20",
  "offset": 0,
  "total": 5,
  "pages": 1,
  "page": 1,
  "_links": {
    "self": {
      "href": "/api/v1/groups?limit=20"
    },
    "first": {
      "href": "/api/v1/groups?limit=20"
    },
    "last": {
      "href": "/api/v1/groups?limit=20"
    }
  },
  "_embedded": {
    "items": [
      {
        "id": 1,
        "title": "Admins - Read",
        "handle": "ADMIN_READ",
        "status": 1
      }      
    ]
  }
}
```

Group Detail:
```json
{
  "id": 2,
  "title": "Admins - Write",
  "handle": "ADMIN_WRITE",
  "privileges": {
    "1": {
      "id": 1,
      "title": "admin access",
      "handle": "ROLE_ADMIN_ACCESS"
    },
    "2": {
      "id": 2,
      "title": "users own read",
      "handle": "USERS_OWN_READ"
    },
  },
  "users": [
   {
      "id": 1,
      "email": "email@example.com",
      "status": 1,
      "privileges": {
         "99": {
           "id": 99,
           "title": "special for only this one user privilege",
           "handle": "SPECIAL_PRIVILEGE"
         }
      }
   }
   ]
}
```
User Resource

```json
{
  "id": 1,
  "email": "email@example.com",
  "privileges": {
    "99": { "user special privildge data" } 
  },
  "groups": {
    "2": {
      "id": 2,
      "title": "Admins - Write",
      "handle": "ADMIN_WRITE",
      "privileges": {
        "1": { "permission data" },
        "2": { "permission data" } 
      }
    }
  },
}

umpirsky:
Thanks @nakashu!

But I am just thinking...we never expose passwords or any sensitive data via API, so they will never be included. Right @pjedrzejewski?

nakashu:
@umpirsky 
Only fields that got @Expose() annotation will be serialized, as the class got @ExclusionPolicy("all"), meaning everythings is exluded by default.

Putting @Expose on a property, means you add it to the "Default" serialization group. That's why the group is there in the controller, so properties that dont have other groups (like userDetail), will be included in the output.

btw.: fixed first 3 lines of my example as the part was stripped as I used html tags there.

umpirsky:
@nakashu Sure thing, I just wanted to check with @pjedrzejewski that we never expose sensitive user data to serializer.

bendavies:
@umpirsky are you using this in your project? I might be interested in using it i mine.

umpirsky:
@bendavies Yes, I used to use it on one.

bendavies:
@umpirsky hey, thanks. is it pretty much this exact code or have you improved it?

umpirsky:
@bendavies No, that is pretty much it.

pjedrzejewski:
Closing this because it won't be in the core for now. We will have a look at this for SyliusShopApiPlugin.

Issue #2693, opened 2015-04-26
Move TranslationLocaleProvider from the CoreBundle to a component.


adamelso:
After upgrading Sylius (which I'm using Sylius like a library rather than the app), while running my Behat scenarios I was getting the exception:

> Request must be defined. (RuntimeException)

After some digging, I found I was able to fix the problem by registering the contextual TranslationLocaleProvider service:

```xml
<service
  id="sylius.translation.locale_provider.contextual"
  class="Sylius\Bundle\CoreBundle\Locale\TranslationLocaleProvider">
    <argument type="service" id="sylius.context.locale" />
</service>
```

I don't have CoreBundle registered in my Symfony app, but I still ended up registering this class, which makes me think that the class could be moved to a component, or a lower-level bundle. Question being where exactly to move it to (Core? LocaleBundle?).

aRn0D:
I think class can be moved into a component because it is not related to Symfony. I think we should update `TranslationBundle` to explain that too.

michalmarcinkowski:
Should be fixed already. /cc @pamil

PR #3778, opened 2015-12-27
[WIP] Merge Translation into Resource + some clean ups

pjedrzejewski:
Ref. #3750.

Issue #3750, opened 2015-12-18
Merge Translation & TranslationBundle into Resource?


pjedrzejewski:
Hey folks!

I would like to propose merging Translation and TranslationBundle into Resource and ResourceBundle respectively. 

Caveats:

* You will still be able to creates resources which are not translatable. Just implement TranslatableInterface or not and configure ``translation`` node or not.

Rationale:

Pros:
* Removes overhead of having 4 packages instead of 2;
* Removes overhead of decoupling configuration and its processing in ResourceBundle;
* We can use TranslatableInterface and other interfaces safely everywhere to check if something supports translations, now we need to use string and check if bundle exists etc. - really annoying;
* Translation & TranslationBundle without ResourceBundle do not make much sense, is anyone actually using them without Resource?

Cons:
* I don't see any right now, but opening this RFC to find out. :)

I know that in perfect scenario we create an extensions system for ResourceBundle and every extension could add its own configuration node to every resource and also process metadata in its own way etc. and I'd love to do that but at this point we don't have time and I really don't see a reason to have it decoupled. It only adds overhead.

If someone thinks of using Translation from Sylius standalone, there are already bundles/components which have similar level of functionality.

pjedrzejewski:
Ref. https://github.com/Sylius/Sylius/issues/3748 and #3721.

gperdomor:
:+1:

bendavies:
+1 sounds good. 
sort of related, there is a lot of duplication of the configuration tree structures at the moment. I wonder if it can be made more DRY?

PR #3940, opened 2016-01-21
[Translation] Move TranslationLocaleProvider from the CoreBundle to Translation

adamelso:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #2693, #4633
| License       | MIT
| Doc PR        | n/a

The class is not dependent on Symfony, but registering this class
as a service would be required in some projects using Sylius
translatable models.

Question: Should the service definition be moved from the CoreBundle to either the TranslationBundle or LocaleBundle, or stay in the CoreBundle?

### DIFF
--- a/src/Sylius/Component/Translation/Provider/TranslationLocaleProvider.php
+++ b/src/Sylius/Component/Translation/Provider/TranslationLocaleProvider.php
@@ -9,10 +9,9 @@
 * file that was distributed with this source code.
 */
 
-namespace Sylius\Bundle\CoreBundle\Locale;
+namespace Sylius\Component\Translation\Provider;
 
 use Sylius\Component\Locale\Context\LocaleContextInterface;
-use Sylius\Component\Translation\Provider\LocaleProviderInterface;
 
 /**
  * @author Paweł Jędrzejewski <pawel@sylius.org>
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
@@ -110,7 +110,7 @@
         <parameter key="sylius.form.type.data_fetcher.user_registration.class">Sylius\Bundle\CoreBundle\Form\Type\DataFetcher\UserRegistrationType</parameter>
 
         <parameter key="sylius.authorization_checker.toggleable.class">Sylius\Bundle\CoreBundle\Authorization\ToggleableAuthorizationChecker</parameter>
-        <parameter key="sylius.translation.locale_provider.contextual.class">Sylius\Bundle\CoreBundle\Locale\TranslationLocaleProvider</parameter>
+        <parameter key="sylius.translation.locale_provider.contextual.class">Sylius\Component\Translation\Provider\TranslationLocaleProvider</parameter>
     </parameters>
 
     <services>


michalmarcinkowski:
The TranslationBundle will be merged into ResourceBundle (see https://github.com/Sylius/Sylius/pull/3778). IMO we should move the definition to LocaleBundle.

adamelso:
@michalmarcinkowski it wouldn't be a hard dependency, much like `symfony/dependency-injection` does not depend on `yaml`, `config` or `expression-language` but its [`YamlFileLoader`](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/DependencyInjection/Loader/YamlFileLoader.php) would expect those packages to be installed if it is being used.

adamelso:
...but agreed that the service registration should remain in the CoreBundle.

src/Sylius/Component/Locale/Provider/TranslationLocaleProvider.php
```diff
@@ -9,7 +9,7 @@
 * file that was distributed with this source code.
 */
 
-namespace Sylius\Bundle\CoreBundle\Locale;
+namespace Sylius\Component\Locale\Provider;
 
 use Sylius\Component\Locale\Context\LocaleContextInterface;
 use Sylius\Component\Translation\Provider\LocaleProviderInterface;
```
michalmarcinkowski: Actually this interface is from `Translation` component that is not required for `Locale` component. We can't move this class to `Locale` component before merging Translation component into Resource component.

adamelso: Oh yeah, I see that there are two different `LocaleProviderInterface`. I've moved it to the Translation component with the intention of having it later moved into Resource. Then anyone wanting to use this class would need to install both `sylius/resource` and `sylius/locale` rather than copying it or installing the whole CoreBundle.

src/Sylius/Component/Translation/Provider/TranslationLocaleProvider.php
```diff
@@ -9,10 +9,9 @@
 * file that was distributed with this source code.
 */
 
-namespace Sylius\Bundle\CoreBundle\Locale;
+namespace Sylius\Component\Translation\Provider;
 
 use Sylius\Component\Locale\Context\LocaleContextInterface;
```
michalmarcinkowski: The `Translation` component shouldn't be coupled to `Locale` component. This class should be moved to `Core` component and after the `Translation` is merged into `Resource` component (#3750) we can move it to `Locale`.

adamelso: Okay, I'll move it to the Core component.

michalmarcinkowski:
@adamelso the `Translation` component is merged into `Resource`, so now we can move this provider to `Locale` component :)

adamelso:
@michalmarcinkowski Great! However, there is another source of confusion that would prevent me from doing so - there are two interfaces named `LocaleProviderInterface` - one in Locale and one in Resource, with different signatures.

The `Sylius\Bundle\CoreBundle\Provider\TranslationLocaleProvider` actually implements the interface in _Resource_.

If I were to mirror the namespaces, so the resulting FQCN is `Sylius\Component\Locale\Provider\TranslationLocaleProvider` it creates the false impression that it is using the interface in the same namespace.

Can we consolidate the two?

```php
namespace Sylius\Component\Locale\Provider;

interface LocaleProviderInterface
{
    public function getAvailableLocales();
    public function isLocaleAvailable($locale);
}
```

```php
namespace Sylius\Component\Resource\Provider;

interface LocaleProviderInterface
{
    public function getCurrentLocale();
    public function getFallbackLocale();
    public function getDefaultLocale();
}
```

adamelso:
Fixed in #5601.

PR #3942, opened 2016-01-21
Di fixes

bendavies:
Just some fixes to invalid services.

EDIT: 
infact, the `ODMTranslatableListener` is still broken after fixing the class name. look at the service definition. Doesn't match at all.
Should `ODMTranslatableListener` extend `AbstractTranslatableListener` like the `ORMTranslatableListener`?

EDIT2: 
it was broken in this commit: https://github.com/Sylius/Sylius/commit/6be08f847f7fea3def754b05c562989f0f06276d


bendavies:
I'm going to split this up.

Issue #3538, opened 2015-11-06
[FlowBundle] Step history lost in some cases


umpirsky:
We started to get some issues with our checkout flow, suddenly users are seeing 404 page in the middle of checkout process. This happens in ~1% of cases.

In order to debug this problem, we override `Coordinator.forward()` method:

```php
try {
    $this->context->rewindHistory();
} catch (NotFoundHttpException $e) {
    $referer = null;
    if (null !== $this->context->getRequest()) {
        $referer = $this->context->getRequest()->headers->get('referer');
    }

    $this->logger->warning(sprintf(
        '%s Referer: %s. Step history: %s.',
        $e->getMessage(),
        null === $referer ? 'N/A' : $referer,
        serialize($this->context->getStepHistory()
    )));
   // ...
}
```

Here are some logs we found:

```
[2015-11-04 21:49:19] flow.WARNING: Step "addressing" not found in step history. Referer: /checkout/addressing. Step history: a:0:{}. [] []
[2015-11-04 21:59:56] flow.WARNING: Step "delivery_and_payment" not found in step history. Referer: /checkout/delivery_and_payment. Step history: a:0:{}. [] []
[2015-11-05 00:49:45] flow.WARNING: Step "delivery_and_payment" not found in step history. Referer: /checkout/delivery_and_payment. Step history: a:0:{}. [] []
[2015-11-05 00:49:54] flow.WARNING: Step "delivery_and_payment" not found in step history. Referer: /checkout/delivery_and_payment. Step history: a:0:{}. [] []
[2015-11-05 00:50:00] flow.WARNING: Step "delivery_and_payment" not found in step history. Referer: /checkout/delivery_and_payment. Step history: a:0:{}. [] []
[2015-11-05 00:50:05] flow.WARNING: Step "delivery_and_payment" not found in step history. Referer: /checkout/delivery_and_payment. Step history: a:0:{}. [] []
[2015-11-05 01:37:53] flow.WARNING: Step "addressing" not found in step history. Referer: /checkout/addressing. Step history: a:0:{}. [] []
[2015-11-05 03:08:55] flow.WARNING: Step "addressing" not found in step history. Referer: N/A. Step history: a:0:{}. [] []
```

@pjedrzejewski Ever had issues like this? May it be session related issue?

pjedrzejewski:
Something new to me, but can be session related issue. I was thinking about saving that on the Order itself, not in the session. (as an alternative storage, sesssion would be still supported)

[...]

PR #3946, opened 2016-01-21
[RFC][FlowBundle] Use Storage component

adamelso:
| Q             | A
| ------------- | ---
| Bug fix?      | -
| New feature?  | yes
| BC breaks?    | Application: no; Bundle: yes
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

At first, I thought FlowBundle was using Storage, but it actually has its own storage classes and interfaces.
The difference appears to be that FlowBundle accesses session variables from a custom bag.

Is it worth merging the two somehow to reduce code duplication and to make it easier to add custom storage? This might also give way to providing a fix for #3538


src/Sylius/Bundle/FlowBundle/spec/Process/Context/ProcessContextSpec.php
```diff
@@ -46,7 +46,7 @@ function it_initializes(
     ) {
         $process->getScenarioAlias()->shouldBeCalled();
         $storage->initialize(Argument::type('string'))->shouldBeCalled();
-        $process->getOrderedSteps()->shouldBeCalled()->willReturn(array($previousStep, $currentStep, $nextStep));
+        $process->getOrderedSteps()->shouldBeCalled()->willReturn([$previousStep, $currentStep, $nextStep]);
```
stloyd: As you change this file anyway, please remove all calls to `shouldBeCalled()` if there are calls to the `willReturn()` like on this line.

pjedrzejewski:
Definitely makes sense.

I know that @peteward will kill me for saying this but I have this crazy idea laying around for some time already... Maybe I will illustrate it with some routing:

Rough idea of routing:

```
sylius_checkout_addresing:
    path: /addressing
    methods: [GET, PUT]
    defaults:
        _controller: sylius.controller.cart:updateAction
        _sylius:
            template: SyliusShopBundle:Checkout:addressing.html.twig
            form: sylius_checkout_addressing
            event: checkout_addressing
            repository:
                method: findOneByCustomerAndCheckoutState
                arguments: ["expr:service('sylius.context.customer').get()", 'started']
            state_machine:
                transition: address
                graph: sylius_cart_checkout
            redirect: sylius_checkout_shipping
```

Before anyone calls me crazy :D I am not entirely convinced too but using tested and simple actions instead of custom controllers gives us less points for possible bugs, with flow bundle, custom update logic etc. All is handled by our good old RC above.

Cons:
* Too much logic in routing?

Pros:
* All FlowSteps are basically this right now, updating the order using appropriate form, dispatching some events, next, that's it. In #3918 we add state machine and that perfectly fits into this scheme.
* Works for both HTML rendering and API. We would have an unified way of handling the checkout and would not need to worry about differences between api / html access that much.
* Easy to customize, probably easier than overriding whole Step class.
* Less code is better. :D
* There are other use cases for introducing state machine into updateAction, for example:

Shipping the shipment. Even if we drop this idea for checkout, I feel like it makes sense for many other actions, shipping the shipment is just an example. Shipping the shipment means executing transition "ship" on shipment and optionally setting tracking code (or whatever you want if you customize the form), so that could be perfectly realized by:

```
sylius_admin_shipment_ship:
    path: /{id}/ship
    defaults:
        _controller: sylius.controller.shipment:updateAction
        _sylius:
            template: SyliusAdminBundle:Shipment:ship.html.twig
            form: sylius_shipment_ship
            event: ship
            state_machine:
                transition: ship
```

No worries, this is not decided yet but I believe that could be used for many other transitions which need to edit something as well. Let me know! :dancer:

src/Sylius/Bundle/FlowBundle/spec/Storage/SessionStorageSpec.php
```diff
@@ -30,42 +30,55 @@ function let(SessionInterface $session, SessionFlowsBag $bag)
     function it_is_initializable()
     {
         $this->shouldHaveType('Sylius\Bundle\FlowBundle\Storage\SessionStorage');
-        $this->shouldHaveType('Sylius\Bundle\FlowBundle\Storage\AbstractStorage');
+    }
+
+    function it_is_a_Sylius_storage()
+    {
+        $this->shouldImplement('Sylius\Component\Storage\StorageInterface');
```
stloyd: `StorageInterface::class` same below.

peteward:
fix provinces first!! magic code later.

pjedrzejewski:
@peteward I have commented in the relevant issue. Checkout code is quite important as well,  that's why I am throwing an idea here.

src/Sylius/Bundle/FlowBundle/spec/Process/Context/ProcessContextSpec.php
bendavies: @stloyd does `willReturn` imply `shouldBeCalled`?

adamelso: @stloyd :+1: 

@bendavies It's part of the [Command-Query Separation principle](http://martinfowler.com/bliki/CommandQuerySeparation.html).

Basically, methods should either:

  * Do something (_Command_, `shouldBeCalled()`)
  * Return something (_Query_, `willReturn()`)

but never both!

However, for methods that don't adhere to this principle, you'll need to do `->shouldBeCalled()->willReturn()`, but this is not the case here.

bendavies: @adamelso thanks!

adamelso:
FlowBundle is currently not deprecated yet, but will be soon. Closing. (and last PR to close, the remaining PR I'll keep open :).

PR #3948, opened 2016-01-21
[WIP][RFC] Choice attribute type

umpirsky:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| License | MIT |

### DIFF
--- a/src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/ChoiceAttributeType.php
+++ b/src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/ChoiceAttributeType.php
@@ -1,0 +1,47 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\AttributeBundle\Form\Type\AttributeType;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @author Saša Stamenković <umpirsky@gmail.com>
+ */
+class ChoiceAttributeType extends AbstractType
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getParent()
+    {
+        return 'choice';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults(array(
+            'label' => false,
+        ));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return 'sylius_attribute_type_choice';
+    }
+}
--- a/src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/Configuration/ChoiceAttributeConfigurationType.php
+++ b/src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/Configuration/ChoiceAttributeConfigurationType.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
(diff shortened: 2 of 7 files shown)


src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/Configuration/ChoiceAttributeConfigurationType.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+class ChoiceAttributeConfigurationType extends AbstractType
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('options', 'collection', [
```
umpirsky: How should we render this? Should we have separate template for each configuration type?

src/Sylius/Component/Attribute/AttributeType/ChoiceAttributeType.php
```diff
@@ -0,0 +1,44 @@
...
+class ChoiceAttributeType implements AttributeTypeInterface
+{
+    const TYPE = 'choice';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getStorageType()
+    {
+        return AttributeValueInterface::STORAGE_TEXT;
```
umpirsky: Text storage cannot handle multiple selection.

pjedrzejewski: I see 2 options:

* Adding JSON storage which will be added anyway and using it here.
* Adding JSON storage but introducing 2 attribute types, single + multiple select, using text + json storages respectively for these two.

umpirsky: @pjedrzejewski If we use same attribute type for single and multiple select, they will be both stored using JSON storage? Maybe we can add configuration to `public function getStorageType(array $configuration)` and return storage type based on `$configuration['multiple']`?

Also, what is JSON storage type? Not same as doctrine array?

xantrix:
hey guys, Is the purpose of this PR to add a sort of Enum Attribute ? 
Like a Status attribute for products with values like [New, Classic, On sale] ?

umpirsky:
@xantrix Yes, something like that. Attribute that can have values from the list of choices.

xantrix:
Great @umpirsky ! Really helpful feature. It would be nice if the enum values would support translations.

src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/ChoiceAttributeType.php
```diff
@@ -0,0 +1,47 @@
...
+            'label' => false,
+        ));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return 'sylius_attribute_type_choice';
```
pjedrzejewski: Ouch, this can be a bit confusing. Maybe we could figure out a more appropriate name for this attribute type. I think Akeneo is calling that "single select" and "multi select", wdyt?

src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/Configuration/ChoiceAttributeConfigurationType.php
pjedrzejewski: I'd suggest form theme. Hm?

src/Sylius/Component/Attribute/AttributeType/ChoiceAttributeType.php
pjedrzejewski: Array uses serialize/unserialize, while json uses json_encode/decode. I'd go for JSON. You get an array in PHP anyway with both of these though.

src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/ChoiceAttributeType.php
umpirsky: We follow Symfony terminology. :)

PR #3954, opened 2016-01-22
Order item api form

bendavies:
The service definition for `sylius.form.type.api.order_item` was incorrect.


pjedrzejewski:
Thank you Ben!

PR #3961, opened 2016-01-23
Introduce CodeAwareTrait into Resource component

stloyd:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| License       | MIT


src/Sylius/Component/Addressing/Model/ZoneMember.php
```diff
@@ -10,6 +10,7 @@
  */
 
 namespace Sylius\Component\Addressing\Model;
+use Sylius\Component\Resource\Model\CodeAwareTrait;
```
adamelso: PSR-2: Missing blank line between `namespace` and `use`.

src/Sylius/Component/Archetype/Model/Archetype.php
```diff
@@ -29,7 +30,7 @@
  */
 class Archetype extends AbstractTranslatable implements ArchetypeInterface
 {
-    use TimestampableTrait;
+    use CodeAwareTrait, TimestampableTrait;
```
adamelso: Nothing specific about importing traits is mentioned in PSR-2, but if there many to import in the future (hopefully not) it might be worth declaring one per statement, [like when importing classes from other namespaces](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#3-namespace-and-use-declarations).

pjedrzejewski:
I am against adding <field>AwareTraits because it is only hiding complexity and we could do it for a lot of fields, like name, slug, description etc. The thing is that trait should be about some specific behavior, timestampable, toggleable etc. Having a code is not a behavior really. That being said, I will be happy to have a discussion here, as always.

Please avoid doing changes not related to the PR, it is really hard to review. This PR is about code aware trait, not reordering use statements and 5 other small tweaks. I know it's tempting to fix small stuff when going over the files but this is not good practice.

michalmarcinkowski:
@stloyd we will be happy to merge "optimize imports" PR if you could extract it to a separate one.

Do you want to discuss AwareTrait concept or this PR can be closed?

stloyd:
Let's ignore that "optimize" commit as it should not land here (my fault by forgetting to switch branches).

About trait's that are "*AwareOfSomething", IMO when it comes to components, it makes developing much easier, i.e. `OriginAwareTrait`, it same approach as Symfony has for `ContainerAwareTrait` (talking about approach not behavior of related code).

Skiping that `CodeAwareTrait` is bad name, same as all those `CodeAwareInterface` (IIRC I told that name is wrong ;)) cause it's not really a "code", it's "internal name" or more correct "internal identifier"...

pjedrzejewski:
I think we will stick with code and eventually rename it in future versions, although I still believe it is most appropriate name. Thanks for the suggestion about traits but as discussed - we don't want to use them like this.

PR #3970, opened 2016-01-25
[UserBundle] Fix default username orm listener

michalmarcinkowski:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3939
| License       | MIT
| Doc PR        |

PR #3972, opened 2016-01-26
Revert "[UserBundle] Fix default username orm listener"

pjedrzejewski:
Seems like PR was green but only Scrutinizer passed and Travis was not run at all and I merged this PR at ~1AM today, oups. :) I will make Travis required in GitHub configuration which is possible now since some time already.

Reverts Sylius/Sylius#3970.


michalmarcinkowski:
Look like we have a bug in our test suite and it uses cached entity.

PR #3978, opened 2016-01-26
[Behat] Fixed sorting orders by customers' last names scenario

pamil:
###### (By removing it temporarily)

It causes false negatives on Travis (and it does not even work correctly :)).


pjedrzejewski:
:+1:

pamil:
In fact we didn't even click on the right link to sort the resources, had the last names autogenerated by faker and was sorting by email, everything that could gone wrong had already gone wrong :)

PR #3987, opened 2016-01-26
[Storage] Get the request from the request stack in CookieStorage.

adamelso:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no for application, yes for component
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        | https://github.com/Sylius/Sylius-Docs/pull/402

In order to read and write from cookies, the Request object needs to be
accessible from the Storage implementation, but instead of using the
deprecated Symfony `request` service in a setter injection, the
CookieStorage is constructed with a RequestStack and the cookie data
accessed from the master request of the RequestStack.

I take it this service was created when Sylius was still using the
Symfony 2.3 LTS which didn't have the request stack.

### DIFF
--- a/src/Sylius/Component/Storage/CookieStorage.php
+++ b/src/Sylius/Component/Storage/CookieStorage.php
@@ -11,24 +11,25 @@
 
 namespace Sylius\Component\Storage;
 
-use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
 
 /**
  * @author Joseph Bielawski <stloyd@gmail.com>
+ * @author Adam Elsodaney <adam.elso@gmail.com>
  */
 class CookieStorage implements StorageInterface
 {
     /**
-     * @var Request
+     * @var RequestStack
      */
-    protected $request;
+    protected $requestStack;
 
     /**
-     * @param Request $request
+     * @param RequestStack $requestStack
      */
-    public function setRequest($request)
+    public function __construct(RequestStack $requestStack)
     {
-        $this->request = $request;
+        $this->requestStack = $requestStack;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Sylius/Component/Storage/CookieStorage.php
```diff
@@ -11,24 +11,25 @@
...
+ * @author Adam Elsodaney <adam.elso@gmail.com>
  */
 class CookieStorage implements StorageInterface
 {
     /**
-     * @var Request
+     * @var RequestStack
      */
-    protected $request;
+    private $requestStack;
```
adamelso: Request stack should be protected instead of private.

pamil:
@adamelso are you sure it really sets the cookie? :) IMHO the new cookies should be set on the Response object, Request is rather read-only and I'd be surprised if the cookies set on Request appears on Response in some magical way.

adamelso:
@pamil Good point but I didn't want to change how it originally behaved. Unless of course it wasn't working to begin with - this is registered as a service in the ResourceBundle but not injected into any other services in Sylius.

I can see it working either between kernel events during the lifecycle of the request but not in between requests, or if there is listener that reads the new values and adds the cookies to the response headers, but this means determining which ones shouldn't be sent again.

pjedrzejewski:
The cookie has to be set on Response. So this should probably collect data during the request lifecycle and listen to kernel.response event and set cookies on the response then. Otherwise it won't work. :(

adamelso:
I realise this has been covered already in #3437 - might be worth combining these PRs.

adamelso:
CookieStorage won't work without some sort of event listener which will make things complex. Closing :)

PR #3991, opened 2016-01-27
[Cart] - itemresolver return correct CartItem from resolve() for items already present in cart.

nakashu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

When you tried to resolve item that is already in the Cart, ItemResolver returned the back the empty CartItem supplied to the resolve() method instead of the correct CartItem from Cart with updated quantity.


nakashu:
Ok the solution here is bad, but I'll like opitions if the current behaviour of itemResolver is bad or not.

What is expected behaviour of ItemResolver? What should it return?

1. new CartItem object (with data from request), without ID, even if it is in cart. (current behaviour)
2. existing CartItem object from cart if it is present, and new one if it is missing.

This fix doesn't work because of the logic inside CartSubscriber, where the quantities from both items are added together inside the cart.
If current behavior is correct, fell free to close this.

Issue #3685, opened 2015-12-06
Using offline payment throws exception


torinaki:
Faced with following exception on order submit:
```
Gateway "offline" does not exist.
```
Seems new version of Payum bundle do not have such gateway configured:
```
$ php ./app/console debug:config payum
payum:
    gateways:
        paypal_express_checkout: { paypal_express_checkout_nvp: { username: db-facilitator_api1.zb.lv, password: 1386079449, signature: AA.EdtBRYFrtdnjw9s-2azOre0X4AYbckaCSL.pubIbuKwTqOX0owTYi, sandbox: true, actions: {  }, apis: {  }, extensions: {  } } }
        be2bill_direct: { be2bill_direct: { identifier: EDITME, password: EDITME, sandbox: true, actions: {  }, apis: {  }, extensions: {  } } }
        be2bill_offsite: { be2bill_offsite: { identifier: EDITME, password: EDITME, sandbox: true, actions: {  }, apis: {  }, extensions: {  } } }
        stripe_checkout: { stripe_checkout: { publishable_key: EDITME, secret_key: EDITME, actions: {  }, apis: {  }, extensions: {  } } }
        dummy: { offline: { actions: {  }, apis: {  }, extensions: {  } } }

```

Sp0ng3b0b:
payum:
    gateways:
        offline:
            offline: ~

Add this to your configuration, works for me :+1:

michalmarcinkowski:
If it didn't work for you please post this question on StackOverflow - it is the official support channel for Sylius. Just tag your question with "sylius" and "payum", maybe someone will be able to help you! http://stackoverflow.com/questions/tagged/sylius

torinaki:
@Sp0ng3b0b thanks you. I will try. More than sure it will work.

@michalmarcinkowski It is not a question and even not Support Request. Currenlty `offline` payment type do not work in default Sylius setup, because of incomplite payum setup(I guess it happen after Payum bundle update till 1.0 version). I have faced with this issue right after Sylius update.
The only question is what should be fixed to solve this issue:
1. Remove offline payment type from Sylius setup
2. Payum configuration should be adjusted as @Sp0ng3b0b wrote.

// cc @pjedrzejewski

pjedrzejewski:
@torinaki Your solution sounds good, are you willing to work on a patch?

torinaki:
Yes, sure. I will prepare PR to adjust default configuration.
[...]

PR #3995, opened 2016-01-27
[Payment] Replace dummy payment gateway with offline

michalmarcinkowski:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3685
| License       | MIT
| Doc PR        | 

Replaced `dummy` gateway with `offline` as it always succeed. Probably later we can think of introducing requirement to approve the offline payment manually by admin and then we can introduce some `dummy` gateway just for tests.


pjedrzejewski:
:+1:

Issue #3988, opened 2016-01-26
[RFC] - CartItemController more flexible redirects/responses


nakashu:
I am currently making some changes in `CartItemController` for a project , that would probably benefit also for the core, and need initial comments for the approach.

#### Problem

Currently there are `addAction` and `RemoveAction` where 95% of time you don't need to change the resolving login / flash handling, as you can override the Resolver itself if needed or just hook to the dispatched events.

The limitation is with the `redirectAfterAdd / redirectToCartSummary` methods:

* `redirectAfterAdd()` is private, so you cannot overriding its login inside your controller, and are left with copy paste 
* `removeAction` calls only redirectToSummary.
* inside the redirect methods you are missing the whole context, namely the `cartItem` that was just added / removed, this limits uses where you need to dynamically redirect to page with current `cartItem `only

#### Usecases:

* you want to handle additional query parameters (passed to the route) on which you'll decide where to redirect. Although it is possible to do now, if you overrid the `redirectToSummary` it seems dirty if in the end you will go to homepage for instance based on the parameters.
* you want do dynamically generate route with current cartItem ID or other data. For instance you add a laptop to cart and want to show page with adjustments/accessories like extended warranty. Here you can buy the same product/variant twice but with different accessories.
* ajax handling of adding and removing, where you need different response for ajax requests.

#### Solution:

* make 2 methods - responseAfterAdd / responseAfterRemove - both protected
* pass second parameter `$cartItem` to these

With this you can handle all above scenarios, just by overriding those two methods instead of whole actions, without any BC.

#### Sample
```php
protected function responseAfterAdd (RequestConfiguration $configuration, CartItemInterface $item = null)
    {
        $request = $configuration->getRequest();

        if ($request->query->has('_redirect_to')) {
            return $this->redirectHandler->redirect($configuration, $request->query->get('_redirect_to'));
        }

       // if you override this you can do what you please like
       if ($request->isXmlHttpRequest())
            return new JsonResponse(['stuff' => 'ajax_stuff']);

        return $this->redirectToCartSummary($configuration);
    }
```

I'll make the PR, if this looks sensible.

Its also partialy related to this issue #2627 concerning the redirect to summary page.

adamelso:
:+1: Came across this issue today; would also like to use it in an AJAX request without duplicating the code too much :)

nakashu:
Just found another issue with cartItemResolver / CartSubscriber

[...]

Issue #2627, opened 2015-03-27
[Enhancement][WebBundle] Add to cart AJAX + mini cart summary


axelvnk:
Hi everyone,

For a clients project I felt the need to use AJAX to add products to the cart without having to redirect to the summary everytime. I also made a smaller version of the cart summary, which also updates after every click on an add to cart button. 

This might be a nice addition to the WebBundle and I would love to contribute my code to the project. Here is a short video of the functionality I created. (Still rough around the edges though...)

https://www.youtube.com/watch?v=YEFh1sshdLg

If there is interest in this functionality, I will refactor this and put it inside a Sylius fork, so it's usable for any project.

pjedrzejewski:
Looks really cool and that's the goal for new theme. :) Are you using default cart actions or something more is needed for this to work?

axelvnk:
@pjedrzejewski Currently the default cart actions always redirect to the cart summary page. So I extended the CartController and CartItemController and added practically the same actions but made some changes to have json responses. Also leaving out the redirect to summary and eliminating the flash messages. Otherwise they would be all printed out when the user visits another page.

axelvnk:
Maybe we can first start by adding the ajax friendly actions to both controllers, or work with a boolean in the current actions, like so.

``````
public function removeAction($id, $ajax = null)
{
    if ($ajax) {
        //do something
     }
}

```````
This would not affect current functionality, but only expands the possibilities.

amenophis:
Maybe use API actions that return JSON or XML ?

axelvnk:
@amenophis Not sure if that's necessary, because the only info that is needed from the ajax call is wether or not the data processing was successful (imo). I notify the user when something went wrong and when the action was successful. 

I also make an ajax call to request a rendered template of the mini cart summary after an item is added or removed from the cart. That's how my mini cart stays up to date. I thought this was the easiest way since the grand total needs to get updated as well.

amenophis:
yes, the API could respond 2xx in case of success, or 4xx/5xx in case of error.
Thinks it can be better to implement this functionnality like this.
It would be usefull if you want to build entirely a front with a Javascript framework like angularjs or other.
All server interactions would work in the same way (getProducts, getProduct, addToCart, getCart, ...)

axelvnk:
@amenophis you're right. Would make more sense :)

nicolasricci:
[...]

PR #3996, opened 2016-01-27
[Cart] - CartItemController more flexible responses

nakashu:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| Enhancment? | yes - DX
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #3988, #2627(partialy)
| License       | MIT

Response methods after add/remove action are protected  instead of private to allow overriding.
Additionally second argument $context is passed with additional data about added/remove item.

This allows for easier ajax responses, without code duplication, as well as dynamic redirect/response generation.

`getEqualCartItemFromCart` method is a workaround for this issue #3991


okwinza:
:+1:

nakashu:
bump

michalmarcinkowski:
Closing as you can now configure redirect in route definition using standard ResourceBundle option (see [here)](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ShopBundle/Resources/config/routing/cart.yml#L35).

PR #3784, opened 2015-12-28
Code style

bendavies:
Is there any reason why CS isn't enforced any more?
Consistency is good!

I've enabled caching. should work, provided that travis retains file modification times on checkout?
It also runs only under php7 to speed things up.

PR #4019, opened 2016-01-29
Code Style: no phpdoc_no_empty_return

bendavies:
Since we use `-empty_return` we should use `-phpdoc_no_empty_return` too.

see https://github.com/Sylius/Sylius/pull/3784#issuecomment-176914186


michalmarcinkowski:
Thanks Ben!

pjedrzejewski:
:+1:

PR #3802, opened 2016-01-04
[Report] Fixed year-change issue with reports

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Sylius/Sylius/issues/3800
| License       | MIT
| Doc PR        |

To prevent this issue occurrence after every year change (:smile:), I've set default date on ``UserRegistrationDataFetcher`` to "now" and given possibility to choose year even 100 years before current date.

Improvement:
- it would be good to change date selecting steps in scenarios, as ``And I select "2010" from "sylius_report_dataFetcherConfiguration_start_year"`` is really not easy to read :)

PR #4026, opened 2016-02-01
[Report] Quick fix after month change

Zales0123:
In https://github.com/Sylius/Sylius/pull/3802 it has been fixed issue with reports scenarios after year change. Sadly, setting default date as current date occurred in different bug, after *month* change :)
Now each day, month and year are set manually in scenario to prevent such issues in the future.


michalmarcinkowski:
Thanks Mateusz!

PR #4029, opened 2016-02-01
[Rbac] Remove softdelete interface and trait from Permission entity

michalmarcinkowski:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

Softdelete is not necessary here, what is more it can introduce unwanted behavior (after removing the permission you can't create permission with the same code). In addition the `deletedAt` field was not even included in the doctrine mapping.


pjedrzejewski:
I think this was accidentally merged in #3264...

pjedrzejewski:
Thank you Michal!

PR #4043, opened 2016-02-03
collect some image info such as file size, file mime,original file name

videni:


src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/Image.orm.xml
```diff
@@ -17,6 +17,13 @@
         <field name="updatedAt" column="updated_at" type="datetime" nullable="true">
             <gedmo:timestampable on="update"/>
         </field>
+
+        <field name="fileName" column="file_name"  length="255"  />
+
+        <field name="fileMime" column="file_mime" type="string" length="20"/>
```
stloyd: IIRC mime can be longer than just 20 characters.

videni: thanks for your remind,  the main reason caused the failed build is that  the Taxon class extends  the Image class , I think , there should be a central table to place image,  if we  want a entity to has a image  , we just add a database mapping,  this is simple  than extending the Image class.

PR #4035, opened 2016-02-02
[Cart][Behat] Products with different rates for different zones

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

Based on https://github.com/Sylius/Sylius/pull/4032 and everything before. Please, hold on with review until it will be reasonable (for the psychical health of all of us :smile:)

PR #4044, opened 2016-02-03
[Cart][Behat] Logging customer's cart

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

Based on https://github.com/Sylius/Sylius/pull/4035. Again, please, don't review this until previous PR's will be merged.

### DIFF
--- a/etc/behat/suites/ui-cart.yml
+++ b/etc/behat/suites/ui-cart.yml
@@ -68,6 +68,6 @@ default:
                     taxCategoryRepository: "@sylius.repository.tax_category"
                     zoneRepository: "@sylius.repository.zone"
 
-                - Sylius\Behat\Context\Ui\UserContext
+                - Sylius\Behat\Context\Ui\UserContext: ~
             filters:
                 tags: "@ui-cart"
--- a/etc/behat/suites/ui-cart.yml
+++ b/etc/behat/suites/ui-cart.yml
@@ -68,5 +68,6 @@ default:
                     taxCategoryRepository: "@sylius.repository.tax_category"
                     zoneRepository: "@sylius.repository.zone"
 
+                - Sylius\Behat\Context\Ui\UserContext
             filters:
                 tags: "@ui-cart"
--- a/features/shop/apply_correct_taxes_based_on_customer_data.feature
+++ b/features/shop/apply_correct_taxes_based_on_customer_data.feature
@@ -1,0 +1,33 @@
+@ui-cart
+Feature: Apply correct taxes based on customer data
+    In order to pay proper amount when buying goods
+    As a Customer
+    I want to have correct taxes applied to my order
+
+    Background:
+        Given the store is operating on a single channel
+        And there is "EU" zone containing all members of European Union
+        And there is rest of the world zone containing all other countries
+        And default currency is "EUR"
+        And default tax zone is "EU"
+        And store ships to "France" and "Australia"
+        And there is user "john@example.com" identified by "password123", with "Australia" as shipping country
+        And store has "EU VAT" tax rate of 23% for "Clothes" within "EU" zone
+        And store has "EU VAT" tax rate of 10% for "Clothes" for the rest of the world
+        And store has a product "PHP T-Shirt" priced at "€100.00"
+        And product "PHP T-Shirt" belongs to "Clothes" tax category
+        And store ships everything for free
+        And store allows paying offline
+
+    Scenario: Proper taxes for taxed product
+        When I add product "PHP T-Shirt" to the cart
+        And I proceed logging as "john@example.com" with "password123" password
+        Then my cart total should be "€110.00"
+        And my cart taxes should be "€10.00"
+
+    Scenario: Proper taxes after specifying shipping address
+        When I add product "PHP T-Shirt" to the cart
+        And I log in as "john@example.com" with "password123" password
+        And I proceed without selecting shipping address
+        Then my cart total should be "€110.00"
+        And my cart taxes should be "€10.00"
--- a/features/shop/apply_correct_taxes_based_on_customer_data.feature
+++ b/features/shop/apply_correct_taxes_based_on_customer_data.feature
@@ -6,18 +6,18 @@ Feature: Apply correct taxes based on customer data
 
     Background:
         Given the store is operating on a single channel
+        And store ships to "France" and "Australia"
         And there is "EU" zone containing all members of European Union
         And there is rest of the world zone containing all other countries
         And default currency is "EUR"
         And default tax zone is "EU"
-        And store ships to "France" and "Australia"
-        And there is user "john@example.com" identified by "password123", with "Australia" as shipping country
         And store has "EU VAT" tax rate of 23% for "Clothes" within "EU" zone
         And store has "EU VAT" tax rate of 10% for "Clothes" for the rest of the world
         And store has a product "PHP T-Shirt" priced at "€100.00"
         And product "PHP T-Shirt" belongs to "Clothes" tax category
         And store ships everything for free
         And store allows paying offline
+        And there is user "john@example.com" identified by "password123", with "Australia" as shipping country
 
     Scenario: Proper taxes for taxed product
         When I add product "PHP T-Shirt" to the cart
--- a/src/Sylius/Behat/Context/Setup/TaxContext.php
+++ b/src/Sylius/Behat/Context/Setup/TaxContext.php
@@ -92,7 +92,7 @@ final class TaxContext implements Context
 
         $taxRate = $this->taxRateFactory->createNew();
         $taxRate->setName($taxRateName);
-        $taxRate->setCode($this->getCodeFromName($taxRateName));
+        $taxRate->setCode($this->getCodeFromNameAndZoneCode($taxRateName, $zone->getCode()));
         $taxRate->setZone($zone);
         $taxRate->setAmount($this->getAmountFromString($taxRateAmount));
         $taxRate->setCategory($taxCategory);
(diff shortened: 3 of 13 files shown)


michalmarcinkowski:
@Zales0123 please rebase.

Zales0123:
@michalmarcinkowski done ;) And awaiting review, of course.

etc/behat/suites/ui-cart.yml
```diff
@@ -68,5 +68,6 @@ default:
                     taxCategoryRepository: "@sylius.repository.tax_category"
                     zoneRepository: "@sylius.repository.zone"
 
+                - Sylius\Behat\Context\Ui\UserContext
```
pamil: Let's put `: ~` at the end to be sure that context class will be interpreted as a key.

Zales0123: :+1:

src/Sylius/Behat/Context/Setup/TaxContext.php
```diff
@@ -146,8 +146,12 @@ private function getAmountFromString($taxRateAmount)
...
      */
-    private function getCodeFromName($taxRateName)
+    private function getCodeFromNameAndZoneCode($taxRateName, $zoneCode = null)
     {
-        return str_replace(' ', '_', strtolower($taxRateName));
+        if (null === $zoneCode) {
+            return str_replace(' ', '_', strtolower($taxRateName));
+        }
+
+        return str_replace(' ', '_', strtolower($taxRateName)).'_'.$zoneCode;
```
stloyd:
```php
$code = str_replace(' ', '_', strtolower($taxRateName));
if (null !== $zoneCode) {
    $code .= '_'.$zoneCode;
}

return $code;

File: src/Sylius/Behat/Context/Setup/UserContext.php [hunk 3]
```diff
@@ -130,6 +130,7 @@ private function createUser(
         $user->setUsername($email);
         $user->setEmail($email);
         $user->setPlainPassword($password);
+        $user->enable();
```
stloyd: IMO it could be useful to make this optional (sometimes you may want to test not enabled users that try to do something).

src/Sylius/Behat/Page/Checkout/CheckoutSecurityStep.php
```diff
@@ -0,0 +1,45 @@
...
+    public function getRouteName()
+    {
+        return 'sylius_checkout_security';
+    }
+
+    /**
+     * @param string $login
+     * @param string $password
+     */
+    public function logInExistingUser($login, $password)
```
pamil: We don't need an existing user to fill these fields and press a button :)

pamil: Oh, I see, there's some weird selectors, but there shouldn't be any option to log in as unexisting user.

Zales0123: However we're logging existing user/customer with passed credentials in this step ;)

src/Sylius/Behat/Context/Setup/TaxContext.php
```diff
@@ -146,8 +146,12 @@ private function getAmountFromString($taxRateAmount)
      *
      * @return string
      */
-    private function getCodeFromName($taxRateName)
+    private function getCodeFromNameAndZoneCode($taxRateName, $zoneCode = null)
```
pamil: Which code is it?

pamil: Anyway, can we make two methods instead, `getCodeFromName` and `getCodeFromNameAndZoneCode` basing on the previous method?

src/Sylius/Behat/Page/Checkout/CheckoutSecurityStep.php
Zales0123: **weird**? right in the feels :(
I tried to keep sylius security page naming:
zrzut ekranu 2016-02-09 o 15 48 14
So we can log in as guest, log in as existing customer or create new account... I guess method can be named ``logInExistingCustomer`` if you mean that, if not, I must disagree :smile:

src/Sylius/Behat/Context/Setup/TaxContext.php
Zales0123: :+1:

src/Sylius/Behat/Page/Checkout/CheckoutSecurityStep.php
pamil: Ok, that seems a little bit more complicated than I thought :D Anyway, we can rename it to `logInAsExistingUser` to be more semantically correct.

Zales0123: :+1: for ``logInAsExistingUser``

src/Sylius/Behat/Context/Setup/UserContext.php
michalmarcinkowski: I would say that usually we would like to have enabled user account as default. If we need to make it disable we can add a separate step for that `And its account is disabled`.

features/shop/apply_correct_taxes_based_on_customer_data.feature
```diff
@@ -0,0 +1,33 @@
...
+    As a Customer
+    I want to have correct taxes applied to my order
+
+    Background:
+        Given the store is operating on a single channel
+        And there is "EU" zone containing all members of European Union
+        And there is rest of the world zone containing all other countries
+        And default currency is "EUR"
+        And default tax zone is "EU"
+        And store ships to "France" and "Australia"
```
michalmarcinkowski: IMO these 3 steps above should be before zones.

features/shop/apply_correct_taxes_based_on_customer_data.feature
```diff
@@ -0,0 +1,33 @@
...
+        And there is user "john@example.com" identified by "password123", with "Australia" as shipping country
```
michalmarcinkowski: Make it the last step

features/shop/apply_correct_taxes_based_on_customer_data.feature
```diff
@@ -0,0 +1,33 @@
...
+        And there is rest of the world zone containing all other countries
+        And default currency is "EUR"
+        And default tax zone is "EU"
+        And store has "EU VAT" tax rate of 23% for "Clothes" within "EU" zone
+        And store has "EU VAT" tax rate of 10% for "Clothes" for the rest of the world
+        And store has a product "PHP T-Shirt" priced at "€100.00"
+        And product "PHP T-Shirt" belongs to "Clothes" tax category
+        And store ships everything for free
+        And store allows paying offline
+        And there is user "john@example.com" identified by "password123", with "Australia" as shipping country
```
pjedrzejewski: Maybe we can move this to separate step:

```
And user/customer "john@example.com" has default shipping address to "Australia"
```

pjedrzejewski:
I did not review everything yet because I'd like to start a quick discussion:

1. I think we are missing one scenario here, when someone is logged in before adding to the cart. The reason is we want to verify that cart is recalculated when someone signs in and has address, but we also want to verify that it is considered at the time of adding to the cart.

2. More natural scenario (short version), maybe that's even separate feature:

```
Background:
    I am logged in customer
    And my default shipping address is to "Australia"

Scenario: Adding item to the cart and getting proper tax information
    When I add product ...
    Then ...
```

E-Mail and password are irrelevant when we are logged in already. What do you think? I know this is something we can add in separate PR but I wanted to start discussion already.

Btw. something I just realized, from scenarios it is perfectly clear what is going on and they are not coupled to UI just like we wanted, but from UI perspective we are checking the "Cart Summary View" right now. We need to remember there is "Checkout Summary View" at the end and we should ensure that correct amounts are displayed there as well. Something to think about as we progress.

michalmarcinkowski:
1. :+1: 
2. IMO this is the same feature, but we could probably add a concept of default user? E.g. in this scenario I don't care about it's email or password, so I would like to read that there is a customer that is logged in. WDYT?

For checkout the "Checkout Summary View" I would say we could have a separate scenarios for verifying that "Checkout Summary View"  displays the same as "Cart Summary View". This way we will not need to check it on every single scenario.

Zales0123:
1. What do you mean by "missing scenario, when someone is logged in before adding to cart"? Wasn't all previous scenarios about already logged in customer? :smile: For example [in this feature](https://github.com/Sylius/Sylius/blob/master/features/shop/apply_correct_tax_for_items_with_the_same_tax_rate.feature#L12) we are checking proper taxes calculating in order for logged in "john@example.com" from "United Kingdom" :)
2. :+1: for default customer, for cases like this one.

michalmarcinkowski:
@Zales0123 

1. I would say we should have this specific scenario (defined shipping address) here. In the feature you linked to there is no need to specify the customer address, it should be removed.

Zales0123:
@michalmarcinkowski Agreed :+1:

michalmarcinkowski:
@Zales0123 

[These two steps](https://github.com/Sylius/Sylius/blob/master/features/shop/apply_correct_tax_for_items_with_the_same_tax_rate.feature#L18-L19) are also irrelevant since we do not even start a checkout there.

Zales0123:
As we can see, first feature must have some childhood disease :smile: 
OK, so *suma summarum*:
- create ``And I am loged in customer`` step with default customer
- separate step ``And my default shipping address is to "Australia"`` from ``And I am logged in customer``
- move here step for logged in user with already defined shipping country
- remove redundant steps from [here](https://github.com/Sylius/Sylius/blob/master/features/shop/apply_correct_tax_for_items_with_the_same_tax_rate.feature)

Am I missing sth?

michalmarcinkowski:
@Zales0123 :+1: 

But let's merge this and do it all in a separate PRs. /cc @pjedrzejewski

pjedrzejewski:
Agreed about separate PR. :+1: Thank you Mateusz!

Issue #4049, opened 2016-02-03
Error with bin/behat


vikey89:
Hi guys,
I'm on the master branch synced with the last commit and with Symfony versions 2.8.2. 
When I launch the command `bin/behat`, I get this error:

````
PHP Fatal error:  Call to private appTestDebugProjectContainer::__clone() from context 'Sylius\Behat\PageObjectExtension\ServiceContainer\PageObjectExtension' in /var/www/Sylius/src/Sylius/Behat/PageObjectExtension/ServiceContainer/PageObjectExtension.php on line 99

PHP Stack trace:
PHP   1. {main}() /var/www/Sylius/vendor/behat/behat/bin/behat:0
PHP   2. Symfony\Component\Console\Application->run() /var/www/Sylius/vendor/behat/behat/bin/behat:31
PHP   3. Behat\Testwork\Cli\Application->doRun() /var/www/Sylius/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:123
PHP   4. Behat\Testwork\Cli\Application->createCommand() /var/www/Sylius/vendor/behat/behat/src/Behat/Testwork/Cli/Application.php:96
PHP   5. Behat\Testwork\Cli\Application->createContainer() /var/www/Sylius/vendor/behat/behat/src/Behat/Testwork/Cli/Application.php:138
PHP   6. Symfony\Component\DependencyInjection\ContainerBuilder->compile() /var/www/Sylius/vendor/behat/behat/src/Behat/Testwork/Cli/Application.php:164
PHP   7. Symfony\Component\DependencyInjection\Compiler\Compiler->compile() /var/www/Sylius/vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/ContainerBuilder.php:589
PHP   8. Sylius\Behat\PageObjectExtension\ServiceContainer\PageObjectExtension->process() /var/www/Sylius/vendor/symfony/symfony/src/Symfony/Component/DependencyInjection/Compiler/Compiler.php:107
PHP   9. Sylius\Behat\PageObjectExtension\ServiceContainer\PageObjectExtension->extractSymfonyApplicationRouter() /var/www/Sylius/src/Sylius/Behat/PageObjectExtension/ServiceContainer/PageObjectExtension.php:63
[...]

PR #4053, opened 2016-02-04
[Sylius/Behat] Refactoring Behat extensions and pages

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #4049
| License       | MIT
| Doc PR        | -

PageObject extension was not defensive enough, so I've done some changes to it in our extension to an extension:

  - Default `Page` class no longer extends Mink's `DocumentElement`, it has protected `Page::getDocument()` method to be used by pages only
  - There's no `Page::getPage($page)` method
  - Removed `FeatureContext`, pages should be automatically injected to the constructor

Most of our contexts should implement empty `Context` interface.


src/Sylius/Behat/.gitignore
```diff
@@ -0,0 +1 @@
+/phpspec.yml
```
pjedrzejewski: We never prefix files with / here. Is this gitignore needed here? I guess you should call bin/phpspec from the root dir.

src/Sylius/Behat/Context/Ui/CartContext.php
```diff
@@ -11,48 +11,101 @@
...
+     */
+    public function myCartTaxesShouldBe($taxTotal)
+    {
+        $this->cartSummaryPage->open();
+
+        expect($this->cartSummaryPage->getTaxTotal())->toBe($taxTotal);
+    }
+
+    /**
+     * @Given my cart shipping fee should be :shippingTotal
```
pjedrzejewski: Then (know it is mostly moving steps around, but let's fix it in this PR)

src/Sylius/Behat/Context/Ui/CartContext.php
```diff
@@ -11,48 +11,101 @@
...
+     */
+    public function myCartShippingFeeShouldBe($shippingTotal)
+    {
+        expect($this->cartSummaryPage->getShippingTotal())->toBe($shippingTotal);
```
pjedrzejewski: Me likes :D

src/Sylius/Behat/.gitignore
pamil: This way I can call `../../../bin/phpspec run` from that directory and have only that specifications runned, what results in a faster feedback cycle.

IMO ignoring absolute paths is better than any file named `phpspec.yml`, I only want to ignore `src/Sylius/Behat/phpspec.yml` but not `src/Sylius/Behat/foo/bar/phpspec.yml`.

src/Sylius/Behat/Context/Ui/CartContext.php
pamil: :+1:

src/Sylius/Behat/.gitignore
pjedrzejewski: bin/phpspec run src/Sylius/Behat/spec?

src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -74,6 +74,7 @@ public function __construct(
     /**
      * @Transform /^product "([^"]+)"$/
      * @Transform /^"([^"]+)" product$/
+     * @Transform :product
```
michalmarcinkowski: `:productName`?

src/Sylius/Behat/Page/Product/ProductShowPage.php
```diff
@@ -20,13 +19,13 @@
 class ProductShowPage extends SymfonyPage
 {
     /**
-     * @param ProductInterface $product
+     * @param array $urlParameters
      *
      * @return ProductShowPage
      */
-    public function openSpecificProductPage(ProductInterface $product)
+    public function open(array $urlParameters = [])
```
michalmarcinkowski: What was wrong with previous signature?

src/Sylius/Behat/PageObjectExtension/PageObject/Page.php
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    /**
+     * @param string $statusCode
+     *
+     * @return bool
+     */
+    protected function isErrorResponse($statusCode)
+    {
+        return in_array(substr($statusCode, 0, 1), ['4', '5'], true);
```
michalmarcinkowski: The `$statusCode` is `int` so we can check `400 <= $statusCode && $statusCode < 600`

src/Sylius/Behat/Context/Setup/ProductContext.php
pamil: We should use `:productName` in step definitions if we want to get product name. E.g.
```gherkin
I create :productName # I want to get product name as a string
I add :product to the cart # I want to get product entity
```

src/Sylius/Behat/Page/Product/ProductShowPage.php
pamil: The `open()` wasn't opening a page and it was the only exception from the rule. If we manage to specify pages as a services, every page will be able to define their own opening methods.

src/Sylius/Behat/Context/Setup/ProductContext.php
michalmarcinkowski: :+1:

pjedrzejewski:
Awesome work, thanks Kamil!

Issue #3929, opened 2016-01-20
[RFC] Resolving channels, selecting currencies, locales and exceptions


pjedrzejewski:
Helloes!

Current channel resolving logic is wrong and same goes for selecting currency, locale etc.
We are simplifying the installer in #3814 so there is no channels, countries, locales etc. by default. So unless you load fixtures you will have a raw Sylius install with only base currency defined.

Overview of current state:

* Look for channel with LIKE hostname, which is just terrible solution implemented by me; (sorry! :D)
* If not found, get the default channel which means first channel found in repository;
* If there are no channels -> throw doctrine exception and 500 error.

Expected behavior:

* Find a channel by hostname. (for now, later I'd like to have ability to attach custom resolvers and decide on different factors than hostname, could be even locale or country of shipping)
* If not found, throw UnresolvedChannelException;
* Handle it via kernel.exception listener and display a 404 for customers and nice information page for admins / devs in dev mode, saying that Sylius could not decide about current channel with links to the backend etc.

Pretty much the same behavior should be done for not configured currencies, locales etc. but that's for future.

Let me know what do you think!

pamil:
:+1:, we may create an instance of `Symfony\Component\HttpFoundation\RequestMatcherInterface` for every channel and configure it like in Symfony Security component (but using CLI Installer for the first time & web backend later).

Also, I would like to get `ChannelContextInterface::setChannel()` removed - it only creates confusion, if one wants to set current in code, they can implement channel context interface with setter.

`Sylius\Component\Core\Channel\ChannelConextInterface` should be removed at the same time, as it only changes docblock.

Last but not least, default `ChannelContext` itself shouldn't include caching (sorry! :D), caching should be implemented as a decorator.

pjedrzejewski:
@pamil :+1: on all points.

michalmarcinkowski:
:+1:

peteward:
It's also worth considering that not everyone will use channel by hostname (we don't). I haven't checked how this is implemented, but consider abstracting it to something like a `ChannelResolverStrategy` that could be replaced more easily for people who use channel differently. Even if you only implement your version above.

pjedrzejewski:
@peteward absolutely, Strategy is a nicer name than custom resolvers, see my first point -> ``(for now, later I'd like to have ability to attach custom resolvers and decide on different factors than hostname, could be even locale or country of shipping)``.

peteward:
apologies, missed that part :)

[...]

PR #4037, opened 2016-02-02
[API] [User] Customers CRUD tests

lchrusciel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | partially #3549 
| License       | MIT
| Doc PR        | [#409](https://github.com/Sylius/Sylius-Docs/pull/409)
This PR contains tested Customer CRUD. During tests I've had to update User/Customer serialization, add Customer serialization to CoreBundle. I wasn't able to force HatoasBundle to serialize User and Customer bidirectional, because of [#136 issue](https://github.com/willdurand/Hateoas/issues/136). If you have any thoughts for some workaround or even better for some solution, please share or contribute. Current solution is to embed User into Customer, but Customer cannot be embedded into User. 

I had to update CustomerType, what results in introducing AddUserTypeSubscriber. This subscriber partially fix #3549 , because UserType is no longer coupled with CustomerType.

**Update**
I have resigned with maxDepth strategy and HATEOAS relations in favor of good serialization approach after considering @nakashu [advices](https://github.com/Sylius/Sylius/pull/4037#issuecomment-178563508)

**Update 2**
Added documentation reference

PR #4056, opened 2016-02-04
[Channel] Brand new channel context!

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #3929
| License       | MIT
| Doc PR        | -

Reworking the channel context, introducing request based channel context that uses request resolvers.


src/Sylius/Component/Channel/Context/RequestBased/ChannelContext.php
```diff
@@ -0,0 +1,79 @@
...
+        return $channel;
+    }
+
+    /**
+     * @return Request
+     */
+    private function getMasterRequest()
+    {
+        $masterRequest = $this->requestStack->getMasterRequest();
+        if (null === $masterRequest) {
```
tristanbes: you should break a new line before the if statement for better redability

src/Sylius/Component/Channel/Context/RequestBased/CompositeRequestResolver.php
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+final class CompositeRequestResolver implements RequestResolverInterface
+{
+    /**
+     * @var PriorityQueue|RequestResolverInterface[]
+     */
+    private $priorityQueue;
```
michalmarcinkowski: `$priorityQueue` -> `$requestResolvers`?

src/Sylius/Bundle/ChannelBundle/Doctrine/ORM/ChannelRepository.php
```diff
@@ -24,28 +22,8 @@ class ChannelRepository extends EntityRepository implements ChannelRepositoryInt
...
-            ->setParameter('hostname', '%'.$hostname.'%')
-        ;
-
-        return $queryBuilder->getQuery()->getOneOrNullResult();
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function findDefault()
```
michalmarcinkowski: It would be good to support single channel without need to configure hostname or anything else, maybe we should have a default channel? We could have the default channel resolver that will be the last resolver and will return the channel that is marked as default.

aramalipoor: :+1:

pjedrzejewski: Let's discuss this today, sounds reasonable, but can lead to some unexpected behavior, especially when poorly documented.

pamil: :+1:, but it should be optional and made at `ChannelContext` level as it does not involve any request-related logic. Anyway, I think we can agree that finding default channel is not the responsibility of channel repository.

src/Sylius/Component/Channel/Context/RequestBased/CompositeRequestResolver.php
pamil: :+1:

src/Sylius/Component/Channel/Context/RequestBased/ChannelContext.php
pamil: It's a common practice not to put a blank line between those if one line is assigning the value to A and the second one is `if` doing some comparison using A.

src/Sylius/Bundle/ChannelBundle/Doctrine/ORM/ChannelRepository.php
pamil: Okay, there it is - `Sylius\Component\Channel\Context\SingleChannel\ChannelContext` :)

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/CompositeRequestResolverPass.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        if (!$container->hasDefinition('sylius.context.channel.request_based.resolver')) {
+            return;
+        }
+
+        $requestResolverDefinition = $container->findDefinition('sylius.context.channel.request_based.resolver');
+
+        $taggedServices = $container->findTaggedServiceIds('sylius.context.channel.request_based.resolver');
+        foreach ($taggedServices as $id => $tags) {
+            foreach ($tags as $attributes) {
```
lchrusciel: Won't it be more readable if you extract it to private method?

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/CompositeRequestResolverPass.php
```diff
@@ -0,0 +1,47 @@
...
+namespace Sylius\Bundle\ChannelBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+final class CompositeRequestResolverPass implements CompilerPassInterface
```
lchrusciel: Why do you used final here?

src/Sylius/Component/Channel/composer.json
```diff
@@ -23,7 +23,8 @@
         "php": "^5.5.9|^7.0",
 
         "doctrine/orm": "^2.4.8,<2.5",
-        "sylius/resource": "0.17.*"
+        "sylius/resource": "0.17.*@dev",
```
pamil: To be reverted

src/Sylius/Component/Channel/composer.json
```diff
@@ -37,7 +38,6 @@
     "autoload-dev": {
         "psr-4": { "Sylius\\Component\\Channel\\spec\\": "spec/" }
     },
-    "minimum-stability": "dev",
```
pamil: Minimum stability + prefer stable

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/CompositeRequestResolverPass.php
pamil: There's no need to extend this service, I haven't ever seen a use-case where extending compiler pass would be a solution.

src/Sylius/Component/Channel/composer.json
lchrusciel: Why did you remove this? Did we change convention?

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/CompositeChannelContextPass.php
```diff
@@ -0,0 +1,47 @@
...
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+final class CompositeChannelContextPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('sylius.context.channel')) {
```
pamil: TODO: Define that service at `sylius.context.channel.composite` instead and proceed only if `sylius.context.channel` does not exist. If someone would like to provide different implementation, we don't want to add method calls on another service. At the end of the function the alias `sylius.context.channel` -> `sylius.context.channel.composite` should be set

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/CompositeRequestResolverPass.php
lchrusciel: :+1:

src/Sylius/Component/Channel/Context/SingleChannel/ChannelContext.php
```diff
@@ -0,0 +1,49 @@
...
+namespace Sylius\Component\Channel\Context\SingleChannel;
+
+use Sylius\Component\Channel\Context\ChannelContextInterface;
+use Sylius\Component\Channel\Context\ChannelNotFoundException;
+use Sylius\Component\Channel\Repository\ChannelRepositoryInterface;
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+final class ChannelContext implements ChannelContextInterface
```
michalmarcinkowski: Not not renaming to `SingleChannelContext` and putting it in `Sylius\Component\Channel\Context` namespace? IMO it's overhead to create a separate namespace for that :)

src/Sylius/Component/Channel/spec/Context/RequestBased/CompositeRequestResolverSpec.php
```diff
@@ -0,0 +1,90 @@
...
+        $this->shouldHaveType('Sylius\Component\Channel\Context\RequestBased\CompositeRequestResolver');
+    }
+
+    function it_implements_request_resolver_interface()
+    {
+        $this->shouldImplement(RequestResolverInterface::class);
+    }
+
+    function it_returns_null_if_there_are_no_nested_request_resolvers_added(
+        Request $request
```
michalmarcinkowski: No need to put it in a separate line :)

michalmarcinkowski:
Well done Kamil! :+1:

src/Sylius/Component/Channel/Context/SingleChannel/ChannelContext.php
pamil: :+1:

src/Sylius/Component/Channel/Context/RequestBased/CompositeRequestResolver.php
```diff
@@ -0,0 +1,58 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function findChannel(Request $request)
+    {
+        foreach ($this->requestResolvers as $requestResolver) {
+            $channel = $requestResolver->findChannel($request);
+
+            if (null === $channel) {
```
pjedrzejewski:
```
if (null !== $channel) {
    return $channel;
}
```

src/Sylius/Component/Channel/spec/Context/SingleChannelContextSpec.php
```diff
@@ -0,0 +1,69 @@
...
+    function it_returns_a_channel_if_it_is_the_only_one_defined(
+        ChannelRepositoryInterface $channelRepository,
+        ChannelInterface $channel
+    ) {
+        $channelRepository->findAll()->willReturn([$channel]);
+
+        $this->getChannel()->shouldReturn($channel);
+    }
+
+    function it_throws_an_channel_not_found_exception_if_there_are_no_channels_defined(
```
lchrusciel: Won't it be better to test all exception throws in a one function?

pamil: There are the same behaviours expected from different backgrounds, specifications should say explicitly when something happens, not only that "it_throws_an_exception".

src/Sylius/Bundle/ChannelBundle/Tests/DependencyInjection/Compiler/CompositeRequestResolverPassTest.php
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+class CompositeRequestResolverPassTest extends AbstractCompilerPassTestCase
+{
+    /**
+     * @test
+     */
+    public function it_collects_tagged_request_based_channel_contexts()
```
lchrusciel: Do we have any convention defined for phpUnit tests? IIRC we should use camel case instead of snake case or am I missed something?

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/PrioritizedCompositeServicePass.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        if (!$container->hasDefinition($this->compositeId)) {
+            return;
+        }
+
+        $channelContextDefinition = $container->findDefinition($this->compositeId);
+
+        $taggedServices = $container->findTaggedServiceIds($this->tagName);
+        foreach ($taggedServices as $id => $tags) {
+            foreach ($tags as $attributes) {
```
lchrusciel: Won't it be cleaner to extract this for each to separate method?

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/CompositeRequestResolverPass.php
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Sylius\Bundle\ChannelBundle\DependencyInjection\Compiler;
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+final class CompositeRequestResolverPass extends PrioritizedCompositeServicePass
+{
+    /**
+     * {@inheritdoc}
```
lchrusciel: Do we need this inheritdoc if constructor doesn't need any parameters?

src/Sylius/Bundle/ChannelBundle/Tests/DependencyInjection/Compiler/CompositeRequestResolverPassTest.php
pamil: I don't know, I used the same convention as in PhpSpec because snake_case makes reading long sentences easier that camelCase.

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/PrioritizedCompositeServicePass.php
pamil: :+1:

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/CompositeRequestResolverPass.php
pamil: In fact I'm not sure here - on the one hand it does not need these arguments, on the other - I consider `{@inheritdoc}` like Java's `@Override`, just as a mark that the method overrides parent method.

src/Sylius/Component/Channel/spec/Context/SingleChannelContextSpec.php
lchrusciel: I'm totally agree for above case in [CompositeChannelContextSpec](https://github.com/Sylius/Sylius/pull/4056/files#diff-cd860143795fd6ba7770e844de50c2ceR37). But here we have just different values returned by repository function. Won't it be better?

```php
    function it_throws_an_channel_not_found_exception_if_there_isnt_one_channel_defined(
        ChannelRepositoryInterface $channelRepository
    ) {
        $channelRepository->findAll()->willReturn([], [$firstChannel, $secondChannel]);

        $this->shouldThrow(ChannelNotFoundException::class)->during('getChannel');
        $this->shouldThrow(ChannelNotFoundException::class)->during('getChannel');
    }
```

src/Sylius/Bundle/ChannelBundle/Tests/DependencyInjection/Compiler/CompositeRequestResolverPassTest.php
lchrusciel: You are right, but on the other hand shouldn't we keep consistency? At the end tests in theme bundle or api tests are written with camel case. Should we force some convention?

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/CompositeRequestResolverPass.php
lchrusciel: I'm not convinced about that. For example in Product class in core bundle we use parent constructor, but don't use `{@inheritdoc}` tag.

src/Sylius/Component/Channel/spec/Context/SingleChannelContextSpec.php
pamil: IMHO it makes it much more difficult to read, if I can make two simple specifications instead of one complex, I'll go for the first option.

src/Sylius/Bundle/ChannelBundle/Tests/DependencyInjection/Compiler/CompositeRequestResolverPassTest.php
pamil: Right now we does not regulated any convention, ThemeBundle has already two of them inside (camelCase in functional, snake_case in configuration). If there will be any, I'll follow it, but for now I'll let this as is and vote for snake_case.

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/CompositeRequestResolverPass.php
pamil: All right, it's not a big deal.

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/CompositeRequestResolverPass.php
```diff
@@ -0,0 +1,28 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\ChannelBundle\DependencyInjection\Compiler;
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+final class CompositeRequestResolverPass extends PrioritizedCompositeServicePass
```
pjedrzejewski: This based class is pretty generic, later we should think about appropriate place for that. We use so much stuff from Symfony DIC that we could have a standalone Sylius DependencyInjection component with all this crazy stuff haha. :)

src/Sylius/Bundle/ChannelBundle/Tests/DependencyInjection/Compiler/CompositeRequestResolverPassTest.php
pjedrzejewski: I did snake_case in #3936 and I think we should use it because it is more readable. That being said, I am not sure how that will work out for API tests in PHPUnit:

```
function it_allows_creating_product()
function it_denies_access_to_product_creation_for_not_authenticated_users()
function it_returns_not_found_response_for_non_existing_product()
function it_creates_customer_without_user_account()
function it_allows_to_delete_users()
function it_allows_to_update_users()
```

Thoughts?

src/Sylius/Bundle/ChannelBundle/Tests/DependencyInjection/Compiler/CompositeRequestResolverPassTest.php
```diff
@@ -0,0 +1,130 @@
...
+            'sylius.context.channel.request_based.resolver',
+            'addResolver',
+            [new Reference('sylius.context.channel.request_based.resolver.tagged_one'), 42]
+        );
+    }
+
+    /**
+     * @test
+     */
+    public function it_does_not_add_method_calls_to_overwriting_service_if_the_composite_service_is_overwrited()
```
pjedrzejewski: overwritten or overridden?

src/Sylius/Bundle/ChannelBundle/composer.json
```diff
@@ -27,7 +27,9 @@
         "symfony/framework-bundle": "^2.7.7"
     },
     "require-dev": {
+        "matthiasnoback/symfony-dependency-injection-test": "^0.7.6",
```
pjedrzejewski: Wouldn't be more appropriate to have this next to PHPUnit? We do not sort deps alphabetically.

src/Sylius/Bundle/ChannelBundle/composer.json
```diff
@@ -41,6 +43,7 @@
         "psr-4": { "Sylius\\Bundle\\ChannelBundle\\spec\\": "spec/" }
     },
     "minimum-stability": "dev",
+    "prefer-stable": true,
```
pjedrzejewski: Why? Symfony components/bundles do not use that flag.

src/Sylius/Bundle/ChannelBundle/Tests/DependencyInjection/Compiler/CompositeRequestResolverPassTest.php
lchrusciel: It can be even better as test names for API can be pretty long. :+1: I will adapt #4037 to snake case.

src/Sylius/Bundle/ChannelBundle/composer.json
pamil: Installing that bundle standalone without prefer-stable downloads a lot of non-stable dependencies. If we want to test bundles standalone, I believe that it is the way to go. 

It does not affect any packages using it as these configuration options - "minimum-stability" and "prefer-stable" - are considered only for root packages. Sylius dependencies requiring `0.17.*` still resolves to the latest, development version.

src/Sylius/Bundle/ChannelBundle/composer.json
pamil: IIRC Sylius' main `composer.json` is sorted in that way, so I decided to stick to that.

src/Sylius/Bundle/ChannelBundle/Tests/DependencyInjection/Compiler/CompositeRequestResolverPassTest.php
pamil: Ooops, something has gone wrong! :)

src/Sylius/Bundle/ChannelBundle/DependencyInjection/Compiler/CompositeRequestResolverPass.php
pamil: Yep, the same with dynamic forms in metadata and so on...

src/Sylius/Bundle/ChannelBundle/Tests/DependencyInjection/Compiler/CompositeRequestResolverPassTest.php
pamil: Awesome :+1:

michalmarcinkowski: :+1: 

Just to clarify, in order to stick to the convention with 'allow' we should rename:
`function it_creates_customer_without_user_account()`
to:
`function it_allows_to_create_customer_without_user_account()`

src/Sylius/Bundle/ChannelBundle/composer.json
pjedrzejewski: Ok, I opened RFC to add this to all bundles/components then - https://github.com/Sylius/Sylius/issues/4088.

pjedrzejewski:
Awesome work Kamil! :+1:

PR #4057, opened 2016-02-04
Add  "doctrine/collections"

jmontoyaa:
According to this:

https://github.com/Sylius/Attribute/blob/master/Model/AttributeSubjectInterface.php#L14

Doctrine collections must be present ...

### DIFF
--- a/src/Sylius/Component/Attribute/composer.json
+++ b/src/Sylius/Component/Attribute/composer.json
@@ -25,7 +25,8 @@
         "sylius/resource": "0.17.*",
         "sylius/translation": "0.17.*",
         "sylius/registry": "0.17.*",
-        "symfony/validator": "^2.7.7"
+        "symfony/validator": "^2.7.7",
+        "doctrine/collections": "~1.2"
     },
     "require-dev": {
         "phpspec/phpspec": "^2.4"


okwinza:
IIRC there was talk about replacing it with something generic/sylius based.

jmontoyaa:
In the mean time 0.16 was released with Collection :smile:

pamil:
IMHO methods like `setCollection` should be removed as they are very rarely used and can be replaced through iterating and calling `addCollectionElement`. They also couple model interfaces with Doctrine Collections.

pamil:
For example, this method is never used in Sylius codebase.

pjedrzejewski:
:+1: for removing the setters, especially if we do not use them but that still leaves the deps here https://github.com/Sylius/Attribute/blob/master/Model/Attribute.php#L63.

michalmarcinkowski:
Yeah, but removing `Collection` from the interface at least do not enforce you to use it in your implementation, so that's a huge benefit.

src/Sylius/Component/Attribute/composer.json
```diff
@@ -25,7 +25,8 @@
         "sylius/resource": "0.17.*",
         "sylius/translation": "0.17.*",
         "sylius/registry": "0.17.*",
-        "symfony/validator": "^2.7.7"
+        "symfony/validator": "^2.7.7",
+        "doctrine/collections": "~1.2"
```
lchrusciel: Can we use lower version?

jmontoyaa: I got that value from here:

https://github.com/Sylius/Sylius/blob/master/composer.json#L28

lchrusciel: :+1:

michalmarcinkowski:
Thanks Julio! :+1:

PR #4061, opened 2016-02-05
[POC] Refund Adjustments

peteward:
**DO NOT REVIEW FOR CS** 

This is just a simple POC for @pjedrzejewski and @michalmarcinkowski to show some issues around refunds. It is copy&paste from a PR on my project so is not fit for proper review.

I've done this simple implementation by adding a `refund` flag to `Adjustment` that marks if it is a refund (or reversal) of an original `Adjustment`. There is currently no relationship between the refund adjustment and the original that it is refunding (I'm not sure that it's necessary). This first implementation only covers an initial use-case for us of refunding shipping and out-of-stock products at dispatch time - there will be more to come.

This very simple model works fine up to a point. There are 2 issues to highlight:

1). Once you get into refunds (and I can't imagine a merchant not needing this facility), you really don't want to be dealing with `Adjustments` at the `OrderItem` level... I'll leave that to you, it's not doing us any harm to have this relationship, but we certainly wouldn't want to be doing the calculations to take an OrderItem-level adjustment and part-refund it at an OrderItemUnit level.

2). More importantly, it became obvious very quickly that we are missing a top-level view that differentiates the `Order.grossTotal` from `Order.netTotal`. Currently there is just total, and with the wonderful calculation logic that you've added and my code above, this effectively becomes a netTotal because all adjustments are summed, giving more of an "order balance".

I don't think there's anything wrong with this, but I do think that for top-level reporting it's important to have a view of gross and net of refunds, to distinguish between the site demand, captured money and refunds.

This could either be by storing `grossTotal` and `netTotal` explicitly or by splitting `adjustmentsTotal` and separating out `refundAdjustmentsTotal` (probably the latter seems more appropriate and could be done at both the OrderItem and Order levels).

If you were doing simple queries for reporting either through ORM or any other means then you have what you need through `order.total` (net) and `order.total - refundAdjustmentsTotal` (gross), and means there is no requirement to cascade down and aggregate at lower levels.

It would be interesting to get your thoughts!


src/Sylius/Bundle/OrderBundle/Modifier/AdjustmentManager.php
```diff
@@ -0,0 +1,90 @@
...
+        $this->entityManager     = $entityManager;
+    }
+
+    /**
+     * @param OrderItemUnit $unit
+     * @param string        $descriptionSuffix
+     */
+    public function buildRefundAdjustmentsForUnit(OrderItemUnit $unit, $descriptionSuffix = 'Return')
+    {
+        if (count($unit->getRefundAdjustments())) {
```
stloyd: Maybe:
```php
if (!$unit->getRefundAdjustments()->isEmpty()) {}

File: src/Sylius/Bundle/OrderBundle/Modifier/AdjustmentManager.php [hunk 2]
```diff
@@ -0,0 +1,90 @@
...
+            // If already has some, do nothing.
+            return;
+        }
+
+        foreach ($unit->getAdjustments() as $adjustment) {
+            $refundAdjustment = clone $adjustment;
```
stloyd: This could be done in factory, just maybe change the method:
```php
public function createUnitRefund(OrderItemUnit $unit, $label = 'Refund', Adjustment $baseAdjustment = null)
{
   $adjustment = null !== $baseAdjustment ? clone $baseAdjustment : $this->createNew();
   // ...
}

File: src/Sylius/Component/Order/Model/Order.php [hunk 3]
```diff
@@ -531,6 +531,16 @@ protected function subtractFromAdjustmentsTotal(AdjustmentInterface $adjustment)
     }
 
     /**
+     * @return Collection|Adjustment[]
+     */
+    public function getRefundAdjustments()
+    {
+        return $this->adjustments->filter(function (Adjustment $adjustment) {
+            return true === $adjustment->isRefund();
```
stloyd: No need to strict comparsion as this is always boolean value.

peteward:
@stloyd - read the top of the description in bold.

stloyd:
@peteward This is not (mostly) CS ;)

Anyway this looks good, but having split of "total" into two "parts" sounds logical IMO (or at least it should not be called "total" but "balance"/"summary").

For first steps, IMO it's acceptable to "skip" first issue, as this at least allows to go with refunds, and will show more details while using it in production it requires such deeper adjustments or it will work well ;)

src/Sylius/Bundle/OrderBundle/Modifier/AdjustmentManager.php
peteward: A unit refund is a bit different from refunding an existing Adjustment, I wouldn't merge the too, I think it obscures what's happening.  Although there's definitely a valid case for moving the cloning into the Factory.

peteward: We were debating about using clone anyway, so that's a point of contention in itself...

src/Sylius/Bundle/OrderBundle/Factory/AdjustmentFactory.php
```diff
@@ -0,0 +1,40 @@
...
+     * @param string        $label
+     *
+     * @return Adjustment
+     */
+    public function createUnitRefund(OrderItemUnit $unit, $label = 'Refund')
+    {
+        /** @var Adjustment $adjustment */
+        $adjustment = $this->createNew();
+        $adjustment->setRefund(true);
+        $adjustment->setAmount($unit->getOrderItem()->getUnitPrice() * -1);
```
michalmarcinkowski: `$unit->getTotal() * -1` - this is a real value that needs to be neutralized.

michalmarcinkowski:
@peteward few words from me:

1. Why do we need `refund` boolean on adjustment? Can't we simply get all adjustments of type `Adjustment::UNIT_REFUND_ADJUSTMENT`?
2. Since the tax applied for shipping is on order level, as well as shipping adjustment then in order to refund shipping we can simply:
```php
$refundAmount = $order->getAdjustmentsTotal(Adjustment::SHIPPING_ADJUSTMENT) + $order->getAdjustmentsTotal(Adjustment::TAX_ADJUSTMENT);
// when we will introduce promotions then also:
$refundAmount += $order->getAdjustmentsTotal(Adjustment::ORDER_PROMOTION_ADJUSTMENT);
...
if (0 !== $refundAmount) {
    $order->addAdjustment(-$refundAdjustment);
}
```
3. In order to refund `OrderItemUnit` we simply create adjustment with amount equal to negative `$unit->getTotal()`, that's all.

peteward:
Although it negates the total, you cannot distinguish between the amount of individual adjustment types for reporting if you go for this simplistic approach.

Here's a few examples:

For US Tax filing at the end of each month we need to report on the amount of tax captured from customers and also refunded to them. If a customer returns a product to us we don't have to pay the tax on it (obviously), so it's important to know the exact tax amount coming back, especially as it is likely to be in the subsequent months tax file.

For internal reporting for promotions you we want to know how much was originally spent on promotions compared to how much was retained or refunded.

The only way to achieve this fine-grain detail fully is to take each individual adjustment and 'reverse' or refund it. This way you have a full record of each amount to report on.

michalmarcinkowski:
Yeah, so it looks like you have to create each refund adjustment on each applied adjustment...

peteward:
yes, it's fun! :)

And one for the original item amount (unit_price), this is what the `Adjustment::UNIT_REFUND_ADJUSTMENT` is intending to represent, although maybe `UNIT_PRICE_REFUND` would be a more accurate name...

Anyway this approach will give a comprehensive audit trail for any kind of reporting you could ask for.

michalmarcinkowski:
Maybe it's even worth going into specialized types like `UNIT_PRICE_REFUND`, `PROMOTION_REFUND`, `TAX_REFUND` etc.

peteward:
We thought about that, but I think a combination of the refund flag against the original type gives you the most flexibility as you can do AND / OR combinations.

If as we talked about this morning you have ORDER_PROMOTION and ORDER_ITEM_PROMOTION then you are doubling the number of types and have to have an explicit negation type for each.

The refund flag means you can just take an adjustment, mark the refund flag and negate the value.

michalmarcinkowski:
Sounds reasonable :+1:

peteward:
Hi @michalmarcinkowski,

I've updated the `Order`, `OrderItem` and `OrderItemUnit` models to "piggy-back" on your wonderful automated recalculation logic for ensuring we have correctly maintained refund totals.

This is still very rough and issues may arise when I look at writing tests for it.

I was just wondering what your opinion was of this approach? By storing the `total` and `totalRefund` separately we now have all 3 values that you could want for reporting:

* Gross total exc. refunds (`total` + `totalRefund`)
* Net total inc. refund (`total`)
* Refund total (`refundTotal`)

I could have chosen another approach and stored grossTotal instead. As long as you have 2 of the 3 values you can obviously calculate the third so it becomes a choice of preference.

I know this isn't going to get merged anytime soon but I'm hoping not to deviate far from how Sylius will eventually do this and work out how best to implement it - either as part of our application or our Sylius fork that the app is based on...

Thanks!

PR #4065, opened 2016-02-05
[Product] VariantFactory naming changes

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

- changed ``VariantFactory`` into ``ProductVariantFactory``, as it's placed in *Product* component and creates only ``ProductVariant`` objects
- changed ``createForProduct`` method into ``createBasedOnProductId``
- change invalid parameter names in factory and factory interface

### DIFF
--- a/src/Sylius/Bundle/ProductBundle/DependencyInjection/Compiler/ServicesPass.php
+++ b/src/Sylius/Bundle/ProductBundle/DependencyInjection/Compiler/ServicesPass.php
@@ -56,12 +56,12 @@ class ServicesPass implements CompilerPassInterface
         $variantFactoryClass = $variantFactoryDefinition->getClass();
         $variantFactoryDefinition->setClass(Factory::class);
 
-        $decoratedVariantFactoryDefinition = new Definition($variantFactoryClass);
-        $decoratedVariantFactoryDefinition
+        $decoratedProductVariantFactoryDefinition = new Definition($variantFactoryClass);
+        $decoratedProductVariantFactoryDefinition
             ->addArgument($variantFactoryDefinition)
             ->addArgument(new Reference('sylius.repository.product'))
         ;
 
-        $container->setDefinition('sylius.factory.product_variant', $decoratedVariantFactoryDefinition);
+        $container->setDefinition('sylius.factory.product_variant', $decoratedProductVariantFactoryDefinition);
     }
 }
--- a/src/Sylius/Component/Product/Factory/ProductVariantFactory.php
+++ b/src/Sylius/Component/Product/Factory/ProductVariantFactory.php
@@ -17,7 +17,7 @@ use Sylius\Component\Resource\Repository\RepositoryInterface;
 /**
  * @author Paweł Jędrzejewski <pawel@sylius.org>
  */
-class VariantFactory implements VariantFactoryInterface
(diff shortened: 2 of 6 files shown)


src/Sylius/Bundle/ProductBundle/DependencyInjection/Compiler/ServicesPass.php
```diff
@@ -56,12 +56,12 @@ public function process(ContainerBuilder $container)
         $variantFactoryClass = $variantFactoryDefinition->getClass();
         $variantFactoryDefinition->setClass(Factory::class);
```
michalmarcinkowski: Shouldn't there be `setClass($variantFactoryClass)`?

src/Sylius/Component/Product/Factory/ProductVariantFactory.php
```diff
@@ -50,15 +50,15 @@ public function createNew()
     /**
      * {@inheritdoc}
      */
-    public function createForProduct($productId)
+    public function createBasedOnProductId($id)
     {
-        if (null === $product = $this->productRepository->find($productId)) {
-            throw new \InvalidArgumentException(sprintf('Product with id "%s" does not exist.', $productId));
+        if (null === $product = $this->productRepository->find($id)) {
```
pamil: Can we extract the assignment before the `if`?

pjedrzejewski:
1. Not sure about the renaming of the factory, currently in Product component we have https://github.com/Sylius/Product/blob/master/Model/Variant.php model, not ProductVariant, because it seemed too obvious to say it for all other models. ProductVariant, ProductAttributeValue, etc. We actually have a mismatch there right now, since introducing ProductAssociation.

So we should first agree on that convention and then rename the factory if needed. I'd vote for leaving out the Product prefix here, cause it seems to be duplicated when we have Product in the namespace already. If we ever end up with 2 Variant concepts in that namespace (I doubt) then it is a much bigger problem than prefixing/not-prefixing stuff.

2. ``createBasedOnProductId`` does not sound really good to me... :(

``createForProductId``
``createForProductWithId``

Saying "Create variant based on product id" may sound reasonable but in reality we are "Creating a variant of a product with ID", but ``createOfProductWithId`` sounds terrible too... :D

Zales0123:
1) Ok, that sound reasonable for me :+1: 
2) ``createForProductWithId`` looks the more appropriate for me (``I create variant for product with id "$id")

pjedrzejewski:
Let's hear few more opinions on the naming because we will go with that convention for other factories as well. :)

pamil:
I'm for adding subject to the name of generic classes like `Variant` as soon as they are aware of it. Like our `Sylius\Component\Product\Model\Variant` has `getProduct()` and `setProduct()` methods, therefore it should be named `ProductVariant`. If it would not have them, it should be still named `Variant`.

Zales0123:
Ok, so finally after hours of discussion we decided to stay with ``ProductVariantFactory`` class and ``createForProductWithId`` method ;)

pjedrzejewski:
Thanks Mateusz! :+1:

PR #4076, opened 2016-02-08
[Behat] Added default behat.yml extending behat.yml.dist

pamil:
Added `behat.yml` boilerplate to make customizing Behat even easier!

### DIFF
--- a/behat.yml
+++ b/behat.yml
@@ -1,0 +1,8 @@
+imports: ["behat.yml.dist"]
+
+# Uncomment to customize base_url:
+
+#default:
+#    extensions:
+#        Behat\MinkExtension:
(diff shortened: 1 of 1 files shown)


behat.yml
```diff
@@ -0,0 +1,8 @@
+imports: ["behat.yml.dist"]
```
pjedrzejewski: Any reason to use double quotes?

pamil: Seems cleaner to me than:
```yaml
imports: [behat.yml.dist]
```

michalmarcinkowski:
:+1:

pjedrzejewski:
:+1:

PR #3737, opened 2015-12-17
[WIP] Promotion Engine Enhancements

peteward:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | yes |
| Deprecations? | no |
| Fixed tickets | #3368 |
| License | MIT |

Implements #3368. Please read for a deeper understanding of what this is trying to achieve.

**It's WIP so don't review code style. Tests will break since @pjedrzejewski merged Driver refactor this morning**

I haven't had as much time as I wanted today to polish this. Despite the incomplete tasks, a large amount of the hard work has been done.
- [x] New core structure of `Actions` which are now groupings of `Benefits` and `Filters`
- [x] Implementation of several base Filters - Taxon, Archetype, Cheapest, Most Expensive, Every, Price
- [x] Many new tests added and existing updated. Core behat tests have been implemented to decouple UI to test domain logic.
- [x] FixedDiscount features file needs work (will do this tonight).
- [ ] `Action` probably needs a `position` to handle priority.
- [ ] Admin UI needs implementing as per the referenced issue. 
- [ ] `PromotionApplicator` needs decoupling as it's currently implemented to expect Orders (`getItems()`), for the purposes of filtering order items within it. Possibly a `getFilterSubjects()` or `getFilterable()` method? 
- [ ] Similarly deciding what should be a PromotionSubject - just an `Order`? 
- [ ] Behat tests need an overhaul in structure. There are too many duplicates spread in different files. There should be features to cover admin UI, some for frontend UI, then the main bulk to cover the permutations of available promotion building.
- [ ] Integration tests to be removed? These were from before @pamil made behat fast.
- [ ] Finally the actual application of Adjustments resulting from Benefits needs to be thought out. Currently it's working at the OrderItem-level, which works but breaks some of the current UI tests. A plan should be worked out how to complete this in a complimentary way to changes suggested in #3712.

Suggested next steps:
- Defined the changes to the Admin UI via a replacement Behat file.
- Implement Admin UI to populate implemented structure.
- The rest!

PR #4085, opened 2016-02-09
[WIP][RFC][Behat][Promotion] Basic scenarios for promotions

Zales0123:
Hello everybody!

Taxes and shipping related scenarios will be, hopefully, fully merged soon, so it's time to introduce some new stuff :smile:
Here are two simple features about promotions, that can be a base for further, more hardcore scenarios. Lots of steps are already implemented, the only left are those related with promotions logic. For now, there are scenarios for "Fixed discount" and "Percentage discount" promotions, but obviously it's needed to cover as many cases as possible...
And this is main thing to discuss. In there nearest future, we'll try to implement big promotions system improvement (inspired by @peteward https://github.com/Sylius/Sylius/pull/3737). There are a lot of possible promotion combinations and will be even more after this work. We should deeply consider, how many cases do we want to cover with behat.

However, for now there are features that are possible to achieve with current promotions system. I will be glad to hear your thoughts, is *force* strong with these scenarios and is it the way we should go with them ;)

### DIFF
--- a/features/shop/apply_correct_simple_promotion_on_order.feature
+++ b/features/shop/apply_correct_simple_promotion_on_order.feature
@@ -1,0 +1,32 @@
+@ui-cart
+Feature: Apply correct simple promotions on order
+    In order to pay proper amount while buying promoted goods
+    As a Customer
+    I want to have promotions applied
+
+    Background:
+        Given the store is operating on a single channel
+        And default currency is "EUR"
+        And store has a product "PHP T-Shirt" priced at "€100.00"
+        And store has a product "Symfony Mug" priced at "€5.00"
+        And there is promotion "Special discount"
+        And it has "€10.00" "fixed discount" for cart with "item total" above "€80.00"
+        And there is promotion "Special discount"
+        And it has "10%" "percentage discount" for cart with "item count" above "5"
+        And there is user "john@example.com" identified by "password123"
+        And I am logged in as "john@example.com"
+
+    Scenario: Applying fixed discount promotion
+        Given I have product "PHP T-Shirt" in the cart
+        Then my cart total should be "€90.00"
+        And my cart promotions should be "-€10.00"
+
+    Scenario: Applying percentage discount promotion
+        When I add 6 products "Symfony Mug" to the cart
+        Then my cart total should be "€27.00"
+        And my cart promotions should be "-€3.00"
+
+    Scenario: Applying both promotions that order fits
+        When I add 6 products "PHP T-Shirt" to the cart
+        Then my cart total should be "€531.00"
+        And my cart promotions should be "-€69.00"
--- a/features/shop/apply_correct_simple_promotion_with_taxes_on_order.feature
+++ b/features/shop/apply_correct_simple_promotion_with_taxes_on_order.feature
@@ -1,0 +1,43 @@
+@ui-cart
+Feature: Apply correct simple promotions with taxes on order
+    In order to pay proper amount while buying promoted taxed goods
+    As a Customer
+    I want to have promotions and taxes applied
+
+    Background:
+        Given the store is operating on a single channel
+        And there is "EU" zone containing all members of European Union
+        And default currency is "EUR"
+        And default tax zone is "EU"
+        And store has a product "PHP T-Shirt" priced at "€100.00"
+        And store has a product "Symfony Mug" priced at "€5.00"
+        And there is promotion "Special discount"
+        And it has "€10.00" "fixed discount" for cart with "item total" above "€80.00"
+        And there is promotion "Special discount"
+        And it has "10%" "percentage discount" for cart with "item count" above "5"
+        And store has "EU VAT" tax rate of 23% for "Clothes" within "EU" zone
+        And store has "Low VAT" tax rate of 5% for "Mugs" within "EU" zone
+        And store has a product "PHP T-Shirt" priced at "€100.00"
+        And product "PHP T-Shirt" belongs to "Clothes" tax category
+        And store has a product "Symfony Mug" priced at "€50.00"
+        And product "Symfony Mug" belongs to "Mugs" tax category
+        And there is user "john@example.com" identified by "password123"
+        And I am logged in as "john@example.com"
+
+    Scenario: Applying fixed discount promotion and taxes
+        Given I have product "PHP T-Shirt" in the cart
+        Then my cart total should be "€110.70"
+        And my cart promotions should be "-€10.00"
+        And my cart taxes should be "€20.70"
+
+    Scenario: Applying percentage discount promotion and taxes
+        When I add 6 products "Symfony Mug" to the cart
+        Then my cart total should be "€28.35"
+        And my cart promotions should be "-€3.00"
+        And my cart taxes should be "€1.35"
+
+    Scenario: Applying both promotions that order fits and taxes
+        When I add 6 products "PHP T-Shirt" to the cart
+        Then my cart total should be "€653,13"
+        And my cart promotions should be "-€69.00"
+        And my cart taxes should be "€122.13"


features/shop/apply_correct_simple_promotion_on_order.feature
```diff
@@ -0,0 +1,32 @@
...
+    As a Customer
+    I want to have promotions applied
+
+    Background:
+        Given the store is operating on a single channel
+        And default currency is "EUR"
+        And store has a product "PHP T-Shirt" priced at "€100.00"
+        And store has a product "Symfony Mug" priced at "€5.00"
+        And there is promotion "Special discount"
+        And it has "€10.00" "fixed discount" for cart with "item total" above "€80.00"
```
pamil: Can we use discount types without wrapping them with quotation marks? `(.+? discount) for cart`

Zales0123: We could think about it, but then we should standardization of discount names :D For example, now we have "Fixed **discount**", "Percentage **discount**" and "Shipping **discount**" actions, but also "Add product" ;)

pamil: Anyway, if we don't want to get into trap like with country names transforming to country codes in old contexts, we may even want to create anemic, simple steps for every discount type.

features/shop/apply_correct_simple_promotion_on_order.feature
```diff
@@ -0,0 +1,32 @@
...
+        And there is promotion "Special discount"
```
michalmarcinkowski: Why duplicated step?

features/shop/apply_correct_simple_promotion_on_order.feature
Zales0123: Sure, we can think about it :+1: Trap is never a good option.
tumblr_inline_nzfevykxev1rg0x34_500

features/shop/apply_correct_simple_promotion_on_order.feature
Zales0123: Well, not duplicated but badly named, it is a different promotion ;)

features/shop/apply_correct_simple_promotion_on_order.feature
pjedrzejewski: This leaks implementation detail and does not sounds natural when you read it. I think we should have steps for different promotions, it is not that big overhead:

```
And store is running promotion "Christmas Sale"
And it gives €10 discount for customers with carts above €80
```

pjedrzejewski: Or:

```
And store is running promotion "Christmas Sale"
And it gives 10% discount for customers with cart total above €100
```

michalmarcinkowski:
I would go for something similiar to:
```
And the store has promotion "Get €10 off for order with total greater that €80"
And it has fixed discount of €10.00 for a cart with total greater than €80.00
```

Later we also support discount and criteria split e.g.
```
And the store has promotion "Get €10 off and free shipping for order with total greater that €80"
And it has fixed discount of €10.00
And it has a free shipping discount
And it it applied to a cart with items total greater than €80.00
```

features/shop/apply_correct_simple_promotion_on_order.feature
pjedrzejewski: Future scenarios:

```
And store is running promotion "Super-duper Lottery"
And it gives 10% discount for all "Dresses" with a valid coupon
And this promotion has following coupons:
   | code  |
   | X653 |
   | T534 |
And coupon "X653" has been used already
```

And so on, we should stay close to this style IMO.

pamil: @pjedrzejewski :+1:

pjedrzejewski:
@michalmarcinkowski In your example the promotion name contains a lot of information which is irrelevant to how it is configured:

1. I noticed " only on the second look.
2. Easy to get desync between promo name and details
3. Generally confusing to have it duplicated

I get bored after reading 20 steps in the background. I think we should prefer short, promotion specific step definitions over implementing builder pattern in the background. :D

features/shop/apply_correct_simple_promotion_on_order.feature
Zales0123: :+1:

michalmarcinkowski: :+1:

Zales0123:
I think that for simple promotion, step like ``And it gives €10 discount for customers with carts above €80`` is really good - we have every information in one line, and it could not be clearer. For more complicated promotions, we could implement a hardly-considered table, as it probably will be more clearer than millions of steps one below another... we should definitely think about ``Background`` size, as it have potential to be a real poem in more specified features ;)

pamil:
Btw. can we extract "simple" promotions to two features? The first one for fixed discount, the second for percentage discount.

pjedrzejewski:
:+1: for splitting them as much as possible.

```
fixed_discount_promotion.feature
percentage_discount_promotion.feature
```

Or perhaps more specific:

```
fixed_discount_promotion_for_carts_with_total_over_X.feature
etc.
```

Not sure about the X thingy, but I can easily get what it means. :D

Zales0123:
Btw, are we ready to use some folders to group features inside ``features/shop/`` etc? :smile: We already have 9 ``.feature`` files inside, 2 two go, only this PR would contain 3 or more... It starts to be quite messy.

Zales0123:
However :+1: for splitting. It is a good question how we should do that, by rule? By action? By rule/action? Features number/Scenarios number balance would be pretty hard to be kept :smile:

pjedrzejewski:
I think it is impossible to come up with a convention that will fit all cases, because you may have combinations of rules, actions, coupons, channels, etc. Let's try to come up with meaningful name for every feature. So:

If we want to describe promotions which have coupons with % discount, we go for:

```
percentage_discount_promotion_applied_with_a_coupon.feature
```

Zales0123:
Ok, so for now I see such features:
- ``apply_fixed_discount_promotion_on_order``
- ``apply_percentage_discount_promotion_on_order``
- ``apply_fixed_discount_promotion_with_taxes_on_order``
- ``apply_percentage_discount_promotion_with_taxes_on_order``

Or shall ``with_taxes`` features be in the *basics* ones? And what with applying multiple promotion? We can handle multiple promotions with the same discount type inside features mentioned above, but there also should be a case with multiple different promotions... ``apply_multiple_different_promotions_on_order``? But there it should be also ``with_taxes`` feature? :smile: In the worst case, there are 6 features for these very simple cases ;/

michalmarcinkowski:
IMO we should divide features to:
 * Verifying promotion rules (I would use a fixed discount action in examples these features):
  * `apply_promotion_based_on_order_total` - scenarios like greater than, equal to, lesser than etc.
  * `apply_promotion_based_on_items_quantity` - scenarios like greater than, equal to, lesser than etc.
  * `apply_promotion_based_on_items_quantity_by_category` - (same as above)
  * `apply_promotion_based_on_order_number` - (same as above)
  * the rest of features for applying single rules
 * Applying promotion actions without any rules
  * `apply_fixed_discount_on_order` - scenarios for edge cases like discount is greater than order total
  * `apply_fixed_discount_on_order_with_shipping_and_taxes` - I would merge verifying shipping and taxes in one feature to reduce the amount of scenarios. We should verify the discount is not applied to shipping and the taxes are correct after the discount is applied

Let's start with one feature, can be `apply_fixed_discount_on_order` and see how this works out. I would avoid writing many features at once, because it will be easier to iterate on them one by one.

pjedrzejewski:
@michalmarcinkowski Makes sense to me. :+1:

michalmarcinkowski:
@Zales0123 I believe this can be closed.

Zales0123:
Yes, I think that everything what should be discussed, have been discussed here ;)

PR #4092, opened 2016-02-10
[Composer] Updated lock for lakion/mink-debug-extension

pamil:


pamil:
Composer could not be authenticated :(

pamil:
Will be fixed by #4151.

Issue #4094, opened 2016-02-10
sylius:install:sample-data  "Channel could not be found!"


peteward:
Latest master from 2 mins ago:

```
php app/console sylius:install:sample-data -v
Warning! This will erase your database. Your current environment is dev.
Load sample data? (y/N) y
Loading sample data...
 0/3 [|                           ]   0%  1 sec

                                                               
  [Sylius\Component\Channel\Context\ChannelNotFoundException]  
  Channel could not be found!                                  
                                                               


Exception trace:
 () at /var/www/reissSylius/src/Sylius/Component/Channel/Context/CompositeChannelContext.php:53
 Sylius\Component\Channel\Context\CompositeChannelContext->getChannel() at /var/www/reissSylius/src/Sylius/Component/Channel/Context/CachedPerRequestChannelContext.php:57
 Sylius\Component\Channel\Context\CachedPerRequestChannelContext->getChannel() at /var/www/reissSylius/src/Sylius/Bundle/CoreBundle/EventListener/OrderChannelListener.php:53
 Sylius\Bundle\CoreBundle\EventListener\OrderChannelListener->processOrderChannel() at n/a:n/a
 call_user_func() at /var/www/reissSylius/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/Debug/WrappedListener.php:61
 Symfony\Component\EventDispatcher\Debug\WrappedListener->__invoke() at n/a:n/a
 call_user_func() at /var/www/reissSylius/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php:158
 Symfony\Component\EventDispatcher\EventDispatcher->doDispatch() at /var/www/reissSylius/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php:46
 Symfony\Component\EventDispatcher\EventDispatcher->dispatch() at /var/www/reissSylius/vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/Debug/TraceableEventDispatcher.php:124
 Symfony\Component\EventDispatcher\Debug\TraceableEventDispatcher->dispatch() at /var/www/reissSylius/src/Sylius/Bundle/FixturesBundle/DataFixtures/ORM/LoadOrdersData.php:142
[...]

PR #4095, opened 2016-02-10
[Fixture] Quick fixes for channel not found bug

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Sylius/Sylius/issues/4094
| License       | MIT
| Doc PR        |

After last updates, problem with fixtures occurred, with ``ChannelNotFoundException``. There were two sources of this evil:
- ``OrderChannelListener`` which got channel from ``ChannelContext`` (which led to exception) and set it on order
- ``PromotionProcessor`` which found active promotions by channel (the same case as above)

Quick fixes for these problems are:
- set channel on ``Order`` in ``OrderChannelListener`` only if it is not set yet (we're setting channel on each order in ``LoadOrdersData``
- pass order's channel to ``findActiveByChannel`` repository method, rather than this from ``ChannelContext``

Of course, these are a little bit workarounds, but fortunately there is huge checkout and promotion's refactoring upcoming, which should kill these nasty bugs forever ;)

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/EventListener/OrderChannelListener.php
+++ b/src/Sylius/Bundle/CoreBundle/EventListener/OrderChannelListener.php
@@ -50,6 +50,8 @@ class OrderChannelListener
             );
         }
 
-        $order->setChannel($this->channelContext->getChannel());
+        if (null === $order->getChannel()) {
+            $order->setChannel($this->channelContext->getChannel());
+        }
     }
 }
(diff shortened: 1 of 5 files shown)


michalmarcinkowski:
Good job! :+1:

pjedrzejewski:
Good to go when green.

michalmarcinkowski:
That was a really quick fix. Thanks Mateusz!

pjedrzejewski:
During review I noticed another issue, reported in https://github.com/Sylius/Sylius/issues/4097.

michalmarcinkowski:
@pjedrzejewski good catch!

PR #4100, opened 2016-02-10
[Sylius/Behat] SharedStorage refactored without overriding protection

pamil:
After a conversation with @Zales0123 and @Arminek a few days ago we came to conclusion that overriding protection in `SharedStorage` is useless for now and causes unnecessary complexity.


pjedrzejewski:
:+1:

michalmarcinkowski:
Let's bring it back only when we find a real use case for that :)

Issue #4083, opened 2016-02-09
[RFC] Rename Shipment.items => Shipment.units?


peteward:
Since in full-stack Sylius a Shipment.item is an OrderItemUnit (not OrderItem) this could get confusing very quickly.

wdyt?

pjedrzejewski:
@peteward Sounds reasonable to me, as it tracks individual units anyway.

michalmarcinkowski:
Thanks for your contribution! :+1:

PR #4102, opened 2016-02-10
[Shipping] Rename ShipmentItem to ShipmentUnit

peteward:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #4083
| License       | MIT

Due to the introduction of `OrderItemUnit`, it is confusing to have `ShippingItem` referring to `OrderItemUnit` when there is also `OrderItem`.

This updates the naming for consistency.

### DIFF
--- a/src/Sylius/Bundle/ShippingBundle/spec/Form/Type/RuleChoiceTypeSpec.php
+++ b/src/Sylius/Bundle/ShippingBundle/spec/Form/Type/RuleChoiceTypeSpec.php
@@ -22,8 +22,8 @@ use Symfony\Component\OptionsResolver\OptionsResolver;
 class RuleChoiceTypeSpec extends ObjectBehavior
 {
     private $choices = [
-        RuleInterface::TYPE_ITEM_TOTAL => 'Order total',
-        RuleInterface::TYPE_ITEM_COUNT => 'Order items count',
+        RuleInterface::TYPE_UNIT_TOTAL => 'Order total',
+        RuleInterface::TYPE_UNIT_COUNT => 'Order units count',
     ];
 
     function let()
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -242,6 +242,34 @@ class BookController extends ResourceController
 * Introduced ``OrderItemQuantityDataMapper``, which attached to ``OrderItemType`` uses proper service to modify ``OrderItem`` quantity;
 * Changed ``Adjustment`` ``description`` field to ``label``;
 
+### Shipping and ShippingBundle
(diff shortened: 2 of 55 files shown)


pjedrzejewski:
Nice, could you add quick not to UPGRADE file? To use ``sylius.*.shipment_unit`` instead of ``sylius.*.shipment_item``, etc. ? Thanks!

peteward:
done

peteward:
hmmm not sure about that fail - any ideas?

pjedrzejewski:
CS check is failing cause use statements are not sorted. I think we need to move this check out of travis or use something like StyleCI. I will have a look.

peteward:
ok I'll sort and update

src/Sylius/Bundle/ShippingBundle/spec/Form/Type/RuleChoiceTypeSpec.php
```diff
@@ -22,8 +22,8 @@
 class RuleChoiceTypeSpec extends ObjectBehavior
 {
     private $choices = [
-        RuleInterface::TYPE_ITEM_TOTAL => 'Order total',
-        RuleInterface::TYPE_ITEM_COUNT => 'Order items count',
+        RuleInterface::TYPE_UNIT_TOTAL => 'Order total',
```
pjedrzejewski: I am not sure about it. This should be probably renamed to ITEMS_TOTAL, no? Units in that context can be confusing?

peteward: Yeah I was not sure about this, for both the TOTAL and the COUNT. But fundamentally it _is_ units. The TYPE_UNIT_TOTAL rule doesn't appear to have any use, but TYPE_UNIT_COUNT certainly does. If you look down at the `PerUnitRateCalculator` `calculate()` method:

```
return (int) ($configuration['amount'] * $subject->getShippingUnitCount());
```

Previously this was `getShippingItemCount()` but it was of course getting the count of `OrderItemUnits` so effectively this is correct to call it `TYPE_UNIT_COUNT`.

I was thinking that you might want to add a CoreBundle extra rule which is TYPE_ITEM_COUNT and works on an OrderItem basis, but for now this is at least accurate to what is being counted?

peteward:
ping @pjedrzejewski

pjedrzejewski:
Thank you Peter! :+1:

PR #4111, opened 2016-02-11
[Sylius/Behat] Removing page objects in favour of plain old Symfony services

pamil:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes (within master)
| Deprecations? | no
| Fixed tickets | #X
| License       | MIT
| Doc PR        | Sylius/Sylius-Docs#414


pamil:
In fact it could be divided into two extensions, one for defining contexts as services and the second one for supporting multiple containers, but I guess it's worth another iteration.

src/Sylius/Behat/MultiContainerExtension/Context/Environment/Handler/ContextServiceEnvironmentHandler.php
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @return string[]
+     *
+     * @throws SuiteConfigurationException If `contexts_as_services` setting is not an array
+     */
+    private function getSuiteContexts(Suite $suite)
+    {
+        if (!is_array($suite->getSetting('contexts_as_services'))) {
+            throw new SuiteConfigurationException(
+                sprintf('`contexts` setting of the "%s" suite is expected to be an array, %s given.',
```
pjedrzejewski: contexts_as_services

src/Sylius/Behat/Page/ElementNotFoundException.php
```diff
@@ -0,0 +1,19 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Behat\Page;
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+final class ElementNotFoundException extends \Exception
```
michalmarcinkowski: No need to declare it final. It's handy to allow creating e.g. exception `DeleteButtonNotFoundException extends ElementNotFoundException`

pamil: :+1:, but since it's not a critical feature, I think it can be merged as is and fixed just after merge, I don't want to wait with this PR for another build to complete as it touches the part that is actively developed right now.

michalmarcinkowski: Agree :+1:

pjedrzejewski:
Crazy work Kamil, thank you!

PR #4112, opened 2016-02-11
[Behat] Added default customer step

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

Here are some changes, we've been talking about [here](https://github.com/Sylius/Sylius/pull/4044#issuecomment-182392303). Main thing, is introducing ``I am logged in Customer`` step, for scenarios that don't need specific information about Customer.

### DIFF
--- a/features/shop/apply_correct_taxes_based_on_customer_data.feature
+++ b/features/shop/apply_correct_taxes_based_on_customer_data.feature
@@ -29,3 +29,11 @@ Feature: Apply correct taxes based on customer data
         And I proceed without selecting shipping address
         Then my cart total should be "€110.00"
         And my cart taxes should be "€10.00"
+
+    Scenario: Proper taxes for logged in Customer with already specified shipping address
+        Given I am logged in customer
+        And my default shipping address is "Australia"
+        When I add product "PHP T-Shirt" to the cart
+        And I proceed without selecting shipping address
+        Then my cart total should be "€110.00"
+        And my cart taxes should be "€10.00"
--- a/src/Sylius/Behat/Context/Setup/SecurityContext.php
+++ b/src/Sylius/Behat/Context/Setup/SecurityContext.php
@@ -16,6 +16,7 @@ use Behat\Mink\Driver\Selenium2Driver;
 use Behat\Mink\Session;
 use Sylius\Behat\Page\Shop\HomePage;
 use Sylius\Bundle\CoreBundle\Test\Services\SecurityServiceInterface;
+use Sylius\Component\Core\Test\Services\SharedStorageInterface;
 
 /**
  * @author Arkadiusz Krakowiak <arkadiusz.krakowiak@lakion.com>
@@ -28,11 +29,6 @@ final class SecurityContext implements Context
     private $securityService;
 
     /**
-     * @var HomePage
-     */
-    private $homePage;
-
-    /**
      * @var Session
      */
     private $minkSession;
@@ -43,17 +39,34 @@ final class SecurityContext implements Context
     private $minkParameters;
 
     /**
+     * @var HomePage
+     */
+    private $homePage;
+
+    /**
+     * @var SharedStorageInterface
+     */
+    private $sharedStorage;
+
+    /**
      * @param SecurityServiceInterface $securityService
      * @param Session $minkSession
      * @param array $minkParameters
      * @param HomePage $homePage
+     * @param SharedStorageInterface $sharedStorage
      */
-    public function __construct(SecurityServiceInterface $securityService, Session $minkSession, array $minkParameters, HomePage $homePage)
-    {
+    public function __construct(
+        SecurityServiceInterface $securityService,
+        Session $minkSession,
+        array $minkParameters,
+        HomePage $homePage,
+        SharedStorageInterface $sharedStorage
+    ) {
         $this->securityService = $securityService;
         $this->minkSession = $minkSession;
         $this->minkParameters = $minkParameters;
         $this->homePage = $homePage;
+        $this->sharedStorage = $sharedStorage;
     }
 
     /**
@@ -63,6 +76,16 @@ final class SecurityContext implements Context
     {
         $this->prepareSessionIfNeeded();
         $this->securityService->logIn($email, 'main', $this->minkSession);
+    }
+
+    /**
+     * @Given /^I am logged in customer$/
+     */
+    public function iAmLoggedInCustomer()
+    {
+        $user = $this->securityService->logInDefaultUser($this->minkSession);
+
+        $this->sharedStorage->setCurrentResource('user', $user);
     }
 
     private function prepareSessionIfNeeded()
--- a/src/Sylius/Behat/Context/Setup/SecurityContext.php
+++ b/src/Sylius/Behat/Context/Setup/SecurityContext.php
@@ -15,8 +15,10 @@ use Behat\Behat\Context\Context;
 use Behat\Mink\Driver\Selenium2Driver;
 use Behat\Mink\Session;
 use Sylius\Behat\Page\Shop\HomePage;
+use Sylius\Bundle\CoreBundle\Test\Factory\TestUserFactoryInterface;
 use Sylius\Bundle\CoreBundle\Test\Services\SecurityServiceInterface;
 use Sylius\Component\Core\Test\Services\SharedStorageInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
 
 /**
  * @author Arkadiusz Krakowiak <arkadiusz.krakowiak@lakion.com>
@@ -44,6 +46,16 @@ final class SecurityContext implements Context
     private $homePage;
 
     /**
+     * @var TestUserFactoryInterface
+     */
+    private $testUserFactory;
+
+    /**
+     * @var RepositoryInterface
+     */
+    private $userRepository;
+
+    /**
      * @var SharedStorageInterface
      */
     private $sharedStorage;
@@ -53,6 +65,8 @@ final class SecurityContext implements Context
      * @param Session $minkSession
      * @param array $minkParameters
      * @param HomePage $homePage
+     * @param TestUserFactoryInterface $testUserFactory
+     * @param RepositoryInterface $userRepository
      * @param SharedStorageInterface $sharedStorage
      */
     public function __construct(
@@ -60,12 +74,16 @@ final class SecurityContext implements Context
         Session $minkSession,
         array $minkParameters,
         HomePage $homePage,
+        TestUserFactoryInterface $testUserFactory,
+        RepositoryInterface $userRepository,
         SharedStorageInterface $sharedStorage
     ) {
         $this->securityService = $securityService;
         $this->minkSession = $minkSession;
         $this->minkParameters = $minkParameters;
         $this->homePage = $homePage;
+        $this->testUserFactory = $testUserFactory;
+        $this->userRepository = $userRepository;
         $this->sharedStorage = $sharedStorage;
     }
 
@@ -83,7 +101,10 @@ final class SecurityContext implements Context
      */
     public function iAmLoggedInCustomer()
     {
-        $user = $this->securityService->logInDefaultUser($this->minkSession);
+        $user = $this->testUserFactory->createDefault();
+        $this->userRepository->add($user);
+
+        $this->securityService->logIn($user->getEmail(), 'main', $this->minkSession);
 
         $this->sharedStorage->setCurrentResource('user', $user);
     }
--- a/src/Sylius/Behat/Context/Setup/SecurityContext.php
+++ b/src/Sylius/Behat/Context/Setup/SecurityContext.php
@@ -93,7 +93,7 @@ final class SecurityContext implements Context
     public function iAmLoggedInAs($email)
     {
         $this->prepareSessionIfNeeded();
-        $this->securityService->logIn($email, 'main', $this->minkSession);
+        $this->securityService->logIn($email, $this->minkSession, 'main');
     }
 
     /**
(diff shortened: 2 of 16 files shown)


features/shop/apply_correct_taxes_based_on_customer_data.feature
```diff
@@ -31,3 +31,11 @@ Feature: Apply correct taxes based on customer data
         And I proceed without selecting shipping address
         Then my cart total should be "€110.00"
         And my cart taxes should be "€10.00"
+
+    Scenario: Proper taxes for logged in Customer with already specified shipping address
+        Given I am logged in Customer
```
michalmarcinkowski: The `Customer` should start with lowercase letter `customer`

src/Sylius/Behat/Context/Setup/SecurityContext.php
```diff
@@ -64,6 +75,16 @@ public function iAmLoggedInAs($email)
     }
 
     /**
+     * @Given /^I am logged in customer$/
+     */
+    public function iAmLoggedInCustomer()
+    {
+        $user = $this->securityService->logInDefaultUser($this->mink->getSession());
```
michalmarcinkowski: `securityService` should not return `user`. What is more it should not create a new user (I suppose it is).

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactory.php
```diff
@@ -0,0 +1,80 @@
...
+    public function __construct(FactoryInterface $customerFactory, FactoryInterface $userFactory)
+    {
+        $this->customerFactory = $customerFactory;
+        $this->userFactory = $userFactory;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function create($firstName, $lastName, $email, $password)
```
michalmarcinkowski: How about `create($email, $password, $firstName = 'John', $lastName = 'Doe')`?

src/Sylius/Bundle/CoreBundle/Test/Services/SecurityService.php
```diff
@@ -45,14 +49,23 @@ public function __construct(
     /**
      * {@inheritdoc}
      */
-    public function logIn($email, $providerKey, Session $minkSession)
+    public function logIn($email, $providerKey = self::DEFAULT_PROVIDER_KEY, Session $minkSession)
```
michalmarcinkowski: Since `$providerKey` is optional it should be the last parameter.

Zales0123: :+1:

src/Sylius/Bundle/CoreBundle/Test/Services/SecurityServiceInterface.php
```diff
@@ -12,6 +12,7 @@
 namespace Sylius\Bundle\CoreBundle\Test\Services;
 
 use Behat\Mink\Session;
+use Sylius\Component\Core\Model\UserInterface;
```
michalmarcinkowski: Should be reverted.

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactory.php
pamil: We need a builder for user, I'm sure there will be `$enabled = true`, `$role = 'administrator'` and many more in the future.

src/Sylius/Bundle/CoreBundle/Test/Services/SecurityService.php
pamil: I'm curious what happens if we set `$providerKey` to something else, can you check if it is required?

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactory.php
michalmarcinkowski: Let's tackle this next time, but :+1:

src/Sylius/Bundle/CoreBundle/Test/Services/SecurityService.php
michalmarcinkowski: It is required. It is the name of the firewall [here](https://github.com/michalmarcinkowski/Sylius/blob/5009463405dee467e3503b210f3a30e274ec9414/app/config/security.yml#L51-L51).

pamil: In fact, at this moment is not needed as the token is already authenticated. We have other setting hardcoded, `_security_user` has a `user` as a context name used in our firewalls, which is processed by `ContextListener`. If provider key is anything but not null, everything goes all right.

michalmarcinkowski: :+1:

src/Sylius/Behat/Context/Setup/SecurityContext.php
```diff
@@ -28,32 +31,60 @@
...
     /**
-     * @var Session
+     * @var TestUserFactoryInterface
      */
-    private $minkSession;
+    private $testUserFactory;
 
     /**
-     * @var array
+     * @var RepositoryInterface
```
lchrusciel: User has custom repository interface. It would be better to use it instead of default one.

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactoryInterface.php
```diff
@@ -0,0 +1,38 @@
...
+namespace Sylius\Bundle\CoreBundle\Test\Factory;
+
+use Sylius\Component\Core\Model\UserInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface TestUserFactoryInterface
+{
+    const DEFAULT_USER_FIRST_NAME = 'John';
```
lchrusciel: Why do you define some variables in interface and other in a class implementation?

src/Sylius/Behat/Context/Setup/SecurityContext.php
Zales0123: In fact, why? ;) the only benefit, which we have from using ``UserRepositoryInterface`` is ``findOneByEmail`` method, which is not used in this class.

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactory.php
```diff
@@ -0,0 +1,78 @@
...
+
+use Sylius\Component\Resource\Factory\FactoryInterface;
+use Symfony\Component\Security\Core\User\UserInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class TestUserFactory implements TestUserFactoryInterface
+{
+    const DEFAULT_USER_EMAIL = 'john.doe@example.com';
```
lchrusciel: Do we use this values outside of this class? What do you think about making them static and private?

src/Sylius/Behat/Context/Setup/SecurityContext.php
lchrusciel: Are you sure, you wont need this method? I didn't check usage of this variable. :+1:

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactoryInterface.php
Zales0123: I think they all should be in class, thanks :+1:

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactory.php
Zales0123: In fact now, but I still think they can be pretty useful in the future and it doesn't make any difference currently is it ``const`` or ``private static`` variable ;)

src/Sylius/Behat/Context/Setup/SecurityContext.php
Zales0123: Yeah, I'm pretty sure it's not needed for now ;)

src/Sylius/Bundle/CoreBundle/Resources/config/test_services.xml
```diff
@@ -28,6 +29,10 @@
             <argument type="service" id="sylius.repository.user" />
             <argument type="service" id="session" />
         </service>
+        <service id="sylius.behat.user_factory" class="%sylius.behat.user_factory.class%">
```
pjedrzejewski: I think it should be ``sylius.test.factory.user``.

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactory.php
```diff
@@ -0,0 +1,80 @@
...
+        $customer->setFirstName($firstName);
+        $customer->setLastName($lastName);
+
+        $user = $this->userFactory->createNew();
+
+        $user->setCustomer($customer);
+        $user->setUsername($email);
+        $user->setEmail($email);
+        $user->setPlainPassword($password);
+        $user->enable();
```
pjedrzejewski: User is enabled by default, no need for that.

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactoryInterface.php
```diff
@@ -0,0 +1,35 @@
...
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\CoreBundle\Test\Factory;
```
pjedrzejewski: Should go to the component.

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactory.php
pamil: It looks like it isn't, `Sylius\Component\User\Model\User::__construct()` overrides the trait default value and sets enabled to false.

Zales0123: I just wanted to point the same thing ;) https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/User/Model/User.php#L123

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactoryInterface.php
Zales0123: :+1:

src/Sylius/Bundle/CoreBundle/Resources/config/test_services.xml
Zales0123: Ok, but then we should change convention in all ``test_services.xml`` file ;)

src/Sylius/Behat/Context/Setup/SecurityContext.php
```diff
@@ -62,7 +93,20 @@ public function __construct(SecurityServiceInterface $securityService, Session $
     public function iAmLoggedInAs($email)
     {
         $this->prepareSessionIfNeeded();
-        $this->securityService->logIn($email, 'main', $this->minkSession);
+        $this->securityService->logIn($email, $this->minkSession, 'main');
```
michalmarcinkowski: The last argument can be skipped.

src/Sylius/Behat/Context/Setup/SecurityContext.php
```diff
@@ -62,7 +93,20 @@ public function __construct(SecurityServiceInterface $securityService, Session $
...
+
+    /**
+     * @Given /^I am logged in customer$/
+     */
+    public function iAmLoggedInCustomer()
+    {
+        $user = $this->testUserFactory->createDefault();
+        $this->userRepository->add($user);
+
+        $this->securityService->logIn($user->getEmail(), $this->minkSession, 'main');
```
michalmarcinkowski: The last argument can be skipped.

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactory.php
lchrusciel: But you are not using them at the moment. According to YAGNI, I would vote to keep them private.   If one has to use them, he will change them to constants. WDYT?

Zales0123: Ok, I agree with this argumentation. @michalmarcinkowski @pjedrzejewski are you happy with ``private static`` variables for default user data? Or maybe we just should *for now* pass use regular strings, not variables?

src/Sylius/Behat/Context/Setup/SecurityContext.php
pamil: IMHO typehinting to `UserRepositoryInterface` would be wise, because user repository should always be implementing that interface and if we will need to use any of the user repository specific method in the future we won't have to do any irrelevant changes.

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactory.php
pjedrzejewski: As discussed offline, we go with consts. :)

src/Sylius/Bundle/CoreBundle/Test/Factory/TestUserFactory.php
pjedrzejewski: My bad! :)

pjedrzejewski:
Thanks Mateusz! :+1:

PR #4116, opened 2016-02-11
[Core][Promotion] Removed multiple promotion coupons on order

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

According to upcoming changes in promotion system and related with them cleanups, there are multiple coupons on Cart deleted.


pjedrzejewski:
Thanks Mateusz! :+1:

kayue:
Why doesn't allow multiple coupon per order?

michalmarcinkowski:
We're reworking the promotions system and multiple coupons introduces complexity that we would like to avoid in the first step. Perhaps we will bring this feature back in the new architecture, but dropping support for this for now to avoid bugs.

PR #4123, opened 2016-02-11
[Behat] Decouple checkout steps in context

michalmarcinkowski:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

* Do not enforce to always select payment method when proceeding with checkout since in most scenarios we do not need to
* Explicitly choose payment method when proceeding with checkout (right now the taxes to shipping fee are only applied after choosing payment method, but it should be applied at the moment the shipping method is chosen, so let's be explicitly aware of that)


pjedrzejewski:
Thanks Michał!

PR #4124, opened 2016-02-11
[product] - originalPrice allow null, fix error 500 on save

nakashu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT

Original price can be set to null with the setOriginalPrice() method.

Previously if you had original price set lets say to 100$, only way to reset/remove it was with filling in 0.
If you tried setting it with empty value, on saving the product an error 500 would be shown.
The error was thrown from the setter method.

### DIFF
--- a/src/Sylius/Component/Core/Model/ProductVariant.php
+++ b/src/Sylius/Component/Core/Model/ProductVariant.php
@@ -185,7 +185,7 @@ class ProductVariant extends BaseVariant implements ProductVariantInterface
      */
     public function setOriginalPrice($originalPrice)
     {
-        if (!is_int($originalPrice)) {
+        if (null !== $originalPrice && !is_int($originalPrice)) {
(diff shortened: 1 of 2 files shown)


src/Sylius/Component/Core/Model/ProductVariant.php
```diff
@@ -185,7 +185,7 @@ public function setPrice($price)
      */
     public function setOriginalPrice($originalPrice)
     {
-        if (!is_int($originalPrice)) {
+        if (!is_null($originalPrice) and !is_int($originalPrice)) {
```
michalmarcinkowski: `null !== $originalPrice && !is_int($originalPrice))`

michalmarcinkowski:
Thanks @nakashu!

PR #4129, opened 2016-02-12
[Travis] Removed CS checker from Travis build

pamil:
It's quite annoying when you need to push once again because empty line has 4 spaces instead of none, some original licensing header needs to be changed or coding standard is enforced even on copied files. It causes a lot of false-negatives.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ before_script:
 
 script:
     - composer validate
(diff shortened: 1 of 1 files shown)


stloyd:
:+1:

pjedrzejewski:
:+1:

bendavies:
:( haha

michalmarcinkowski:
We're going to setup StyleCI and configure auto-PR with CS fixes :)

bendavies:
:+1:

PR #4131, opened 2016-02-12
[Behat] Renamed ui-* suites to ui_*

pamil:
Running `bin/behat -s ui-cart` returns:

screen shot 2016-02-12 at 12 01 44


pjedrzejewski:
:+1:

Issue #4082, opened 2016-02-08
[Behat] Replace DefaultFranceChannelFactory with DefaultChannelFactory


michalmarcinkowski:
Current `DefaultFranceChannelFactory` creates single "France" channel and also "France" country and zone, but it shouldn't. The idea is to replace it with `DefaultChannelFactory` whose sole responsibility will be creating default channel. Then in the separate steps we will define zones and shipping countries (as we do already).

michalmarcinkowski:
After giving it some thoughts together with @pjedrzejewski we agreed to have the following:
 * `Given the store operates on a single channel` - this step will only create an 'empty' channel
 * `Given the store operates on a single channel in "France"` - this step will create a preconfigured channel for France, it means:
  * Set EUR as default currency
  * Create "France" country and zone

This will reduce the backgrounds for scenarios where this details are not necessary.

pjedrzejewski:
:+1:

PR #4138, opened 2016-02-12
[Behat] Rework default channel factories

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Sylius/Sylius/issues/4082
| License       | MIT
| Doc PR        |

According to mentioned issue, there are now two default channel factories:
- ``DefaultChannelFactory`` creating only *empty* **Default** channel
- ``DefaultFranceChannelFactory`` creating **France** channel, with **EUR** as default currency, **France** shipping country and proper zone with zone member


etc/behat/suites/ui_cart.yml
```diff
@@ -9,8 +9,9 @@ default:
                     entityManager: "@doctrine.orm.entity_manager"
 
                 - Sylius\Behat\Context\Setup\ChannelContext:
+                    defaultFranceChannelFactory: "@sylius.behat.factory.default_france_channel"
+                    defaultChannelFactory: "@sylius.behat.factory.default_channel"
                     sharedStorage: "@sylius.behat.shared_storage"
```
pamil: If some context uses `SharedStorage`, can we pass it as the first argument?

src/Sylius/Behat/Context/Setup/ChannelContext.php
```diff
@@ -22,25 +23,42 @@
 final class ChannelContext implements Context
 {
     /**
-     * @var SharedStorageInterface
+     * @var DefaultStoreDataInterface
```
pamil: Why these two factories have different interfaces?

src/Sylius/Behat/Context/Setup/ChannelContext.php
```diff
@@ -22,25 +23,42 @@
...
      */
-    private $sharedStorage;
+    private $defaultFranceChannelFactory;
```
pamil: IMHO `$franceChannelFactory` whould be more suitable.

src/Sylius/Component/Core/Test/Services/DefaultFranceChannelFactory.php
```diff
@@ -109,6 +154,29 @@ private function createChannel()
...
+
+    /**
+     * @return CurrencyInterface
+     */
+    private function createCurrency()
+    {
+        $currency = $this->currencyFactory->createNew();
+        $currency->setCode(self::DEFAULT_CURRENCY_CODE);
+        $currency->setExchangeRate(1.00);
+
```
michalmarcinkowski: Add also `$currency->setBase(true);` ;)

src/Sylius/Component/Core/Test/Services/DefaultChannelFactoryInterface.php
```diff
@@ -0,0 +1,25 @@
...
+
+use Sylius\Component\Channel\Repository\ChannelRepositoryInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface DefaultChannelFactoryInterface
+{
+    /**
+     * @return ChannelRepositoryInterface
```
pamil: ?

src/Sylius/Component/Core/Test/Services/DefaultChannelFactory.php
```diff
@@ -0,0 +1,57 @@
...
+     * {@inheritdoc}
+     */
+    public function create()
+    {
+        $channel = $this->channelFactory->createNamed(self::DEFAULT_CHANNEL_NAME);
+        $channel->setCode(self::DEFAULT_CHANNEL_CODE);
+
+        $this->channelRepository->add($channel);
+
+        return ['channel' => $channel];
```
pamil: Why can't we return a Channel instance?

src/Sylius/Component/Core/Test/Services/DefaultFranceChannelFactory.php
```diff
@@ -24,6 +26,8 @@
 class DefaultFranceChannelFactory implements DefaultStoreDataInterface
```
pamil: `implements DefaultChannelFactoryInterface` instead?

pamil:
As you're messing with channel factories, I have a dream... that `DefaultFranceChannelFactory` returns `ChannelInterface` instance, contexts using it sets the shared storage variables by using some getters on returned channel and `SharedStorage::setClipboard()` is removed :)

src/Sylius/Component/Core/Test/Services/DefaultChannelFactory.php
Zales0123: If we want to keep unified interface for ``DefaultFranceChannelFactory`` and ``DefaultChannelFactory``, we need to return array as the result of function is used as ``setClipboard`` parameter.

Zales0123:
Sure, ``setClipboard()`` is pretty lame, as it persuade factory to return ``array`` (in spite of being called e.g. "Default**Channel**Factory"). I think it should be deeply considered and reworked in the nearest future. For now, we can do that, as not all objects returned by ``DefaultFranceChannelFactory`` is property of channel (``zone``, for example) ;/

src/Sylius/Component/Core/Test/Services/DefaultFranceChannelFactory.php
pjedrzejewski: :+1:

Zales0123: Already changes, I don't know why github haven't marked it as outdated :|

pjedrzejewski:
Thank you Mateusz!

PR #4142, opened 2016-02-14
[RFC][Association] Introduce AssociableTrait with ability to get associated objects by association type

aramalipoor:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | no |
| License | MIT |
| Doc PR | - |

I've extracted some code related to association from Product, Idk why it wasn't done and I couldn't figure out why so do you think this is good? And why `AssociableInterface` does not have method signatures such as `addAssociation`, `removeAssociation`, etc? Any specific reason? 

Also having `getAssociatedObjects($type = null)` method is useful in twig.

``` twig
{% for relatedProduct in product.associatedObjects('RELATED_PRODUCTS') %}
```

/cc @lchrusciel

### DIFF
--- a/src/Sylius/Component/Association/Traits/AssociableTrait.php
+++ b/src/Sylius/Component/Association/Traits/AssociableTrait.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Sylius\Component\Association\Traits;
+
+use Doctrine\Common\Collections\Collection;
+use Sylius\Component\Association\Model\AssociationInterface;
+use Sylius\Component\Association\Model\AssociationTypeInterface;
+
+trait AssociableTrait
+{
+    /**
+     * @var Collection|AssociationInterface[]
+     */
+    protected $associations;
+
+    /**
+     * @return Collection|AssociationInterface[]
+     */
+    public function getAssociations()
+    {
+        return $this->associations;
+    }
+
+    /**
+     * @param null|string|AssociationTypeInterface $type
+     *
+     * @return array|AssociationInterface[]
+     */
+    public function getAssociatedObjects($type = null)
+    {
+        if (null === $type) {
+            return $this->associations;
+        }
+
+        if ($type instanceof AssociationTypeInterface) {
+            $type = $type->getCode();
+        }
+
+        $associatedObjects = [];
+
+        $this->associations->forAll(function (AssociationInterface $association) use (&$associatedObjects, $type) {
+            if ($association->getType()->getCode() === $type) {
+                return;
+            }
+
+            $associatedObjects = array_merge($associatedObjects, $association->getAssociatedObjects());
+        });
+
+        return $associatedObjects;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function addAssociation(AssociationInterface $association)
+    {
+        if (!$this->hasAssociation($association)) {
+            $this->associations->add($association);
+            $association->setOwner($this);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function removeAssociation(AssociationInterface $association)
+    {
+        if ($this->hasAssociation($association)) {
+            $association->setOwner(null);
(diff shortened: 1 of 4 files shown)


peteward:
looks good :+1:

src/Sylius/Component/Product/Model/ProductInterface.php
```diff
@@ -28,8 +29,7 @@
     SoftDeletableInterface,
     TimestampableInterface,
     ToggleableInterface,
-    ProductTranslationInterface,
-    AssociableInterface
```
lchrusciel: You should leave it here, because ProductInterface should force user to implement associations methods.

src/Sylius/Component/Product/Model/ProductInterface.php
```diff
@@ -13,6 +13,7 @@
 
 use Sylius\Component\Archetype\Model\ArchetypeSubjectInterface;
 use Sylius\Component\Association\Model\AssociableInterface;
+use Sylius\Component\Association\Model\AssociationTypeInterface;
```
lchrusciel: Redundant use statement.

src/Sylius/Component/Product/Model/Product.php
```diff
@@ -26,9 +28,9 @@
  * @author Paweł Jędrzejewski <pawel@sylius.org>
  * @author Gonzalo Vilaseca <gvilaseca@reiss.co.uk>
  */
-class Product extends AbstractTranslatable implements ProductInterface
+class Product extends AbstractTranslatable implements ProductInterface, AssociableInterface
```
lchrusciel: AssociableInterface should be extended in a product interface.

src/Sylius/Component/Association/Traits/AssociableTrait.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function hasAssociation(AssociationInterface $association)
+    {
+        return $this->associations->contains($association);
+    }
+}
```
lchrusciel: Missing blank line

lchrusciel:
I'm not convinced about using traits. You are correct, that it will reduce code duplication, but on the other hand it will hide complexity of class. At the end, you have to initialize collection in a constructor, but you don't have a associations field visible on a class. And this code is not duplicated anywhere in a whole Sylius. 
Another thing is, that association bundle is pretty generic. So you are able to have multiple association in a one project to same class. [Removing these lines](https://github.com/Sylius/Sylius/pull/4142/files#diff-47887d888eb1c11de60fa3230c7399e8L62) will force us to have only one association on a product. 
Anyway your changes require also to delete this [class](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Product/Model/ProductAssociationInterface.php). It is a dead code, after your changes. 
To summarize, I am not against your changes. Im just not convinced. Maybe flexibility of multiple association is really useless and it will be easier to implement such trait. Lets here some other contributors. 

@aramalipoor can you tag this PR with RFC?

aramalipoor:
I have the same feeling too about this trait, and the thing you say about multiple associations, I cannot think of a real-life example because distinction is somehow done with association *types*.

What if we name this trait `SingularAssociableTrait` or something similar indicating this is only useful where you need a single associations relation?

By removing `ProductAssociationInterface` you mean removing ProductAssociation resource completely? Or just use the generic interface?

Another way that we can do instead of `getAssociatedObjects` (which is mostly useful in templating) is to write a twig extension like `sylius_associations(product, 'RELATED_PRODUCTS')`, WDYT?

aramalipoor:
@lchrusciel @pjedrzejewski ping ;)

michalmarcinkowski:
AssociationBundle was merged to ProductBundle.

Issue #2678, opened 2015-04-22
Possible to return a more meaningful payment error message?


kayue:
Right now Sylius only able to tell whether the payment is successful or not. (by looking at payment state)

Is it possible to return a more meaningful error message back to the user?

/cc @makasim

makasim:
There are two solutions I have in mind:

* Extend current GetStatusInterface with some method for it.
* Create a separate request (GetErrorMessage).

I personally prefer the second one.

kayue:
@makasim Thanks for the feedback. For the 2nd solution, I think you meant a request for extracting error message from payment details. This solution assume all information is available in payment details. Did I get it right?

makasim:
Yes sure,

PR #4150, opened 2016-02-15
[RFC][Core] Displaying payment error message when payment is failed.

ahmadrabie:
Q  | A
------------- | -------------
Bug fix?  | no
New feature?  | yes
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

fixes: #2678

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/StateMachineCallback/FailPaymentCallback.php
+++ b/src/Sylius/Bundle/CoreBundle/StateMachineCallback/FailPaymentCallback.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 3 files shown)


pjedrzejewski:
Sounds good although I am not sure it is responsibility of state machine to display flashes... We should think more about it.

ahmadrabie:
I agree and I have the same feeling but unfortunately I couldn't find a better solution, maybe @makasim can help here!

src/Sylius/Bundle/CoreBundle/StateMachineCallback/FailPaymentCallback.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $this->session = $session;
+        $this->translator = $translator;
+    }
+
+    public function addFailFlash(PaymentInterface $payment)
+    {
+
+        $details = $payment->getDetails();
+        if (isset($details['error']) && !empty($details['error'])) {
```
makasim: This is not correct. Payment details contains details in a payment gateway specific format, and hence they are different for paypal and stripe for example and may not have `error` key.

ahmadrabie:
:+1:

PR #4151, opened 2016-02-15
[Travis] Added default Github Oauth Token for Composer

pamil:
No more failing PRs caused by not being authenticated by Github. Please do not use this key in other projects than Sylius. Also updates `lakion/mink-debug-extension` version to `^1.0.3`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,12 +39,11 @@ before_install:
 
     - composer self-update
 
-    # Add Github OAuth token to Composer config if that variable exists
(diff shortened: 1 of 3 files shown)


gabiudrescu:
why don't you configure it as a secret env var in Travis?

pamil:
There is no way to pass secret variable to pull requests and it makes updating dependencies nearly impossible as the build fails on downloading packages from Github.

pjedrzejewski:
Thanks Kamil!

Issue #4149, opened 2016-02-14
[RFC] Behat tags


pjedrzejewski:
I just noticed that we are using tags like ``ui-cart`` - https://github.com/Sylius/Sylius/blob/master/features/shop/apply_correct_tax_for_items_with_the_same_tax_rate.feature#L1.

My initial idea with these tags was to combine them.

```
@cart @ui @domain
Feature: XYZ

@taxation
Feature: ABC

@ui @domain
Scenario: A

@ui
Scenario: B

@domain @api
Scenario: C
```

https://github.com/Sylius/Sylius/blob/master/etc/behat/suites/ui_cart.yml#L26 This could be changed to:

```
tags: @cart && @ui
```

This has following benefits:

* We can easily mark whole features as "cart" or "checkout" suite, etc.
* We can easily make specific scenarios run through UI and Domain, without repeating the suite name

I am pretty sure Behat supports that.

michalmarcinkowski:
:+1: realized the same yesterday! :smile:

pamil:
:+1:, this way we can run `bin/behat --tags cart` to run both UI and domain tests for cart.

PR #4159, opened 2016-02-15
[Behat] Retag behat features

lchrusciel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #4149

### DIFF
--- a/features/legacy/backend/metadata/metadata_customization.feature
+++ b/features/legacy/backend/metadata/metadata_customization.feature
@@ -1,4 +1,4 @@
-@metadata @ui
+@legacy_metadata
 Feature: Metadata customization
     In order to customize metadata
     As a store owner
--- a/features/shop/round_taxes_on_order_item_level.feature
+++ b/features/shop/round_taxes_on_order_item_level.feature
@@ -1,4 +1,4 @@
-@ui-cart
+@cart
 Feature: Round taxes on order item level
     In order to avoid taxes amount inaccuracy
     As a Visitor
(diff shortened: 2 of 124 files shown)


features/shop/round_taxes_on_order_item_level.feature
```diff
@@ -1,4 +1,4 @@
-@ui-cart
+@ui @cart
```
pjedrzejewski: I think it should be the following order ``@cart @domain @api @ui``. Thoughts?

pamil: `@cart` should be the first one, I'm not sure about the other ones order.

michalmarcinkowski: :+1:

lchrusciel: We will implement *ui* before *api* probably, so I would vote to swap them. Rest is :+1:

pjedrzejewski: Yeah, let's do this.

lchrusciel: :+1: done

features/legacy/backend/metadata/metadata_customization.feature
```diff
@@ -1,4 +1,4 @@
-@metadata @ui
+@legacy_metadata @ui
```
michalmarcinkowski: Why do we need `ui` tag here? /cc @pamil

pamil: It's not needed, can be deleted.

lchrusciel: :+1:

PR #4174, opened 2016-02-16
[Behat] MCE: Friendly exception message if context is not tagged as context

pamil:


src/Sylius/Behat/MultiContainerExtension/ContextRegistry.php
```diff
@@ -40,7 +40,11 @@ public function add($serviceId, $serviceClass)
     public function getClass($serviceId)
     {
         if (!isset($this->registry[$serviceId])) {
-            throw new \InvalidArgumentException();
+            throw new \InvalidArgumentException(sprintf(
+                'Could not find class for service with id "%s". Make sure you tagged the context with "%s".',
+                $serviceId,
+                'sylus.behat.context'
```
pjedrzejewski: What's the point of sprintfing it? :D

pamil: I like it separated. Anyway, there's a typo! :D

pjedrzejewski: Well, not a big deal for me, but still I think it is pointless. Not really more readable. I always assume that sprintf arguments are something dynamic.

michalmarcinkowski:
:+1:

PR #4177, opened 2016-02-16
Catch the ChannelNotFoundException in WDT

pjedrzejewski:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Sylius/Bundle/ChannelBundle/Context/FakeChannel/FakeChannelCollector.php
+++ b/src/Sylius/Bundle/ChannelBundle/Context/FakeChannel/FakeChannelCollector.php
@@ -12,6 +12,7 @@
 namespace Sylius\Bundle\ChannelBundle\Context\FakeChannel;
 
 use Sylius\Component\Channel\Context\ChannelContextInterface;
+use Sylius\Component\Channel\Context\ChannelNotFoundException;
 use Sylius\Component\Channel\Model\ChannelInterface;
 use Sylius\Component\Channel\Repository\ChannelRepositoryInterface;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ChannelBundle/Resources/views/FakeChannelCollector/layout.html.twig
```diff
@@ -4,8 +4,8 @@
     {% set icon %}
         <a href="#">
             {# fake image span #}<span style="width: 0; height: 28px; vertical-align: middle;"></span>
-            <span class="sf-toolbar-status">{{ collector.channels|length }}</span> channels,
-            <b>{{ collector.currentChannel.name }}</b>
+            <span class="sf-toolbar-status">{{ collector.channels|length }}</span> channel(s),
+            <b>{{ collector.currentChannel.name|default('Undefined') }}</b>
```
pamil: Will it work as expected? `collector.currentChannel` may be null, won't `collector.currentChannel.name` throw an error then?

pamil: Anyway, we may save only `current_channel_name` instead of `current_channel` in WDT.

pjedrzejewski: I checked, it works. ;) I'd leave the channel, it gives us more info and this tab of WDT will be improved with more info I think.

pamil: Weird Twig magic :tada:

src/Sylius/Bundle/ChannelBundle/Context/FakeChannel/FakeChannelCollector.php
```diff
@@ -30,7 +31,12 @@
     public function __construct(ChannelRepositoryInterface $channelRepository, ChannelContextInterface $channelContext)
     {
         $this->data['channels'] = $channelRepository->findAll();
-        $this->data['current_channel'] = $channelContext->getChannel();
+
+        try {
+            $this->data['current_channel'] = $channelContext->getChannel();
+        } catch (ChannelNotFoundException $exception) {
+            $this->data['current_channel'] = null;
```
stloyd: Other option instead of using `default()` filter in Twig, would be to set "fake" channel here, WDYT?

michalmarcinkowski:
:+1:

PR #4170, opened 2016-02-16
[Behat][Promotions] Promotions with fixed discount scenarios

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

First portion of new scenarios for *Promotion*, describing simple fixed discount on customer's cart.
It uses "@todo" tag introduced by @pamil in https://github.com/Sylius/Sylius/pull/4169.

PR #4199, opened 2016-02-17
[Behat][Promotion] Fixed discount promotion scenarios

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

Based on https://github.com/Sylius/Sylius/pull/4170
First portion of new promotion scenarios, for very simple cases - receiving fixed discount without any rules and receiving discount equal to cart items total.

### DIFF
--- a/src/Sylius/Behat/Context/Setup/PromotionContext.php
+++ b/src/Sylius/Behat/Context/Setup/PromotionContext.php
@@ -79,18 +79,18 @@ final class PromotionContext implements Context
     public function thereIsPromotion($promotionName)
     {
         $promotion = $this->testPromotionFactory->createPromotion($promotionName);
-        $promotion->addChannel($this->sharedStorage->getCurrentResource('channel'));
+        $promotion->addChannel($this->sharedStorage->get('channel'));
 
         $this->promotionRepository->add($promotion);
-        $this->sharedStorage->setCurrentResource('promotion', $promotion);
+        $this->sharedStorage->set('promotion', $promotion);
     }
 
     /**
      * @Given /^it gives "(?:€|£|\$)([^"]+)" fixed discount to every order$/
      */
-    public function itGivesFixedDiscountForCustomersWithCartsAbove($amount)
+    public function itGivesFixedDiscountToEveryOrder($amount)
     {
-        $currentPromotion = $this->sharedStorage->getCurrentResource('promotion');
+        $currentPromotion = $this->sharedStorage->get('promotion');
 
         $action = $this->testPromotionFactory->createFixedDiscountAction($amount, $currentPromotion);
         $this->actionRepository->add($action);
--- a/src/Sylius/Behat/Context/Setup/PromotionContext.php
+++ b/src/Sylius/Behat/Context/Setup/PromotionContext.php
@@ -79,8 +79,9 @@ final class PromotionContext implements Context
      */
     public function thereIsPromotion($promotionName)
     {
-        $promotion = $this->testPromotionFactory->create($promotionName);
-        $promotion->addChannel($this->sharedStorage->get('channel'));
+        $promotion = $this->testPromotionFactory
+            ->createForChannel($promotionName, $this->sharedStorage->get('channel'))
+        ;
 
         $this->promotionRepository->add($promotion);
         $this->sharedStorage->set('promotion', $promotion);
--- a/src/Sylius/Behat/Context/Setup/PromotionContext.php
+++ b/src/Sylius/Behat/Context/Setup/PromotionContext.php
@@ -1,0 +1,100 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Behat\Context\Setup;
+
+use Behat\Behat\Context\Context;
+use Doctrine\Common\Persistence\ObjectManager;
+use Sylius\Component\Core\Model\PromotionInterface;
+use Sylius\Component\Core\Test\Factory\TestPromotionFactoryInterface;
+use Sylius\Component\Core\Test\Services\SharedStorageInterface;
+use Sylius\Component\Promotion\Model\ActionInterface;
+use Sylius\Component\Promotion\Model\RuleInterface;
+use Sylius\Component\Promotion\Repository\PromotionRepositoryInterface;
+use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+final class PromotionContext implements Context
+{
+    /**
+     * @var SharedStorageInterface
+     */
+    private $sharedStorage;
+
+    /**
+     * @var PromotionRepositoryInterface
+     */
+    private $promotionRepository;
+
+    /**
+     * @var RepositoryInterface
+     */
+    private $actionRepository;
+
+    /**
+     * @var TestPromotionFactoryInterface
+     */
+    private $testPromotionFactory;
+
+    /**
+     * @var ObjectManager
+     */
+    private $objectManager;
+
+    /**
+     * @param SharedStorageInterface $sharedStorage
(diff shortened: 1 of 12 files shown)


src/Sylius/Behat/spec/Context/Setup/PromotionContextSpec.php
```diff
@@ -0,0 +1,92 @@
...
+        $promotionRepository->add($promotion)->shouldBeCalled();
+        $sharedStorage->setCurrentResource('promotion', $promotion)->shouldBeCalled();
+
+        $this->thereIsPromotion('Super promotion');
+    }
+
+    function it_creates_fixed_discount_action_for_promotion(
+        $actionRepository,
+        $objectManager,
+        $sharedStorage,
```
pamil: Can we have the same order of arguments as in `let`?

src/Sylius/Component/Core/Test/Factory/TestPromotionFactoryInterface.php
```diff
@@ -0,0 +1,37 @@
...
+     *
+     * @return PromotionInterface
+     */
+    public function createPromotion($name);
+
+    /**
+     * @param string $discount
+     * @param PromotionInterface $promotion
+     *
+     * @return ActionInterface
```
pamil: Why does promotion factory creates an promotion action?

Zales0123: This is a good question, *de facto* about interface/class name.
My point was, that I don't see a need of separate test factories for promotions **and** actions **and** rules. They are highly connected with each other and IMO they should be created in the same service.
So the problem is - how such a service should be named? :smile:

src/Sylius/Behat/spec/Context/Setup/PromotionContextSpec.php
Zales0123: :+1:

src/Sylius/Behat/Context/Setup/PromotionContext.php
```diff
@@ -0,0 +1,100 @@
...
+     * @Given /^it gives "(?:€|£|\$)([^"]+)" fixed discount to every order$/
+     */
+    public function itGivesFixedDiscountToEveryOrder($amount)
+    {
+        $currentPromotion = $this->sharedStorage->get('promotion');
+
+        $action = $this->testPromotionFactory->createFixedDiscountAction($amount, $currentPromotion);
+        $this->actionRepository->add($action);
+
+        $this->objectManager->flush();
```
michalmarcinkowski: No need to `flush` since `repository` on `add` does persist and flush :)

Zales0123:
Ok, after some brainstorm we have some conclusion, how **for now** it should be done :smile: 
- per-type methods should be implemented on custom action factory in *Core* (which will be done in separate PR)
- ``TestPromotionFactory`` should create only promotion and have no knowledge about actions
- ``PromotionContext`` should have ``testPromotionFactory``, ``promotionRepository`` and ``actionFactory`` (implemented in point 1)... and of course in the future custom ``ruleFactory``

We believe this concept is the best approach **at this moment**, but surely it's really big thing to discuss, how these services should be implemented in the perfect *Sylius* ;)

Zales0123:
**UPDATE:** After changing ``TestPromotionFactory`` logic slightly, this PR is now dependent on https://github.com/Sylius/Sylius/pull/4223, which should be first reviewed and merged ;)

src/Sylius/Behat/Context/Setup/PromotionContext.php
```diff
@@ -0,0 +1,111 @@
...
+        $this->objectManager = $objectManager;
+    }
+
+    /**
+     * @Given there is a promotion :promotionName
+     */
+    public function thereIsPromotion($promotionName)
+    {
+        $promotion = $this->testPromotionFactory->create($promotionName);
+        $promotion->addChannel($this->sharedStorage->get('channel'));
```
michalmarcinkowski: How about `$this->testPromotionFactory->createForChannel($promotionName, $this->sharedStorage->get('channel'));`?

src/Sylius/Behat/Context/Setup/PromotionContext.php
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * @Given /^it gives "(?:€|£|\$)([^"]+)" fixed discount to every order$/
+     */
+    public function itGivesFixedDiscountToEveryOrder($amount)
+    {
+        $currentPromotion = $this->sharedStorage->get('promotion');
+
+        $action = $this->actionFactory->createFixedDiscount($this->getPriceFromString($amount));
+        $currentPromotion->addAction($action);
```
michalmarcinkowski: Didn't we agree that the action factory should also receive `Promotion?  `createFixedDiscount($currentPromotion, $this->getPriceFromString($stringAmount))`?

src/Sylius/Behat/spec/Context/Ui/CartContextSpec.php
```diff
@@ -54,4 +54,12 @@ function it_checks_if_cart_has_given_tax_total(CartSummaryPage $cartSummaryPage)
...
         $this->myCartTaxesShouldBe('$50.00');
     }
+
+    function it_checks_if_cart_has_given_promotion_total(CartSummaryPage $cartSummaryPage)
+    {
+        $cartSummaryPage->open()->shouldBeCalled();
+        $cartSummaryPage->getPromotionTotal()->willReturn('$50.00');
+
+        $this->myCartPromotionsShouldBe('$50.00');
+    }
```
michalmarcinkowski: Missing spec for use case when the amounts does not match:
```
$cartSummaryPage->getPromotionTotal()->willReturn('$50.00');

$this->myCartPromotionsShouldBe('$100.00');
```

src/Sylius/Component/Core/spec/Test/Factory/TestPromotionFactorySpec.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $this->shouldHaveType('Sylius\Component\Core\Test\Factory\TestPromotionFactory');
+    }
+
+    function it_implements_test_promotion_factory_interface()
+    {
+        $this->shouldImplement(TestPromotionFactoryInterface::class);
+    }
+
+    function it_creates_promotion_based_on_provided_data($promotionFactory, PromotionInterface $promotion)
```
michalmarcinkowski: ...based_on_provided_**name**

src/Sylius/Behat/Context/Setup/PromotionContext.php
Zales0123: I think that method ``createFixedDiscount`` should only create fixed discount action, not also add it to promotion. We could of course create ``createFixedDiscountForPromotion`` method...
However I'm not sure about it. It could leads to some confusion when reading ``itGivesFixedDiscountToEveryOrder`` method, as it would create action and flush ``objectManager``... and first question could be "why? we've just created ``$action``, why flushing manager?". Adding ``$action`` to ``$promotion`` at this level it is clear what is it used for and why ``objectManager`` should be flushed.

src/Sylius/Behat/Context/Setup/PromotionContext.php
Zales0123: :+1:

src/Sylius/Behat/spec/Context/Ui/CartContextSpec.php
Zales0123: Fixed :+1:

src/Sylius/Behat/Context/Setup/PromotionContext.php
pjedrzejewski: Yeah, I think it should only create action and then we should ``$promotion->addAction($action)``.

pjedrzejewski:
Thank you Mateusz! :+1:

src/Sylius/Behat/Context/Setup/PromotionContext.php
michalmarcinkowski: My idea was that the `action` doesn't make sense without promotion, so perhaps `createFixedDiscountForPromotion` would be a better method for that, but let's stay with this for now :)

PR #4205, opened 2016-02-17
Always declare HTML encoding immediately after the opening head tag.

adamelso:
From the [W3C documentation][1] :

> The declaration should fit completely within the first 1024 bytes at
> the start of the file, so it's best to put it immediately after the
> opening head tag.

If the Twig title blocks exceed this (which would be rare), it's possible
for pages using those templates to not be encoded as UTF-8 when sent.

[1]: https://www.w3.org/International/questions/qa-html-encoding-declarations.en#quickanswer


pjedrzejewski:
Thanks Adam! :+1:

PR #4214, opened 2016-02-18
[Payment] Adding PaymentFactory 

lchrusciel:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| License       | MIT

### DIFF
--- a/src/Sylius/Bundle/PaymentBundle/DependencyInjection/SyliusPaymentExtension.php
+++ b/src/Sylius/Bundle/PaymentBundle/DependencyInjection/SyliusPaymentExtension.php
@@ -12,13 +12,14 @@
 namespace Sylius\Bundle\PaymentBundle\DependencyInjection;
 
 use Sylius\Bundle\ResourceBundle\DependencyInjection\Extension\AbstractResourceExtension;
+use Sylius\Component\Resource\Factory\Factory;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
+use Symfony\Component\DependencyInjection\Parameter;
 
 /**
- * Payments extension.
- *
  * @author Paweł Jędrzejewski <pawel@sylius.org>
  */
 class SyliusPaymentExtension extends AbstractResourceExtension
@@ -42,5 +43,11 @@ class SyliusPaymentExtension extends AbstractResourceExtension
         }
 
         $container->setParameter('sylius.payment_gateways', $config['gateways']);
+
+        $factoryDefinition = new Definition(Factory::class, [new Parameter('sylius.model.payment.class')]);
+        $paymentFactoryClass = $container->getParameter('sylius.factory.payment.class');
+        $decoratedPaymentFactoryDefinition = new Definition($paymentFactoryClass, [$factoryDefinition]);
+
+        $container->setDefinition('sylius.factory.payment', $decoratedPaymentFactoryDefinition);
     }
 }
--- a/src/Sylius/Component/Core/OrderProcessing/PaymentProcessor.php
+++ b/src/Sylius/Component/Core/OrderProcessing/PaymentProcessor.php
@@ -13,29 +13,24 @@ namespace Sylius\Component\Core\OrderProcessing;
 
 use Sylius\Component\Core\Model\OrderInterface;
 use Sylius\Component\Core\Model\PaymentInterface;
+use Sylius\Component\Payment\Factory\PaymentFactoryInterface;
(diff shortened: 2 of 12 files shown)


src/Sylius/Bundle/PaymentBundle/DependencyInjection/SyliusPaymentExtension.php
```diff
@@ -42,5 +45,14 @@ public function load(array $config, ContainerBuilder $container)
         }
 
         $container->setParameter('sylius.payment_gateways', $config['gateways']);
+
+        $factoryDefinition = new Definition(Factory::class);
+        $factoryDefinition->setArguments([new Parameter('sylius.model.payment.class')]);
+        $paymentFactoryDefinition = $container->getDefinition('sylius.factory.payment');
+        $paymentFactoryClass = $paymentFactoryDefinition->getClass();
+        $decoratedPaymentFactoryDefinition = new Definition($paymentFactoryClass);
+        $decoratedPaymentFactoryDefinition->setArguments([$factoryDefinition]);
```
michalmarcinkowski: Should this work?
```
$paymentFactoryDefinition = $container->getDefinition('sylius.factory.payment');
$paymentFactoryClass = $paymentFactoryDefinition->getClass();
$decoratedPaymentFactoryDefinition = new Definition($paymentFactoryClass);
$decoratedPaymentFactoryDefinition->setArguments([$paymentFactoryDefinition]);
```

pjedrzejewski:
@lchrusciel Can we also remove the dispatcher from the listener? It's never used.

pjedrzejewski:
Please take a look at the failing build, otherwise looks good to me. In a separate PR we should definitely figure out a better name than PaymentProcessor, it's more like OrderPaymentsUpdater, but still far from perfect. :/

src/Sylius/Bundle/PaymentBundle/DependencyInjection/SyliusPaymentExtension.php
michalmarcinkowski: How about squeezing this to one line:
`$decoratedPaymentFactoryDefinition = new Definition($paymentFactoryClass, [$factoryDefinition]);`?

src/Sylius/Bundle/PaymentBundle/DependencyInjection/SyliusPaymentExtension.php
```diff
@@ -42,5 +45,14 @@ public function load(array $config, ContainerBuilder $container)
         }
 
         $container->setParameter('sylius.payment_gateways', $config['gateways']);
+
+        $factoryDefinition = new Definition(Factory::class);
+        $factoryDefinition->setArguments([new Parameter('sylius.model.payment.class')]);
+        $paymentFactoryDefinition = $container->getDefinition('sylius.factory.payment');
+        $paymentFactoryClass = $paymentFactoryDefinition->getClass();
```
michalmarcinkowski: `$paymentFactoryClass = $container->getParameter('sylius.factory.payment.class');`?

lchrusciel: :+1:

src/Sylius/Bundle/PaymentBundle/DependencyInjection/SyliusPaymentExtension.php
lchrusciel: :+1:

pjedrzejewski:
Thank you Łukasz! :+1:

src/Sylius/Component/Core/OrderProcessing/PaymentProcessor.php
```diff
@@ -43,15 +38,11 @@ public function __construct(FactoryInterface $paymentFactory)
     /**
      * {@inheritdoc}
      */
-    public function createPayment(OrderInterface $order)
+    public function processOrderPayments(OrderInterface $order)
```
peteward: Hi, I know this was merged a while ago, but I'm just working through an update...

Is this class (or this method) actually doing any processing? 

It seems a bit pointless and could go into the Factory (or a Core extension/decoration of the Factory), because the naming is confusing now. There's no processing of order payments going on here that I can see - just creating them.

peteward: also the name implies multiple payments but this method only creates a single new payment for the total of an order. Could be something like `PaymentFactory::createForOrderTotal()` ?

pjedrzejewski:
@peteward Yeah, the naming is unfortunate, but we want to have it separated from payment factory because we want to allow custom implementations, which may create multiple payments or do whatever they want. That's why we have this extra service. That being said, the naming and this as a whole must be improved, I agree.

peteward:
Understood, perhaps `Provider` or `Builder`?

Issue #4211, opened 2016-02-18
Administration Pagination Broken in every Page


DanPhyxius:
Hi all,

first of all, thank you for this amazing framework, you guys have done a great job!

I found a critical bug on many views of administration zone and i saw that the bug exists in demo version too. Actually every view with paginations doesn't work.

<img width="1217" alt="schermata 2016-02-18 alle 13 02 21" src="https://cloud.githubusercontent.com/assets/4693301/13142794/e843104e-d63f-11e5-9c0f-7bfca40d6380.png">

Is there some fix that i could use for that?

Thanks a lot,
Dan

DanPhyxius:
Hi @pjedrzejewski ,

while I'm waiting for the final fix, could i ask you a quick to fix the problem? 
I would have some business urgency and I would be able to resolve that.

Thanks a lot for any answer,
Dan

pjedrzejewski:
@DanPhyxius Hey Dan! Looking into this right now. :) Thanks for reporting! I will get back to you with some feedback.

DanPhyxius:
We found a temporary fix, thanks a lot @pjedrzejewski! :+1:

nakashu:
@DanPhyxius please share :)

DanPhyxius:
@nakashu It's just a temporary raw workaround for the Products Pagination, i think it would be better to get a true fix for that :)

pjedrzejewski:
I added a test case for API - https://github.com/Sylius/Sylius/pull/4216/files, seems like it is problem only with HTML version or perhaps specific to how Sylius uses ResourceBundle. I will investigate further and hopefully fix it in admin too...

xantrix:
it seems related to setting the correct current page query param to the PagerFanta paginator as it is already being done on https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php#L60 for all resources without a specific Repository like http://demo.sylius.org/administration/countries/?page=2

pjedrzejewski:
@xantrix You are right, working on a fix right now. The issue is that we do not handle paginator correctly if it is returned from a custom method. :) Thanks!

xantrix:
you're welcome @pjedrzejewski ;)

pjedrzejewski:
Should be fixed when #4216 is merged. Thanks for reporting!

michalmarcinkowski:
Should be fixed now.

[...]

PR #4216, opened 2016-02-18
Add test for pagination in ResourceBundle (via API)

pjedrzejewski:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/Sylius/Sylius/issues/4211
| License       | MIT
| Doc PR        | -

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourcesCollectionProviderSpec.php
+++ b/src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourcesCollectionProviderSpec.php
@@ -164,4 +164,30 @@ class ResourcesCollectionProviderSpec extends ObjectBehavior
 
         $this->get($requestConfiguration, $repository)->shouldReturn($paginatedRepresentation);
     }
+
+    function it_sets_current_page_on_paginator_from_custom_method(
+        RequestConfiguration $requestConfiguration,
+        RepositoryInterface $repository,
+        Pagerfanta $paginator,
+        Request $request,
+        ParameterBag $queryParameters
+    ) {
+        $requestConfiguration->isHtmlRequest()->willReturn(true);
+        $requestConfiguration->getRepositoryMethod()->willReturn('findAll');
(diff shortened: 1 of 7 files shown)


src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourcesCollectionProviderSpec.php
```diff
@@ -164,4 +164,66 @@ function it_creates_a_pagniated_respresentation_for_pagerfanta_for_non_html_requ
...
+        $requestConfiguration->isHtmlRequest()->willReturn(false);
+        $requestConfiguration->getRepositoryMethod()->willReturn('findAll');
+        $requestConfiguration->getRepositoryArguments()->willReturn(['foo']);
+
+        $requestConfiguration->isPaginated()->willReturn(true);
+        $requestConfiguration->isLimited()->willReturn(false);
+        $requestConfiguration->getCriteria()->willReturn([]);
+        $requestConfiguration->getSorting()->willReturn([]);
+
+        $repository->findAll('foo')->willReturn($paginator);
```
michalmarcinkowski: How about renaming the `findAll` method to custom one, e.g. `createPaginator`?

pjedrzejewski: The point was to demonstrate that it works with a custom method, other than ``createPaginator``. If I change it to ``createPaginator`` that spec would pass even if custom method is not used at all/not working, etc. Ideally, we can introduce "TestRepositoryInterface" just for this spec with something really custom. I'd do that later though and get this fixed now.

michalmarcinkowski: I meant `createCustomPaginator` but as you noticed it will require introducing `TestRepositoryInterface`, so topic for later :+1:

Issue #4213, opened 2016-02-18
[Variation] Introduce code for option values and replace presentation with name


pjedrzejewski:
Option should have:

* code
* name (translatable)

OptionValue should have:

* code
* value (translatable)

Somehow we have ended with ``name``, ``presentation`` and ``code`` at the same time. :)

vikey89:
@pjedrzejewski the goal of this issue is rename `value` of OptionValue in `name`. It's correct?

michalmarcinkowski:
Not exactly. The goal of this issue is to remove current `name` from `Option` and rename `presentation` to `name`. The current `name` of `Option` is exactly the same as `code`, so it's duplicated. The `OptionValue` entity is correct :)

vikey89:
ah ok, now I understand thanks :)

nakashu:
this should be closed

PR #4237, opened 2016-02-19
[Variation] Introduce code for option values and replace presentation with name 

vikey89:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

This PR is based from [#4213](https://github.com/Sylius/Sylius/issues/4213)

### DIFF
--- a/app/migrations/Version20160220110242.php
+++ b/app/migrations/Version20160220110242.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20160220110242 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_product_option DROP name');
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_product_option ADD name VARCHAR(255) NOT NULL COLLATE utf8_unicode_ci');
+    }
+}
--- a/src/Sylius/Component/Variation/Model/OptionValue.php
+++ b/src/Sylius/Component/Variation/Model/OptionValue.php
@@ -105,13 +105,13 @@ class OptionValue extends AbstractTranslatable implements OptionValueInterface
     /**
      * {@inheritdoc}
      */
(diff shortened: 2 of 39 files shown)


vikey89:
@pjedrzejewski I think I've finished on this PR.

vikey89:
@pjedrzejewski @michalmarcinkowski I would like to help with something else :)
Any suggestion?

app/migrations/Version20160220110242.php
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_product_option DROP name');
+        $this->addSql('DROP INDEX fulltext_search_idx ON sylius_search_index');
```
pjedrzejewski: This and line below seem incorrect.

pjedrzejewski:
@vikey89 Great, thank you so much for your time! Could you please squash the commits and have a look at the migration? Otherwise looks good to go!

src/Sylius/Component/Variation/Model/OptionValue.php
```diff
@@ -105,13 +105,13 @@ public function setValue($value)
     /**
      * {@inheritdoc}
      */
-    public function getName()
+    public function getCodeOption()
```
pjedrzejewski: This should be ``getOptionCode`` I think, wdyt?

vikey89: yes, you're right.

vikey89:
@pjedrzejewski I squashed the commits and fixes migration and option code.

src/Sylius/Component/Variation/Model/OptionValue.php
```diff
@@ -105,13 +105,13 @@ public function setValue($value)
     /**
      * {@inheritdoc}
      */
-    public function getName()
+    public function getOptionCode()
```
okwinza: :+1: this is much better name.

okwinza:
@pjedrzejewski as a sidenote with optionvalues now being translatable we have to be careful when displaying all option values at once otherwise we risk facing something like this(800+ queries FTW)

zomg

In light of this my #4158 quickly becomes quite important to merge IMO.
Or maybe we can introduce some caching mechanism like memcached for doctrine or something.

nakashu:
@okwinza is this something that can be handled by setting EAGER loading on the translations and optionValues? It should be then loaded with a single query.

http://doctrine-orm.readthedocs.org/projects/doctrine-orm/en/latest/reference/annotations-reference.html#onetomany

okwinza:
ping @pjedrzejewski 
this seems to be ready to merge. And i also need your thoughts on my comment above.

vikey89:
@pjedrzejewski @michalmarcinkowski I would like to collaborate with you more often.
Do you have some issues with more priority? :)

okwinza:
@vikey89 try to come up with some decent price slider for example :D 

On a serious note it would be nice to have it and #4285 merged already :)
Double ping @pjedrzejewski

pjedrzejewski:
Thanks Vincenzo! :+1:

psyray:
I think this PR was merged too fast.
Migrations are not correctly done.

What is called migration is not only **Create and drop**, in some case, like this one, migrations must be **Create, select, insert and finally drop** like this [migration] (https://github.com/Sylius/Sylius/blob/master/app/migrations/Version20141203091118.php)

I am rewriting the migration to copy all values from the option_value table in option_value_translation table because in this state, this PR break my project and certainly other project

psyray:
Issue & PR created

pjedrzejewski:
Hey @psyray, sorry for that. Most migrations are autogenerated right now. We want to fix that soon, as we are minimizing the changes to schema right now. All future migrations will migrate the data as well. Thanks for your contribution!

PR #4238, opened 2016-02-20
[Review] - average rating recalculation on change, decouple model

nakashu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

Average rating on `review subject` (Product) was only recalculated when approved (state machine).
This PR adds events and fixes remover code to recalculate also when:

* `review` is deleted
* `review` rating was changed - its possible in administration
* `customer` was removed and associated `reviews` where deleted.

There is still one problem with recalculation when updating `review` in administration and `review subject `is changed. In that situation, average rating on previous `review subject` is not modified.
Possibility to change the subject is in my opinion pretty weird one so there are two solutions:

* disable change of subject on review once it is set.
* add onFlush doctrine listener to handle updating both old and new subject average rating.

### DIFF
--- a/src/Sylius/Bundle/ReviewBundle/EventListener/ReviewChangeListener.php
+++ b/src/Sylius/Bundle/ReviewBundle/EventListener/ReviewChangeListener.php
@@ -19,7 +19,7 @@ use Symfony\Component\EventDispatcher\GenericEvent;
 /**
  * @author Mateusz Zalewski <mateusz.p.zalewski@gmail.com>
  */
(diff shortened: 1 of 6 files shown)


nakashu:
bump

pjedrzejewski:
I think we will remove the ability to update/create reviews in the new admin, at least initially. Thanks @nakashu! :+1:

PR #4239, opened 2016-02-20
[Resource][Pagination] - respect paginate parameter from config

nakashu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

`Paginator` on `Resource` now respects and sets number of items to paginate on page, set in route configuration. eg:

```yml

sylius_backend_product_index:
    path: /
    methods: [GET]
    defaults:
        _sylius:
            template: SyliusWebBundle:Backend/Product:index.html.twig
            sortable: true
            paginate: 5 # works now
            repository:
                method: createFilterPaginator
                arguments: [$criteria, $sorting, $deleted]
```

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourcesCollectionProviderSpec.php
+++ b/src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourcesCollectionProviderSpec.php
@@ -115,6 +115,7 @@ class ResourcesCollectionProviderSpec extends ObjectBehavior
         $requestConfiguration->getRepositoryMethod()->willReturn(null);
 
         $requestConfiguration->isPaginated()->willReturn(true);
+        $requestConfiguration->getPaginationMaxPerPage()->willReturn(5);
         $requestConfiguration->isLimited()->willReturn(false);
         $requestConfiguration->getCriteria()->willReturn([]);
         $requestConfiguration->getSorting()->willReturn([]);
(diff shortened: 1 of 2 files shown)


steffenbrem:
Isnt this what limit is for or am I wrong?

steffenbrem:
Nvm, I too had this problem indeed.

steffenbrem:
@pjedrzejewski  Maybe we should get rid of the `paginate` all together? Since it does the same as `limit`, it may be confusing.

src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourcesCollectionProviderSpec.php
```diff
@@ -187,11 +192,12 @@ function it_sets_current_page_on_paginator_from_custom_method(
         $queryParameters->get('page', 1)->willReturn(8);
 
         $paginator->setCurrentPage(8)->shouldBeCalled();
+        $paginator->setMaxPerPage(5)->shouldBeCalled();
 
         $this->get($requestConfiguration, $repository)->shouldReturn($paginator);
     }
 
-    function it_creates_a_pagniated_respresentation_for_pagerfanta_for_non_html_requests_with_a_custom_repository_method(
+    function it_creates_a_ppaginated_representation_for_pagerfanta_for_non_html_requests_with_a_custom_repository_method(
```
steffenbrem: `ppaginated` -> `paginated`

pjedrzejewski:
Thank you @nakashu!

@steffenbrem :+1: for getting rid of paginate.

nakashu:
@steffenbrem @pjedrzejewski 
even if you get rid of it you still need two confguration options. 

1. 'give-me-the-paginator-object' - currently paginate with any value other than false
2. 'limit-number-of-items-per-page' - currently limit and paginate, depends on if paginate is false

PR #4242, opened 2016-02-21
[CoreBundle] Decode url when finding entity based on request URI

aramalipoor:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | maybe!?
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

We've created a custom slug `Transliterator` to have Persian characters in URL but when visit a Taxon URI needs to be decoded.

Do you think this a right place to decode it?

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Routing/RouteProvider.php
+++ b/src/Sylius/Bundle/CoreBundle/Routing/RouteProvider.php
@@ -73,7 +73,9 @@ class RouteProvider extends DoctrineProvider implements RouteProviderInterface
         }
 
         foreach ($this->getRepositories() as $className => $repository) {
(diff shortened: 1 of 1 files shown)


aramalipoor:
@pjedrzejewski ;)

aramalipoor:
@pjedrzejewski ping ;)

pjedrzejewski:
I merge this as a hotfix, but the idea is to remove these route providers... So we will need to figure out another way when we do that. :) Thanks Aram!

aramalipoor:
@pjedrzejewski Thanks, we needed this hotfix ;)
It is a good point to get rid of these providers!

Issue #4148, opened 2016-02-14
[RFC] Rework Settings bundle


pjedrzejewski:
Before going stable we need to tweak a bit how Settings are handled. We also need that for Themes configuration.

Problems:

* Schema is coupled to namespace.
* Namespace is stored on every single parameter.

Solution:

* [x] Introduce Settings model, which has a collection of Parameter(s). It would look like this:
    * namespace (immutable and unique)
    * schema (immutable)
    * parameters (collection)
* [x] Decouple schema from namespace. Every settings schema would be registered in container, just it is now. It will be identified by unique alias.
* [x] Make editing/loading settings dependent on namespace & schema.

Example schema registered in container:

```xml
<service id="app.settings_schema.seo" class="AppBundle\Settings\\SeoSettingsSchema">
    <tag name="sylius.settings_schema" alias="app_seo" />
</service>
```

This adds given schema to the registry.

@pamil For themes, we could have a compiler pass/extension that goes over all themes and if specific file exists, like "ThemeSettingsSchema.php" it gets the class from it and registers it as a service in container. Only problem I see is with autoloading these classes.

Routing to update settings:

```
app_update_seo_settings:
    path: /settings/seo
    methods: [GET, PUT, PATCH]
    defaults:
        _controller: sylius.controller.settings:updateAction
        _sylius:
             namespace: banana
             schema: app_seo
```

This will use schema ``app_seo`` and load/save stuff under ``banana``.

PHP:

```php
$seoSettings = $this->settingsManager->load('banana', 'app_seo'); // Loads from database if exists or creates.
$seoSettings['keywords'] = ['foo', 'bar'];
$this->settingsManager->save($seoSettings);
```

You can skip schema, but it won't create settings if they do not exist, it will throw an exception.

This allows multiple namespaces using the same schema:

```php
$userNotificationPreferences = $this->settingsManager->load('user_notification_preferences_'.$user->getId(), 'app_user_notification_preferences'); // Loads from database if exists or creates.
$userPreferences['notify_about_new_order'] = false;
$userPreferences['notify_about_new_customer'] = true;
$this->settingsManager->save($userNotificationPreferences);
```

Thoughts? It actually requires just small changes to existing codebase.

gperdomor:
:+1: Awesome... With this, we can have settings for every model :D

aramalipoor:
:+1: This is great.

[...]

PR #4231, opened 2016-02-19
[SettingsBundle] Decouple schema from namespace

tuka217:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets |  #4148
| License       | MIT
| Doc PR        | -

PR #4246, opened 2016-02-21
[SettingsBundle] Settings rework

steffenbrem:
| Q             | A
| ------------- | ---
| Bug fix?      | no|yes
| New feature?  |yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #4148
| License       | MIT

If using a `namespace` for your settings is not enough, you can use specific resolvers for a schema that can be used to retrieve the correct settings.

The interface for a settings resolver looks like this:
```php
interface SettingsResolverInterface
{
    /**
     * Resolves settings based on schema alias.
     *
     * @param string $schemaAlias
     * @param string $namespace
     *
     * @return SettingsInterface
     */
    public function resolve($schemaAlias, $namespace = null);
}
```

You can register a settings resolver like this:
```xml
<service id="acme.settings.theme_resolver" class="Acme\AppBundle\Settings\ThemeSettingsResolver">
    <tag name="sylius.settings_resolver" schema="theme"/>
</service>
```

When you have a schema defined as `theme` (alias), it will then use the above resolver instead of the default one (the default resolver just loads by schema name). You can also easily switch out the default resolver and implement different default behaviour, depending on your application needs.

There are a few BC breaks, so is `loadSettings` changed to just `load`, and the same for `saveSettings` to `save`. Note that previously you had to pass in the schema name to the `save` method. This is no longer required, since you have the whole `Settings` object when you load a setting. This means that we already know what schema it has.

Also the `SettingsEvent` has changed, it now only holds a reference to the `Settings` object, from which you can retrieve it's parameters and schema alias.


steffenbrem:
@pjedrzejewski Do you think it is okay to use a doctrine listener to manage the transformation of parameter values. It greatly reduces bugs, since when you flush before you call `save` on the `settingsManager` it will not transform it. See [this file](https://github.com/steffenbrem/Sylius/blob/feature/settings-rework/src/Sylius/Bundle/SettingsBundle/Doctrine/EventListener/TransformParameterListener.php) for implementation details.

steffenbrem:
@pjedrzejewski May I ask why you use both the `OptionsResolver` and the `FormBuilder` in a `Schema` class? What purpose do they both fill?

ahilles107:
:+1:

steffenbrem:
I am currently rewriting a bit to solve the issue with the creation of new parameters through the `Settings` model. My idea was to introduce a new class, the `ParameterCollection` which extends `ArrayCollection`. This class holds a reference to the `Settings` model, but you do not work directly with it in your application, you work with the `ParameterCollection`. This has some nice benefits, since you do not want to flush changes to the db on accident when you call `$em->flush()` before `$settingsManager->save()`. This change also removes the needed doctrine listener, since that bug no longer is present when using the parameter collection class.

steffenbrem:
Ready!

pjedrzejewski:
@steffenbrem I will do my best to have a look at this as soon as possible, to avoid duplicated effort, see #4231.

steffenbrem:
@pjedrzejewski I saw that PR, if you like I could make the schema registry extend the service registry. I think it is good too keep the nice exception messages that you get now with the current schema registry.

sjonkeesse:
:+1:

takeit:
we also need this, the idea and the implementation looks :+1:

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
```diff
@@ -35,14 +38,19 @@ class SettingsManager implements SettingsManagerInterface
     protected $schemaRegistry;
```
tuka217: You can use ServiceRegistryInterface. The SchemaRegistry should be removed. See  #4231

pjedrzejewski:
@steffenbrem We discussed with @tuka217 in the office and we will go with your solution. Could you adapt some small tweaks from our PR into your? https://github.com/Sylius/Sylius/pull/4246#discussion_r53754633

I will review the PR and merge it soon, thanks!

src/Sylius/Bundle/CoreBundle/Provider/DefaultTaxZoneProvider.php
```diff
@@ -37,8 +37,8 @@ public function __construct(Settings $settings)
      */
     public function getZone()
     {
-        if ($this->settings->has('default_tax_zone')) {
-            return $this->settings->get('default_tax_zone');
+        if ($zone = $this->settings->get('default_tax_zone')) {
+            return $zone;
         }
 
         return null;
```
michalmarcinkowski: You can simply do `return $this->settings->get('default_tax_zone');`

src/Sylius/Bundle/SettingsBundle/Model/Settings.php
```diff
@@ -1,119 +1,105 @@
...
 
     /**
-     * @param array $parameters
+     * @var string
      */
-    public function __construct(array $parameters)
-    {
-        $this->parameters = $parameters;
-    }
+    protected $schema;
```
michalmarcinkowski: Isn't it `$schemaAlias`?

src/Sylius/Bundle/SettingsBundle/Resolver/DefaultResolver.php
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Sylius\Bundle\SettingsBundle\Resolver;
+
+use Doctrine\ORM\NonUniqueResultException;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+/**
+ * @author Steffen Brem <steffenbrem@gmail.com>
+ */
+class DefaultResolver implements SettingsResolverInterface
```
michalmarcinkowski: I would call it `DefaultSettingsResolver`.

src/Sylius/Bundle/SettingsBundle/Resolver/SettingsResolverInterface.php
```diff
@@ -0,0 +1,29 @@
...
+interface SettingsResolverInterface
+{
+    /**
+     * Resolves settings based on schema.
+     *
+     * @param string $schema
+     *
+     * @return SettingsInterface
+     */
+    public function resolve($schema);
```
michalmarcinkowski: Isn't it `$schemaAlias`?

src/Sylius/Bundle/SettingsBundle/Resources/config/services.xml
```diff
@@ -21,33 +21,44 @@
...
         <parameter key="sylius.settings.schema_registry.class">Sylius\Bundle\SettingsBundle\Schema\SchemaRegistry</parameter>
+        <parameter key="sylius.settings.resolver.default.class">Sylius\Bundle\SettingsBundle\Resolver\DefaultResolver</parameter>
+        <parameter key="sylius.settings.resolver_registry.class">Sylius\Component\Registry\ServiceRegistry</parameter>
     </parameters>
 
     <services>
         <service id="sylius.controller.settings" class="%sylius.controller.settings.class%">
             <call method="setContainer">
-                <argument type="service" id="service_container" />
+                <argument type="service" id="service_container"/>
```
michalmarcinkowski: This should be reverted.

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
```diff
@@ -11,96 +11,116 @@
...
-    function it_should_allow_to_get_parameters_via_array_access()
+    function it_can_check_if_it_has_a_parameter(ParameterInterface $parameter)
     {
-        $this['page']->shouldReturn(1);
+        $parameter->setSettings($this)
+            ->shouldBeCalled()
+        ;
+
+        $parameter->getName()
+            ->shouldBeCalled()
```
michalmarcinkowski: Should be removed, `willReturn` if enough.

src/Sylius/Bundle/SettingsBundle/Controller/SettingsController.php
```diff
@@ -37,17 +36,11 @@ class SettingsController extends FOSRestController
      */
     public function showAction(Request $request)
     {
-        $namespace = $request->get('namespace');
+        $schema = $request->get('namespace');
```
michalmarcinkowski: `$schemaName = $request->attributes->get('schema')`

src/Sylius/Bundle/SettingsBundle/Controller/SettingsController.php
```diff
@@ -59,35 +52,29 @@ public function showAction(Request $request)
...
      *
      * @return Response
      */
-    public function updateAction(Request $request, $namespace)
+    public function updateAction(Request $request, $schema)
     {
-        $this->isGrantedOr403($namespace);
+        $this->isGrantedOr403($schema);
 
         $manager = $this->getSettingsManager();
```
michalmarcinkowski: `$settingsManager`

src/Sylius/Bundle/SettingsBundle/DependencyInjection/Compiler/RegisterResolversPass.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        if (!$container->hasDefinition('sylius.settings.resolver_registry')) {
+            return;
+        }
+
+        $resolverRegistry = $container->getDefinition('sylius.settings.resolver_registry');
+
+        foreach ($container->findTaggedServiceIds('sylius.settings_resolver') as $id => $tags) {
+            foreach ($tags as $attributes) {
+                if (!array_key_exists('schema', $attributes)) {
```
michalmarcinkowski: `!isset($attributes['schema'])`

src/Sylius/Bundle/SettingsBundle/Model/Settings.php
tuka217: We can have in Sylius Schemas object. For this reason this name can be confusing. Maybe the better will be ``schemaName``, WDYT? :)

src/Sylius/Bundle/SettingsBundle/Model/Settings.php
```diff
@@ -1,119 +1,105 @@
...
 
-/*
- * This file is part of the Sylius package.
- *
- * (c) Paweł Jędrzejewski
- *
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-
```
tuka217: You shouldn't remove this lines.

src/Sylius/Bundle/SettingsBundle/Resources/config/doctrine/model/Settings.orm.xml
```diff
@@ -0,0 +1,31 @@
...
+-->
+
+<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping">
+
+    <mapped-superclass name="Sylius\Bundle\SettingsBundle\Model\Settings" table="sylius_settings">
+        <id name="id" column="id" type="integer">
+            <generator strategy="AUTO"/>
+        </id>
+
+        <field name="schema" column="schema_name" type="string" nullable="false"/>
```
tuka217: Every Settings should have namespace and schema, because in the future we want to create many settings object for one schema.

src/Sylius/Bundle/SettingsBundle/Model/Settings.php
steffenbrem: I think it is good to be consistent, I vote for `schemaAlias`, since we use that everywhere else.

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
```diff
@@ -101,20 +102,29 @@ public function __construct(
...
+        if ($this->resolverRegistry->has($schemaAlias)) {
+            $resolver = $this->resolverRegistry->get($schemaAlias);
         }
 
-        if ($this->cache->contains($namespace)) {
-            $parameters = $this->cache->fetch($namespace);
-        } else {
-            $parameters = $this->getParameters($namespace);
-            $this->cache->save($namespace, $parameters);
+        $settings = $resolver->resolve($schemaAlias);
```
tuka217: It ok, but we want to have mulitple Settings object, which will have the same schema. They will have different namespaces only.

src/Sylius/Bundle/SettingsBundle/Resources/config/doctrine/model/Settings.orm.xml
steffenbrem: You could do the same with the resolvers, could you give a specific use case where a resolver could not fit the need for having namespaces on settings? As I see it, implementing a resolver for a specific schema could do the same as what you want to do with the namespace.

steffenbrem: If you like, I can implement both, but I am not sure if it fills a real need there. The namespace should be optional too right?

src/Sylius/Bundle/SettingsBundle/Resources/config/services.xml
```diff
@@ -21,33 +21,44 @@
...
             </call>
         </service>
 
         <service id="sylius.settings.form_factory" class="%sylius.settings.form_factory.class%">
-            <argument type="service" id="sylius.settings.schema_registry" />
-            <argument type="service" id="form.factory" />
```
tuka217: You should revert it.

steffenbrem: In this file both are both used, I think it is good to stick with one or the other. In the rest of Sylius a space after an empty tag is used. Or do you prefer fixing those things in a seperate PR? If you still want to I can revert it.

src/Sylius/Bundle/SettingsBundle/Resolver/DefaultSettingsResolver.php
```diff
@@ -0,0 +1,47 @@
...
+     * {@inheritdoc}
+     */
+    public function resolve($schemaAlias)
+    {
+        try {
+            return $this->settingRepository->findOneBy([
+                'schemaAlias' => $schemaAlias
+            ]);
+        } catch (NonUniqueResultException $e) {
+            throw new \LogicException(sprintf('Multiple schemas found for "%s". You should probably define a custom settings resolver for this schema.', $schemaAlias));
```
tuka217: It won't be easier if you allow create many Settings objects for one schema? Now you should create custom reslover every time when you create more than one Settings object for schema. It generates a lot of additional code.

steffenbrem: True @tuka217 almost finished adding extra namespace to the resolver interface and then the default resolver uses the namespace too.

steffenbrem: Done, everything is in place now. Could you verify if the latest default resolver implementation is okay?

src/Sylius/Bundle/SettingsBundle/Resources/config/doctrine/model/Settings.orm.xml
tuka217: You need to create new reslover always if you have many Settings for one schema. When you allow create many Settings for one schema you don't have to create new class.

steffenbrem:
You want me to rebase this? Or should I wait till it's good to go?

pjedrzejewski:
@steffenbrem Would be good to squash it but please wait a bit, I will review it soon. Thanks!

steffenbrem:
@pjedrzejewski Okay will do! Thanks :+1:

steffenbrem:
Rebased against master, it was out of sync. @pjedrzejewski Do you have an indication when this might get merged? The way I have to implement theme settings depends on if this gets merged or not.

src/Sylius/Bundle/SettingsBundle/Form/Factory/SettingsFormFactoryInterface.php
```diff
@@ -19,11 +19,11 @@
 interface SettingsFormFactoryInterface
 {
     /**
-     * @param string     $namespace
+     * @param string     $schemaAlias
```
lchrusciel: Wa are not aligning parameters vertically anymore.

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
```diff
@@ -184,37 +184,25 @@ public function saveSettings($namespace, Settings $settings)
...
 
-        $this->cache->save($namespace, $parameters);
+        $this->settingsManager->flush();
+        $this->eventDispatcher->dispatch(SettingsEvent::POST_SAVE, new SettingsEvent($settings));
     }
 
     /**
-     * @param string $namespace
+     * @param SettingsBuilder $settingsBuilder
+     * @param array           $parameters
```
lchrusciel: Do not align parameters vertically

src/Sylius/Bundle/SettingsBundle/Model/SettingsInterface.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * @return Collection|ParameterInterface[]
+     */
+    public function getParameters();
+
+    /**
+     * @param string $name
+     *
+     * @throws \InvalidArgumentException
```
lchrusciel: Throws declarations shouldn't go to class? it is implementation detail at the end .

src/Sylius/Bundle/SettingsBundle/Resolver/SettingsResolverInterface.php
```diff
@@ -0,0 +1,30 @@
...
+
+use Sylius\Bundle\SettingsBundle\Model\SettingsInterface;
+
+/**
+ * @author Steffen Brem <steffenbrem@gmail.com>
+ */
+interface SettingsResolverInterface
+{
+    /**
+     * Resolves settings based on schema alias.
```
lchrusciel: Redundant comment

src/Sylius/Bundle/SettingsBundle/Resources/config/doctrine/model/Settings.orm.xml
```diff
@@ -0,0 +1,36 @@
...
+            <generator strategy="AUTO"/>
+        </id>
+
+        <field name="schemaAlias" column="schema_alias" type="string" nullable="false"/>
+        <field name="namespace" type="string" nullable="true"/>
+
+        <one-to-many target-entity="Sylius\Bundle\SettingsBundle\Model\ParameterInterface" mapped-by="settings"
+                     field="parameters" index-by="name" fetch="EAGER">
+            <cascade>
+                <cascade-persist/>
```
lchrusciel: Shouldn't parameters be deleted together with settings?

src/Sylius/Bundle/SettingsBundle/spec/Manager/SettingsManagerSpec.php
```diff
@@ -11,15 +11,15 @@
 
 namespace spec\Sylius\Bundle\SettingsBundle\Manager;
 
-use Doctrine\Common\Cache\Cache;
 use Doctrine\Common\Persistence\ObjectManager;
 use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
```
lchrusciel: Have you used it?

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
```diff
@@ -11,96 +11,128 @@
...
-    function it_should_overwrite_parameter()
+    function its_namespace_is_null_by_default()
     {
-        $this->set('page', 12);
-        $this->get('page')->shouldReturn(12);
+        $this->getNamespace()->shouldReturn(null);
     }
 
-    function it_should_complain_when_trying_to_remove_non_existing_parameter()
+    public function its_namespace_should_be_immutable_after_it_is_set()
```
lchrusciel: Remove visibility

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
```diff
@@ -11,96 +11,128 @@
...
-    function it_should_implement_array_access_interface()
+    function it_can_get_parameter_by_name(ParameterInterface $parameter)
     {
-        $this->shouldImplement('ArrayAccess');
+        $parameter->getName()
+            ->shouldBeCalled()
+            ->willReturn('left_side_products')
+        ;
+
+        $parameter->setSettings($this)
```
lchrusciel: This one will fit the line perfectly.

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
```diff
@@ -11,96 +11,128 @@
...
+        $this->addParameter($parameter);
+        $this->getParameters()->shouldHaveCount(1);
     }
 
-    function it_should_implement_array_access_interface()
+    function it_can_get_parameter_by_name(ParameterInterface $parameter)
     {
-        $this->shouldImplement('ArrayAccess');
+        $parameter->getName()
+            ->shouldBeCalled()
```
lchrusciel: Do we need this assertion? `getName` is not a builder.

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
```diff
@@ -11,96 +11,128 @@
...
+            ->willReturn('left_side_products')
+        ;
+
+        $parameter->setSettings($this)
+            ->shouldBeCalled()
+        ;
+
+        $this->addParameter($parameter);
```
lchrusciel: Missing `->shouldBeCalled()`

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
```diff
@@ -11,96 +11,128 @@
...
+
+        $this->getParameter('left_side_products')
```
lchrusciel: It should be in one line

src/Sylius/Bundle/SettingsBundle/Model/SettingsInterface.php
pamil: It's the way it communicates. You can either do `@return string|null` or `@return string @throws \FooBarException`, the second one is less error prone. 

It's good, but I'll vote for a custom exception (`ParameterNotFoundException`?).

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
```diff
@@ -11,96 +11,128 @@
...
-        $this['page']->shouldReturn(1);
+        $parameter->setSettings($this)
+            ->shouldBeCalled()
+        ;
+
+        $parameter->getName()
+            ->willReturn('title')
+        ;
+
+        $this->addParameter($parameter);
```
lchrusciel: Missing `->shouldBeCalled()`

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
```diff
@@ -11,96 +11,128 @@
...
 
-    function it_should_allow_to_get_parameters_via_array_access()
+    function it_can_check_if_it_has_a_parameter(ParameterInterface $parameter)
     {
-        $this['page']->shouldReturn(1);
+        $parameter->setSettings($this)
+            ->shouldBeCalled()
+        ;
+
+        $parameter->getName()
```
lchrusciel: put it in one line

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
```diff
@@ -11,96 +11,128 @@
...
+        $this->getParameter('left_side_products')
+            ->shouldReturn($parameter)
+        ;
     }
 
-    function it_should_allow_to_get_parameters_via_array_access()
+    function it_can_check_if_it_has_a_parameter(ParameterInterface $parameter)
     {
-        $this['page']->shouldReturn(1);
+        $parameter->setSettings($this)
```
lchrusciel: Put it in one line

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
lchrusciel: Same should be applied to specs below

src/Sylius/Bundle/SettingsBundle/Resolver/DefaultSettingsResolver.php
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @author Steffen Brem <steffenbrem@gmail.com>
+ */
+class DefaultSettingsResolver implements SettingsResolverInterface
+{
+    /**
+     * @var RepositoryInterface
+     */
+    protected $settingRepository;
+
```
pamil: Missing docblock.

src/Sylius/Bundle/SettingsBundle/spec/Manager/SettingsManagerSpec.php
```diff
@@ -27,15 +27,25 @@
...
     function let(
-        SchemaRegistryInterface $registry,
-        Cache $cache,
-        ObjectManager $manager,
-        RepositoryInterface $repository,
-        FactoryInterface $factory,
+        ServiceRegistryInterface $registry,
+        ServiceRegistryInterface $resolverRegistry,
+        ObjectManager $settingsManager,
+        FactoryInterface $settingsfactory,
```
pamil: settings**F**actory

src/Sylius/Bundle/SettingsBundle/spec/Manager/SettingsManagerSpec.php
```diff
@@ -27,15 +27,25 @@
 class SettingsManagerSpec extends ObjectBehavior
 {
     function let(
-        SchemaRegistryInterface $registry,
-        Cache $cache,
-        ObjectManager $manager,
-        RepositoryInterface $repository,
-        FactoryInterface $factory,
+        ServiceRegistryInterface $registry,
```
pamil: What is `$registry` for?

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
pamil: Not needed, it's a call to SUS.

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
pamil: Not needed, call to SUS.

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
pamil: :+1:

src/Sylius/Bundle/SettingsBundle/spec/Model/SettingsSpec.php
lchrusciel: :+1:

src/Sylius/Bundle/SettingsBundle/Resources/config/doctrine/model/Settings.orm.xml
steffenbrem: Isn't the `CASCADE DELETE` that is in `Parameter.orm.xml` enough? Why would we also need to cascade a doctrine remove here?

src/Sylius/Bundle/SettingsBundle/spec/Manager/SettingsManagerSpec.php
steffenbrem: It's the schemaRegistry, I will change it's name so its more clear.

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
```diff
@@ -71,29 +79,22 @@ class SettingsManager implements SettingsManagerInterface
...
-        SchemaRegistryInterface $schemaRegistry,
-        ObjectManager $parameterManager,
-        RepositoryInterface $parameterRepository,
+        ServiceRegistryInterface $schemaRegistry,
+        ServiceRegistryInterface $resolverRegistry,
+        ObjectManager $settingsManager,
+        FactoryInterface $settingsFactory,
         FactoryInterface $parameterFactory,
-        Cache $cache,
+        SettingsResolverInterface $defaultResolver,
```
pamil: Can't we get the default resolver from the resolver registry?

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
```diff
@@ -71,29 +79,22 @@ class SettingsManager implements SettingsManagerInterface
...
-    /**
-     * @param SchemaRegistryInterface  $schemaRegistry
-     * @param ObjectManager            $parameterManager
-     * @param RepositoryInterface      $parameterRepository
-     * @param FactoryInterface         $parameterFactory
-     * @param Cache                    $cache
-     * @param ValidatorInterface       $validator
-     * @param EventDispatcherInterface $eventDispatcher
-     */
     public function __construct(
```
pamil: It may be suitable for another iteration, but 8 dependencies seems bad.

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
```diff
@@ -101,20 +102,30 @@ public function __construct(
...
+        $settings = $resolver->resolve($schemaAlias, $namespace);
+
+        if (!$settings) {
+            $settings = $this->settingsFactory->createNew();
+            $settings->setSchemaAlias($schemaAlias);
         }
 
-        $schema = $this->schemaRegistry->getSchema($namespace);
+        // map parameters to a plain php array
+        $parameters = $settings->getParameters()
```
pamil: It would be great to put this logic somewhere inside Settings / ParameterCollection.

src/Sylius/Bundle/SettingsBundle/Resources/config/doctrine/model/Settings.orm.xml
steffenbrem: The only times you will need the ORM delete cascading is when you want to do some additional operations when you delete a specific parameter. For example when you have a gallery that has images and you remove that gallery, you want to also remove those images from your filesystem. I don't think we need to have the extra complexity of having doctrine cascade through all relations and fire delete events etc. The current setup is faster.

If you prefer it I can add it.

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
steffenbrem: I had it like that first. But I felt like this was a more flexible approach. Since you could more easily swap out the default resolver with something different. Otherwise you have to first `unregister` the `default` resolver from the registry and then register it again. Right now you can just replace the argument on the settings manager, I felt like this was more developer friendly.

src/Sylius/Bundle/SettingsBundle/Resources/config/doctrine/model/Settings.orm.xml
pamil: IMHO `on-delete="CASCADE"` on `Parameter` is enough.

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
steffenbrem: What about we just let `Settings::getParameters()` return the plain values array? I don't see why you would want a whole `Parameter` collection. Only `Settings::addParameter` makes sense to me to accept real `Parameter` objects.

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
pamil: We can leave the default resolver concept as is, but for me the best solution will be to move resolver resolving responsibility from settings manager to resolver registry. 

Eg. if I call `$resolverRegistry->get('foo')` and it does not have `foo` resolver, it should return the default one (it can be done as a decorator, so there's no need to write a custom registry). This way settings manager is not aware of that logic (and there's less dependencies! :tada:) and there's a single point of truth for getting a settings resolver.

Moreover, if we will go with that approach, we won't need to register `default` resolver in the registry, so the issue with unregistering and registering again will never happen :)

steffenbrem: That sounds like a great solution! But if people wanted to change the default resolver they would need to change the service that decorates the resolver registry service right?

What do you think about adding a configuration for the `sylius_settings` extension to easily change the service used as the default resolver in the decorator for the resolver registry? I think it might be a nice feature to have and easy to implement.

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
pamil: I'm not too much into Sylius' Settings, at the first glance the ParameterInterface / Parameter / ParameterCollection seems overcomplicated as we transform it to an array and again from an array to parameter objects. Is there some reason why it isn't just a simple array inside Settings model (that can be mapped as `object` in Doctrine)?

steffenbrem: Yes it is indeed too much for something so simple. If we want to make it simpeler we would need, as you were saying, remove the parameter model all together and maybe make is a serialized array inside the settings model. While this is faster and might make it simpelere you cannot easily extend a single parameter. Right now you can override the `Parameter` model and do specific stuff there. 

It really depends what you would prefer. If extension of the parameter model is not required, then we can get rid of the parameter factory, it's model etc.

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
pamil: The configuration option you've mentioned before is a good solution. Anyway, there comes another one:

 - let's say the default resolver is named `sylius.settings.default_resolver`, then the end user can "decorate" it without passing the "decorated" service by:

```xml
<service id="app.settings.default_resolver" class="Foo\Bar" decorates="sylius.settings.default_resolver">
  <argument type="service" id="custom.service" />
</service>
```

It's quite difficult to come up with a name for aliased services (like `basic` and `default` feels similar), but if we will name it meaningfully, that can be a nice solution too.

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
steffenbrem: My preference would go to letting `Settings` extend `ArrayObject` or implementing `ArrayAccess` etc. by itself and simply map `parameters` as an array in the Doctrine mapping. The API to work with settings would then be:

```php
// set parameter
$settings["new_parameter"] = "Hello";

// get plain array, to be used when transforming etc.
$parameters = $settings->getArrayCopy();

// replace the transformed parameters
$settings->exchangeArray($parameters);
```

pamil: I can't come up with an idea for specific stuff that can be done on Parameter level but not on Settings level. It will simplify all the things with introducing just a little bit of restrictions. Overriding `Parameter` also means that the custom stuff is done on each one.

Reducing that complexity we get rid of all Parameter* models, tricky relations and we reduce all the issues related to transforming array <=> ParameterCollection/Parameter.

WDYT? /cc @pjedrzejewski what was behind the inital idea?

steffenbrem: @pjedrzejewski ping

pjedrzejewski: We discussed with @pamil a bit in the office and I think it would be indeed good idea to simplify this whole thing and remove Parameter model. Data is serialized so it cannot be queried against anyway. Having Settings entity with parameters array field is enough. :+1: Good idea!

steffenbrem: @pjedrzejewski Great! I will roll out all feedback today and rebase it.

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
steffenbrem: :+1:

src/Sylius/Bundle/SettingsBundle/Manager/SettingsManager.php
steffenbrem: They are reduced now.

steffenbrem:
Waiting for the build status, but everything is in place now. The parameter model and everything related to it is removed in favor of storing parameters as key-value array on the settings model. Another important change is the use of a doctrine listener to transform the parameters instead of doing that in the `load` and `save` methods of the `SettingsManager`. This is a better approach since the previous solution can cause bugs when flushing the UOW before calling `save` on the `SettingsManager`, you do not have this problem now.

steffenbrem:
Anyone knows why the build is failing?

gperdomor:
@steffenbrem the problem is related to cart

lchrusciel:
Your problems can be connected to #3854 or #4444

steffenbrem:
@pjedrzejewski Could you try and rerun the CI build? I think PR #4444 has fixed it.

pjedrzejewski:
@steffenbrem You still need to rebase it in order to get these changes. ;)

steffenbrem:
@lchrusciel I don't think it is connected to #3854, because everything was green a few days ago. Not sure why it still keeps failing on the `CartSummaryPage`. I have no idea what changed might affect the cart.

pamil:
@steffenbrem I'm just curious, could you temporarily set `twig.strict_variables` to `false` in `app/config/config_test_cached.yml`? We've found it today, that there were fails on `test` environment which was silenced in `test_cached`, it may be something related to this issue.

Anyway, there's a lot of failing tests before the fatal error, do you have any of those locally (will need check both for `test` and `test_cached`). I'll send a fix for that fatal error in the morning :tada:

steffenbrem:
@pamil I will try that, but I cannot test it locally since I cannot get Selenium to work locally (I get `The path to the driver executable must be set by the webdriver.chrome.driver`). Otherwise it would be much easier for me to work out what might be wrong.

steffenbrem:
@pamil @pjedrzejewski I already know what the problem is. When adding an item to a cart it expects the `default_tax_zone` property to be set. And since we throw an exception there, it will fail.

In the old `SettingsManager` parameters where transformed inside the settings manager, and when you would set a transformer it would set that parameter correctly. But since the transforming now only happens when loading and saving settings to the DB the parameter is now missing, hence the `ParameterNotFoundException` and the failing of the Behat tests.

To be more explicit, it happens in `sylius_taxation` schema. For reference I will paste it's settings code here:
```php
$builder
    ->setOptional([
        'default_tax_zone',
    ])
    ->setAllowedTypes([
        'default_tax_zone' => ['null', ZoneInterface::class],
    ])
    ->setTransformer('default_tax_zone', new ObjectToIdentifierTransformer($this->zoneRepository))
;
```

The issue it that the `default_tax_zone` is defined as optional, so when calling `$settingsBuilder->resolve([])` it will not add the optional `default_tax_zone` field to the resolved parameters.

Should we leave the exception on the `Settings::get()` method when trying to get a undefined parameter or should we return `null`? Or should we just map over all defined settings in the `SettingsManager::load` and fill in the missing parameters ourself (and set their value to `null`)?

What is the best approach for this?

michalmarcinkowski:
@steffenbrem with Selenium, you just need ChromeDriver to get it work (see [this article](http://www.kenst.com/2015/03/installing-chromedriver-on-mac-osx/)).

Regarding the settings, I'm for the last (strict) solution, but only if it will not require to do any 'hacks' in the code or introduce some inconvenient extra logic (I'm not familiar with implementation details, so that's just my random thought).

steffenbrem:
@michalmarcinkowski Well it is kinda hacky, I will loop over `$settingsBuilder->getDefinedOptions()` and when it is not present on the `$parameters` array I will set it with a value of `null`. But that is just the way the options resolver component works. 

If you want no extra code the schema should use `setDefaults` instead of `setDefined`, and if you then use `setDefined` it will give you an undefined index error when doing `echo $settings['undefined_option']` for example. It really depends on how you want it to behave, but those two options are the only ones I see.

steffenbrem:
Yeey it's green again :tada:

steffenbrem:
@Sylius/core-team I've rebased it again. It is ready :+1:

src/Sylius/Bundle/SettingsBundle/DependencyInjection/Compiler/RegisterResolversPass.php
```diff
@@ -0,0 +1,46 @@
...
+ * @author Steffen Brem <steffenbrem@gmail.com>
+ */
+class RegisterResolversPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('sylius.settings.resolver_registry')) {
```
pjedrzejewski: For registries we use convention ``sylius.registry.X``.

src/Sylius/Bundle/SettingsBundle/Controller/SettingsController.php
```diff
@@ -59,35 +52,29 @@ public function showAction(Request $request)
...
      * @param Request $request
-     * @param string  $namespace
      *
      * @return Response
      */
-    public function updateAction(Request $request, $namespace)
+    public function updateAction(Request $request)
     {
-        $this->isGrantedOr403($namespace);
+        $schemaAlias = $request->get('schema');
```
pjedrzejewski: That should come from request attributes I think, not from query/post. ``$request->attributes->...``?

src/Sylius/Bundle/CoreBundle/Resources/config/app/main.yml
```diff
@@ -13,7 +13,6 @@ twig:
 
 doctrine_cache:
     providers:
-        sylius_settings: %sylius.cache%
```
pjedrzejewski: Why?

src/Sylius/Bundle/SettingsBundle/Doctrine/EventListener/ParameterTransformerListener.php
```diff
@@ -0,0 +1,132 @@
...
+    {
+        $registry = $this->container->get('sylius.settings.schema_registry');
+        $schema = $registry->get($settings->getSchemaAlias());
+
+        $settingsBuilder = new SettingsBuilder();
+        $schema->buildSettings($settingsBuilder);
+
+        return $settingsBuilder->getTransformers();
+    }
+}
```
pjedrzejewski: I have a feeling that this (not this specific implementation, my idea in general ;p) will hurt us in the long-run. I want to move default tax zone setting to Channel (see #4446), is there any other use-case for transformers at the moment? I think we could safely store codes of specific entities and do transformation/comparison whenever needed. I am really reluctant to add more doctrine listeners. Post load is called on every entity...

src/Sylius/Bundle/SettingsBundle/Controller/SettingsController.php
```diff
@@ -59,35 +52,29 @@ public function showAction(Request $request)
...
 
-        $manager = $this->getSettingsManager();
+        $this->isGrantedOr403($schemaAlias);
```
pjedrzejewski: I am not sure about the naming, ``namespace`` was not perfect, but the goal was to separate ``schema`` (how things are structured) from namespace (where they are saved). I know we have resolvers, but ``load(schema)`` can be confusing. SchemaResolvers make sense - absolutely. Perhaps we could come up with something more clear for the loading - ``type`` of settings? Not sure.

src/Sylius/Bundle/SettingsBundle/Model/SettingsInterface.php
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * @return string
+     */
+    public function getNamespace();
+
+    /**
+     * @param string $namespace
+     */
+    public function setNamespace($namespace);
```
pjedrzejewski: Do we still need it then?

src/Sylius/Bundle/SettingsBundle/Resolver/DefaultSettingsResolver.php
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * @param RepositoryInterface $settingRepository
+     */
+    public function __construct(RepositoryInterface $settingRepository)
+    {
+        $this->settingRepository = $settingRepository;
```
pjedrzejewski: setting**s**Repository

src/Sylius/Bundle/SettingsBundle/Resolver/DefaultSettingsResolver.php
```diff
@@ -0,0 +1,57 @@
...
+            $criteria = [
+                'schemaAlias' => $schemaAlias,
+            ];
+
+            if (null !== $namespace) {
+                $criteria['namespace'] = $namespace;
+            }
+
+            return $this->settingRepository->findOneBy($criteria);
+        } catch (NonUniqueResultException $e) {
```
pjedrzejewski: This effectively couples us to Doctrine ORM. :/

src/Sylius/Bundle/SettingsBundle/Resources/config/doctrine/model/Settings.orm.xml
```diff
@@ -13,18 +13,18 @@
...
 
-        <field name="namespace" column="namespace" type="string" />
-        <field name="name" column="name" type="string" />
-        <field name="value" column="value" type="object" nullable="true" />
-
-        <unique-constraints>
-            <unique-constraint columns="namespace,name" name="name_idx" />
-        </unique-constraints>
+        <field name="schemaAlias" column="schema_alias" type="string" nullable="false"/>
+        <field name="namespace" type="string" nullable="true"/>
```
pjedrzejewski: So the namespace can be null now?

src/Sylius/Bundle/SettingsBundle/Controller/SettingsController.php
pjedrzejewski: Or why not just "name" of settings, lol? :D

src/Sylius/Bundle/SettingsBundle/Resolver/DefaultSettingsResolver.php
pamil: Anyway, `findOneBy` in the default implementation does not throw this exception.

src/Sylius/Bundle/SettingsBundle/Resources/config/doctrine/model/Settings.orm.xml
steffenbrem: Yes, this is for global settings, basically I switched out the schema alias with what previously was `namespace`. The `namespace` now is just used to isolate settings so you can, as the name says, namespace them. Is this wrong?

src/Sylius/Bundle/SettingsBundle/Resolver/DefaultSettingsResolver.php
steffenbrem: I wil remove that try block. Since the namespace and schema alias already have a unique constraint I only need to properly send the query containing the schema and the namespace. Right now it skips the namespace criteria when it's null, but I think that is wrong and we need to always send it.

src/Sylius/Bundle/SettingsBundle/Resolver/DefaultSettingsResolver.php
steffenbrem: :+1:

src/Sylius/Bundle/SettingsBundle/Resolver/DefaultSettingsResolver.php
pamil: @steffenbrem good point :+1:

src/Sylius/Bundle/SettingsBundle/Model/SettingsInterface.php
steffenbrem: Hmm yeah I think we could safely remove this from the interface, same for `setSchemaAlias`? Since they should be immutable. Should we keep the setters on the concrete classes? Or should I look into constructor injection and fully remove those setters and use a factory method for this.

src/Sylius/Bundle/SettingsBundle/Controller/SettingsController.php
steffenbrem: I think `name` a good name.  Not sure if it's descriptive enough though, since you can easily mix it with thinking you are loading by `namespace`, which you are obviously not doing.

src/Sylius/Bundle/SettingsBundle/Resolver/DefaultSettingsResolver.php
steffenbrem: Well I updated my comment a bit about the unique constraint on the `schema_alias,namespace` combination. Since when you would use theme settings, you most likely only have `schema_alias` defined (and leave namespace on `null`) and then make a one-directional many-to-one (basically a many-to-many in doctrine with a unique constraint) with a `Theme` or whatever relationship you want to have with settings.

So in short, I will remove the unique constraint from the model. Is that ok?

src/Sylius/Bundle/SettingsBundle/Doctrine/EventListener/ParameterTransformerListener.php
steffenbrem: Well theme settings is a good use case, also the only one I can think of. With themes you don't really want to roll out your own separate settings management system, so re-using SettingsBundle is really good here. But if we want to use SettingsBundle with theme settings you want to be able to transform setting values. This is because themes can have settings that sets a specific `Taxon` for example, or a theme setting where you can link a `Page` to it. With data transformers on the settings bundle level it is very easy to implement all that for themes. Combine this with the resolver based schema loading in this PR and I think you can build settings for the craziest things. 

I am very biased to themes, but that is because I wanted to re-use the settings bundle for my own theming system. But I think it a good use case for Sylius too.

About the performance, I don't know if it's much overhead, I haven't profiled it yet. I think it is negligible, but I may be wrong here.

src/Sylius/Bundle/CoreBundle/Resources/config/app/main.yml
steffenbrem: I removed this because you can't be sure if the given `schemaAlias` is the only one. Since you now have settings resolvers it is not simply `namespace` anymore to make it unique. We could cache them by settings id now? But doesn't doctrine already provide us with an identity map and have this build in?

src/Sylius/Bundle/SettingsBundle/Controller/SettingsController.php
steffenbrem: :+1: Makes sense

src/Sylius/Bundle/SettingsBundle/DependencyInjection/Compiler/RegisterResolversPass.php
steffenbrem: :+1:

pjedrzejewski: ``sylius.registry.settings_resolver``, ``sylius.registry.settings_schema``, we generally avoid prefixing services with bundle name, to keep things consistent with:

``sylius.repository.tax_rate``, ``sylius.manager.shipping_category``, ``sylius.controller.product``, etc. As developer using Sylius you don't want to think from which bundle the service comes.

src/Sylius/Bundle/CoreBundle/Resources/config/app/main.yml
pjedrzejewski: OK, makes sense.

src/Sylius/Bundle/SettingsBundle/DependencyInjection/Compiler/RegisterResolversPass.php
steffenbrem: Okay! :+1:

steffenbrem:
Is this something to be merged in `0.18`? I've read that Sylius is now in stabilisation fase, I would like to now what is the planning for this PR, since I need it in a short term.

michalmarcinkowski:
@steffenbrem we will try to do the final review this week and get it merged ASAP :)

steffenbrem:
@michalmarcinkowski Any news if this can get squashed and merged?

pjedrzejewski:
Thanks for your work Steffen! I still have some ideas how we can improve Settings, but this is definitely big step in good direction. :+1:

pjedrzejewski:
Ouch, this should have been squashed indeed. :/ Although the commit history is quite clean except 2 commits, so let's keep it this way.

jmontoyaa:
Can anyone update the documentation?

PR #4247, opened 2016-02-21
[Resource][Paginator] - setMaxPerPage is set before setCurrentPage to get proper results

nakashu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no|
| Deprecations? | no
| License       | MIT

Fixing yesterday fix, sorry -1:
MaxPerPage needs to be set before current page, so proper number of pages is calculated and current page would be set successfully.


pjedrzejewski:
Thanks @nakashu! :+1: Good catch!

PR #4252, opened 2016-02-22
[Addressing] Country name to code converter extraction

lchrusciel:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

### DIFF
--- a/etc/behat/services/contexts.xml
+++ b/etc/behat/services/contexts.xml
@@ -46,6 +46,7 @@
         <service id="sylius.behat.context.setup.geographical" class="%sylius.behat.context.setup.geographical.class%" scope="scenario">
             <argument type="service" id="sylius.factory.country" container="symfony" />
             <argument type="service" id="sylius.repository.country" container="symfony" />
+            <argument type="service" id="sylius.converter.country_name" container="symfony" />
             <tag name="sylius.behat.context" />
         </service>
 
--- a/src/Sylius/Behat/spec/Context/Setup/GeographicalContextSpec.php
+++ b/src/Sylius/Behat/spec/Context/Setup/GeographicalContextSpec.php
@@ -13,6 +13,7 @@ namespace spec\Sylius\Behat\Context\Setup;
 
 use Behat\Behat\Context\Context;
 use PhpSpec\ObjectBehavior;
+use Sylius\Component\Addressing\Converter\CountryNameConverterInterface;
 use Sylius\Component\Addressing\Model\CountryInterface;
 use Sylius\Component\Resource\Factory\FactoryInterface;
 use Sylius\Component\Resource\Repository\RepositoryInterface;
@@ -22,9 +23,12 @@ use Sylius\Component\Resource\Repository\RepositoryInterface;
  */
 class GeographicalContextSpec extends ObjectBehavior
 {
-    function let(FactoryInterface $countryFactory, RepositoryInterface $countryRepository)
-    {
-        $this->beConstructedWith($countryFactory, $countryRepository);
+    function let(
+        FactoryInterface $countryFactory,
+        RepositoryInterface $countryRepository,
+        CountryNameConverterInterface $nameToCodeConverter
+    ) {
+        $this->beConstructedWith($countryFactory, $countryRepository, $nameToCodeConverter);
     }
 
     function it_is_initializable()
@@ -42,7 +46,8 @@ class GeographicalContextSpec extends ObjectBehavior
         RepositoryInterface $countryRepository,
         CountryInterface $australia,
         CountryInterface $china,
-        CountryInterface $france
+        CountryInterface $france,
+        CountryNameConverterInterface $nameToCodeConverter
     ) {
         $countryFactory->createNew()->willReturn($australia, $china, $france);
 
@@ -53,6 +58,10 @@ class GeographicalContextSpec extends ObjectBehavior
         $countryRepository->add($australia)->shouldBeCalled();
         $countryRepository->add($china)->shouldBeCalled();
         $countryRepository->add($france)->shouldBeCalled();
+
+        $nameToCodeConverter->convertToCode('Australia')->willReturn('AU');
+        $nameToCodeConverter->convertToCode('China')->willReturn('CN');
+        $nameToCodeConverter->convertToCode('France')->willReturn('FR');
 
         $this->storeShipsTo('Australia', 'China', 'France');
     }
--- a/src/Sylius/Component/Addressing/Converter/CountryNameConverterInterface.php
+++ b/src/Sylius/Component/Addressing/Converter/CountryNameConverterInterface.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Addressing\Converter;
+
+/**
+ * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
+ */
+interface CountryNameConverterInterface
+{
+    /**
+     * @param string $name
+     * @param string $locale
+     *
+     * @return string
+     *
+     * @throws \InvalidArgumentException If name is not found in country code registry.
+     */
+    public function convertToCode($name, $locale = 'en');
+}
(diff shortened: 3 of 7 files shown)


src/Sylius/Behat/CountryNameToCodeConverterInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
+ */
+interface CountryNameToCodeConverterInterface
+{
+    /**
+     * @param string $name
+     *
+     * @return string
+     *
+     * @throws \InvalidArgumentException If name is not found in country code registry.
```
Zales0123: Should the ``@throws`` docblock be in interface? I think it's an implementation detail.

src/Sylius/Behat/CountryNameToCodeConverterInterface.php
```diff
@@ -0,0 +1,26 @@
...
+     */
+    public function convert($name);
```
michalmarcinkowski: `convertToCode($name)`

src/Sylius/Behat/CountryNameToCodeConverterInterface.php
```diff
@@ -0,0 +1,26 @@
...
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace Sylius\Behat;
```
michalmarcinkowski: Is this class really specific to Behat?

src/Sylius/Behat/CountryNameToCodeConverterInterface.php
pamil: IMHO the interface is right place for it, but I'd comment it with "If conversion fails"

src/Sylius/Behat/CountryNameToCodeConverterInterface.php
lchrusciel: It is used only in Context setups. Where should it be placed? CoreBundle?

src/Sylius/Behat/CountryNameToCodeConverterInterface.php
Zales0123: :+1:

etc/behat/services.xml
```diff
@@ -31,6 +32,8 @@
             <argument>%mink.parameters%</argument>
         </service>
 
+        <service id="sylius.behat.converter.country_name_to_code" class="%sylius.behat.converter.country_name_to_code.class%" scope="scenario" public="false" />
```
pamil: No need for scope scenario as the service and its all dependencies (none) are stateless.

src/Sylius/Behat/CountryNameToCodeConverter.php
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
+ */
+final class CountryNameToCodeConverter implements CountryNameToCodeConverterInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function convert($name)
```
michalmarcinkowski: `convertToCode($name, $locale = 'en');`

etc/behat/services/contexts.xml
```diff
@@ -46,6 +46,7 @@
         <service id="sylius.behat.context.setup.geographical" class="%sylius.behat.context.setup.geographical.class%" scope="scenario">
             <argument type="service" id="sylius.factory.country" container="symfony" />
             <argument type="service" id="sylius.repository.country" container="symfony" />
+            <argument type="service" id="sylius.behat.converter.country_name_to_code" container="symfony" />
```
pamil: The service is defined in default, "behat", container, it should throw an exception if trying to get it from symfony container.

src/Sylius/Behat/CountryNameToCodeConverterInterface.php
lchrusciel: I think, that @Zales0123 is right. As far, as we don't depend on this exception it is an implementation detail. It should be placed in class.

src/Sylius/Behat/CountryNameToCodeConverterInterface.php
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
+ */
+interface CountryNameToCodeConverterInterface
```
michalmarcinkowski: `CountryNameConverterInterface`

michalmarcinkowski:
:+1: for extracting, after this is merged we can reuse it in contexts, fixtures and few other places in Sylius also :)

src/Sylius/Behat/CountryNameToCodeConverterInterface.php
pamil: So what's the other way to tell the user that given country name cannot be transformed? It's one of the obvious cases that should be a part of a contract, otherwise there is no way to handle an transformation error.

michalmarcinkowski: The interface know nothing about the exceptions since it's only abstraction layer. The exceptions are dependent on implementation and should be declared there.

src/Sylius/Behat/CountryNameToCodeConverterInterface.php
michalmarcinkowski: It's not dependent on `Core`, so I would place it in `Addressing` **component**.

src/Sylius/Behat/CountryNameToCodeConverterInterface.php
pamil: It knows nothing, but we can declare, that implementations will throw given type of exception if something bad happens, so the interface consumer can catch it and handle it. Throwing an exception is one of the ways to communicate between objects just like returning a value.

src/Sylius/Behat/CountryNameToCodeConverterInterface.php
pjedrzejewski: :+1: for putting it in the component.

src/Sylius/Behat/CountryNameToCodeConverterInterface.php
michalmarcinkowski: I agree that while defining contract we can specify which exception should be thrown (it's a pity that PHP doesn't enforce it on implementations...). In that particular case it is totally valid to specify the exception :+1:

etc/behat/services.xml
```diff
@@ -17,6 +17,7 @@
         <import resource="services/pages.xml"/>
     </imports>
     <parameters>
+        <parameter key="sylius.behat.converter.country_name_to_code.class">Sylius\Behat\CountryNameToCodeConverter</parameter>
```
michalmarcinkowski: This should be removed

src/Sylius/Behat/spec/Context/Setup/GeographicalContextSpec.php
```diff
@@ -22,9 +23,9 @@
  */
 class GeographicalContextSpec extends ObjectBehavior
 {
-    function let(FactoryInterface $countryFactory, RepositoryInterface $countryRepository)
+    function let(FactoryInterface $countryFactory, RepositoryInterface $countryRepository, CountryNameConverterInterface $nameToCodeConverter)
```
michalmarcinkowski: To long line

src/Sylius/Component/Addressing/Converter/CountryNameConverterInterface.php
```diff
@@ -0,0 +1,27 @@
...
+interface CountryNameConverterInterface
+{
+    /**
+     * @param string $name
+     *
+     * @return string
+     *
+     * @throws \InvalidArgumentException If name is not found in country code registry.
+     */
+    public function convertToCode($name);
```
michalmarcinkowski: Should be `($name, $locale)`

src/Sylius/Component/Addressing/Converter/CountryNameConverterInterface.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    /**
+     * @param string $name
+     * @param string $locale
+     *
+     * @return string
+     *
+     * @throws \InvalidArgumentException If name is not found in country code registry.
+     */
+    public function convertToCode($name, $locale = 'en');
```
michalmarcinkowski: We should not define defaults on interface.

lchrusciel:
It is ready to merge

michalmarcinkowski:
Thanks Łukasz! :+1:

Issue #4248, opened 2016-02-22
[review] - average rating NULL or 0 - model vs migration inconsistency


nakashu:
Generated migration for product.average_rating is `DEFAULT NULL` and model and mapping are for NOT NULL.

What should be fixed, migration or model?

Migration:
```
$this->addSql('ALTER TABLE sylius_product ADD average_rating DOUBLE PRECISION DEFAULT NULL');
```

Mapping Product.orm.xml
```
<field name="averageRating" column="average_rating" type="float" />
```

Model Product.php
```
protected $averageRating = 0;
```

pjedrzejewski:
Migration, it should be set to 0 I believe. Thanks for reporting this bug!

michalmarcinkowski:
Fixed in #4254

PR #4254, opened 2016-02-22
[review] - migration fix to be consistent with model

nakashu:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #4248
| License       | MIT


app/migrations/Version20160205083520.php
```diff
@@ -30,7 +30,7 @@ public function up(Schema $schema)
         $this->addSql('CREATE TABLE sylius_product_review (id INT AUTO_INCREMENT NOT NULL, product_id INT NOT NULL, author_id INT DEFAULT NULL, title VARCHAR(255) NOT NULL, rating INT NOT NULL, comment LONGTEXT DEFAULT NULL, status VARCHAR(255) NOT NULL, created_at DATETIME NOT NULL, updated_at DATETIME DEFAULT NULL, INDEX IDX_C7056A994584665A (product_id), INDEX IDX_C7056A99F675F31B (author_id), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB');
         $this->addSql('ALTER TABLE sylius_product_review ADD CONSTRAINT FK_C7056A994584665A FOREIGN KEY (product_id) REFERENCES sylius_product (id) ON DELETE CASCADE');
         $this->addSql('ALTER TABLE sylius_product_review ADD CONSTRAINT FK_C7056A99F675F31B FOREIGN KEY (author_id) REFERENCES sylius_customer (id)');
-        $this->addSql('ALTER TABLE sylius_product ADD average_rating DOUBLE PRECISION DEFAULT NULL');
+        $this->addSql('ALTER TABLE sylius_product ADD average_rating DOUBLE PRECISION NOT NULL');
```
michalmarcinkowski: How about ...DOUBLE PRECISION NOT NULL **DEFAULT 0**?

nakashu: Can change it, but this is how doctrine would generate it on its own, and also how the mapping is set currently

nakashu: changed

pjedrzejewski:
Thanks @nakashu! I am not sure we should define default values in Doctrine mapping, but I haven not idea for a better fix atm.

PR #4259, opened 2016-02-22
[WIP][SettingBundle] Introduce Setting model

tuka217:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #4231
| License       | MIT
| Doc PR        | -


steffenbrem:
I already have something for this, see #4246.

src/Sylius/Bundle/SettingsBundle/Controller/SettingsController.php
```diff
@@ -46,7 +46,7 @@ public function showAction(Request $request)
         } catch (MissingOptionsException $e) {
             // When a Settings is not persisted yet, it won't have any initial value in database,
             // so we create a new empty instance.
-            $settings = new Settings([]);
+            $settings = new Settings([], '', null);
```
steffenbrem: With this way you cannot override the Settings model, since it is hardcoded in the controller. Wouldn't it be better to use the factory here or just let the settings manager make sure you have a proper settings model, and inject the settings factory there?

pjedrzejewski:
Will be implemented in #4246.

PR #4167, opened 2016-02-16
Remove payment methods configuration

Arminek:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | no
| License       | MIT

PR #4261, opened 2016-02-22
Simplify Sylius configuration and parameters.yml.dist

pjedrzejewski:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

I will add more details later cause I am leaving the office in 2 mins, but long story short: I am upset everytime Composer asks me to fill all these parameters, when almost none of them is required. There is also no point in using ``sylius.`` prefix for parameters, we should stick to symfony-standard as much as possible. Another thing is that not everyone is using ECB, Open Exchange Rates, Amazon/Facebook/Google login with Sylius, same as with all payment gateways (removed recently in #4167), so this should be well documented but not included in the default configuration.

Probably there will be couple fixes needed but I will wait for the build.


michalmarcinkowski:
:+1: for making the file thinner. Let's add cookbook article about configuring OAuth and it's good to merge :)

liverbool:
:+1:

src/Sylius/Bundle/CoreBundle/Kernel/Kernel.php
```diff
@@ -112,7 +112,6 @@ public function registerBundles()
             new \Payum\Bundle\PayumBundle\PayumBundle(),
             new \JMS\SerializerBundle\JMSSerializerBundle(),
             new \JMS\TranslationBundle\JMSTranslationBundle(),
-            new \HWI\Bundle\OAuthBundle\HWIOAuthBundle(),
```
pamil: Composer should be updated too.

pjedrzejewski: Done.

UPGRADE.md
```diff
@@ -3,6 +3,27 @@ UPGRADE
 
 ## From 0.17 to 0.18.x
 
+### Application
+
+* Moved some of the parameters our of parameters.yml.dist file, please check your configurations
```
michalmarcinkowski: `the parameters our of` -> `the parameters out of`

src/Sylius/Behat/Context/Ui/ProductContext.php
```diff
@@ -69,7 +69,7 @@ public function iShouldBeAbleToAccessProduct(ProductInterface $product)
     {
         $this->productShowPage->tryToOpen(['product' => $product]);
 
-        expect($this->productShowPage->isOpen(['product' => $product]))->toBe(true);
+        $this->productShowPage->isOpen(['product' => $product]);
```
michalmarcinkowski: `Assert::true`

composer.json
```diff
@@ -96,7 +95,8 @@
         "phpspec/phpspec": "^2.4",
         "phpunit/phpunit": "~4.1",
         "se/selenium-server-standalone": "^2.52",
-        "mikey179/vfsStream": "^1.6"
+        "mikey179/vfsStream": "^1.6",
+        "hwi/oauth-bundle": "~0.3"
```
michalmarcinkowski: Why moved to dev?

src/Sylius/Bundle/WebBundle/Resources/config/routing/frontend/main.yml
```diff
@@ -64,3 +64,24 @@ sylius_search:
...
+    resource: "@CmfMediaBundle/Resources/config/routing/image.xml"
+
+payum_capture:
+    resource: "@PayumBundle/Resources/config/routing/capture.xml"
+
+payum_notify:
+    resource: "@PayumBundle/Resources/config/routing/notify.xml"
+
+fos_oauth_server_token:
+    resource: "@FOSOAuthServerBundle/Resources/config/routing/token.xml"
```
michalmarcinkowski: Shouldn't be moved to the ApiBundle?

michalmarcinkowski:
Good job Paweł! 👍

peteward:
wow, this really was a huge BC break.

@pjedrzejewski, what was the reason behind moving these parameters into sylius.yml?

```
parameters:
    sylius.inventory.backorders_enabled: true
    sylius.inventory.tracking_enabled: true
    sylius.inventory.holding.duration: 15 minutes
    sylius.order.allow_guest_order: false
    sylius.order.pending.duration: 3 hours
    sylius.cache:
        type: file_system
    sylius.uploader.filesystem: sylius_image
    phpcr_backend:
        type: doctrinedbal
        connection: default
        caches:
            meta: doctrine_cache.providers.phpcr_meta
            nodes: doctrine_cache.providers.phpcr_nodes
    phpcr_workspace: default
```

I know you were trying to clean things up, but why put these somewhere that's difficult to override?

We used to include all of Sylius's core configuration in our app config.yml, but with our own parameters. A simplified version looks like:

```
imports:
    - { resource: parameters/parameters.php }
    - { resource: "@SyliusCoreBundle/Resources/config/app/main.yml" }
```

Now that these parameters have been hard-coded into the top of sylius.yml file, how do you suggest we override them?

UPGRADE.md
```diff
@@ -3,6 +3,27 @@ UPGRADE
...
+
+# etc.
+```
+
+After:
+
+```yaml
+%database_host%
+%locale%
+```
```
peteward: I would definitely add a section about HWIOAuth since you've removed both the routing and the firewall config which will break anyone depending on it.

You should highlight the changes in `security.yml` and `routing.yml` telling people to restore these in their own apps if they are using these features.

pjedrzejewski:
@peteward:

You can still override them with another file, but this sucks, I know. Perhaps we should include them in ``CoreBundle/Resources/config/app/parameters.yml`` and include before main parameters.yml file? So that we have defaults in Sylius and not in parameters.yml.dist, but you still can override them? WDYT?

That being said:

```
sylius.inventory.backorders_enabled: true
```
This one needs to be removed completely, because it is confusing and redundant with ``availableOnDemand`` flag on the variant.

```
sylius.inventory.tracking_enabled: true
```

Should be configurable in admin and per variant basis, so will be removed.

```
    sylius.inventory.holding.duration: 15 minutes
    sylius.order.allow_guest_order: false
    sylius.order.pending.duration: 3 hours
```

All is application setting, not suitable for parameters.yml at all, should be moved to admin.

```
    sylius.cache:
        type: file_system
```
This one is bad, because it is an array, but could make sense in parameters.yml.dist. Still I think we should make the installation as simple as possible, if you configure cache, so with my solution with ``parameters.yml`` file shipped with Sylius and included before your parameters.yml, you can override that easily. We can add cookbook articles - How to configure Redis cache for Sylius, etc.

```
    sylius.uploader.filesystem: sylius_image
    phpcr_backend:
        type: doctrinedbal
        connection: default
        caches:
            meta: doctrine_cache.providers.phpcr_meta
            nodes: doctrine_cache.providers.phpcr_nodes
    phpcr_workspace: default
```

Pretty much same as above, I'd use same approach as for params above. If you want to use PHPCR without doctrinedbal, you will need to do some configs anyway, so I don't think it is a good idea to have these in main parameters.yml. People are often confused what is PHPCR and what they should put there during the installation.

peteward:
I agree that there's a separation of your shop's settings and your hosting environment configuration, and I can see that the 3 you highlighted will move to the admin UI settings at some point.

```
sylius.inventory.backorders_enabled: true
```

This broke for us because it's used in `AvailabilityChecker` which we use.

We cache in Redis so defaulting to `file_system` also broke our app.

It depends how you're planning the installation process, and I'm sure you have some ideas, but a lot of the infrastructure stuff could be wrapped up into something more user-friendly `What is your caching mechansim?: File (default), Redis, etc`.  You could also have `Just use defaults` as an option.

You can also separate parameters into different files - then only use Incenteev script on the ones you want composer to ask you about and write some other script for others.

We split our configuration into 3 files:
* `environment.yml` we have `live` and `non-live` versions of this which are installed based on which server you are on
* `server.yml` we have specific version for `production`, `testing`, `pr`, `dev` where params are different for each env.
* `main.yml` stuff that doesn't vary by server and environment (much of which will move to settings later). 


Having said all this, in the short term it would be great to at least move those parameters somewhere that they can be overridden then go forwards.

PR #4266, opened 2016-02-23
[Promotion] Removed "equal" flag from rules

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        | https://github.com/Sylius/Sylius-Docs/pull/418

Flag "equal" on promotions rules is redundant. For example, ``count: 5`` with ``equal: true`` is equivalent to ``count: 4`` without this flag. Moreover, in my opinion it's more natural to set promotion for "sth or more" rather than "for more than sth".


michalmarcinkowski:
:+1:

pjedrzejewski:
Thanks Mateusz!

PR #4257, opened 2016-02-22
[Behat][Promotion] Discount based on cart quantity

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

Based on https://github.com/Sylius/Sylius/pull/4249.
There is few things done in this PR connected with implementation of ``receiving_discount_based_on_cart_quantity`` feature:
- implementation of custom ``RuleFactory`` - as it have been done with ``ActionFactory`` but already in proper namespace
- added new promotions-checking step definition - changed ``my cart promotions`` to ``my discount``, changes in other scenarios will be done in separate PR
- set ``CartQuantityRuleChecker`` to always check ``>=`` condition - this change should be done for all checkers and it will be proposed and explained in separate PR

PR #4267, opened 2016-02-23
[Behat][Promotion] Discount based on items total

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

Further promotion scenarios - this time for ``item count`` rule.

### DIFF
--- a/features/promotion/receiving_discount_based_on_items_total.feature
+++ b/features/promotion/receiving_discount_based_on_items_total.feature
@@ -11,21 +11,21 @@ Feature: Receiving discount based on items total
 
     @ui
     Scenario: Receiving discount when buying items for the required total value
-        Given the promotion gives "€10.00" fixed discount to every order with item total at least "€80.00"
+        Given the promotion gives "€10.00" fixed discount to every order with items total at least "€80.00"
         When I add product "PHP T-Shirt" to the cart
         Then my cart total should be "€70.00"
         And my discount should be "-€10.00"
 
     @ui
     Scenario: Receiving discount when buying items for more than required total value
-        Given the promotion gives "€10.00" fixed discount to every order with item total at least "€90.00"
+        Given the promotion gives "€10.00" fixed discount to every order with items total at least "€90.00"
         When I add 2 products "PHP T-Shirt" to the cart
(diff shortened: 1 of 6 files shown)


features/promotion/receiving_discount_based_on_items_total.feature
```diff
@@ -5,35 +5,35 @@ Feature: Receiving discount based on items total
...
-        Given the store is operating on a single "France" channel
+        Given the store operates on a single channel in "France"
         And the store has a product "PHP T-Shirt" priced at "€80.00"
         And there is a promotion "Holiday promotion"
 
-    @todo
+    @ui
     Scenario: Receiving discount when buying items for the required total value
-        Given the promotion gives "€10.00" fixed discount to every cart with items total at least "€80.00"
+        Given the promotion gives "€10.00" fixed discount to every order with item total at least "€80.00"
```
michalmarcinkowski: `cart` was correct. Are you sure `items` is not proper?

Zales0123: *Primo:* As I've written [here](https://github.com/Sylius/Sylius/pull/4257#discussion_r53742662), I try to be consistent with previously implemented steps - if we would put ``cart`` here, we should put it everywhere else.
*Secundo:* I was following [rule type name](https://github.com/Sylius/Sylius/blob/43dc7551317d2c7919a29982ec9da7b401ec489a/src/Sylius/Component/Promotion/Model/RuleInterface.php#L21). Perhaps we should discuss, should it be named ``item total`` or ``items total`` ;)

michalmarcinkowski: 1. Agree
2. IMO we should use `items total` everywhere

Zales0123: Ok, so I will write here ``items total`` but rule type constant should be change in nearest future ;)

michalmarcinkowski:
:+1:

Issue #4225, opened 2016-02-19
[Behat][RFC] Proposal of LexicalContext


Zales0123:
We're all now participants of a great adventure with new behat scenarios and contexts ;) There are some really big changes and many other ideas are created over and over. Which is obviously good :smile: 

I've been thinking about implementing something like ``LexicalContext``, which could hold some reusable steps (or probably mainly ``@Transform``ers). Two examples:
- ``@Transform this|that {resourceName}`` - return last implemented/hold in ``sharedStorage`` resource with proper type
- ``@Transform it`` - any last implemted/hold in ``sharedStorage`` resource

It could be useful for steps like ``Given there is a promotion "Christmas Sale"/And it gives "€106.00" fixed discount to every order`` and many others.
Moreover, there could be some data transforming ``@Transform``ers. For example, usually we need to get integer from ``"€106.00"`` string, or float from ``10%``, so:
- ``@Transform "(€|£|\$)([^"]+)"``: ``"€106.00"`` -> ``10600``
- ``@Transform "([^"]+)%"``: ``"10%"`` -> ``0.1``

I think it could be quite useful and would avoid duplication of some logic/private methods ;)

pamil:
:+1:, but I'd like to keep those two related to shared storage apart to the other ones, just to let developers choose what they want to use.

lchrusciel:
:+1: Splitting SharedStorageContext and LexicalContext sounds reasonable too.

Zales0123:
Agreed :+1:

pjedrzejewski:
Nice idea and implementation. :+1:

PR #4269, opened 2016-02-23
[RFC][Behat] Extract reusable transformers

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #4225
| License       | MIT
| Doc PR        |

In order to avoid (current or future) duplications and make new behat scenarios even easier to write and it (they're already pretty natural :smile:), here is propose of two additional contexts, containing reusable transformers. Concept is widely described in related issue, just to make it clear about new features:

``SharedStorageContext``:
- ``it`` -> transform to last resource from ``SharedStorage``
- ``this|that|the {resource}`` -> transform to specific resource from ``SharedStorage``

``LexicalContext``:
- ``€103.00`` (string) -> ``10300`` (int)
- ``20%`` (string) -> ``0.2`` (float)

I'm sure we could find lots of more use cases that can be placed in one of this context. Moreover, I hope that by using them wisely, we could get rid of injecting ``SharedContext`` into most of contexts and minimize number of private functions.

### DIFF
--- a/etc/behat/services/contexts.xml
+++ b/etc/behat/services/contexts.xml
@@ -24,6 +24,7 @@
         <parameter key="sylius.behat.context.setup.shipping.class">Sylius\Behat\Context\Setup\ShippingContext</parameter>
         <parameter key="sylius.behat.context.setup.zone.class">Sylius\Behat\Context\Setup\ZoneContext</parameter>
         <parameter key="sylius.behat.context.setup.tax.class">Sylius\Behat\Context\Setup\TaxContext</parameter>
+        <parameter key="sylius.behat.context.setup.lexical.class">Sylius\Behat\Context\Setup\LexicalContext</parameter>
 
         <parameter key="sylius.behat.context.hook.doctrine_orm.class">Sylius\Behat\Context\Hook\DoctrineORMContext</parameter>
 
@@ -126,6 +127,10 @@
             <argument type="service" id="sylius.repository.tax_rate" container="symfony" />
             <argument type="service" id="sylius.repository.tax_category" container="symfony" />
             <argument type="service" id="sylius.repository.zone" container="symfony" />
+            <tag name="sylius.behat.context" />
+        </service>
+
+        <service id="sylius.behat.context.setup.lexical" class="%sylius.behat.context.setup.lexical.class%">
             <tag name="sylius.behat.context" />
         </service>
 
--- a/etc/behat/services/contexts.xml
+++ b/etc/behat/services/contexts.xml
@@ -25,6 +25,7 @@
         <parameter key="sylius.behat.context.setup.zone.class">Sylius\Behat\Context\Setup\ZoneContext</parameter>
         <parameter key="sylius.behat.context.setup.tax.class">Sylius\Behat\Context\Setup\TaxContext</parameter>
         <parameter key="sylius.behat.context.setup.lexical.class">Sylius\Behat\Context\Setup\LexicalContext</parameter>
+        <parameter key="sylius.behat.context.setup.shared_storage.class">Sylius\Behat\Context\Setup\SharedStorageContext</parameter>
 
         <parameter key="sylius.behat.context.hook.doctrine_orm.class">Sylius\Behat\Context\Hook\DoctrineORMContext</parameter>
 
@@ -131,6 +132,11 @@
         </service>
 
         <service id="sylius.behat.context.setup.lexical" class="%sylius.behat.context.setup.lexical.class%">
+            <tag name="sylius.behat.context" />
+        </service>
+
+        <service id="sylius.behat.context.setup.shared_storage" class="%sylius.behat.context.setup.shared_storage.class%">
+            <argument type="service" id="sylius.behat.shared_storage" container="symfony" />
             <tag name="sylius.behat.context" />
         </service>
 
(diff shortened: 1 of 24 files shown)


etc/behat/services/contexts.xml
```diff
@@ -24,6 +24,8 @@
         <parameter key="sylius.behat.context.setup.shipping.class">Sylius\Behat\Context\Setup\ShippingContext</parameter>
         <parameter key="sylius.behat.context.setup.zone.class">Sylius\Behat\Context\Setup\ZoneContext</parameter>
         <parameter key="sylius.behat.context.setup.tax.class">Sylius\Behat\Context\Setup\TaxContext</parameter>
+        <parameter key="sylius.behat.context.setup.lexical.class">Sylius\Behat\Context\Setup\LexicalContext</parameter>
```
pamil: IMHO new namespace `sylius.behat.context.transform.*` would be the most suitable for contexts that only transforms arguments

Zales0123: Agreed :+1: I couldn't come up with some good name, bad ``transform`` is simple and accurate :smile:

src/Sylius/Behat/spec/Context/Setup/LexicalContextSpec.php
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    function it_implements_context_interface()
+    {
+        $this->shouldImplement(Context::class);
+    }
+
+    function it_transforms_price_string_to_integer()
+    {
+        $this->getPriceFromString('10.00')->shouldReturn(1000);
```
pamil: It would be nice to have more cases covered as converting string to float and then float to int is quite unreliable.

pamil:
:+1:, we will have to use shared context directly only to set a resource after this PR gets merged! :)

src/Sylius/Behat/spec/Context/Setup/LexicalContextSpec.php
michalmarcinkowski: :+1: 

I would add check for at least the following values:
 - `0.10`
 - `0.30`

src/Sylius/Behat/Context/Transform/LexicalContext.php
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * @Transform /^"(?:€|£|\$)([^"]+)"$/
+     */
+    public function getPriceFromString($price)
+    {
+        return (int) round(($price * 100), 2);
+    }
+
+    /**
+     * @Transform /^"([^"]+)%"$/
```
lchrusciel: Won't it be better to use here regex which will match only floats? `([0-9]*\.)?[0-9]+`

src/Sylius/Behat/spec/Context/Transform/LexicalContextSpec.php
```diff
@@ -0,0 +1,48 @@
...
+    function it_implements_context_interface()
+    {
+        $this->shouldImplement(Context::class);
+    }
+
+    function it_transforms_price_string_to_integer()
+    {
+        $this->getPriceFromString('10.00')->shouldReturn(1000);
+        $this->getPriceFromString('0.30')->shouldReturn(30);
+        $this->getPriceFromString('0.0134')->shouldReturn(1);
```
lchrusciel: Are you sure, that this is expected behaviour? Maybe it would be better to throw some exception, that it cannot be casted to integer without losing precision

Zales0123: I think it's fine, there is no possibility to operate on "parts of cents/pennies" so it should be casted to ``1``, it's not an error. At least I think so :smile:

src/Sylius/Behat/Context/Transform/LexicalContext.php
Zales0123: :+1:

pamil: We can make it `/^"((?:\d+\.)?\d+)%"$/` :tada:

pamil: Anyway, same for price from string :)

src/Sylius/Behat/spec/Context/Transform/LexicalContextSpec.php
pamil: I'm :+1: for throwing an exception, $0.1456 is an invalid amount of money.

src/Sylius/Behat/Context/Transform/LexicalContext.php
lchrusciel: :+1:

src/Sylius/Behat/spec/Context/Transform/LexicalContextSpec.php
Zales0123: Is it?
stock

Zales0123: I think it's not, we just don't operate on such a small amounts.

pamil: For our usecase - it is. If we transform `0.0134` into `0.01` (`1` as int) and lose that `0.0034`, we should tell the user that it's a wrong amount.

michalmarcinkowski: If the system does not support such precision it should inform the developer by throwing exception and do not round it without asking. There is no point in writing "0.01234" in your scenario.

Zales0123: Ok, I can leave with this exception (you outnumber me :smile:)

pjedrzejewski:
Thanks Mateusz! :+1:

PR #4270, opened 2016-02-23
[Behat][Promotion] Shipping discount promotions

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

Issue #4229, opened 2016-02-19
[Promotion] ActionFactory should go to Core & ActionInterface::TYPE_FIXED_DISCOUNT too


pjedrzejewski:
Promotion component does not know any types of discounts, so:

* ActionFactory added in #4223 should go to the core component
* ActionInterface::TYPE_FIXED_DISCOUNT should be moved to FixedDiscountAction::TYPE

Zales0123:
We must remember to add missing promotion types' constants (for example for ``shipping discount``) and moreover use them everywhere where it should be used - as mentioned in [this discussion](https://github.com/Sylius/Sylius/pull/4275#discussion_r53915764)

PR #4275, opened 2016-02-24
[Behat][Promotion] Reapplying promotion on cart change

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets |
| License       | MIT
| Doc PR        |

PR is based on https://github.com/Sylius/Sylius/pull/4270.

### DIFF
--- a/features/promotion/reapplying_promotion_on_cart_change.feature
+++ b/features/promotion/reapplying_promotion_on_cart_change.feature
@@ -5,23 +5,23 @@ Feature: Reapplying promotion on cart change
     I want to have proper discount applied after every operation on my cart
 
     Background:
-        Given the store is operating on a single "France" channel
+        Given the store operates on a single channel in "France"
         And the store has a product "PHP T-Shirt" priced at "€100.00"
         And there is a promotion "Holiday promotion"
         And I am logged in customer
 
-    @todo
+    @ui
     Scenario: Not receiving discount on shipping after removing last item from cart
         Given the store has "DHL" shipping method with "€10.00" fee
         And the promotion gives "100%" percentage discount on shipping to every order
         And I have product "PHP T-Shirt" in the cart
         When I proceed selecting "DHL" shipping method
         And I remove product "PHP T-Shirt" from the cart
-        Then my cart total should be "€0.00"
-        And my cart shipping fee should be "€0.00"
-        And my discount should be "€0.00"
+        Then cart should be empty with no value
+        And there should be no shipping fee
+        And there should be no discount
 
-    @todo
+    @ui
     Scenario: Receiving discount on shipping after shipping method change
         Given the store has "DHL" shipping method with "€10.00" fee
         And the store has "FedEx" shipping method with "€30.00" fee
@@ -33,28 +33,28 @@ Feature: Reapplying promotion on cart change
         And my cart shipping fee should be "€30.00"
         And my discount should be "-€30.00"
 
-    @todo
+    @ui
     Scenario: Receiving discount after removing an item from the cart and then adding another one
         Given the store has a product "Symfony T-Shirt" priced at "€150.00"
         And the promotion gives "€10.00" fixed discount to every order
         And I had product "PHP T-Shirt" in the cart
-        But I removed this product from the cart
+        But I removed product "PHP T-Shirt" from the cart
         When I add product "Symfony T-Shirt" to the cart
         Then my cart total should be "€140.00"
         And my discount should be "-€10.00"
 
-    @todo
+    @ui
     Scenario: Not receiving discount when cart does not meet the required total value after removing an item
-        Given the promotion gives "€10.00" fixed discount to every cart with items total at least "€120.00"
+        Given the promotion gives "€10.00" fixed discount to every order with items total at least "€120.00"
         And I have 2 products "PHP T-Shirt" in the cart
         When I change "PHP T-Shirt" quantity to 1
         Then my cart total should be "€100.00"
-        And my discount should be "€0.00"
+        And there should be no discount
 
-    @todo
+    @ui
     Scenario: Not receiving discount when cart does not meet the required quantity after removing an item
-        Given the promotion gives "€10.00" fixed discount to every cart with quantity at least 3
+        Given the promotion gives "€10.00" fixed discount to every order with quantity at least 3
         And I have 3 products "PHP T-Shirt" in the cart
         When I change "PHP T-Shirt" quantity to 1
         Then my cart total should be "€100.00"
-        And my discount should be "€0.00"
+        And there should be no discount
--- a/src/Sylius/Behat/Context/Ui/CartContext.php
+++ b/src/Sylius/Behat/Context/Ui/CartContext.php
@@ -46,9 +46,8 @@ final class CartContext implements Context
 
     /**
      * @Given I added product :product to the cart
-     * @Given I had product :product in the cart
+     * @Given /^I (?:have|had) (product "([^"]+)") in the cart$/
      * @When I add product :product to the cart
-     * @When I have product :product in the cart
      */
     public function iAddProductToTheCart(ProductInterface $product)
     {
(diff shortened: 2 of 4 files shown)


features/promotion/reapplying_promotion_on_cart_change.feature
```diff
@@ -33,28 +33,28 @@ Feature: Reapplying promotion on cart change
...
         And my discount should be "-€30.00"
 
-    @todo
+    @ui
     Scenario: Receiving discount after removing an item from the cart and then adding another one
         Given the store has a product "Symfony T-Shirt" priced at "€150.00"
         And the promotion gives "€10.00" fixed discount to every order
         And I had product "PHP T-Shirt" in the cart
-        But I removed this product from the cart
+        But I removed product "PHP T-Shirt" from the cart
```
Zales0123: This is a temporary change. After merging https://github.com/Sylius/Sylius/pull/4269 we could use ``this product`` or ``it`` without injecting ``SharedStorage`` to ``CartContext``.

features/promotion/reapplying_promotion_on_cart_change.feature
```diff
@@ -5,23 +5,23 @@ Feature: Reapplying promotion on cart change
...
     Scenario: Not receiving discount on shipping after removing last item from cart
         Given the store has "DHL" shipping method with "€10.00" fee
         And the promotion gives "100%" percentage discount on shipping to every order
         And I have product "PHP T-Shirt" in the cart
         When I proceed selecting "DHL" shipping method
         And I remove product "PHP T-Shirt" from the cart
-        Then my cart total should be "€0.00"
-        And my cart shipping fee should be "€0.00"
-        And my discount should be "€0.00"
+        Then cart total should not be displayed
```
michalmarcinkowski: `displayed` is coupled to `UI` we should find better step definition. Let's brainstorm on that.
`Then cart should be empty with no value` - do anyone have a better idea?

features/promotion/reapplying_promotion_on_cart_change.feature
michalmarcinkowski: :+1:

src/Sylius/Behat/spec/Context/Ui/CartContextSpec.php
```diff
@@ -89,6 +89,22 @@ function it_throws_not_equal_exception_if_promotion_total_is_incorrect(CartSumma
...
+
+        $this->cartTotalShouldNotBeDisplayed();
+    }
+
+    function it_ensures_there_is_no_shipping_fee_info_on_the_page(CartSummaryPage $cartSummaryPage)
+    {
+        $cartSummaryPage->open()->shouldBeCalled();
+        $cartSummaryPage->getShippingTotal()->willThrow(new ElementNotFoundException('"shipping total" element is not present on the page'));
+
+        $this->thereShouldBeNoShippingFee();
```
michalmarcinkowski: We also need to check the case when the shipping fee will be found on the page. Same for the above.

src/Sylius/Component/Promotion/Factory/ActionFactory.php
```diff
@@ -63,4 +63,16 @@ public function createPercentageDiscount($percentage)
...
         return $action;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createPercentageShippingDiscount($percentage)
+    {
+        $action = $this->createNew();
+        $action->setType('shipping_discount');
```
michalmarcinkowski: We should extract it to the constant. Do we set up a convention for that?

src/Sylius/Behat/Context/Ui/CartContext.php
```diff
@@ -68,6 +69,7 @@ public function iAddProductToTheCartSelectingVariant($variant, ProductInterface
 
     /**
      * @Given I remove product :productName from the cart
+     * @Given I removed product :productName from the cart
```
michalmarcinkowski: Can we merge this with the previous? `@Given I (remove|removed) product :productName from the cart` WDYT?

Zales0123: :+1:

michalmarcinkowski: BTW. we should have:
```
@Given I removed ...
@When I remove ...
```

src/Sylius/Component/Promotion/Factory/ActionFactory.php
Zales0123: It should definitely be a constant, however I don't know when to put it :smile: Obvious choice is ``ShippingDiscountAction``, but it is in *Core* and for now ``ActionFactory`` is in *Promotion* component - which [should be changed](https://github.com/Sylius/Sylius/issues/4229).
So maybe the best solution would be to solve mentioned issue and then do it as it should be done ;)

src/Sylius/Behat/Context/Ui/CartContext.php
```diff
@@ -46,6 +46,7 @@ public function __construct(
 
     /**
      * @Given I added product :product to the cart
+     * @Given I had product :product in the cart
      * @When I add product :product to the cart
      * @When I have product :product in the cart
```
michalmarcinkowski: Do we have a use case when `When I have product ...` makes sense to be used instead of `When I add product ...`?

src/Sylius/Component/Promotion/Factory/ActionFactory.php
michalmarcinkowski: Please add a comment about this in the mentioned issue, so we remember to change that :)

features/promotion/reapplying_promotion_on_cart_change.feature
Zales0123: Good for me :+1:

features/promotion/reapplying_promotion_on_cart_change.feature
Zales0123: Ok, I lied :D we must have product name in this step, as ``removeProduct`` method use it to find product to remove on products index page.

src/Sylius/Behat/Context/Ui/CartContext.php
Zales0123: It used in [several steps](https://github.com/Sylius/Sylius/blob/master/features/shipping/apply_correct_shipping_fee_on_order.feature#L16), where having a product in cart should be presumption, not an action performed.

src/Sylius/Behat/spec/Context/Ui/CartContextSpec.php
Zales0123: You mean case when we're assuming there is no shipping fee, but there is?

michalmarcinkowski: Exactly :)

src/Sylius/Behat/Context/Ui/CartContext.php
michalmarcinkowski: It is used as `Given`, so what do you think about:
```
@Given I (added|have|had) product :product to the cart
@When I add product :product to the cart
```

Zales0123: :+1:

features/promotion/reapplying_promotion_on_cart_change.feature
michalmarcinkowski: So let's create a new method that will receive `Product` object :)

Zales0123: It would require to inject ``SharedStorage`` to ``CartContext`` and set product on it while adding... This is thing to discuss, maybe we could merge it as it is and make it perfect in separate, refactoring PR? ;)

michalmarcinkowski: But `this product`will not be transformed into `Product` by transformer in `SharedStorageContext`?

michalmarcinkowski: :+1:

pjedrzejewski:
Thanks Mateusz! :+1:

PR #4285, opened 2016-02-24
[Core][ProductAccessor] Use option code instead of name

okwinza:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no|yes
| BC breaks?    | no|yes
| Deprecations? | no|yes
| License       | MIT

At this moment this class only used in SearchBundle and it is wrong to match option by presentation since it might contain non-latin characters that cannot be used in YAML. 

It is better to use code field in this case.

This PR is based on #4237


pjedrzejewski:
Nice, let's wait for the base PR to be merged and tweaked. (my comment will affect this PR too - https://github.com/Sylius/Sylius/pull/4237/files#r54012174)

okwinza:
@pjedrzejewski updated & squashed. Should be g2g once related PR is merged.

okwinza:
@pjedrzejewski greeen :tada:

pjedrzejewski:
Thank you Oleg!

PR #4288, opened 2016-02-24
[WebBundle] Hide option values on product show page in backend

okwinza:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no|yes
| BC breaks?    | no|yes
| Deprecations? | no|yes
| License       | MIT

IMO this is not only redundant but also could(and will) explode if you have too many values attached to an option.


pjedrzejewski:
Exploding is never good, thanks Oleg! :D

PR #4289, opened 2016-02-24
[MailerBundle] Introduce SendGrid Adapter

gperdomor:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Fixed tickets | - |
| License | MIT |
| Doc PR | Sylius/Sylius-Docs#422 |

Introduce SendGrid Adapter.

Phpspec fails because an authentication error (sendgrid api key related).

### DIFF
--- a/src/Sylius/Bundle/MailerBundle/DependencyInjection/SyliusMailerExtension.php
+++ b/src/Sylius/Bundle/MailerBundle/DependencyInjection/SyliusMailerExtension.php
@@ -12,6 +12,7 @@
 namespace Sylius\Bundle\MailerBundle\DependencyInjection;
 
 use Sylius\Bundle\ResourceBundle\DependencyInjection\Extension\AbstractResourceExtension;
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
@@ -40,6 +41,17 @@ class SyliusMailerExtension extends AbstractResourceExtension
 
         foreach ($configFiles as $configFile) {
             $loader->load($configFile);
+        }
+
+        if ($config['sender_adapter'] == 'sylius.email_sender.adapter.sendgrid') {
+            if (!isset($config['sendgrid_api'])) {
+                throw new InvalidConfigurationException('You must configure a sendgrid_api for mailer adapter');
+            }
+
+            $container
+                ->getDefinition('sylius.email_sender.adapter.sendgrid')
+                ->addArgument($config['sendgrid_api'])
+            ;
         }
 
         $container->setAlias('sylius.email_sender.adapter', $config['sender_adapter']);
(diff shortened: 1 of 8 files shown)


src/Sylius/Bundle/MailerBundle/DependencyInjection/SyliusMailerExtension.php
```diff
@@ -42,6 +43,16 @@ public function load(array $config, ContainerBuilder $container)
             $loader->load($configFile);
         }
 
+        if ($config['sender_adapter'] == 'sylius.email_sender.adapter.sendgrid') {
+            if (is_null($config['sendgrid_api'])) {
```
michalmarcinkowski: `if(!isset($config['sendgrip_api']))` - this will check for `null` value also.

stloyd:
This should be optional not hard dependency IMO.

src/Sylius/Bundle/MailerBundle/DependencyInjection/SyliusMailerExtension.php
```diff
@@ -42,6 +43,16 @@ public function load(array $config, ContainerBuilder $container)
...
+                throw new InvalidConfigurationException('You must configure a sendgrid_api for mailer adapter');
+            }
+
+            $container
+                ->getDefinition('sylius.email_sender.adapter.sendgrid')
+                ->addArgument($config['sendgrid_api']);
```
michalmarcinkowski: `;` should be in a separate line

src/Sylius/Bundle/MailerBundle/Resources/config/services.xml
```diff
@@ -67,5 +68,7 @@
         <service id="sylius.email_sender.adapter.swiftmailer" class="%sylius.email_sender.adapter.swiftmailer.class%" parent="sylius.email_sender.adapter.abstract">
             <argument type="service" id="mailer" />
         </service>
+        <service id="sylius.email_sender.adapter.sendgrid" class="%sylius.email_sender.adapter.sendgrid.class%" parent="sylius.email_sender.adapter.abstract">
+        </service>
```
michalmarcinkowski: `<service id= ..... />` should work fine.

pjedrzejewski:
@gperdomor That's really cool, I am glad we finally we use of these adapters! That being said, I agree with @stloyd that this should not really go to the core. Perhaps that could be a separate bundle?

gperdomor:
@pjedrzejewski @stloyd create a new bundle only to add this adapter?... Hmmm... i don't know, sounds a little unnecessary to me, i don't like the idea, or i don't understand your idea xD

pjedrzejewski:
OK, maybe a bundle is a bit overkill, I guess we need to wait (can be a bit long :>) for Sylius plugins, haha. Everything that lands in core must be maintained and that includes the 1.0 compatibility promise. So I am hesitant to add more than necessary to this repo... I hope you understand. It's pretty late and I am out of ideas, let's discuss it tomorrow and maybe we can figure out something that is not an overkill. :>

gperdomor:
@pjedrzejewski sure, don't worry!

peteward:
similarly, we've got a Mandrill adapter that we could contribute when the best method to do so is identified.

pjedrzejewski:
@peteward just FYI, for Mandrill I always use https://github.com/AccordGroup/MandrillSwiftMailerBundle (or similar bundle).

peteward:
There's a good reason why we didn't, I just can't remember what it is right now...

PR #4296, opened 2016-02-24
[CoreBundle] Add enabled is true when retrieve latest products

Niiko:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

Add enabled is true on findLatest productRepository function.

I think we have to hide "disabled" products in frontend.

This PR will only "filter" results, so, we have to control access in ProductController:detailsAction.

The simplier would be to throw an exception "NotEnabled" or something like this. What do you think ?


pjedrzejewski:
Thanks @Niiko!

Issue #4293, opened 2016-02-24
[Behat] Add selenium-server-standalone in composer require-dev


Niiko:
Hi,

What about add selenium in require-dev ?

Using a package like this one : https://packagist.org/packages/se/selenium-server-standalone

michalmarcinkowski:
Sounds good for me :+1:

PR #4299, opened 2016-02-24
[Behat] Add selenium server standalone in require-dev

Niiko:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #4293 
| License       | MIT
| Doc PR        | -


pjedrzejewski:
What do you folks think? Sounds good to me.

pamil:
Looks good, we can also adjust Travis to use it instead of downloading it itself (`etc/travis/prepare-javascript`). It would be nice to have Chromedriver downloaded by Composer as well.

pjedrzejewski:
Thanks @Niiko! :+1:

PR #4311, opened 2016-02-25
[SearchBundle] Add boolOr filter for range filters for elasticsearch engine

Niiko:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
| License       | MIT
| Doc PR        | -

Indead, when you select two values in facets like price "from 0 to 20" and "from 20.01 to 50" the results can't find the correct results.

In fact, it seems to keep the last request :/

I also fix the missing $termFilters variable initialization


pjedrzejewski:
Thanks Nicolas! :+1:

PR #4328, opened 2016-02-26
[Customer] Added phone number to customer

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT
| Doc PR          |

### DIFF
--- a/src/Sylius/Bundle/UserBundle/Resources/config/doctrine/model/Customer.orm.xml
+++ b/src/Sylius/Bundle/UserBundle/Resources/config/doctrine/model/Customer.orm.xml
@@ -61,6 +61,8 @@
             </join-table>
         </many-to-many>
 
+        <field name="phoneNumber" column="phone_number" type="string" nullable="true" />
+
     </mapped-superclass>
 
 </doctrine-mapping>
--- a/app/migrations/Version20160229091707.php
+++ b/app/migrations/Version20160229091707.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Sylius\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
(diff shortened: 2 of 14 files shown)


michalmarcinkowski:
Migration file is missing.

aramalipoor:
:+1:

src/Sylius/Bundle/UserBundle/Resources/config/doctrine/model/Customer.orm.xml
```diff
@@ -61,6 +61,8 @@
             </join-table>
         </many-to-many>
 
+        <field name="phoneNumber" column="phone_number" type="string" nullable="true" />
```
Zales0123: Don't we need any validation configured for this field?

pjedrzejewski: We will add it later when we cover this with scenarios in frontend.

pjedrzejewski:
Thanks Grzesiu!

gperdomor:
Sorry, I'm late but this is my opinion:

What happen if we need handle multiple phone numbers?, mobile and home number for example. In that case, i don't like the idea of creation of a new entity <code>phoneNumber</code>, at least not in <code>UserBundle</code>, i think this should be managed this in the <code>CoreBundle</code> and leave the <code>UserBundle</code> more generic possible.

peteward:
I think this is a good change.

Customers have Phone numbers in the same way that they have email addresses, names and addresses. It's definitely not a Core attribute to say that a customer has a phone number.

If you want multiple phone numbers then perhaps this is an extension of your own on top of the existing phoneNumber?

PR #4331, opened 2016-02-26
Remove ObjectToIdentifierTransformer from  ShippingMethodChoiceType f…

videni:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |

Remove ObjectToIdentifierTransformer from  ShippingMethodChoiceType for ChoiceType handles it internally and add a indentifier option , so we can use shipment method code name instead id, the value of id changes every time, when we load fixtures, it is not good for API

### DIFF
--- a/src/Sylius/Bundle/ShippingBundle/Form/Type/ShippingMethodChoiceType.php
+++ b/src/Sylius/Bundle/ShippingBundle/Form/Type/ShippingMethodChoiceType.php
@@ -53,6 +53,15 @@ class ShippingMethodChoiceType extends AbstractType
     protected $repository;
 
     /**
+     * {@inheritdoc}
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ShippingBundle/Form/Type/ShippingMethodChoiceType.php
```diff
@@ -53,6 +53,15 @@ class ShippingMethodChoiceType extends AbstractType
...
 
     /**
+     * {@inheritdoc}
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        if ($options['multiple']) {
+            $builder->addModelTransformer( new CollectionToArrayTransformer());
+        }
+    }
```
peteward: missing new line

PR #4336, opened 2016-02-27
[WebBundle] Added graph to routing to solving state machine update issue.

ahmadrabie:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

I know that you are refactoring state machines but for now in Order backend there is some bugs, so I added graphs to routings to solve this issues.


pjedrzejewski:
Thank you Ahmad! :+1:

PR #4337, opened 2016-02-27
replace ThemeInterface with Theme entity in doctrine mapping

ahilles107:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | no
| License         | MIT
| Doc PR          | no

With ThemeInterface used in mapping we can't generate update sql - error: 

```
$ php app/console doctrine:schema:update --dump-sql


  [Doctrine\Common\Persistence\Mapping\MappingException]
  Class 'Sylius\Bundle\ThemeBundle\Model\ThemeInterface' does not exist
```

In profiler there is also error with mapping:
```
The target entity 'Sylius\Bundle\ThemeBundle\Model\ThemeInterface' specified on Sylius\Bundle\ThemeBundle\Model\Theme#parents is unknown or not an entity.
```

### DIFF
--- a/src/Sylius/Bundle/ThemeBundle/Resources/config/doctrine/model/Theme.orm.xml
+++ b/src/Sylius/Bundle/ThemeBundle/Resources/config/doctrine/model/Theme.orm.xml
@@ -29,7 +29,7 @@
         <field name="description" type="string" nullable="true" />
         <field name="authors" type="object" nullable="true" />
 
-        <many-to-many field="parents" target-entity="Sylius\Bundle\ThemeBundle\Model\ThemeInterface">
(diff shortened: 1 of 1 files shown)


steffenbrem:
I don't think this would be accepted, you won't be able easily override the `Theme` model this way. Not sure why you get this error, are you on master?

pjedrzejewski:
@steffenbrem is right. If you are using ThemeBundle standalone, make sure it is configured before DoctrineBundle, together with ResourceBundle. This should fix the error. Please close this issue when resolved, thanks!

ahilles107:
Yep, initializing it before DoctrineBundle solved problem. Thanks for help!

PR #4357, opened 2016-02-29
Fix various missing deps and not needed coupling to Symfony

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT
| Doc PR          | -

### DIFF
--- a/src/Sylius/Bundle/ThemeBundle/composer.json
+++ b/src/Sylius/Bundle/ThemeBundle/composer.json
@@ -36,14 +36,16 @@
         "phpunit/phpunit": "^4.1",
         "symfony/twig-bundle": "^2.7.7"
     },
-    "minimum-stability": "dev",
-    "prefer-stable": true,
+    "config": {
+        "bin-dir": "bin"
+    },
     "autoload": {
         "psr-4": { "Sylius\\Bundle\\ThemeBundle\\": "" }
     },
     "autoload-dev": {
         "psr-4": { "Sylius\\Bundle\\ThemeBundle\\spec\\": "spec/" }
     },
+    "minimum-stability": "dev",
     "extra": {
         "branch-alias": {
             "dev-master": "0.17-dev"
--- a/src/Sylius/Component/Metadata/composer.json
+++ b/src/Sylius/Component/Metadata/composer.json
@@ -20,10 +20,19 @@
         }
     ],
     "require": {
-        "php": "^5.5.9|^7.0"
+        "php": "^5.5.9|^7.0",
+
+        "sylius/resource": "0.17.*",
+        "symfony/property-access": "^2.7",
+        "symfony/options-resolver": "^2.7",
+        "twig/twig": "~1.11"
     },
     "require-dev": {
-        "phpspec/phpspec": "^2.4"
+        "phpspec/phpspec": "^2.4",
+        "phpunit/phpunit": "^4.1"
+    },
+    "config": {
+        "bin-dir": "bin"
     },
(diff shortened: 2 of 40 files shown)


src/Sylius/Bundle/ThemeBundle/composer.json
```diff
@@ -36,14 +36,16 @@
         "phpunit/phpunit": "^4.1",
         "symfony/twig-bundle": "^2.7.7"
     },
-    "minimum-stability": "dev",
-    "prefer-stable": true,
```
pamil: Why?

src/Sylius/Component/Metadata/composer.json
```diff
@@ -20,10 +20,19 @@
...
+
+        "sylius/resource": "0.17.*",
+        "symfony/property-access": "^2.7",
+        "symfony/options-resolver": "^2.7",
+        "twig/twig": "~1.11"
     },
     "require-dev": {
-        "phpspec/phpspec": "^2.4"
+        "phpspec/phpspec": "^2.4",
+        "phpunit/phpunit": "^4.1"
```
pamil: We don't use PhpUnit in Metadata component IIRC.

src/Sylius/Bundle/ThemeBundle/composer.json
pjedrzejewski: We do not have ``prefer-stable`` in other bundles and Symfony doesn't use it - https://github.com/symfony/http-foundation/blob/master/composer.json. If we decide it is good idea to add it then we should do it for all bundles and components.

michalmarcinkowski:
:+1:

src/Sylius/Bundle/ThemeBundle/composer.json
pamil: It's definitely a good idea, without prefer-stable we will install unstable versions of all our dependencies, whereas stable versions suits as well. We should install dev versions only if there isn't any other version available.

src/Sylius/Component/Metadata/composer.json
pjedrzejewski: You used assert classes.

src/Sylius/Bundle/ThemeBundle/composer.json
pamil: Btw. that Symfony package doesn't have `.travis.yml` too, it's meant to be tested as a part of `symfony/symfony`.

src/Sylius/Component/Metadata/composer.json
pamil: :+1:, should be updated to `expect` in the future :)

src/Sylius/Bundle/ThemeBundle/composer.json
pamil: As an example: https://travis-ci.org/Sylius/SyliusResourceBundle/jobs/111649988

```
- Installing symfony/twig-bridge (3.0.x-dev 2a46495)
  Cloning 2a4649506e9e538b6f42b5f1c7545fca99ff6e5a
- Installing symfony/twig-bundle (2.8.x-dev 0901cdf)
  Cloning 0901cdfd83379cedde8cd57a24dd44969c0a38a2
```

Twig bridge `3.0.x-dev` mixed up with twig bundle `2.8.x-dev`, what could possibly go wrong? :)

PR #4371, opened 2016-03-02
[WebBundle] Removed Payment creation in admin panel

ahmadrabie:
Q  | A
------------- | -------------
Bug fix?  | yes
New feature?  | no
BC breaks? | no
Deprecations? | no
Fixed tickets | -
License | MIT
Doc PR | -

Creating payment in backend needs to set `currency` and `order`, so it return exception:
`Integrity constraint violation: 1048 Column 'currency/order' cannot be null`.
IMO you should add this feature in backend order creation.


michalmarcinkowski:
This action doesn't make sense right now, thank you Ahmad for removing it :+1:

PR #4377, opened 2016-03-02
[Promotion][Core] Action decoupling issues

### LINKED ISSUE #4229: [Promotion] ActionFactory should go to Core & ActionInterface::TYPE_FIXED_DISCOUNT too

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | fixes #4229 
| License         | MIT
| Doc PR          |

- moved ``ActionFactory`` to *Core*
- move actions types' constants to theirs classes
- implemented missing ``shipping_discount`` constant

#### Note
There is a quick fix with action form in this PR. ``ActionType`` is implemented in ``PromotionBundle`` but has been using ``fixed_discount`` constant as default type (similar issue concerns ``RuleType``). Because of moving this to *Core*, there is now quick js solution to handle proper default action type configuration form rendering. I think it's acceptable due to upcoming changes in admin panel ui. However, we should deeply rethink actions/rules configurations forms logic.


lchrusciel:
:+1:

src/Sylius/Bundle/PromotionBundle/Behat/PromotionContext.php
```diff
@@ -152,6 +152,18 @@ public function iShouldNotSeeProductInCartSummary($productName)
     }
 
     /**
+     * @Given /^I add "([^"]*)" action$/
+     */
+    public function iAddAction($action)
+    {
+        $this->clickLink('Add action');
+
+        $this->getSession()->wait(100);
```
pamil: Can we wait for an element to appear instead of fixed time?

src/Sylius/Bundle/PromotionBundle/Form/Type/ActionType.php
```diff
@@ -38,7 +36,11 @@ public function buildForm(FormBuilderInterface $builder, array $options = [])
                 ],
             ])
             ->addEventSubscriber(
-                new BuildActionFormSubscriber($this->registry, $builder->getFormFactory(), $options['configuration_type'])
+                new BuildActionFormSubscriber(
+                    $this->registry,
+                    $builder->getFormFactory(),
+                    (isset($options['configuration_type'])) ? $options['configuration_type'] : null
```
pamil: `(isset(...))` -> `isset(...)`?

pjedrzejewski:
Thanks Mateusz!

src/Sylius/Bundle/WebBundle/Resources/views/Backend/Promotion/create.html.twig
```diff
@@ -2,6 +2,15 @@
 
 {% from 'SyliusResourceBundle:Macros:actions.html.twig' import create %}
 
+{% block javascripts %}
+    {{ parent() }}
+    {% javascripts output='assets/compiled/backend_promotions.js'
+    'bundles/syliuspromotion/js/sylius-promotion.js'
+    %}
+    <script type="text/javascript" src="{{ asset_url }}"></script>
```
pamil: Can we just link it by `asset()` instead of using Assetic?

Zales0123: It's already a quick fix, we'll change whole CRUD pages in weeks, so I think there is no need to change it ;)

rvanlaak:
This merge results in errors if you upgrade from `0.17` to `0.18`. We do not make use of the `Core` but only use `PromotionBundle` in our Symfony application. The configuration now is not able to parse because of this changed use statement: https://github.com/Sylius/Sylius/pull/4377/files#diff-e937bf0d4c6f47fba4fb8be6bc8e70d1R22

@Zales0123 can you please provide upgrade instructions in `UPGRADE.md` ? @pjedrzejewski  basically this code change means we can not use `PromotionBundle` without `Core` anymore.

Zales0123:
@rvanlaak This is definitely a bug, *PromotionBundle* should not depend on *Core* component. I'll fix this as soon as possible, thank you for reporting that!
In fact we have tests running for separate packages, but these are only *phpspec* and does not help if we don't have spec for class (which is totally correct, as specing ``Configuration`` class would be ridiculous).

rvanlaak:
Thank, great! You think this will result in a hotfix release? Let me know if you fixed this so I can try to upgrade again :+1:

Issue #4381, opened 2016-03-03
[Product Review] Using interface instead of class implementation


peteward:
I'm fairly sure that these should be interfaces not class implementations?:

https://github.com/Sylius/Sylius/blob/3c02eef2a12a76062317af94b971bb0b183d3476/src/Sylius/Bundle/CoreBundle/Form/Type/ProductReviewAdminType.php#L35

I get an error because it's trying to use a `sylius_customer` database table instead of a `reiss_customer` one, I presume the same is true of Product.

Although, I can't be sure that changing it definitely works since trying to load 1,000,000 customers and 100,000 products into select boxes kinda blows up my webserver :)

michalmarcinkowski:
Of course these should be interfaces, thanks for reporting :+1: 

Regarding select boxes, the autocomplete is coming soon! :)

nakashu:
These form types have probably the same problem i guess?

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Form/Type/OrderItemType.php#L55
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Form/Type/Api/OrderItemType.php#L32
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Form/Type/Api/ProductType.php#L30

michalmarcinkowski:
Fixed already.

PR #4383, opened 2016-03-03
[ProductReview] Changed classes to interfaces in ProductReviewAdminType

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #4381 
| License         | MIT
| Doc PR          |

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Form/Type/ProductReviewAdminType.php
+++ b/src/Sylius/Bundle/CoreBundle/Form/Type/ProductReviewAdminType.php
@@ -11,8 +11,8 @@
 
 namespace Sylius\Bundle\CoreBundle\Form\Type;
 
-use Sylius\Component\Core\Model\Customer;
-use Sylius\Component\Core\Model\Product;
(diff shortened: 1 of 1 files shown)


peteward:
:+1:  :smile:

GSadee:
I guess that this PR was not a good idea, we could not use interfaces instead of classes, because it have to be an entity mapped by doctrine. @michalmarcinkowski WDYT? Is there any other option to solve problem from #4381 ?

pjedrzejewski:
@GSadee Yes, removing this feature. I don't think we want adding reviews in backend. For editing: We can generate ``sylius_customer_choice`` type. Just add ``choice`` configuration node under ``customer -> node``. And I think we should remove the ability to select products, it does not make sense. Still not sure if it makes sense to edit them in admin.

nakashu:
Edit in admin for review is viable but only for certain stuff, like title and comment. So the admin can fix typos in the review post by customer. Rating and review subject should not be editable imho.

PR #4386, opened 2016-03-03
[AddressingBundle] Added sorting countries by name in form type

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT
| Doc PR          |

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Form/Type/CountryCodeChoiceType.php
+++ b/src/Sylius/Bundle/AddressingBundle/Form/Type/CountryCodeChoiceType.php
@@ -64,6 +64,8 @@ class CountryCodeChoiceType extends CountryChoiceType
             $countryCodes[$country->getCode()] = Intl::getRegionBundle()->getCountryName($country->getCode());
         }
 
+        asort($countryCodes);
+
         return $countryCodes;
     }
 }


src/Sylius/Bundle/AddressingBundle/Form/Type/CountryCodeChoiceType.php
```diff
@@ -64,6 +64,6 @@ private function getCountryCodes(array $countries)
             $countryCodes[$country->getCode()] = Intl::getRegionBundle()->getCountryName($country->getCode());
         }
 
-        return $countryCodes;
+        return asort($countryCodes);
```
peteward: Maybe better to sort when you retrieve from doctrine on lines 33 and 35?

GSadee: Unfortunately, I receive from doctrine only country codes, but I want to sort by country names. However I'm not sure that it is the best solution.

peteward: ah ok good point :+1:

michalmarcinkowski:
Thanks Grzegorz! :+1:

PR #4406, opened 2016-03-07
[Taxation] Tax Calculation Strategies

### LINKED ISSUE #4454: [RFC] Double-rounding on tax
Hi,

I know me and @michalmarcinkowski have had some fun conversations about tax rounding in the past :)  but currently I think there's a flaw in the implementation as integer rounding is done twice.

The process of calculation and distribution are split into 2 separate services which are called here:

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Taxation/OrderItemsTaxesByZoneApplicator.php#L80

By default the [`IntegerDistributor`](https://github.com/Sylius/Sylius/blob/777ed6f4d2d855ee8b2c49d7be8427e4f193352e/src/Sylius/Component/Core/Distributor/IntegerDistributor.php#L33) will take the tax amount and split it across units to ensure the tax for individual units adds up as closely as possible to the original tax amount. However this amount has already been rounded by the [`DefaultCalculator`](https://github.com/Sylius/Sylius/blob/322157b1f02dd1b101c2b13588b876de79b7dbdb/src/Sylius/Component/Taxation/Calculator/DefaultCalculator.php#L31) so it's not actually that accurate.

In our case we are writing a different distributor, because we cannot charge customers different tax amounts for the same item because they may have bought 2 instead of 1 - it's a business rule.

Scenario from @mmckelviereiss on my team:

* An item with an amount with quantity of 1, amount  395 and tax of 20% = 65.83333333333333 tax
* Which gets rounded by the calculator to 65.83 and then applied to 1 unit at 65.83.
* If we quantity of 2 then it's an amount of 790 and tax of 20% = 131.6666666666667
* Which gets rounded by the calculator to 131.67.
* The distributor will then use this rounded 131.67 / 2 = 65.835
* Our business rounding gives us 65.84

If the unrounded value was returned from the calculator:
* 131.6666666666667 / 2 = 65.8333333335
* Our distributor will then round to a consistent 65.83.

So the double-rounding is a problem.

Ideally we don't want to override the entire `OrderItemsTaxesByZoneApplicator`, really we should only have to override the `IntegerDistributor` to override from the unrounded (non-integer) value.

Ideas:
* Inject the distribution into the calculator and pass in the number to split the calculation into (defaulting to 1).
* Add (yet another) service layer here to do it - seems excessive.

Thoughts?

peteward:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | fixes #4454
| License         | MIT

NOTE: As we are looking for your input on this approach first, this work so far does not include all updated and new PHPSpec or any Behat scenario updates.

## Overview

This PR introduces the notion of `TaxCalculationStrategy`.

Currently Sylius has a single strategy which calculates Tax through Applicators at the OrderItem-level. Mark has implemented an architecture for the following use cases:

* Sylius OrderItem-level tax application
* Sylius OrderItemUnit-level tax application
* A Custom application strategy (using an external provider for tax calculation)
* A combination of custom application and one of the Sylius built-in application strategies (this is the REISS use-case).

This PR satisfies all of these use-cases through the use of *Prioritized Strategies*.  By prioritizing we can allow a custom strategy to assess the order first to decide whether it should be applied, then 'fallback' to built-in Sylius strategies.

In our case we will look to see if the order is in the US, if so use our third-party engine, otherwise we will let one of the built-in strategies calculate tax for our remaining orders (which have simpler taxation rules). 

All of the above should be possible through configuration without overriding any classes. 

## Implementation

There is now a global taxation setting for the default tax calculation strategy which will be order items based by default.

The `OrderTaxesApplicator` has now been renamed to `OrderTaxesProcessor`, which accepts a prioritised set of tax calculation strategies. There are now two main tax calculation strategies, one that it order items based and the other is order item units based. A service for the `OrderItemUnitsTaxesApplicator` has now been added too. All tax applicators inherit a single unified `OrderTaxesApplicatorInterface`.

I hope you like what you see here and would appreciate your feedback on this so far as I'm now working on the PHPSpec.


peteward:
any thoughts on this one @pjedrzejewski. It shouldn't be hard to make this a bit more flexible and allow people to more easily do their own thing without lots of overrides.

mmckelviereiss:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | mentioned in #4454
| License         | MIT
| Doc PR          | 

@michalmarcinkowski @pjedrzejewski Based on the conversation in RFC #4454 I have worked out a solution to use different tax calculation strategies with @peteward.

NOTE: As we are looking for your input on this approach first, this work so far does not include all updated and new PHPSpec or any Behat scenario updates.

## Overview

We now have a global taxation setting for the default tax calculation strategy which will be order items based by default.

The `OrderTaxesApplicator` has now been renamed to `OrderTaxesProcessor`, which accepts a prioritised set of tax calculation strategies. By default there are now two main tax calculation strategies, one that it order items based and the other is order item units based. A service for the `OrderItemUnitsTaxesApplicator` has now been added too. All tax applicators now inherit a single unified `OrderTaxesApplicatorInterface` now.

The good thing about having a prioritised tax calculation strategy is that client projects can easily add a new custom strategy and support for it can be checked against the order and/or zone.

I hope you like what you see here and would appreciate your feedback on this so far as I'm now working on the PHPSpec.

mmckelviereiss:
I have now made the taxation Settings an optional dependency for the `TaxCalculationStrategy`.

peteward:
@mmckelviereiss I've updated the PR descriptionand title to include your updates.

@pjedrzejewski, @michalmarcinkowski, @Zales0123 - ready for review

peteward:
ping @michalmarcinkowski @pjedrzejewski

src/Sylius/Bundle/CoreBundle/Form/Type/Taxation/TaxCalculationStrategyChoiceType.php
```diff
@@ -0,0 +1,62 @@
...
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\CoreBundle\Form\Type\Taxation;
```
pjedrzejewski: Can we put it in the same namespace as other form types? I agree we need to separation in the core, but I'd like to do it in separate PR if we agree that's a good idea.

src/Sylius/Bundle/CoreBundle/EventListener/OrderTaxationListener.php
```diff
@@ -50,6 +50,6 @@ public function applyTaxes(GenericEvent $event)
             );
         }
 
-        $this->orderTaxesApplicator->apply($order);
+        $this->orderTaxesProcessor->applyTaxes($order);
```
pjedrzejewski: processTaxes? Having 2 interfaces related to taxes that have apply method can be confusing.

src/Sylius/Bundle/CoreBundle/Settings/TaxationSettingsSchema.php
```diff
@@ -64,6 +62,10 @@ public function buildForm(FormBuilderInterface $builder)
                 'required' => false,
                 'label' => 'sylius.form.settings.taxation.default_tax_zone',
             ])
+            ->add('default_tax_calculation_strategy', 'sylius_tax_calculation_strategy_choice', [
```
pjedrzejewski: It is fine if we put it on taxation settings for now, just a note that I was considering moving default tax zone to Channel, we could do the same with this one, this would be even more flexible - having different strategies for different channels.

pjedrzejewski:
@peteward @mmckelviereiss I now fully understand the idea, sorry it took so much time. I think good to go after applying my comments. With namespaces, I just want to give it a bit more thought, so please revert, but we will probably do it this way for all components. Thanks for your work!

peteward:
When you say revert, you mean change `OrderTaxesProcessor` back to `OrderTaxesApplicator`? 

The reason that we made this change was to avoid having OrderTaxesApplicator calling a TaxStrategy calling Applicators (Applicator -> Strategy -> Applicators seemed wrong).  Although processor can be a bit generic, it avoids this and the Process does removing as well as applying.

pjedrzejewski:
@peteward No no, I only meant revert moving stuff to ``Taxation`` namespace in Core, we will do that for all components at once I think. :)

mmckelviereiss:
ping @michalmarcinkowski @pjedrzejewski @peteward 

The changes based on the feedback above have now been made. To summarise the last commit contains the following changes to pass tests:

* Undo the namespace change for the tax form types.
* Rename the OrderTaxesProcessorInterface::applyTaxes method to 'process'.
* Complete the PHPSpec of the tax calculation strategies.
* Correct the namespace of the UnsupportedTaxCalculationStrategyException.
* Update the TaxCalculationStrategy to work with the new settings rework.
* Abstract out the TaxCalculationStrategy into the Component/Core and CoreBundle since the primary implementation depended upon Settings and can extend an AbstractTaxCalculationStrategy.

src/Sylius/Bundle/CoreBundle/Settings/TaxationSettingsSchema.php
mmckelviereiss: Having a default tax calculation strategy on a per channel basis would be useful and offers a lot more power and flexibility. :+1:

michalmarcinkowski:
Nice work Mark! :+1: 

We are out of office right now and will review this PR in the next week.

pjedrzejewski:
Looks good, thank you very much Mark and Peter! I will open an issue about moving this configuration to channels. :+1:

src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
```diff
@@ -237,22 +240,63 @@
             <argument type="service" id="sylius.repository.payment_method" />
         </service>
 
-        <service id="sylius.taxation.order_shipment_taxes_by_zone_applicator" class="%sylius.taxation.order_shipment_taxes_by_zone_applicator.class%">
+        <service id="sylius.taxation.order_shipment_taxes_applicator" class="%sylius.taxation.order_shipment_taxes_applicator.class%">
```
michalmarcinkowski: Why removing **by_zone**?

peteward: when would taxes not be by zone?

michalmarcinkowski: E.g. you can use some external service for tax calculation that can identify the rates using country/province (that can be taken from order entity) and know nothing about the Sylius' zones. I would at least leave `byZone` in the interfaces (e.g. [here](https://github.com/Sylius/Sylius/pull/4406/files#diff-cb3e1d18d47a2ce77e48662ae8172afdL20)), wdyt?

peteward: Our change of naming here evolved from trying to work through the pattern, I don't think it was an immediate decision, but an attempt at trying to simplify the naming when understanding all of the components involved.

As it stands, I would say KISS: If someone wants an alternative mechanism that falls outside of Sylius's now wide  scope of functionality then let them implement and name it accordingly. Even for the simplest of situations most people would just create a single tax zone and use this rather than implement their own version.

If you disagree then of course I'm not going to lose any sleep if you decide to change it back ;)

michalmarcinkowski: I'm for keeping it simple and since we don't plan to have any other types of applicators in core I think it's fine to leave it as it is now :)

psyray:
Hi all,

Thanks for this :+1: 

Silly question ....
How could I register a custom application strategy (using an external provider for tax calculation) ?

Thanks for reply ;)

peteward:
Hi @psyray, 

taken straight from our project:

```
    reiss.tax_calculation.onesource_taxes_applicator:
        class: Reiss\Bundle\StoreBundle\Taxation\Applicator\OneSourceTaxesApplicator
        arguments:
            - "@reiss.tax_calculation.onesource_calculator"
            - "@doctrine.orm.entity_manager"
            - "@sylius.factory.adjustment"

    reiss.tax_calculation.onesource_strategy:
        class: %reiss.tax_calculation.onesource_strategy.class%
        arguments:
            type: "onesource"
            applicators:
                - "@reiss.tax_calculation.onesource_taxes_applicator"
            entityManager: "@doctrine.orm.entity_manager"
            logger: "@logger"
        tags:
            - { name: "sylius.taxation.calculation_strategy", type: "onesource", label: "OneSource", priority: 99 }
```

It's obviously up to you what your strategy needs or what applicators you want it to use. You may be happy with the standard sylius applicators - we needed our own here for US tax calculation.

Inside the strategy you need to implement the `supports()` method which determines whether this strategy should be used or not.

In our case we check whether a TaxRate is configured to use our OneSource calculator:

```
    /**
     * {@inheritdoc}
     */
    public function supports(OrderInterface $order, ZoneInterface $zone)
    {
        $rate = $zone->getTaxRates()->first();

        if ($rate instanceof TaxRateInterface) {
            return $rate->getCalculator() === $this->type;
        }

        return false;
    }
```

We have a complex scenario of custom external calculation for the US, everything else goes through Sylius build-in strategies and calculators.

Your use-case might be simpler.

I hope that helps.

psyray:
fast & furious :+1:

pjedrzejewski:
@peteward Thanks for clarifying that, awesome!

mmckelviereiss:
Hi @psyray,

Just to point out. The key to registering your custom strategy is to make sure your custom strategy service implements `TaxCalculationStrategyInterface` or extends `AbstractTaxCalculationStrategy`.

Your new strategy must be tagged with the name `sylius.taxation.calculation_strategy` and have a `type` and `label` attribute in order for the `RegisterTaxCalculationStrategiesPass` to find and register it.

I hope that helps.

peteward:
The other subtlety that I forgot to mention is the `priority` on the tag can be important if you do want to use multiple strategies. The strategy registry will try to apply strategies in priority order, with the first one that `supports()` being applied.

This gives you a lot of flexibility for different configurations.

In our case, the orders that we wanted to use the US calculator would also be supported by default Sylius strategy, so it's important that our custom strategy has a higher priority and is checked first.

psyray:
Great, thanks @peteward @mmckelviereiss 
I think you should put this in the upgrade.md or in the docs.

psyray:
Implemented in my project ;)
Your new taxation system is really awesome, simple and we don't need anymore to override any class, thanks for that :+1: 

For others who just want to add a simple taxation system over the standard one to calculate tax on added product/service (in my case a clothing marking system added on the product page with his own admin section - price, marking, position...) here is the complete way to do

**service file in xml version**
```
<parameters>
        <parameter key="your_bundle.taxation.your_system_taxes_applicator.class">YourBundle\Bundle\CoreBundle\Taxation\Applicator\YourSystemTaxesApplicator</parameter>
        <parameter key="your_bundle.taxation.your_system_strategy.class">YourBundle\Bundle\CoreBundle\Taxation\Strategy\YourSystemTaxCalculationStrategy</parameter>
</parameters>

<services>
        <!-- taxation -->
        <service id="your_bundle.taxation.your_system_taxes_applicator" class="%your_bundle.taxation.your_system_taxes_applicator.class%">
            <argument type="service" id="sylius.tax_calculator" />
            <argument type="service" id="sylius.factory.adjustment" />
            <argument type="service" id="sylius.tax_rate_resolver" />
        </service>

        <service id="your_bundle.taxation.your_system_strategy" class="%your_bundle.taxation.your_system_strategy.class%">
            <argument key="type">your_system</argument>
            <argument type="collection">
                <argument type="service" id="your_bundle.taxation.your_system_taxes_applicator" />
                <argument type="service" id="sylius.taxation.order_items_taxes_applicator" />
                <argument type="service" id="sylius.taxation.order_shipment_taxes_applicator" />                
            </argument>
            <tag name="sylius.taxation.calculation_strategy" type="your_system" label="Your system based" priority="99" />
        </service>
<services>
```

**strategy class** file is simple, **supports** method only returns true because it's the same as the general order calculation strategy
```
<?php

namespace YourBundle\Bundle\CoreBundle\Taxation\Strategy;

use Sylius\Component\Core\Taxation\Strategy\AbstractTaxCalculationStrategy;
use Sylius\Component\Addressing\Model\ZoneInterface;

class YourSystemLotTaxCalculationStrategy extends AbstractTaxCalculationStrategy
{
    /**
     * {@inheritdoc}
     */
    public function supports(OrderInterface $order, ZoneInterface $zone)
    {
        return true;
    }    
}
```

**applicator class** is as follow (inspired from the orderItemsUnitsTaxesAppplicator
```
<?php

namespace YourBundle\Bundle\CoreBundle\Taxation\Applicator;

use Sylius\Component\Addressing\Model\ZoneInterface;
use Sylius\Component\Core\Model\AdjustmentInterface;
use Sylius\Component\Order\Factory\AdjustmentFactoryInterface;
use Sylius\Component\Taxation\Calculator\CalculatorInterface;
use Sylius\Component\Taxation\Resolver\TaxRateResolverInterface;
use YourBundle\Bundle\YourSystemBundle\Entity\YourSystem;
use Sylius\Component\Core\Taxation\Applicator\OrderTaxesApplicatorInterface;

class YourSystemTaxesApplicator implements OrderTaxesApplicatorInterface
{
    /**
     * @var CalculatorInterface
     */
    private $calculator;

    /**
     * @var AdjustmentFactoryInterface
     */
    private $adjustmentFactory;

    /**
     * @var TaxRateResolverInterface
     */
    private $taxRateResolver;

    /**
     * @param CalculatorInterface $calculator
     * @param AdjustmentFactoryInterface $adjustmentFactory
     * @param TaxRateResolverInterface $taxRateResolver
     */
    public function __construct(CalculatorInterface $calculator, AdjustmentFactoryInterface $adjustmentFactory, TaxRateResolverInterface $taxRateResolver)
    {
        $this->calculator = $calculator;
        $this->adjustmentFactory = $adjustmentFactory;
        $this->taxRateResolver = $taxRateResolver;
    }

    /**
     * {@inheritdoc}
     */
    {
        $yourDatas = $order->getYourDatas();
        foreach ($yourDatas as $yourData) {
            $orderItem = $yourData->getOrderItems()->first();
            
            $taxRate = $this->taxRateResolver->resolve($orderItem->getVariant(), ['zone' => $zone]);
    
            if (null === $taxRate) {
                return;
            }
    
            $totalTaxAmount = $this->calculator->calculate($yourData->getTotal(), $taxRate);
    
            $this->addAdjustment($yourData->getOrder(), $totalTaxAmount, $taxRate->getLabel(), $taxRate->isIncludedInPrice());
        }
    }

    /**
     * @param yourEntity $yourData
     * @param int $taxAmount
     * @param string $label
     * @param bool $included
     */
    private function addAdjustment(OrderInterface $order, $taxAmount, $label, $included)
    {
        $yourDataTaxAdjustment = $this->adjustmentFactory->createWithData(AdjustmentInterface::TAX_ADJUSTMENT, $label, $taxAmount, $included);
        $order->addAdjustment($yourDataTaxAdjustment);
    }

}

```
@peteward @mmckelviereiss 
If you have any remark, you're welcome ;)

If it can help someone, enjoy ;)

peteward:
looks great, well done. 

We should definitely add this to some docs about Tax at some point. I'd be happy to contribute on further cases too.

psyray:
You could use my example without any problems ;)

Issue #4408, opened 2016-03-07
[RFC] Prefer stable in bundles' & components' composer.json


pamil:
Right now, we download unstable dependencies and this may end with dev dependencies everywhere, failing tests and some unusual packages installed together:

```
- Installing symfony/twig-bridge (3.0.x-dev 75a8e36)
    Cloning 75a8e368c3bba891f9f8ce9dfc92a47a2bef869d
- Installing symfony/twig-bundle (2.8.x-dev 166c91b)
    Cloning 166c91ba49b2e7272052fc15327eab3a1947e3c1
```

The solution is to add `"prefer-stable": true` so we won't install any dev dependencies unless it is really needed. It should also install faster.

pjedrzejewski:
:+1:

pamil:
Already applied

PR #4415, opened 2016-03-08
[Travis] Continuous integration for each Sylius package included in the build

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #4408
| License         | MIT
| Doc PR          | -

- [x] Set `prefer-stable` to `true` in all packages' `composer.json`
- [x] Updated Sylius packages versioning from different formats (mostly `0.18.*`, but there was `0.18.*@dev` and `^0.18.0` too) to simple`^0.18`)
- [x] Removed vertical alignment in `composer.json`
- [x] Fixed not installable packages
- [x] Added dependencies downloading for each package
- [x] Added Composer validation for each package
- [x] Added isolated tests running for each package (phpspec & phpunit support)
- [x] Added caching for packages' `composer.lock` (only the first builds will take much more time :tada:)


bendavies:
You could run all your shell scripts through http://www.shellcheck.net/

etc/concurrent.lib.sh
```diff
@@ -0,0 +1,1067 @@
...
+        echo '>>> DRY RUN (concurrent): The "$CONCURRENT_DRY_RUN" environment variable is set. <<<'
+    fi
+
+    __crt__hide_cursor
+    __crt__start_all_tasks
+    __crt__wait_for_all_tasks
+    __crt__show_cursor
+
+    exit ${__crt__final_status}
+)
```
pjedrzejewski: Any chance we can avoid committing that? Downloading?

pamil: Yeah, I've just get rid of that in favour of GNU parallel :tada:

src/Sylius/Bundle/AdminBundle/composer.json
```diff
@@ -7,23 +7,23 @@
...
             "homepage": "http://github.com/Sylius/Sylius/contributors"
         }
     ],
     "require": {
         "php": "^5.5.9|^7.0",
 
-        "sylius/ui-bundle": "^0.17.0",
-        "sylius/core-bundle": "^0.17.0",
+        "sylius/ui-bundle": "^0.18",
+        "sylius/core-bundle": "^0.18",
```
liverbool: should be `0.18.*`?

pjedrzejewski:
Thanks Kamil! Awesome work! :+1: This will help a lot...

Issue #4410, opened 2016-03-07
[Behat] Not all scenarios are included in travis build


lchrusciel:
_edited_
During work on #3854 I had plenty places where my code could not work, and because of that I decided to run behats feature after feature on my local machine. It seems that some of suits are not included to travis build. For example tags `@legacy_product` or `@legacy_orders`
![zrzut ekranu 2016-03-07 o 17 37 04](https://cloud.githubusercontent.com/assets/6213903/13576042/fba9ec7a-e48b-11e5-841c-5023861f86a3.png)

pamil:
There's no `legacy_products` suite. The `@legacy_products` is a tag. Running Behat with specific file will test the feature using every suitable suite, in this case - `@legacy_products` is still parsed by `products` suite. This may lead to confusion :)

IMHO the suites with tags `@legacy_products` should be tagged as `@legacy @product`, the suite name should be renamed from `products` to `legacy_product` and it should filter the scenarios by `@legacy && @product`.

This way we can run all product suites by `behat --tags @product` - both the legacy ones and the future ones in each possible environment (`ui`, `domain`, etc.)

lchrusciel:
@pamil, ok I didn't explain problem correctly. And sure, what have you written make sense. But still It doesn't resolve problem of broken tests. I have found at least 3 places where the code is broken, but tests are still green.

pamil:
Where are these broken tests? The one scenario from the screenshot is executed on Travis and it is passing.

pjedrzejewski:
I don't think we should spend too much time working on the legacy, let's leave it as is. ;) We will remove them soon. :D

michalmarcinkowski:
@lchrusciel I verified that all scenarios are executed on travis. I also checked it on my local env and the scenariou you mentioned wasn't passing on `master` branch for me also, but after cache clear it is green now :)

lchrusciel:
Commands I used to reproduce this issue:
```
git pull upstream master
rm -fr app/cache/*
app/console do:da:dr --env test --force
app/console do:da:cr --env test
app/console do:sc:cr --env test
bin/behat features/legacy/backend/orders.feature --stop-on-failure 
```
Result?
![bugzol](https://cloud.githubusercontent.com/assets/6213903/13602779/fffd1d80-e538-11e5-8e28-5170bed0e310.png)

It has to be broken, because I forgot to remove deleted property from orders twig template... I want to fix it right now, but still I guess there is an issue.

[...]

PR #4438, opened 2016-03-09
[Behat] Set "twig.strict_variables" to true in test_cached

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #4410
| License         | MIT
| Doc PR          | -

As it defaults to `%kernel.debug%`, we don't get any error if referencing unexisting variable / attribute in Twig.

### DIFF
--- a/app/config/config_test_cached.yml
+++ b/app/config/config_test_cached.yml
@@ -19,3 +19,6 @@ doctrine_phpcr:
             type: memcached
             host: localhost
             port: 11211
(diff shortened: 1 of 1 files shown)


lchrusciel:
:+1:

pjedrzejewski:
Good catch!

PR #4453, opened 2016-03-10
[Behat][Promotion] Fixed discount on products in price range scenarios

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT
| Doc PR          |

### DIFF
--- a/features/promotion/receiving_fixed_discount_on_items_on_price_range.feature
+++ b/features/promotion/receiving_fixed_discount_on_items_on_price_range.feature
@@ -1,0 +1,97 @@
+@promotion
+Feature: Receiving fixed discount on products from specific price range
+    In order to pay less while buying goods from specific price range
+    As a Customer
+    I want to receive discount for my purchase
+
+    Background:
+        Given the store operates on a single channel in "France"
+        And the store has a product "PHP T-Shirt" priced at "€100.00"
+        And the store has a product "PHP Mug" priced at "€20.00"
+        And there is a promotion "Christmas promotion"
+
+    @todo
+    Scenario: Receiving fixed discount on a single item fulfilling minimum price criteria
+        Given the promotion gives "€10.00" off on every product with minimum price at "€50.00"
+        When I add product "PHP T-Shirt" to the cart
+        Then its price should be decreased by "€10.00"
+        And my cart total should be "€90.00"
+        And my discount should be "-€10.00"
+
+    @todo
+    Scenario: Receiving fixed discount on a single item with price equal to filter minimum criteria
+        Given the promotion gives "€10.00" off on every product with minimum price at "€100.00"
+        When I add product "PHP T-Shirt" to the cart
+        Then its price should be decreased by "€10.00"
+        And my cart total should be "€90.00"
+        And my discount should be "-€10.00"
+
+    @todo
+    Scenario: Receiving fixed discount on a single item fulfilling range price criteria
+        Given the promotion gives "€10.00" off on every product priced between "€15.00" and "€50.00"
+        When I add product "PHP Mug" to the cart
+        Then its price should be decreased by "€10.00"
+        And my cart total should be "€10.00"
+        And my discount should be "-€10.00"
+
+    @todo
+    Scenario: Receiving fixed discount on a single item with price equal to filter range minimum criteria
+        Given the promotion gives "€10.00" off on every product priced between "€20.00" and "€50.00"
+        When I add product "PHP Mug" to the cart
+        Then its price should be decreased by "€10.00"
+        And my cart total should be "€10.00"
+        And my discount should be "-€10.00"
+
+    @todo
+    Scenario: Receiving fixed discount on a single item with price equal to filter range maximum criteria
+        Given the promotion gives "€10.00" off on every product priced between "€15.00" and "€20.00"
+        When I add product "PHP Mug" to the cart
+        Then its price should be decreased by "€10.00"
+        And my cart total should be "€10.00"
+        And my discount should be "-€10.00"
+
+    @todo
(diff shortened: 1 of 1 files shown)


features/promotion/receiving_fixed_discount_on_items_on_price_range.feature
```diff
@@ -0,0 +1,81 @@
...
+
+    Background:
+        Given the store operates on a single channel in "France"
+        And the store has a product "PHP T-Shirt" priced at "€100.00"
+        And the store has a product "PHP Mug" priced at "€20.00"
+        And there is a promotion "Christmas promotion"
+
+    @todo
+    Scenario: Receiving fixed discount on a item fulfilling minimum price criteria
+        Given this promotion gives "€10.00" on every product priced at more than "€50.00"
```
lchrusciel: `priced more` / `more expensive`?

Zales0123: ``more expensive`` is the most natural for me.

lchrusciel: :+1:

features/promotion/receiving_fixed_discount_on_items_on_price_range.feature
```diff
@@ -0,0 +1,81 @@
...
+    @todo
+    Scenario: Receiving fixed discount on a single item fulfilling range price criteria
+        Given this promotion gives "€10.00" on every product priced at more than "€15.00" and less than "€50.00"
+        When I add product "PHP Mug" to the cart
+        Then its price should be decreased by "€10.00"
+        And my cart total should be "€40.00"
+        And my discount should be "-€10.00"
+
+    @todo
+    Scenario: Receiving fixed discount on a multiple items fulfilling maximum price criteria
```
pjedrzejewski: remove ``a`` before ``multiple``.

features/promotion/receiving_fixed_discount_on_items_on_price_range.feature
```diff
@@ -0,0 +1,81 @@
...
+    @todo
+    Scenario: Receiving fixed discount equal to the items total of my cart even if the discount is bigger than the items total
+        Given this promotion gives "€30.00" on every product priced at more than "€30.00" and less than "€50.00"
+        When I add 2 products "PHP Mug" to the cart
+        Then theirs price should be decreased by "€40.00"
+        Then my cart total should be "€0.00"
+        And my discount should be "-€40.00"
+
+    @todo
+    Scenario: Receiving fixed discount only on items that fits price range criteria
```
pjedrzejewski: fit

features/promotion/receiving_fixed_discount_on_items_on_price_range.feature
```diff
@@ -0,0 +1,81 @@
...
+
+    Background:
+        Given the store operates on a single channel in "France"
+        And the store has a product "PHP T-Shirt" priced at "€100.00"
+        And the store has a product "PHP Mug" priced at "€20.00"
+        And there is a promotion "Christmas promotion"
+
+    @todo
+    Scenario: Receiving fixed discount on a item fulfilling minimum price criteria
+        Given this promotion gives "€10.00" off on every product more expensive than "€50.00"
```
michalmarcinkowski: Given **the** promotion gives... `this` should be used when the preceding line `talk` about promotion.

michalmarcinkowski:
I'm wondering if we should have scenarios for promotions with rule `for every product less expensive than "€X"`, does it makes sense at all?

IMO we should only have scenarios for:
 * more expensive than €X
 * more expensive than €X but less expensive than €Y (where Y > X)

WDYT?

michalmarcinkowski:
Then we should consider the following cases:
* more expensive than €X:
 * T-shirt equal to €X
 * T-shirt more expensive than €X
* more expensive than €X and less expensive than €Y
 * T-shirt equal to €X
 * T-shirt equal to €Y
 * T-shirt priced at €Z ( X < Z < Y)

Zales0123:
@michalmarcinkowski I was thinking about it, but therefore I think we should forbid to set only ``max`` price? I also don't think that having "product less expensive than (...)" promotion is reasonable, but I'm not an expert :smile: @pjedrzejewski @peteward what do you think about not implementing this feature?

michalmarcinkowski:
I think it could have sense in scenarios for [percentage discount on item](https://github.com/Sylius/Sylius/pull/4463), but with fixed discount it can lead to the case when you sell something for free unless the discount amount is lower than the cheapest product in your store.

peteward:
The only use-case I can think of for having a maximum value limit would be staggered discounts (although it's not something we would do). e.g.

You have multiple promotions such as:
* €1 off everything < €10
* €2 off everything between €10 and €20.
* etc.

You could achieve the same thing by just rewarding the promotion based on the greatest discount, but I'm not sure if you considered this way of applying promotions?  I know before that only the first qualifying exclusive promotion would be applied arbitrarily, it wouldn't consider which actually offers the best discount for the customer (which it should really, but this opens up a new can of fun...)

michalmarcinkowski:
Thanks for your input Peter. That is a good example, but I think we're not going to support that kind of promotions right now. This could be handled by a custom action if someone needs it, but making it generic based on current architecture will introduce a lot of logic and edge cases we would need to handle. IMO we can skip the `less than` scenarios.

peteward:
:+1:

Zales0123:
Ok, for me it's good to remove ``less than`` scenarios for now, as *primo* it can be delivered later and *secundo* it's really specific case. So also :+1:

pjedrzejewski:
Closing as it will be merged as part of #4457.

PR #4455, opened 2016-03-10
[RFC][Taxation] Skip 0 value tax adjustment on shipping.

### LINKED ISSUE #4452: [RFC] Should we skip 0 value tax adjustments on shipping?
Skipping 0 value tax amounts is done for OrderItemUnits, but it is not done for Shipping.

Skipping 0 value in general is an interesting thought, but it reduces Adjustment bloat and in any case you want to display it explicitly you can fallback to 0 in a template so I think it's OK :+1: 

However it should probably be consistent?

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Taxation/OrderShipmentTaxesByZoneApplicator.php#L76

If agreed we'll PR this since it's trivial to check for 0 here.

Thoughts?

mmckelviereiss:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #4452
| License         | MIT
| Doc PR          | 

The shipment tax adjustment will not be added if the resulting tax amount calculated for it is 0.

### DIFF
--- a/src/Sylius/Component/Core/spec/Taxation/OrderShipmentTaxesByZoneApplicatorSpec.php
+++ b/src/Sylius/Component/Core/spec/Taxation/OrderShipmentTaxesByZoneApplicatorSpec.php
@@ -105,7 +105,7 @@ class OrderShipmentTaxesByZoneApplicatorSpec extends ObjectBehavior
 
         $calculator->calculate(1000, $taxRate)->willReturn(0);
 
-        $adjustmentsFactory->createWithData(AdjustmentInterface::TAX_ADJUSTMENT, 'Simple tax (0%)', 0, false)->shouldNotBeCalled();
+        $adjustmentsFactory->createWithData(Argument::any())->shouldNotBeCalled();
 
         $this->apply($order, $zone);
     }
--- a/src/Sylius/Component/Core/spec/Taxation/OrderShipmentTaxesByZoneApplicatorSpec.php
+++ b/src/Sylius/Component/Core/spec/Taxation/OrderShipmentTaxesByZoneApplicatorSpec.php
@@ -106,6 +106,7 @@ class OrderShipmentTaxesByZoneApplicatorSpec extends ObjectBehavior
         $calculator->calculate(1000, $taxRate)->willReturn(0);
 
         $adjustmentsFactory->createWithData(Argument::any())->shouldNotBeCalled();
+        $order->addAdjustment(Argument::any())->shouldNotBeCalled();
 
         $this->apply($order, $zone);
     }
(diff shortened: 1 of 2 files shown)


src/Sylius/Component/Core/spec/Taxation/OrderShipmentTaxesByZoneApplicatorSpec.php
```diff
@@ -81,6 +82,34 @@ function it_applies_shipment_taxes_on_order_based_on_shipment_adjustments_and_ra
...
+        $taxRateResolver->resolve($shippingMethod, ['zone' => $zone])->willReturn($taxRate);
+
+        $order->getAdjustments(AdjustmentInterface::SHIPPING_ADJUSTMENT)->willReturn($shippingAdjustments);
+        $shippingAdjustments->isEmpty()->willReturn(false);
+        $shippingAdjustments->last()->willReturn($shippingAdjustment);
+        $shippingAdjustment->getAmount()->willReturn(1000);
+
+        $calculator->calculate(1000, $taxRate)->willReturn(0);
+
+        $adjustmentsFactory->createWithData(AdjustmentInterface::TAX_ADJUSTMENT, 'Simple tax (0%)', 0, false)->shouldNotBeCalled();
```
pjedrzejewski: I think here we can just say ``$adjustmentsFactory->createWithData(Argument:any())``, because none should be created.

michalmarcinkowski: And add also `$order->addAdjustment(Argument::any())->shouldNotBeCalled()`.

mmckelviereiss: OK, this has now been corrected.

mmckelviereiss: Done. :)

pamil: `$adjustmentsFactory->createWithData(Argument:any())` will match `$adjustmentsFactory->createWithData('foo')` but not `$adjustmentsFactory->createWithData('foo', 'bar')`. It's better to use `$adjustmentsFactory->createWithData(Argument::cetera())` instead as it will match every `createWithData` method call.

pjedrzejewski: @pamil is right. We should use cetera. :)

mmckelviereiss: Good, I've learnt something new here. Done. :)

pjedrzejewski: @mmckelviereiss :+1:

pjedrzejewski:
Thanks Mark! :+1:

Issue #4456, opened 2016-03-10
A coupon can only be used in a single order


gonzalovilaseca:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        |

In this commit https://github.com/Sylius/Sylius/commit/d537d28abed34138484de2410162a97fc4c6b8d6 the `many to many` relationship between orders and coupons was changed to a `one to one`. 
Therefore even if a coupon can be used several times you get an exception when saving the cart.
I guess it should be changed to `many to one`. 
If so I can make a PR for it.

pjedrzejewski:
Argh, that's a nasty bug. PR highly appreciated!

PR #4460, opened 2016-03-11
[CoreBundle] Fix coupon association

inssein:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/4456
| License         | MIT
| Doc PR          | --

I know @gonzalovilaseca said he would submit a PR, but I just came across this and needed the fix.

### DIFF
--- a/app/migrations/Version20160310181749.php
+++ b/app/migrations/Version20160310181749.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Sylius\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20160310181749 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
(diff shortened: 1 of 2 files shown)


michalmarcinkowski:
:+1: but migration file is missing :)

inssein:
@michalmarcinkowski ah good catch, I even generated one, but forgot to commit since I did this over the GitHub UI. Will add it when I get a chance.

inssein:
@michalmarcinkowski should be good to go now.

pjedrzejewski:
Restarted the build on PHP7.

inssein:
@pjedrzejewski failed again :( but it's not only just this PR

inssein:
ping @pjedrzejewski - is there anything else I need to do?

app/migrations/Version20160310181749.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace Sylius\Migrations;
```
Zales0123: I think there should be license block here - @pjedrzejewski did we agreed about this?

michalmarcinkowski:
Thanks Hussein! :+1:

app/migrations/Version20160310181749.php
michalmarcinkowski: Several migration files don't have this block, so I suppose we can add it to all of them in a separate PR.

Zales0123: :+1:

PR #4475, opened 2016-03-12
Province api tests

acrobat:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Tests/Controller/ProvinceApiTest.php
+++ b/src/Sylius/Bundle/CoreBundle/Tests/Controller/ProvinceApiTest.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\CoreBundle\Tests\Controller;
+
+use Lakion\ApiTestCase\JsonApiTestCase;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * @author Jeroen Thora <jeroen.thora@intracto.com>
+ */
+class ProvinceApiTest extends JsonApiTestCase
+{
+    /**
+     * @var array
+     */
+    private static $authorizedHeaderWithContentType = [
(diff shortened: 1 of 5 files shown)


src/Sylius/Bundle/CoreBundle/Tests/Controller/ProvinceApiTest.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
```
lchrusciel: Missing license block

acrobat: Fixed!

src/Sylius/Bundle/CoreBundle/Tests/Controller/ProvinceApiTest.php
```diff
@@ -0,0 +1,55 @@
...
+{
+    /**
+     * @var array
+     */
+    private static $authorizedHeaderWithContentType = [
+        'HTTP_Authorization' => 'Bearer SampleTokenNjZkNjY2MDEwMTAzMDkxMGE0OTlhYzU3NzYyMTE0ZGQ3ODcyMDAwM2EwMDZjNDI5NDlhMDdlMQ',
+        'CONTENT_TYPE' => 'application/json',
+    ];
+
+    public function testGetProvinceAccessDeniedResponse()
```
lchrusciel: In [CustomerApiTest](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Tests/Controller/CustomerApiTest.php) I introduced slightly different convention of naming tests. But unfortunately we wern't consistent with this change. I'm ok with these names but we should open a RFC about this convention

acrobat: Yes we weren't sure what the convention was, but as more tests had camelcase function names we assumed that camelcase would be the standard

lchrusciel: Absolutely. I just wanted to highlined a problem. Anyway, good work :+1:

pjedrzejewski:
Thanks Jeroen! :+1:

Issue #3989, opened 2016-01-27
Migrations - Code


inssein:
Did Sylius figure out migrations? I know the current ones are stored in app/migrations, but they are all the automatically generated ones.

The current problem is that if you are upgrading from before all the entities had a code field, your migrations will fail because a lot of entities add a code field, and then create a unique index. Creating the index will fail if there are more than two rows in the table, as both the code fields are empty.

I manually modified my migrations to copy over a default value, either the name, id, or something else that was relevant.

awerffeli:
i'm having the same issue when upgrading to v0.16. will probably go for the same solution as you did.

PR #4493, opened 2016-03-14
[Migration] Added missing license blocks to migrations

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT
| Doc PR          |

We should definitely keep in mind to check this in every PR, to avoid such a trivial pull request in the future ;)

### DIFF
--- a/app/migrations/Version20160212150554.php
+++ b/app/migrations/Version20160212150554.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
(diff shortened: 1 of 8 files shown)


inssein:
99% of the time these files are auto-generated - would it be trivial to auto-generate with the license?

michalmarcinkowski:
Thanks :+1:

michalmarcinkowski:
@inssein not sure if it is possible to add license to a file auto-generated by doctrine migrations. 

I know we didn't respond you in #3989, but we plan to start writing migrations that will actually migrate the data and help with upgrades (not only generate the schema update). I'm just not sure when we will be able to find time to figure out the proper workflow and setup tests for that.

PR #4507, opened 2016-03-15
[FixtureBundle] Throw exception when method does not exists

jverdeyen:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Throw an exception when a method in Datafixture class does not exists.
Currently I get segmentation faults when calling a "wrong" function.


pjedrzejewski:
Thanks Joeri! :+1:

PR #4514, opened 2016-03-15
[OptionValue][Core] - various fixes after option value model changes

### LINKED ISSUE #4449: [Variation] Product Options are broken on cyrillic locale
Prior to #4237 we had `name` field for internal use and translatable `presentation` to display for options. That internal name was useful to store transliterated name of an option which is then passed to a form builder and so on.

But with it gone and replaced with proper translatable `getPresentaion()` we now get an empty form name exception because Behat's `Urlizer::urlize()` cannot into cyrillic(it just returns empty string). 

Here is the [Breaking point](
https://github.com/Sylius/Sylius/blob/92949a767fd26ef54ecaade570eb86662998a80e/src/Sylius/Bundle/VariationBundle/Form/Type/VariantMatchType.php#L45)

Proposed solutions:

- Replace `Urlizer::urlize()` with something decent like `Slugify` which is available somewhere in `vendor` dir IIRC.
- Reintroduce a field to store transliterated name for internal use.

WDYT?

nakashu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #4449
| License         | MIT

Fixed:
* in inventory, where the script will throw an exception,
* cyrilc translations referenced in #4449 
* translation and validation strings and keys

Maybe for consistency all mentions of `presentation` should be changed to `name` in OptionValue and probably in productVariant. As it is proxied to the name anyway.

```php
    // OptionValue.php
    /**
     * {@inheritdoc}
     */
    public function getPresentation()
    {
        if (null === $this->option) {
            throw new \BadMethodCallException('The option have not been created yet so you cannot access proxy methods.');
        }

        return $this->option->getName();
    }
```

### DIFF
--- a/src/Sylius/Bundle/WebBundle/Resources/views/Backend/Inventory/macros.html.twig
+++ b/src/Sylius/Bundle/WebBundle/Resources/views/Backend/Inventory/macros.html.twig
@@ -72,7 +72,7 @@
                                     <td>
                                         <ul>
                                         {% for option in variant.options %}
-                                            <li><strong>{{ option.name }}</strong>: {{ option.value }}</li>
+                                            <li><strong>{{ option.presentation }}</strong>: {{ option.value }}</li>
                                         {% endfor %}
                                         </ul>
                                     </td>
--- a/src/Sylius/Bundle/WebBundle/Resources/views/Backend/ProductOption/macros.html.twig
+++ b/src/Sylius/Bundle/WebBundle/Resources/views/Backend/ProductOption/macros.html.twig
@@ -9,8 +9,6 @@
             <tr>
                 <th>{{ sylius_resource_sort('id', '#id') }}</th>
                 <th>{{ sylius_resource_sort('code', 'sylius.ui.code'|trans) }}</th>
-                {#TODO allow sort by translation field#}
-                {#<th>{{ sylius_resource_sort('presentation', 'sylius.product_option.presentation'|trans) }}</th>#}
                 <th>{{ 'sylius.product_option.name'|trans }}</th>
                 <th>{{ 'sylius.product_option.values'|trans }}</th>
                 <th>{{ sylius_resource_sort('updatedAt', 'sylius.product_option.updated_at'|trans) }}</th>
--- a/src/Sylius/Bundle/VariationBundle/Form/Type/OptionValueType.php
+++ b/src/Sylius/Bundle/VariationBundle/Form/Type/OptionValueType.php
@@ -45,7 +45,7 @@ class OptionValueType extends AbstractResourceType
         $builder
             ->add('translations', 'a2lix_translationsForms', [
                 'form_type' => sprintf('sylius_%s_option_value_translation', $this->variableName),
-                'label' => 'sylius.form.option.presentation',
+                'label' => 'sylius.form.option.name',
             ])
             ->addEventSubscriber(new AddCodeFormSubscriber())
         ;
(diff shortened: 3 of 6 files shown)


okwinza:
@nakashu can you please fix #4449 aswell since you're on it?

nakashu:
@okwinza done

okwinza:
@nakashu great, thanks! :+1:

src/Sylius/Bundle/WebBundle/Resources/views/Backend/ProductOption/macros.html.twig
```diff
@@ -10,7 +10,7 @@
                 <th>{{ sylius_resource_sort('id', '#id') }}</th>
                 <th>{{ sylius_resource_sort('code', 'sylius.ui.code'|trans) }}</th>
                 {#TODO allow sort by translation field#}
-                {#<th>{{ sylius_resource_sort('presentation', 'sylius.product_option.presentation'|trans) }}</th>#}
+                {#<th>{{ sylius_resource_sort('name', 'sylius.product_option.name'|trans) }}</th>#}
```
michalmarcinkowski: I assume this line can be removed since it's commented anyway.

src/Sylius/Bundle/WebBundle/Resources/views/Backend/Inventory/macros.html.twig
```diff
@@ -72,7 +72,7 @@
                                     <td>
                                         <ul>
                                         {% for option in variant.options %}
-                                            <li><strong>{{ option.name }}</strong>: {{ option.value }}</li>
+                                            <li><strong>{{ option.presentation }}</strong>: {{ option.value }}</li>
```
michalmarcinkowski: I suppose it was correct, wasn't it?

nakashu: It wasn't. It was throwing exception. Twig naming is misleading - here the option is `OptionValue` which now doesn't have a `name` only `value`. Presentation will get the name from assocatiated `Option` see codeblock in description.

nakashu:
@michalmarcinkowski commented part removed

src/Sylius/Bundle/WebBundle/Resources/views/Backend/Inventory/macros.html.twig
michalmarcinkowski: Yeah, that's confusing... Maybe to highlight it's not property we can explicitly call `option.getPresentation` here, wdyt?

nakashu: No problem, But there are more places of confusion like this I would say. 
* all options, which in real are optionValue should be renamed in twig and other places.  `{% for optionValue in variant.options %}`. wdyt?
* rename the method `getPresentation` to `getOptionName` which would be consitent with `getOptionCode`  already in the object, and resolve confusion of what you can expect from presentation method.

wdyt?

michalmarcinkowski: Agree with all :+1: 

**Notice:** There is on-going effort on moving to the new admin with grids, so if it's just simple rename then fine, but I wouldn't invest much time in current templates since they will be removed soon :wink:

nakashu: rename is simple. but the object level changes are more important. will update it later todat.

michalmarcinkowski: Method rename is definitely more important change, so I would definitely start with that :)

michalmarcinkowski:
Thanks @nakashu! :+1:

PR #4516, opened 2016-03-16
[Bug Fix] Fix orphan removal of channels and taxons on product model

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Channels and taxons relationships of a product should be removed when they are unset (orphaned). Right now they will stay in the join table, this is not the right behaviour.

It is late and I did not test this code, but it didn't make sense when I saw it. I have had that issue before with `many-to-many` relationships and it is important to properly define if it's orphans should live or not.


steffenbrem:
Nvm, I think I had a dream or something, seeing ghosts..

steffenbrem:
..

pjedrzejewski:
:D

PR #4522, opened 2016-03-16
Allow gedmo 3.0.x-dev to be installed 

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

I made this change since you want to use the `NestedSetTraits` etc. to easily move up nodes etc. Normally you would extend the `NestedSetRepository` but since Sylius uses a base repository class for all resources this is impossible with gedmo `2.4.x`. 

This way you can use the un-released `3.0.x` of gedmo extensions. So you could optionally use that when requiring Sylius in your project and use those traits!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
             "homepage": "http://github.com/Sylius/Sylius/contributors"
         }
     ],
+    "prefer-stable": true,
     "require": {
         "php": "^5.5.9|^7.0",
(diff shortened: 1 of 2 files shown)


steffenbrem:
Closed because I don't think this PR is valuable.

PR #4523, opened 2016-03-16
[Docker] Docker improvements

jverdeyen:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

* Changed docker php container timezone to Etc/UTC
* Added SYLIUS_APP_DEV_PERMITTED in php-fpm pool settings to allow access in app_dev.php
* Added zip library in php container


peteward:
Looking forward to giving Docker for Sylius a try in the near future, thanks for sharing your work here :+1:

pjedrzejewski:
Thanks Joeri! I tried it with Docker Toolbox, but I got some error and did not try again, will do this weekend maybe. :)

jverdeyen:
@pjedrzejewski Contact me if you need any help. Are you developing on a Mac? I can help you setting up an easy and fast way to develop Sylius in Docker containers. Dev page loads at 100-600ms.

pjedrzejewski:
@jverdeyen Yes, whole Lakion team is on macs, so that would be really awesome, do you have any link or smth? :)

jverdeyen:
@pjedrzejewski Yes, this is a post about how we develop with docker and PHP: http://tech.yappa.be/docker-php-development

But we are using our own tailermade images (which could also work for Sylius).

Our Vagrantfile is tweaked for performance over nfs and installs dnsdock and dockerui by default.
We also create `/vendor` in our php container and symlink this folder in our project folder (in the container). This way all vendors are installed inside the container and don't need to be read from the nfs mount. Which is a huge performance difference.

PR #4527, opened 2016-03-17
Prevent from twig InvalidParameterException

ahmadrabie:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT
| Doc PR          | -

If someone is in checkout process and the order is not finalized yet (order number is null) it rise an `InvalidParameterException: Parameter "number" for route "sylius_account_order_show" must match "[^/]++" ("" given) to generate a corresponding URL.`

### DIFF
--- a/src/Sylius/Bundle/WebBundle/Resources/views/Frontend/Account/Order/_list.html.twig
+++ b/src/Sylius/Bundle/WebBundle/Resources/views/Frontend/Account/Order/_list.html.twig
@@ -17,29 +17,31 @@
             </thead>
             <tbody>
             {% for order in orders %}
-                <tr class="order" id="order-{{ order.number }}">
-                    <td>{{ order.createdAt|date }}</td>
-                    <td>{{ order.number }}</td>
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
Not completed orders should not be displayed there, so that is a bug in account section loading wrong orders.

nakashu:
I'll just add when its here, that same is happening in Admin Dashboard

ahmadrabie:
@pjedrzejewski How can I fix this problem?
I have 2 soloution:
1. Adding new method to order repository to select only customer orders that have number
2- Changing the account/OrderController

What is your suggestion?

aramalipoor:
IMO we can introduce `findComfirmedOrders` in OrderRepository !?

michalmarcinkowski:
@aramalipoor I would call this method simply `findConfirmed`.

michalmarcinkowski:
Thanks for reporting, that was fixed in #4902.

PR #4542, opened 2016-03-17
[ResourceBundle] Allow applying transition when updating resources

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | -
| License         | MIT
| Doc PR          | -

Summary:

* Renamed ``updateStateAction`` to ``applyStateMachineTransitionAction`` in ResourceController;
* Extracted logic to separate service;
* Reimplemented the action from scratch BDD wayyy;
* Added ability to execute transition during updateAction :tada: 

The most important is the last point. Use cases:

Any action that needs to edit something on the resource and apply transition. Simplest example: shipping a shipment. We want to specify tracking code and apply "ship" transition - doable with routing configuration alone now.

More crazy use-case: simplified checkout process. Our checkout right now is:

* edit order via form (change address, select shipping method, select payment method)
* apply transition
* save
* redirect

Using FlowBundle for all that is quite an overkill and does not work via API nicely. With this action we can do it with HTML/XML/JSON. :D I recall pasing some examples in another issue, but can't find it.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
@@ -557,4 +557,28 @@ class RequestConfiguration
 
         return $this->parameters->get('grid');
     }
+
+    /**
+     * @return bool
+     */
+    public function hasStateMachine()
+    {
+        return $this->parameters->has('state_machine');
+    }
+
+    /**
+     * @return string
+     */
+    public function getStateMachineGraph()
+    {
+        return $this->parameters->get('state_machine[graph]', null, true);
+    }
+
+    /**
+     * @return string
+     */
+    public function getStateMachineTransition()
+    {
+        return $this->parameters->get('state_machine[transition]', null, true);
+    }
 }
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
@@ -404,7 +404,7 @@ class ResourceController extends Controller
         $this->eventDispatcher->dispatchPostEvent(ResourceActions::UPDATE, $configuration, $resource);
 
         if (!$configuration->isHtmlRequest()) {
-            return $this->viewHandler->handle($configuration, View::create(null, 204));
+            return $this->viewHandler->handle($configuration, View::create($resource, 200));
         }
 
         $this->flashHelper->addSuccessFlash($configuration, ResourceActions::UPDATE, $resource);
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
@@ -103,6 +103,11 @@ class ResourceController extends Controller
     protected $eventDispatcher;
 
     /**
+     * @var StateMachineInterface
+     */
+    protected $stateMachine;
+
+    /**
      * @param MetadataInterface $metadata
      * @param RequestConfigurationFactoryInterface $requestConfigurationFactory
      * @param ViewHandlerInterface $viewHandler
@@ -117,6 +122,7 @@ class ResourceController extends Controller
      * @param FlashHelperInterface $flashHelper
      * @param AuthorizationCheckerInterface $authorizationChecker
      * @param EventDispatcherInterface $eventDispatcher
+     * @param StateMachineInterface $stateMachine
      */
     public function __construct(
         MetadataInterface $metadata,
@@ -132,7 +138,8 @@ class ResourceController extends Controller
         RedirectHandlerInterface $redirectHandler,
         FlashHelperInterface $flashHelper,
         AuthorizationCheckerInterface $authorizationChecker,
-        EventDispatcherInterface $eventDispatcher
+        EventDispatcherInterface $eventDispatcher,
+        StateMachineInterface $stateMachine
     ) {
         $this->metadata = $metadata;
         $this->requestConfigurationFactory = $requestConfigurationFactory;
@@ -148,6 +155,7 @@ class ResourceController extends Controller
         $this->flashHelper = $flashHelper;
         $this->authorizationChecker = $authorizationChecker;
         $this->eventDispatcher = $eventDispatcher;
+        $this->stateMachine = $stateMachine;
     }
 
     /**
@@ -298,6 +306,10 @@ class ResourceController extends Controller
                 return $this->redirectHandler->redirectToResource($configuration, $resource);
             }
 
+            if ($configuration->hasStateMachine()) {
+                $this->stateMachine->apply($configuration, $resource);
+            }
+
             $this->manager->flush();
             $this->eventDispatcher->dispatchPostEvent(ResourceActions::UPDATE, $configuration, $resource);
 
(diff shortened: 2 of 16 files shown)


liverbool:
How do you think about adding some thing like `showPossibleTransitionAction` to get possible transitions?

Response:

```
{
    resource:  { .... },
    transitions: [transitionA, transitionB, ..],
    translations: {
        transitionA: 'label text',
        ....
    }
}
```

src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
```diff
@@ -557,4 +557,28 @@ public function getGrid()
...
         return $this->parameters->get('grid');
     }
+
+    /**
+     * @return bool
+     */
+    public function hasStateMachine()
+    {
+        return $this->parameters->has('state_machine');
+    }
```
liverbool: The function `hasStateMachine` should return something like this instead:
`return $this->parameters->get('state_machine[transition]', false, true);`
Because some cases would have no transition to continue and it may throw an error. 

For example, if an object is in the end of states, when we submit(update) the form to `updateAction` then the key `state_machine[transition]` will be null but the key `state_machine` will still exists.

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
```diff
@@ -368,6 +380,43 @@ public function deleteAction(Request $request)
...
+            return $this->redirectHandler->redirectToResource($configuration, $resource);
+        }
+
+        $this->stateMachine->apply($configuration, $resource);
+        $this->manager->flush();
+
+        $this->eventDispatcher->dispatchPostEvent(ResourceActions::UPDATE, $configuration, $resource);
+
+        if (!$configuration->isHtmlRequest()) {
+            return $this->viewHandler->handle($configuration, View::create(null, 204));
```
liverbool: Amm, I think should return `200` status and response `resource` to get new state of resource, The client can't know about what's new state changed to.

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
```diff
@@ -298,6 +306,10 @@ public function updateAction(Request $request)
                 return $this->redirectHandler->redirectToResource($configuration, $resource);
             }
 
+            if ($configuration->hasStateMachine()) {
+                $this->stateMachine->apply($configuration, $resource);
+            }
```
liverbool: Will throw exception (can't apply) when transition reach the end.

liverbool:
ping @pjedrzejewski ;)

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
aramalipoor: Maybe show a fail flash message on exception? (e.g. Cannot transition to X)

liverbool: When we have no transition to apply `X = null`, It's not an exception.

pjedrzejewski: And that seems correct, you should not apply it twice or from invalid state, hm?

src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
pjedrzejewski: The idea is to only have dedicated routes for applying these transitions, not generic update action, so this should fail if you try to apply invalid transition.

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
liverbool: `$configuration->hasStateMachine()` away return `true` when we try to update form even though it has no transition.

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
pjedrzejewski: Changed! 👍

src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourceControllerSpec.php
```diff
@@ -1242,6 +1305,222 @@ function it_does_not_delete_a_resource_and_throws_http_exception_for_non_html_re
...
+        $authorizationChecker->isGranted($configuration, 'sylius.product.update')->willReturn(true);
+        $singleResourceProvider->get($configuration, $repository)->willReturn(null);
+
+        $this
+            ->shouldThrow(new NotFoundHttpException())
+            ->during('applyStateMachineTransitionAction', [$request])
+        ;
+    }
+
+    function it_applies_state_machine_transition_to_resource_and_redirects_by_for_html_request(
```
michalmarcinkowski: redirects_~~by_~~for_html

src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourceControllerSpec.php
```diff
@@ -1242,6 +1305,222 @@ function it_does_not_delete_a_resource_and_throws_http_exception_for_non_html_re
...
+        $eventDispatcher->dispatchPostEvent(ResourceActions::UPDATE, $configuration, $resource)->shouldNotBeCalled();
+        $flashHelper->addSuccessFlash($configuration, ResourceActions::UPDATE, $resource)->shouldNotBeCalled();
+
+        $flashHelper->addFlashFromEvent($configuration, $event)->shouldBeCalled();
+        $redirectHandler->redirectToResource($configuration, $resource)->willReturn($redirectResponse);
+
+        $this->applyStateMachineTransitionAction($request)->shouldReturn($redirectResponse);
+    }
+
+    function it_applies_state_machine_transition_on_resource_and_returns_204_for_non_html_requests(
```
michalmarcinkowski: returns_20~~4~~**0**

src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourceControllerSpec.php
```diff
@@ -1242,6 +1305,222 @@ function it_does_not_delete_a_resource_and_throws_http_exception_for_non_html_re
...
+
+        $stateMachine->apply($configuration, $resource)->shouldNotBeCalled();
+        $objectManager->flush()->shouldNotBeCalled();
+
+        $eventDispatcher->dispatchPostEvent(Argument::any())->shouldNotBeCalled();
+        $flashHelper->addSuccessFlash(Argument::any())->shouldNotBeCalled();
+        $flashHelper->addFlashFromEvent(Argument::any())->shouldNotBeCalled();
+
+        $this
+            ->shouldThrow(new HttpException(500, 'Cannot approve this product.'))
```
michalmarcinkowski: IMO we should throw `400` for API request when state machine transition cannot be applied.

pjedrzejewski: This code is configurable when stopping action via event. ;) 500 is just example here.

michalmarcinkowski:
That will be useful, nice work! 👍

PR #4548, opened 2016-03-18
[Core] Checkout flow and errors improvements

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT
| Doc PR          |

- previous implementation led to confusion; if transition is invalid, then state machine is wrongly configured or wrongly used and it should be shown
- added checking order state before applying some transitions

### DIFF
--- a/src/Sylius/Component/Core/Exception/InvalidTransitionException.php
+++ b/src/Sylius/Component/Core/Exception/InvalidTransitionException.php
@@ -12,6 +12,7 @@
 namespace Sylius\Component\Core\Exception;
 
 use SM\SMException;
+use SM\StateMachine\StateMachineInterface;
 
 /**
  * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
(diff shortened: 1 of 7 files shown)


src/Sylius/Component/Core/Exception/InvalidTransitionException.php
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class InvalidTransitionException extends SMException
+{
+    /**
+     * @param string $transition
+     */
+    public function __construct($transition)
```
pamil: DX: Can we also pass an optional list of available states? Or whole `StateMachine` and build even more helpful message? "Transition foo is invalid for bar state machine. Available states are: rynkowski, krawczyk, hołdys"?

Zales0123: Not a bad idea :+1:

okwinza: :+1: on this

pjedrzejewski:
Thanks Mateusz!

Issue #4374, opened 2016-03-02
[Admin] Managing countries


pjedrzejewski:
* [x] Adding country
* [x] Adding country with provinces
* [x] Validation of country code uniqueness
* [x] Validation of province required fields
* [x] Validation of province code uniqueness
* [x] Editing country (checkbox)
* [x] Immutability of country code while editing
* [x] Adding and removing provinces while editing
* [ ] Toggling country (via Ajax) on index

PR #4565, opened 2016-03-21
[Admin] Country code immutability

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #4374 
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Page/Admin/Country/UpdatePage.php
+++ b/src/Sylius/Behat/Page/Admin/Country/UpdatePage.php
@@ -55,12 +55,9 @@ class UpdatePage extends BaseUpdatePage implements UpdatePageInterface
      */
     public function isCodeFieldDisabled()
     {
-        try {
-            $codeField = $this->getElement('code');
-            return $codeField->getAttribute('disabled') === 'disabled';
-        } catch (ElementNotFoundException $exception) {
-            return false;
-        }
+        $codeField = $this->getElement('code');
+
+        return $codeField->getAttribute('disabled') === 'disabled';
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/Sylius/Behat/Service/Accessor/NotificationAccessor.php
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * @return NodeElement
+     *
+     * @throws ElementNotFoundException
+     */
+    private function getMessageElement()
+    {
+        $messageElement = $this->createMessageElement();
```
pamil: Can't we just use `$this->session->find('css', '.message')` instead?

Arminek: `$this->session->getDriver()->find()` won't work with css it operates on xpath only but i think we can change this, cuz i can get `selectorsHandler` from session.

pamil: Ouch, I meant `$this->session->getPage()->find('css', '.message')` :) It works with the css selector.

src/Sylius/Behat/Page/Admin/Country/UpdatePage.php
```diff
@@ -49,6 +51,19 @@ public function disable()
...
     /**
+     * {@inheritdoc}
+     */
+    public function isCodeFieldDisabled()
+    {
+        try {
+            $codeField = $this->getElement('code');
+            return $codeField->getAttribute('disabled') === 'disabled';
+        } catch (ElementNotFoundException $exception) {
+            return false;
```
michalmarcinkowski: Do not catch exception, it will indicate that something is wrong instead of incorrectly saying that code field is not disabled.

pjedrzejewski: :+1:

Arminek: Method name indicate that it should return boolean value so IMO we shouldn't throw exception.

pjedrzejewski:
Thank you Arek!

PR #4571, opened 2016-03-21
Drop support for PHP 5.5

### LINKED ISSUE #4567: [RFC] Dropping support for PHP 5.5
Hello!

As PHP 5.5 loses the security support on 10 Jul 2016, we need to consider whether Sylius should support it longer than that.

Supporting PHP 5.5 requires us to check every commit against it, which results in longer feedback cycle (especially now, while there are single components and bundles tested too). It also prevents us from using awesome, new features like variadics or constant expressions.

IMHO Sylius as a project should enforce some good practices - dropping support together with PHP security updates is one of them. Personally I'd like to support only PHP 7, but we also need the people that uses the project :)

We must have in mind that when Sylius hits 1.0, we will have to support the minimal PHP version until 2.0 comes.

Moreover, PHP 5.6 is LTS version (security support ends up on 31 Dec 2018) and PhpSpec 3 will require at least that one.

So, the question is: Should we support PHP 5.5 after the PHP drops security support for it or should we go for PHP 5.6?

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | fixes #4567
| License         | MIT
| Doc PR          | -

As discussed in #4567 :tada:


michalmarcinkowski:
@pamil let's refresh this PR! 💃

pamil:
Rebased :tada:

bendavies:
variadics were introduced in 5.6, so all instances of `func_get_args()` in the code could be replaced.

pjedrzejewski:
@bendavies Good idea, but let's do this in a separate PR! 👍

pjedrzejewski:
Thank you Kamil!

PR #4575, opened 2016-03-21
[Travis] Reset all composer.lock caches on composer.json update

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT
| Doc PR          | -

Eg. if ResourceBundle adds a new dependency, and ThemeBundle depends on ResourceBundle, ThemeBundle cache should be invalidated just as ResourceBundle one

### DIFF
--- a/etc/bash/packages.lib.sh
+++ b/etc/bash/packages.lib.sh
@@ -32,3 +32,25 @@ cast_package_argument_to_package_path() {
 
     echo "${package_path}"
 }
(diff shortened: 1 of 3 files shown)


pjedrzejewski:
Nice, thanks Kamil!

PR #4597, opened 2016-03-22
[DX] Add 'an' to administrator log in step

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | 
| License         | MIT
| Doc PR          |

### DIFF
--- a/src/Sylius/Behat/Context/Setup/SecurityContext.php
+++ b/src/Sylius/Behat/Context/Setup/SecurityContext.php
@@ -82,9 +82,9 @@ final class SecurityContext implements Context
     }
 
     /**
-     * @Given I am logged in as administrator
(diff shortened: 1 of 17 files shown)


src/Sylius/Behat/Context/Setup/SecurityContext.php
```diff
@@ -82,7 +82,7 @@ public function iAmLoggedInCustomer()
     }
 
     /**
-     * @Given I am logged in as administrator
+     * @Given I am logged in as an administrator
      */
     public function iAmLoggedInAsAdministrator()
```
TheMadeleine: in the function name also, maybe?

PR #4620, opened 2016-03-24
[WIP][ThemeBundle] Context aware themes

### LINKED ISSUE #4280: [ThemeBundle] Context aware themes
Hi,

I found your ThemeBundle meeting most of my requirements and it looks really nice - good job guys! Although, it misses one major feature imo.

I'm developing a multi-tenancy app, and one of my requirements is to make themes which are tenant context aware (it can be also some other context aware, e.g. channel context in your case, so in general: context aware themes).

This will give me more flexibility:
- it will be possible to have a theme per channel (in your case)/tenant (in my case)
- themes synchronization could be done for current context only. For example: once I add a new theme for `tenant/channel A` I would want to synchronize all themes only for the `A` context - I don't need to synchronize it for all channels/tenants while it was only changed for single tenant/channel.

This would be also nice to make it configurable so you can use this feature or just the basic behaviour.

What do you think about it? Or maybe this can be achieved already and I'm not aware of it?

Best,

takeit:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | fixes #4280 
| License         | MIT
| Doc PR          | -

New features:
- [x] - possibility to store themes under context aware directories

If a config setting `context_aware_paths` under `sylius_theme -> context` is set to `true` in config file and if a name of the current context is given (see `ChannelBasedThemeContext` class - `getName()` method), themes will be loaded from the directory: e.g. `/themes/<context_name>/<theme>`.

If a config setting `context_aware_paths` is set to `true` and theme context name is `null`, it will throw an exception:
```
You have enabled "context_aware_paths" setting but your "%s" method must return a name!
```
**Note**
The code will break if someone uses a custom theme context with current implementation (without these changes), because in this PR I introduced `getName` method in `ThemeContextInterface`, so it can allow to get the context name on which we can rely later on to keep themes per channel (channel name in Sylius case) under the different subdirectories.

- [x] - implemented theme events: update / remove / add

TODO
- [ ] - fix RecursiveFileLocator spec
- [ ] - add unit tests for configuration (phpunit)

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Theme/ChannelBasedThemeContext.php
+++ b/src/Sylius/Bundle/CoreBundle/Theme/ChannelBasedThemeContext.php
@@ -18,6 +18,7 @@ use Sylius\Component\Core\Model\ChannelInterface;
 
 /**
  * @author Kamil Kokot <kamil.kokot@lakion.com>
+ * @author Rafał Muszyński <rafal.muszynski@sourcefabric.org>
  */
 final class ChannelBasedThemeContext implements ThemeContextInterface
 {
@@ -41,7 +42,7 @@ final class ChannelBasedThemeContext implements ThemeContextInterface
     {
         try {
             /** @var ChannelInterface $channel */
-            $channel = $this->channelContext->getChannel();
+            $channel = $this->getChannel();
 
             return $channel->getTheme();
         } catch (ChannelNotFoundException $exception) {
@@ -49,5 +50,25 @@ final class ChannelBasedThemeContext implements ThemeContextInterface
         } catch (\Exception $exception) {
             return null;
         }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        try {
+            /** @var ChannelInterface $channel */
+            $channel = $this->getChannel();
+
+            return $channel->getCode();
+        } catch (ChannelNotFoundException $exception) {
+            return null;
+        }
+    }
+
+    private function getChannel()
+    {
+        return $this->channelContext->getChannel();
     }
 }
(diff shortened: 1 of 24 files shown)


takeit:
Please let me know what you guys think about it /cc @pamil @pjedrzejewski

pamil:
I may not fully understand the whole "context" thingy, but it looks quite complicated. I was also doing multi-tenant application using ThemeBundle and you need the context (tenant?) only when you are letting the user choose one. IMHO there is no need for including this logic in the bundle.

As for the domain events, I can't see where these events are dispatched. It's a good idea that will give us more flexibility while changing themes, but I will be rather careful with those optional parameters (like `ThemeSynchronizerInterface::synchronize(ThemeInterface $theme = null)`) - for me it will be more suitable to define independent `synchronize()` and `synchronizeTheme(ThemeInterface $theme)` methods.

takeit:
> I may not fully understand the whole "context" thingy, but it looks quite complicated. I was also doing multi-tenant application using ThemeBundle and you need the context (tenant?) only when you are letting the user choose one. IMHO there is no need for including this logic in the bundle.

How do you store themes per tenant then in your filesystem? do you keep all themes under e.g. `app/themes` dir for all tenants?

What If I want to keep themes which I want to be available only for the tenant that I want? meaning I want to separate how the themes are stored to not use themes which should belong to other tenants.

Currently, when you synchronize themes, it will load (from the configured directory) and persist all of them into the db, which imo is wrong in case of multi-tenancy app, instead themes should be separated per tenant:

- `app/themes/<tenant1>/<theme>`
- `app/themes/<tenant2>/<theme>`

Another thing is that if you want to use the same theme (same `name` in composer.json of the theme) for let's say 5 different tenants, you won't be able to do it because synchronization command will initialize the first one matching the name found in `composer.json`.

> As for the domain events, I can't see where these events are dispatched. It's a good idea that will give us more flexibility while changing themes, but I will be rather careful with those optional parameters (like ThemeSynchronizerInterface::synchronize(ThemeInterface $theme = null)) - for me it will be more suitable to define independent synchronize() and synchronizeTheme(ThemeInterface $theme) methods.

:+1:

PR #4622, opened 2016-03-25
[RFC][DX] Introduce NotificationChecker

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | 
| License         | MIT
| Doc PR          | 

Just a thought. Because of usage of NotificationAccessor we had to duplicate some line, so I just extracted it to separated service, just to keep DRY. I haven't coupled it with NotificationAccessor, because it can be useful somewhere else and writing test for it is pretty painful also.

_Edit_
After introducing new Exception output for developer will look like this:
zrzut ekranu 2016-03-30 o 16 58 05

### DIFF
--- a/etc/behat/services.xml
+++ b/etc/behat/services.xml
@@ -25,6 +25,7 @@
         <parameter key="sylius.behat.mocker.class">Sylius\Behat\Service\Mocker\Mocker</parameter>
         <parameter key="sylius.behat.response_loader.class">Sylius\Behat\Service\ResponseLoader</parameter>
         <parameter key="sylius.behat.notification_accessor.class">Sylius\Behat\Service\Accessor\NotificationAccessor</parameter>
+        <parameter key="sylius.behat.service.notification_checker.class">Sylius\Behat\Service\NotificationChecker</parameter>
     </parameters>
     <services>
         <service id="mink.default_session" class="Behat\Mink\Session" factory-service="mink" factory-method="getSession" lazy="true" scope="scenario" public="false" />
@@ -58,6 +59,9 @@
 
         <service id="sylius.behat.notification_accessor" class="%sylius.behat.notification_accessor.class%" scope="scenario" public="false">
             <argument type="service" id="mink.default_session" />
+        </service>
+        <service id="sylius.behat.notification_checker" class="%sylius.behat.service.notification_checker.class%" scope="scenario" public="false">
+            <argument type="service" id="sylius.behat.notification_accessor" />
         </service>
     </services>
 </container>
--- a/src/Sylius/Behat/Exception/NotificationExpectationMismatchException.php
+++ b/src/Sylius/Behat/Exception/NotificationExpectationMismatchException.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Behat\Exception;
+
+/**
+ * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
+ */
+final class NotificationExpectationMismatchException extends \RuntimeException
+{
+    /**
+     * @param string $missingNotificationType
+     * @param string $missingNotification
+     * @param string $foundedNotificationType
+     * @param string $foundedNotification
+     * @param int $code
+     * @param \Exception|null $previous
+     */
+    public function __construct(
+        $missingNotificationType,
+        $missingNotification,
+        $foundedNotificationType,
+        $foundedNotification,
+        $code = 0, 
+        \Exception $previous = null
+    ) {
+        $message = sprintf(
+            "Expected *%s* notification with a \"%s\" message was not found.\n *%s* notification with a \"%s\" message has been found.",
+            $missingNotificationType,
+            $missingNotification,
+            $foundedNotificationType,
+            $foundedNotification
+        );
+
+        parent::__construct($message, $code, $previous);
+    }
+}
--- a/src/Sylius/Behat/Service/NotificationChecker.php
+++ b/src/Sylius/Behat/Service/NotificationChecker.php
@@ -1,0 +1,100 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Behat\Service;
+
+use Sylius\Behat\Exception\NotificationExpectationMismatchException;
+use Sylius\Behat\Service\Accessor\NotificationAccessorInterface;
+use Webmozart\Assert\Assert;
+
+/**
+ * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
+ */
+final class NotificationChecker implements NotificationCheckerInterface
+{
+    /**
+     * @var NotificationAccessorInterface
+     */
+    private $notificationAccessor;
+
+    /**
+     * @param NotificationAccessorInterface $notificationAccessor
+     */
+    public function __construct(NotificationAccessorInterface $notificationAccessor)
+    {
+        $this->notificationAccessor = $notificationAccessor;
+    }
+
+    /**
+     * {@inheritdoc}
+     * 
+     * @throws NotificationExpectationMismatchException
+     */
+    public function checkCreationNotification($resource)
+    {
+        $message = sprintf('Success %s has been successfully created.', $this->humanizeResourceName($resource));
+
+        $this->checkSuccessNotificationMessage($message);
+    }
+
+    /**
+     * {@inheritdoc}
+     * 
+     * @throws NotificationExpectationMismatchException
+     */
+    public function checkDeletionNotification($resource)
+    {
+        $message = sprintf('Success %s has been successfully deleted.', $this->humanizeResourceName($resource));
+
+        $this->checkSuccessNotificationMessage($message);
+    }
+
+    /**
+     * {@inheritdoc}
+     * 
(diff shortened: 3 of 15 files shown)


etc/behat/services.xml
```diff
@@ -30,6 +30,7 @@
         <parameter key="sylius.behat.page.admin.crud.create.class">Sylius\Behat\Page\Admin\Crud\CreatePage</parameter>
         <parameter key="sylius.behat.page.admin.crud.update.class">Sylius\Behat\Page\Admin\Crud\UpdatePage</parameter>
         <parameter key="sylius.behat.service.accessor.notification_accessor.class">Sylius\Behat\Service\Accessor\NotificationAccessor</parameter>
+        <parameter key="sylius.behat.service.notification_validator.class">Sylius\Behat\Service\NotificationValidator</parameter>
```
pjedrzejewski: Perhaps we should be consistent and name these:

``sylius.behat.service.accessor.notification``
``sylius.behat.service.validator.notification``

?

Or maybe name this... Asserter!!! :smiling_imp:

pamil: IMO `sylius.behat.service.notification_{validator,accessor}` seems the best for me, as grouping them by accessor / validator doesn't really help - I need a notification validator, not a validator for notifications.

michalmarcinkowski: :+1: for services names. Asserter doesn't sound good for me :smile: I would leave Validator :)

pamil: As for name - asserter / validator - the verb used in the interface suggests that it's a notification checker :)

src/Sylius/Behat/Service/NotificationValidator.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * @param string $resource
+     */
+    public function checkDeletionNotification($resource)
+    {
+        $this->assertPositiveMessage();
+
+        Assert::true(
```
pamil: Can we use a custom exception?

src/Sylius/Behat/Service/NotificationValidatorInterface.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+namespace Sylius\Behat\Service;
+
+/**
+ * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
+ */
+interface NotificationValidatorInterface
+{
+    /**
+     * @param string $resource
```
pamil: So that we can define that it really gives a feedback, right now it shouldn't throw any exception or return anything. `@throws NotificationNotFoundException`.

etc/behat/services.xml
pjedrzejewski: Asserter was a joke, but I am +1 for Checker. Validator can be confusing.

src/Sylius/Behat/Service/NotificationValidatorInterface.php
pamil: The exception constructor may receive the searched notification and all existing notifications, so that we can display a really nice message for the developer, like `Expected *success* notification with "lorem ipsum" message, got *error* "abc def", *success* "ghi jpg"`.

etc/behat/services.xml
michalmarcinkowski: +1 for Checker

src/Sylius/Behat/Service/NotificationValidator.php
```diff
@@ -0,0 +1,78 @@
...
+            $this->notificationAccessor->isSuccessfullyDeletedFor($resource),
```
pamil: Should notification accessor still have issers?

src/Sylius/Behat/Service/NotificationValidator.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        $this->notificationAccessor = $notificationAccessor;
+    }
+
+    /**
+     * @param string $resource
+     */
+    public function checkDeletionNotification($resource)
+    {
+        $this->assertPositiveMessage();
```
pamil: Anyway, what if there is more than one notification?

etc/behat/services.xml
lchrusciel: :+1:

src/Sylius/Behat/Service/NotificationValidator.php
lchrusciel: Sure, because we are still relying on this assertions. Even if we introduce custom exception

src/Sylius/Behat/Service/NotificationValidator.php
lchrusciel: I guess it is not a possible scenario... /cc @michalmarcinkowski

michalmarcinkowski: Probably right now there is no scenario where this will happen, but it is definitely a possible scenario. However I would leave it for now, there will be only one place to change this logic if we will find it buggy.

src/Sylius/Behat/Exception/NotificationExpectationMismatchException.php
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Behat\Exception;
+
+use Exception;
```
Zales0123: I don't think it's needed.

src/Sylius/Behat/Service/NotificationChecker.php
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    /**
+     * @param string $resource
+     *
+     * @throws NotificationExpectationMismatchException
+     */
+    public function checkCreationNotification($resource)
+    {
+        $message = sprintf('Success %s has been successfully created.', $this->humanizeResourceName($resource));
```
Zales0123: "**Success** %s has been successfully created."? :smile: Sth has gone wrong here.

src/Sylius/Behat/Service/NotificationChecker.php
```diff
@@ -0,0 +1,105 @@
...
+     */
+    public function checkEditionNotification($resource)
+    {
+        $message = sprintf('Success %s has been successfully updated.', $this->humanizeResourceName($resource));
+
+        return $this->checkSuccessNotificationMessage($message);
+    }
+
+    /**
+     * @param $message
```
Zales0123: Missing type hint.

src/Sylius/Behat/Service/NotificationCheckerInterface.php
```diff
@@ -0,0 +1,32 @@
...
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+namespace Sylius\Behat\Service;
```
Zales0123: Missing blank line before.

src/Sylius/Behat/Service/NotificationChecker.php
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * @param NotificationAccessorInterface $notificationAccessor
+     */
+    public function __construct(NotificationAccessorInterface $notificationAccessor)
+    {
+        $this->notificationAccessor = $notificationAccessor;
+    }
+
+    /**
+     * @param string $resource
```
Zales0123: Use ``{@inheritdoc}`` here and in all methods below.

src/Sylius/Behat/Service/NotificationCheckerInterface.php
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * @param string $resource
+     */
+    public function checkCreationNotification($resource);
+
+    /**
+     * @param string $resource
+     */
+    public function checkEditionNotification($resource);
+}
```
Zales0123: Shouldn't ``checkSuccessNotificationMessage`` be also in this interface?

src/Sylius/Behat/Exception/NotificationExpectationMismatchException.php
```diff
@@ -0,0 +1,45 @@
...
+    public function __construct(
+        $missingNotificationType,
+        $missingNotification,
+        $foundedNotificationType,
+        $foundedNotification,
+        $code = 0, 
+        \Exception $previous = null
+    ) {
+        $message = sprintf(
+            "Expected *%s* notification with a \"%s\" message was not found.\n *%s* notification with a \"%s\" message has been founded.",
```
tristanbes: `has been found` instead of `has been founded` no ? Unless the message has raised a lot of money and indeed has been funded :D

src/Sylius/Behat/Service/NotificationChecker.php
lchrusciel: Unfortunately, everything is correct. 
zrzut ekranu 2016-03-31 o 08 58 43

Zales0123: It's just misleading that message starts with "Success" and then, without any blank line (or sth similar), there is flash message text. Can't we just check ``'%s has been successfully created.'``?

etc/behat/services.xml
```diff
@@ -25,6 +25,7 @@
         <parameter key="sylius.behat.mocker.class">Sylius\Behat\Service\Mocker\Mocker</parameter>
         <parameter key="sylius.behat.response_loader.class">Sylius\Behat\Service\ResponseLoader</parameter>
         <parameter key="sylius.behat.notification_accessor.class">Sylius\Behat\Service\Accessor\NotificationAccessor</parameter>
+        <parameter key="sylius.behat.service.notification_checker.class">Sylius\Behat\Service\NotificationChecker</parameter>
```
pjedrzejewski: In other PR you removed "service" prefix. I agree.

pjedrzejewski:
:+1:

Issue #4614, opened 2016-03-24
[Addressing] Missing locale code converter


Arminek:
Based on https://github.com/Sylius/Sylius/pull/4594

Arminek:
Solved

pjedrzejewski:
:+1:

PR #4640, opened 2016-03-29
[Locale] Locale name converter

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | https://github.com/Sylius/Sylius/issues/4614
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | https://github.com/Sylius/Sylius/issues/4614
| License         | MIT


pjedrzejewski:
I am not sure we should merge these 2 interfaces. NameCoverterInterface is generic indeed, but 2 similar interfaces (even if identical) are not the same abstraction.

Also, Addressing does not seem like an appropriate component for the locale converter. Perhaps that should go to LocaleBundle (still not perfect, but more logical).

pjedrzejewski:
Thanks Arek!

Issue #3268, opened 2015-09-11
[BUG] Pagerfanta Outputwalker breaks sorting in doctrine repository


koemeet:
The PR #3185 is causing a lot of problems for me. Whevener I apply sorting to the repository, the whole SQL query is wrong. This is the error I am getting:

``` sql
An exception occurred while executing 
    'SELECT DISTINCT id_4 
        FROM (
                SELECT m0_.available_on AS available_on_0, m0_.created_at AS created_at_1, 
                    m0_.updated_at AS updated_at_2, m0_.deleted_at AS deleted_at_3, m0_.id AS id_4, 
                    m0_.variant_selection_method AS variant_selection_method_5, m1_.name AS name_6, 
                    m1_.slug AS slug_7, m1_.description AS description_8, m1_.meta_keywords AS meta_keywords_9, 
                    m1_.meta_description AS meta_description_10, m1_.id AS id_11, 
                    m1_.short_description AS short_description_12, m1_.locale AS locale_13,
                    m2_.is_master AS is_master_14, m2_.presentation AS presentation_15, m2_.created_at AS created_at_16, 
                    m2_.updated_at AS updated_at_17, m2_.deleted_at AS deleted_at_18, m2_.id AS id_19, 
                    m2_.available_on AS available_on_20, m2_.sku AS sku_21, m2_.on_hold AS on_hold_22, m2_.on_hand AS on_hand_23, 
                    m2_.sold AS sold_24, m2_.available_on_demand AS available_on_demand_25, m2_.price AS price_26, 
                    m2_.pricing_calculator AS pricing_calculator_27, m2_.pricing_configuration AS pricing_configuration_28, 
                    m2_.width AS width_29, m2_.height AS height_30, m2_.depth AS depth_31, m2_.weight AS weight_32 
                FROM mango_product m0_ LEFT JOIN mango_product_translation m1_ ON m0_.id = m1_.translatable_id LEFT JOIN mango_product_variant m2_ ON m0_.id = m2_.product_id
            ) dctrn_result 
        ORDER BY m0_.updated_at DESC 
        LIMIT 10 OFFSET 0':\n\nSQLSTATE[42S22]: Column not found: 1054 Unknown column 'm0_.updated_at' in 'order clause'
```

I use the `resource-bundle` and the `product-bundle` and have setup a route exactly like the `core-bundle` does:

``` yaml
mango_api_product_index:
    path: /products
    methods: [GET]
    defaults:
[...]

PR #4646, opened 2016-03-30
[Promotion] Don't lock Doctrine to version 2.4.x only

stloyd:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Without this change I cannot use newest versions of this standalone component.

### DIFF
--- a/src/Sylius/Component/Promotion/composer.json
+++ b/src/Sylius/Component/Promotion/composer.json
@@ -22,7 +22,7 @@
     "require": {
         "php": "^5.5.9|^7.0",
 
-        "doctrine/orm": "^2.4.8,<2.5",
+        "doctrine/orm": "^2.4.8",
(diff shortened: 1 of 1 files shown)


stloyd:

pjedrzejewski:
See https://github.com/Sylius/Sylius/issues/3268, we cannot use 2.5 until this is resolved. :/

stloyd:
@pjedrzejewski But this has __nothing__ to do with __standalone__ usage as this component __don't__ depend on this functionality! Even bundle don't depend on this IIRC...

pamil:
IMO the component itself can be used with 2.5, it does not use any sorting on translatable fields. The only references to ORM are in `CouponGenerator`.

Issue #4544, opened 2016-03-18
Promotions - Fairly broken in 0.18


inssein:
Since my upgrade to Sylius 0.18, promotions have broken down quite a bit.

Errors:
 - Migration of promotions has left with a lot of broken rules (a lot of the actions and the rules are now renamed)
 - ~~The taxon rule doesn't work. When you add a taxon rule, the next page has rendering issues due to an error with Taxon::__toString throwing an exception, and furthermore, the TaxonRuleChecker fails because ```$configuration['taxons']->contains``` is not a function.~~
 - The PromotionEligibilityChecker does not differentiate between pre-qualified promotions and checks when in the cart (one should throw flash messages, the other should not).

The first error is reproducible on the demo, and presumable the second is too, but I cannot test because you can no longer edit any promotion with older actions or rules.

manact:
Don't know if I had to report it here but, in 0.18:
- Promotion based on user group report the following validation error "Client group must be numeric"

The select tag is correctly formed.
Label type:
`<option selected="selected" value="customer_group">User Group</option>`
Label groupe_clients:
`<option selected="selected" value="1">Revendeur</option>`

`<span class="help-block form-error"> Cette valeur doit être de type numeric. </span>`

manact:
I don't know if I have to open another issue but:
When trying to validate a promotion based on user group the validation error is "Value must be numeric":

`<span class="help-block form-error"> Cette valeur doit être de type numeric. </span>`

PR #4651, opened 2016-03-31
Fix TaxonsToCodesTransformer

inssein:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | https://github.com/Sylius/Sylius/issues/4544
| License         | MIT

Fixed the TaxonsToCodesTransformer by adjusting the array check to allow for ArrayCollection.


michalmarcinkowski:
:+1: could you fix the failing spec?

src/Sylius/Bundle/CoreBundle/Form/DataTransformer/TaxonsToCodesTransformer.php
```diff
@@ -61,20 +61,25 @@ public function reverseTransform($value)
             throw new UnexpectedTypeException($value, Collection::class);
         }
 
-        if ($value->isEmpty() || null === $value->get('taxons')) {
+        $taxons = $value->get('taxons');
+
+        // ensure not empty / null
+        if ($value->isEmpty() || null === $taxons) {
```
Zales0123: Shouldn't be ``$taxons->isEmpty()``?

src/Sylius/Bundle/CoreBundle/Form/DataTransformer/TaxonsToCodesTransformer.php
```diff
@@ -61,20 +61,25 @@ public function reverseTransform($value)
...
             return [];
         }
 
-        if (!is_array($value->get('taxons'))) {
-            throw new \InvalidArgumentException('"taxons" element of collection should be an array.');
+        // ensure taxons is traversable
```
Zales0123: I think it's redundant comment.

src/Sylius/Bundle/CoreBundle/Form/DataTransformer/TaxonsToCodesTransformer.php
```diff
@@ -61,20 +61,25 @@ public function reverseTransform($value)
...
+        if (!(is_array($taxons) || $taxons instanceof \Traversable)) {
+            throw new \InvalidArgumentException('"taxons" element of collection should be Traversable');
         }
 
-        $taxons = [];
+        // convert to array of taxon codes
```
Zales0123: So do it.

Zales0123:
But in general :+1:

src/Sylius/Bundle/CoreBundle/Form/DataTransformer/TaxonsToCodesTransformer.php
michalmarcinkowski: We don't know if `$taxons` is `ArrayCollection` instance. There is no need to check for empty.

Zales0123: Ok, but then it should be checked before ``$taxons = $value->get('taxons');``.

michalmarcinkowski: I would say that we don't need this check at all since we're only interested if `$value->get('taxons')` is not null.

inssein:
Fixed based on the comments :+1:

inssein:
ping @michalmarcinkowski - anything else before this can get merged?

michalmarcinkowski:
Please [squash your commits](http://docs.sylius.org/en/latest/contributing/code/patches.html#rework-your-patch) and it will be ready to merge :)

inssein:
Reopened in https://github.com/Sylius/Sylius/pull/4708

PR #4653, opened 2016-03-31
[Composer] Add missing dependency + alphabetically sorting

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,7 @@
         "friendsofsymfony/oauth-server-bundle": "~1.4",
         "friendsofsymfony/rest-bundle": "~1.5",
         "fzaninotto/faker": "~1.2",
+        "gedmo/doctrine-extensions": "^2.4.12",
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
Thanks Mateusz!

composer.json
```diff
@@ -52,9 +53,9 @@
         "payum/omnipay-bridge": "~1.0",
         "payum/payum": "~1.0",
         "payum/payum-bundle": "~1.0",
+        "polishsymfonycommunity/symfony-mocker-container": "^1.0",
```
michalmarcinkowski: This dependency is also missing in `CoreBundle`.

Btw. it should be only required in dev, but it's a separate topic.

PR #4510, opened 2016-03-15
[WebBundle] Add missing field availableUntil

Niiko:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT
| Doc PR          | -

The field was missing.

In order to be more consistent, i rework just a little the form to match the masterVariant one.

Also, i added form_rest() at the end of form to prevent any missing fields (in usage, we will see them quickly)

PR #4656, opened 2016-03-31
[Admin] Don't render whole form by default on create/update page

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

As ``{{ form_end(form) }}`` has been added at the end of every form, it was impossible to render only part of form's fields. I've changed it into ``{{ form_end(form, {'render_rest': false}) }}`` and added {{ form_row(form._token) }} (as token is always needed) to prevent such a inconvenience.

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Country/_form.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Country/_form.html.twig
@@ -3,4 +3,3 @@
     {{ form_row(form.code) }}
     {{ form_row(form.enabled) }}
 </div>
-{{ form_rest(form) }}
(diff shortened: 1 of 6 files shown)


Niiko:
I ran into the exact opposite position as explained in #4510 

I think it's a bad idea to not render the whole form and can have side effects.

Is that not manageable with an override of form in your use case ? Maybe i miss something here.

pjedrzejewski:
@Zales0123 Thanks!

@Niiko I plan this to be temporary, so that we have full control over what is exposed in new admin. We will expose more and more functionality as we implement it. :)

Niiko:
@pjedrzejewski Okay got it

PR #4669, opened 2016-04-01
New Sylius default theme

adamelso:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

I think it looks more playful than the current default

screen shot 2016-04-01 at 13 10 30
screen shot 2016-04-01 at 13 10 38


pamil:
Looks great, Comic Sans makes checkout so readable!

lchrusciel:
We need to have this one merged ASAP. Finally working with Sylius won't be so painful

Zales0123:
What's more, blue background made it looks much more professional! That's what we need, to make Sylius respectable e-commerce. Good job! :+1:

pbrun-mobizel:
49ba358484e7ba5e4de5286da087f64f_large

liverbool:
Why I missed this one cool! 👻

nakashu:
The build is ok, when would this be merged?

pjedrzejewski:
@nakashu Sorry, I am still considering introducing Polymer and AngularJS to new frontend, so I will need to work closely with @adamelso on adding this together with ``jQuery('head').attachJsFrameworkOfTheMonth()`` plugin.

adamelso:
We also want to encourage users to create their own themes, and I fear this theme being as awesome as it is, users may not want to change it for anything else.

I'm going to keep it open for the time being as I noticed a few issues with the docs that need adding or clarification, and so I can use this as a demo to address them.

A few that I remember are:
  * How to add the theme to the database, so it can be assigned to a channel.
  * How to include assets that are [being extended on or overriding the parent assets](https://github.com/Sylius/Sylius/pull/4669/files#diff-5f169bc2bf9cad3d991651a58650285dR21)
  * How to use the Twig `embed` tag to reduce code duplication when overriding templates.
  * How `sylius:install:assets` and `sylius:theme:assets:install` should be merged or work independently.
  * The docs [mentions](https://github.com/Sylius/Sylius/blob/master/docs/bundles/SyliusThemeBundle/your_first_theme.rst#theme-structure) a `theme.json` but this should actually be `composer.json`(?)

pamil:
@adamelso thanks for feedback! Documentation is incoming just after I delete some unnecessary code form ThemeBundle like  `FailsafeRepository` and that whole synchronizing feature, that only causes confusion and greatly increases complexity of already complex bundle :)

pjedrzejewski:
:)

PR #4670, opened 2016-04-01
[Metadata] Fix modifying metadata by serializing it early

### LINKED ISSUE #4472: [Metadata] Update form is broken
Update won't overwrite the values, it doesn't seem to save them.

I've tried digging into the ResourceController and after submitting the form the Metadata object doesn't have the updated values, so something about the form?

Sorry, I don't have more time to look at this now.

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #4472
| License         | MIT

Metadata cannot be stored as `object` ORM type, because objects are
compared (===) by reference, not by value, what means that changes
would not be persisted.

The only working way I have found is to store metadata as `text`
type and serialize it on MetadataContainer::setMetadata() method
call. However, after every metadata modification we have to use that
method once again, if not - changes won't be persisted.

### DIFF
--- a/features/legacy/backend/metadata/metadata_customization.feature
+++ b/features/legacy/backend/metadata/metadata_customization.feature
@@ -24,6 +24,16 @@ Feature: Metadata customization
             | Description | <empty>           |
             | Twitter     | <empty>           |
 
+    Scenario: Modifying existing page metadata
+        Given there is the following metadata "ZBIGNIEW-WODECKI":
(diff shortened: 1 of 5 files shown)


pjedrzejewski:
Thank you Kamil! /cc @peteward

PR #4674, opened 2016-04-03
[Bug Fix] Make sure doctrine removes orphan tax rates

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

There is a bug now where you cannot remove existing tax rates from a tax category instance. It will try to set the tax rates category to `null`. We want to remove the tax rate entirely when it's not connected to a tax category anymore. This PR fixes this.

### DIFF
--- a/src/Sylius/Bundle/TaxationBundle/Resources/config/doctrine/model/TaxCategory.orm.xml
+++ b/src/Sylius/Bundle/TaxationBundle/Resources/config/doctrine/model/TaxCategory.orm.xml
@@ -23,7 +23,8 @@
         <field name="name" column="name" type="string" />
         <field name="description" column="description" type="text" nullable="true" />
 
-        <one-to-many field="rates" target-entity="Sylius\Component\Taxation\Model\TaxRateInterface" mapped-by="category">
+        <one-to-many field="rates" target-entity="Sylius\Component\Taxation\Model\TaxRateInterface" mapped-by="category"
+            orphan-removal="true">
             <cascade>
                 <cascade-all/>
             </cascade>


steffenbrem:
@pjedrzejewski ping. I guess this one is easy to merge.

lchrusciel:
Please, be patient. It can be quite hard to merge something during this week. It should be merged at the beginning of the next week.

michalmarcinkowski:
Not sure if it's the best UX to silently remove all tax rates from the deleted tax category. IMO the current behavior should stay for now and later we will implement something like `The following tax rates will also be removed: X, Y, Z. Do you confirm deletion?`.

You should first remove all tax rates from this tax category and then you should be able to delete the tax category without any problem.

steffenbrem:
@michalmarcinkowski Right now when you remove a tax category in the admin it will also remove the tax rates associated with it. I think they have a strong relationship with each other and a tax rate should not exist without being associated with a tax category, at least that is how everything is build right now. With this in mind I have build an API around tax rates and categories where you can update the tax rates of a tax category with `PATCH /tax-categories/1`. When I leave out a tax rate in the payload and send the request I would expect the tax rate to be deleted too. I think if we would leave out the orphan removal, we should allow tax rates to be alive individually, which is not the case at the moment.

michalmarcinkowski:
Not sure if I understand correctly. The current behavior is the following:
```
Given the store has a tax category "Sports gear"
And the store has "European Union Sales Tax" tax rate of 20% for "Sports gear"
When I delete tax category "Sports gear"
Then tax rate "European Union Sales Tax" should no longer exist
```

So it should work fine for your use case. Are you able to write a test reproducing the bug you're experiencing?

steffenbrem:
```
Given the store has a tax category "Sports gear"
And the store has "European Union Sales Tax" tax rate of 20% for "Sports gear"
When I remove tax rate "European Union Sales Tax" from category "Sports gear"
And I save tax category "Sports gear"
Then tax category "Sports gear" should no longer have tax rate "European Union Sales Tax"
``` 

This specific case will fail and throw this error:
```
SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "category_id" violates not-null constraint DETAIL
```

This is simply happening because tax rates can be orphaned now, since the `orphan-removal` is not enabled for a tax rate.

michalmarcinkowski:
```
Given the store has a tax category "Sports gear"
And the store has "European Union Sales Tax" tax rate of 20% for "Sports gear"
When I remove tax rate "European Union Sales Tax" from category "Sports gear"
```
That is fine.
```
And I save tax category "Sports gear"
```
That do you mean by that?

I tried:
```
Given the store has a tax category "Sports gear"
And the store has "European Union Sales Tax" tax rate of 20% for "Sports gear"
When I remove tax rate "European Union Sales Tax" from category "Sports gear"
And I modify tax category "Sports gear"
And I save tax category "Sports gear"
Then tax category "Sports gear" should no longer have tax rate "European Union Sales Tax"
```
And it works fine. I can't reproduce your issue... Are you using the newest master?

steffenbrem:
Well I am not using the admin panel, only using the resources made available to me. I have a case where I save all tax rates embedded inside the tax category. I have my own `POST /tax-categories` endpoint that does that for me. And since my understanding is that a tax rate always belongs to a tax category, it should be removed when it's tax rate is `null`, hence the `orphan-removal` addition.

steffenbrem:
As you can see in this file, tax rates always needs a tax category: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/TaxationBundle/Resources/config/doctrine/model/TaxRate.orm.xml#L29

michalmarcinkowski:
Yeah I think that this is correct, but wanted to verify your issue 😄  Anyway thanks for the explanation and the PR! 👍

steffenbrem:
Great! 👍  Next time I will put some more time into giving a better and more to-the-point explanation.

PR #4677, opened 2016-04-04
[Core] Get last payment's method also from failed payments

aramalipoor:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

Because of this bug customers would see an exception due to `lastPayment->method` was not set (`null`) correctly.

### DIFF
--- a/src/Sylius/Component/Core/OrderProcessing/PaymentProcessor.php
+++ b/src/Sylius/Component/Core/OrderProcessing/PaymentProcessor.php
@@ -39,6 +39,10 @@ class PaymentProcessor implements PaymentProcessorInterface
      */
     public function processOrderPayments(OrderInterface $order)
     {
+        if ($order->getLastPayment(PaymentInterface::STATE_NEW)) {
+            return;
+        }
+
         /** @var $payment PaymentInterface */
         $payment = $this->paymentFactory->createWithAmountAndCurrency($order->getTotal(), $order->getCurrency());
         $this->setPaymentMethodIfNeeded($order, $payment);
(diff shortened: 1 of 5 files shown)


src/Sylius/Component/Core/OrderProcessing/PaymentProcessor.php
```diff
@@ -51,11 +51,22 @@ public function processOrderPayments(OrderInterface $order)
...
-            $payment->setMethod($lastCancelledPayment->getMethod());
+        if ($lastNewPayment) {
+            return;
         }
+
+        $lastCancelledOrFailedPayment = $order->getLastPayment(PaymentInterface::STATE_CANCELLED);
+
+        if (!$lastCancelledOrFailedPayment) {
+            $lastCancelledOrFailedPayment = $order->getLastPayment(PaymentInterface::STATE_FAILED);
+        }
```
Zales0123: How about extracting four lines above into separate method? Then we could use returned variable in following if. Moreover, variable should be named ``$lastPayment`` or sth like this, I suppose, as ``$lastCancelledOrFailedPayment`` smells bad :)

src/Sylius/Component/Core/OrderProcessing/PaymentProcessor.php
```diff
@@ -51,11 +55,28 @@ public function processOrderPayments(OrderInterface $order)
...
+            return;
         }
+
+        $payment->setMethod($lastPayment->getMethod());
+    }
+
+    /**
+     * @param OrderInterface $order
+     *
+     * @return false|PaymentInterface
```
lchrusciel: I think something is wrong here. This method should not return false at all.

aramalipoor:
@michalmarcinkowski ping!

pjedrzejewski:
This method will probably need to go away, but we will see, looks good to me for now, thank you Aram!

PR #4681, opened 2016-04-04
[Twig] Sematic money widget

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | 
| License         | MIT

This widget rendering:
zrzut ekranu 2016-04-04 o 13 48 55
will be changed to this one:
zrzut ekranu 2016-04-04 o 13 49 15

_Update_
Currency extension was extended to provide current currency instead of hardcoded _$_

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig
+++ b/src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig
@@ -27,3 +27,10 @@
         </div>
     </div>
 {%- endblock checkbox_row %}
+
+{% block money_widget -%}
+    <div class="ui labeled input">
+        <div class="ui label">$</div>
+        {{- form_widget(form) -}}
+    </div>
+{%- endblock money_widget %}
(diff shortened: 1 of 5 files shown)


src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig
```diff
@@ -27,3 +27,10 @@
         </div>
     </div>
 {%- endblock checkbox_row %}
+
+{% block money_widget -%}
+    <div class="ui labeled input">
+        <div class="ui label">$</div>
```
pjedrzejewski: We should not hardcode USD here.

lchrusciel: Should I write a custom function or there is a way to receive it from config?

michalmarcinkowski:
Thanks Łukasz! :+1:

pjedrzejewski:
This is wrong - money form type does not convert the amount to the base currency. So this form type should always display the symbol of base currency. Otherwise if you switch currency in the frontend, you will see the change in the backend and this can lead to some interesting mistakes. Please open a PR which uses base currency @lchrusciel. Thanks!

PR #4688, opened 2016-04-04
[Migration] Auto-generate code in db field while running migration

psyray:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

Automatically add incremental code based on the db name.
eg: for sylius_shipping_category, generate SC1, SC2, SC3 .....

This is to prevent migration fails due to the UNIQUE INDEX statement.
When the migration fails you have to manually add code to db for each line, comment SQL query already ran and run migration script again... :(

Tested and fully working

Enjoy

### DIFF
--- a/app/migrations/Version20151203134947.php
+++ b/app/migrations/Version20151203134947.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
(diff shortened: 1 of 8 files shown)


pjedrzejewski:
Thank you Raynald!

PR #4702, opened 2016-04-06
[Resource] Fixed proviously spoiled transformer

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Ok, 7th and 8th of October were bad days for the Republic :smile: These days I've fixed ``ObjectToIdentifierTransformer`` in https://github.com/Sylius/Sylius/pull/3406... but *de facto* I ruined it :( It has been written good, as in ``reverseTransform`` method value is object from form and in ``transform`` it's identifier - not otherwise. The reason why it was working is that if we pass object to ``findOneBy`` method instead of id, doctrine would return object based on passed object id :boom: Now it should be good, I really hope :)

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Form/DataTransformer/ObjectToIdentifierTransformer.php
+++ b/src/Sylius/Bundle/ResourceBundle/Form/DataTransformer/ObjectToIdentifierTransformer.php
@@ -48,26 +48,6 @@ class ObjectToIdentifierTransformer implements DataTransformerInterface
      */
     public function transform($value)
     {
-        if (empty($value)) {
-            return '';
-        }
-
-        $class = $this->repository->getClassName();
-
-        if (!$value instanceof $class) {
-            throw new UnexpectedTypeException($value, $class);
-        }
-
-        $accessor = PropertyAccess::createPropertyAccessor();
-
-        return $accessor->getValue($value, $this->identifier);
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function reverseTransform($value)
-    {
         if (!$value) {
             return null;
         }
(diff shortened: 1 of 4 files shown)


src/Sylius/Bundle/ShippingBundle/Form/Type/ShippingMethodChoiceType.php
```diff
@@ -90,10 +93,19 @@ public function configureOptions(OptionsResolver $resolver)
             return new ObjectChoiceList($methods, null, [], null, 'id');
         };
 
+        $dataClass = function (Options $options) {
+            if ($options['multiple']) {
+                return null;
+            }
+
+            return ShippingMethod::class;
```
pjedrzejewski: This is definitely wrong, we cannot hardcode ShippingModel class here.

src/Sylius/Bundle/ResourceBundle/Form/DataTransformer/ObjectToIdentifierTransformer.php
```diff
@@ -48,39 +48,39 @@ public function __construct(ObjectRepository $repository, $identifier = 'id')
...
-            return '';
+        if (!$value) {
+            return null;
         }
 
-        $class = $this->repository->getClassName();
-
-        if (!$value instanceof $class) {
-            throw new UnexpectedTypeException($value, $class);
+        if (null === $entity = $this->repository->findOneBy([$this->identifier => $value])) {
```
pamil: Extracting the assignment before the `if` would make it clearer.

michalmarcinkowski:
Thanks Mateusz!

PR #4704, opened 2016-04-06
[Semantic Ui] [RFC] Fix problematic dropdown generation

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/private/js/app.js
+++ b/src/Sylius/Bundle/UiBundle/Resources/private/js/app.js
@@ -31,7 +31,6 @@
     ;
 
     $('.ui.checkbox').checkbox();
-    $('select').dropdown();
 
     $('.form button').on('click', function() {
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
Could you elaborate on how this fixes the problem?

lchrusciel:
I just removed semantic wrapper on select tag. I looks similar to other fields, but it is not as pretty as it can be. But it is the simplest solution.
Pros:
* Simple
* Does not depend on any variables 
* Does not require any additional code
* Super easy to maintain
Cons:
* Could be prettier 
* Could be problematic if we decide to implement Multiple Search Selection

pjedrzejewski:
:tada: Thanks Łukasz!

PR #4709, opened 2016-04-06
[Behat] Remove contexts specs

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

As we decided not to write specifications for pages and contexts, it would be nice to delete them all at once than partially in multiple PRs when noticed, that they are failing.


pjedrzejewski:
^_^

Issue #4710, opened 2016-04-06
[CartBundle] Repository is not mapped (and thus not implementing CartRepositoryInterface)


inssein:
```sylius.repository.cart``` is a default resource bundle entity repository, and not Sylius\Bundle\CartBundle\Doctrine\ORM\CartRepository

pjedrzejewski:
Thank you Hussein for reporting!

inssein:
This is still not fixed. We set sylius.repository.cart.class, but never reference it.

inssein:
@pamil I thought the fix was to add ```repository: %sylius.repository.cart.class%``` in CoreBundle/Resources/config/app/sylius.yml``` for the sylius_cart declaration, but ```sylius.repository.cart.class``` is not yet defined.

PR #4711, opened 2016-04-06
[CartBundle] Map CartRepository

inssein:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/4710
| License         | MIT


pamil:
Thanks for contribution! Unfortunately, we use a quite different convention for modifying repository class in order to be able to set it per database driver, I've implemented it in #4721.

inssein:
@pamil makes sense - saw this in another bundle, but didn't do it because the odm.xml files aren't implemented anyways.

PR #4731, opened 2016-04-09
[CoreBundle] Sitemap - Fix priority setter

inssein:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Fixed the setter so that it maches the exception. Previously a priority of 1.0 would not be allowed.


pamil:
Shouldn't it [allow both `0.0` and `1.0`](http://www.sitemaps.org/protocol.html#prioritydef)?

inssein:
@pamil good catch - fixed.

pjedrzejewski:
Thank you Hussein!

PR #4732, opened 2016-04-10
[RFC][Behat] Introduce semantic ui driver

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

I was thinking a lot about this problems related to semantic ui and testing it with selenium driver. Finally i decided to implement semantic ui driver it will keep our page object implementation nice and clean without this sick checking which driver is currently running. There is another approach to solve this problem. Setting environment flag in body tag and then simple catch this using JS to switch off those fireworks from semantic but since it will be set in html it may cause problems for example in production environment.
To use this driver simple tag your scenario with `@mink:semantic_ui`.

### DIFF
--- a/etc/behat/profiles/default.yml
+++ b/etc/behat/profiles/default.yml
@@ -36,6 +36,18 @@ default:
                 firefox:
                     selenium2:
                         browser: firefox
+                semantic_ui:
+                    semantic_ui:
+                        browser: chrome
+                        capabilities:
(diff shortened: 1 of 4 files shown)


pjedrzejewski:
Closing in favor of #4704.

PR #4739, opened 2016-04-11
[Ui][Critical] Sort messages keys

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

I also fixed indent (I think it should be 4 spaces, here and in some other files). We should definitely keep this order, as it's not hard but makes file much more readable ;)


pjedrzejewski:
Thanks Mateusz!

PR #4746, opened 2016-04-11
[Order] Fix bug in Order model

### LINKED ISSUE #4727: [Order] Sylius\Component\Order\Model\Order::removeItem() calculates itemsTotal incorrectly
The removeItem function calls `$item->setOrder(null)`, which will call `removeItem($this)` again, if the item already has an order. This causes the item price to be subtracted twice from the cart total.
```
public function removeItem(OrderItemInterface $item)
    {
        if ($this->hasItem($item)) {
            $item->setOrder(null);
            $this->items->removeElement($item);
            $this->itemsTotal -= $item->getTotal();
            $this->recalculateTotal();
        }
    }
```
https://github.com/Sylius/Order/blob/master/Model/Order.php#L209

```
    public function setOrder(OrderInterface $order = null)
    {
        $currentOrder = $this->getOrder();
        if ($currentOrder === $order) {
            return;
        }
        $this->order = null;
        if (null !== $currentOrder) {
            $currentOrder->removeItem($this);
        }
        if (null === $order) {
            return;
        }
        $this->order = $order;
        if (!$order->hasItem($this)) {
            $order->addItem($this);
        }
    }
```
https://github.com/Sylius/Order/blob/master/Model/OrderItem.php#L116

The tests do pass. I believe this is because in the test cases `null !== $currentOrder` is never true and therefore `removeItem()` is never called twice.

sblaut:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #4727
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Order/Model/Order.php
+++ b/src/Sylius/Component/Order/Model/Order.php
@@ -206,11 +206,10 @@ class Order implements OrderInterface
     public function removeItem(OrderItemInterface $item)
     {
         if ($this->hasItem($item)) {
-            $item->setOrder(null);
             $this->items->removeElement($item);
             $this->itemsTotal -= $item->getTotal();
-
             $this->recalculateTotal();
+            $item->setOrder(null);
         }
     }
 


pjedrzejewski:
Nice! Good to go when green on Travis, thank you!

pjedrzejewski:
Thank you @sblaut for reporting and fixing. :+1:

PR #4747, opened 2016-04-11
[Bug Fix] Added customer validator initializer to set canonical email

### LINKED ISSUE #4733: Customer email canonical validation case insensitivity
The email unique validation on the `Customer` model is currently checking against the `email` field, but it should check agains the `emailCanonical` field. This causes bugs when using PostgreSQL which does not have a case-insensitive collation. Is this something that should be fixed on the Sylius side? In my opinion I think it does, but I let that decision over to you guys :+1:  

We can't simply validate against the `emailCanonical` field since it is being set on `prePersist`. So I am not sure what a good solution is to handle this case.
```xml
<class name="Sylius\Component\User\Model\Customer">
        <constraint name="Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity">
            <option name="message">sylius.customer.email.unique</option>
            <option name="fields">email</option>
        </constraint>
<!-- ... -->
```

Maybe something like http://symfony.com/doc/current/reference/dic_tags.html#validator-initializer could be used to achieve this:

```xml
<constraint name="Symfony\Bridge\Doctrine\Validator\Constraints\UniqueEntity">
    <option name="message">sylius.customer.email.unique</option>
    <option name="fields">emailCanonical</option> <!-- This is what you want -->
    <option name="errorPath">email</option>
</constraint>
```

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #4733 
| License         | MIT

Added customer validator initializer that sets email canonical on initialisation. Customer object is now also validating against `emailCanonical` and uses `email` as it's error path.

### DIFF
--- a/src/Sylius/Bundle/UserBundle/Resources/config/services.xml
+++ b/src/Sylius/Bundle/UserBundle/Resources/config/services.xml
@@ -59,6 +59,7 @@
         <parameter key="sylius.controller.user_security.class">Sylius\Bundle\UserBundle\Controller\SecurityController</parameter>
 
         <parameter key="sylius.user.validator.unique_registered_user.class">Sylius\Bundle\UserBundle\Validator\Constraints\RegisteredUserValidator</parameter>
+        <parameter key="sylius.user.validator.customer_initializer.class">Sylius\Bundle\UserBundle\Validator\Initializer\CustomerInitializer</parameter>
     </parameters>
 
     <services>
@@ -207,6 +208,11 @@
         <service id="validator.unique.registered_user" class="%sylius.user.validator.unique_registered_user.class%">
             <argument type="service" id="sylius.repository.customer" />
             <tag name="validator.constraint_validator" alias="registered_user_validator" />
+        </service>
+
+        <service id="sylius.user.validator.customer_initializer" class="%sylius.user.validator.customer_initializer.class%">
+            <argument type="service" id="sylius.user.canonicalizer" />
+            <tag name="validator.initializer" />
         </service>
     </services>
 
--- a/src/Sylius/Bundle/UserBundle/spec/Validator/Initializer/CustomerInitializerSpec.php
+++ b/src/Sylius/Bundle/UserBundle/spec/Validator/Initializer/CustomerInitializerSpec.php
@@ -1,0 +1,53 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace spec\Sylius\Bundle\UserBundle\Validator\Initializer;
+
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Sylius\Component\User\Canonicalizer\CanonicalizerInterface;
+use Sylius\Component\User\Model\CustomerInterface;
+use Symfony\Component\Validator\ObjectInitializerInterface;
+
+/**
+ * @author Steffen Brem <steffenbrem@gmail.com>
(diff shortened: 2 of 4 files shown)


src/Sylius/Bundle/UserBundle/Resources/config/services.xml
```diff
@@ -208,6 +209,11 @@
             <argument type="service" id="sylius.repository.customer" />
             <tag name="validator.constraint_validator" alias="registered_user_validator" />
         </service>
+
+        <service id="sylius.user.validator.customer_initializer" class="Sylius\Bundle\UserBundle\Validator\Initializer\CustomerInitializer">
```
pamil: Should use the class set above :)

steffenbrem: haha, wow, that is stupid..

src/Sylius/Bundle/UserBundle/spec/Validator/Initializer/CustomerInitializerSpec.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        $this->shouldImplement(ObjectInitializerInterface::class);
+    }
+
+    function it_sets_canonical_email_when_initializing_customer($canonicalizer, CustomerInterface $customer)
+    {
+        $customer->getEmail()->willReturn('sTeFfEn@gMaiL.CoM');
+        $canonicalizer->canonicalize('sTeFfEn@gMaiL.CoM')
+            ->shouldBeCalled()
+            ->wilLReturn('steffen@gmail.com')
```
stloyd: Typo: `wilLReturn` vs `willReturn`

steffenbrem: Good catch :+1:

pjedrzejewski:
Great, could you please squash it to a single commit? Thank you Steffen!

stloyd:
@pjedrzejewski You can do it too now: https://github.com/blog/2141-squash-your-commits

pjedrzejewski:
@stloyd For some reason it does not work, I get that Travis check is required and we should not disable that.

src/Sylius/Bundle/UserBundle/spec/Validator/Initializer/CustomerInitializerSpec.php
```diff
@@ -0,0 +1,56 @@
...
+    function it_implements_symfony_validator_initializer_interface()
+    {
+        $this->shouldImplement(ObjectInitializerInterface::class);
+    }
+
+    function it_sets_canonical_email_when_initializing_customer($canonicalizer, CustomerInterface $customer)
+    {
+        $customer->getEmail()->willReturn('sTeFfEn@gMaiL.CoM');
+        $canonicalizer->canonicalize('sTeFfEn@gMaiL.CoM')
+            ->shouldBeCalled()
```
michalmarcinkowski: `shouldBeCalled` should be removed, `willReturn` is enough.

steffenbrem:
You have to click on merge and then you will be able to select if you want to squash it.

pjedrzejewski:
@steffenbrem The problem is that I have configured Travis as required check, and this prevents me from squashing it via UI. :/

steffenbrem:
That is strange, I have two other projects with the same setup, also a CI check and enabled both merge commit + squash merging. Then I could simply click on `Merge pull request` and then it lets me choose which merge strategy to use.

G

But np, I will squash it :+1:

pjedrzejewski:
Thanks Steffen!

PR #4752, opened 2016-04-11
[RFC] [Ui] Cancel button

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

It was annoying, that it wasn't possible to just go back on create/update form. It is not the prettiest solution, but I wasn't able to find a better one. Also I have not add second parameter as a fallback, because I didn't want to add parameter to this templates. But on the other hand I don't have any good idea, how to solve this. 

Before:
zrzut ekranu 2016-04-12 o 08 19 02

After:
zrzut ekranu 2016-04-12 o 08 22 05


pjedrzejewski:
Nice, thanks Łukasz!

pjedrzejewski:
Perhaps later we can extract this to separate macro/template to avoid duplication. :+1:

lchrusciel:
👍

PR #4754, opened 2016-04-11
Doctrine option value translation migration fix

### LINKED ISSUE #4753: Option value translation migration does not copy data
This PR #4237 has bad migration file that delete all values by dropping the **value** column in the **product_option_value** table instead of copying it to the newly created **product_option_value_translation** table in the **presentation** column.

A good migration is like [this one] (https://github.com/Sylius/Sylius/blob/master/app/migrations/Version20141203091118.php)

PR coming soon

psyray:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #4753
| License         | MIT

Tested and working in my project (up & down)

### DIFF
--- a/app/migrations/Version20151203134947.php
+++ b/app/migrations/Version20151203134947.php
@@ -1,5 +1,14 @@
 <?php
 
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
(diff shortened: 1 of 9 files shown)


psyray:
some branches errors in this PR, I close it

PR #4758, opened 2016-04-12
[Behat] [DX] Table row counter on index page

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Page/Admin/Crud/IndexPage.php
+++ b/src/Sylius/Behat/Page/Admin/Crud/IndexPage.php
@@ -69,6 +69,14 @@ class IndexPage extends SymfonyPage implements IndexPageInterface
     }
 
     /**
+     * @return int
+     */
+    public function countItems()
+    {
+        return $this->getTableAccessor()->countTableBodyRows($this->getElement('table'));
+    }
+
+    /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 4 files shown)


src/Sylius/Behat/Page/Admin/Crud/IndexPage.php
```diff
@@ -76,6 +76,14 @@ public function isResourceOnPage(array $parameters)
     }
 
     /**
+     * @return int
+     */
+    public function countTableRecords()
```
pjedrzejewski: We should not assume that it is a table. Index page contains a collection (paginated or not) of items, so we should have something like ``countItems``?

pjedrzejewski: I meant that on the interface, we of course can use table accessor in our index page implementation, but the interface should not assume that it is a table.

lchrusciel: 👍

michalmarcinkowski:
Looks good to me :)

pjedrzejewski:
Thank you Łukasz!

PR #4760, opened 2016-04-12
[Taxation] Tax calculation minor enhancements

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Few minor enhancements to #4406:
* Removed check for `0` quantity because order item with this quantity shouldn't ever exist
* Reveal invalid order items (with quantity `0`) passed to order items taxes applicator
* Removed PHPUnit dependency from specs
* Replaced manual assertions with Webmozart library

### DIFF
--- a/src/Sylius/Component/Core/Taxation/Applicator/OrderItemUnitsTaxesApplicator.php
+++ b/src/Sylius/Component/Core/Taxation/Applicator/OrderItemUnitsTaxesApplicator.php
@@ -57,13 +57,7 @@ class OrderItemUnitsTaxesApplicator implements OrderTaxesApplicatorInterface
     public function apply(OrderInterface $order, ZoneInterface $zone)
     {
         foreach ($order->getItems() as $item) {
-            $quantity = $item->getQuantity();
-            if (0 === $quantity) {
-                continue;
-            }
-
             $taxRate = $this->taxRateResolver->resolve($item->getVariant(), ['zone' => $zone]);
-
             if (null === $taxRate) {
                 continue;
             }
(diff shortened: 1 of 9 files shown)


pjedrzejewski:
Nice! :+1:

src/Sylius/Component/Core/Taxation/Applicator/OrderItemUnitsTaxesApplicator.php
```diff
@@ -57,13 +57,7 @@ public function __construct(CalculatorInterface $calculator, AdjustmentFactoryIn
     public function apply(OrderInterface $order, ZoneInterface $zone)
     {
         foreach ($order->getItems() as $item) {
-            $quantity = $item->getQuantity();
-            if (0 === $quantity) {
-                continue;
-            }
```
mmckelviereiss: I completely agree with removing this pointless check. I had tried to keep it consistent with the other tax applicator `Sylius\Component\Core\Taxation\Applicator\OrderItemsTaxesApplicator`. I suggest you update that one too. :)

michalmarcinkowski: Actually in `Sylius\Component\Core\Taxation\Applicator\OrderItemsTaxesApplicator` there must be a check because we distribute the tax on units, but I changed the implementation to throw an exception when order item with `0` quantity is passed in order to reveal the possible bugs :)

pjedrzejewski:
:+1:

PR #4762, opened 2016-04-12
[mailer] - fix mailer configuration to allow adding new templates

nakashu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

- configuration for templates uses keys now, and new templates can be properly added. Before the fix new templates from the configuration would be resolved only with number index instead of actual key.
- choiceform deprecation fixed.


pjedrzejewski:
:+1:

pjedrzejewski:
Thank you @nakashu!

Issue #4461, opened 2016-03-11
Remove softdelete filters toggling in the repositories


michalmarcinkowski:
We have almost finished removing softdelete from all entities, but the filters are still being disabled in the repositories (e.g. [here](https://github.com/michalmarcinkowski/Sylius/blob/4cb6495742ac4b05bea3d312a374d19203232747/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/PaymentRepository.php#L28-L28)). Disabling this filter doesn't make sense anymore since the entities are not softdeleteable, so we should remove that from repositories.

pjedrzejewski:
As soon as last soft-delete is removed we should disable this filter in general and we will get exception whenever it is disabled/enabled.

michalmarcinkowski:
Sounds like a plan :)

michalmarcinkowski:
Fixed in #4767.

Issue #4140, opened 2016-02-13
[RFC] Deleting Resources


michalmarcinkowski:
We need to resolve issue with removing things.

### Problems:

 - We want to be able to remove stuff from admin
 - We want to keep order, product history intact
 - Deleting resources is sensitive and in many cases should be considered individually, ‘generic’ softdelete is not going to help here
 - Softdeleteable Doctrine Behavior is causing a lot of problems since we still need to manually handle all relations that can be softdeleted. That makes this ‘generic’ approach a source of many bugs and exceptions in admin

### Solutions:

 - Removing Softdeleteable Behavior
 - Ensure proper cascading for all entities
 - Introducing concept of Archivization:
  - You can delete something completely if it is not used anywhere else
  - You can delete something completely if it is used somewhere else, but it won’t break order history, we just need to notify user that it will have side effects (deleting TaxCategory will make all products not taxed)
  - You cannot delete something completely if it breaks order history, for example, shipping method which is associated with previous orders, but you can archive it so it does not show up in the backend and cannot be used in new stuff

##### Stage 1: Remove softdelete from
* [ ] Order
* [ ] Coupon
* [ ] Promotion
* [ ] Payment
* [ ] Product
* [ ] Variant
* [ ] Taxon (already covered in #3773)
* [ ] User and Customer (already covered in #4009)

##### Stage 2: Introduce Archivization concept on
* [ ] Shipping Method
* [ ] Payment Method
* [ ] Channel
* … to be continued

##### Related issues:
 - https://github.com/Sylius/Sylius/issues/3654
 - https://github.com/Sylius/Sylius/issues/3523
 - https://github.com/Sylius/Sylius/issues/3514
 - https://github.com/Sylius/Sylius/issues/3349
 - https://github.com/Sylius/Sylius/issues/3144
 - https://github.com/Sylius/Sylius/issues/3065
 - https://github.com/Sylius/Sylius/issues/2950
 - https://github.com/Sylius/Sylius/issues/2778
 - https://github.com/Sylius/Sylius/issues/2774
 - https://github.com/Sylius/Sylius/issues/2537
 - https://github.com/Sylius/Sylius/issues/2423
 - https://github.com/Sylius/Sylius/issues/1799
 - https://github.com/Sylius/Sylius/issues/1326
 - https://github.com/Sylius/Sylius/issues/1049
 - https://github.com/Sylius/Sylius/issues/1040
 - https://github.com/Sylius/Sylius/issues/950

[...]

PR #4767, opened 2016-04-12
Remove the remaining soft deleteable code

### LINKED ISSUE #4761: What is the status of the SoftDeleteableFilter?
I get some errors on the current master branch due to calls to the softdeleteable filter.

What is the status of the usage of this filter? I would like to clean up the codebase to avoid errors, but I'm not sure what components still are allowed to have a `deletedAt` field.

coudenysj:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | ?
| Deprecations?   | yes
| Related tickets | fixes #4761 and #4461
| License         | MIT

## TODO
- [ ] Cleanup Behat tests
- [x] Check if [DoctrineSluggablePass](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/DependencyInjection/Compiler/DoctrineSluggablePass.php) can be removed


src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
```diff
@@ -46,10 +46,8 @@ public function indexByCustomerAction(Request $request, $id)
 
         // Fetch and cache deleted orders
         $entityManager = $this->container->get('doctrine.orm.entity_manager');
```
michalmarcinkowski: `$entityManager` is not needed anymore, as well as comment above.

coudenysj:
Looks like TravisCI is having a lot of difficulties?

Zales0123:
@coudenysj yes, sadly it seems to fail with latest chrome version. All should be fixed after merging https://github.com/Sylius/Sylius/pull/4786.

coudenysj:
As far as I can tell, the soft deleteable magic is not tested in any Behat feature file.

We have to wait on #4786 to see if anything breaks, I guess.

michalmarcinkowski:
@coudenysj #4786 was merged, please rebase.

coudenysj:
@michalmarcinkowski Done!

michalmarcinkowski:
Great work! Thanks Jachim! :+1:

pjedrzejewski:
Awesome work Jachim, thank you so much!

rvanlaak:
Great! Thanks, the `softdeleteable` also resulted into errors when the `promotion` bundle was used stand-alone.

pjedrzejewski:
@rvanlaak I am not sure what you mean? Soft-deleteable behavior is gone from Sylius.

rvanlaak:
@pjedrzejewski yes I know, but I'm still on `0.17` and the change was merged after that release. So basically my question is whether, until another version is released, I can just enable it before and disable the filter afterwards?

Oh wait, I see `0.18` is released yesterday. Think I'm going to upgrade right away. Major breaks on `promotion`?

rvanlaak:
The upgrade to `0.18` wasn't easy, got a missing factory error while compiling the container. Probably has something to do with a missing dependency if `promotion` is used stand-alone.

michalmarcinkowski:
@rvanlaak did you resolve this issue?

rvanlaak:
@michalmarcinkowski no we postponed the upgrade for now, but will try again soon. Did hope that the documentation on the upgrade can be improved, given `0.18` has that many changes related to the `promotion` component: https://github.com/Sylius/Sylius/blob/master/CHANGELOG.md

rvanlaak:
We updated from `0.17` to `0.19` today, basically rewriting the `softDelete` functionality took us quite some time. We are fine with removing it from the `Promotion` entity, but we ourselves actually have a use-case for it on the `Coupon` entity. When customers bought a coupon from us but don't want to use it anymore, we should be able to "delete" it.

@michalmarcinkowski are there any plans to simulate the `softDelete` behaviour for the `Coupon` entity, but without using the `DoctrineExtensions`? The only thing about the decision proces I can find on it is you saying that is should be removed in https://github.com/Sylius/Sylius/issues/4761#issuecomment-208777651, but I'm not exactly sure why :)

coudenysj:
@rvanlaak I think a solution could be to use the ToggleableInterface.

michalmarcinkowski:
@rvanlaak if coupon is not used you should be able to delete it, if it was already used it should be archived. See https://github.com/Sylius/Sylius/issues/4140 and our idea for archiving.

michalmarcinkowski:
Btw. for coupons you can simply set "expiresAt" to make it archived :)

rvanlaak:
Ah thanks for referencing #4140, I agree with the problem statement given there 👍 For now I was able to reuse the `deletedAt` field that already exists by overriding the `Coupon` class used by the `promotion` component:

    /**
     * @ORM\Table(name="sylius_promotion_coupon")
     * @ORM\Entity
     *
     * @Gedmo\SoftDeleteable(fieldName="deletedAt")
     */
    class Coupon extends BaseCoupon
    {
        use SoftDeleteableEntity;
    }

After the first alpha is released we will migrate it to `expiresAt`. Wouldn't it be better to "improve" the [`ToggleableTrait`](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Resource/Model/ToggleableTrait.php) and use an `expiresAt` datetime instead of the boolean that is used now? The bool doesn't tell us anything about the coupon's history (read: archive date).

michalmarcinkowski:
Not sure what do you mean? `PromotionCoupon` is not toggleable.

rvanlaak:
Yes that's correct, so I'd like to propose to make [`PromotionCoupon`](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Promotion/Model/PromotionCoupon.php) toggleable and would like to know if you agree on that too. 

The second thing I'd like to mention (aside of making it toggleable itself), is that it would be great if `ToggleableTrait::$enabled` could become a `DateTime` to provide more info about the lifecycle of the entity. In other words: so we can see when the coupon was archived. That would also make archiving in the future possible:

```php
/**
 * @return Boolean
 */
public function isEnabled()
{
	if (null !== $this->enabled) {
		return $this->enabled <= (new \DateTime());
	}

	return false;
}
```

michalmarcinkowski:
That seems redundant. Why not simply use `expiresAt` for this? The logic is already implemented and working.

rvanlaak:
Okay thanks 👍 For our use-case the archiving has a different use-case than the `expiresAt` field would have so I'll stick with my custom implementation for now.

PR #4768, opened 2016-04-12
[Core][Promotion] Distribute promotions on units

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets |
| License         | MIT

To fulfill new order units approach, we should distribute promotions into units not only in ``item`` promotions, but also in those applied on ``Order`` level. Therefore, in ``fixed_discount`` and ``percentage_discount`` promotions, ``order_promotion`` adjustments are added on order units (while applying percentage discount distribution is equal, while applying fixed discount - promotion amount is distributed on order's items proportionally and then equally on units).

### DIFF
--- a/features/legacy/frontend/cart_promotions_percentage.feature
+++ b/features/legacy/frontend/cart_promotions_percentage.feature
@@ -8,12 +8,12 @@ Feature: Checkout percentage discount promotions
         Given store has default configuration
         And the following promotions exist:
             | code | name    | description                                   |
-            | PR1  | 5 items | 25% Discount for orders with at least 5 items |
+            | PR1  | 5 units | 25% Discount for orders with at least 5 units |
             | PR2  | 300 EUR | 10% Discount for orders over 300 EUR          |
-        And promotion "5 items" has following rules defined:
+        And promotion "5 units" has following rules defined:
             | type          | configuration |
             | Cart quantity | Count: 5      |
-        And promotion "5 items" has following actions defined:
+        And promotion "5 units" has following actions defined:
             | type                      | configuration  |
             | Order percentage discount | percentage: 25 |
         And promotion "300 EUR" has following rules defined:
@@ -69,5 +69,5 @@ Feature: Checkout percentage discount promotions
         And I added product "Buzz" to cart, with quantity "1"
         When I add product "Woody" to cart, with quantity "3"
         Then I should still be on the cart summary page
-        And "Promotion total: -€586.25" should appear on the page
-        And "Grand total: €1,088.75" should appear on the page
+        And "Promotion total: -€544.38" should appear on the page
+        And "Grand total: €1,130.62" should appear on the page
--- a/src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
+++ b/src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Core\Distributor;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class ProportionalIntegerDistributor implements ProportionalIntegerDistributorInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function distribute($total, array $elements, $amount)
+    {
+        if ($total !== array_sum($elements)) {
+            throw new \InvalidArgumentException('Element sum should be equal with total.');
+        }
+
+        $distributedAmounts = [];
+
+        $sign = $amount < 0 ? -1 : 1;
+        $amount = abs($amount);
+
+        $distributedSum = 0;
+        foreach ($elements as $element) {
+            $distributedAmount = (int) ($sign * floor(($element * $amount) / $total));
+            $distributedSum += $distributedAmount;
+
+            $distributedAmounts[] = $distributedAmount;
+        }
+
+        if ($distributedSum < $amount) {
+            for ($i = 0; $i < ($amount - $distributedSum); $i++) {
+                $distributedAmounts[$i]++;
+            }
+        }
+
+        return $distributedAmounts;
+    }
+}
--- a/src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
+++ b/src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
@@ -11,22 +11,25 @@
 
 namespace Sylius\Component\Core\Distributor;
 
+use Webmozart\Assert\Assert;
+
 /**
  * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
  */
-class ProportionalIntegerDistributor implements ProportionalIntegerDistributorInterface
+final class ProportionalIntegerDistributor implements ProportionalIntegerDistributorInterface
 {
     /**
      * {@inheritdoc}
      */
-    public function distribute($total, array $elements, $amount)
+    public function distribute(array $integers, $amount)
     {
-        if ($total !== array_sum($elements)) {
-            throw new \InvalidArgumentException('Element sum should be equal with total.');
-        }
+        Assert::allInteger($integers);
+        Assert::integer($amount);
 
+        $total = array_sum($integers);
         $distributedAmounts = [];
-        foreach ($elements as $element) {
+
+        foreach ($integers as $element) {
             $distributedAmounts[] = (int) round(($element * $amount) / $total, 0, PHP_ROUND_HALF_DOWN);
         }
 
--- a/src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
+++ b/src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
@@ -26,22 +26,13 @@ class ProportionalIntegerDistributor implements ProportionalIntegerDistributorIn
         }
 
         $distributedAmounts = [];
-
-        $sign = $amount < 0 ? -1 : 1;
-        $amount = abs($amount);
-
-        $distributedSum = 0;
         foreach ($elements as $element) {
-            $distributedAmount = (int) ($sign * floor(($element * $amount) / $total));
-            $distributedSum += $distributedAmount;
-
-            $distributedAmounts[] = $distributedAmount;
+            $distributedAmounts[] = (int) round(($element * $amount) / $total, 0, PHP_ROUND_HALF_DOWN);
         }
 
-        if (abs($distributedSum) < $amount) {
-            for ($i = 0; $i < ($amount - abs($distributedSum)); $i++) {
-                (1 === $sign) ? $distributedAmounts[$i]++ : $distributedAmounts[$i]--;
-            }
+        $missingAmount = $amount - array_sum($distributedAmounts);
+        for ($i = 0; $i < abs($missingAmount); $i++) {
+            $distributedAmounts[$i] += $missingAmount >= 0 ? 1 : -1;
         }
 
         return $distributedAmounts;
--- a/src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
+++ b/src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
@@ -40,7 +40,7 @@ class ProportionalIntegerDistributor implements ProportionalIntegerDistributorIn
 
         if (abs($distributedSum) < $amount) {
             for ($i = 0; $i < ($amount - abs($distributedSum)); $i++) {
-                (1 === $sign) ? $distributedAmounts[$i]++ : $distributedAmounts[$i]++;
+                (1 === $sign) ? $distributedAmounts[$i]++ : $distributedAmounts[$i]--;
             }
         }
 
--- a/src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
+++ b/src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
@@ -38,9 +38,9 @@ class ProportionalIntegerDistributor implements ProportionalIntegerDistributorIn
             $distributedAmounts[] = $distributedAmount;
         }
 
-        if ($distributedSum < $amount) {
-            for ($i = 0; $i < ($amount - $distributedSum); $i++) {
-                $distributedAmounts[$i]++;
+        if (abs($distributedSum) < $amount) {
+            for ($i = 0; $i < ($amount - abs($distributedSum)); $i++) {
+                (1 === $sign) ? $distributedAmounts[$i]++ : $distributedAmounts[$i]++;
             }
         }
 
--- a/src/Sylius/Component/Core/Promotion/Action/DiscountAction.php
+++ b/src/Sylius/Component/Core/Promotion/Action/DiscountAction.php
@@ -13,11 +13,13 @@ namespace Sylius\Component\Core\Promotion\Action;
 
 use Sylius\Component\Core\Model\AdjustmentInterface;
 use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Model\OrderItemUnitInterface;
 use Sylius\Component\Originator\Originator\OriginatorInterface;
 use Sylius\Component\Promotion\Action\PromotionActionInterface;
 use Sylius\Component\Promotion\Model\PromotionInterface;
 use Sylius\Component\Promotion\Model\PromotionSubjectInterface;
 use Sylius\Component\Resource\Exception\UnexpectedTypeException;
+use Webmozart\Assert\Assert;
 
 /**
  * @author Saša Stamenković <umpirsky@gmail.com>
@@ -39,6 +41,11 @@ abstract class DiscountAction implements PromotionActionInterface
     }
 
     /**
+     * @param array $configuration
+     */
+    abstract protected function isConfigurationValid(array $configuration);
+
+    /**
      * {@inheritdoc}
      */
     public function revert(PromotionSubjectInterface $subject, array $configuration, PromotionInterface $promotion)
@@ -49,11 +56,7 @@ abstract class DiscountAction implements PromotionActionInterface
 
         foreach ($subject->getItems() as $item) {
             foreach ($item->getUnits() as $unit) {
-                foreach ($unit->getAdjustments(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT) as $adjustment) {
-                    if ($promotion === $this->originator->getOrigin($adjustment)) {
-                        $unit->removeAdjustment($adjustment);
-                    }
-                }
+                $this->removeUnitOrderPromotionAdjustmentsByOrigin($unit, $promotion);
             }
         }
     }
@@ -65,19 +68,21 @@ abstract class DiscountAction implements PromotionActionInterface
      */
     protected function isSubjectValid(PromotionSubjectInterface $subject)
     {
-        if (!$subject instanceof OrderInterface) {
-            throw new UnexpectedTypeException($subject, OrderInterface::class);
-        }
+        Assert::implementsInterface($subject, OrderInterface::class);
 
-        if (0 === $subject->countItems()) {
-            return false;
-        }
-
-        return true;
+        return 0 !== $subject->countItems();
     }
 
     /**
-     * @param array $configuration
+     * @param OrderItemUnitInterface $unit
+     * @param PromotionInterface $promotion
      */
-    protected abstract function isConfigurationValid(array $configuration);
+    private function removeUnitOrderPromotionAdjustmentsByOrigin(OrderItemUnitInterface $unit, PromotionInterface $promotion)
+    {
+        foreach ($unit->getAdjustments(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT) as $adjustment) {
+            if ($promotion === $this->originator->getOrigin($adjustment)) {
+                $unit->removeAdjustment($adjustment);
+            }
+        }
+    }
 }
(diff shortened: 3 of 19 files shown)


features/legacy/frontend/cart_promotions_percentage.feature
```diff
@@ -69,5 +69,5 @@ Feature: Checkout percentage discount promotions
         And I added product "Buzz" to cart, with quantity "1"
         When I add product "Woody" to cart, with quantity "3"
         Then I should still be on the cart summary page
-        And "Promotion total: -€586.25" should appear on the page
-        And "Grand total: €1,088.75" should appear on the page
+        And "Promotion total: -€544.38" should appear on the page
+        And "Grand total: €1,130.62" should appear on the page
```
pjedrzejewski: Distribution changes the discounts, by that much?

Zales0123: Well, in fact it does not :smile: It's a funny thing:
- we have order containing 4x **Potato** (€200.00 each), 1x **Buzz** (€500.00) and 3x **Woody** (€125.00)
- it gives us total of €1675.00
- promotion **PR1** gives 25% off on order (we have 8 units, 5 required) -> €1675.00 - 25% = €1256.25
- promotion **PR2** gives 10% off on order (as its total is above €300.00) -> €1256.25 - 10% = €1130.62

so as we see, order is calculated correctly :D I have no idea how it worked before, by for me it's good now.

src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
```diff
@@ -0,0 +1,49 @@
...
+        foreach ($elements as $element) {
+            $distributedAmount = (int) ($sign * floor(($element * $amount) / $total));
+            $distributedSum += $distributedAmount;
+
+            $distributedAmounts[] = $distributedAmount;
+        }
+
+        if (abs($distributedSum) < $amount) {
+            for ($i = 0; $i < ($amount - abs($distributedSum)); $i++) {
+                (1 === $sign) ? $distributedAmounts[$i]++ : $distributedAmounts[$i]--;
```
pamil: `$distributedAmounts[$i] += $sign;`?

src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
```diff
@@ -0,0 +1,49 @@
...
+
+        $distributedSum = 0;
+        foreach ($elements as $element) {
+            $distributedAmount = (int) ($sign * floor(($element * $amount) / $total));
+            $distributedSum += $distributedAmount;
+
+            $distributedAmounts[] = $distributedAmount;
+        }
+
+        if (abs($distributedSum) < $amount) {
```
pamil: Is that `if` really needed? If `$amount` is smaller than `abs($distributedSum)` it will result in `$amount - abs($distributedSum)` being below 0, so that for loop won't be run even once.

src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
```diff
@@ -0,0 +1,49 @@
...
+
+        $distributedAmounts = [];
+
+        $sign = $amount < 0 ? -1 : 1;
+        $amount = abs($amount);
+
+        $distributedSum = 0;
+        foreach ($elements as $element) {
+            $distributedAmount = (int) ($sign * floor(($element * $amount) / $total));
+            $distributedSum += $distributedAmount;
```
pamil: May we calculate the sum by `$distributedSum = array_sum($distributedAmounts)` after the loop?

src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
```diff
@@ -0,0 +1,49 @@
...
+
+            $distributedAmounts[] = $distributedAmount;
+        }
+
+        if (abs($distributedSum) < $amount) {
+            for ($i = 0; $i < ($amount - abs($distributedSum)); $i++) {
```
pamil: Can expression `($amount - abs($distributedSum))` be extracted to a meaningful variable? `$missingAmount` or something like that

src/Sylius/Component/Core/Promotion/Action/DiscountAction.php
```diff
@@ -53,34 +43,41 @@ public function __construct(FactoryInterface $adjustmentFactory, OriginatorInter
...
+        if (!$this->isSubjectValid($subject)) {
+            return;
         }
 
-        foreach ($subject->getAdjustments(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT) as $adjustment) {
-            if ($promotion === $this->originator->getOrigin($adjustment)) {
-                $subject->removeAdjustment($adjustment);
+        foreach ($subject->getItems() as $item) {
+            foreach ($item->getUnits() as $unit) {
+                foreach ($unit->getAdjustments(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT) as $adjustment) {
```
lchrusciel: Can you extract it? I know, that current implementation require less lines, but I bet, extracting some of the logic will improve readability.

src/Sylius/Component/Core/Promotion/Action/DiscountAction.php
```diff
@@ -53,34 +43,41 @@ public function __construct(FactoryInterface $adjustmentFactory, OriginatorInter
...
     {
-        $adjustment = $this->adjustmentFactory->createNew();
-        $adjustment->setType($type);
-        $adjustment->setLabel($promotion->getName());
+        if (!$subject instanceof OrderInterface) {
+            throw new UnexpectedTypeException($subject, OrderInterface::class);
+        }
 
-        $this->originator->setOrigin($adjustment, $promotion);
+        if (0 === $subject->countItems()) {
```
lchrusciel: `return 0 !== $subject->countItems()`

src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class ProportionalIntegerDistributor implements ProportionalIntegerDistributorInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function distribute($total, array $elements, $amount)
```
tuka217: I'm not sure, but if `$elements` are array of prices of elements, then maybe it should have name like `$elementsPrices` or something less misleading?

pamil: Btw. why do we need both `$total` and `$elements` if `$total === array_sum($elements)`?

Zales0123: In fact, ``$elements`` is not array of prices, but array of integers, as it's generic solution, not only for orders ;) However, I agree that ``$total`` parameter is not necessary.

src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
```diff
@@ -0,0 +1,38 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Core\Distributor;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class ProportionalIntegerDistributor implements ProportionalIntegerDistributorInterface
```
lchrusciel: Maybe it would be worthy to close this implementation with `final` keyword? What could change here? Even if somebody would like to override this class, he can just create a new one.

src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
```diff
@@ -0,0 +1,38 @@
...
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Core\Distributor;
```
lchrusciel: We don't depends on any class from `Core` component. IMHO we should put this class to promotion component, it could be useful to anybody, who is using it.

src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function distribute(array $integers, $amount)
+    {
+        $total = array_sum($integers);
+        $distributedAmounts = [];
+
+        foreach ($integers as $element) {
```
lchrusciel: naming of this variable is a little bit confusing to me, but I don't have better idea for them

src/Sylius/Component/Core/Promotion/Action/DiscountAction.php
```diff
@@ -51,34 +44,46 @@ public function __construct(FactoryInterface $adjustmentFactory, OriginatorInter
...
-            );
+        if (!$this->isSubjectValid($subject)) {
+            return;
         }
 
-        foreach ($subject->getAdjustments(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT) as $adjustment) {
-            if ($promotion === $this->originator->getOrigin($adjustment)) {
-                $subject->removeAdjustment($adjustment);
+        foreach ($subject->getItems() as $item) {
+            foreach ($item->getUnits() as $unit) {
```
lchrusciel: :(

src/Sylius/Component/Core/Promotion/Action/DiscountAction.php
```diff
@@ -51,34 +44,46 @@ public function __construct(FactoryInterface $adjustmentFactory, OriginatorInter
...
-     * @return AdjustmentInterface
+     * @return bool
      */
-    protected function createAdjustment(PromotionInterface $promotion, $type = AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT)
+    protected function isSubjectValid(PromotionSubjectInterface $subject)
     {
-        $adjustment = $this->adjustmentFactory->createNew();
-        $adjustment->setType($type);
-        $adjustment->setLabel($promotion->getName());
+        if (!$subject instanceof OrderInterface) {
```
lchrusciel: Assert::

src/Sylius/Component/Core/Promotion/Action/DiscountAction.php
```diff
@@ -51,34 +44,46 @@ public function __construct(FactoryInterface $adjustmentFactory, OriginatorInter
...
 
-        $this->originator->setOrigin($adjustment, $promotion);
+        return 0 !== $subject->countItems();
+    }
 
-        return $adjustment;
+    /**
+     * @param array $configuration
+     */
+    protected abstract function isConfigurationValid(array $configuration);
```
lchrusciel: I would suggest tot put this method at the top of class. Even if its protected, is it pretty important... There is even some [example](http://www.php-fig.org/psr/psr-2/#4-5-abstract-final-and-static) of fig

src/Sylius/Component/Core/Promotion/Action/FixedDiscountAction.php
```diff
@@ -54,13 +87,23 @@ public function getConfigurationFormType()
     }
 
     /**
+     * {@inheritdoc}
+     */
+    protected function isConfigurationValid(array $configuration)
+    {
+        if (!isset($configuration['amount']) || !is_int($configuration['amount'])) {
```
lchrusciel: Assert? Not sure, it there would be suitable method for such a case

src/Sylius/Component/Core/Promotion/Action/ShippingDiscountAction.php
```diff
@@ -43,8 +67,44 @@ public function execute(PromotionSubjectInterface $subject, array $configuration
     /**
      * {@inheritdoc}
      */
+    public function revert(PromotionSubjectInterface $subject, array $configuration, PromotionInterface $promotion)
+    {
+        if (!$subject instanceof OrderInterface && !$subject instanceof OrderItemInterface) {
```
lchrusciel: Asserts? I have one problem with this library. IIRC it can only throws invalid argument exceptions...

src/Sylius/Component/Core/Promotion/Applicator/UnitsPromotionAdjustmentsApplicator.php
```diff
@@ -0,0 +1,115 @@
...
+        $this->distributor = $distributor;
+        $this->originator = $originator;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function apply(OrderInterface $order, PromotionInterface $promotion, array $adjustmentsAmounts)
+    {
+        if ($order->countItems() !== count($adjustmentsAmounts)) {
```
lchrusciel: Assert

src/Sylius/Component/Core/Promotion/Applicator/UnitsPromotionAdjustmentsApplicator.php
```diff
@@ -0,0 +1,115 @@
...
+            throw new \InvalidArgumentException(
+                'Number of adjustments amount to distribute must be equal with number of order items.'
+            );
+        }
+
+        $i = 0;
+        foreach ($order->getItems() as $item) {
+            if (0 === $adjustmentsAmounts[$i]) {
```
lchrusciel: Can we have `0` value in a middle of `adjustmentsAmounts` array? 
If yes, this code has a bug. For example for amounts: 1, 0, 1 and 3 items, this loop till apply adjustment to first item, but  when it will check the second element of array it will stop in this loop, because we don't increment index.

If no, break will be much more appropriate here.

src/Sylius/Component/Core/Promotion/Applicator/UnitsPromotionAdjustmentsApplicator.php
```diff
@@ -0,0 +1,115 @@
...
+     * @param PromotionInterface $promotion
+     * @param int $itemPromotionAmount
+     */
+    private function applyAdjustmentsOnItemUnits(OrderItemInterface $item, PromotionInterface $promotion, $itemPromotionAmount)
+    {
+        $splitPromotionAmount = $this->distributor->distribute($itemPromotionAmount, $item->getQuantity());
+
+        $i = 0;
+        foreach ($item->getUnits() as $unit) {
+            if (0 === $splitPromotionAmount[$i]) {
```
lchrusciel: same as above

src/Sylius/Component/Core/spec/Promotion/Applicator/UnitsPromotionAdjustmentsApplicatorSpec.php
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    function it_applies_promotion_adjustments_on_all_units_of_given_order(
+        AdjustmentFactoryInterface $adjustmentFactory,
+        AdjustmentInterface $firstAdjustment,
+        AdjustmentInterface $secondAdjustment,
+        AdjustmentInterface $thirdAdjustment,
+        IntegerDistributorInterface $distributor,
+        OrderInterface $order,
+        OrderItemInterface $coltItem,
```
lchrusciel: Why do you want to have a colt as a order item? Do you want to buy a gun? Guns are dangerous... But not as dangerous as a people who has them. 

> Guns don't kill people...

src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php
```diff
@@ -0,0 +1,38 @@
...
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class ProportionalIntegerDistributor implements ProportionalIntegerDistributorInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function distribute(array $integers, $amount)
+    {
+        $total = array_sum($integers);
```
bendavies:
```
Assert::allInteger($integers);
Assert::numeric($amount);
`

File: src/Sylius/Component/Core/Promotion/Action/DiscountAction.php (hunk 12 above)
Comment (Zales0123) [inline src/Sylius/Component/Core/Promotion/Action/DiscountAction.php]: I also don't like nested loops, but I'm not sure is it good for readability to extract one more function here. Making ``function``, which contains ``loop``, which contains ``function``, which contains ``loop``, which contains ``function``, which contains ``loop`` which contains a main logic doesn't seem as better way to do that 😄

Comment (lchrusciel) [inline src/Sylius/Component/Core/Promotion/Action/DiscountAction.php]: I understand why you want to avoid loopoception. I'm just sad, that I have seen it :(

File: src/Sylius/Component/Core/Promotion/Action/ShippingDiscountAction.php (hunk 16 above)
Comment (Zales0123) [inline src/Sylius/Component/Core/Promotion/Action/ShippingDiscountAction.php]: Sadly we check two interfaces here and probably it's not possible to make it shiny and bright with ``Assert`` :(

File: src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php (hunk 10 above)
Comment (Zales0123) [inline src/Sylius/Component/Core/Distributor/ProportionalIntegerDistributor.php]: Agreed 👍 However, it's placed here because previously implemented ``IntegerDistributor`` has been placed here too. Maybe we can move them together in separate PR to avoid making them sad and alone?

File: src/Sylius/Component/Core/Promotion/Applicator/UnitsPromotionAdjustmentsApplicator.php (hunk 18 above)
Comment (Zales0123) [inline src/Sylius/Component/Core/Promotion/Applicator/UnitsPromotionAdjustmentsApplicator.php]: In current implementation there is no possibility of such a situation, but we should definitely it could happen (e.g. with different distributor implementation). Then it's a bug - good catch! 👍

File: src/Sylius/Component/Core/Promotion/Action/FixedDiscountAction.php [hunk 22]
```diff
@@ -54,13 +88,22 @@ public function getConfigurationFormType()
...
+    /**
      * @param int $promotionSubjectTotal
      * @param int $targetPromotionAmount
      *
      * @return int
      */
     private function calculateAdjustmentAmount($promotionSubjectTotal, $targetPromotionAmount)
     {
-        return -1 * min($promotionSubjectTotal, $targetPromotionAmount);
+        return -1 * min(abs($promotionSubjectTotal), abs($targetPromotionAmount));
```
michalmarcinkowski: Why do we need `abs`? Both values should be positive, we can assert that.

src/Sylius/Component/Core/spec/Promotion/Applicator/UnitsPromotionAdjustmentsApplicatorSpec.php
```diff
@@ -0,0 +1,212 @@
...
+        OrderItemUnitInterface $coltUnit,
+        OrderItemUnitInterface $magnumUnit,
+        OriginatorInterface $originator,
+        PromotionInterface $promotion
+    ) {
+        $order->countItems()->willReturn(2);
+
+        $order
+            ->getItems()
+            ->willReturn(new \ArrayIterator([$coltItem->getWrappedObject(), $magnumItem->getWrappedObject()]))
```
michalmarcinkowski: `ArrayIterator` => `ArrayCollection` everywhere.

src/Sylius/Component/Core/spec/Distributor/ProportionalIntegerDistributorSpec.php
```diff
@@ -0,0 +1,67 @@
...
+
+    function it_distributes_integer_based_on_elements_participation_in_total_even_if_it_can_be_divided_easily()
+    {
+        $this->distribute([4300, 1400, 2300], 300)->shouldReturn([162, 52, 86]);
+    }
+
+    function it_distributes_negative_integer_based_on_elements_participation_in_total_even_if_it_can_be_divided_easily()
+    {
+        $this->distribute([4300, 1400, 2300], -300)->shouldReturn([-162, -52, -86]);
+    }
```
michalmarcinkowski: Please add at least one case when the amount is indivisible by the number of items. E.g. distribute `299` instead of `300`.

src/Sylius/Component/Core/Promotion/Action/PercentageDiscountAction.php
```diff
@@ -11,35 +11,66 @@
...
-        $adjustment->setAmount(-$adjustmentAmount);
+        $this->isConfigurationValid($configuration);
+
+        $promotionAmount = $this->calculateAdjustmentAmount($subject->getPromotionSubjectTotal(), $configuration['percentage']);
+        if (0 === $promotionAmount) {
+            return;
+        }
 
-        $subject->addAdjustment($adjustment);
+        $splitPromotion = $this->distributor->distribute($promotionAmount, $subject->countItems());
```
michalmarcinkowski: That looks wrong, this should be `->distribute($itemsTotals, $promotionAmount)` as above.

src/Sylius/Component/Core/Promotion/Action/FixedDiscountAction.php
```diff
@@ -11,38 +11,72 @@
...
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
  */
 class FixedDiscountAction extends DiscountAction
 {
     const TYPE = 'order_fixed_discount';
 
     /**
+     * @var ProportionalIntegerDistributorInterface
+     */
+    private $distributor;
```
michalmarcinkowski: Maybe we could call it `$proportionalDistributor` to avoid confusion with ordinary `IntegerDistributor`

src/Sylius/Component/Core/Promotion/Applicator/UnitsPromotionAdjustmentsApplicator.php
```diff
@@ -0,0 +1,112 @@
...
+        Assert::eq($order->countItems(), count($adjustmentsAmounts));
+
+        $i = 0;
+        foreach ($order->getItems() as $item) {
+            if (0 === $adjustmentsAmounts[$i]) {
+                continue;
+            }
+
+            $this->applyAdjustmentsOnItemUnits($item, $promotion, $adjustmentsAmounts[$i]);
+            $i++;
```
michalmarcinkowski: Incrementing here will make an infinite loop on `0` adjustment amount. We should do the following:
```
$adjustmentAmount = $adjustmentsAmounts[$i++];
if (0 === $adjustmentAmount) {
    continue;
}
...
```
and add a spec for that.

pjedrzejewski:
Nice work Mateusz, thank you! 👍

PR #4769, opened 2016-04-12

SyliusBot:


src/Sylius/Bundle/ContentBundle/Resources/translations/messages.th.yml
```diff
@@ -88,9 +88,9 @@ sylius:
             title: หัวข้อ
         static_content:
             body: Body
-            id: ไอดี
+            id: ID
             internal_name: ชื่อภายใน
-            menu_nodes: เมนูต่างๆ
+            menu_nodes: Menu nodes
```
michalmarcinkowski: Is this change correct?

src/Sylius/Bundle/CoreBundle/Resources/translations/messages.th.yml
```diff
@@ -7,7 +7,7 @@ sylius:
             add_product_configuration:
                 price: ราคาต่อหน่วย
                 quantity: จำนวน
-                variant: ตัวแปร
+                variant: Variant
```
michalmarcinkowski: As well as this?

src/Sylius/Bundle/ContentBundle/Resources/translations/messages.th.yml
pamil: It is correct in terms of SyliusBot work, the Thai translation was changed on CrowdIn. Since we don't verify translations, there's no guarantee that translations make sense.

michalmarcinkowski: I believe it's not SyliusBot fault 😄  just wondering if it's ok to accept such changes... /cc @pjedrzejewski

pjedrzejewski:
These mistakes need to be corrected on Crowdin. We will have some until we only import approved translations. Probably someone accidentally clicked the english suggestion.

PR #4785, opened 2016-04-14
[Promotion][UX] Sort promotion rules and actions in selectbox

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/PromotionBundle/DependencyInjection/Compiler/RegisterPromotionActionsPass.php
+++ b/src/Sylius/Bundle/PromotionBundle/DependencyInjection/Compiler/RegisterPromotionActionsPass.php
@@ -31,7 +31,10 @@ class RegisterPromotionActionsPass implements CompilerPassInterface
         $registry = $container->getDefinition('sylius.registry.promotion_action');
         $actions = [];
 
-        foreach ($container->findTaggedServiceIds('sylius.promotion_action') as $id => $attributes) {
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/PromotionBundle/DependencyInjection/Compiler/RegisterPromotionActionsPass.php
```diff
@@ -16,8 +16,6 @@
 use Symfony\Component\DependencyInjection\Reference;
 
 /**
- * Registers all promotion actions in registry service.
```
michalmarcinkowski: This comment can be useful, I think we should leave it.

src/Sylius/Bundle/PromotionBundle/DependencyInjection/Compiler/RegisterRuleCheckersPass.php
```diff
@@ -16,8 +16,6 @@
 use Symfony\Component\DependencyInjection\Reference;
 
 /**
- * Registers all rule checkers in registry service.
```
michalmarcinkowski: Same as above.

pjedrzejewski:
Thank you Mateusz!

PR #4789, opened 2016-04-14
[Grid] Introduced grid datetime field type

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

It could be useful for fields like ``createdAt`` etc.

### DIFF
--- a/src/Sylius/Component/Grid/FieldTypes/DatetimeFieldType.php
+++ b/src/Sylius/Component/Grid/FieldTypes/DatetimeFieldType.php
@@ -13,11 +13,12 @@ namespace Sylius\Component\Grid\FieldTypes;
 
 use Sylius\Component\Grid\DataExtractor\DataExtractorInterface;
 use Sylius\Component\Grid\Definition\Field;
+use Webmozart\Assert\Assert;
 
 /**
- * @author Paweł Jędrzejewski <pawel@sylius.org>
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
  */
-class StringFieldType implements FieldTypeInterface
+class DatetimeFieldType implements FieldTypeInterface
 {
     /**
      * @var DataExtractorInterface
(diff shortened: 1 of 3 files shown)


src/Sylius/Component/Grid/FieldTypes/DatetimeFieldType.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function render(Field $field, $data)
+    {
+        /** @var \DateTime $value */
+        $value = $this->dataExtractor->get($field, $data);
+        if (null === $value) {
```
michalmarcinkowski: Shouldn't we validate if `$value` is instance of `\DateTime`?

src/Sylius/Component/Grid/spec/FieldTypes/DatetimeFieldTypeSpec.php
```diff
@@ -0,0 +1,67 @@
...
+
+        $dateTime->format('Y-m-d')->willReturn('2001-10-10');
+
+        $this->render($field, ['foo' => 'bar'])->shouldReturn('2001-10-10');
+    }
+
+    function it_returns_null_if_property_accessor_returns_null(DataExtractorInterface $dataExtractor, Field $field)
+    {
+        $dataExtractor->get($field, ['foo' => 'bar'])->willReturn(null);
+        $field->getOptions()->willReturn(['format' => 'Y-m-d']);
```
michalmarcinkowski: Should not be called.

src/Sylius/Component/Grid/FieldTypes/DatetimeFieldType.php
```diff
@@ -0,0 +1,56 @@
...
+            return null;
+        }
+
+        return $value->format($field->getOptions()['format']);
```
michalmarcinkowski: Is `format` a required option?

src/Sylius/Component/Grid/FieldTypes/DatetimeFieldType.php
Zales0123: Sure! I didn't add it, because I wasn't sure where we are now with fields validation... it would be good to validate ``format`` variable to be proper datetime format :smile: but for now - validation of value type :+1:

bendavies:
I was about do do this! great stuff.

liverbool:
I thinks datetime should be intl/localizeddate with dateformat,timeformat.

pjedrzejewski:
Thank you Mateusz!

@liverbool Absolutely, we will add that in future PRs. ;)

PR #4798, opened 2016-04-15
[Twig] Currency and Money helpers clean up

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/spec/Templating/Helper/MoneyHelperSpec.php
+++ b/src/Sylius/Bundle/CoreBundle/spec/Templating/Helper/MoneyHelperSpec.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace spec\Sylius\Bundle\CoreBundle\Templating\Helper;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\Bundle\CurrencyBundle\Templating\Helper\MoneyHelper;
+use Sylius\Bundle\MoneyBundle\Formatter\MoneyFormatterInterface;
+use Sylius\Bundle\MoneyBundle\Templating\Helper\MoneyHelperInterface;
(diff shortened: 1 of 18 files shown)


pamil:
Wow, Travis feedback is as fast as it has never been before! :tada:

pjedrzejewski:
AmountFormatter vs. MoneyFormatter? Amount is quite wide term.

src/Sylius/Bundle/CoreBundle/spec/Templating/Helper/MoneyHelperSpec.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        $this->shouldHaveType('Sylius\Bundle\CoreBundle\Templating\Helper\MoneyHelper');
+    }
+
+    function it_implements_money_helper_interface()
+    {
+        $this->shouldImplement(MoneyHelperInterface::class);
+    }
+
+    function it_allows_to_format_money_in_different_currencies(
```
pamil: Shouldn't it be divided into more sections:
 - [x] it formats money using context currency and locale if only amount is given
 - [x] it formats money using given currency and locale if they are given
 - [x] it formats money using default currency if not given
 - [x] it formats money using default locale if not given

src/Sylius/Bundle/MoneyBundle/Formatter/AmountFormatter.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    /**
+     * @param int $amount
+     * @param string $currency
+     * @param bool $decimal
+     * @param string $locale
+     *
+     * @return string
+     */
+    public function format($amount, $currency, $decimal = false, $locale = 'en')
```
pamil: IMO `$decimal` is redundant here - if currency is given, use `\NumberFormatter::CURRENCY`, if not  `\NumberFormatter::DECIMAL`. If we require currency to be passed, we can just delete that option, because currency won't be used anyway.

pamil:
@pjedrzejewski :+1: for MoneyFormatter

michalmarcinkowski:
:+1:

pjedrzejewski:
Thanks Łukasz, really nice work!

PR #4742, opened 2016-04-11
[Admin][PaymentMethod] Payment method features

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Partially based on #4736

PR #4604, opened 2016-03-23
[Admin][Currency] New behats for Currency

tuka217:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #4375
| License         | MIT
| Doc PR          | -

PR #4805, opened 2016-04-15
[Admin] Channel features

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Based on #4742 and #4604

### DIFF
--- a/features/channel/managing_channels/adding_channel.feature
+++ b/features/channel/managing_channels/adding_channel.feature
@@ -7,7 +7,7 @@ Feature: Adding a new channel
     Background:
         Given I am logged in as an administrator
 
-    @todo
+    @ui
     Scenario: Adding a new channel
         Given I want to create a new channel
         When I specify its code as "MOBILE"
--- a/features/channel/managing_channels/adding_channel.feature
+++ b/features/channel/managing_channels/adding_channel.feature
@@ -1,0 +1,29 @@
+@managing_channels
+Feature: Adding a new channel
+    In order to sell through multiple websites or mobile applications
+    As an Administrator
+    I want to add a new channel to the registry
+
+    Background:
+        Given I am logged in as an administrator
+
+    @todo
+    Scenario: Adding a new channel
+        Given I want to create a new channel
+        When I specify its code as "MOBILE"
+        And I name it "Mobile channel"
+        And I add it
+        Then I should be notified that it has been successfully created
+        And the channel "Mobile channel" should appear in the registry
+
+    @ui
+    Scenario: Adding a new channel with addtional fields
+        Given I want to create a new channel
+        When I specify its code as "MOBILE"
+        And I name it "Mobile channel"
+        And I describe it as "Main distribution channel for mobile apps"
+        And I set its hostname as "m.avengers-gear.com"
+        And I define its color as "blue"
+        And I add it
+        Then I should be notified that it has been successfully created
+        And the channel "Mobile channel" should appear in the registry
--- a/features/channel/managing_channels/not_being_able_to_disable_last_available_channel.feature
+++ b/features/channel/managing_channels/not_being_able_to_disable_last_available_channel.feature
@@ -1,0 +1,18 @@
+@managing_channels
+Feature: Toggling a channel
+    In order to change channels which are available to my customers
+    As an Administrator
+    I want to be able to switch state of channel between enable and disable
+
+    Background:
+        Given the store operates on a channel named "Web Channel"
+        And I am logged in as an administrator
+
+    @ui
+    Scenario: Disabling the last available channel
+        Given the channel "Web Channel" is enabled
+        And I want to modify this channel
+        When I disable it
+        And I save my changes
+        Then I should be notified that at least one channel has to be defined is required
+        And channel with name "Web Channel" should still be enabled
--- a/features/channel/managing_channels/selecting_currencies_available_for_channel.feature
+++ b/features/channel/managing_channels/selecting_currencies_available_for_channel.feature
@@ -1,0 +1,28 @@
+@managing_channels
+Feature: Selecting available currencies for a channel
+    In order to give possibility to pay in different currencies on my stores website
+    As an Administrator
+    I want to be able to select available currencies
+
+    Background:
+        Given the store has currency Euro
+        And I am logged in as an administrator
+
+    @ui
+    Scenario: Adding a new channel with currencies
+        Given I want to create a new channel
+        When I specify its code as MOBILE
+        And I name it "Mobile store"
+        And I allow to paying Euro for this channel
+        And I add it
+        Then I should be notified that it has been successfully created
+        And paying in Euro should be possible for the "Mobile store" channel
+
+    @ui
+    Scenario: Adding currencies to an existing channel
+        Given the store operates on a channel named "Web store"
+        And I want to modify this channel
+        And I allow to paying Euro for this channel
+        And I save my changes
+        Then I should be notified that it has been successfully edited
+        And paying in Euro should be possible for the "Web store" channel
--- a/features/channel/managing_channels/toggling_channel.feature
+++ b/features/channel/managing_channels/toggling_channel.feature
@@ -1,0 +1,28 @@
+@managing_channels
+Feature: Toggling a channel
+    In order to change channels which are available to my customers
+    As an Administrator
+    I want to be able to switch state of channel between enable and disable
+
+    Background:
+        Given the store operates on a channel named "Web Channel"
+        And the store operates on another channel named "Mobile Channel"
+        And I am logged in as an administrator
+
+    @ui
+    Scenario: Disabling the channel
+        Given the channel "Web Channel" is enabled
+        And I want to modify this channel
+        When I disable it
+        And I save my changes
+        Then I should be notified that it has been successfully edited
+        And this channel should be disabled
+
+    @ui
+    Scenario: Enabling the channel
+        Given the channel "Web Channel" is disabled
+        And I want to modify this channel
+        When I enable it
+        And I save my changes
+        Then I should be notified that it has been successfully edited
+        And this channel should be enabled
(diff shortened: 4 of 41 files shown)


features/channel/managing_channels/adding_channel.feature
```diff
@@ -0,0 +1,17 @@
+@managing_channels
+Feature: Adding a new channel
+    In order to offer products in different ways
```
pjedrzejewski: In order to sell through multiple websites or mobile applications

features/channel/managing_channels/adding_channel.feature
```diff
@@ -0,0 +1,17 @@
...
+        Given I am logged in as an administrator
+
+    @todo
+    Scenario: Adding a new channel
+        Given I want to create a new channel
+        When I specify its code as "MOBILE"
+        And I name it "Mobile channel"
+        And I add it
+        Then I should be notified that it has been successfully created
+        And the channel "FedEx Carrier" should appear in the registry
```
pjedrzejewski: :D

features/channel/managing_channels/not_being_able_to_disable_last_available_channel.feature
```diff
@@ -0,0 +1,18 @@
...
+        And I am logged in as an administrator
+
+    @ui
+    Scenario: Disabling the last available channel
+        Given the channel "Web Channel" is enabled
+        And I want to modify this channel
+        When I disable it
+        And I save my changes
+        Then I should be notified that at least one channel has to be defined is required
+        And channel with name "Web Channel" should still be enabled
```
tuka217: There shouldn't be the second scenario about enabling a channel?

features/channel/managing_channels/selecting_available_locales_for_channel.feature
```diff
@@ -0,0 +1,28 @@
+@managing_channels
+Feature: Selecting available locales for a channel
+    In order to sell have stores website available in different languages
```
tuka217: Without `sell`?

src/Sylius/Behat/Context/Setup/ChannelContext.php
```diff
@@ -89,13 +99,34 @@ public function storeOperatesOnASingleChannel()
...
+     */
+    public function theChannelIsEnabled(ChannelInterface $channel)
+    {
+        $channel->enable();
+        $this->channelManager->flush();
+        $this->sharedStorage->set('channel', $channel);
+    }
+
+    /**
+     * @Given the channel :channel is disabled
```
Zales0123: What about giving there parameter ``(enabled|disabled)`` and use ``$channel->setEnabled(true|false)``? It would allows us to do this two steps in one function ;)

features/channel/managing_channels/toggling_channel.feature
```diff
@@ -0,0 +1,28 @@
...
+        Then I should be notified about successful edition
+        And this channel should be disabled
+
+    @ui
+    Scenario: Enabling the channel
+        Given the channel "Web Channel" is disabled
+        And I want to modify this channel
+        When I enable it
+        And I save my changes
+        Then I should be notified about successful edition
```
tuka217: `I should be notified that it has been successfully edited`

src/Sylius/Behat/Context/Ui/Admin/ManagingChannelsContext.php
```diff
@@ -0,0 +1,409 @@
...
+
+    /**
+     * @Then I should see :numberOfChannels channels in the list
+     */
+    public function iShouldSeeChannelsInTheList($numberOfChannels)
+    {
+        $foundRows = $this->indexPage->countItems();
+
+        Assert::true(
+            ((int) $numberOfChannels) === $foundRows,
```
tuka217: `Assert::eq` ?

src/Sylius/Behat/Context/Ui/Admin/ManagingChannelsContext.php
```diff
@@ -0,0 +1,409 @@
...
+        );
+    }
+
+    /**
+     * @Then the code field should be disabled
+     */
+    public function theCodeFieldShouldBeDisabled()
+    {
+        Assert::true(
+            $this->updatePage->isCodeDisabled(),
```
tuka217: Shouldn't we have logic about checking the state of element in context and pages should only provide states without assert it?

src/Sylius/Behat/Page/Admin/Channel/CreatePageInterface.php
```diff
@@ -11,30 +11,44 @@
...
+     * @param string $color
+     */
+    public function defineColor($color);
+
+    /**
+     * @param string $language
+     */
+    public function makeItAvailableFor($language);
+
+    public function enable();
```
tuka217: It should be at the beginning of this interface.

src/Sylius/Behat/Page/Admin/Channel/UpdatePageInterface.php
```diff
@@ -30,7 +30,21 @@ public function setTheme($themeName);
...
+     * @param string $language
+     */
+    public function makeItAvailableFor($language);
+
+    /**
+     * @param string $language
+     */
+    public function isAvailableFor($language);
+
+    public function enable();
```
tuka217: Same here

src/Sylius/Behat/Context/Setup/ChannelContext.php
lchrusciel: I'm not convinced. We could do it in one function, but we would have to add if statement. I think it is more readable this way. When we can check condition a little bit earlier. 
Going further that way, I could us `(enabled|disabled)` as a string parameter `state`, cut off last letter and just call `$channel-$state()`. PHP magic...
Maybe extracting 
```
        $this->channelManager->flush();
        $this->sharedStorage->set('channel', $channel);
```
would be a good solution?

Zales0123: I'm 100% against magic :D "Cutting a letter to call a function" smells really bad for me. I think that we have ``setEnabled`` function to use it in situations like this one. We have parameter, that can be true or false (we could even write transformer for such a case :D) and it can be used to set resource ``enabled`` state. If you're still not convinced, I'm for extracting common logic ;)

src/Sylius/Behat/Context/Ui/Admin/ManagingChannelsContext.php
lchrusciel: I think it would be better to leave this check on page side. It would make page replacement easier.

features/channel/managing_channels/adding_channel.feature
```diff
@@ -0,0 +1,29 @@
...
+        Then I should be notified that it has been successfully created
+        And the channel "Mobile channel" should appear in the registry
+
+    @ui
+    Scenario: Adding a new channel with addtional fields
+        Given I want to create a new channel
+        When I specify its code as "MOBILE"
+        And I name it "Mobile channel"
+        And I describe it as "Main distribution channel for mobile apps"
+        And I set its hostname as "http://m.avengers-gear.com"
```
pamil: Hostname should be without `http://` part

features/channel/managing_channels/selecting_currencies_available_for_channel.feature
```diff
@@ -0,0 +1,28 @@
...
+        And paying in Euro should be possible for the "Mobile store" channel
+
+    @ui
+    Scenario: Adding currencies to an existing channel
+        Given the store operates on a channel named "Web store"
+        And I want to modify this channel
+        And I allow to paying Euro for this channel
+        And I save my changes
+        Then I should be notified that it has been successfully edited
+        And paying in Euro should be possible for the "Web store" channel
```
pamil: `Euro` -> `euro`?

src/Sylius/Behat/Context/Ui/Admin/ManagingChannelsContext.php
```diff
@@ -0,0 +1,482 @@
...
+     */
+    public function iShouldBeNotifiedItHasBeenSuccessfulCreation()
+    {
+        $this->notificationChecker->checkCreationNotification(self::RESOURCE_NAME);
+    }
+
+    /**
+     * @Then the channel :channelName should appear in the registry
+     * @Then the channel :channelName should be in the registry
+     * @Then /^(this channel) should still be in the registry$/
```
pamil: It will return channel instance

src/Sylius/Behat/Context/Ui/Admin/ManagingChannelsContext.php
```diff
@@ -0,0 +1,482 @@
...
+     */
+    public function theChannelShouldAppearInTheRegistry($channelName)
+    {
+        $this->iWantToBrowseChannels();
+
+        Assert::true(
+            $this->indexPage->isResourceOnPage(['name' => $channelName]),
```
pamil: And we assume, that channel has `_toString()` that returns its name, but it's not a part of channel interface

src/Sylius/Behat/Context/Ui/Admin/ManagingChannelsContext.php
```diff
@@ -0,0 +1,482 @@
...
+    /**
+     * @Then channel with :element :value should not be added
+     */
+    public function channelWithShouldNotBeAdded($element, $value)
+    {
+        $this->iWantToBrowseChannels();
+
+        Assert::false(
+            $this->indexPage->isResourceOnPage([$element => $value]),
+            sprintf('Channel with %s %s was created, but it should not.', $element, $value)
```
pamil: `Channel with %s "%s"`?

pjedrzejewski:
Nice work Łukasz, thanks! 👍

Issue #2057, opened 2014-10-27
support configuration a PK lookup field for SyliusResourceBundle


lsmith77:
it would be useful to be able to define what field should be used for PK lookups in which case the Controller should use ``find()`` rather than ``findByOne()``. Also notice that we need to add a "/" to the beginning of the id. In fact it might be useful to be able to configure a root path.

Here just an ugly hack to illustrate what I mean.

```
diff --git a/Controller/ResourceController.php b/Controller/ResourceController.php
index f7ec683..b1dce26 100644
--- a/Controller/ResourceController.php
+++ b/Controller/ResourceController.php
@@ -351,16 +351,22 @@ class ResourceController extends FOSRestController
 
         $criteria = array_merge($default, $criteria);
 
-        if (!$resource = $this->resourceResolver->getResource(
-            $this->getRepository(),
-            'findOneBy',
-            array($this->config->getCriteria($criteria)))
-        ) {
+        if (count($criteria) === 1 && array_key_exists('id', $criteria)) {
+            $criteria = '/'.$criteria['id'];
+            $method = 'find';
+        } else {
+            $criteria = $this->config->getCriteria($criteria);
+            $method = 'findOneBy';
+        }
+
+        $resource = $this->resourceResolver->getResource($this->getRepository(), $method, array($criteria));
+
+        if (!$resource) {
             throw new NotFoundHttpException(
                 sprintf(
                     'Requested %s does not exist with these criteria: %s.',
                     $this->config->getResourceName(),
-                    json_encode($this->config->getCriteria($criteria))
+                    json_encode($criteria)
                 )
             );
         }
```

PR #4618, opened 2016-03-24
[ResourceBundle] Merge request criteria in SingleResourceProvider

dragosprotung:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT
| Doc PR          | no

Before the introduction of the new ResourceController this was the expected behaviour.
This is still reflected in the documentation

PR #4810, opened 2016-04-16
[SingleResourceProvider] Use repository find() method.

dantleech:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Currently the `SingleResourceProvider` uses the `findBy` method of the repository in order to locate an item by ID - this works fine with the ORM but will fail with the Doctrine PHPCR ODM which does not map the `id` to a field.

This PR changes the behavior to call the `find` method when an ID is given rather than `findBy`


pjedrzejewski:
:+1:

dantleech:
Also think this would close https://github.com/Sylius/Sylius/issues/2057

pjedrzejewski:
Thanks @dantleech!

aramalipoor:
@pjedrzejewski What about #2910 ?

dragosprotung:
@dantleech @pjedrzejewski What about #4618 ?

Issue #4809, opened 2016-04-16
[GridBudnle] has hard dependency on Doctrine ORM.


dantleech:
The ORM driver's DI service has a dependency on `doctrine.orm.entity_manager`, and it therefore not possible to use this bundle without the Doctrine ORM (I would like to add support for PHPCR-ODM).

Would it be possible to either only load the Doctrine ORM / DBAL configurations when ORM/DBAL is existing, or add a configuration option which enables drivers?

pjedrzejewski:
Absolutely, good idea!

PR #4812, opened 2016-04-16
[GridBundle] Remove unsupported drivers during container compilation

dantleech:
| Q               | A
| --------------- | ---
| Bug fix?        | -
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #4809 
| License         | MIT

This PR allows the drivers in the GridBundle to be explicitly enable.

The default behavior is to enable both dbal and orm drivers:

```yaml
sylius_grid:
    drivers: [ "doctrine/orm", "doctrine/dbal" ]
```

### DIFF
--- a/src/Sylius/Bundle/GridBundle/Tests/DependencyInjection/Compiler/RegisterDriversPassTest.php
+++ b/src/Sylius/Bundle/GridBundle/Tests/DependencyInjection/Compiler/RegisterDriversPassTest.php
@@ -1,0 +1,97 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\GridBundle\Tests\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\Container;
+use Sylius\Bundle\GridBundle\DependencyInjection\Compiler\RegisterDriversPass;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+class RegisterDriversPassTest extends \PHPUnit_Framework_TestCase
+{
+    private $container;
+    private $pass;
+
+    public function setUp()
+    {
+        $this->container = new ContainerBuilder();
+        $this->container->register('sylius.registry.grid_driver', '\stdClass');
+        $this->pass = new RegisterDriversPass();
+    }
+
+    /**
+     * @test
+     * @expectedException \InvalidArgumentException
+     * @expectedExceptionMessage needs to have `alias` attribute
+     */
+    public function it_should_throw_an_exception_if_the_driver_has_no_alias_attribute()
+    {
+        $definition = $this->container->register('driver.1', '\stdClass');
+        $definition->addTag(
+            'sylius.grid_driver'
+        );
+
+        $this->pass->process($this->container);
+    }
+
+    /**
+     * @test
+     */
+    public function it_should_register_drivers()
+    {
+        $definition = $this->container->register('driver.1', '\stdClass');
+        $definition->addTag(
+            'sylius.grid_driver',
+            [
+                'alias' => 'foobar',
+            ]
+        );
+
+        $this->pass->process($this->container);
+
+        $definition = $this->container->getDefinition('sylius.registry.grid_driver');
+        $calls = $definition->getMethodCalls();
+        $this->assertCount(1, $calls);
+        $this->assertEquals(
+            [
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/GridBundle/DependencyInjection/Configuration.php
```diff
@@ -38,6 +39,20 @@ public function getConfigTreeBuilder()
     /**
      * @param ArrayNodeDefinition $node
      */
+    private function addDriversSection(ArrayNodeDefinition $node)
+    {
+        $node
+            ->children()
+                ->arrayNode('drivers')
```
dantleech: Not sure how ambiguous this is -- I assume the user could equally register their own drivers, in whch case they would not need to include them here.

Should add an `->info` element.

dantleech:
Or maybe it would be better to use either `class_exists` or a compiler pass to activate the drivers.

dantleech:
@pjedrzejewski do you think this would be best done as a compiler pass? e.g. if the `doctrine.orm.entuty_manager` exists, then add the `doctrine/orm` driver?

pjedrzejewski:
There is a chance we get into trouble with compiler pass, but that would be nicer DX indeed. Let's try it!

dantleech:
Updated. I updated the `RegisterDriverPass` to check the dependencies/references of the driver definitions, if the reference is not available in the container, then the driver will be removed (and not added to the registry).

This is still slightly bad DX as it hides the avialability of the drivers - 

An alternative could be to replace the invalid definitions with an `UnsupportedDriver` class which would throw an exception upon instantiation asking the user to install the required dependencies (e.g. doctrine ORM, PHPCR ODM). But I feel that might be over-engineering for this specific bundle.

pjedrzejewski:
I agree that removing the services without notice to the developer can be a bit confusing. If someone adds a custom driver and make a typo or reference wrong service, his driver will not appear. So maybe more explicit would be your initial solution where you define enabled drivers in yaml. At least for the ones supplied by SyliusGridBundle default, all custom drivers should be loaded automatically. I'd just split ``drivers.xml`` and put doctrine/orm and doctrine/dbal in separate files and load them or not based on this configuration, wdyt?

pamil:
In the end, I think the best solution for both ResourceBundle and GridBundle drivers is to make the drivers like [security authentication listeners in Symfony SecurityBundle](http://symfony.com/doc/current/cookbook/security/custom_authentication_provider.html#the-factory) or [configuration sources in Sylius ThemeBundle](http://docs.sylius.org/en/latest/bundles/SyliusThemeBundle/configuration_sources/creating_custom.html). The one I've written for ThemeBundle only modifies the container if it was activated, so you can enforce having some dependencies while there's still possibility to disable problematic driver.

src/Sylius/Bundle/GridBundle/Tests/DependencyInjection/Compiler/RegisterDriversPassTest.php
```diff
@@ -0,0 +1,97 @@
...
+
+namespace Sylius\Bundle\GridBundle\Tests\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\Container;
+use Sylius\Bundle\GridBundle\DependencyInjection\Compiler\RegisterDriversPass;
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Reference;
+
+class RegisterDriversPassTest extends \PHPUnit_Framework_TestCase
```
pamil: This lib makes testing compiler passes & extensions a lot easier - https://github.com/matthiasnoback/SymfonyDependencyInjectionTest#testing-a-compiler-pass :)

pjedrzejewski:
@pamil Yeah, that would be the best, but will require a lot of work now, with not that big benefit currently. Let's start with something simpler to implement. :)

src/Sylius/Bundle/GridBundle/Tests/DependencyInjection/Compiler/RegisterDriversPassTest.php
dantleech: tbh, I used to use that library myself but over time I (personally) found that it was just obfuscating the tests. imo it's not worth abstracting.

dantleech:
I also like the "provider" approach - it means that all the "drivers" will be available (i.e. if the user makes a typo it means that we can show them all the correct "answers" - good discoverability) and the providers can also throw good exceptions (i.e. "You need to composer require "doctrine/phpcr-odm") in order to use the `doctrine/phpcr-odm` driver.

On the other hand, by explicitly enabling the "drivers" we oblige the user to do this for every single bundle which has drivers - which is not such good DX.

But also don't mind implementing the explicit solution for now.

dantleech:
Replaced by https://github.com/Sylius/Sylius/pull/4972

PR #4821, opened 2016-04-18
[Translations] Add required locales option sylius translations type

### LINKED ISSUE #4822: [Has PR] Translations type should be able to define its own required locales
I have a use case where you want to be able to define the required locales at runtime on the `sylius_translations` type. Before you could also do this with the `a2lix` bundle, but since it was removed recently, I could no longer use it. I have already a PR ready that fixes this 👍

koemeet:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Related tickets | fixes #4822 |
| License | MIT |

This adds a `required_locales` option to the `sylius_translations` form type. This can be very useful for when you want to define what locales are required at runtime. For example, I have a use case where I want to make the locales required of a product based on the unique sum of locales of all linked channels.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Form/Type/ResourceTranslationsType.php
+++ b/src/Sylius/Bundle/ResourceBundle/Form/Type/ResourceTranslationsType.php
@@ -15,6 +15,7 @@ use Sylius\Bundle\ResourceBundle\Form\EventSubscriber\ResourceTranslationsSubscr
 use Sylius\Component\Locale\Provider\LocaleProviderInterface;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\FormBuilderInterface;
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/ResourceBundle/Form/Type/ResourceTranslationsType.php
```diff
@@ -63,6 +63,32 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     }
 
     /**
+     * @param string $locale
+     * @param array $options
+     *
+     * @return boolean
```
michalmarcinkowski: `bool`

michalmarcinkowski:
Looks good! 👍

steffenbrem:
PR is rebased again and ready. You think it will be in 1.0?

PR #4829, opened 2016-04-19
[Reporting] Fix UserRepository coupling with sylius default table naming structure

peteward:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Allows the use of extended User entities.


pjedrzejewski:
Thank you Peter, good catch!

PR #4830, opened 2016-04-19
[Taxation] Entrepreneur tax rate setting

sjonkeesse:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Related tickets |  |
| License | MIT |
### Tax rate settings for individuals or entrepreneurs / resellers

It is currently not possible to apply taxes only to entrepreneurs or individuals. When you sell to a country within the European Union, the tax rate changes if you are an entrepreneur. Therefore I added two properties to the `tax rate` model; `applyToIndividuals` and `applyToEntrepreneursAndResellers`. With this it's possible to assign specific tax rates especially to individuals or entrepreneurs / resellers only.

**Examples**

| Origin | Destination | Legal form | Tax Rate |
| --- | --- | --- | --- |
| The Netherlands (EU) | The Netherlands (EU) | individual | 21% |
| The Netherlands (EU) | The Netherlands (EU) | entrepreneur | 21% |
| The Netherlands (EU) | The Germany (EU) | individual | 21% |
| The Netherlands (EU) | The Germany (EU) | entrepreneur | 0% |
| The Netherlands (EU) | Switzerland (NON EU) | individual | 0% |
| The Netherlands (EU) | Switzerland (NON EU) | entrepreneur | 0% |
| US | US | individual | (differs per state) |
| US | US | reseller | 0% |

In the EU there is always a VAT even if it is 0%.
More information about tax rates can be found [here](https://gist.github.com/sjonkeesse/9253ebfe9c8df9693a6e096955edb38e).
#### Calculations

In Sylius it's possible to create `tax rates` that are already included in the product or shipping price. So if you have a `tax rate` that is based on an `individual` only and the customer is an `entrepreneur/reseller` than the amount of calculated taxes should be taken of the order item price and visa versa.

### DIFF
--- a/src/Sylius/Component/Core/Taxation/Applicator/OrderItemUnitsTaxesApplicator.php
+++ b/src/Sylius/Component/Core/Taxation/Applicator/OrderItemUnitsTaxesApplicator.php
@@ -13,11 +13,14 @@ namespace Sylius\Component\Core\Taxation\Applicator;
 
 use Sylius\Component\Addressing\Model\ZoneInterface;
 use Sylius\Component\Core\Model\AdjustmentInterface;
+use Sylius\Component\Core\Model\CustomerInterface;
 use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Model\OrderItemInterface;
 use Sylius\Component\Core\Model\OrderItemUnitInterface;
+use Sylius\Component\Core\Model\TaxRateInterface;
 use Sylius\Component\Order\Factory\AdjustmentFactoryInterface;
 use Sylius\Component\Taxation\Calculator\CalculatorInterface;
-use Sylius\Component\Taxation\Resolver\TaxRateResolverInterface;
+use Sylius\Component\Taxation\Resolver\TaxRatesResolverInterface;
 
 /**
  * @author Mark McKelvie <mark.mckelvie@reiss.com>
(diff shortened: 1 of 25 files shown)


src/Sylius/Component/Core/Taxation/Applicator/OrderItemUnitsTaxesApplicator.php
```diff
@@ -57,18 +63,17 @@ public function __construct(CalculatorInterface $calculator, AdjustmentFactoryIn
     public function apply(OrderInterface $order, ZoneInterface $zone)
     {
         foreach ($order->getItems() as $item) {
-            $taxRate = $this->taxRateResolver->resolve($item->getVariant(), ['zone' => $zone]);
-            if (null === $taxRate) {
-                continue;
-            }
+            $taxRates = $this->taxRatesResolver->resolve($item->getVariant(), ['zone' => $zone]);
```
steffenbrem: It does not make much sense to apply multiple tax rate to a single order item unit. Not sure what use case you are trying to cover here, but I think this should always be a single tax rate.

steffenbrem:
I think it would be a better idea to let the tax rate resolver first filter out the tax rates that are not enabled for `VAT`, so the `TaxRateResolverInterface::resolve()` signature needs to be changed to accept a `Customer` (or something smaller, but for now lets go with customer). When you have that, you know that the tax rate in there has to apply to the current customer. Then you can just keep the code as it was and you have it much cleaner.

I've made some scenarios to outline the use cases in a understandable view:
```cucumber
Background:
        Given there is a zone "NL" containing country "Netherlands"
        And there is a zone "REST" containing countries "Belgium" and "Germany"
        And the store has "Low VAT" tax rate of 6% for "Clothes" within "NL" zone
        And that tax rate does apply to VAT customers
        And that tax rate is included in price
        And the store has "Low VAT" tax rate of 6% for "Clothes" within "REST" zone
        And that tax rate does not apply to VAT customers
        And that tax rate is included in price
        And the store has a product "T-Shirt" priced at "€100.00"
        And there is customer "john@example.com"

    Scenario: Applies 6% VAT for individual customer to NL
        Given customer "john@example.com" has "Netherlands" as shipping country
        When I add product "T-Shirt" to the cart
        And I proceed without selecting shipping address
        Then my cart total should be "€100.00"
        And my cart taxes should be "€6.00"

    Scenario: Applies 6% VAT for business customer to NL
        Given customer "john@example.com" has "Netherlands" as shipping country
        And that customer has a VAT number
        When I add product "T-Shirt" to the cart
        And I proceed without selecting shipping address
        Then my cart total should be "€100.00"
        And my cart taxes should be "€6.00"

    Scenario: Applies 6% VAT for individual customer to BE
        Given customer "john@example.com" has "Belgium" as shipping country
        When I add product "T-Shirt" to the cart
        And I proceed without selecting shipping address
        Then my cart total should be "€100.00"
        And my cart taxes should be "€6.00"

    Scenario: Applies 0% VAT for business customer to BE
        Given customer "john@example.com" has "Belgium" as shipping country
        When I add product "T-Shirt" to the cart
        And I proceed without selecting shipping address
        Then my cart total should be "€94.34"
        And my cart taxes should be "€0.00"
```

aramalipoor:
Is this generic enough to be in the core?

c33s:
it would also be good to have a "company name" field. often i sell to a company and not to an individual

Issue #4814, opened 2016-04-17
[ServiceRegistry] Add context to the constructor


dantleech:
Currently if a "service" is not found then the following exception is thrown:

```
Service of type "boolean" does not exist.
```

This would be less confusing as:

```
Grid field type "boolean" does not exist.
```

Would suggest adding a context to the service registry constructor, e.g.

```php
new ServiceRegistry('Grid field', FieldTypeInterface::class);
```

brunoroux:
:+1:

pjedrzejewski:
Sounds good to me!

pjedrzejewski:
Thank you for good idea and implementation! 👍

PR #4836, opened 2016-04-20
[Registry] Added service registry context

dantleech:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #4814 
| License         | MIT

Added additional argument to the service registry indicating the context in
order to give more helpful exception messages.

So instead of:

```
Service of type "boolean" does not exist.
```

You get:

```
Grid field of type "boolean" does not exist.
```

The context is optional in order to preserve BC (I would personally consider making it mandatory).

### DIFF
--- a/src/Sylius/Bundle/ReportBundle/Resources/config/services.xml
+++ b/src/Sylius/Bundle/ReportBundle/Resources/config/services.xml
@@ -35,10 +35,12 @@
 
         <service id="sylius.registry.report.data_fetcher" class="%sylius.registry.report.class%">
             <argument>Sylius\Component\Report\DataFetcher\DataFetcherInterface</argument>
+            <argument>data fetcher</argument>
         </service>
 
         <service id="sylius.registry.report.renderer" class="%sylius.registry.report.class%">
             <argument>Sylius\Component\Report\Renderer\RendererInterface</argument>
+            <argument>report renderer</argument>
         </service>
 
         <!-- Report -->
--- a/src/Sylius/Bundle/SettingsBundle/Resolver/ResolverServiceRegistry.php
+++ b/src/Sylius/Bundle/SettingsBundle/Resolver/ResolverServiceRegistry.php
@@ -27,10 +27,11 @@ class ResolverServiceRegistry extends ServiceRegistry
     /**
      * @param string $interface
      * @param SettingsResolverInterface $defaultResolver
+     * @param string
      */
-    public function __construct($interface, SettingsResolverInterface $defaultResolver)
+    public function __construct($interface, SettingsResolverInterface $defaultResolver, $context = 'resolver service')
     {
-        parent::__construct($interface);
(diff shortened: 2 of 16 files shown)


src/Sylius/Bundle/ReportBundle/Resources/config/services.xml
```diff
@@ -35,10 +35,12 @@
 
         <service id="sylius.registry.report.data_fetcher" class="%sylius.registry.report.class%">
             <argument>Sylius\Component\Report\DataFetcher\DataFetcherInterface</argument>
+            <argument>Data fetcher</argument>
         </service>
 
         <service id="sylius.registry.report.renderer" class="%sylius.registry.report.class%">
             <argument>Sylius\Component\Report\Renderer\RendererInterface</argument>
+            <argument>Renderer</argument>
```
dantleech: s/Renderer/Report renderer

src/Sylius/Bundle/SettingsBundle/Resolver/ResolverServiceRegistry.php
```diff
@@ -30,7 +30,7 @@ class ResolverServiceRegistry extends ServiceRegistry
      */
     public function __construct($interface, SettingsResolverInterface $defaultResolver)
     {
-        parent::__construct($interface);
+        parent::__construct($interface, 'Settings resolver');
```
dantleech: Wonder if this should be also defined in the DI configuration.

src/Sylius/Component/Registry/PrioritizedServiceRegistry.php
```diff
@@ -32,12 +32,18 @@ class PrioritizedServiceRegistry implements PrioritizedServiceRegistryInterface
     protected $interface;
 
     /**
+     * @var string
+     */
+    protected $context;
+
+    /**
      * @param $interface
```
dantleech: Missing `@param`

src/Sylius/Bundle/SettingsBundle/Resolver/ResolverServiceRegistry.php
adamelso: It is possible to have more than one instance of this registry - so I would put it in the DI config. :+1:

pjedrzejewski:
👍

dantleech:
Whilst I am here, what do you think about modifying the `NotExistingServiceException` by passing the keys of the registered services, so the message becomes:

```
Grid field of type "bool" does not exist, existing grid field types: "boolean", "string", "number", "foobar"
```

pjedrzejewski:
@dantleech Sounds great!

dantleech:
Updated.

pjedrzejewski:
Thanks @dantleech!

PR #4861, opened 2016-04-24
[Bug Fix] Remove order item inventory listener

### LINKED ISSUE #4860: [Bug] Settings parameters listener onFlush not called when order item inventory is updated
When you have for example the `sylius_taxation` setting saved in the database and the `OrderItemInventoryListener::onFlush` is called, it will disable all `onFlush` event listeners for the current transaction (see: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/EventListener/OrderItemInventoryListener.php#L63). 

This is causing some big issues, because if you have the `Settings` model in the current change set for the unit of work it will never reach the transformation process (that is in the `ParameterTransformerListener` class).

I am not sure if the code inside the `OrderItemInventoryListener::onFlush` is even necessary, as everything seems to work fine without that code.

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | fixes #4860
| License         | MIT

I don't see why this code needs to be there, not sure what the goal was a few years ago, but I don't think it is relevant anymore. Removing it solves the issue mentioned above. Having this removed will make the lifecycle of entities in Doctrine way more predictable! 👍


steffenbrem:
This has been fixed in https://github.com/Sylius/Sylius/pull/5271

PR #4881, opened 2016-04-26
[Resource] Do not make repositories lazy by default

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | -
| License         | MIT

Generated container for environment **prod** with debug **false** gets smaller by 33%, from 3MiB to 2MiB.


pjedrzejewski:
👍

src/Sylius/Bundle/CoreBundle/DependencyInjection/Compiler/LazyCacheWarmupPass.php
```diff
@@ -26,17 +27,20 @@ public function process(ContainerBuilder $container)
...
 
     /**
      * @param ContainerBuilder $container
-     * @param $id
+     * @param string $id
      */
     private function markServiceAsLazy(ContainerBuilder $container, $id)
     {
-        if ($container->hasDefinition($id)) {
+        try {
```
Richtermeister: I am just curious.. what is the advantage/difference of doing the try/catch over the hasDefinition check? Thank you.

PR #4889, opened 2016-04-27
[Bug Fix] Elastica product listener orphaned objects throws error

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

Fixes bug where a "product aware" object that has it's product detached (such as an attribute value) throwed an error when updating ES product index.

If you would for example remove an attribute value object from a product and save that product the listener will fail, since the `ProductAttributeValue` has been orphaned (ready to be deleted).


pjedrzejewski:
Thanks Steffen!

PR #4893, opened 2016-04-27
[Travis] Do not run "cli" and "todo" tagged scenarios

pamil:


pamil:
Supplements #4886, it seems Behat `--tags` arguments overwrites the default tags.

pjedrzejewski:
Thank you Kamil!

PR #4897, opened 2016-04-27
[Theme] Remove database synchronization & simplify workflow

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | -
| License         | MIT

- [x] Remove database synchronization and theme persistence
- [x] Remove dependency on ResourceBundle
- [x] Allow to add own sources of themes configurations (just like in SecurityBundle for authentication providers)
- [x] Rewrite `filesystem` themes configurations source as an optional, lazy loaded source
- [x] Remove `ThemeHydrator`
- [x] Rename `Theme::getCode()` to `Theme::getId()`
- [x] Ensure every theme has a name and an identifier
- [x] Persist themes through one Behat scenario
- [x] Write documentation for the bundle

Please do not merge unless I write the documentation :tada:


src/Sylius/Behat/Context/Setup/ThemeContext.php
```diff
@@ -51,24 +53,32 @@
...
-        ObjectManager $channelManager
+        ObjectManager $channelManager,
+        ContainerInterface $container
     ) {
         $this->sharedStorage = $sharedStorage;
         $this->themeRepository = $themeRepository;
         $this->themeFactory = $themeFactory;
         $this->channelRepository = $channelRepository;
         $this->channelManager = $channelManager;
+        $this->cacheDir = $container->getParameter('kernel.cache_dir');
```
pjedrzejewski: Why not tmp dir?

pamil: It's still a quite dirty solution, for now it uses the path defined in `config_test.yml` and we can't acces the temporary directory path there. I still seek for the solution that will allow us to persist themes configurations in test environment.

app/config/config_test.yml
```diff
@@ -39,6 +39,14 @@ sylius_rbac:
 sylius_channel:
     fake_channel_support: true
 
+sylius_theme:
+    sources:
+        filesystem:
+            locations:
+                - "%kernel.root_dir%/themes"
+                - "%kernel.root_dir%/../vendor/sylius/themes"
+                - "%kernel.cache_dir%/_themes"
```
pamil: Will be rewritten, it's not the best idea for persisting themes through Behat scenario.

src/Sylius/Bundle/CoreBundle/EventListener/ChannelThemeListener.php
```diff
@@ -0,0 +1,61 @@
...
+        if (!$channel instanceof ChannelInterface) {
+            return;
+        }
+
+        $objectManager = $eventArgs->getObjectManager();
+
+        $channelReflection = $objectManager->getClassMetadata(get_class($channel))->getReflectionClass();
+
+        $channelThemeIdReflection = $channelReflection->getProperty('themeId');
+        $channelThemeIdReflection->setAccessible(true);
```
pjedrzejewski: I think we should consider adding a getter here, but that can be next iteration.

pjedrzejewski:
Please rebase.

pjedrzejewski:
Thank you Kamil!

PR #4901, opened 2016-04-28
[Behat] Notifcation context extraction

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| License         | MIT

### DIFF
--- a/etc/behat/services/contexts/ui.xml
+++ b/etc/behat/services/contexts/ui.xml
@@ -27,6 +27,7 @@
         <parameter key="sylius.behat.context.ui.admin.managing_taxons.class">Sylius\Behat\Context\Ui\Admin\ManagingTaxonsContext</parameter>
         <parameter key="sylius.behat.context.ui.admin.managing_tax_rate.class">Sylius\Behat\Context\Ui\Admin\ManagingTaxRateContext</parameter>
         <parameter key="sylius.behat.context.ui.admin.managing_zones.class">Sylius\Behat\Context\Ui\Admin\ManagingZonesContext</parameter>
+        <parameter key="sylius.behat.context.ui.admin.notification.class">Sylius\Behat\Context\Ui\Admin\NotificationContext</parameter>
 
         <parameter key="sylius.behat.context.ui.cart.class">Sylius\Behat\Context\Ui\CartContext</parameter>
         <parameter key="sylius.behat.context.ui.channel.class">Sylius\Behat\Context\Ui\ChannelContext</parameter>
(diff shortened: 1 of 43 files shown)


etc/behat/services/contexts/ui.xml
```diff
@@ -27,6 +27,7 @@
         <parameter key="sylius.behat.context.ui.admin.managing_taxons.class">Sylius\Behat\Context\Ui\Admin\ManagingTaxonsContext</parameter>
         <parameter key="sylius.behat.context.ui.admin.managing_tax_rate.class">Sylius\Behat\Context\Ui\Admin\ManagingTaxRateContext</parameter>
         <parameter key="sylius.behat.context.ui.admin.managing_zones.class">Sylius\Behat\Context\Ui\Admin\ManagingZonesContext</parameter>
+        <parameter key="sylius.behat.context.ui.admin.notification.class">Sylius\Behat\Context\Ui\Admin\NotificationContext</parameter>
```
lchrusciel: Can we use it only on admin side?

src/Sylius/Behat/Context/Ui/Admin/NotificationContext.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $this->notificationChecker = $notificationChecker;
+    }
+
+    /**
+     * @Then I should be notified that it has been successfully created
+     */
+    public function iShouldBeNotifiedItHasBeenSuccessfullyCreated()
+    {
+        $this->notificationChecker->checkNotification('has been successfully created.', NotificationType::success());
```
lchrusciel: Is there any benefit in checking 'has been'? Or even `has been successfully`? We assert a notification type and a key word(_created/updated/deleted_) anyway.

etc/behat/services/contexts/ui.xml
Arminek: Right now we use it only in managing contexts so it is related with admin side. I think we won't have different use case.

lchrusciel: Just wonder if there are any crud notification on show pages. But we can move it later.

michalmarcinkowski: That is correct for now :)

src/Sylius/Behat/Context/Ui/Admin/NotificationContext.php
michalmarcinkowski: I would leave it as it is now. There is one edge case when we could benefit from that, there could be a notification `has not been successfully created.` 💥  :trollface:

pjedrzejewski:
Awesome, thanks Arek!

PR #4902, opened 2016-04-28
[Order] Add custom repository methods to obtain only placed orders

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets |
| License         | MIT

Currently used repository methods in `.../Frontend/OrderController` also find carts. That is not a desired behavior since they are used in "My account" section where only placed orders should appear. When a user has a cart the page breaks because carts don't have `number` which is required by order details route.


pjedrzejewski:
Thanks Michał!

PR #4907, opened 2016-04-28
synchronize RepositoryInterface method signatures

bendavies:
| Q             | A
| ------------- | ---
| Bug fix?      | sorta
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| License       | MIT

This works around a bug in php i just encountered in php 7.0.6: https://bugs.php.net/bug.php?id=72119


pjedrzejewski:
Thanks Ben! 👍

Issue #4874, opened 2016-04-25
[Resource] Catch foreing key on resource deletion


gperdomor:
Hi...

I found a possible bug when you try to delete a resource from api, if that resource is used by other resource, the <code> $this->repository->remove($resource);</code> method throw an exception and the response is an "Internal server error"... I think we should throw a proper message using: 

<pre><code>
try {
    $this->repository->remove($resource);
    $this->eventDispatcher->dispatchPostEvent(ResourceActions::DELETE, $configuration, $resource);
} catch(ForeignKeyConstraintViolationException $e){
    if (!$configuration->isHtmlRequest()) {
        return $this->viewHandler->handle($configuration, View::create([
            'error' => [
                'code' => $e->getSQLState(),
                'message' => $e->getMessage(),
            ]
        ], 500));
    }
}
</code></pre>

michalmarcinkowski:
Fixed in #4911. @gperdomor thanks for reporting and fixing! :+1:

PR #4911, opened 2016-04-28
[Resource] catch foreing key deletion error

gperdomor:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | #4874
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/EventListener/ResourceDeleteSubscriber.php
+++ b/src/Sylius/Bundle/ResourceBundle/EventListener/ResourceDeleteSubscriber.php
@@ -12,9 +12,12 @@
 namespace Sylius\Bundle\ResourceBundle\EventListener;
 
 use Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException;
+use FOS\RestBundle\View\View;
+use FOS\RestBundle\View\ViewHandlerInterface as RestViewHandlerInterface;
 use Sylius\Component\Resource\ResourceActions;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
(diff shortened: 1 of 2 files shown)


michalmarcinkowski:
@gperdomor this is already implemented using [subscriber](https://github.com/michalmarcinkowski/Sylius/blob/26e41f3a9ec3bf608883210980c1ba73737c2d10/src/Sylius/Bundle/ResourceBundle/EventListener/ResourceDeleteSubscriber.php#L75-L75).

gperdomor:
Mmm... I get error for non html requests... We still need fix this, but in
that case, should be fixed in the listener... I will update the PR soon...

michalmarcinkowski:
Exactly, you need to move this fix to the subscriber. Thanks!

gperdomor:
@michalmarcinkowski updated, squashed and ready to review :D

pjedrzejewski:
Looks good, although I am pretty sure we should not expose the sql exception in the API response.

gperdomor:
@pjedrzejewski agree... We can use the translator to return a custom message.

Do you have a proposal for the message?

pjedrzejewski:
@gperdomor I think we should reuse the translation that we use for html view. "entity is in use" or something like that. What do you think?

gperdomor:
@pjedrzejewski updated!

src/Sylius/Bundle/ResourceBundle/EventListener/ResourceDeleteSubscriber.php
```diff
@@ -42,18 +45,26 @@ class ResourceDeleteSubscriber implements EventSubscriberInterface
     private $translator;
 
     /**
+     * @var RestViewHandler
+     */
+    protected $viewHandler;
```
pjedrzejewski: This should be private and also an interface?

gperdomor:
@pjedrzejewski ready

pjedrzejewski:
Thanks @gperdomor!

PR #4915, opened 2016-04-29
[Travis] Improve caching

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

The cache from "application" suite was overwritten and saved as the same from "packages" suite. Joining them all may use more space, but it will be faster (~1min on "application" builds).


pjedrzejewski:
Thank you Kamil!

PR #4936, opened 2016-05-04
[Core][Product] Remove master variant

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets |
| License         | MIT

Due to upcoming changes in Sylius, it has been decided to remove concept of **master** variant and base on products with normal variants. For now, most of usage of **master** has been replaced with **first** variant, but it surely should be deeply considered to make it perfect.

To-do:
- [x] update UPGRADE file
- [x] introduce changes in documentation


nakashu:
:+1: 

Getting first variant is simple approach, but I would suggest making variants also sortable.
Then we would be able to change which one will be first, and thus which would be facing customers in the frontend.

There is also usecase (out of scope for now anyway)
- you have variants A, B, C
- you order them B, C, A - so in detail page user will see variant choice in this order
- you want variant C to be shown to customer in listings, eg you would need to keep the isMaster flag, but without duplication of variants as it is now.

Zales0123:
@nakashu sure, there are many things that can be improved in this case 👍  sorting variants is complex approach and I think can be implemented in the future. Also proper way, but much more simple could be selecting variant with the lowest price, as it's what customer should see to encourage him to buy a product 😄  I suppose there will be many discussions about this (as they should be).

pjedrzejewski:
@nakashu Yeah, we should definitely sort the variants and this could be implemented relatively easy with Sortable extension, but let's do it after this PR is merged. ;) Good points, thanks!

src/Sylius/Bundle/CoreBundle/Cart/ItemResolver.php
```diff
@@ -112,11 +112,6 @@ public function resolve(CartItemInterface $item, $data)
         $form = $this->formFactory->create('sylius_cart_item', $item, ['product' => $product]);
         $form->submit($data);
 
-        // If our product has no variants, we simply set the master variant of it.
-        if (null === $item->getVariant() && !$product->hasVariants()) {
-            $item->setVariant($product->getMasterVariant());
-        }
-
         if (null === $item->getVariant() && $product->hasVariants()) {
```
michalmarcinkowski: The `&& $product->hasVariants()` condition should be removed.

src/Sylius/Bundle/SearchBundle/Resources/views/SearchResultSnippets/product.html.twig
```diff
@@ -4,7 +4,13 @@
         <img class="img-rounded img-responsive"
              src="{{ result.image ? result.image.path|imagine_filter('sylius_medium') : 'http://placehold.it/262x255' }}"
              alt="{{ result.name }}"/>
-        <span class="label label-primary">{{ sylius_calculate_price(result.masterVariant)|sylius_price }}</span>
+        {% if result.variants|length == 1 and result.variants|length > 0 %}
```
michalmarcinkowski: `and result.variants|length > 0` - seems redundant.

src/Sylius/Bundle/WebBundle/Resources/views/Backend/Product/show.html.twig
```diff
@@ -134,13 +130,7 @@
...
                 <tr>
                     <td>{{ 'sylius.ui.tax_category'|trans }}</td>
-                    <td>{{ product.masterVariant.taxCategory|default('sylius.ui.no_tax_category'|trans) }}</td>
-                </tr>
-                <tr>
-                    <td>{{ 'sylius.ui.available_on_demand'|trans }}</td>
-                    <td>
-                        {{ misc.state_label(product.masterVariant.availableOnDemand) }}
-                    </td>
+                    <td>{{ product.firstVariant.taxCategory|default('sylius.ui.no_tax_category'|trans) }}</td>
```
michalmarcinkowski: That should not be displayed here.

src/Sylius/Behat/Context/Setup/OrderContext.php
```diff
@@ -210,7 +210,7 @@ public function theCustomerChoseShippingWithPayment(
      */
     public function theCustomerBoughtSingleProduct(ProductInterface $product)
     {
-        $this->addProductVariantToOrder($product->getMasterVariant(), $product->getPrice(), 1);
+        $this->addProductVariantToOrder($product->getFirstVariant(), $product->getFirstVariant()->getPrice(), 1);
```
lchrusciel: Won't it be better to leave `getPrice` method on product? We would have SSoT, and if anybody would decide that he need another logic for price displaying rather then this from the first variant, one could override it easily.

src/Sylius/Bundle/CoreBundle/EventListener/ImageUploadListener.php
```diff
@@ -39,16 +39,11 @@ public function __construct(ImageUploaderInterface $uploader)
     public function uploadProductImage(GenericEvent $event)
     {
         $subject = $event->getSubject();
-        if (!$subject instanceof ProductInterface && !$subject instanceof ProductVariantInterface) {
-            throw new UnexpectedTypeException(
-                $subject,
-                'Sylius\Component\Core\Model\ProductInterface or Sylius\Component\Core\Model\ProductVariantInterface'
-            );
+        if (!$subject instanceof ProductVariantInterface) {
```
lchrusciel: assert?

src/Sylius/Bundle/CoreBundle/Generator/ProductVariantGenerator.php
```diff
@@ -0,0 +1,38 @@
...
+namespace Sylius\Bundle\CoreBundle\Generator;
+
+use Sylius\Bundle\ProductBundle\Generator\VariantGenerator;
+use Sylius\Component\Core\Model\ProductVariantInterface;
+use Sylius\Component\Variation\Model\VariableInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class ProductVariantGenerator extends VariantGenerator
```
michalmarcinkowski: Is this class used anywhere? I don't think we need variant generator that sets random price.

src/Sylius/Bundle/FixturesBundle/DataFixtures/ORM/LoadProductsData.php
```diff
@@ -261,14 +262,13 @@ protected function generateVariants(ProductInterface $product)
...
      * @param ProductInterface $product
-     * @param string           $sku
+     * @param string $sku
      */
-    protected function addMasterVariant(ProductInterface $product, $sku = null)
+    protected function addVariant(ProductInterface $product, $sku = null)
     {
-        $variant = $product->getMasterVariant();
+        /** @var ProductVariantInterface $variant */
+        $variant = $product->getFirstVariant();
```
michalmarcinkowski: In `createProduct` I would create Product without Variant and here simply create variant and add it to the Product.
`$variant = $this->variantFactory->createNew();`

src/Sylius/Bundle/FixturesBundle/DataFixtures/ORM/LoadProductsData.php
```diff
@@ -368,13 +366,11 @@ protected function getUniqueISBN()
...
 
     /**
-     * Create new product instance.
-     *
      * @return ProductInterface
      */
     protected function createProduct()
     {
-        return $this->getProductFactory()->createNew();
+        return $this->getProductFactory()->createWithVariant();
```
michalmarcinkowski: See my comment above.

src/Sylius/Bundle/SearchBundle/Resources/views/SearchResultSnippets/product.html.twig
```diff
@@ -4,7 +4,13 @@
         <img class="img-rounded img-responsive"
              src="{{ result.image ? result.image.path|imagine_filter('sylius_medium') : 'http://placehold.it/262x255' }}"
              alt="{{ result.name }}"/>
-        <span class="label label-primary">{{ sylius_calculate_price(result.masterVariant)|sylius_price }}</span>
+        {% if result.variants|length == 1 %}
```
michalmarcinkowski: Why do we need this `if`? The `for` loop will work with all use cases.

src/Sylius/Bundle/VariationBundle/spec/Validator/VariantCombinationValidatorSpec.php
```diff
@@ -93,9 +80,9 @@ function it_should_add_violation_if_variant_with_given_same_options_already_exis
             'message' => 'Variant with given options already exists',
         ]);
 
-        $variant->isMaster()->willReturn(false);
         $variant->getObject()->willReturn($variable);
-        $variant->getOptions()->willReturn([$option]);
+
+        $variant->getOptions()->willReturn(new \ArrayIterator([$option->getWrappedObject()]));
```
michalmarcinkowski: `ArrayCollection`

src/Sylius/Component/Product/Model/Product.php
```diff
@@ -315,55 +316,26 @@ public function getAttributeByCode($attributeCode)
...
-        }
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function hasVariants()
     {
-        return !$this->getVariants()->isEmpty();
+        return 1 < $this->getVariants()->count();
```
michalmarcinkowski: Why do we need to change that?

src/Sylius/Bundle/CoreBundle/Generator/ProductVariantGenerator.php
Zales0123: Unfortunately we need for now, as price is required to persist variant ;/  However, after fixes from [this PR](https://github.com/Sylius/Sylius/pull/5059) it indeed won't be necessary ;)

app/migrations/Version20160425094419.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('DROP INDEX fulltext_search_idx ON sylius_search_index');
+        $this->addSql('CREATE INDEX fulltext_search_idx ON sylius_search_index (item_id)');
```
pamil: Is it related to this PR?

src/Sylius/Bundle/CoreBundle/Behat/CoreContext.php
```diff
@@ -551,6 +551,18 @@ public function productIsAvailableInAllVariations($productName)
     }
 
     /**
+     * @Given /^product "([^""]*)" has no variants$/
```
tuka217: `/^ product :productName` ?

app/migrations/Version20160425094419.php
Zales0123: Not at all, you're right 😄

pjedrzejewski:
Thank you Mateusz! Finally!!! :)

src/Sylius/Bundle/WebBundle/Resources/views/Backend/Order/show.html.twig
```diff
@@ -81,7 +81,8 @@
         <tr>
             <td class="picture text-center">
                 <a href="{{ path('sylius_backend_product_show', {'id': product.id}) }}" title="View {{ product.name|escape('html_attr') }} details.">
-                    <img src="{{ product.image ? product.image.path|imagine_filter('sylius_small') : 'http://placehold.it/90x60' }}" alt="" class="imgmedia-object" />
+                    {% set image = (product.variants|length > 0 and product.image) ? product.image.path|imagine_filter('sylius_medium') : 'http://placehold.it/90x60' %}
```
michalmarcinkowski: `(product.variants|length > 0 and product.image)` -> `product.image`

src/Sylius/Bundle/WebBundle/Resources/views/Backend/Product/macros.html.twig
```diff
@@ -34,7 +34,7 @@
                 <div class="media">
                     <a href="{{ path('sylius_backend_product_show', {'id': product.id}) }}" class="pull-left"
                        title="{{ 'sylius.ui.view_product_details'|trans({'%product%': product.name})|escape('html_attr') }}">
-                        <img src="{{ product.image ? product.image.path|imagine_filter('sylius_small') : 'http://placehold.it/50x40' }}" alt="" class="imgmedia-object" />
+                        <img src="{{ (product.variants|length > 0 and product.firstVariant.images.first()) ? product.firstVariant.images.first().path|imagine_filter('sylius_small') : 'http://placehold.it/50x40' }}" alt="" class="imgmedia-object" />
```
michalmarcinkowski: `product.image ? ...`

src/Sylius/Bundle/WebBundle/Resources/views/Frontend/Product/_single.html.twig
```diff
@@ -1,7 +1,8 @@
 <div class="row product">
     <div class="col-md-3">
         <a href="{{ path(product) }}" class="thumbnail">
-            <img src="{{ product.image ? product.image.path|imagine_filter('sylius_medium') : 'http://placehold.it/200x200' }}" alt="{{ product.name }}" />
+            {% set image = (product.variants|length > 0 and product.image) ? product.image.path|imagine_filter('sylius_medium') : 'http://placehold.it/200x200' %}
```
michalmarcinkowski: `product.image` is enough

PR #4940, opened 2016-05-04
Recompile container on configuration class modification

### LINKED ISSUE #4853: [DX] Recompiling container on configuration class change
**Steps to reproduce:**
 - Modify `DependencyInjection/Configuration` class in the bundle eg. change the default factory class

**Actual result:**
 - Container isn't recompiled in `dev` enviroment and the application still uses the old factory class

**Expected result:**
 - In `dev` environment container should be automatically rebuilt and the application should use new factory class

It can be technically solved by calling `$containerBuilder->addObjectResource()` or `$containerBuilder->addClassResource()` in the bundle extension.

pamil:
Fixes #4853, uses [`Symfony\Component\DependencyInjection\Extension\Extension::getConfiguration()`](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/DependencyInjection/Extension/Extension.php#L79).


pjedrzejewski:
Thank you Kamil!

PR #4942, opened 2016-05-04
Remove semantic coupling in metadata spec

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Resource/spec/Metadata/MetadataSpec.php
+++ b/src/Sylius/Component/Resource/spec/Metadata/MetadataSpec.php
@@ -12,9 +12,6 @@
 namespace spec\Sylius\Component\Resource\Metadata;
 
 use PhpSpec\ObjectBehavior;
-use Sylius\Bundle\ProductBundle\Form\Type\ProductType;
-use Sylius\Bundle\ResourceBundle\Form\Type\ResourceChoiceType;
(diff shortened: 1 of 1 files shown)


src/Sylius/Component/Resource/spec/Metadata/MetadataSpec.php
```diff
@@ -32,10 +29,10 @@ function let()
                 'driver' => 'doctrine/orm',
                 'templates' => 'SyliusProductBundle:Product',
                 'classes' => [
-                    'model' => Product::class,
+                    'model' => 'Sylius\Component\Product\Model\Product',
```
pamil: IMHO it is still coupled, we should avoid any reference to existing components. Can we use `\Example\Resource` instead here and something similar everything else?

lchrusciel: IMHO real example is much more readable. Should it be changed? ping @michalmarcinkowski @pjedrzejewski

pjedrzejewski: Yup.

pjedrzejewski:
Thanks Łukasz!

PR #4943, opened 2016-05-04
[Setup][Order] Fixed names of some steps in OrderContext

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Setup/OrderContext.php
+++ b/src/Sylius/Behat/Context/Setup/OrderContext.php
@@ -119,9 +119,9 @@ final class OrderContext implements Context
     }
 
     /**
-     * @Given the customer :customer placed an order :orderNumber
+     * @Given there is a customer :customer that placed an order :orderNumber
      */
-    public function theCustomerPlacedAnOrder(CustomerInterface $customer, $orderNumber)
+    public function thereIsCustomerThatPlacedOrder(CustomerInterface $customer, $orderNumber)
     {
         $order = $this->createOrder($customer, $orderNumber);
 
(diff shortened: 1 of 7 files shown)


src/Sylius/Behat/Context/Setup/OrderContext.php
```diff
@@ -119,9 +119,9 @@ public function __construct(
     }
 
     /**
-     * @Given the customer :customer placed an order :orderNumber
+     * @Given there is a customer :customer that placed an order :orderNumber
      */
-    public function theCustomerPlacedAnOrder(CustomerInterface $customer, $orderNumber)
+    public function aCustomerPlacedAnOrder(CustomerInterface $customer, $orderNumber)
```
pamil: Should we write articles in camelCased method names? `customerPlacedOrder` instead of `aCustomerPlacedAnOrder` and `customerBoughtSingleProduct` instead of `theCustomerBoughtASingleProduct` looks cleaner to me. Typing `$this->custom...` will show me all actions that customer may perform, without thinking if I should look at `aCustom...` or `theCustom...`.

Moreover, we do not know in the method if it is a new customer or the one defined earlier, because it is the responsibility of parameters transformers to provide us customer instance in any case. We just get `customer`, nothing more or less.

src/Sylius/Behat/Context/Setup/OrderContext.php
```diff
@@ -119,9 +119,9 @@ public function __construct(
     }
 
     /**
-     * @Given the customer :customer placed an order :orderNumber
+     * @Given there is a customer :customer that placed an order :orderNumber
```
michalmarcinkowski: `there is a customer` - suggests that we will create a customer, not that this customer already exist. IMO the previous version was fine.

pjedrzejewski:
Thanks Grzesiu!

PR #4957, opened 2016-05-06
[Channel] Cache channel not found exceptions in CachedPerRequestChannelContext

pamil:
It caused ~200-300 database requests sent while being in store administration panel with undefined channel.


pjedrzejewski:
Thank you Kamil!

PR #4962, opened 2016-05-06
[Settings] Make schema & resolver registries lazy

pamil:
Schema or resolvers depending on the Doctrine repositories throws an exception about unexisting database during cache warmup.


michalmarcinkowski:
Thank you Kamil! 👍

PR #4967, opened 2016-05-06
[Product] fix ProductController::searchAction

lighta971:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | --
| License         | MIT

Exception when product has no image

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Controller/ProductController.php
+++ b/src/Sylius/Bundle/CoreBundle/Controller/ProductController.php
@@ -275,7 +275,7 @@ class ProductController extends ResourceController
             $results[] = [
                 'id' => $product->getMasterVariant()->getId(),
                 'name' => $product->getName(),
-                'image' => $product->getImage()->getPath(),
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/CoreBundle/Controller/ProductController.php
```diff
@@ -275,7 +275,7 @@ public function searchAction(Request $request)
             $results[] = [
                 'id' => $product->getMasterVariant()->getId(),
```
lighta971: Why not `$product->getId()` instead ?

pjedrzejewski:
Thanks Aurel! 👍

src/Sylius/Bundle/CoreBundle/Controller/ProductController.php
pjedrzejewski: Not sure to be honest, I believe there was some reason, but can't remember now.

PR #4970, opened 2016-05-06
[Performance] Prevent fetch locales from db every time when get available.

liverbool:
| Q | A |
| --- | --- |
| Bug Fix? | n |
| New Feature? | n |
| BC Breaks? | n |
| Deprecations? | n |
| Tests Pass? | n |
| Fixed Tickets |  |
| License | MIT |
| Doc PR |  |


stefandoorn:
How much performance gain does this give? It would be great if Sylius has some sort of `cache.app` or `cache.framework` implementation out of the box for these things. WDYT @pamil?

PR #4972, opened 2016-05-07
[GridBundle] Explicitly enable drivers

dantleech:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | #4809

As discussed in PR #4812.

This PR adds an `enabled_drivers` configuration key to the GridBundle.

The drivers are organized in folders in `Resources/config` as:

```
config/
    driver/
        doctrine/
            orm.xml
            phpcr-odm.xml
```

Which means we can use the actual value for the driver name (`doctrine/orm`) to locate the configuration file.

In addition the `Configuration` class will scan the `driver` directory to determine the valid values for the configuration key.


pjedrzejewski:
@dantleech Nice, you just need to update config tests. :)

src/Sylius/Bundle/GridBundle/DependencyInjection/Configuration.php
```diff
@@ -29,12 +29,41 @@ public function getConfigTreeBuilder()
...
+    {
+        // determine which drivers are distributed with this bundle
+        $driverDir = __DIR__ . '/../Resources/config/driver';
+        $iterator = new \RecursiveDirectoryIterator($driverDir);
+        foreach (new \RecursiveIteratorIterator($iterator) as $file) {
+            if ($file->getExtension() !== 'xml') {
+                continue;
+            }
+
+            $validDrivers[] = substr($file->getPathname(), 1 + strlen($driverDir), -4);
```
michalmarcinkowski: `substr($file->getFilename(), 0, strrpos($file->getFilename(), '.'))`

src/Sylius/Bundle/GridBundle/DependencyInjection/SyliusGridExtension.php
```diff
@@ -46,5 +45,10 @@ public function load(array $config, ContainerBuilder $container)
 
         $container->setAlias('sylius.grid.renderer', 'sylius.grid.renderer.twig');
         $container->setAlias('sylius.grid.data_extractor', 'sylius.grid.data_extractor.property_access');
+
+        foreach ($config['enabled_drivers'] as $enabledDriver) {
+            $path = 'driver/' . $enabledDriver . '.xml';
```
michalmarcinkowski: `$path = sprintf('driver/%s.xml', $enabledDriver);`

pamil: `$loader->load(sprintf('driver/%s.xml', $enabledDriver));`?

src/Sylius/Bundle/GridBundle/DependencyInjection/Configuration.php
```diff
@@ -29,12 +29,41 @@ public function getConfigTreeBuilder()
...
+        }
+
+        $node
+            ->children()
+                ->arrayNode('enabled_drivers')
```
pamil: What do you think about renaming it to `drivers` instead? 

IMO this:
```yaml
sylius_grid:
  drivers:
    - doctrine/orm
```

looks better than this:

```yaml
sylius_grid:
  enabled_drivers:
    - doctrine/orm
```
And it is understandable easily as well

dantleech: Well, I think the issue here is that we are just "turning on" drivers
which are distributed with the bundle. If the user has their own drivers
then they may think to add those here too, which would break things.

`drivers` implies all drivers, including those which are not distributed
with the bundle.

`enabled_drivers` is slightly more specific imo, not by much.

Ultimately I like the idea of the `driver provider` pattern better, in
which case all drivers would be turned on but lazily instantiated. But
this require more work.

But in general I don't really mind which term we use here.

dantleech:
Updated and added some tests, also added the mattiasnoback dependencies to `composer.json` and added the autoloader to `phpunit.xml.dist`.

dantleech:
Still an open question about if to use `drivers` or `enabled_drviers` or a better term.

pjedrzejewski:
I'd go with ``drivers``! We can still iterate on this for a while, but let's merge it. :) Great work @dantleech, thank you!

dantleech:
Updated

pjedrzejewski:
Nice, thanks a lot Dan!

PR #4982, opened 2016-05-09
[PayumBundle] Prevent hard-dependency error to CoreBundle

aramalipoor:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Related tickets | - |
| License | MIT |


pjedrzejewski:
I think SyliusPayumBundle should depend on Core. Are you using it without the core? :D

pamil:
PayumBundle requires `sylius/core` but should require `sylius/core-bundle`.

aramalipoor:
:D Yeah we're creating a very simple app and need some stuff in PayumBundle, so do you think moving them in our own AppBundle is OK?

pamil:
I would just define own `sylius.invoice_number_generator` service then (interface is placed in Core component) or override the service that uses it.

pjedrzejewski:
@aramalipoor Yeah and if it is generic and not related to Sylius, then I'd consider contributing it to PayumBundle. :)

PR #4987, opened 2016-05-09
[ResourceBundle] Use default manager as configured by Doctrine

dantleech:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

Currently the ResourceBundle uses the entity manager named "default" by default - this is wrong as the "default" entity manager can be called anything and it
is only "default" by default.

This PR changes the logic to use the alias of the default manager (e.g. `doctrine.entity_manager`) 
when no object manager name has been explicitly configured.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/Doctrine/AbstractDoctrineDriver.php
+++ b/src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/Doctrine/AbstractDoctrineDriver.php
@@ -53,13 +53,16 @@ abstract class AbstractDoctrineDriver extends AbstractDriver
     }
 
     /**
+     * Return the configured object managre name, or NULL if the default
+     * manager should be used.
+     *
      * @param MetadataInterface $metadata
      *
-     * @return string
+     * @return string|null
      */
     protected function getObjectManagerName(MetadataInterface $metadata)
     {
-        $objectManagerName = 'default';
+        $objectManagerName = null;
 
         if ($metadata->hasParameter('options') && isset($metadata->getParameter('options')['object_manager'])) {
             $objectManagerName = $metadata->getParameter('options')['object_manager'];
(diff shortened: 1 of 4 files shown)


pjedrzejewski:
👍

pjedrzejewski:
Thank you @dantleech!

bendavies:
@dantleech This doesn't work and has broken my custom repositoryies, when only one entity manager is defined. Investigating...

bendavies:
Yeah, this doesn't work.
You need to use the `%doctrine.default_entity_manager%` parameter for the default manager name, which Doctrine bundle defaults to 'default'.
https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/Configuration.php#L80

using the alias `doctrine.orm.entity_manager` doesn't seem to work.

Suggest this is reverted for now.

dantleech:
I wonder why the alias doesn't work?

https://github.com/doctrine/DoctrineBundle/blob/master/DependencyInjection/DoctrineExtension.php#L355

But yeah, just using the actual parameter would probably be better.

bendavies:
@dantleech ok, the issue occurs when retrieving a repository from the manager, because the manager will use the 'default' manager: https://github.com/doctrine/common/blob/master/lib/Doctrine/Common/Persistence/AbstractManagerRegistry.php#L168-L184

dantleech:
But the default manager is the same as `doctrine.orm.entity_manager` no?

bendavies:
sure is, but it doesn't work 🎱 
I'll dig a bit more.

bendavies:
@dantleech i must apologise, i think i'm totally wrong and there's nothing wrong with this! haah

bendavies:
@dantleech i accidentally introduced a bug in https://github.com/Sylius/Sylius/pull/5047 which i didn't pick up on. Apologise again!

dantleech:
No worries :)

PR #4988, opened 2016-05-09
Array accessible and countable Service Registry.

adamelso:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    |yes
| BC breaks?      | no
| Related tickets |
| License         | MIT

The idea behind this is to make it almost effortless to replace arrays
of services with a service registry.  The implementations of SPL's
`ArrayAccess` are called from the implementations of the
`RegistryInterface`, the only difference being that the `ArrayAccess`
implementations do not throw exceptions as to align the behavior with
array usage in PHP.

The countable interface allows checking how many items are in the
registry without having to get all of them.

### DIFF
--- a/src/Sylius/Component/Registry/ServiceRegistry.php
+++ b/src/Sylius/Component/Registry/ServiceRegistry.php
@@ -13,8 +13,9 @@ namespace Sylius\Component\Registry;
 
 /**
  * @author Paweł Jędrzejewski <pawel@sylius.org>
+ * @author Adam Elsodaney <adam.elso@gmail.com>
  */
-class ServiceRegistry implements ServiceRegistryInterface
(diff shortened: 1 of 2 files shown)


src/Sylius/Component/Registry/ServiceRegistry.php
```diff
@@ -72,38 +73,78 @@ public function register($identifier, $service)
...
+    {
         return $this->services[$identifier];
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function offsetSet($identifier, $service)
+    {
+        $this->services[$identifier] = $service;
```
pamil: I'm not sure whether we should allow replacing services.

adamelso:
ArrayAccess actually made using the registry more confusing given the current API (but make sense for something like Pimple.) Closing.

PR #4990, opened 2016-05-10
[Composer] Bump symfony/symfony to ^2.7.9 due to a security issue (CVE-2016-1902)

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

http://symfony.com/blog/cve-2016-1902-securerandom-s-fallback-not-secure-when-openssl-fails


src/Sylius/Bundle/AddressingBundle/Form/Type/CountryCodeChoiceType.php
```diff
@@ -61,10 +87,11 @@ private function getCountryCodes(array $countries)
 
         /* @var CountryInterface $country */
         foreach ($countries as $country) {
-            $countryCodes[$country->getCode()] = Intl::getRegionBundle()->getCountryName($country->getCode());
+            $countryName = Intl::getRegionBundle()->getCountryName($country->getCode());
+            $countryCodes[$countryName] = $country->getCode();
```
michalmarcinkowski: Why do we need to swap country name with code?

pamil: It has started to fail since I've updated from 2.7.7 to 2.7.9 and `choice_as_values` is deprecated and defaults to `true` in Symfony >=3.0.

michalmarcinkowski: So in Symfony >=3.0 you must use array keys as values and there is no option to reverse that?

pamil: [It looks so](http://symfony.com/doc/current/reference/forms/types/choice.html#choices-as-values), but you can use `choice_label` / `choice_value` options to manipulate the choices.

michalmarcinkowski: Not a fan of this approach, but we can do nothing about that 😞

pjedrzejewski:
Thanks Kamil!

PR #4998, opened 2016-05-11
Do no translate choice form type values that do not need it

### LINKED ISSUE #4995: [Forms] - Disable choice options translations for performance
For all choice options that are already translated and/or doesn't need to be translated the translation domain on form type should be disabled.

```php
{
    $resolver->setDefaults([
        'choice_translation_domain' => false,
    ])
}
```

In dev environment with around 500 options in a select box the performance increase is around 1 second at times.

reference:
http://symfony.com/blog/new-in-symfony-2-7-form-and-validator-updates#added-choice-translation-domain-domain-to-avoid-translating-options
http://symfony.com/doc/current/reference/forms/types/choice.html#choice-translation-domain

Not complete list of choice types
* taxon
* option
* attribute
* country
* zone members

peteward:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #4995 
| License         | MIT

Disable choice_translation_domain on choice form types that do not need translation (e.g. entity names that are translated at source)

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
+++ b/src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
@@ -52,6 +52,7 @@ class CountryChoiceType extends AbstractType
 
         $resolver
             ->setDefaults([
+                'choice_translation_domain' => false,
                 'choice_list' => $choices,
                 'enabled' => true,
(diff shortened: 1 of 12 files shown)


pjedrzejewski:
Great, thank you! 👍

PR #4999, opened 2016-05-11
[Legacy Admin] Fix login assetic dependencies

peteward:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

Remove CDN CSS dependencies from assetic filter, include as stylesheets in their own right


pjedrzejewski:
Thank you Peter!

Issue #4791, opened 2016-04-14
[Cart] - saving cart will change persist, product variant options - if variant field is present


nakashu:
**usecase:**
- customer should be able to change product variant inside cart summary
- You have t-shirts with selection method "variant match"
- options "color"(blue, red) and "size"(S, XL)
- customer adds product variant to cart with options: `color => blue`, `size => S`
- when you extend the cart summary with with variant field - same as in product detail page.
  change options to `color => red`, `size => XL` and save

**problem**
- the original `ProductVariant` will now have new options, in the process creating a duplicate in the database and keep its original price
- So in a way you can add cheap product to cart, change it to expensive variant and go forward with the order.

**expected behaviour**
- After changing options in the cart and saving, a new matching for `ProductVariant` will occur and replace the reference on the `CartItem object` without touching the `ProductVariant` object
- Also without changing the CartItem ID - eg. removing it and adding again to cart.

It seems that some that there is a referenced OptionsCollection or ProductVariant somewhere in the code and It get updated, instead of just being transformed. in the VariantMatchType.
If anyone got experience with similar use case please share how you solved it.

**code to reproduce**

``` php
// Sylius\Bundle\CoreBundle\Form\Type\CartItemType.php
public function buildForm(FormBuilderInterface $builder, array $options)
    {
        parent::buildForm($builder, $options);

        if (isset($options['product']) && $options['product']->hasVariants()) {
            $type = Product::VARIANT_SELECTION_CHOICE === $options['product']->getVariantSelectionMethod() ? 'sylius_product_variant_choice' : 'sylius_product_variant_match';

            $builder->add('variant', $type, [
                'variable' => $options['product'],
            ]);
        }
// ADDED part - for it to be available in Cart summary page.
        $builder->addEventListener(FormEvents::POST_SET_DATA, function(FormEvent $event) use ($options) {

            $builder = $event->getForm();
            if (!$event->getData()->getVariant()) {
                return;
            }

            $product = $event->getData()->getVariant()->getProduct();

            if (isset($product) && $product->hasVariants()) {
                $type = Product::VARIANT_SELECTION_CHOICE === $product->getVariantSelectionMethod() ? 'sylius_product_variant_choice' : 'sylius_product_variant_match';

                $builder->add('variant', $type, [
                    'variable' => $product,
                ]);
            }
        });
    }
```

``` twig
{# WebBundle/Resources/view/Frontend/Cart/_item.html.twig #}
...
... current code 
...
<tr>
[...]

PR #5009, opened 2016-05-12
[Variation] - return transformed options as array to avoid side-effects

nakashu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | partially #4791
| License         | MIT

As described in #4791 - when changes are made on `Option` returned from the transformer, it can have side effects - eg. `Variant` option values will change upon persist

Returning options as array will, remove reference of the persistent collection and no side effects will occur.


pjedrzejewski:
Nice! I have restarted one build - it failed for no reason. :)

michalmarcinkowski:
Thanks you @nakashu! 👍

PR #4496, opened 2016-03-14
[CoreBundle] Change enabled = 1 in ProductRepository to true

Niiko:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT
| Doc PR          | -

I changed condition from value "1" (integer) to boolean (true).

Indeed, in mysql, there is no problem, but with more strict sgbd (as postgresql), that need to be casted.

PR #5011, opened 2016-05-12
[Promotion] implement `ToggleableTrait`, fixes #3871

### LINKED ISSUE #2739: [Promotion] Applied to orders after adding

rvanlaak:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Related tickets | fixes #2739 , #3098 |
| License | MIT |

Implement `ToggleableTrait` on `Promotion`, also see https://github.com/Sylius/Sylius/pull/3871#issuecomment-179804903 of @pjedrzejewski

### DIFF
--- a/features/promotion/managing_coupons/coupon_generate_validation.feature
+++ b/features/promotion/managing_coupons/coupon_generate_validation.feature
@@ -8,6 +8,7 @@ Feature: Coupon generate instruction validation
         Given the store operates on a single channel in "France"
         And there is a promotion "Christmas sale"
         And it is coupon based promotion
+        And it is enabled promotion
         And I am logged in as an administrator
 
     @ui
--- a/src/Sylius/Bundle/PromotionBundle/Doctrine/ORM/PromotionRepository.php
+++ b/src/Sylius/Bundle/PromotionBundle/Doctrine/ORM/PromotionRepository.php
@@ -80,7 +80,7 @@ class PromotionRepository extends EntityRepository implements PromotionRepositor
                     $queryBuilder->expr()->gt($this->getPropertyName('endsAt'), ':date')
                 )
             )
-            ->andWhere($this->getPropertyName('enabled') . ' = 1')
+            ->andWhere($this->getPropertyName('enabled') . ' = true')
             ->setParameter('date', $date)
         ;
     }
--- a/src/Sylius/Bundle/PromotionBundle/Doctrine/ORM/PromotionRepository.php
+++ b/src/Sylius/Bundle/PromotionBundle/Doctrine/ORM/PromotionRepository.php
@@ -80,6 +80,7 @@ class PromotionRepository extends EntityRepository implements PromotionRepositor
                     $queryBuilder->expr()->gt($this->getPropertyName('endsAt'), ':date')
                 )
             )
+            ->andWhere($this->getPropertyName('enabled') . ' = 1')
             ->setParameter('date', $date)
         ;
     }
--- a/features/promotion/managing_coupons/adding_coupon.feature
+++ b/features/promotion/managing_coupons/adding_coupon.feature
@@ -8,6 +8,7 @@ Feature: Adding a new coupon
         Given the store operates on a single channel in "France"
         And there is a promotion "Christmas sale"
         And it is coupon based promotion
+        And it is enabled promotion
         And I am logged in as an administrator
 
     @ui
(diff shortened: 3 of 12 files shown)


pjedrzejewski:
Do you think you could add a Behat scenario for that? Would require only adding a new step saying if promotion is enabled and rest should be just matter of using existing steps.

rvanlaak:
@pjedrzejewski added another commit for the Behat scenarios. Given this was my first Behat experience, please let me know if it makes any sense.

src/Sylius/Bundle/PromotionBundle/Doctrine/ORM/PromotionRepository.php
```diff
@@ -80,6 +80,7 @@ protected function filterByActive(QueryBuilder $queryBuilder, \Datetime $date =
                     $queryBuilder->expr()->gt($this->getPropertyName('endsAt'), ':date')
                 )
             )
+            ->andWhere($this->getPropertyName('enabled') . ' = 1')
```
lchrusciel: `'o.enabled = 1'`

features/promotion/managing_coupons/coupon_generate_validation.feature
```diff
@@ -8,6 +8,7 @@ Feature: Coupon generate instruction validation
         Given the store operates on a single channel in "France"
         And there is a promotion "Christmas sale"
         And it is coupon based promotion
+        And it is enabled promotion
```
lchrusciel: Promotion is enabled by default, so there is no point in writing it explicit in every scenario. This steps should be removed.

lchrusciel:
You have missed migration file.

And behats. Can you write a new scenarios? Your PR introduce a new feature, so you should add the 'toggling_promotion.feature'. You can check [currency toggling](https://github.com/Sylius/Sylius/blob/master/features/currency/managing_currency/toggling_currency.feature) as a sample implementation. Also, you can find [this section](http://docs.sylius.org/en/latest/behat/index.html?highlight=behat) of our documentation useful.

pamil:
@rvanlaak it would be great if you share your experiences with creating Behat scenarios (which are *quite* customized in Sylius) afterwards, I'm very interested in how the first impression looks like :)

src/Sylius/Bundle/PromotionBundle/Doctrine/ORM/PromotionRepository.php
rvanlaak: That was my first idea too, until I saw that all other fields use `getPropertyName` too.

rvanlaak:
My first experience with Behat went like this:

> Okay, I need to add a Behat scenario, lets see how far the existing Sylius scenarios bring me. I've heard about Behat before and know about BDD, can't differ that much from PHPUnit tests. But hey wait, I have no idea how to determine what a proper Behat scenario _should_ look like. Does it has similar rules like PHPUnit has with the "you at least need two assertions for each test"? Lets dive into the test suite a bit more, and find a parameter similar to `enabled` that has tests too. So, then once again I found out that PHPStorm is awesome because it allows you to CTRL+click Behat steps and points you to the right context with the related annotation directly. This clarified a lot to me, so based on that I created the extra steps and some sort of "validated" them by assuming that if they are clickable in PHPStorm that would mean that the added Behat step would work. The last thing worth mentioning is that it takes Travis quite some time to complete the Behat test suite, so maybe I need to make a local test environment for Behat the next time too.

Niiko:
Referring to #4496 it would be better to set enabled = true instead of enabled = 1

rvanlaak:
@Niiko I've changed it to `enabled = true`. :+1: Can someone else fix the Behat scenarios, I'm not that known to it and next to that do not really understand what you guys expect.

PR #3980, opened 2016-01-26
[Review] SyliusReviewBundle

GSadee:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 
| License       | MIT
| Doc PR        | 

Initial implementation of SyliusReviewBundle, which allows to easily manage review system. It's based on @Zales0123 #3300.

"Reviewable" and "Reviewer" classes can be configured under ``sylius_review`` tag:
```
sylius_review:
    resources:
        product:
            subject: %sylius.model.product.class%
            review:
                classes:
                    model: Sylius\Component\Core\Model\ProductReview
                    form:
                        default: Sylius\Bundle\CoreBundle\Form\Type\ProductReviewType
                        admin: Sylius\Bundle\CoreBundle\Form\Type\ProductReviewAdminType
            reviewer:
                classes:
                    model: Sylius\Component\Core\Model\Customer
```

I have added changes from #3372 and extended this PR to use CustomerGuestType in ReviewType.

PR #5012, opened 2016-05-12
[WIP] SyliusLikeBundle - User can like or dislike a resource

loicmobizel:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | yes |
| Related tickets |  |
| License | MIT |

This is an initial implementation of SyliusLikeBundle, which allows to easily manage like or dislike system on any entity. It's based on #3980.

I suggest my contribution to this new feature. It's a work in progress feature :)


loicmobizel:
@pjedrzejewski Is this feature can be useful for sylius ?

pjedrzejewski:
Hi @loicmobizel, I really appreciate the effort put into this bundle, but I think it is out of scope for Sylius core. As with every feature, we need to maintain it and Sylius already has plenty of bundles. That being said, I think you could definitely move this bundle to your own namespace and make it a plugin to Sylius or a generic bundle with optional integration to Sylius. You still can use other Sylius bundles (like resource or grids) as dependencies! 👍  Have a look at https://github.com/webburza/sylius-article-bundle

loicmobizel:
@pjedrzejewski Ok, I really understand you have enough bundles for Sylius Core to maintain. I agree to make this feature as a sylius plugin. I need this feature for a project anyway :)

PR #5016, opened 2016-05-12
[Doc][Behat] Add basic usage for behat guide

Arminek:
Add some example usage of our new behat. feat @TheMadeleine

### DIFF
--- a/docs/behat/basic-usage.rst
+++ b/docs/behat/basic-usage.rst
@@ -1,0 +1,640 @@
+Basic Usage
+===========
+
+The best way of understanding how things work in detail is showing and analyzing examples, that is why this section gathers all the knowledge from the previous chapters.
+Let's assume that we are going to implement the functionality of managing countries in our system.
+Now let us show you the flow.
+
+Describing features
+-------------------
+Let's start with writing our feature file, which will contain answers to the most important questions: Why (benefit), who (actor using the feature) and what (the feature itself). It should also include scenarios, which serve as examples of how things supposed to work.
+``features/addressing/managing_countries/adding_country.feature``
+Scenario::
+
+    @managing_countries
+    Feature: Adding a new country
+        In order to sell my goods to different countries
+        As an Administrator
+        I want to add a new country to the store
+
+        Background:
+            Given I am logged in as an administrator
+
+        @ui
+        Scenario: Adding country
+            Given I want to add a new country
+            When I choose "France"
+            And I add it
+            Then I should be notified that it has been successfully created
+            And the country "France" should appear in the store
+
+Pay attention to the form of these sentences. From the developer point of view they are hiding the details of the feature's implementation. Instead of describing "When I click on the select box And I choose France from the dropdown Then I should see the France country in the table" - we are using sentences that are less connected with the implementation, but more focused on the effects of our actions.
+A side effect of such approach is that it results in steps being really generic, therefore if we want to add another way of testing this feature for instance in the domain or api context, it will be extremely easy to apply. We just need to add a different tag (in this case "@domain") and of course implement the proper steps in the domain context of our system.
+To be more descriptive let's imagine that we want to check if a country is added properly in two ways. First we are checking if the adding works via frontend, so we are implementing steps that are clicking, opening pages, filling fields on forms and similar, but also we want to check this action regardlessly of the frontend, for that we need the domain, which allows us to perform actions only on objects.
+
+Choosing a correct suite
+------------------------
+After we are done with a feature file, we have to create a new suite for it. At the beginning we have decided that it will be a frontend/user interface feature, that is why we are placing it in "etc/behat/suites/ui/addressing/managing_countries.yml".
+
+.. code-block:: yaml
+
+    default:
+        suites:
+            ui_managing_countries:
+                contexts_as_services:
+                    - sylius.behat.context.hook.doctrine_orm
+                    # This service is responsible for clearing database before each scenario,
+                    # so that only data from the current and its background is available.
+
+                    - sylius.behat.context.transform.country
+                    - sylius.behat.context.transform.shared_storage
+                    # The transformer contexts services are responsible for all the transformations of data in steps:
+                    # For instance "And the country "France" should appear in the store" transforms "(the country "France")" to a proper Country object, which is from now on available in the scope of the step.
+
+                    - sylius.behat.context.setup.geographical
+                    - sylius.behat.context.setup.security
+                    # The setup contexts here are preparing the background, adding available countries and users or administrators.
+                    # These contexts have steps like "I am logged in as an administrator" already implemented.
+
+                    # Lights, Camera, Action!
+                    - sylius.behat.context.ui.admin.managing_countries
+                    - sylius.behat.context.ui.admin.notification
+                    # Those contexts are essential here we are placing all action steps like "When I choose "France" and I add it Then I should ne notified that...".
+                filters:
+                    tags: "@managing_countries && @ui"
+
+A very important thing that is done here is the configuration of tags, from now on Behat will be searching for all your features tagged with ``@managing_countires`` and your scenarios tagged with ``@ui``.
+Second thing is ``contexts_as_services:`` in this section we will be placing all our services with step implementation.
+
+We have mentioned with the generic steps we can easily switch our testing context to @domain. Have a look how it looks:
+
+.. code-block:: yaml
+
+    default:
+        suites:
+            domain_managing_countries:
+                contexts_as_services:
+                    - sylius.behat.context.hook.doctrine_orm
+
+                    - sylius.behat.context.transform.country
+                    - sylius.behat.context.transform.shared_storage
+
+                    - sylius.behat.context.setup.geographical
+                    - sylius.behat.context.setup.security
+
+                    - sylius.behat.context.domain.admin.managing_countries # Domain step implementation.
+                filters:
+                    tags: "@managing_countries && @domain"
+
+We are almost finished with the suite configuration.
+Now we need to register our first Behat context as a service, but beforehand we need
+
+Registering Pages
+-----------------
+The page object approach allows us to hide all the detailed interaction with ui (html, javascript, css) inside.
+
+We have three kinds of pages:
+    - Page - First layer of our pages it knows how to interact with DOM objects. It has a method ``->getUrl(array $urlParameters)`` where you can define a raw url to open it.
+    - SymfonyPage - This page extends the Page. It has a router injected so that the ``->getUrl()`` method generates a url from the route name which it gets from the ``->getRouteName()`` method.
+    - Base Crud Pages (IndexPage, CreatePage, UpdatePage) - These pages extend SymfonyPage and they are specific to the Sylius resources. They have a resource name injected and therefore they know about the route name.
+
+There are two ways to manipulate UI - by using ``->getDocument()`` or ``->getElement('your_element')``.
+First method will return a ``DocumentElement`` which represents an html structure of the currently opened page,
+second one is a bit more tricky because it uses the ``->getDefinedElements()`` method and it will return a ``NodeElement`` which represents only the restricted html structure.
+
+Usage example of ``->getElement('your_element')`` and ``->getDefinedElements`` methods.
+
+.. code-block:: php
+
+    <?php
+
+    class CreatePage extends SymfonyPage implements CreatePageInterface
+    {
+        // This method returns a simple associative array, where the key is the name of your element and the value is its locator.
+        protected function getDefinedElements()
+        {
+            return array_merge(parent::getDefinedElements(), [
+                'provinces' => '#sylius_country_provinces',
+            ]);
+        }
+
+        // By default it will assume that your locator is css.
+        // Example with xpath.
+        protected function getDefinedElements()
+        {
+            return array_merge(parent::getDefinedElements(), [
+                'provinces' => ['xpath' => '//*[contains(@class, "provinces")]'] // Now your value is an array where key is your locator type.
+            ]);
+        }
+
+        // Like that you can easily manipulate your page elements.
+        public function addProvince(ProvinceInterface $province)
+        {
+            $provinceSelectBox = $this->getElement('provinces');
+
+            $provinceSelectBox->selectOption($province->getName());
+        }
+    }
+
+Let's get back to our main example and analyze our scenario.
+We have steps like
+"When I choose "France"
+And I add it
+Then I should be notified that it has been successfully created
+And the country "France" should appear in the store".
+
+.. code-block:: php
+
+    <?php
+
+    namespace Sylius\Behat\Page\Admin\Country;
+
+    use Sylius\Behat\Page\Admin\Crud\CreatePage as BaseCreatePage;
+
+    class CreatePage extends BaseCreatePage implements CreatePageInterface
+    {
+        /**
+         * @param string $name
+         */
+        public function chooseName($name)
+        {
+            $this->getDocument()->selectFieldOption('Name', $name);
+        }
+
+        public function create()
+        {
+            $this->getDocument()->pressButton('Create');
+        }
+    }
+
+.. code-block:: php
+
+    <? php
+
+    namespace Sylius\Behat\Page\Admin\Country;
+
+    use Sylius\Behat\Page\Admin\Crud\IndexPage as BaseIndexPage;
+
+    class IndexPage extends BaseIndexPage implements IndexPageInterface
+    {
+        /**
+         * @return bool
+         */
+        public function isSingleResourceOnPage(array $parameters)
+        {
+            try {
+                $rows = $this->tableAccessor->getRowsWithFields($this->getElement('table'), $parameters);
+                // Table accessor is a helper service which is responsible for all html table operations.
+
+                return 1 === count($rows);
+            } catch (ElementNotFoundException $exception) { // Table accessor throws this exception when cannot find table element on page.
+                return false;
+            }
+        }
+    }
+
+.. warning::
+
+    There is one small gap in this concept - PageObjects is not a concrete instance of the currently opened page, they only mimic its behaviour (dummy pages).
+    This gap will be more understandable on the below code example.
+
+.. code-block:: php
+
+    <?php
+
+    // Of course this is only to illustrate this gap.
+
+    class HomePage
+    {
+        // In this context on home page sidebar you have for example weather information in selected countries.
+        public function readWeather()
+        {
+            return $this->getElement('sidebar')->getText();
+        }
+
+        protected function getDefinedElements()
+        {
+            return ['sidebar' => ['css' => '.sidebar']]
+        }
+
+        protected function getUrl()
+        {
+            return 'http://your_domain.com';
+        }
+    }
+
+    class LeagueIndexPage
+    {
+        // In this context you have for example football match results.
+        public function readMatchResults()
+        {
+            return $this->getElement('sidebar')->getText();
+        }
+
+        protected function getDefinedElements()
+        {
+            return ['sidebar' => ['css' => '.sidebar']]
+        }
+
+        protected function getUrl()
+        {
+            return 'http://your_domain.com/leagues/'
+        }
+    }
+
+    final class GapContext implements Context
+    {
+        private $homePage;
+        private $leagueIndexPage;
+
+        /**
+         * @Given I want to be on Homepage
+         */
+        public function iWantToBeOnHomePage() // After this method call we will be on "http://your_domain.com".
+        {
+            $this->homePage->open(); //When we add @javascript tag we can actually see this thanks to selenium.
+        }
+
+        /**
+         * @Then I want to see the sidebar and get information about the weather in France
+         */
+        public function iWantToReadSideBarOnHomePage($someInformation) // Still "http://your_domain.com".
+        {
+            $someInformation === $this->leagueIndexPage->readMatchResults() // This returns true, but wait a second we are on home page (dummy pages).
+
+            $someInformation === $this->homePage->readWeather() // This also returns true.
+        }
+    }
+
+Registering contexts
+--------------------
+As it was shown in the previous section we have registered a lot of contexts, so we will show you only some of the steps implementation.
+
+Scenario::
+
+    Given I want to add a new country
+    And I add it
+    Then I should be notified that it has been successfully created
+    And the country "France" should appear in the store
+
+Let's start with essential one ManagingCountriesContext
+
+Ui contexts
+~~~~~~~~~~~
+
+.. code-block:: php
+
+    <?php
+
+    namespace Sylius\Behat\Context\Ui\Admin
+
+    final class ManagingCountriesContext implements Context
+    {
+        /**
+         * @var IndexPageInterface
+         */
+        private $indexPage;
+
+        /**
+         * @var CreatePageInterface
+         */
+        private $createPage;
+
+        /**
+         * @var UpdatePageInterface
+         */
+        private $updatePage;
+
+        /**
+         * @param IndexPageInterface $indexPage
+         * @param CreatePageInterface $createPage
+         * @param UpdatePageInterface $updatePage
+         */
+        public function __construct(
+            IndexPageInterface $indexPage,
+            CreatePageInterface $createPage,
+            UpdatePageInterface $updatePage
+        ) {
+            $this->indexPage = $indexPage;
+            $this->createPage = $createPage;
+            $this->updatePage = $updatePage;
+        }
+
+        /**
+         * @Given I want to add a new country
+         */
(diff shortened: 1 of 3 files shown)


docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,513 @@
+Basic Usage
+===========
+
+In our opinion the best way to understand how some stuff works is by showing some examples so this section gather all knowledge form previous chapters.
+Assume that we are implementing managing countries in our system.
+Let the fun begin.
```
TheMadeleine: The best way of understanding how things work in detail is showing and analyzing examples, that is why this section gathers all the knowledge from the previous chapters. 
Let's assume that we are going to implement the functionality of managing countries in our system. Now let us show you the flow.

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,513 @@
...
+
+Describing feature
+------------------
+First of all let's start with writing our feature file the most crucial thing here is ubiquitous language. We don't want to specify how something is done,
+it doesn't really matter in case of business.
```
TheMadeleine: First of all let's start with writing our feature file - the most crucial thing here is the ubiquitous language, the main principle of BDD. We have to describe the functionalities of our system in a way all the project stakeholders will understand, by using words and sentences that sound natural and will in the end form a user story.

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,513 @@
...
+        Scenario: Adding country
+            Given I want to add a new country
+            When I choose "France"
+            And I add it
+            Then I should be notified that it has been successfully created
+            And the country "France" should appear in the store
+
+Pay special attention to form of those sentences, they are specially constructed that way to hide details like ``I click on select box and i click on France country, I see France country in box with css .table``.
+The side effect of this is that we got more generic steps and when we want to change way of testing this feature for example in domain context it will be extremely easy all we need to do is adding ``@domain`` tag and of course implement proper steps in domain context.
+
```
TheMadeleine: Pay attention to the form of these sentences. From the developer point of view they are hiding the details of the feature's implementation. Instead of describing 'When I click on the select box And I choose France from the dropdown Then I should see the France country in the table" - we are using sentences that are less connected with the implementation, but more focused on the effects of our actions.
A side effect of such approach  is that it results in steps being really generic, therefore if we want to add another way of testing this feature for instance in the domain context, it will be extremely easy to apply. We just need to add a different tag (in this case "@domain") and of course implement the proper steps in the domain context of our system.
To be more descriptive let's imagine that we want to check if a country is added properly in two ways. First we are checking if the adding works via frontend, so we are implementing steps that are clicking, opening pages, filling fields on forms and similar, but also we want to check this action regardlessly of the frontend, for that we need the domain, which allows us to perform actions only in the database.

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
+Basic Usage
+===========
+
+In our opinion the best way to understand how some stuff works is by showing some examples so this section gather all knowledge form previous chapters.
+Assume that we are implementing managing countries in our system.
+Let the fun begin.
+
+Describing feature
```
TheMadeleine: Describing features

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+            Given I want to add a new country
+            When I choose "France"
+            And I add it
+            Then I should be notified that it has been successfully created
+            And the country "France" should appear in the store
+
+Pay special attention to form of those sentences, they are specially constructed that way to hide details like ``I click on select box and i click on France country, I see France country in box with css .table``.
+The side effect of this is that we got more generic steps and when we want to change way of testing this feature for example in domain context it will be extremely easy all we need to do is adding ``@domain`` tag and of course implement proper steps in domain context.
+
+Choosing correct suite
```
TheMadeleine: Choosing a correct suite

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+----------------------
+Next thing is creating suite, in our example at the beginning we decide that it will be user interface feature.
+So this suite will be placed in ``etc/behat/suites/ui/addressing/managing_countries.yml`` and it will looks like
```
TheMadeleine: After we are done with a feature file, we have to create a new suite for it. At the beginning we have decided that it will be a frontend/user interface feature, that is why we are placing it in "etc/behat/suites/ui/addressing/managing_countries.yml".

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+
```
TheMadeleine: We need sth more: what is a suite? Add some description maybe.

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+.. code-block:: yaml
+
+    default:
+        suites:
+            ui_managing_countries:
+                contexts_as_services:
+                    - sylius.behat.context.hook.doctrine_orm //This service is responsible for clearing database before scenario.
```
TheMadeleine: `This service is responsible for clearing the database before each scenario, so that only data from the current scenario and its background is available.`

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+
+                    - sylius.behat.context.transform.country
+                    - sylius.behat.context.transform.shared_storage
+                    // Those contexts services are responsible for all transformation for example  "And the country "France" should appear in the store" transforms "(the country "France")" to proper Country object.
```
TheMadeleine: `The transformer contexts services are responsible for all the transformations of data in steps: For instance "And the country "France" should appear in the store" transforms "(the country "France")" to a proper Country object, which is from now on available in the scope of the step. `

pamil:
Have a look at the failing build, please :tada:

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+
+                    - sylius.behat.context.setup.geographical
+                    - sylius.behat.context.setup.security
+                    // Setup contexts in our case they are building our background scene. Preparing all actors.
```
TheMadeleine: `The setup contexts here are preparing the background, adding available countries and users or administrators. These contexts have steps like "I am logged in as an administrator" already implemented.`

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+
+                    // Lights, Camera, Action!
```
TheMadeleine: well... 💃

docs/behat/basic-usage.rst
pamil: YAML comments begin with `#`, not `//`

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+                    - sylius.behat.context.ui.admin.managing_countries
+                    - sylius.behat.context.ui.admin.notification
+                    // Those contexts are essential here we are placing all action steps like "When I choose "France" and I add it Then I should ne notified that...".
+                filters:
+                    tags: "@managing_countries && @ui"
+
+First essential thing that happens here is configuration of tags, from now behat will be searching all your features tagged with ``@managing_countires`` and your scenarios tagged with ``@ui``.
+Second thing is ``contexts_as_services:`` in this section we will be placing all our services with step implementation.
```
TheMadeleine: A very important thing that is done here is the configuration of tags, from now on Behat will be searching for all your features tagged with ``@managing_countires`` and your scenarios tagged with ``@ui``.

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+
+I mentioned about side effects of ubiquitous language that's why we can very easy switch our testing context to ``@domain``. Domain suite will look like:
```
TheMadeleine: `We have mentioned the with the generic steps we can easily switch our testing context to @domain. Have a look how it looks:`

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+                    - sylius.behat.context.transform.shared_storage
+
+                    - sylius.behat.context.setup.geographical
+                    - sylius.behat.context.setup.security
+
+                    - sylius.behat.context.domain.admin.managing_countries // Domain step implementation.
+                filters:
+                    tags: "@managing_countries && @domain"
+
+We have almost finished with suite configuration now we need to register our first behat context service, but before we will do it we need one more thing because like we said this is ui feature so we need page objects.
```
TheMadeleine:
```
We are almost finished with the suite configuration. 
Now we need to register our first Behat context as a service, but beforehand we need one more thing - page objects for the @ui suites.
```

docs/behat/basic-usage.rst
Arminek: This section is only about suite.

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+            CountryNameConverterInterface $countryNameConverter,
+            RepositoryInterface $countryRepository
+        ) {
+            $this->countryNameConverter = $countryNameConverter;
+            $this->countryRepository = $countryRepository;
+        }
+
+        /**
+         * @Transform /^country "([^"]+)"$/
+         * @Transform /^"([^"]+)" country$/
```
lchrusciel: We should definitely  show example with `:country` as they are the most common.

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+         */
+        public function iShouldBeNotifiedItHasBeenSuccessfullyCreated()
+        {
+            $this->notificationChecker->checkNotification('has been successfully created.', NotificationType::success());
+        }
+    }
+
+Transformer contexts
+~~~~~~~~~~~~~~~~~~~~
+
```
lchrusciel: You should also add warning, that no object should be created in transformers. Only allowed exceptions are Customer and Address.

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+        /**
+         * @Given /^the store has disabled country "([^"]*)"$/
+         */
+        public function theStoreHasDisabledCountry($countryName) // This method save country in data base.
+        {
+            $country = $this->createCountryNamed(trim($countryName));
+            $country->disable();
+
+            $this->sharedStorage->set('country', $country);
+            // Shared storage is an helper service for transferring objects between steps.
```
lchrusciel: I would add another paragraph with little shared storage explanation. We could point out a name convention for keys to  help users use _SharedStorageContext_ and also inform them, that `with great power comes great responsibility` so you shouldn't over use a SharedStorage instance.

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+
+                    - sylius.behat.context.domain.admin.managing_countries // Domain step implementation.
+                filters:
+                    tags: "@managing_countries && @domain"
+
+We have almost finished with suite configuration now we need to register our first behat context service, but before we will do it we need one more thing because like we said this is ui feature so we need page objects.
+
+Registering Pages
+-----------------
+Page object pattern allows us to hide all interaction with html behind those objects so all ui things happens here.
```
TheMadeleine: The page object approach allows us to hide all the detailed interaction with ui (html, javascript, css) inside.

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+         * @param RepositoryInterface $countryRepository
+         * @param CountryNameConverterInterface $countryNameConverter
+         */
+        public function __construct(
+            SharedStorageInterface $sharedStorage,
+            FactoryInterface $countryFactory,
+            RepositoryInterface $countryRepository,
+            CountryNameConverterInterface $countryNameConverter
+        ) {
+            $this->sharedStorage = $sharedStorage;
```
lchrusciel: We should also mention, that shared storage should be declared as a first argument in constructor.

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+                filters:
+                    tags: "@managing_countries && @domain"
+
+We have almost finished with suite configuration now we need to register our first behat context service, but before we will do it we need one more thing because like we said this is ui feature so we need page objects.
+
+Registering Pages
+-----------------
+Page object pattern allows us to hide all interaction with html behind those objects so all ui things happens here.
+
+We got three kinds of page:
```
TheMadeleine: `We have three kinds of pages:`

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+    - Page // First layer of our pages it knows how to interact with dom objects. It has method ->getUrl(array $urlParameters) where you can define raw url to open it.
```
TheMadeleine: - Page // First layer of our pages it knows how to interact with DOM objects. It has a method ->getUrl(array $urlParameters) where you can define a raw url to open it.

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+    - SymfonyPage // This page extends Page and it has router injected so from now ->getUrl() generates url from route name which gets from ->getRouteName() method.
```
TheMadeleine: ` - SymfonyPage // This page extends the Page. It has a router injected so that the ->getUrl() method generates a url from the route name which it gets from the ->getRouteName() method.`

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+    - Base Crud Pages (IndexPage, CreatePage, UpdatePage) // This page extends SymfonyPage and it's specific to sylius resources it has resource name injected and from now it knows about route name.
```
TheMadeleine: `- Base Crud Pages (IndexPage, CreatePage, UpdatePage) // These pages extend SymfonyPage and ithey are specific to the Sylius resources. They have a resource name injected and therefore they know about the route name.`

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+
+There are two ways to manipulate html. By using ``->getDocument()`` or ``->getElement('your_element')`` first method will return ``DocumentElement`` which represents html structure of currently opened page,
+second one is more tricky it use ``->getDefinedElements()`` method and it will return ``NodeElement`` which represents only restricted html structure.
```
TheMadeleine:
```
There are two ways to manipulate UI - by using ``->getDocument()`` or ``->getElement('your_element')``.
First method will return a ``DocumentElement`` which represents an html structure of the currently opened page,
second one is a bit more tricky because it uses the ``->getDefinedElements()`` method and it will return a ``NodeElement`` which represents only the restricted html structure.
```

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+Usage example of ``->getElement('your_element')`` and ``->getDefinedElements`` methods.
+
+.. code-block:: php
+
+    class CreatePage extends SymfonyPage implements CreatePageInterface
+    {
+        /**
+         *{@inheritdoc}
+         */
+        protected function getDefinedElements() // This method returns simple associative array, where key is name of your element and value is locator.
```
TheMadeleine: `// This method returns a simple associative array, where the key is the name of your element and the value is its locator.`

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+         *{@inheritdoc}
+         */
+        protected function getDefinedElements()
+        {
+            return array_merge(parent::getDefinedElements(), [
+                'provinces' => ['xpath' => '//*[contains(@class, "provinces")]'] // Now your value is an array where key is your locator type.
+            ]);
+        }
+
+        // Form now you can easy manipulate with your page elements.
```
TheMadeleine: `// Like that you can easily manipulate your page elements.`

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+        {
+            $provinceSelectBox = $this->getElement('provinces');
+
+            $provinceSelectBox->selectOption($province->getName());
+        }
+    }
+
+Let's back to our main example and analyse our scenario.
+We got steps like "When I choose "France" and I add it Then I should be notified that it has been successfully created And the country "France" should appear in the store".
+
```
TheMadeleine:
```
Let's get back to our main example and analyze our scenario.
We have steps like 
"When I choose "France" 
And I add it 
Then I should be notified that it has been successfully created 
And the country "France" should appear in the store".
```

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+            } catch (ElementNotFoundException $exception) {
+                return false;
+            }
+        }
+    }
+
+.. warning::
+
+    There is one small gap in this concept PageObjects aren't the concrete instance of current opened page they only mimic this behaviour (dummy pages).
+    This gap will be more understandable on code example.
```
TheMadeleine:
```
There is one small gap in this concept - PageObjects is not a concrete instance of the currently opened page, they only mimic its behaviour (dummy pages).
This gap will be more understandable on the below code example.
```

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+
+    // Of course this is only to illustrate this gap.
+
+    final class GapContext implements Context
+    {
+        private $homePage;
+        private $leagueIndexPage;
+
+        /**
+         * @Given I want to be on home page
```
TheMadeleine: I want to be on the Homepage

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+         */
+        public function iWantToBeOnHomePage() // After this method call we will be on "http://your_domain.com".
+        {
+            $this->homePage->open(); //When we add @javascript tag we can actually see this thanks to selenium.
+        }
+
+        /**
+         * @Then I want to read side bar and get The weather in France information
```
TheMadeleine: `Then I want to see the sidebar and get information about the weather in France`

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+         */
+        public function iWantToReadSideBarOnHomePage($someInformation) // Still "http://your_domain.com".
+        {
+            $someInformation === $this->leagueIndexPage->readMatchResults() // This returns true, but wait a second we are on home page (dummy pages).
```
TheMadeleine: Don;t you do different things than specified in the step? Weather !== Match results, i think. :)

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+            return ['sidebar' => ['css' => '.sidebar']]
+        }
+
+        protected function getUrl()
+        {
+            return 'http://your_domain.com/leagues/'
+        }
+    }
+
+Registering context
```
TheMadeleine: `Registering contexts`

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,510 @@
...
+-------------------
+Like you see in previous section we have registered a lot of contexts so we will show you only some of step implementation.
```
TheMadeleine: `As it was shown in the previous section we have registered a lot of contexts, so we will show you only some of the step implementation.`

docs/behat/basic-usage.rst
Arminek: IMO this section is precise enough.
http://docs.sylius.org/en/latest/behat/how-to-use-transformers.html

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,544 @@
+Basic Usage
+===========
+
+The best way of understanding how things work in detail is showing and analyzing examples, that is why this section gathers all the knowledge from the previous chapters.
+Let's assume that we are going to implement the functionality of managing countries in our system.
+Now let us show you the flow.
+
+Describing features
+-------------------
+First of all let's start with writing our feature file the most crucial thing here is ubiquitous language. We don't want to specify how something is done,it doesn't really matter in case of business.
```
pjedrzejewski: I am not sure what you mean here. Ubiquitous Language is about building a common language between users and developers. http://martinfowler.com/bliki/UbiquitousLanguage.html

Perhaps something like:

```
Let's start with writing our feature file, which will contain answers to the most important questions: Why (benefit), who (actor using the feature) and what (the feature itself). It should also include scenarios, which serve as examples of how things supposed to work.
```

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,540 @@
...
+                    # These contexts have steps like "I am logged in as an administrator" already implemented.
+
+                    # Lights, Camera, Action!
+                    - sylius.behat.context.ui.admin.managing_countries
+                    - sylius.behat.context.ui.admin.notification
+                    # Those contexts are essential here we are placing all action steps like "When I choose "France" and I add it Then I should ne notified that...".
+                filters:
+                    tags: "@managing_countries && @ui"
+
+A very important thing that is done here is the configuration of tags, from now on Behat will be searching for all your features tagged with ``@managing_countiresand`` your scenarios tagged with ``@ui``.
```
tuka217: Shouldn't be '@managing_countries' ?

pamil: Sand may be tired sometimes -.-

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,540 @@
...
+        @ui
+        Scenario: Adding country
+            Given I want to add a new country
+            When I choose "France"
+            And I add it
+            Then I should be notified that it has been successfully created
+            And the country "France" should appear in the store
+
+Pay attention to the form of these sentences. From the developer point of view they are hiding the details of the feature's implementation. Instead of describing "When I click on the select box And I choose France from the dropdown Then I should see the France country in the table" - we are using sentences that are less connected with the implementation, but more focused on the effects of our actions.
+A side effect of such approach is that it results in steps being really generic, therefore if we want to add another way of testing this feature for instance in the domain context, it will be extremely easy to apply. We just need to add a different tag (in this case "@domain") and of course implement the proper steps in the domain context of our system.
```
michalmarcinkowski: domain **or api** context

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,540 @@
...
+         * @Transform :country
+         * @Transform /^"([^"]+)" country$/
+         */
+        public function getCountryByName($countryName) // Thanks to this method we got in our ManagingCountries an Country object.
+        {
+            $countryCode = $this->countryNameConverter->convertToCode($countryName);
+            $country = $this->countryRepository->findOneBy(['code' => $countryCode]);
+
+            if (null === $country) {
+                throw new \InvalidArgumentException(sprintf('Country with name %s does not exist', $countryName));
```
michalmarcinkowski: `Assert::notNull`

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,540 @@
...
+        {
+            try {
+                $rows = $this->tableAccessor->getRowsWithFields($this->getElement('table'), $parameters);
+                // Table accessor is a helper service which is responsible for all html table operations.
+
+                return 1 === count($rows);
+            } catch (\InvalidArgumentException $exception) {
+                return false;
+            } catch (ElementNotFoundException $exception) {
+                return false;
```
michalmarcinkowski: I would add a comment that we catch `ElementNotFoundException` since if there are no resources the `table` is not rendered (both in the docs and the code 😄 ).

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,540 @@
...
+         */
+        public function isSingleResourceOnPage(array $parameters)
+        {
+            try {
+                $rows = $this->tableAccessor->getRowsWithFields($this->getElement('table'), $parameters);
+                // Table accessor is a helper service which is responsible for all html table operations.
+
+                return 1 === count($rows);
+            } catch (\InvalidArgumentException $exception) {
+                return false;
```
michalmarcinkowski: Why do we catch `\ InvalidArgumentException`?

docs/behat/basic-usage.rst
```diff
@@ -0,0 +1,540 @@
...
+            RepositoryInterface $countryRepository
+        ) {
+            $this->countryNameConverter = $countryNameConverter;
+            $this->countryRepository = $countryRepository;
+        }
+
+        /**
+         * @Transform /^country "([^"]+)"$/
+         * @Transform :country
+         * @Transform /^"([^"]+)" country$/
```
pamil: I guess there's way too many `@Transform` for a tutorial :)

lchrusciel: Maybe it would be worth, to present two examples and discuss differences between them?

pjedrzejewski:
Nice work, thanks Arek and everyone for the review! 👍

PR #5018, opened 2016-05-13
[Travis] More verbose docs build

pamil:
`-W` turns on error code returning on warnings, but shows only the first one


pjedrzejewski:
👍

PR #5020, opened 2016-05-13
[Travis] Innovational, intelligent suite executing ;)

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | -
| License         | MIT

- if changing files only in `docs/*`, only `docs` suite will be runned
- if changing files outside `src/Sylius/Bundle` / `src/Sylius/Component`, `packages` suite will not be runned

For example, this PR should not run both `docs` and `packages` suites, so the Travis build should end ~10 minutes earlier. If a PR is documentation-only, it should end just after a few minutes.


lchrusciel:
Why did you change access level to install.sh and script.sh?

pamil:
@lchrusciel they should be executable just like every other script inside `etc/travis`

pjedrzejewski:
Nice one! 👍

PR #5025, opened 2016-05-13
[Travis] Disable documentation cache

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | #5020
| License         | MIT

Since documentation is build only while it was changed, caching, that may cause some serious issues while building (like hiding previous warnings, not regenerating the root toctree), is disabled as it will no longer affect the performance of non-documentation builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,6 @@ cache:
         - vendor
         - bin
         - node_modules
-        - docs/build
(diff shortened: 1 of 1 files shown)


michalmarcinkowski:
:+1:

PR #5030, opened 2016-05-13
[Translations] - frontend/legacy admin fix unnecessary double translating

nakashu:
| Q | A |
| --- | --- |
| Bug fix? | yes? |
| New feature? | no |
| BC breaks? | yes - menubuilders |
| Related tickets | # |
| License | MIT |

Mainly changed menu builder label translation and moved it to template, similar as in new admin.
Added new argument in Menu - to control if string should be translated or not
- translation_domain (similar as in forms)
- translation_parameters - to pass variables
- see menu.html.twig

99% percent of missing translation that where false possitives are now removed.
-> tiny performance boost + easier to identify string for translation that are missing for real

### DIFF
--- a/src/Sylius/Bundle/CartBundle/Form/Type/CartType.php
+++ b/src/Sylius/Bundle/CartBundle/Form/Type/CartType.php
@@ -31,7 +31,9 @@ class CartType extends AbstractResourceType
             ->add('items', 'collection', [
                 'type' => 'sylius_cart_item',
             ])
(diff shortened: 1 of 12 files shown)


nakashu:
I was also thinking about moving all translation from `menu` domain to `messages` domain same as it is in new admin. If it won't break crowdin translation - I can add it to this PR if interested.

pamil:
No legacy admin means no more issues with legacy admin! :)

PR #5032, opened 2016-05-14
Allow multiple search indexes

inspiran:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

If you have two search indexes, eg. 'sylius' and 'site', this patch prevents sylius from loading an event listener for site.product if the product type doesn't exists.   It is not the ideal patch, but at least now you can use multiple FOS indexes.


pjedrzejewski:
Thank you Daniel! We should definitely take a look at proper fix soon. ;)

PR #5035, opened 2016-05-15
[Doc][Translation] Add example for usage with SyliusTranslationBundle

SamFleming:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Related tickets |  |
| License | MIT |

After spending an long time figuring out why my translation entities weren't being persisted, I discovered they have to be configured through the SyliusResourceBundle.

Could we add something like this to the Translation component documentation? Hopefully it'll prevent others wasting as much time as I did.

(Unless I've been stupid and this has been documented elsewhere?)


pamil:
Translation component/bundle was merged into Resource.

PR #5038, opened 2016-05-16
[Currency] Do not use booleans as money formatter's locales

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

It crashes when `ext-intl` is not available. Anyway, `false` and `true` does not look like a valid locales :)


pjedrzejewski:
👍

PR #5041, opened 2016-05-16
[Travis] More verbose Symfony setup

pamil:
Makes debugging easier; especially these ones, that happens during container compilation and only if there's no database.


pjedrzejewski:
👍

PR #5044, opened 2016-05-17
Fix Archetype parent field mapping

mihaichitic:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets |
| License         | MIT

'joinColumn' key is ignored, so onDelete is not applied in DB.


pjedrzejewski:
Archetypes have been removed.

PR #5045, opened 2016-05-17
[Resource] Fix broken spec in ResourceBundle

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

It should be deeply investigated, why this nasty bug has not been caught by **Travis** build.


michalmarcinkowski:
Thanks Mateusz! 👍 

**Travis** and the build is fine. The broken build on master was caused by merging the outdated PR (https://github.com/Sylius/Sylius/pull/4618). We could prevent that by enforcing the PRs to be always rebased with the newest master, but this will require to rebase all PRs after each merge, so for practical reasons we must live with this happening from time to time 😉

PR #5047, opened 2016-05-17
Fix and refactor Load*MetadataSubscriber 

### LINKED ISSUE #5040: What is the purpose of Load(ORM/ODM)MetadataSubscriber?
Hi!

I'm using Sylius Resource Bundle to serve a rest API.

Simply, `LoadORMMetadataSubscriber` is breaking some the associations in my custom entities, specifically the `unsetAssociationMappings` method.

Before trying to fix the behaviour, I need to understand what it is for!
What is the purpose of these listeners?

Thanks

bendavies:
| Q               | A
| --------------- | ---
| Bug fix?      |  yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #5040
| License         | MIT

I've split the `Load*MetadataSubscriber` out into two different subscribers.
`*MappedSuperClassSubscriber`, which should only run on Sylius entities.
`*RepositoryClassSubscriber`, which should run on all entities.

Unfortunately there are no existing specs for mapped super class stuff.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/EventListener/AbstractDoctrineSubscriber.php
+++ b/src/Sylius/Bundle/ResourceBundle/EventListener/AbstractDoctrineSubscriber.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\ResourceBundle\EventListener;
+
+use Doctrine\Common\EventSubscriber;
+use Doctrine\Common\Persistence\Mapping\ClassMetadata;
+use Sylius\Component\Resource\Metadata\RegistryInterface;
+
+/**
+ * @author Ben Davies <ben.davies@gmail.org>
+ */
+abstract class AbstractDoctrineSubscriber implements EventSubscriber
+{
+    /**
+     * @var RegistryInterface
+     */
+    protected $resourceRegistry;
+
+    /**
+     * @param RegistryInterface $resourceRegistry
+     */
+    public function __construct(RegistryInterface $resourceRegistry)
+    {
+        $this->resourceRegistry = $resourceRegistry;
+    }
+
+    /**
+     * @param ClassMetadata $metadata
+     *
+     * @return bool
+     */
+    protected function isSyliusClass(ClassMetadata $metadata)
+    {
+        return 0 === strpos($metadata->getName(), 'Sylius\\');
+    }
+}
--- a/src/Sylius/Bundle/ResourceBundle/EventListener/ODMMappedSuperClassSubscriber.php
+++ b/src/Sylius/Bundle/ResourceBundle/EventListener/ODMMappedSuperClassSubscriber.php
@@ -11,38 +11,25 @@
 
 namespace Sylius\Bundle\ResourceBundle\EventListener;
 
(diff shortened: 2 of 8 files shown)


bendavies:
@pamil hello Sylius CI travis expert!
This is failing because I have a spec which is using Doctrine Mongo ODM classes, but suggested packages are not installed by the travis job. What do you think? We could install them with `--ignore-platform-reqs`?

bendavies:
@pjedrzejewski ready for review.

pamil:
@bendavies I'll try to get `mongodb` extension up and running on Travis and get rid of existing `--ignore-platform-reqs` for ResourceBundle, then we'll be able to add the packages requiring mongodb to `require-dev`.

src/Sylius/Bundle/ResourceBundle/EventListener/AbstractDoctrineSubscriber.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+namespace Sylius\Bundle\ResourceBundle\EventListener;
+
+use Doctrine\Common\EventSubscriber;
+use Doctrine\Common\Persistence\Mapping\ClassMetadata;
+use Sylius\Component\Resource\Metadata\RegistryInterface;
+
+/**
+ * @author Paweł Jędrzejewski <pawel@sylius.org>
```
lchrusciel: Is it true?

bendavies: well, i'm not sure of your convention. Symfony seemed to have a convention of putting Fabien as as the author of all files, regardless of who wrote the code.
I'll put myself if you want :)

bendavies:
thanks @pamil !

src/Sylius/Bundle/ResourceBundle/EventListener/AbstractDoctrineSubscriber.php
lchrusciel: I think you should add yourself as an author :)

pjedrzejewski: I am not sure if it is true for Symfony (probably no), but in Sylius we should definitely only mention the real authors of the file, not me everywhere. :)

pamil:
@bendavies unfortunately there's no compatibile Doctrine MongoDB support for the PHP 7 driver, so it'll be more hacky solution :) We already have [Akeneo PhpSpecSkipExampleExtension](https://github.com/akeneo/PhpSpecSkipExampleExtension) included in main PhpSpec configuration, so adding `@require` annotations to specs requiring those classes will prevent running them on application level, whereas they will be run during ResourceBundle package tests (the only `--ignore-platform-reqs` package).

bendavies:
@pamil done!

src/Sylius/Bundle/ResourceBundle/EventListener/AbstractDoctrineSubscriber.php
bendavies: done :)

pjedrzejewski:
Thank you Ben! One thing we should look at later is loading the ORM/ODM stuff only when doctrine ORM/ODM is actually installed in Symfony. Right now it just ignores the tags, but why put more services than needed. Also, ODM is not only MongoDB, so probably we should make the naming more specific. That's for the future I guess.

bendavies:
you're welcome! yep i noticed that it could support mongo and phpcr. I guess lost of stuff needs  renaming to be Mongo specific.
Cheers!

PR #5058, opened 2016-05-18
Executable tests on Sylius-Standard & remove duplicated configs

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | -
| License         | MIT


pjedrzejewski:
General note, I have not reviewed it in full yet: We need to have same structure as Symfony Standard whenever possible, even if this means duplication of some configs. I don't want to write a whole chapter in docs about why configs are splitted and included separately and answering questions: "where is configuration X, in Symfony it is right there!". I'd rather still live with a small duplication.

pamil:
@pjedrzejewski I agree, but there are over 1500 lines of configuration compared to 100 lines of Symfony Standard, so we must think through what we can duplicate and what rather not.

pjedrzejewski:
The Sylius defaults should not be duplicated of course and I believe they could stay in the Sylius[Shop/Admin/Core accordingly]Bundle, just like they are right now. In main config files I would leave the basic Symfony/Doctrine settings you find in all Symfony apps and anything that's cannot be overriden (fos_rest paths IIRC are an example)

app/AppKernel.php
```diff
@@ -23,15 +24,153 @@ class AppKernel extends Kernel
...
+            new \Payum\Bundle\PayumBundle\PayumBundle(),
+            new \JMS\SerializerBundle\JMSSerializerBundle(),
+            new \JMS\TranslationBundle\JMSTranslationBundle(),
+            new \Stof\DoctrineExtensionsBundle\StofDoctrineExtensionsBundle(),
+            new \WhiteOctober\PagerfantaBundle\WhiteOctoberPagerfantaBundle(),
+
+            new \Doctrine\Bundle\MigrationsBundle\DoctrineMigrationsBundle(),
+            new \Doctrine\Bundle\FixturesBundle\DoctrineFixturesBundle(),
+            new \Sylius\Bundle\FixturesBundle\SyliusFixturesBundle(),
+            new \Sylius\Bundle\PayumBundle\SyliusPayumBundle(), // must be added after PayumBundle.
```
michalmarcinkowski: This comment doesn't make sense 😄

app/config/config.yml
```diff
@@ -2,88 +2,54 @@
 # (c) Paweł Jędrzejewski
 
 imports:
-    - { resource: parameters.yml }
-    - { resource: security.yml }
-    - { resource: migrations.yml }
-    - { resource: @SyliusCoreBundle/Resources/config/app/main.yml }
-    - { resource: @SyliusAdminBundle/Resources/config/app/main.yml }
+    - { resource: "security.yml" }
+
```
michalmarcinkowski: `parameters.yml` shouldn't be there also?

app/config/config_test.yml
```diff
@@ -65,3 +38,29 @@ payum:
...
+    disable_delivery: true
+
+sylius_channel:
+    fake_channel_support: true
+
+sylius_payment:
+    gateways:
+        cc: Credit Card
+        test: Test
+        stripe: Stripe
```
michalmarcinkowski: Paypal is missing.

src/Sylius/Bundle/AddressingBundle/Resources/config/config.yml
```diff
@@ -0,0 +1,9 @@
+# This file is part of the Sylius package.
+# (c) Paweł Jędrzejewski
+
+jms_serializer:
+    metadata:
+        directories:
+            sylius-addressing:
+                namespace_prefix: "Sylius\\Component\\Addressing"
+                path: "@SyliusAddressingBundle/Resources/config/serializer"
```
michalmarcinkowski: Why do we need this file?

src/Sylius/Bundle/CoreBundle/Resources/config/config.yml
```diff
@@ -0,0 +1,357 @@
...
+            timestampable: true
+            loggable: true
+            sortable: true
+
+twig:
+    debug: "%kernel.debug%"
+    strict_variables: "%kernel.debug%"
+    exception_controller: "FOS\\RestBundle\\Controller\\ExceptionController::showAction"
+
+winzou_state_machine:
```
michalmarcinkowski: IMO this should be extracted to a separate file.

pjedrzejewski:
* I think we should separate app configurations from services/parameters. Not sure if current ``config/app`` directory is a good idea, but having one big ``config.yml`` next to ``services.xml``, ``form.xml``, etc. can be confusing.
* State machines should definitely be in separate files.
* Why removing base Kernel class?

src/Sylius/Bundle/CoreBundle/Resources/config/sylius.yml
```diff
@@ -138,101 +39,122 @@ sylius_archetype:
...
-            variant:
+            subject: "%sylius.model.product.class%"
+            association:
                 classes:
-                    model: Sylius\Component\Core\Model\ProductVariant
-                    repository: Sylius\Bundle\CoreBundle\Doctrine\ORM\ProductVariantRepository
-                    form:
-                        default: Sylius\Bundle\CoreBundle\Form\Type\ProductVariantType
+                    model: Sylius\Component\Product\Model\ProductAssociation
+                    interface: Sylius\Component\Product\Model\ProductAssociationInterface
```
michalmarcinkowski: Not a concern of this PR, but this shouldn't live in `ProductBundle`?

pamil: This way we would couple ProductBundle with AssociationBundle.

pamil:
@pjedrzejewski:

 - Moved back these files to bundles' `config/app/*`
 - Extracted state machine configuration to bundles' `config/app/state_machine.yml`
 - The kernel class from CoreBundle was moved to `/app/`, because it also includes WebBundle / AdminBundle etc. and it's definitely not a property of CoreBundle

pamil:
I think it's all for this iteration, if there aren't any more issues, it would be nice to get this PR merged so I can open the another one on Sylius-Standard.

src/Sylius/Behat/Resources/config/profiles/default.yml
```diff
@@ -40,8 +40,8 @@ default:
 
         Behat\Symfony2Extension:
             kernel:
-                class: TestKernel
-                path: app/TestKernel.php
+                class: TestAppKernel
+                path: app/TestAppKernel.php
```
lchrusciel: Shouldn't path have `"`?

pamil: It's not required in this case.

src/Sylius/Bundle/CoreBundle/Resources/config/sylius.yml
michalmarcinkowski: `Sylius\Component\Product\Model\ProductAssociation` - it is already coupled :)

michalmarcinkowski: And probably with the bundle ([composer.json](https://github.com/michalmarcinkowski/Sylius/blob/f2ef70f73186b6ca73d658abde1d5310c8fcf270/src/Sylius/Bundle/ProductBundle/composer.json#L29-L29)).

app/migrations/Version20160523143611.php
```diff
@@ -17,7 +8,7 @@
 /**
  * Auto-generated Migration: Please modify to your needs!
  */
-class Version20160516131553 extends AbstractMigration
+class Version20160523143611 extends AbstractMigration
```
michalmarcinkowski: Why this migration was not removed?

pamil: It's the first one made after v0.18 was tagged

lchrusciel:
Can we setup test environment based on migration rather than doctrine schema generation?

bendavies:
@lchrusciel I would say the travis build should do that too. That way, travis can run `doctrine:schema:validate` a check if any schema changes have not been accompanied with a migration. which has happened more than once.

lchrusciel:
@bendavies this is exactly what I meant :D We should ensure, that this won't happen again.

bendavies:
@lchrusciel ah right! i thought you meant the demo env or something.

src/Sylius/Bundle/ShippingBundle/Resources/config/app/state_machine.yml
```diff
@@ -0,0 +1,88 @@
+# This file is part of the Sylius package.
+# (c) Paweł Jędrzejewski
+
+jms_serializer:
```
pjedrzejewski: This should not be here.

pamil: 👍

michalmarcinkowski:
Good job! 👍

nakashu:
@pamil why where all migrations removed ??? That seems a bit drastic for upgrading.

psyray:
Agree with @nakashu, remove migrations is a really bad idea for those who want to upgrade from very old sylius version.
And this part was not really discussed her.
I think this part should be quickly reverted.

pjedrzejewski:
@psyray We will publish old migrations in a separate repo probably, because we need to generate new ones and figure out proper upgrade strategy that considers application + sylius migrations together.

psyray:
OK, right, but meanwhile there is no possibility to cleanly migrate old sylius for those who wants.
Maybe you could put a dir named old with old migrations and a README textfile in migrations folder to permit user to manually copy/paste migrations in main folder ?

peteward:
The old migrations didn't work for anyone with a DB that has content because they didn't consider any data - they just did structure. So you would get a same result by just doing a `doctrine:migrations:diff` and then working through your issues.

psyray:
I'm not agree with you.
Some old migrations have logic and move data from a table to another, you can look at those with code migrations or translations (like [Version20160112154949](https://github.com/Sylius/Sylius/blob/v0.18.0/app/migrations/Version20160112154949.php) [Version20151203134947](https://github.com/Sylius/Sylius/blob/v0.18.0/app/migrations/Version20151203134947.php) or [Version20141203091118](https://github.com/Sylius/Sylius/blob/v0.18.0/app/migrations/Version20141203091118.php)).

psyray:
Another example is channel, which insert default channel and update the whole channel tables.
[Version20150202213852](https://github.com/Sylius/Sylius/blob/v0.18.0/app/migrations/Version20150202213852.php)

psyray:
More recent option_value translation
[Version20160203224648](https://github.com/Sylius/Sylius/blob/v0.18.0/app/migrations/Version20160203224648.php)

I'm agree that a lot of migrations was obsolete and was just a `doctrine:migrations:diff`but some are not, and drop the entire migration folder was a mistake.

PR #5062, opened 2016-05-18
[WIP] [Order] Order improvements

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

This PR will be included in #4768. It is a first step to get proper amounts for all the stuff on order that we want.

* [x] Added `getTaxTotal` method to `Order` which returns sum of tax adjustments regardless neutrality - useful when you want to know the value of tax that was charged
* [x] Added `getShippingTotal` method to `Order` which returns shipping fee together with shipping tax decreased by shipping discount - useful when you want to know the end value that was charged for shipping
* [x] Made `getAdjustmentsTotal` and `getAdjustmentsTotalRecursively` methods consistent - when called without argument they returned sum of only non neutral adjustments, when called with `$type` argument all adjustments were summed (both non neutral and neutral). Right now always only non neutral adjustments are summed.
* [x] Removed `getPromotionsTotalRecursively` in favor of `getOrderPromotionTotal` - this method returns total amount of order level promotion (e.g. $5 off to every order with items total at least $50.00). This does not include item level promotions (e.g. $1 off every T-Shirt) and does not include shipping discount which is included in `getShippingTotal`.
* [ ] Add `getDiscountedUnitPrice` method to `OrderItem` that will return the unit price decreased by item promotions.
* [ ] Add `getSubtotal` method to `OrderItem` that will return `discountedUnitPrice * quantity`

/cc @peteward

### DIFF
--- a/src/Sylius/Component/Core/Model/Order.php
+++ b/src/Sylius/Component/Core/Model/Order.php
@@ -527,17 +527,6 @@ class Order extends Cart implements OrderInterface
     }
 
     /**
-     * @return int
-     */
-    public function getPromotionsTotalRecursively()
-    {
-        return
-            $this->getAdjustmentsTotalRecursively(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT) +
-            $this->getAdjustmentsTotalRecursively(AdjustmentInterface::ORDER_ITEM_PROMOTION_ADJUSTMENT)
-        ;
-    }
-
-    /**
      * Returns sum of neutral and non neutral tax adjustments on order and total tax of order items.
      *
      * {@inheritdoc}
@@ -556,14 +545,36 @@ class Order extends Cart implements OrderInterface
         return $taxTotal;
     }
 
+    /**
+     * Returns shipping fee together with taxes decreased by shipping discount.
+     *
+     * @return int
+     */
     public function getShippingTotal()
     {
         $shippingTotal = $this->getAdjustmentsTotal(AdjustmentInterface::TAX_ADJUSTMENT);
+        $shippingTotal += $this->getAdjustmentsTotal(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT);
 
         foreach ($this->getAdjustments(AdjustmentInterface::SHIPPING_ADJUSTMENT) as $shippingAdjustment) {
             $shippingTotal += $shippingAdjustment->getAmount();
         }
 
         return $shippingTotal;
+    }
+
+    /**
+     * Returns amount of order discount. Does not include order item and shipping discounts.
+     *
+     * @return int
+     */
+    public function getOrderPromotionTotal()
+    {
+        $orderPromotionTotal = 0;
+
+        foreach ($this->items as $item) {
+            $orderPromotionTotal += $item->getAdjustmentsTotalRecursively(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT);
+        }
+
+        return $orderPromotionTotal;
     }
 }
--- a/src/Sylius/Component/Core/Model/Order.php
+++ b/src/Sylius/Component/Core/Model/Order.php
@@ -536,4 +536,24 @@ class Order extends Cart implements OrderInterface
             $this->getAdjustmentsTotalRecursively(AdjustmentInterface::ORDER_ITEM_PROMOTION_ADJUSTMENT)
         ;
     }
+
+    /**
+     * Returns sum of neutral and non neutral tax adjustments on order and total tax of order items.
+     *
+     * {@inheritdoc}
+     */
+    public function getTaxTotal()
+    {
+        $taxTotal = 0;
+
+        foreach ($this->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT) as $taxAdjustment) {
+            $taxTotal += $taxAdjustment->getAmount();
+        }
+
+        foreach ($this->items as $item) {
+            $taxTotal += $item->getTaxTotal();
+        }
+
+        return $taxTotal;
+    }
 }
--- a/src/Sylius/Component/Core/Model/Order.php
+++ b/src/Sylius/Component/Core/Model/Order.php
@@ -549,11 +549,21 @@ class Order extends Cart implements OrderInterface
         foreach ($this->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT) as $taxAdjustment) {
             $taxTotal += $taxAdjustment->getAmount();
         }
-
         foreach ($this->items as $item) {
             $taxTotal += $item->getTaxTotal();
         }
 
         return $taxTotal;
+    }
+
+    public function getShippingTotal()
+    {
+        $shippingTotal = $this->getAdjustmentsTotal(AdjustmentInterface::TAX_ADJUSTMENT);
+
+        foreach ($this->getAdjustments(AdjustmentInterface::SHIPPING_ADJUSTMENT) as $shippingAdjustment) {
+            $shippingTotal += $shippingAdjustment->getAmount();
+        }
+
+        return $shippingTotal;
     }
 }
--- a/src/Sylius/Component/Core/Model/OrderItem.php
+++ b/src/Sylius/Component/Core/Model/OrderItem.php
@@ -57,4 +57,24 @@ class OrderItem extends CartItem implements OrderItemInterface
             && $item->getVariant() === $this->variant
         );
     }
+
+    /**
+     * Returns sum of neutral and non neutral tax adjustments on order item and total tax of units.
+     *
+     * {@inheritdoc}
+     */
+    public function getTaxTotal()
+    {
+        $taxTotal = 0;
+
+        foreach ($this->getAdjustments(AdjustmentInterface::TAX_ADJUSTMENT) as $taxAdjustment) {
+            $taxTotal += $taxAdjustment->getAmount();
+        }
+
+        foreach ($this->units as $unit) {
+            $taxTotal += $unit->getTaxTotal();
+        }
+
+        return $taxTotal;
+    }
 }
--- a/src/Sylius/Component/Order/Model/Order.php
+++ b/src/Sylius/Component/Order/Model/Order.php
@@ -435,7 +435,9 @@ class Order implements OrderInterface
 
         $total = 0;
         foreach ($this->getAdjustments($type) as $adjustment) {
-            $total += $adjustment->getAmount();
+            if (!$adjustment->isNeutral()) {
+                $total += $adjustment->getAmount();
+            }
         }
 
         return $total;
@@ -448,7 +450,9 @@ class Order implements OrderInterface
     {
         $total = 0;
         foreach ($this->getAdjustmentsRecursively($type) as $adjustment) {
-            $total += $adjustment->getAmount();
+            if (!$adjustment->isNeutral()) {
+                $total += $adjustment->getAmount();
+            }
         }
 
         return $total;
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Order/_items.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Order/_items.html.twig
@@ -89,8 +89,7 @@
         </th>
         <th colspan="4" id="shipping-total" class="right aligned">
             <strong>{{ 'sylius.ui.shipping_total'|trans }}</strong>:
-            {% set shippingTotal = order.adjustmentsTotal(taxAdjustment) + order.adjustmentsTotal(shippingAdjustment) %}
-            {{ shippingTotal|sylius_price(order.currency) }}
+            {{ order.shippingTotal|sylius_price(order.currency) }}
         </th>
     </tr>
     <tr>
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Order/_items.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Order/_items.html.twig
@@ -130,7 +130,7 @@
         </th>
         <th colspan="4" id="promotion-total" class="right aligned">
             <strong>{{ 'sylius.ui.promotion_total'|trans }}</strong>:
-            {% set orderDiscount = order.adjustmentsTotal(orderPromotionAdjustment) %}
+            {% set orderDiscount = order.orderPromotionTotal %}
             {% if not orderDiscount == 0 %}-{% endif %}{{ (-1 * orderDiscount)|sylius_price(order.currency) }}
         </th>
     </tr>
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Order/_items.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Order/_items.html.twig
@@ -53,7 +53,7 @@
                 {{ (item.unitPrice * item.quantity)|sylius_price(order.currency) }}
             </td>
             <td class="center aligned">
-                {{ item.getAdjustmentsTotalRecursively(taxAdjustment)|sylius_price(order.currency) }}
+                {{ item.taxTotal|sylius_price(order.currency) }}
             </td>
             <td class="center aligned discount">
                 {% set itemDiscount = item.getAdjustmentsTotalRecursively(itemPromotionAdjustment) %}
@@ -111,7 +111,7 @@
         </th>
         <th colspan="4" id="tax-total" class="right aligned">
             <strong>{{ 'sylius.ui.tax_total'|trans }}</strong>:
-            {{ order.getAdjustmentsTotalRecursively(taxAdjustment)|sylius_price(order.currency) }}
+            {{ order.taxTotal|sylius_price(order.currency) }}
         </th>
     </tr>
     <tr>
--- a/src/Sylius/Bundle/WebBundle/Resources/views/Backend/Order/_form.html.twig
+++ b/src/Sylius/Bundle/WebBundle/Resources/views/Backend/Order/_form.html.twig
@@ -148,7 +148,7 @@
         </td>
         <td class="text-right" colspan="3">
             <strong>{{ 'sylius.ui.shipping_total'|trans }}</strong>:
-            <span class="amount">{{ order.adjustmentsTotal(shippingAdjustment)|sylius_price(order.currency) }}</span>
+            <span class="amount">{{ order.shippingTotal|sylius_price(order.currency) }}</span>
         </td>
     </tr>
     <tr class="promotion-discount">
--- a/src/Sylius/Bundle/WebBundle/Resources/views/Backend/Order/_form.html.twig
+++ b/src/Sylius/Bundle/WebBundle/Resources/views/Backend/Order/_form.html.twig
@@ -130,7 +130,7 @@
         </td>
         <td class="text-right" colspan="3">
             <strong>{{ 'sylius.ui.tax_total'|trans }}</strong>:
-            <span class="amount">{{ order.getAdjustmentsTotalRecursively(taxAdjustment)|sylius_price(order.currency) }}</span>
+            <span class="amount">{{ order.taxTotal|sylius_price(order.currency) }}</span>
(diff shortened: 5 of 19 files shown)


src/Sylius/Component/Core/Model/Order.php
```diff
@@ -527,13 +527,54 @@ public function getPromotions()
...
+     * Returns amount of order discount. Does not include order item and shipping discounts.
+     *
+     * @return int
+     */
+    public function getOrderPromotionTotal()
+    {
+        $orderPromotionTotal = 0;
+
+        foreach ($this->items as $item) {
+            $orderPromotionTotal += $item->getAdjustmentsTotalRecursively(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT);
```
peteward: I thought `ORDER_PROMOTION_ADJUSTMENT` only applies to `Orders` not `OrderItems` and `OrderItemUnits`?

src/Sylius/Component/Core/Model/Order.php
```diff
@@ -527,13 +527,54 @@ public function getPromotions()
...
+    public function getShippingTotal()
     {
-        return
-            $this->getAdjustmentsTotalRecursively(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT) +
-            $this->getAdjustmentsTotalRecursively(AdjustmentInterface::ORDER_ITEM_PROMOTION_ADJUSTMENT)
-        ;
+        $shippingTotal = $this->getAdjustmentsTotal(AdjustmentInterface::TAX_ADJUSTMENT);
+        $shippingTotal += $this->getAdjustmentsTotal(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT);
+
+        foreach ($this->getAdjustments(AdjustmentInterface::SHIPPING_ADJUSTMENT) as $shippingAdjustment) {
```
peteward: why not just:
```
$shippingTotal += $this->getAdjustmentsTotal(AdjustmentInterface::SHIPPING_ADJUSTMENT);
```

src/Sylius/Component/Order/Model/Order.php
```diff
@@ -435,7 +435,9 @@ public function getAdjustmentsTotal($type = null)
 
         $total = 0;
         foreach ($this->getAdjustments($type) as $adjustment) {
-            $total += $adjustment->getAmount();
+            if (!$adjustment->isNeutral()) {
+                $total += $adjustment->getAmount();
+            }
```
peteward: What's the rationale behind this change? 

The only use-case I can think of for neutrality is taxes, maybe there could be other adjustments such as handling fees which could be inclusive, or commission fees?

In any case, I think the totalling individual adjustment types should not exclude neutral adjustments. If you want to find out how much of something there is on an order, you should just be able to call this method for that type.

While I appreciate that this is different from the overall total which _does_ consider neutrality, that's because it's trying to sum up different sources of amount to give an overall balance. I think it's different from focussing on a particular adjustment type.

It feels wrong that every time you have a possibly neutral adjustment type you're going to have to add your own method for `getTaxTotal()` to do it explicitly with a different implementation.

We have added quite a few helpful wrapper methods such as `getTaxTotal()` which are useful for templates, but this simply wraps a fixed `return getAdjustmentsTotalRecursively(Adjusment::TAX_ADJUSTMENT);`, it doesn't require a separate implementation.

src/Sylius/Component/Core/Model/OrderItem.php
```diff
@@ -57,4 +57,24 @@ public function equals(BaseOrderItemInterface $item)
             && $item->getVariant() === $this->variant
         );
     }
+
+    /**
+     * Returns sum of neutral and non neutral tax adjustments on order item and total tax of units.
+     *
+     * {@inheritdoc}
+     */
+    public function getTaxTotal()
```
peteward: Yeah just looking at having to add these methods at all 3 levels suggests the design is wrong to me

src/Sylius/Component/Core/Model/Order.php
Zales0123: Sure, but ``ORDER_PROMOTION_ADJUSTMENT``s should be distributed on ``OrderItemUnits`` (see https://github.com/Sylius/Sylius/pull/4768) to fulfill e.g. refund requirements.

peteward: I thought `ORDER_PROMOTION_ADJUSTMENT` was for things like shipping discounts that aren't directly associated with an `OrderItemUnit`.

Can you please give me a definition of `ORDER_PROMOTION_ADJUSTMENT` and how it differs from `ORDER_ITEM_PROMOTION_ADJUSTMENT`?

Zales0123: For shipping there is a special ``SHIPPING_ADJUSTMENT`` ;)
Well, ``ORDER_PROMOTION_ADJUSTMENT`` is an adjustment calculated on ``Order`` level, but distributed on units. The easiest example is **$10 discount on each order** - and this $10 is distributed on all units in order on background. Then, ``ORDER_ITEM_PROMOTION_ADJUSTEMENT`` is also distributed (due to the same reasons), but calculated on ``OrderItem`` level - **$10 discount on each item containing dress**.

peteward: Thanks for the explanation @Zales0123.

So it's really more of a distinction to help promotions as it doesn't offer any clarification outside of this domain.

I meant shipping promotional discounts. From the looks of the code above these will always be `ORDER_PROMOTION_ADJUSTMENT` but stored directly on the `Order` - is that correct?

Zales0123: According to [ShippingDiscountAction](https://github.com/Sylius/Sylius/blob/1b91d94b8aaa404a6e25d2113497f0405a7ca291/src/Sylius/Component/Core/Promotion/Action/ShippingDiscountAction.php#L37) adjustment is indeed applied on ``Order``, but it's type is ``SHIPPING_ADJUSTMENT``, not ``ORDER_PROMOTION_ADJUSTMENT``. So for now, the only adjustment applied *de facto* on ``Order`` is ``SHIPPING_ADJUSTMENT``, rest of them are distributed on units.

peteward: Hmm...

So how do you know how much promotional discount was applied to shipping for reporting?

michalmarcinkowski: @peteward that is correct, the `ORDER_PROMOTION_ADJUSTMENT` applied to `Order` = shipping discount. I thought about about creating `SHIPPING_PROMOTION_ADJUSTMENT`, so we do not assume that all promotion adjustments on `Order` belongs to shipping, but we would need to do the same with taxes. Not sure if we want to create, so many types of adjustments, but we should give it more thoughts later on.

peteward: hmm that doesn't sound great :(  It will get a lot more fun with refunds (which we've implemented).  Relying on a positive or negative sign isn't a great system.

It also means your code above is wrong I think

src/Sylius/Component/Core/Model/Order.php
```diff
@@ -527,13 +527,54 @@ public function getPromotions()
...
      */
-    public function getPromotionsTotalRecursively()
+    public function getShippingTotal()
     {
-        return
-            $this->getAdjustmentsTotalRecursively(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT) +
-            $this->getAdjustmentsTotalRecursively(AdjustmentInterface::ORDER_ITEM_PROMOTION_ADJUSTMENT)
-        ;
+        $shippingTotal = $this->getAdjustmentsTotal(AdjustmentInterface::TAX_ADJUSTMENT);
+        $shippingTotal += $this->getAdjustmentsTotal(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT);
```
peteward: If shipping promotional discounts are truly handled only through `SHIPPING_ADJUSTMENT` then this should be removed?

peteward:
`getRealUnitPrice` => `getDiscountedUnitPrice` or `getUnitPriceAfterDiscounts`? Something more explicit as 'real' sounds a bit vague

src/Sylius/Component/Core/Model/Order.php
peteward: I think that if you can't use `ORDER_PROMOTION_ADJUSTMENT` for shipping discounts then you'll need a third kind of promotional Adjustment - `SHIPPING_PROMOTION_ADJUSTMENT`.  I don't like having lots of different types of adjustment either, but I think it will cause more pain and confusion without this clarity.

src/Sylius/Component/Core/Model/OrderItem.php
mmckelviereiss: Maybe we need some kind of `AdjustablesCalculator` service, whereby we pass in a set of adjustables and we can perform calculation methods like this on it. This implementation here will include refunds too, sometimes we want to know what the tax was before refunds so again requiring a different method. Lots of these types of calculation methods will begin to clutter the Adjustable on all three levels. Separating the calculation logic would be better in my opinion; just a thought.

src/Sylius/Component/Core/Model/Order.php
michalmarcinkowski: But the confusion with tax will still remain. I don't think we should introduce `SHIPPING_TAX_ADJUSTMENT` as it will add another condition to tax calculation/obtaining.

Current logic assumes that all adjustments `ORDER_PROMOTION` adjustments and `TAX` adjustments that are on an `Order` level belongs to shipping. With current implementation it's straightforward and the simplest thing we can do.

Is there any use case where `SHIPPING_PROMOTION` adjustment can be useful?

src/Sylius/Component/Order/Model/Order.php
michalmarcinkowski: But we also need to obtain the adjustments total taking into account the neutrality. E.g.:
```
$shippingTotal = $this->getAdjustmentsTotal(AdjustmentInterface::SHIPPING_ADJUSTMENT);
$shippingTotal += $this->getAdjustmentsTotal(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT);
$shippingTotal += $this->getAdjustmentsTotal(AdjustmentInterface::TAX_ADJUSTMENT); // if tax is already included in shipping fee I don't want to add it once again
```

If we will make `getAdjustmentsTotal` method that will return only sum of non neutral adjustments and then we will have `getAdjustmentsTotalRecursively` that will return sum of all adjustments (both neutral and non neutral) it will be confusing.

Right now `getAdjustmentsTotal()` will return sum of only non neutral adjustments on current level, but `getAdjustmentsTotal($type)` will return sum of all adjustments on current level. This can be confusing. 

How about renaming method `getAdjustmentsTotal` to `getNonNeturalAdjustmentsTotal` and leave current logic of `getAdjustmentsTotalRecursively` and add `getNonNeturalAdjustmentsTotalRecursively` method if needed?

src/Sylius/Component/Core/Model/OrderItem.php
michalmarcinkowski: Current adjustments system implementation isn't perfect, that's why we need this "duplication". We were thinking about introducing `AdjustmentsBag` that will have all these helper methods, but that is a random idea for eventual later refactoring.

Regarding this method I responded above :)

src/Sylius/Component/Core/Model/Order.php
michalmarcinkowski: `SHIPPING_ADJUSTMENT` is the shipping fee, `ORDER_PROMOTION_ADJUSTMENT` on order level is a shipping discount, so in order to know the end cost of shipping we should decrease shipping fee by discount amount, so that is correct.

michalmarcinkowski:
`getDiscountedUnitPrice` - IMO is the best. I was hoping someone will find a better name 😄

src/Sylius/Component/Core/Model/OrderItem.php
michalmarcinkowski: @mmckelviereiss we're thinking about extracting this to a separate service, because you can have different strategies, but we decided to start with this simple implementation and think about improvements later on :) You can easily override this method in your model and implement the logic you need, so I believe it can stay on a model level for now 😉

src/Sylius/Component/Core/Model/Order.php
peteward: Why would you need to introduce a `SHIPPING_TAX_ADJUSTMENT`? There's no requirement for this. Ideally there should be no requirement for `SHIPPING_PROMOTION_ADJUSTMENT` either, it should all just be done as a `PROMOTION_ADJUSTMENT` that is applied to the relevant item (`Order`, `OrderItem`, `OrderItemUnit`), but you needed to separate out Adjustment types for logic internal to promotions.

There are many use cases where you want to know how much discount there has been on shipping, whether it's on a single order, part of a group of orders, a request for total discount or just shipping discounts.

It's more about consistency. If all other promotional discounts are represented by adjustment types, why would you add a special business rule for shipping discounts which is "look for all negative SHIPPING_ADJUSTMENTS". It just feels like a bad design.

In our implementation, and I'm guessing most people's now, the only adjustments that can really be on an `Order` object are shipping adjustments anyway. So any TAX_ADJUSTMENT, or X_PROMOTION_ADJUSTMENT at this level would be for shipping - since all other adjustments are split down to `OrderItemUnits`.

src/Sylius/Component/Core/Model/Order.php
peteward: oh, hold on. 

This contradicts what @Zales0123 said below:

> According to ShippingDiscountAction adjustment is indeed applied on Order, but it's type is SHIPPING_ADJUSTMENT, not ORDER_PROMOTION_ADJUSTMENT. So for now, the only adjustment applied de facto on Order is SHIPPING_ADJUSTMENT, rest of them are distributed on units.

Can you please clarify the following:

Applied shipping promotional discounts are recorded as negative `SHIPPING_ADJUSTMENT` or `ORDER_PROMOTION_ADJUSTMENT`?

Zales0123: @peteward My bad, I looked bad on the action implementation. Of course applied adjustment is not ``SHIPPING_ADJUSTMENT`` but ``ORDER_PROMOTION_ADJUSTMENT``, sorry for the confusion.

peteward: OK :) 

In that case you can ignore my comments here 👍

src/Sylius/Component/Core/Model/Order.php
peteward: and you can ignore my comments here, see above confusion. 👍

michalmarcinkowski: The discount on shipping = `$this->getAdjustmentsTotal(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT);` on an order level (after [this PR](https://github.com/Sylius/Sylius/pull/4768). There are no `"...negative SHIPPING_ADJUSTMENTS"` in current implementation.

> In our implementation, and I'm guessing most people's now, the only adjustments that can really be on an Order object are shipping adjustments anyway. So any TAX_ADJUSTMENT, or X_PROMOTION_ADJUSTMENT at this level would be for shipping - since all other adjustments are split down to OrderItemUnits.

That is exactly the logic we will have after distribution PR is merged. My question was why we need `SHIPPING_PROMOTION_ADJUSTMENT` if we assume that all `ORDER_PROMOTION_ADJUSTMENT`s on order level are actually shipping promotion?

peteward: no, see above, we don't

peteward: Can you clarify the following though:

`ORDER_PROMOTION_ADJUSTMENT` adjustment type is applied ONLY on `Order`?
`ORDER_PROMOTION_ADJUSTMENT`adjustment type is applied on `OrderItem` or `OrderItemUnit` but not on `Order`?

Am I correct here?

michalmarcinkowski: `ORDER_PROMOTION_ADJUSTMENT` = order discount e.g. `$5 off every order above $100` OR `Free shipping for order above $200` - shipping discount.

`ORDER_PROMOTION_ADJUSTMENT` - applied on `Order` = shipping discount
`ORDER_PROMOTION_ADJUSTMENT` - applied to `OrderItem` or `OrderItemUnit` = distributed order discount.

src/Sylius/Component/Order/Model/Order.php
peteward: I think the problem here (and you have already mentioned it) is the lack of clarity around adjustments at the `Order` level.

From my perspective, these can **only** be related to shipping. Any other adjustments would then be associated eventually with `OrderItemUnit`s. 

If you accept this then your use-case above isn't required, because you don't need to do:

```
$shippingTotal = $this->getAdjustmentsTotal(AdjustmentInterface::SHIPPING_ADJUSTMENT);
$shippingTotal += $this->getAdjustmentsTotal(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT);
$shippingTotal += $this->getAdjustmentsTotal(AdjustmentInterface::TAX_ADJUSTMENT); 
```

You just do:

```
$this->getAdjustmentsTotal()
```

Because this is the neutral-adjustment total for Shipping.
It's the same for `OrderItem` and `OrderItemUnit` where you want their totals then you want them neutrally adjusted. It's just perhaps unfortunate that Shipping adjustments are on the `Order` object.

If you reject this rule of only Shipping adjustments directly on an order then I agree that `getNonNeutralAdjustmentsTotal` pattern is probably the solution.

src/Sylius/Component/Core/Model/Order.php
peteward: OK so it's not what they're distributed on, it's a 'type' of promotion effectively - fine.

In which case I think you need to be very careful with the naming of this method. `getOrderPromotionTotal()` suggest to me the total amount of discount for the `Order`. This method which gets only promotions of a specific type excluding shipping is a bit different (and I'm not clear what the use-case is for it, but I guess you must have one).

We might want to distinguish with:

```
function getPromotionTotal();

function getOrderBasedPromotionsTotal();  // this one

function getItemBasedPromotionsTotal();

function getShippingPromotionTotal();
```

I know it's a lot of methods, but might be clearer.

wdyt?

src/Sylius/Component/Order/Model/Order.php
michalmarcinkowski: The only problem with simple `$this->getAdjustmentsTotal()` is when you add your custom adjustment e.g. `COMMISION_FEE` and apply it on `Order` then it will be included in shipping total, but with:
```
$shippingTotal = $this->getAdjustmentsTotal(AdjustmentInterface::SHIPPING_ADJUSTMENT);
$shippingTotal += $this->getAdjustmentsTotal(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT);
$shippingTotal += $this->getAdjustmentsTotal(AdjustmentInterface::TAX_ADJUSTMENT);
```
It will not be included. There is already big assumption that `TAX` and `ORDER_PROMOTION` adjustments belong to shipping, so we should at least not limit the above use case.

However we can write a short note in the docs that if you want to add your own adjustment on `Order` change implementation of `getShippingTotal` logic to "this". Worse customization DX / little implementation overhead, we have to choose 😄  @pjedrzejewski any thoughts on this?

I like the idea to return `adjustmentsTotal` as shipping total since it even does not require to fetch adjustments collection or calculate anything :)

peteward: Yes it's definitely a domain decision and one that I understand. But if you're making assumptions already then I think you either make them fully, or don't make them at all.

It starts to make me wonder if shipping adjustments should have been placed on the `Shipment` instead....

michalmarcinkowski:
Closing in favor of #5084.

PR #5063, opened 2016-05-19
switch 'transform' and 'reverseTransform' in ObjectToIdentifierTranformer

### LINKED ISSUE #5027: [Resource][Form] Why did ObjectToIdentifierTransformer get reversed?
The `DataTransformerInterface` [on github](https://github.com/symfony/form/blob/master/DataTransformerInterface.php) specifies:

* `transform()`  Transforms a value from the original representation to a transformed representation.
* `reverseTransform()` Transforms a value from the transformed representation to its original representation.

This seems logical and sensible to me, but `ObjectToIdentifierTransformer` seems to have been reversed so the opposite happens - and it blows up our forms as a result.

It happened in [this PR](https://github.com/Sylius/Sylius/pull/4702), it seems very wrong to me.

I can only think that the other changes in this PR which add an extra condition for `multiple` means that it wasn't actually ever called after the change so no one noticed it was broken.

Unless I'm wrong somehow?

bendavies:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #5027
| License         | MIT

Reverses what was done in #4702.
Let's see what breaks...

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Form/DataTransformer/ObjectToIdentifierTransformer.php
+++ b/src/Sylius/Bundle/ResourceBundle/Form/DataTransformer/ObjectToIdentifierTransformer.php
@@ -48,6 +48,26 @@ class ObjectToIdentifierTransformer implements DataTransformerInterface
      */
     public function transform($value)
     {
+        if (empty($value)) {
+            return '';
+        }
+
+        $class = $this->repository->getClassName();
+
+        if (!$value instanceof $class) {
+            throw new UnexpectedTypeException($value, $class);
+        }
+
+        $accessor = PropertyAccess::createPropertyAccessor();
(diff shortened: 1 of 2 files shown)


bendavies:
Well, there we go, there are some failures.

patrick-mcdougle:
Should probably use the [ReversedTransformer](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Form/ReversedTransformer.php) for the cases where you need identifier-to-object

bendavies:
@pjedrzejewski can be closed? 3951

pjedrzejewski:
@bendavies Yes, thanks for the ping! :)

PR #5065, opened 2016-05-19
[Doc][ThemeBundle] renamed theme assets command

takeit:
Command `sylius:assets:install` does not exist when using ThemeBundle in a standalone mode. Thus, `sylius:theme:assets:install` should be used.


pjedrzejewski:
Thanks Rafał! 👍

PR #5069, opened 2016-05-19
[Metadata] Add hierarchical archetype support to Product Metadata

### LINKED ISSUE #4483: [Metadata] ProductHierarchyProvider doesn't retrieve Archetypes recursively
The following only retrieves the immediate archetype, but metadata should be resolved recursively for archetypes:

```
    public function getHierarchyByMetadataSubject(MetadataSubjectInterface $product)
    {
        $hierarchy = [
            $product->getMetadataIdentifier(),
        ];

        /** @var ArchetypeInterface $productArchetype */
        $productArchetype = $product->getArchetype();
        if (null !== $productArchetype) {
            $hierarchy[] = $productArchetype->getMetadataIdentifier();
        }

        $hierarchy[] = $product->getMetadataClassIdentifier();
        $hierarchy[] = 'DefaultPage';

        return $hierarchy;
    }
```

If you have an archetype hierarchy such as:

* Accessories
    * Bags
        * Handbags

It should iterate up the tree to find a suitable metadata set. If the product us a `Handbag` it currently just  tries `Handbags` and won't consider metadata for Bags or Accessories if metadata hasn't been written at their specific level.

The fix for this should be trivial, but it does start to bloat out the calls for retrieving metadata, partly because the hierarchy is built up-front without checking if each level has any content first. There's also no caching yet which would be another way of offsetting the efficiency issue.

We should probably just do the obvious fix for now and come back to efficiency later.

peteward:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | fixes #4483 
| License         | MIT

Allows you to define and use product metadata at any level of the Archetype hierarchy.

### DIFF
--- a/src/Sylius/Component/Core/Metadata/HierarchyProvider/ProductHierarchyProvider.php
+++ b/src/Sylius/Component/Core/Metadata/HierarchyProvider/ProductHierarchyProvider.php
@@ -34,6 +34,10 @@ class ProductHierarchyProvider implements MetadataHierarchyProviderInterface
         $productArchetype = $product->getArchetype();
         if (null !== $productArchetype) {
             $hierarchy[] = $productArchetype->getMetadataIdentifier();
+            
+            while ($productArchetype = $productArchetype->getParent()) {
+                $hierarchy[] = $productArchetype->getMetadataIdentifier();
+            }
         }
 
         $hierarchy[] = $product->getMetadataClassIdentifier();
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
Nice, looks good to me! 👍

pjedrzejewski:
Thanks Peter!

PR #5074, opened 2016-05-20
[Metadata] Add hierarchical archetype support to ProductVariant Metadata

peteward:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| License         | MIT

The same as #5069, but for ProductVariant - keeping them consistent.

Could consider injecting the Product in here and using that at some point in future to stop the duplication but this should be fine for now.


src/Sylius/Component/Core/spec/Metadata/HierarchyProvider/ProductVariantHierarchyProviderSpec.php
```diff
@@ -53,6 +53,7 @@ function it_generates_correct_hierarchy_when_product_variant_has_archetype(
         $product->getMetadataIdentifier()->shouldBeCalled()->willReturn('Product-42');
         $product->getMetadataClassIdentifier()->shouldBeCalled()->willReturn('Product');
 
+        $archetype->getParent()->shouldBeCalled()->willReturn(false);
```
pamil: Shouldn't it return `null`?

peteward:
fixed

pjedrzejewski:
Thank you Peter!

PR #5075, opened 2016-05-20
fix mapped superclass subscribers for overidden resources

### LINKED ISSUE #5040: What is the purpose of Load(ORM/ODM)MetadataSubscriber?

bendavies:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    |yes
| BC breaks?      |yes
| Related tickets | fixes #5040 (again)
| License         | MIT

I think this is correct now.

1. Convert MappedSuperClass to Entity if the class is the configured Sylius resource.
2. If class is not a MappedSuperClass, move associations  from MappedSuperClass parents, but only if parent is a Sylius entity.
3. If Class is MappedSuperClass, remove associations, but only if class is a Sylius entities.

cc @liverbool

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/EventListener/ODMMappedSuperClassSubscriber.php
+++ b/src/Sylius/Bundle/ResourceBundle/EventListener/ODMMappedSuperClassSubscriber.php
@@ -39,9 +39,6 @@ class ODMMappedSuperClassSubscriber extends AbstractDoctrineSubscriber
     public function loadClassMetadata(LoadClassMetadataEventArgs $eventArgs)
     {
         $metadata = $eventArgs->getClassMetadata();
-        if (false === $this->isSyliusClass($metadata)) {
-            return;
-        }
 
         $this->convertToDocumentIfNeeded($metadata);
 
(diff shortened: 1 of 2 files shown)


lbaey:
+1 Please Merge, this fixes all my mapping validation problems 👍

liverbool:
Yah! Green again.
screen shot 2559-05-20 at 15 28 39

i don't understand why there are no problems on https://github.com/pjedrzejewski/Sylius/tree/reproduce-issue-with-mapping.

pjedrzejewski:
Great, thank you Ben! 👍

cdaguerre:
With the current implementation, the `DoctrineTargetEntitiesResolverPass` won't map interfaces that are not in the `Sylius\` namespace anymore, which is a BC break (a quite big one for us ;)
Actually, the question is more whether it's a registered sylius resource than a sylius class.
I think this means you have to check if the class is :
- a resource registered with sylius; or,
- a parent class of any registered resource.

If you agree that's the right way to check, I'd be glad to make a PR.
Any chance this could be released as a bug fix on 0.18?

cdaguerre:
Maybe we could use the `getModelNamespace()` method of the `AbstractResourceBundle` to keep track of classes that should be considered as sylius classes by the [AbtsractDoctrineSubscriber](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ResourceBundle/EventListener/AbstractDoctrineSubscriber.php#L43)

bendavies:
Sorry @cdaguerre . was hard to track all possible outcomes here as there were no tests so i was flying blind. I'd welcome a PR. I'd struggle to fix properly as I'm not running full sylius in my app. Will be happy to take a look once you've done it.

cdaguerre:
I'm just not sure about the right way to do it...
We could also enhance the resource registry to let it answer this question.
Let's hear what @pjedrzejewski thinks ;)

PR #5084, opened 2016-05-20
[Order][Core] Order improvements

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets |
| License         | MIT

This is continuation of PR https://github.com/Sylius/Sylius/pull/5062.

Already done:
- ``getTaxTotal`` method to Order which returns sum of tax adjustments regardless neutrality - useful when you want to know the value of tax that was charged
- ``getShippingTotal`` method to Order which returns shipping fee together with shipping tax decreased by shipping discount - useful when you want to know the end value that was charged for shipping
- made ``getAdjustmentsTotal`` and ``getAdjustmentsTotalRecursively`` methods consistent - when called without argument they returned sum of only non neutral adjustments, when called with ``$type`` argument all adjustments were summed (both non neutral and neutral)
- removed ``getPromotionsTotalRecursively`` in favor of ``getOrderPromotionTotal`` - this method returns total amount of order level promotion, but does not include item level promotions and does not include shipping discount which is included in ``getShippingTotal``.
- ``getDiscountedUnitPrice`` method to OrderItem that will return the unit price decreased by item promotions.
- ``getSubtotal`` method to OrderItem that will return ``$discountedUnitPrice`` * ``$quantity``

To do:
- [x] use newly implemented methods in **order** template
- [x] write scenarios for checking new **order** template data

### DIFF
--- a/features/promotion/applying_promotion_rules/reapplying_promotion_on_cart_change.feature
+++ b/features/promotion/applying_promotion_rules/reapplying_promotion_on_cart_change.feature
@@ -30,8 +30,7 @@ Feature: Reapplying promotion on cart change
         And I chose "DHL" shipping method
         When I change shipping method to "FedEx"
         Then my cart total should be "€100.00"
-        And my cart shipping fee should be "€30.00"
-        And my discount should be "-€30.00"
+        And my cart shipping fee should be "€0.00"
 
     @ui
     Scenario: Receiving discount after removing an item from the cart and then adding another one
--- a/features/promotion/applying_promotion_rules/reapplying_promotion_on_cart_change.feature
+++ b/features/promotion/applying_promotion_rules/reapplying_promotion_on_cart_change.feature
@@ -30,7 +30,7 @@ Feature: Reapplying promotion on cart change
         And I chose "DHL" shipping method
         When I change shipping method to "FedEx"
         Then my cart total should be "€100.00"
-        And my cart shipping fee should be "€0.00"
+        And my cart shipping should be for free
 
     @ui
     Scenario: Receiving discount after removing an item from the cart and then adding another one
--- a/features/promotion/receiving_discount/receiving_percentage_discount_on_shipping.feature
+++ b/features/promotion/receiving_discount/receiving_percentage_discount_on_shipping.feature
@@ -17,7 +17,7 @@ Feature: Receiving percentage discount on shipping
         When I add product "PHP T-Shirt" to the cart
         And I proceed selecting "DHL" shipping method
         Then my cart total should be "€108.00"
-        And my cart shipping fee should be "€8.00"
+        And my cart shipping total should be "€8.00"
 
     @ui
     Scenario: Receiving free shipping
@@ -25,4 +25,4 @@ Feature: Receiving percentage discount on shipping
         When I add product "PHP T-Shirt" to the cart
         And I proceed selecting "DHL" shipping method
         Then my cart total should be "€100.00"
-        And my cart shipping fee should be "€0.00"
+        And my cart shipping total should be "€0.00"
--- a/features/promotion/receiving_discount/receiving_percentage_discount_on_shipping.feature
+++ b/features/promotion/receiving_discount/receiving_percentage_discount_on_shipping.feature
@@ -17,8 +17,7 @@ Feature: Receiving percentage discount on shipping
         When I add product "PHP T-Shirt" to the cart
         And I proceed selecting "DHL" shipping method
         Then my cart total should be "€108.00"
-        And my cart shipping fee should be "€10.00"
-        And my discount should be "-€2.00"
+        And my cart shipping fee should be "€8.00"
 
     @ui
     Scenario: Receiving free shipping
@@ -26,13 +25,4 @@ Feature: Receiving percentage discount on shipping
         When I add product "PHP T-Shirt" to the cart
         And I proceed selecting "DHL" shipping method
         Then my cart total should be "€100.00"
-        And my cart shipping fee should be "€10.00"
-        And my discount should be "-€10.00"
-
-    @ui
-    Scenario: Not receiving percentage discount on shipping before selecting shipping method
-        Given the promotion gives "100%" discount on shipping to every order
-        When I add product "PHP T-Shirt" to the cart
-        Then my cart total should be "€100.00"
         And my cart shipping fee should be "€0.00"
-        And there should be no discount
--- a/src/Sylius/Component/Core/Model/OrderItem.php
+++ b/src/Sylius/Component/Core/Model/OrderItem.php
@@ -81,21 +81,13 @@ class OrderItem extends CartItem implements OrderItemInterface
     /**
      * {@inheritdoc}
      */
-    public function getDiscountedUnitPrice()
-    {
-        $orderItemDiscounts = 0;
-        foreach ($this->units as $unit) {
-            $orderItemDiscounts += $unit->getAdjustmentsTotal(AdjustmentInterface::ORDER_ITEM_PROMOTION_ADJUSTMENT);
-        }
-
-        return $this->unitPrice - $orderItemDiscounts;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function getSubtotal()
     {
-        return $this->getDiscountedUnitPrice() * $this->quantity;
+        $subtotal = $this->unitPrice * $this->quantity;
+        foreach ($this->units as $unit) {
+            $subtotal += $unit->getAdjustmentsTotal(AdjustmentInterface::ORDER_ITEM_PROMOTION_ADJUSTMENT);
+        }
+
+        return $subtotal;
     }
 }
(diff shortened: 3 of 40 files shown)


src/Sylius/Component/Core/Model/OrderItem.php
```diff
@@ -57,4 +57,45 @@ public function equals(BaseOrderItemInterface $item)
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDiscountedUnitPrice()
+    {
+        $orderItemDiscounts = 0;
+        foreach ($this->units as $unit) {
+            $orderItemDiscounts += $unit->getAdjustmentsTotal(AdjustmentInterface::ORDER_ITEM_PROMOTION_ADJUSTMENT);
```
peteward: I thought promotional discounts were all applied at the `OrderItemUnit` level?  In which case I'm not sure what this method really means.

src/Sylius/Component/Core/Model/OrderItem.php
```diff
@@ -57,4 +57,45 @@ public function equals(BaseOrderItemInterface $item)
...
+
+        return $this->unitPrice - $orderItemDiscounts;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSubtotal()
+    {
+        return $this->getDiscountedUnitPrice() * $this->quantity;
```
peteward: Is this total after discount but before tax?

Again, not sure what this means as it's implemented here.

src/Sylius/Component/Core/Model/OrderItemUnit.php
```diff
@@ -135,4 +135,20 @@ public function setShippingState($state)
     {
         $this->shippingState = $state;
     }
+
+    /**
+     * Returns sum of neutral and non neutral tax adjustments.
+     *
+     * {@inheritdoc}
+     */
+    public function getTaxTotal()
```
peteward: As per our discussion, I recommend creating a `getNeutrallyInclusiveAdjustmentsTotalRecursively` method at the Order level to cater for a wider scope of possibilities and then add `getTaxTotal()` methods which wraps these at each level.

peteward:
Although 0.18 is released and we intend to freeze on this for now, I will try to mimic most of these changes as they are fundamental to finance and definitely wise to try and keep aligned as far as possible.

src/Sylius/Component/Core/Model/OrderItem.php
Zales0123: Surely there is a bug, as ``ORDER_ITEM_PROMOTION``s are distributed on units, we should not take whole this discount to count ``discountedUnitPrice``... However, there is a problem, as in specific cases promotions distributed on units can be not equal. We'll rethink how to solve this issue ;)

Zales0123:
@peteward, according to [your comment](https://github.com/Sylius/Sylius/pull/5084#discussion_r64130491)
What we want to achieve right now, is to create detailed order item template with this information:
- Item - item product name and image
- unit price - base price of each unit in item
- discounted unit price - unit price lowered by ``ORDER_ITEM_PROMOTION``
- quantity
- subtotal - discounted price of each unit multiplied by quantity
- tax - taxes distributed on this item
- discount - discount from ``ORDER_PROMOTION`` adjustments
- total - subtotal + taxes + discount

so newly added methods are here to help us getting required information without struggling with some nasty filters etc. in templates.

src/Sylius/Component/Core/Model/OrderItem.php
michalmarcinkowski: @peteward the discounts are applied at the `OrderItemUnit` level. This method should return discounted unit price e.g. `$1 discount on each T-Shirt`
* `getUnitPrice` = $10
* `getDiscountedUnitPrice` = $9

Proposed implementation:
```php
if ($this->units->isEmpty()) {
    return $this->unitPrice;
}

$unit = $this->units->first();
$unitDiscount = $unit->getNonNeutralAdjustmentsTotal(AdjustmentInterface::ORDER_ITEM_UNIT_PROMOTION_ADJUSTMENT);

return $this->unitPrice + $unitDiscount; // we add discount since it is negative
```

@Zales0123 we should introduce the `ORDER_ITEM_UNIT_PROMOTION` adjustment and use it instead of `ORDER_ITEM_PROMOTION`. Unit promotion adjustment should be equally distributed on all units, so there shouldn't be any unit with different discount. This type of adjustment should be always used for "per unit" discounts. But let's do this change in a separate PR.

peteward: Is the distribution always equal?

Are there ever any rounding issues where you distribute a fixed discount (say $10 across 3 units) and you have 0.01 extra on one unit?

src/Sylius/Component/Core/Model/OrderItem.php
michalmarcinkowski: For prices with tax included, "subtotal" will return amount with tax, for prices with taxes not included it will return amount without tax.

peteward: OK I understand, a very difficult one to name, and I can't think of anything better than subtotal! :)

src/Sylius/Component/Core/Model/OrderItem.php
michalmarcinkowski: Right now there is no action that would give you a fixed discount on `OrderItem`, all actions are per unit. There is no rounding at all because the discount is always equal on all units.

peteward: I was more thinking of $10 off your order split between 1 `OrderItem` with 3 `OrderItemUnits` but in this case I guess it's an `ORDER_PROMOTION_ADJUSTMENT`.

And you're excluding these because you want only the discounts that were directly associated with this `OrderItem` not those that were derived against it's units as a way of financially associating discount?

michalmarcinkowski: Exactly. @Zales0123 we should describe all that in the docs :)

peteward: OK yeah it makes sense, will be challenging to document the whole financials at some point :)

features/promotion/applying_promotion_rules/reapplying_promotion_on_cart_change.feature
```diff
@@ -30,8 +30,7 @@ Feature: Reapplying promotion on cart change
         And I chose "DHL" shipping method
         When I change shipping method to "FedEx"
         Then my cart total should be "€100.00"
-        And my cart shipping fee should be "€30.00"
-        And my discount should be "-€30.00"
+        And my cart shipping fee should be "€0.00"
```
michalmarcinkowski: `And the shipping should be for free`?

features/promotion/receiving_discount/receiving_percentage_discount_on_shipping.feature
```diff
@@ -17,22 +17,12 @@ Feature: Receiving percentage discount on shipping
         When I add product "PHP T-Shirt" to the cart
         And I proceed selecting "DHL" shipping method
         Then my cart total should be "€108.00"
-        And my cart shipping fee should be "€10.00"
-        And my discount should be "-€2.00"
+        And my cart shipping fee should be "€8.00"
```
michalmarcinkowski: "shipping fee" -> "shipping total" - let's be consistent with new naming.

src/Sylius/Component/Core/Model/OrderItem.php
```diff
@@ -57,4 +57,37 @@ public function equals(BaseOrderItemInterface $item)
...
+     * {@inheritdoc}
+     */
+    public function getSubtotal()
+    {
+        $subtotal = $this->unitPrice * $this->quantity;
+        foreach ($this->units as $unit) {
+            $subtotal += $unit->getAdjustmentsTotal(AdjustmentInterface::ORDER_ITEM_PROMOTION_ADJUSTMENT);
+        }
+
+        return $subtotal;
```
michalmarcinkowski: `return $this->getDiscountedUnitPrice() * $this->quantity;`

Zales0123: It would be nice if we had ``getDiscountedUnitPrice()`` method 😄  its not implemented yet, as it is going to be introduced in separate PR (as I mentioned in [this comment](https://github.com/Sylius/Sylius/pull/5084/#issuecomment-220902478) UPDATE).

michalmarcinkowski: Yeah... I forgot about that 😄  so this implementation can stay for now :)

michalmarcinkowski:
We have to remember to add back all discount assertion that are removed in this PR, when implementing discounted price.

src/Sylius/Component/Core/Model/OrderItemUnit.php
Zales0123: This is good idea, but maybe we could introduce this in separate PR, to not hold this one longer and to make review simpler? This could also give us possibility to discuss about such an approach in more comfortable circumstances ;)

michalmarcinkowski:
Thank you Mateusz!

PR #3339, opened 2015-09-23
[Promotion] Making coupon code length configurable.

Arminek:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | no
| License       | MIT

Rebased PR from @Richtermeister https://github.com/Sylius/Sylius/pull/2389

PR #5087, opened 2016-05-23
Coupon generation fix

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | mentioned in [comment](https://github.com/Sylius/Sylius/pull/3339/files#r64144189)
| License         | MIT


pamil:
This way memory taken by generated codes will never be freed until the process dies. We can remove `$this->generatedCoupons` property and pass a local variable as an argument to the other functions without affecting the performance.

pjedrzejewski:
@pamil 👍 , but in a separate PR.

pjedrzejewski:
Thanks Łukasz! /cc @peteward I will retag v0.18.0 so that it includes this. :)

peteward:
Thanks for the quick turnaround on this guys! 👍

PR #5088, opened 2016-05-23
Bump version to 0.19-dev

pjedrzejewski:


michalmarcinkowski:
👍

loic425:
I thought 0.18 was the last one. Did you change your roadmap ?

pjedrzejewski:
@loic425 We decided that there must be one more version - v0.19.0, because it will introduce ShopBundle, which is a BC break (template's variables will mostly be compatible, but routing will change and template names) too big for alpha. We have a lot of new shop implemented already, so should not be long until we tag v0.19.0 and switch to alpha.

PR #5112, opened 2016-05-25
[RBAC] Ensure Permissions::__toString() returns a string 

peteward:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

Since the description field is nullable, it's possible that description could be null.

If so it blows up the Edit Role form as Symfony ChoiceType requires a string representation.

### DIFF
--- a/src/Sylius/Component/Rbac/Model/Permission.php
+++ b/src/Sylius/Component/Rbac/Model/Permission.php
@@ -78,7 +78,7 @@ class Permission implements PermissionInterface
 
     public function __toString()
     {
-        return $this->description;
+        return $this->code;
     }
 
     /**


src/Sylius/Component/Rbac/Model/Permission.php
```diff
@@ -78,7 +78,7 @@ public function __construct()
 
     public function __toString()
     {
-        return $this->description;
+        return (string) $this->description;
```
michalmarcinkowski: I think the best fix will be to return `code` instead of `description`, wdyt?

peteward:
updated

pjedrzejewski:
Rbac has been removed.

PR #5117, opened 2016-05-25
[Fixture] Set EUR as base currency in fixtures

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

While loading fixtures (without *Sylius* installation) none of currencies has been set as **base**, which led to some nasty bugs like this one:
zrzut ekranu 2016-05-25 o 10 52 51


michalmarcinkowski:
Thanks!

PR #5119, opened 2016-05-25
[TaxonomyBundle] Fix display bug in choice list and enlarge area

psyray:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | none
| License         | MIT

Taxon choice list is badly sorted by id.
To keep tree consistent I've made this PR

**My taxon list**
selection_490

**Before PR**
selection_488

**After PR**
selection_489

I've also enlarge the choice list height on product page, it's more readable whan you have a lot of taxons

Enjoy ;)

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Form/Type/LegacyProductType.php
+++ b/src/Sylius/Bundle/CoreBundle/Form/Type/LegacyProductType.php
@@ -42,6 +42,7 @@ class LegacyProductType extends BaseProductType
             ->add('taxons', 'sylius_taxon_choice', [
                 'label' => 'sylius.form.product.taxons',
                 'multiple' => true,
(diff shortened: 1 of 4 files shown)


pjedrzejewski:
@psyray I noticed this bug recently too, quite annoying! I think that proper fix would be on the query level (sort taxons properly), wdyt?

psyray:
I think the same thing before working on this PR, it's more elegant and a good practice to create a method in the Taxon repository.
You convinced me ;)

I will rework on it.
Sorting on the tree left seems to be good no ?

psyray:
Done !
More concise, more elegant, same result ;)

pjedrzejewski:
Nice, thanks Raynald! 👍

PR #5121, opened 2016-05-25
[Metadata] Fixes, Multiple Types, Improved Twig Interface

### LINKED ISSUE #4481: [Metadata] No translation support
My initial thought was that this would be very challenging but it's probably not _that_ bad.

I'm not sure whether this would follow the generic pattern (I haven't caught up with the recent refactoring of Translations) or just use a composite key of `id` & `locale`.

The UI would probably need the most work, but could be considered as part of other rework.

I think this is extremely important, there's not much point in a wonderful SEO Metadata feature if you can't translate it, it would be unusable to anyone with a multi-language store.

### LINKED ISSUE #4482: [Metadata] Using DefaultPage sitewide
There are 2 intertwined issues here:
- You can't just render the page defaults without having an object implementing the MetadataSubjectInterface to pass (so you can't use your page defaults on any page).
- SEO Metadata settings should be removed once you can as they are confusing and duplication.

On the product page currently it is using this quite ugly looking call:

```
{% block head %}
    {{ sylius_metadata_render(product, null, { 'defaults': {
        'title': settings.title|default('sylius.meta.frontend_title'|trans),
        'description': settings.meta_description|default('sylius.meta.frontend_description'|trans),
        'keywords': settings.meta_description|default('sylius.meta.frontend_keywords'|trans)
    } })|raw }}
{% endblock %}
```

The `defaults` are being passed in through the template which shouldn't be necessary.
- We should remove these settings entirely from Sylius and rely on the DefaultPage SEO Container settings.
- This special root SEO settings should not be deletable, to prevent you destroying SEO for your site. They should also probably be defaulted on install via Fixtures.
- Any rendering object could then call the page as follows because defaults will always be used as a final fallback:

```
{% block head %}
    {{ sylius_metadata_render(product) }}
{% endblock %}
```
- The default call for the parent theme layout should then be as follows so that if a subject is not passed, it just uses the `DefaultPage`:

```
{% block head %}
    {{ sylius_metadata_render() }}
{% endblock %}
```

We could also consider renaming this twig function for more clarity to something like `sylius_page_metadata_render` because I can see many more uses of rendering like ALT tags for images and clearly distinguishing from these would be sensible.

Based on #4481 this would also mean that SEO Metadata will be translatable sitewide and you don't have to worry about trying to accommodate this through settings.

### LINKED ISSUE #4487: [Metadata] Support for ImageMetadata
The other important use-case for this feature is adding a consistent interface to add HTML attributes such as alt and title tags to product images whenever they are rendered.

Something like
```
    <img src="X.jpg" {{ sylius_metadata_render(product, 'image') }} />
```

As there's no documentation for this feature yet, I'm just glancing at the code. It looks like we would need to add the following:

* Custom Model
* Custom Renderer
* Form type

I'm not sure then if it needs a separate Twig extension function to specifically call for the ImageMetadata type or not?

Finally caching is definitely going to be an issue with this - if you're displaying 200 products on a page it's going to be working very hard...

Any pointers here would be great.

### LINKED ISSUE #5101: [Metadata] Child MetadataSubjects of different types throws Exception
```
abstract class AbstractMetadata implements MetadataInterface
{
    /**
     * {@inheritdoc}
     */
    public function merge(MetadataInterface $metadata)
    {
        if (!$metadata instanceof $this || !$this instanceof $metadata) {
            throw new \InvalidArgumentException(
                sprintf(
                    'You can only merge instances of the same classes. Tried to merge "%s" with "%s".',
                    get_class($this),
                    get_class($metadata)
                )
            );
        }
```

The problem with this is the use-case of a child subject within the hierarchy having a different type of Twitter Card.

e.g. You could have a `ProductArchetype` with a `SummaryCard` but then override this at the individual `Product` level with a `SummaryLargeImageCard`.

The exception could be caught, but because this is a recursive action it would break all outstanding merging. 

My proposal is not to throw an Exception and instead just return silently to allow the recursive processing to continue. It won't merge something it can't so should be harmless.

peteward:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | fixes #5101, fixes #4487, fixes #4482, fixes #4481 
| License         | MIT

**Please don't review for code style.**

This PR is intended as the first set of changes to take @pamil's excellent code architecture into an awesome feature. 

I need to complete this work and a lot more for my project. I am willing to contribute this back if there is interest and you agree with my rough approach. If not I will not make any further PR's and just continue the work privately - it's up to you @pamil @pjedrzejewski @michalmarcinkowski.

### Caveats
* No tests
* No intentional code styling
* A few comments left in that need neatening up.

If interested, I'll sort these points out - please don't review for style.

### Highlights:
* Main change is to introduce the `type` dimension for Metadata so that there can be more than just `Page` metadata (see #4487). It can be used as a system for generating product Urls, image alt tags, titles etc. These have NOT been implemented yet, but the support is there to make this simple to drop in.
* Rework around the TwigExtension and how the metadata is rendered (#4482).
    * A dedicated `sylius_metadata_render_page()` method to simplify calling.
    * The `defaults` argument to rendering methods has been reversed so that now you pass `values` which override the metadata properties.
    * Defaults are handled under the hood as the `DefaultPage` metadata. 
    * `values` are specific parts of the metadata that you want to change on the page (which makes much more sense and the Hierarchy handles defaults).
    * Handling the DefaultPage metadata when there is none for your object removes the edge-case requirement for a Core MetadataProvider.
    * Made all twig methods safe for HTML rendering
* Other fixes:
    * Optimised GenericTwitterMetadataRenderer for large array to be static
    * Fixed merge exception when using different types of Twitter card at different levels of the hierarchy.
* Removed the settings related to metadata and make the Sylius defaults installed as `DefaultPage` metadata instead.
* Translations support. This simple version uses the Translation system power. It would probably be better to have translations that work by hierarchy (as language is probably king in SEO, over fallbacks to other languages). However this is a good fully working start.
* Caching support. I've implemented this at the MetadataAccessor level.

### Twig Calling Convention Changes

Previously we had this quite heavy calling convention with defaults:

```
   {{ sylius_metadata_render(product, null, { 'defaults': {
       'title': settings.title|default('sylius.meta.frontend_title'|trans),
       'description': settings.meta_description|default('sylius.meta.frontend_description'|trans),
       'keywords': settings.meta_description|default('sylius.meta.frontend_keywords'|trans)
   } })|raw }}
```

Now we can simply do this for a product page:
```
   {{ sylius_metadata_render_page(product) }}
```

If you want to override parts of the default metadata for a particular page (maybe a stores page):

```
    {{ sylius_metadata_render_page(null, { 'values': {
        'title': 'Sylius - Our lovely stores across the Globe',
        'keywords': ['Sylius', 'stores' 'global']
    ) }}
```

The same conventions have been applied to individual property fetching or rendering.

### Remaining steps

* Gracefully handling rendering exceptions where the underlying data doesn't work in the metadata formula.

There's a lot more to do with the Admin UI but since that is being rewritten I won't be spending any time on this beyond what is required.


peteward:
Updated with:
* Removal of Metadata from general settings.
* Translation support
* Caching

michalmarcinkowski:
We no longer have Metadata component and bundle

PR #5127, opened 2016-05-27
[HOTFIX] Fix core data collector

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

It fix following issue: 
zrzut ekranu 2016-05-27 o 12 35 05

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/DataCollector/SyliusDataCollector.php
+++ b/src/Sylius/Bundle/CoreBundle/DataCollector/SyliusDataCollector.php
@@ -11,7 +11,6 @@
 
 namespace Sylius\Bundle\CoreBundle\DataCollector;
 
-use Sylius\Bundle\CoreBundle\Kernel\Kernel;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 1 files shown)


lchrusciel:
I am not sure, if it is a best solution, but for now it would be enough. Overriding const in AppKernel smells bad, but it is topic for separate RFC.

michalmarcinkowski:
Thanks, that is a good fix! 👍

pjedrzejewski:
We cannot assume there is an AppKernel class...

michalmarcinkowski:
@pjedrzejewski good point, also the version should be taken from `Core` since I can have `0.18` in my `\AppKernel` and update the `Core` to `0.19`, but the version in the collector will remain `0.18` unless I change it manually...

lchrusciel:
@pjedrzejewski @michalmarcinkowski I know, but it was the simplest and fastest solution. As I wrote before, we have a time now to figure out a better one.

PR #5128, opened 2016-05-27
[Sylius] Fixes for changes from #5058

pamil:
Related PR: #5058


src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml
```diff
@@ -21,6 +21,7 @@ imports:
     - { resource: "@SyliusUserBundle/Resources/config/app/config.yml" }
     - { resource: "@SyliusVariationBundle/Resources/config/app/config.yml" }
 
+    - { resource: "parameters.yml" }
```
michalmarcinkowski: These parameters should not override `app/config/parameters.yml`, so we should load core bundle config before app them.

michalmarcinkowski:
Looks good to me 👍  
/cc @pjedrzejewski

pjedrzejewski:
Thanks Kamil!

PR #5137, opened 2016-05-29
Added has and hasByClass methods to Metadata registry

### LINKED ISSUE #5135: [Resource] (RFC) Metadata Registry - hasByClass / specific exception when not found.
Currently when I need to determine if Sylius has metadata for a specific object I need to do the following:

``` php
try {
    $metadata = $this->registry->getByClass($classFqn);
} catch (\InvalidArgumentException $e) {
    return false;
}
```

This is not ideal as an `InvalidArgumentException` is ambiguous.

I propose adding a `hasByClass` method or throwing a `ResourceMetadataNotFound`
exception.

dantleech:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no/yes |
| Related tickets | fixes #5135 |
| License | MIT |
- Adds `has` and `hasByClass` methods to registry and registry interface.
- Indexes by class in addition to alias.

Potential BC break if anybody was implementing the `RegistryInterface`

### DIFF
--- a/src/Sylius/Component/Resource/Metadata/Registry.php
+++ b/src/Sylius/Component/Resource/Metadata/Registry.php
@@ -22,6 +22,11 @@ class Registry implements RegistryInterface
     private $metadata = [];
 
     /**
+     * @var array
+     */
+    private $metadataByClass = [];
+
+    /**
      * {@inheritdoc}
      */
     public function getAll()
@@ -34,7 +39,7 @@ class Registry implements RegistryInterface
      */
     public function get($alias)
     {
-        if (!array_key_exists($alias, $this->metadata)) {
+        if (!isset($this->metadata[$alias])) {
             throw new \InvalidArgumentException(sprintf('Resource "%s" does not exist.', $alias));
         }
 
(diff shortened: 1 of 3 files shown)


src/Sylius/Component/Resource/Metadata/Registry.php
```diff
@@ -34,7 +39,7 @@ public function getAll()
      */
     public function get($alias)
     {
-        if (!array_key_exists($alias, $this->metadata)) {
+        if (!isset($this->metadata[$alias])) {
```
Zales0123: You could use ``!$this->has($alias)`` here.

src/Sylius/Component/Resource/Metadata/Registry.php
```diff
@@ -44,15 +49,29 @@ public function get($alias)
...
+    /**
+     * {@inheritdoc}
+     */
     public function getByClass($className)
     {
-        foreach ($this->metadata as $metadata) {
-            if ($className === $metadata->getClass('model')) {
-                return $metadata;
-            }
+        if (!isset($this->metadataByClass[$className])) {
```
Zales0123: And here ``!$this->hasByClass($className)``.

src/Sylius/Component/Resource/spec/Metadata/RegistrySpec.php
```diff
@@ -52,12 +54,24 @@ function it_throws_an_exception_if_resource_is_not_registered()
...
     {
         $metadata->getAlias()->willReturn('app.shipping_method');
+        $metadata->getClass('model')->willReturn('ShippingMethod');
 
         $this->add($metadata);
 
         $this->get('app.shipping_method')->shouldReturn($metadata);
     }
 
+    function it_should_say_if_it_has_metadata_for_a_specific_metadata(MetadataInterface $metadata)
```
Zales0123: As method is ``has($alias)`` maybe it should be named ``it_should_say_if_it_has_metadata_for_a_specific_alias``?

src/Sylius/Component/Resource/Metadata/Registry.php
dantleech: True, but its (admitedly insignificant) overhead. Don't see any advantage in using the pulic API when both pathways are tested. But don't feel strongly about it.

dantleech:
Application tests failing with:

```
When I browse orders # features/order/managing_orders/browsing_orders.feature:19
      Could not open the page: "http://localhost:8080/admin/orders/". Received an error status code: 500 (Sylius\Behat\Page\UnexpectedPageException)
```

dantleech:
Updated, havn't changed the internal method calls to use the public API, but can do if you feel strongly about it.

dantleech:
Rebased.

PR #5142, opened 2016-05-30
[Promotion][Core] Discounted unit price feature

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

- changed ``item promotions`` to ``unit promotions`` as they're applied on unit level
- changed ``UnitPercentageDiscountAction`` to count discount on unit level, rather than on item
- implement ``getDiscountedUnitPrice`` method on ``OrderItem``, which consist of unit price lowered by ``ORDER_UNIT_PROMOTION``s on one unit

### DIFF
--- a/src/Sylius/Component/Core/Model/OrderItem.php
+++ b/src/Sylius/Component/Core/Model/OrderItem.php
@@ -79,10 +79,16 @@ class OrderItem extends CartItem implements OrderItemInterface
     }
 
     /**
+     * Returns single unit price lowered by order unit promotions (each unit must have the same unit promotion discount)
+     *
      * {@inheritdoc}
      */
     public function getDiscountedUnitPrice()
     {
+        if ($this->units->isEmpty()) {
+            return $this->unitPrice;
+        }
+
         return
             $this->unitPrice +
             $this->units->first()->getAdjustmentsTotal(AdjustmentInterface::ORDER_UNIT_PROMOTION_ADJUSTMENT)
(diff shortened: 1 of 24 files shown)


src/Sylius/Component/Core/Model/OrderItem.php
```diff
@@ -81,13 +81,19 @@ public function getTaxTotal()
...
      * {@inheritdoc}
      */
-    public function getSubtotal()
+    public function getDiscountedUnitPrice()
     {
-        $subtotal = $this->unitPrice * $this->quantity;
-        foreach ($this->units as $unit) {
-            $subtotal += $unit->getAdjustmentsTotal(AdjustmentInterface::ORDER_ITEM_PROMOTION_ADJUSTMENT);
-        }
+        return
```
michalmarcinkowski: We should add 
```php
if ($this->units->isEmpty()) {
    return $this->unitPrice;
}
```

src/Sylius/Component/Core/Model/OrderItem.php
```diff
@@ -81,13 +81,19 @@ public function getTaxTotal()
     /**
      * {@inheritdoc}
      */
-    public function getSubtotal()
+    public function getDiscountedUnitPrice()
```
michalmarcinkowski: We should also add some relevant comments to these methods, as it's done [here](https://github.com/Zales0123/Sylius/blob/348885f8364a28ce5c3f35d5b4b617dc1026f1f5/src/Sylius/Component/Core/Model/OrderItem.php#L62).

peteward:
Will `ORDER_ITEM_PROMOTION` type still be used after these changes?

michalmarcinkowski:
In the Core it will not be used in the nearest future, maybe later we will implement some order item promotions, but I don't think it will happen anytime soon.

pjedrzejewski:
Thank you Mateusz, nice work!

PR #5143, opened 2016-05-30
[Grid] Submit the data to fill the form properly

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

When filtering by entities, we get the ids, not objects, so passing the criteria as data was not enough to preselect values properly.


michalmarcinkowski:
Thanks Paweł! 👍

PR #5155, opened 2016-06-01
[UiBundle] Fix choice type hardcoded class

TheMadeleine:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

Using twig form type rendering we couldn't overwrite the class of Choice Types' rows. Now it will append the class.


pjedrzejewski:
Thanks Magda!

PR #5161, opened 2016-06-02
[Resource] Updating resources response code

bendavies:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no?
| Related tickets | 
| License         | MIT

When updating resources, either 200 `OK` or 204 `No Content` can be returned. But we were mixing them, returning content with 204.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
@@ -314,7 +314,7 @@ class ResourceController extends Controller
             $this->eventDispatcher->dispatchPostEvent(ResourceActions::UPDATE, $configuration, $resource);
 
             if (!$configuration->isHtmlRequest()) {
-                return $this->viewHandler->handle($configuration, View::create($resource, 204));
+                return $this->viewHandler->handle($configuration, View::create(null, 204));
             }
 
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
Actually, I think that for convenience, we could return 200 with the updated entity. I always assumed 204 is the only correct for update most of the time, but seems like no:

diagram

WDYT? That would be 1 less call for APIs.

bendavies:
I actually did that first, but then switched to 204...hah. Whatever you want - I was under the impression 204 was fine for updating...

bendavies:
@pjedrzejewski shall i switch it? The tests are written in a way that they expect 204 (they don't assert response contents, only code)

michalmarcinkowski:
@bendavies From practical point of view returning the updated entity will be useful when e.g. applying state machine transition and getting the next state of the entity in the response already without need to query for it in the next call. On the other hand there are few situations where we apply transitions and in majority we don't need the updated entity in the response (we already have all the data). The 204 responses will be faster since they don't need to serialize the entities and send any data to the receiver, so I would leave `204` and empty response. I think @pjedrzejewski will agree with me.

Later we can think about making this configurable since e.g. almost all cart operations will require to obtain the updated cart after update (recalculated totals, promotions, taxes etc.), but I would leave this improvement for the future.

pjedrzejewski:
Thanks Ben! 👍

bendavies:
yw!

PR #5167, opened 2016-06-02
[User] Customer validation fix

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets |
| License         | MIT

**The issue:**
The `CustomerType` should allow for empty first and last name since we can have customers without first or last name e.g. guest customers. When enforcing first and last name to be always filled we can't edit customers that didn't specify these values.

**What has been done:**
- [x] Removed default first and last name from Behat contexts (this releaved the bug)
- [x] Corrected the scenarios to fit current behavior
- [x] Added scenarios for validating first and last name minimum length
- [x] Fixed bug in validation checks in Behat's create and update pages
- [x] Corrected validation groups for `CustomerType` and for constraints on `Customer`

**Notice**
`CurrentPageResolver` should be used in `ManagingCustomersContext` to ensure we always operate on a correct page.

### DIFF
--- a/features/legacy/user/users.feature
+++ b/features/legacy/user/users.feature
@@ -75,12 +75,6 @@ Feature: Customers management
         When I follow "Create customer"
         Then I should be on the customer creation page
 
-    Scenario: Submitting empty form
-        Given I am on the customer creation page
-        When I press "Create"
-        Then I should still be on the customer creation page
-        And I should see "Please enter your first name"
-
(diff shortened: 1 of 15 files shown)


pjedrzejewski:
Nice work Michał! Thanks! 👍

PR #5172, opened 2016-06-03
[Backend] Fix displaying orders on dashboard

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
@@ -385,6 +385,23 @@ class OrderRepository extends CartRepository implements OrderRepositoryInterface
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function findCompleted(array $sorting = [], $limit = 5)
+    {
+        $queryBuilder = $this->createQueryBuilder('o');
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/WebBundle/Resources/views/Backend/Order/macros.html.twig
```diff
@@ -97,10 +97,14 @@
                 {% include 'SyliusWebBundle:Backend/Order:_paymentState.html.twig' %}
             </td>
             <td style="width: 20%">
-                {{ order.customer.fullname }}<br>
-                <a href="{{ path('sylius_backend_customer_show', {'id': order.customer.id}) }}">
-                    {{ order.customer.email }}
-                </a>
+                {% if order.customer is not null %}
```
michalmarcinkowski: The order should always have customer, why we need this if?

I recall some dashboard issue, but that's probably query problem. We should always query for placed orders, not carts (which it the only possible order representation that could not have customer assigned yet).

GSadee: Yes, this issue is caused by carts, which don't have assigned customer. So I fixed that problem in different way 👍

michalmarcinkowski:
Thank you Grzesiu! :+1:

PR #5174, opened 2016-06-03
[Documentation] The Book - Installation

TheMadeleine:
| Q               | A
| --------------- | ---
| Doc fix?        | yes
| New docs?    | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/docs/book/installation.rst
+++ b/docs/book/installation.rst
@@ -4,11 +4,13 @@
 Installation
 ============
 
-The Sylius main application can serve as end-user app, as well as a foundation
+The Sylius main application can serve as an end-user app, as well as a foundation
 for your custom e-commerce application.
 
-This article assumes you're familiar with `Composer`_, a dependency manager
-for PHP. It also assumes you have `Composer installed globally`_.
+.. warning::
+
+    This article assumes you're familiar with `Composer`_, a dependency manager
+    for PHP. It also assumes you have `Composer installed globally`_.
 
 .. note::
 
@@ -16,7 +18,7 @@ for PHP. It also assumes you have `Composer installed globally`_.
     ``php composer.phar`` where this guide uses ``composer``.
 
 
-It can be installed using two different approaches, depending on your use case.
+**Sylius** can be installed using two different approaches, depending on your use case.
 
 Install to Contribute
 ---------------------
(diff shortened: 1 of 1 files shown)


docs/book/installation.rst
```diff
@@ -74,8 +95,41 @@ browser.
 
 .. tip::
 
-    You can also use the Symfony built-in webserver by running the ``php app/console server:run --env dev``
-    command and then accessing ``http://localhost:8000``.
+    We strongly recommend using the Symfony abuilt-in webserver by running the
+    ``php app/console server:run --env dev``
```
michalmarcinkowski: `server:start` is more handy (runs process in the background). Also we don't need `--env dev` since it's default value.

docs/book/installation.rst
```diff
@@ -74,8 +95,41 @@ browser.
...
+    command and then accessing ``http://localhost:8000``
```
michalmarcinkowski: IMO `http://127.0.0.1:8000` will be more appropriate.

docs/book/installation.rst
```diff
@@ -74,8 +95,41 @@ browser.
...
+
+.. note::
+
+    The localhost's 8000 port may be already occupied by some other process.
+    If so you should try other ports, like for instance:
+    ``php app/console server:run localhost:8081 --env dev``
```
michalmarcinkowski: I would remove this sentence and write something like "Want to know more about using build in server, go to http://symfony.com/doc/current/cookbook/web_server/built_in.html"

docs/book/installation.rst
```diff
@@ -74,8 +95,41 @@ browser.
...
+
+You should be already logged in as an administrator, with the credentials you have provided during the installation process.
```
michalmarcinkowski: "You should be already logged" - there is no autologin during installation :)

docs/book/installation.rst
```diff
@@ -74,8 +95,41 @@ browser.
...
+Since now you can test your clean Sylius installation.
```
michalmarcinkowski: `test` -> `play with`?

docs/book/installation.rst
```diff
@@ -74,8 +95,41 @@ browser.
...
+
+In order to see a fully functional Administration you will need to install `Gulp`_.
+
+**Sylius** already has a ``gulpfile.js``, therefore you just need to get Gulp using `Node.js`_.
+
+Having Node.js installed in your project directory run:
+
+.. code-block:: bash
+
+    $ npm install
```
michalmarcinkowski: We should also add a link, how to install `npm`.

docs/book/installation.rst
Arminek: IMO ``127.0.0.1:8000``.

docs/book/installation.rst
```diff
@@ -74,8 +95,41 @@ browser.
...
-    You can also use the Symfony built-in webserver by running the ``php app/console server:run --env dev``
-    command and then accessing ``http://localhost:8000``.
+    We strongly recommend using the Symfony abuilt-in webserver by running the
+    ``php app/console server:run --env dev``
+    command and then accessing ``http://localhost:8000``
+
+.. note::
+
+    The localhost's 8000 port may be already occupied by some other process.
+    If so you should try other ports, like for instance:
```
Arminek: Or kill process.

Zales0123: 💀

docs/book/installation.rst
TheMadeleine: What if we leave this one and add the one you've suggested? I think it is a common mistake while using built in server for devs that are newbies :)

docs/book/installation.rst
TheMadeleine: There is a link -> https://github.com/Sylius/Sylius/pull/5174#discussion-diff-65704525R116 here.

michalmarcinkowski: 👍  Sorry I missed that 😖

docs/book/installation.rst
Arminek: to find occupied process use for example ``lsof -n -i:8000 | grep LISTEN``.

pjedrzejewski:
Thank you Magda! 👍

PR #5178, opened 2016-06-04
[ResourceBundle]Add the ability to register a custom AuthorizationChecker for a resource

BboyKeen:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Related tickets | None |
| License | MIT |

Heys guys,

With this pull request, we give the ability to register a custom authorization checker on a case by case basis following the same way as other services :

``` yml
sylius_resource:
    resources:
        app.book:
            classes:
                model: AppBundle\Entity\Book
                authorization_checker: AppBundle\Authorization\BookAuthorizationChecker
```

For example, it could be useful to implement an AuthorizationChecker which checks if the logged user has the ownership of a resource.
Could you provide me some hints if I need to implement tests please ?


Grome13:
I found this feature interesting. But in the ResourceController the method isGrantedOr403 is always  called before the findOr404 one. 
How do you envisage to check if a logged user has the ownership on a resource ?

i just begin with Sylius and i don't know all subtleties of it yet.

Jerome

PR #5182, opened 2016-06-05
[RFC] [Settings] Change Settings parameters type to json_array

bendavies:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      |yes
| Related tickets | 
| License         | MIT

I'm overriding this in my own app, because:

1. json is easily manually manipulated in the db directly. serialized data is not. (not recommended i know, but a use case nonetheless) 
2. json is easily query-able by RDBMS engines that support json natively, i.e postgres and mysql 5.7.

Opening it up here for RFC to see if it might be thought useful.

### DIFF
--- a/src/Sylius/Bundle/SettingsBundle/Resources/config/doctrine/model/Settings.orm.xml
+++ b/src/Sylius/Bundle/SettingsBundle/Resources/config/doctrine/model/Settings.orm.xml
@@ -24,7 +24,7 @@
 
         <field name="schemaAlias" column="schema_alias" type="string" nullable="false"/>
         <field name="namespace" type="string" nullable="true"/>
-        <field name="parameters" type="array"/>
+        <field name="parameters" type="json_array"/>
     </mapped-superclass>
 
 </doctrine-mapping>


pjedrzejewski:
Sounds great to me, wdyt @steffenbrem?

steffenbrem:
I think this is indeed a good improvement, you got my 👍

bendavies:
@pjedrzejewski just note the large BC break for existing installations.

bendavies:
actually, the data could probably be converted in a migration

bendavies:
@pjedrzejewski what do you need from me here? do you want the data migrated?

pjedrzejewski:
Thank you Ben! I will bring back migrations for this release, but after that they will be gone.

bendavies:
Did you not want a migration for the column change @pjedrzejewski?

michalmarcinkowski:
The build was broken after this merge, looks like the storage type influenced the logic in `ParameterTransformerListener`. @bendavies will you be able to have a look at the issue?

pamil:
The build should be fixed by #5271 - nevertheless settings transformation logic is way too complicated and probably quite cost expensive.

bendavies:
:+1:

PR #5185, opened 2016-06-06
[Review] Remove creating reviews in backend and product review form types

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | 
| License         | MIT


src/Sylius/Bundle/CoreBundle/Form/Type/ProductReviewAdminType.php
```diff
@@ -21,27 +20,14 @@
...
-                'class' => Product::class,
-                'label' => 'sylius.form.review.product',
-                'property' => 'name',
-            ])
-            ->add('author', 'entity', [
-                'class' => Customer::class,
-                'label' => 'sylius.form.review.author',
-                'property' => 'email',
-            ])
-        ;
```
Zales0123: If we remove all this, we don't really need to override ``buildForm`` method.

pjedrzejewski: Do we need this form at all?

GSadee: Generally, we don't need this, but I left that because we also have empty `ProductReviewType`. What do you think? Should I remove them and use basic `ReviewType`?

pjedrzejewski: Yes, let's do this.

GSadee: Done 👍

pjedrzejewski:
Thank you Grzesiu!

PR #5192, opened 2016-06-07
[Resource] Repository interface fix

TheMadeleine:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets |
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Resource/Repository/InMemoryRepository.php
+++ b/src/Sylius/Component/Resource/Repository/InMemoryRepository.php
@@ -167,7 +167,7 @@ class InMemoryRepository implements RepositoryInterface
         $resources = $this->findAll();
 
         if (!empty($sorting)) {
-            $resources = $this->applyOrder($resources, $orderBy);
+            $resources = $this->applyOrder($resources, $sorting);
         }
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
@TheMadeleine Why removing the constants?

TheMadeleine:
@pjedrzejewski - Since they are used just in one place and that in my opinion makes them a bit redundant.

michalmarcinkowski:
We should use them in all new development and when refactoring the existing repository methods. I would leave them :)

TheMadeleine:
Corrected :)

michalmarcinkowski:
This PR should fix [this bug](https://github.com/michalmarcinkowski/Sylius/blob/90f267c8edfc0d6ce48813b2adc619017ce50682/src/Sylius/Component/Resource/Repository/InMemoryRepository.php#L170-L170), but know it doesn't.

TheMadeleine:
now, it does @michalmarcinkowski

pjedrzejewski:
Thanks Magda!

PR #5196, opened 2016-06-07
[Core] Implement simple ShopperContext

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | -
| License         | MIT

All services, form types, etc. should be stateless, but sometimes we need to have access (in routing, or other contexts) to currently active channel, currency, locale and customer. This service simplifies this and I have also added this info to the web debug toolbar. So as a developer, you can always check the currently selected context.

This is the first step to clean up all the contexts together and unify their behavior and usage.

zrzut ekranu 2016-06-07 o 23 20 22

### DIFF
--- a/src/Sylius/Component/Core/Context/ShopperContext.php
+++ b/src/Sylius/Component/Core/Context/ShopperContext.php
@@ -1,0 +1,93 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Core\Context;
+
+use Sylius\Component\Channel\Context\ChannelContextInterface;
+use Sylius\Component\Currency\Context\CurrencyContextInterface;
+use Sylius\Component\Locale\Context\LocaleContextInterface;
+use Sylius\Component\User\Context\CustomerContextInterface;
+
+/**
+ * @author Paweł Jędrzejewski <pawel@sylius.org>
+ */
+final class ShopperContext implements ShopperContextInterface
+{
+    /**
+     * @var ChannelContextInterface
+     */
+    private $channelContext;
+
+    /**
+     * @var CurrencyContextInterface
+     */
+    private $currencyContext;
+
+    /**
+     * @var LocaleContextInterface
+     */
+    private $localeContext;
+
+    /**
+     * @var CustomerContextInterface
+     */
+    private $customerContext;
+
+    /**
+     * @param ChannelContextInterface $channelContext
+     * @param CurrencyContextInterface $currencyContext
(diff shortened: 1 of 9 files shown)


src/Sylius/Component/Core/Context/ShopperContextInterface.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public function getCurrencyCode();
+
+    /**
+     * @return string
+     */
+    public function getLocaleCode();
+
+    /**
+     * @return CustomerInterface
```
michalmarcinkowski: `CustomerInterface|null`

src/Sylius/Component/Core/Context/ShopperContextInterface.php
```diff
@@ -0,0 +1,41 @@
...
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Core\Model\CustomerInterface;
+
+/**
+ * @author Paweł Jędrzejewski <pawel@sylius.org>
+ */
+interface ShopperContextInterface
+{
+    /**
+     * @return ChannelInterface
```
michalmarcinkowski: Missing `@throws ...`

src/Sylius/Component/Core/Context/ShopperContext.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    private $customerContext;
+
+    /**
+     * @param ChannelContextInterface $channelContext
+     * @param CurrencyContextInterface $currencyContext
+     * @param LocaleContextInterface $localeContext
+     * @param CustomerContextInterface $customerContext
+     */
+    public function __construct(ChannelContextInterface $channelContext, CurrencyContextInterface $currencyContext, LocaleContextInterface $localeContext, CustomerContextInterface $customerContext)
```
michalmarcinkowski: Too long line.

src/Sylius/Component/Core/Context/ShopperContextInterface.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace Sylius\Component\Core\Context;
+
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Core\Model\CustomerInterface;
+
+/**
+ * @author Paweł Jędrzejewski <pawel@sylius.org>
+ */
+interface ShopperContextInterface
```
pamil: What do you think about normalizing methods in current interfaces (one has `getCurrentX()`, the other one - `getX()`) and then just extending them here?

pjedrzejewski: @pamil I think it could be good, but let's leave that for a separate PR, we definitely need to normalize methods on all these contexts.

michalmarcinkowski:
Nice! 👍

PR #5197, opened 2016-06-07
[Cart][Core] Rework the logic behind the cart

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | -
| License         | MIT

Based on #5196.

Summary of changes:

* Simplified getting the cart to ``sylius.context.cart`` service and ``getCart`` method;
* Context uses a prioritized list of services tagged as ``sylius.cart_provider``;
* It returns the cart from the first provider that actually figures out how to get the cart

Currently there is:

* NewCartProvider - it always returns an empty cart (Cart component)
* NewCartProvider (decorates the 1st one) - it always returns an empty cart, but also assigns the channel, currency, customer (if defined) and in the near future also the locale (Core component)
* SessionCartProvider - it tries to find the cart id in the session (CartBundle), shall be used for not logged in customers;
* ExistingCartProvider - Tries to find the cart of currently logged customer (this will give us persistent carts quite easily)

Previous logic was a mess, so I am trying to clean up these things a bit and make it more consistent with the rest of the platform.

Todo:

- [x] Update all other occurrences of CartProvider in both specs and code
- [x] Finish implementation of existing cart provider
- [x] Figure out a nice way to save the id of the cart in the session (for the sake of SessionCartProvider)

Example usage:

``` yaml
services:
    app.cart_provider.custom:
        class: App\Cart\Provider\CustomCartProvider
        tags:
            - { name: sylius.cart_provider, priority: 1000 }
```

And next ``sylius.context.cart::getCart()`` call uses your logic to get the cart, if you do not return it - it fallbacks to Sylius logic (further providers try to return the cart)... Simple as that. :)


peteward:
How does the ExistingCartProvider work in practice? I see the logic for it, but at what point what you use it, or when does a session-based cart move to an existing one?

Whenever a customer first lands on the site, the cart provider logic will kick in and create them a cart (presumably a new one which then gets put into session) before they login. So in theory someone from a web context will always have a session-based cart.  They could add items to this cart and then login later on. At which point, what happens to their session cart if they find an existing one from a previous login? 

Our solution for this was a Security post-login listener which merged items from any previous abandoned login carts into your session one, that way you have consistency in always ending up with a session-based cart.

On the positive side I see that having alternative providers means you can detach the association with session when working from different context such as CLI or behat.

src/Sylius/Bundle/CartBundle/DependencyInjection/Compiler/RegisterCartProvidersPass.php
```diff
@@ -0,0 +1,41 @@
...
+ * @author Paweł Jędrzejewski <pawel@sylius.org>
+ */
+class RegisterCartProvidersPass implements CompilerPassInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('sylius.registry.cart_provider')) {
```
pamil: `hasDefinition` -> `has` (also checks for existence of aliases)

src/Sylius/Bundle/CartBundle/DependencyInjection/Compiler/RegisterCartProvidersPass.php
```diff
@@ -0,0 +1,41 @@
...
+            return;
+        }
+
+        $registry = $container->findDefinition('sylius.registry.cart_provider');
+        $providers = [];
```
pamil: It's not used

src/Sylius/Bundle/CoreBundle/Resources/config/cart_providers.xml
```diff
@@ -0,0 +1,26 @@
...
+
+-->
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services
+                               http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="sylius.cart_provider.new.decorated" class="Sylius\Component\Core\Cart\Provider\NewCartProvider" decorates="sylius.cart_provider.new">
```
Arminek: Is there any reason why you keep two definition of ``cart_provider.new``? IMO it should replace the ``sylius.cart_provider.new``.

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
```diff
@@ -471,4 +472,56 @@ private function createQueryBuilderBetweenDates(\DateTime $from, \DateTime $to,
 
         return $queryBuilder;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function findOneByChannelAndCustomer(ChannelInterface $channel, CustomerInterface $customer)
```
michalmarcinkowski: findOne**Cart**ByChannelAnd... or find~~One~~**Cart**ByChannelAnd...

src/Sylius/Component/Core/Cart/Provider/NewCartProvider.php
```diff
@@ -0,0 +1,78 @@
...
+        $cart = $this->decoratedCartProvider->getCart();
+
+        if (null === $cart) {
+            throw new \LogicException('Decorated cart provider must return a cart instance, null given.');
+        }
+
+        try {
+            $cart->setChannel($this->shopperContext->getChannel());
+        } catch (ChannelNotFoundException $exception) {
+            $cart->setChannel();
```
michalmarcinkowski: I would add comment with `@TODO` tag here, because we should never have cart not assigned to any channel.

src/Sylius/Bundle/CoreBundle/EventListener/CartBlamerListener.php
```diff
@@ -1,105 +0,0 @@
...
-     */
-    public function onInteractiveLogin(InteractiveLoginEvent $interactiveLoginEvent)
-    {
-        $user = $interactiveLoginEvent->getAuthenticationToken()->getUser();
-
-        if (!$user instanceof UserInterface) {
-            return;
-        }
-
-        $this->blame($user);
```
michalmarcinkowski: After this change, how do we assign cart to customer after login?

src/Sylius/Component/Core/Cart/Provider/NewCartProvider.php
pamil: Anyway, `$cart->setChannel(null);`

pamil: IMHO it should fail if there's no channel.

src/Sylius/Bundle/CartBundle/Provider/SessionCartProvider.php
```diff
@@ -0,0 +1,70 @@
...
+    private $cartRepository;
+
+    /**
+     * @param SessionInterface $session
+     * @param string $sessionKeyName
+     * @param CartRepositoryInterface $cartRepository
+     */
+    public function __construct(
+        SessionInterface $session,
+        $sessionKeyName,
```
pamil: We can extract `CartStorageInterface` with methods `set(CartInterface $cart): void` and `get(): CartInterface` so we won't have to pass these two separately and this provider can become storage type independent.

src/Sylius/Bundle/CartBundle/EventListener/SessionCartSubscriber.php
```diff
@@ -0,0 +1,87 @@
...
+    private $sessionKeyName;
+
+    /**
+     * @param CartContextInterface $cartContext
+     * @param string $sessionKeyName
+     */
+    public function __construct(CartContextInterface $cartContext, $sessionKeyName)
+    {
+        $this->cartContext = $cartContext;
+        $this->sessionKeyName = $sessionKeyName;
```
pamil: Same as for `SessionCartProvider`, extracting the storage layer interface may make this subscriber storage independent.

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
pamil: It's both `Order` and `Cart` so the current name is the best :)

src/Sylius/Component/Cart/Provider/CartProviderInterface.php
```diff
@@ -14,39 +14,14 @@
...
-     * @return bool
-     */
-    public function hasCart();
-
-    /**
-     * Returns current cart.
-     * If none found is by storage, it should create new one and save it.
-     *
-     * @return CartInterface
+     * @return null|CartInterface
```
pamil:
```
@return CartInterface

@throws CartNotFoundException
```

src/Sylius/Component/Core/Cart/Provider/NewCartProvider.php
```diff
@@ -0,0 +1,78 @@
...
+        $this->shopperContext = $shopperContext;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCart()
+    {
+        if (null !== $this->cart) {
+            return $this->cart;
```
pamil: It looks fragile.

tests/Responses/Expected/sitemap/show_sitemap.xml
```diff
@@ -2,19 +2,19 @@
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
   <url>
     <loc>http://localhost/p/mug-star-wars</loc>
-    <lastmod>2015-10-10T00:00:00+00:00</lastmod>
+    <lastmod>@string@.isDateTime()</lastmod>
```
pamil: :+1:, but may be contributed standalone :)

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
michalmarcinkowski: No it's not https://github.com/Sylius/Sylius/pull/5197/files#diff-b3c4fa070df7e585e4d320386f45b379R517, It finds only carts 😝

pamil: Hate that order-cart mess :/

src/Sylius/Bundle/CartBundle/EventListener/SessionCartSubscriber.php
```diff
@@ -0,0 +1,87 @@
...
+    public function onKernelResponse(FilterResponseEvent $event)
+    {
+        if (!$event->isMasterRequest()) {
+            return;
+        }
+
+        /** @var Request $request */
+        $request = $event->getRequest();
+        // Hacky hack. Until there is a better solution.
+        if (!$this->isHtmlRequest($request)) {
```
michalmarcinkowski: How about excluding admin section for this subscriber?

`if (!$this->isHtmlRequest($request) || $this->isAdminRoute($request) {`

and inside method:
```
function isAdminRoute(Request $request)
{
    $route = $request->attributes->get('_route');

    return false !== strpos($route, 'admin');
}
```

This should remove `ChannelNotFound` try-catch from here https://github.com/Sylius/Sylius/pull/5197/files#diff-574f7dbf46d15fff5561f80bb9121870R68

src/Sylius/Component/Core/Cart/Provider/NewCartProvider.php
```diff
@@ -0,0 +1,78 @@
...
+use Sylius\Component\Cart\Model\CartInterface;
+use Sylius\Component\Cart\Provider\CartProviderInterface;
+use Sylius\Component\Channel\Context\ChannelNotFoundException;
+use Sylius\Component\Core\Context\ShopperContextInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+
+/**
+ * @author Paweł Jędrzejewski <pawel@sylius.org>
+ */
+class NewCartProvider implements CartProviderInterface
```
pamil: Class name does not say exactly what it really does :)

src/Sylius/Bundle/CartBundle/EventListener/SessionCartSubscriber.php
pamil: Nope nope nope nope nope...

pamil: That `ChannelNotFoundException` should rethrow it as `CartNotFoundException` and here, if we receive that `CartNotFoundException` we should simply catch it and do not try to persist it in session.

michalmarcinkowski: Agree with the first part, but still this subscriber shouldn't be enabled for admin section, so we can exclude this routes.

I already see an issue if we don't do that. After split to Shop and Admin users we will receive cart saved in session for ShopUser while browsing admin panel.

pamil: > Nope nope nope nope nope...

Do you mean that part? 🎉 

Excluding / including anything by route sounds really, really bad and should be avoided at all costs (just ask @lchrusciel about his romance with SiteAccess :)). 

We already have channel contexts reorganised and they're working like a charm - the channel context always throws an exception when in AdminBundle (you can see that by looking at debug toolbar while browsing the admin panel).

We can use that to our advantage and make cart providers return a new cart (`NewCartProvider` in this PR) only if there is a channel found. So `CartProvider` / `CartContext` will always throw an exception if trying to get a current cart.

If there's no cart passed here, we cannot persist it in session :)

pamil: I'm talking about this cart provider: https://github.com/Sylius/Sylius/pull/5197/files#r71517527

pamil: Anyway, if cart was assigned on shop and then we are having the same session for in admin panel, the change you propose wouldn't change anything, as we will already have that cart persisted in session.

michalmarcinkowski: After giving it more thoughts I withdraw my idea with route checking 😄  I just wanted to reduce overhead on admin section, since this subscriber is triggered on every request, but let's find a better way during optimization phase :)

Right now we should at least skip writing to session on requests in admin section.

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
pjedrzejewski: Actually, this method should be removed, as I dropped the existing cart provider for now. The Cart / Order mess will be fixed in another PR.

src/Sylius/Component/Core/Cart/Provider/NewCartProvider.php
pjedrzejewski: Suggestions?

pamil: Got another idea, if we'll inject a cart factory here instead of decorating another provider, it can still be named `NewCartProvider`. If not, we should emphasize, that it is a decorator and does all the blaming on passed cart.

src/Sylius/Component/Cart/Provider/CartProviderInterface.php
```diff
@@ -14,39 +14,14 @@
 use Sylius\Component\Cart\Model\CartInterface;
 
 /**
- * Interface for object that is accessor for cart.
- * It should retrieve existing cart or create new one based on storage.
+ * Used by context to obtain a cart.
  *
  * @author Paweł Jędrzejewski <pawel@sylius.org>
  */
 interface CartProviderInterface
```
pamil: Btw. it looks like it has the same methods as `CartContextInterface`, can we drop the concepts of providers and just use composition in order to create contexts that relies on another contexts? (Just like CompositeChannelContext, CompositeCurrencyContext in #5497, it will make all our contexts using the same approach).

src/Sylius/Bundle/CartBundle/Controller/Controller.php
```diff
@@ -58,19 +58,19 @@ protected function getCartSummaryRoute()
...
         return $this
-            ->getProvider()
+            ->getContext()
             ->getCart()
         ;
     }
 
     /**
-     * Get cart provider.
+     * Get cart context.
```
lchrusciel: I guess, it is redundant.

src/Sylius/Bundle/CartBundle/Provider/SessionCartProvider.php
```diff
@@ -0,0 +1,70 @@
...
+namespace Sylius\Bundle\CartBundle\Provider;
+
+use Sylius\Component\Cart\Provider\CartProviderInterface;
+use Sylius\Component\Cart\Repository\CartRepositoryInterface;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+
+/**
+ * @author Paweł Jędrzejewski <pawel@sylius.org>
+ */
+class SessionCartProvider implements CartProviderInterface
```
lchrusciel: final?

src/Sylius/Bundle/CoreBundle/Cart/ItemResolver.php
```diff
@@ -60,22 +60,22 @@ class ItemResolver implements ItemResolverInterface
     protected $channelContext;
 
     /**
-     * @param CartProviderInterface          $cartProvider
+     * @param CartContextInterface          $cartContext
```
lchrusciel: One of the last places, where we still have parameters aligned vertically... Good, old times. Can we remove this alignment?

src/Sylius/Component/Cart/spec/Context/CartContextSpec.php
```diff
@@ -0,0 +1,72 @@
...
+use Sylius\Component\Cart\Model\CartInterface;
+use Sylius\Component\Cart\Provider\CartProviderInterface;
+use Sylius\Component\Registry\PrioritizedServiceRegistryInterface;
+
+/**
+ * @mixin CartContext
+ *
+ * @author Paweł Jędrzejewski <pawel@sylius.org>
+ */
+class CartContextSpec extends ObjectBehavior
```
lchrusciel: With #5541 we will introduce a final key word to every spec. So, missing final

src/Sylius/Component/Core/Cart/Provider/NewCartProvider.php
lchrusciel: ImmutableCartProvider? Considering [this line](https://github.com/Sylius/Sylius/pull/5197/files#diff-574f7dbf46d15fff5561f80bb9121870R55) we should not name it NewCartProvider. Or CurrentCartProvider? At the end, this class return a current cart provider, if it instance exists, or create a new one.

src/Sylius/Bundle/CartBundle/EventListener/SessionCartSubscriber.php
lchrusciel: @pamil @michalmarcinkowski I have a trauma after that relationship. Don't wanna to talk about it 😭 
We should avoid depending on a route name in all the cases.

pjedrzejewski:
😢 🌵

PR #5198, opened 2016-06-07
[Core] Remove restricted zone feature

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | -
| License         | MIT

Unfortunately, this is not implemented in a clean way and should not go to 1.0. If you have been using this feature - sorry, but should be quite simple to reimplement it. (can even be this way if works for you, it is just not generic enough and not tested properly)


michalmarcinkowski:
Agree with that.

harikt:
Hi @pjedrzejewski ,

not using it, but out of curiosity how do you think it could be implemented.

Thank you

PR #5202, opened 2016-06-08
[Core] Implement object usefull for string formatting

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets |
| License         | MIT

There are many places where we need to format string to some *code-like* form and it is nice to have an object with such a responsibility. I encourage you all to talk about this class name. I've had many ideas like:
- ``SpacesToUnderscoreFormatter``
- ``SnakeCaseFormatter``
- ``Snaker`` (:smile:)

but finally I decided to something more neutral. I would appreciate brainstorm about that ;)

### DIFF
--- a/src/Sylius/Behat/Context/Setup/ProductContext.php
+++ b/src/Sylius/Behat/Context/Setup/ProductContext.php
@@ -15,6 +15,7 @@ use Behat\Behat\Context\Context;
 use Behat\Gherkin\Node\TableNode;
 use Doctrine\Common\Persistence\ObjectManager;
 use Sylius\Component\Attribute\Factory\AttributeFactoryInterface;
+use Sylius\Component\Core\Formatter\StringInflector;
 use Sylius\Component\Core\Model\ChannelInterface;
 use Sylius\Component\Core\Model\ProductInterface;
 use Sylius\Component\Core\Model\ProductVariantInterface;
@@ -428,6 +429,6 @@ final class ProductContext implements Context
      */
     private function convertToCode($productName)
     {
-        return strtoupper(str_replace(' ', '_', $productName));
+        return StringInflector::nameToUpercaseCode($productName);
     }
 }
--- a/src/Sylius/Component/Core/Formatter/StringInflector.php
+++ b/src/Sylius/Component/Core/Formatter/StringInflector.php
@@ -1,0 +1,48 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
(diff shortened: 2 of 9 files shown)


pjedrzejewski:
```StringInflector::nameTo(Lowercase|Uppercase)Code```

https://www.google.pl/#q=inflect

WDYT?

**Comment by Zales0123** [comment]
Looks nice 👍  I'm only not convinced with ``Code`` at the end of function, what if we want to use this ``Inflector`` not for code but sth else? It could lead to confusion and a huge disaster :(

**Comment by pjedrzejewski** [comment]
@Zales0123 I believe separation of concerns is good, so if we define behavior of this object as transforming resource names to codes, if anything about that behavior changes, we change it here. Converting spaces to underscores is generic string manipulation functionality and could go to other object if we need it, but let's think about it when we actually need it. In Behat's case, we always (I think) need to transform name to code.

**Comment by Zales0123** [comment]
@pjedrzejewski convinced and agreed 👍

**Comment by michalmarcinkowski** [comment]
Since the object is in `Core` component we should not use it in bundles that are not dependent on `Core`.  Please remove the usages from `ProductContext`, `PromotionContext` etc. they will be removed soon anyway.

File: src/Sylius/Component/Core/Formatter/StringInflector.php [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+    {
+        return str_replace(' ', '_', $value);
+    }
+
+    /**
+     * @param string $value
+     *
+     * @return string
+     */
+    public static function nameToLowercaseCode($value)
```
lchrusciel: nameToLowercase**d**Code? WDYT?

TheMadeleine: Nope :D lowercase is enough

pjedrzejewski:
Thank you Mateusz!

src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -428,6 +429,6 @@ private function saveProduct(ProductInterface $product)
      */
     private function convertToCode($productName)
     {
-        return strtoupper(str_replace(' ', '_', $productName));
+        return StringInflector::nameToUpercaseCode($productName);
```
michalmarcinkowski: Why we use uppercase for products code?

PR #5205, opened 2016-06-08
[Documentation] README update - Installation

TheMadeleine:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | mentioned in #5174 
| License         | MIT

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -18,14 +18,21 @@ The documentation uses [GitHub issues](https://github.com/Sylius/Sylius/issues).
 Build
 -----
 
-If you contribute to documentation you will want to check how it looks after your changes.  
-To be able to build the documentation, install [Sphinx](http://sphinx-doc.org/).
+To test the documentation before a commit:
 
-1. [Install `pip`, Python package manager](https://pip.pypa.io/en/stable/installing/)
-2. Download documentation requirements: `$ pip install -r requirements.txt`
-3. Build the documentation:`$ sphinx-build -b html . build`
+* [Install `pip`, Python package manager](https://pip.pypa.io/en/stable/installing/)
 
-Documentation index is `build/index.html`.
+* Download the documentation requirements: 
+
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -18,14 +18,19 @@ The documentation uses [GitHub issues](https://github.com/Sylius/Sylius/issues).
...
-To be able to build the documentation, install [Sphinx](http://sphinx-doc.org/).
+To test the documentation before a commit:
 
-1. [Install `pip`, Python package manager](https://pip.pypa.io/en/stable/installing/)
-2. Download documentation requirements: `$ pip install -r requirements.txt`
-3. Build the documentation:`$ sphinx-build -b html . build`
+* Install [Python](https://www.python.org/)
 
-Documentation index is `build/index.html`.
+    `$ brew install python`
```
pamil: It implies that you are using OS X

docs/README.md
```diff
@@ -18,14 +18,19 @@ The documentation uses [GitHub issues](https://github.com/Sylius/Sylius/issues).
...
+
+* Install [Sphinx](http://www.sphinx-doc.org/en/stable/)
+
+    `$ sudo pip install Sphinx`
```
pamil: `pip install -r requirements.txt` ensures that Sphinx >=1.4.2 is installed

pamil: Anyway, we should not use `sudo` with `pip`: http://stackoverflow.com/questions/21055859/what-are-the-risks-of-running-sudo-pip

pamil: Last but not least, some Python packages may not have `pip` bundled so we should put another step between installing Python and installing Sphinx.

TheMadeleine: Precious remarks @pamil ! Thanks! :)

pjedrzejewski:
Thank you Magda!

PR #5210, opened 2016-06-09
[Resource] Allow explicit enabling of drivers

dantleech:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| License         | MIT

This PR allows the explicit enabling of drivers for the ResourceBundle, in the same way that has been done for the GridBundle (https://github.com/Sylius/Sylius/pull/4972).

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/DependencyInjection/Configuration.php
+++ b/src/Sylius/Bundle/ResourceBundle/DependencyInjection/Configuration.php
@@ -39,6 +39,7 @@ class Configuration implements ConfigurationInterface
         $this->addResourcesSection($rootNode);
         $this->addSettingsSection($rootNode);
         $this->addTranslationsSection($rootNode);
+        $this->addDriversSection($rootNode);
 
         $rootNode
             ->children()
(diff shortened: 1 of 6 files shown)


src/Sylius/Bundle/ResourceBundle/DependencyInjection/Configuration.php
```diff
@@ -174,4 +175,35 @@ private function addTranslationsSection(ArrayNodeDefinition $node)
             ->end()
         ;
     }
+
+    private function addDriversSection(ArrayNodeDefinition $node)
```
dantleech: This method is copied verbatim from the GridBundle. Not really sure how to avoid that without adding a new shared dependency just for this section of code..

src/Sylius/Bundle/ResourceBundle/DependencyInjection/SyliusResourceExtension.php
```diff
@@ -60,7 +62,32 @@ public function load(array $config, ContainerBuilder $container)
...
+        $this->loadPersistence($config['drivers'], $config['resources'], $loader);
+        $this->loadResources($config['resources'], $container);
+    }
+
+    private function loadPersistence(array $enabledDrivers, array $resources, LoaderInterface $loader)
+    {
+        foreach ($resources as $alias => $resource) {
+            if (!in_array($resource['driver'], $enabledDrivers)) {
+                throw new InvalidArgumentException(sprintf(
+                    'Resource "%s" uses driver "%s", but this driver has not been enabled.',
```
dantleech: This is a sanity check. If resources have been configured and their drivers have not been enabeld, throw an exception. We magically add any missing drivers, but I prefer explicitness.

dantleech: This logic is missing from the GridBundle however.

dantleech:
I need to add tests for this. All of this logic is duplciated from the GridBundle, so I do wonder if there is a way we can share it (preferably without requiring `CoreBundle`) ?

dantleech:
@pjedrzejewski any ideas on the above comment? Shall we duplicate this logic for now?

dantleech:
Maybe we could put all the "driver" logic that is shared between bundles into an own component (or maybe even a bundle which can handle global settings, e.g. use `doctrine/phpcr-odm` by default).

pjedrzejewski:
Thanks Dan! Yes, I am fine with that duplication for now. We have other prios, but there is plenty of place for improvement in whole resource/driver layer. :)

PR #5211, opened 2016-06-09
Default to empty array on translator resource provider

dantleech:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

`$resources` is an array, but it defaults to NULL.

If no resources are available, an `array_merge` action is later performed and an error occurs.

This PR changes the default to an empty array.


michalmarcinkowski:
Thanks @dantleech!

PR #3773, opened 2015-12-23
[Taxonomy] Remove Taxonomy entity in favor of Taxon

aramalipoor:
Q  | A
------------- | -------------
Bug fix?  | no
New feature?  | refactor
BC breaks? | yes
Deprecations? | no
Fixed tickets | #3761 
License | MIT
Doc PR | -

Requires Atlantic18/DoctrineExtensions#1514

PR #5170, opened 2016-06-03
[Core][Product] Fix Inexistent method

h-sn:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Related tickets |  |
| License | MIT |

PR #5215, opened 2016-06-09
Use getRoot on taxons (replacing the remove getTaxonomy)

coudenysj:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | relates to #3773
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Core/Model/Product.php
+++ b/src/Sylius/Component/Core/Model/Product.php
@@ -130,11 +130,11 @@ class Product extends BaseProduct implements ProductInterface, ReviewableProduct
     /**
      * {@inheritdoc}
      */
-    public function getTaxons($rootTaxon = null)
+    public function getTaxons($rootTaxonCode = null)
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
Good catch, but I think we should use ``code`` here and not the name, wdyt?

coudenysj:
Makes sense. Updated the code.

pjedrzejewski:
Thanks Jachim! 👍

michalmarcinkowski:
There was already a fix for that in https://github.com/Sylius/Sylius/pull/5170. @coudenysj could you write spec for this method to prevent from regression in the future?

coudenysj:
Sorry @michalmarcinkowski & @h-sn, I wasn't aware of that. If @h-sn is willing to rebase then the test would still be his code. If not, I'm happy to supply the test for this.

PR #5216, opened 2016-06-09
[Core][Shipping] Fix shipping method choice type

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

There was a problem with ``ShippingMethodChoiceType``, as it didn't use channel to list shipping methods and also not available for channel methods have been... available.
Instead of quick fix, I've rebuild deeply shipping methods resolver logic:
- ``ShippingMethodChoiceType`` uses prioritized registry to get proper shipping methods resolver
- there are now two shipping methods resolver:
    - resolver from *Shipping* returns all enabled shipping methods that are eligible (eligibility checker used(
    - resolver from *Core* returns all enabled shipping methods for shipment order address (zone matcher) and channel

This changes allowed to remove ``ShippingMethodChoiceType`` instance from *Core*. Also it's now highly simple to use own shipping methods resolver (it has to be just implemented and tagged with proper priority).


src/Sylius/Bundle/ShippingBundle/Form/Type/ShippingMethodChoiceType.php
```diff
@@ -82,10 +81,16 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
-                $methods = $this->resolver->getSupportedMethods($options['subject'], $options['criteria']);
-            } else {
-                $methods = $this->repository->findBy($options['criteria']);
+                /** @var MethodsResolverInterface $resolver */
+                foreach ($this->resolversRegistry->all() as $resolver) {
+                    if ($resolver->supports($options['subject'])) {
+                        $methods = $resolver->getSupportedMethods($options['subject']);
+                        break;
+                    }
+                }
```
pjedrzejewski: Let's move this logic to the "main" methods resolver and inject it here. It will support every shipment and use exactly this logic to get the right methods. (CompositeMethodsResolver as name, wdyt?)

src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
```diff
@@ -622,6 +624,12 @@
             <argument type="service" id="sylius.sql_logger.purger" />
         </service>
 
+        <service id="sylius.shipping_methods_resolver.by_zones_and_channel" class="%sylius.shipping_methods_resolver.by_zones_and_channel.class%">
+            <argument type="service" id="sylius.repository.shipping_method" />
+            <argument type="service" id="sylius.zone_matcher" />
+            <tag name="sylius.shipping_method_resolver" type="by_zones_and_channel" label="By zones and channel" priority="1" />
```
pjedrzejewski: Label should be a translation key: ``sylius.shipping_methods_resolver.by_zones_and_channel``.

src/Sylius/Bundle/ShippingBundle/Form/Type/ShippingMethodChoiceType.php
Zales0123: Sure 👍  Less logic in form type is always a good choice 😄

src/Sylius/Bundle/ShippingBundle/SyliusShippingBundle.php
```diff
@@ -47,6 +48,7 @@ public function build(ContainerBuilder $container)
 
         $container->addCompilerPass(new RegisterCalculatorsPass());
         $container->addCompilerPass(new RegisterRuleCheckersPass());
+        $container->addCompilerPass(new RegisterShippingMethodsResolverPass());
```
pjedrzejewski: We usually use plural form here: Resolver**s**.

src/Sylius/Component/Core/Resolver/ShippingMethodsByZonesAndChannelResolver.php
```diff
@@ -0,0 +1,102 @@
...
+        $zones = $this->getZonesIdsForAddress($subject->getOrder());
+        if (empty($zones)) {
+            return [];
+        }
+
+        /** @var ChannelInterface $channel */
+        $channel = $subject->getOrder()->getChannel();
+
+        $methods = [];
+        foreach ($this->shippingMethodRepository->findBy(['enabled' => true, 'zone' => $zones]) as $method) {
```
pjedrzejewski: ``zone`` should work fine with a collection of zones, you should not need to convert them to ids... IIRC.

src/Sylius/Behat/Context/Setup/ShippingContext.php
```diff
@@ -183,6 +183,11 @@ private function createShippingMethod(
 
         $this->shippingMethodRepository->add($shippingMethod);
         $this->sharedStorage->set('shipping_method', $shippingMethod);
+
+        if ($this->sharedStorage->has('channel')) {
+            $channel = $this->sharedStorage->get('channel');
+            $channel->addShippingMethod($shippingMethod);
+        }
```
michalmarcinkowski: This `if` should be before `$this->shippingMethodRepository->add($shippingMethod);`. Right now the relation is not flushed in this step.

src/Sylius/Bundle/ShippingBundle/Resources/config/services.xml
```diff
@@ -67,12 +70,18 @@
             <argument>shipping rule checker</argument>
         </service>
 
+        <service id="sylius.registry.shipping_methods_resolver" class="%sylius.registry.shipping_methods_resolver.class%">
+            <argument>%sylius.shipping_methods_resolver.interface%</argument>
+            <argument>Shipping methods resolver</argument>
```
michalmarcinkowski: `shipping_methods_resolver`?

src/Sylius/Component/Core/Resolver/ShippingMethodsByZonesAndChannelResolver.php
michalmarcinkowski: We should have custom repository method `findEnabledByZones(array|ZoneInterface[] $zones)`.

michalmarcinkowski: Even better `findEnabledForChannelByZones(ChannelInterface $channel, array|ZoneInterface[] $zones)`

src/Sylius/Component/Core/Resolver/ShippingMethodsByZonesAndChannelResolver.php
```diff
@@ -0,0 +1,102 @@
...
+        ZoneMatcherInterface $zoneMatcher
+    ) {
+        $this->shippingMethodRepository = $shippingMethodRepository;
+        $this->zoneMatcher = $zoneMatcher;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSupportedMethods(ShippingSubjectInterface $subject)
```
michalmarcinkowski: Why not simply `getShippingMethods`?

src/Sylius/Component/Shipping/Resolver/MethodsResolver.php
```diff
@@ -44,11 +43,11 @@ public function __construct(ObjectRepository $repository, ShippingMethodEligibil
     /**
      * {@inheritdoc}
      */
-    public function getSupportedMethods(ShippingSubjectInterface $subject, array $criteria = [])
+    public function getSupportedMethods(ShippingSubjectInterface $subject)
     {
         $methods = [];
 
-        foreach ($this->getMethods($criteria) as $method) {
+        foreach ($this->repository->findBy(['enabled' => true]) as $method) {
```
michalmarcinkowski: `repository` -> `shippingMethodRepository`

src/Sylius/Bundle/ShippingBundle/Resources/config/services.xml
Zales0123: According to [service implementation](https://github.com/Sylius/Sylius/blob/3e0fd8522683bc6c893cc583fdb9d543b56f96c2/src/Sylius/Component/Registry/PrioritizedServiceRegistry.php#L35) ``$context``parameter should be "human readable".

michalmarcinkowski: Fine for me, but then let's lowercase the first letter 😄

pjedrzejewski: @michalmarcinkowski This is for exception message, so uppercase letter is ok.

src/Sylius/Component/Core/Resolver/ShippingMethodsByZonesAndChannelResolver.php
pjedrzejewski: That is original name of the method. Maybe we can keep it for now and also suggests what kind of methods we get from this resolver. ;)

src/Sylius/Bundle/ShippingBundle/Resources/config/services.xml
michalmarcinkowski: 👍  I recall that PR :)

src/Sylius/Component/Core/Resolver/ShippingMethodsByZonesAndChannelResolver.php
Zales0123: Aggreed, however it would require new custom repository on *Core* and, primarily, some query builder magic with subqueries... so maybe we could leave this as it is and improve in separate PR?

pjedrzejewski:
Nice work Mateusz, thanks!

Note: in separate PR, we could remove the CompositeMethodsResolverInterface - This could be an implementation of ShippingMethodsResolverInterface - no need for new contract. Anyway, this is a minor improvement. 👍

Issue #1214, opened 2014-03-17
Redirect to Index with no parameters


pdias:
Hello,

I'm using SyliusResourceBundle. I have this routes:

```
catalogue_product_index:
    pattern: /admin/catalogue/products
    methods: [GET]
    defaults:
        _controller: catalogue.controller.product:indexAction
        _sylius:
            method: filterPaginator
            arguments: [$criteria, $sorting, $deleted]

catalogue_product_create:
    pattern: /admin/catalogue/products/new
    methods: [GET, POST]
    defaults:
        _controller: catalogue.controller.product:createAction
        _sylius:
            redirect: catalogue_product_índex
```

When a create a "product", with this path, it redirect to **"catalogue_product_index"** and in the **browser url** shows **"/admin/catalogue/products?id=32"**, where **"32"** is the new product **ID**.

It is possible redirect to índex and not show the **ID** in url?

Thanks.

Paulo Dias

PR #5218, opened 2016-06-09
[Resource] Allow to redirect to route without default query parameters

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets |
| License         | MIT

Default redirect route parameters are useful, but sometimes can lead to errors (for example with ``isOpen()`` method on *Behat* ``Page``) and it should be possible to define them as empty array.


pjedrzejewski:
Thanks Mateusz! 👍

tchapi:
Hello
I have been having the problem described in https://github.com/Sylius/Sylius/issues/1214, namely, that it was not possible to redirect with empty parameters — does this patch fix this ? Looking at the code, it seems it does but I'm not sure

Thanks !

PR #5220, opened 2016-06-09
[Behat][Shop] New checkout shipping step

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets |
| License         | MIT

Contains commit from https://github.com/Sylius/Sylius/pull/5218.

To-do:
- [x] scenarios for not being able to select shipping method not available for current channel (need [this](https://github.com/Sylius/Sylius/pull/5216) changes), could be done in separate PR


features/checkout/addressing_order/filling_shipping_and_billing_address_details.feature
```diff
@@ -7,14 +7,16 @@ Feature: Addressing an order
...
+        And the store has "Raven Post" shipping method with "€10.00" fee
         And I am logged in customer
 
     @ui
     Scenario: Address an order without different billing address
         Given I have product "PHP T-Shirt" in the cart
         And I am at the checkout addressing step
         When I specify the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
-        And I proceed to the next step
+        And I proceed to the shipping step
```
pjedrzejewski: Why changing "next" to "shipping"?

features/checkout/shipping_order/preventing_not_available_shipping_method_selection.feature
```diff
@@ -0,0 +1,46 @@
+@checkout
+Feature: Preventing not available shipping method selection
+    In order to ship my order properly
+    As a Customer
+    I want to not be able to choose not available shipping methods
+
+    Background:
+        Given the store operates on a single channel in "France"
+        And the store has a product "Targaryen T-Shirt" priced at "$19.99"
```
pjedrzejewski: Dracarys all the bugs! 💃

features/checkout/addressing_order/filling_shipping_and_billing_address_details.feature
Zales0123: As we should somehow differentiate "next step" action between steps (for now its addressing and shipping)... Or maybe we could use ``CurrentPageResolver`` for that? which option is better in this case?

features/checkout/shipping_order/preventing_not_available_shipping_method_selection.feature
```diff
@@ -0,0 +1,46 @@
...
+        And there is a zone "The Rest of the World" containing all other countries
+        And the store has "Dragon Post" shipping method with "€30.00" fee for the rest of the world
+        And the store has disabled "Raven Post" shipping method with "€10.00" fee
+        And I have product "Targaryen T-Shirt" in the cart
+        When I am at the checkout addressing step
+        And I specify the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
+        And I proceed to the shipping step
+        Then I should not be able to select "Raven Post" shipping method
+        And I should not be able to select "Dragon Post" shipping method
+        And I should be alerted, that there is no shipping method available for me
```
pjedrzejewski: And I should be informed that my order cannot be shipped to this address

src/Sylius/Behat/Context/Ui/CheckoutContext.php
```diff
@@ -156,23 +165,100 @@ public function iSpecifyTheBillingAddressAs(AddressInterface $address)
...
+    }
+
+    /**
+     * @When I specify the billing's city as :cityName
+     * @When I do not specify the billing's city
+     */
+    public function iSpecifyTheBillingSCityAs($cityName = null)
+    {
+        $this->addressingPage->specifyBillingAddressCity($cityName);
+    }
```
pjedrzejewski: Mistake during rebase? These steps are removed IIRC and we use full address object.

src/Sylius/Behat/Context/Ui/CheckoutContext.php
```diff
@@ -156,23 +165,100 @@ public function iSpecifyTheBillingAddressAs(AddressInterface $address)
...
+     * @When I proceed to the shipping step
+     * @When I try to proceed to the shipping step
+     */
+    public function iProceedWithTheShippingStep()
     {
         $this->addressingPage->nextStep();
     }
 
     /**
+     * @When I proceed with the payment step
```
pjedrzejewski: I see that we need to use appropriate page to confirm the step, but I think it would be nice to decouple the step definitions from what is next. So this could be:

```
I complete the addressing step
I try to complete the addressing step
```

$this->addressingPage->completeStep();

This way we have a bit cleaner structure and if order of steps changes for whatever reason - we can easily reflect that in Behat. WDYT?

Zales0123: Sure, that makes sense to me 👍

features/checkout/shipping_order/selecting_order_shipping_method.feature
```diff
@@ -0,0 +1,19 @@
...
+        And the store has "Raven Post" shipping method with "€10.00" fee
+        And the store has "Dragon Post" shipping method with "€30.00" fee
+        And I am logged in customer
+
+    @ui
+    Scenario: Selecting one of available shipping method
+        Given I have product "Targaryen T-Shirt" in the cart
+        And I specified the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
+        When I select "Raven Post" shipping method
+        And I proceed with the payment step
```
michalmarcinkowski: proceed ~~with~~ **to**

src/Sylius/Component/Core/Resolver/DefaultShippingMethodResolver.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $this->shippingMethodRepository = $shippingMethodRepository;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefaultShippingMethod()
+    {
+        $shippingMethods = $this->shippingMethodRepository->findAll();
```
michalmarcinkowski: How about using `ShopperContext` to get current channel and use `findEnabledForChannel` repository method?

michalmarcinkowski: How about passing channel object from order to this method and use `findEnabledForChannel` repository method?

michalmarcinkowski: Or we could rename this method to `getDefaultShippingMethodForOrder(OrderInterface $order)` and get the channel from order inside this method.

Zales0123: I agreed with the first proposition (passing a channel to method). We could want to use this service somewhere else, to get default shipping method for channel, not necessarily for channel from order, wdyt?

src/Sylius/Bundle/CoreBundle/Form/EventSubscriber/AvailableShippingMethodsSubscriber.php
```diff
@@ -0,0 +1,82 @@
...
+        $data = $event->getData();
+        $form = $event->getForm();
+
+        $form->add('shipments', 'collection', [
+            'type' => 'sylius_checkout_shipment',
+            'label' => false,
+            'options' => [
+                'criteria' => [
+                    'enabled' => true,
+                    'zone' => $this->getZonesForOrder($data),
```
michalmarcinkowski: We should only pass `order` to the `sylius_checkout_shipment` type and these criteria should be removed.

src/Sylius/Component/Core/Resolver/DefaultShippingMethodResolver.php
michalmarcinkowski: After quick brainstorm we agreed to have `getDefaultShippingMethod(ShipmentInterface $shipment)`. In the component we should simply return first enabled shipping method and in the `Core` we should return first enabled shipping method for channel which is taken from order that is taken from shipment :)

src/Sylius/Bundle/CoreBundle/Form/EventSubscriber/AvailableShippingMethodsSubscriber.php
pjedrzejewski: Not exactly:

Here we should only have ``collection`` type using ``sylius_shipment`` (we can most likely remove the ``sylius_checkout_shipment`` type). And ``sylius_shipment`` type will add a ``sylius_shipping_method_choice`` type with option ``subject`` equal to currently edited shipment - so it has to be done via listener.

features/checkout/shipping_order/selecting_order_shipping_method.feature
```diff
@@ -0,0 +1,19 @@
+@checkout
+Feature: Selecting order shipping method
+    In order to ship my order properly
+    As a Customer
+    I want to be able to choose shipping method
```
tuka217: a shipping method ?

src/Sylius/Bundle/CoreBundle/Form/EventSubscriber/AvailableShippingMethodsSubscriber.php
michalmarcinkowski: @pjedrzejewski that is for sure a better option :)

features/checkout/addressing_order/filling_shipping_and_billing_address_details.feature
```diff
@@ -7,14 +7,16 @@ Feature: Addressing an order
     Background:
         Given the store operates on a single channel in "France"
         And the store has a product "PHP T-Shirt" priced at "$19.99"
+        And the store has "Raven Post" shipping method with "€10.00" fee
```
pjedrzejewski: Why do we need to define it here?

pjedrzejewski: OK, I see, but maybe we can use ``And the store ships everything for free`` as the fee is not relevant to this scenario.

src/Sylius/Behat/Context/Setup/ShippingContext.php
```diff
@@ -155,14 +171,17 @@ public function theShippingMethodIsDisabled(ShippingMethodInterface $shippingMet
...
     private function createShippingMethod(
         $name,
         $code = null,
         ZoneInterface $zone = null,
         $locale = 'en',
         $configuration = ['amount' => 0],
-        $calculator = DefaultCalculators::FLAT_RATE
+        $calculator = DefaultCalculators::FLAT_RATE,
+        $enabled = true,
+        $addForCurrentChannel = true
```
pjedrzejewski: Why do we need this?

Zales0123: We need to create shipping method not assigned to any channel to check if its not available - see [this scenario](https://github.com/Sylius/Sylius/pull/5220/files#diff-35c7c77e6ee2837180922a35f25cf2d0R35);

src/Sylius/Behat/Page/Shop/Checkout/ShippingPage.php
```diff
@@ -0,0 +1,78 @@
...
+
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function hasNoShippingMethodsMessage()
+    {
+        return null !== $this->getDocument()->find('css', '#no_shipping_methods');
```
pjedrzejewski: This should be an element. #sylius-order-cannot-be-shipped?

src/Sylius/Bundle/CoreBundle/Checkout/Step/ShippingStep.php
```diff
@@ -108,13 +108,6 @@ protected function createCheckoutShippingForm(OrderInterface $order)
...
 
-        return $this->createForm('sylius_checkout_shipping', $order, [
-            'criteria' => [
-                'zone' => !empty($this->zones) ? array_map(function ($zone) {
-                    return $zone->getId();
-                }, $this->zones) : null,
-                'enabled' => true,
-            ],
-        ]);
+        return $this->createForm('sylius_checkout_shipping', $order);
```
pjedrzejewski: 😍

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/ShippingType.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder->add('shipments', 'collection', [
+            'type' => 'sylius_checkout_shipment',
+            'label' => false,
+
```
pjedrzejewski: Extra blank line.

src/Sylius/Bundle/ShopBundle/Resources/views/Checkout/shipping.html.twig
```diff
@@ -0,0 +1,45 @@
+
```
pjedrzejewski: Extra blank line.

src/Sylius/Behat/Context/Setup/ShippingContext.php
pjedrzejewski: 👍

src/Sylius/Behat/Page/Shop/Checkout/ShippingPageInterface.php
```diff
@@ -0,0 +1,39 @@
...
+
+use Sylius\Behat\Page\SymfonyPageInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface ShippingPageInterface extends SymfonyPageInterface
+{
+    /**
+     * @param $shippingMethod
```
tuka217: missing type hint

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/ShipmentType.php
```diff
@@ -50,25 +50,15 @@ public function __construct($dataClass, TranslatorInterface $translator)
...
     {
-        $criteria = $options['criteria'];
-        $channel = $options['channel'];
-
-        $notBlank = new NotBlank(['groups' => ['sylius']]);
-        $notBlank->message = $this->translator->trans('sylius.checkout.shipping_method.not_blank', [], 'validators');
-
         $builder
-            ->addEventListener(FormEvents::PRE_SET_DATA, function (FormEvent $event) use ($notBlank, $criteria, $channel) {
+            ->addEventListener(FormEvents::PRE_SET_DATA, function (FormEvent $event) {
```
tuka217: Maybe EventSubscriber?

src/Sylius/Component/Core/Resolver/DefaultShippingMethodResolver.php
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public function __construct(ShippingMethodRepositoryInterface $shippingMethodRepository)
+    {
+        $this->shippingMethodRepository = $shippingMethodRepository;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefaultShippingMethod(ShipmentInterface $shippingSubject)
```
tuka217: If it is shippingSubject then the interface shouldn't be ShippingSubjectInterface?

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/ShippingType.php
```diff
@@ -0,0 +1,51 @@
...
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults([
+            'validation_groups' => ['sylius'],
```
michalmarcinkowski: This should not be hardcoded, but injected as it is e.g. in `UserChangePasswordType`

src/Sylius/Component/Core/Resolver/DefaultShippingMethodResolver.php
```diff
@@ -0,0 +1,65 @@
...
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Core\Resolver;
+
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Shipping\Model\ShipmentInterface;
```
tuka217: Shouldn't be `Core\Model\ShipmentInterface`?

src/Sylius/Bundle/ShippingBundle/Form/Type/ShippingMethodChoiceType.php
```diff
@@ -82,26 +82,22 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     public function configureOptions(OptionsResolver $resolver)
     {
         $choiceList = function (Options $options) {
-            $resolvedMethods = [];
+            $methods = $this->repository->findAll();
```
michalmarcinkowski: We shouldn't query for all methods everytime. Let's use `if-else` or extract it to a separate function.

src/Sylius/Component/Core/Resolver/DefaultShippingMethodResolver.php
Zales0123: The problem is, it implements interface from *Shipping* component. There is no problem with calling ``getOrder()``, as we [assert that subject is shipment from Core](https://github.com/Sylius/Sylius/pull/5220/files#diff-8f588422e1d608e78fb3477db05afd6fR43), but we must keep types compatibility between this *Core* service and *Shipping* interface.

src/Sylius/Component/Core/Resolver/DefaultShippingMethodResolver.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        $this->shippingMethodRepository = $shippingMethodRepository;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefaultShippingMethod(ShipmentInterface $shippingSubject)
+    {
+        Assert::isInstanceOf($shippingSubject, \Sylius\Component\Core\Model\ShipmentInterface::class);
```
michalmarcinkowski: We should use `use` here :)

src/Sylius/Component/Core/Resolver/DefaultShippingMethodResolver.php
```diff
@@ -0,0 +1,65 @@
...
+            if (!$channel->hasShippingMethod($shippingMethod)) {
+                unset($shippingMethods[$key]);
+            }
+        }
+
+        if (empty($shippingMethods)) {
+            return null;
+        }
+
+        return array_values($shippingMethods)[0];
```
michalmarcinkowski: I would say that current implementation can be simply:
```php
foreach ($shippingMethods as $shippingMethod) {
    if ($channel->hasShippingMethod($shippingMethod)) {
        return $shippingMethod;
    }
}

return null;
```

michalmarcinkowski: Instead of returning `null` shouldn't we throw exception when there is no shipping method available? /cc @pjedrzejewski

src/Sylius/Component/Shipping/Model/ShipmentInterface.php
```diff
@@ -48,7 +48,7 @@ public function getMethod();
     /**
      * @param ShippingMethodInterface $method
      */
-    public function setMethod(ShippingMethodInterface $method);
+    public function setMethod(ShippingMethodInterface $method = null);
```
michalmarcinkowski: I believe setting `null` should not be possible. We should also prevent from creating Shipment without shipping method on a database level.

Zales0123: Unfortunately we must allow setting ``null`` on shipment method to make validation works. It must get object with invalid state to add violation, otherwise selecting no shipping method in shipping step would lead to an exception.

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/ShipmentType.php
Zales0123: Unfortunately, ``addEventSubscriber`` method requires ``EventSubscriber`` object as an argument. As it's legacy shipping step type and will be probably removed in the nearest future, I would leave it as it is.

src/Sylius/Component/Shipping/Resolver/DefaultShippingMethodResolverInterface.php
```diff
@@ -0,0 +1,29 @@
...
+namespace Sylius\Component\Shipping\Resolver;
+
+use Sylius\Component\Core\Model\ShippingMethodInterface;
+use Sylius\Component\Shipping\Model\ShipmentInterface;
+use Sylius\Component\Shipping\Model\ShippingSubjectInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface DefaultShippingMethodResolverInterface
```
pamil: Do we need that "Default" part?

src/Sylius/Component/Shipping/Resolver/DefaultShippingMethodResolverInterface.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    /**
+     * @param ShipmentInterface $shipment
+     *
+     * @return ShippingMethodInterface|null
+     */
+    public function getDefaultShippingMethod(ShipmentInterface $shipment);
```
pamil: Same here.

src/Sylius/Component/Shipping/Resolver/DefaultShippingMethodResolverInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface DefaultShippingMethodResolverInterface
+{
+    /**
+     * @param ShipmentInterface $shipment
+     *
+     * @return ShippingMethodInterface|null
```
pamil: Shouldn't we throw an exception instead?

src/Sylius/Component/Core/Resolver/DefaultShippingMethodResolver.php
```diff
@@ -0,0 +1,62 @@
...
+
+        $shippingMethods = $this->shippingMethodRepository->findBy(['enabled' => true]);
+        if (empty($shippingMethods)) {
+            return null;
+        }
+
+        /** @var ChannelInterface $channel */
+        $channel = $shipment->getOrder()->getChannel();
+
+        foreach ($shippingMethods as $key => $shippingMethod) {
```
pamil: Looks like we don't use the key.

src/Sylius/Component/Shipping/Model/ShipmentInterface.php
michalmarcinkowski: You're right... :(

michalmarcinkowski:
As I've already mentioned, we should add not nullable constraint to shipping method on Shipment.

src/Sylius/Component/Shipping/Resolver/DefaultShippingMethodResolverInterface.php
michalmarcinkowski: If we decide on throwing an exception instead of returning `null` then I would leave the `Default` prefix.

Zales0123:
Happily, I [remembered about it](https://github.com/Sylius/Sylius/pull/5220/files#diff-f8dfd8f1bed7287bd860bcc9f9a5b87cR95) 😄

pjedrzejewski:
Thanks Mateusz! 👍  Nice work. :)

michalmarcinkowski:
@Zales0123 I meant the database level :)

Zales0123:
[Also done](https://github.com/Sylius/Sylius/blob/f5d1b5f7fe845adfcc2fac485d6e19cd1708e03b/src/Sylius/Bundle/ShippingBundle/Resources/config/doctrine/model/Shipment.orm.xml#L23), in fact long time ago 🐎

michalmarcinkowski:
@Zales0123 great! 💥  I didn't saw it in this PR and assumed it was nullable same as it is on [Payment](https://github.com/michalmarcinkowski/Sylius/blob/514cf8b18d496f40ce691e58cf4b89901327819f/src/Sylius/Bundle/PaymentBundle/Resources/config/doctrine/model/Payment.orm.xml#L23-L23) with payment method. Sorry about the hassle! 🐗 🔨

Zales0123:
*Better check than regret* 😄

PR #5221, opened 2016-06-09
[Bug Fix][ShippingBundle] Default shipping method choice form type set to correct class

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no|yes
| BC breaks?      | no|yes
| License         | MIT

Set correct default form choice type for shipping method. It expects this type since it is overriden in the services.xml (https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ShippingBundle/Resources/config/services.xml#L74); This will prevent some future headaches for other people trying to use standalone shipping bundle.


steffenbrem:
Should I close this in favor of #5216?

pjedrzejewski:
@steffenbrem Let's keep it open until we merge #5216 to make sure this is fixed.

steffenbrem:
This has been fixed in #5216

Issue #5173, opened 2016-06-03
Invalid method name that leads to confusion and bugs


michalmarcinkowski:
After changing the logic to always save currency code instead of `Currency` object on `Order` the method `get/setCurrency` was never renamed to `get/setCurrencyCode`. That leads to confusion and bugs. Example can be found in the new [Setup/OrderContext](https://github.com/michalmarcinkowski/Sylius/blob/5b4f09a6f5a636aab8ec6f7f3a088f0cb6a20cb9/src/Sylius/Behat/Context/Setup/OrderContext.php#L469-L469) where we set and typehint to `CurrencyInterface` which later leads to bugs which were not revealed in the main repo yet.

The same situation can be found on `Payment` model which also has `get/setCurrency` method that actually accepts currency code not an object.

### The solution 
- [x] Rename all `get/setCurrency` methods that accept currency code to `get/setCurrencyCode`
- [x] In `setCurrencyCode` method we should use `Assert::string` to ensure we don't set any object
- [x] Add [Currency validation constraint](http://symfony.com/doc/current/reference/constraints/Currency.html) to models, in order to prevent from setting invalid code
- [x] Rename all variables that represents currency code to `currencyCode` instead of `currency` that indicates an object

pjedrzejewski:
👍 👍 👍

michalmarcinkowski:
Done in #5228.

PR #5228, opened 2016-06-10
Change currency to currency code

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | mentioned in #5173 
| License         | MIT


src/Sylius/Component/Core/Model/Customer.php
```diff
@@ -55,17 +55,17 @@ public function __construct()
     /**
      * {@inheritdoc}
      */
-    public function setCurrency($currency)
+    public function setCurrencyCode($currencyCode)
     {
-        $this->currency = $currency;
```
michalmarcinkowski: `Assert::string` missing

src/Sylius/Bundle/WebBundle/Resources/views/Backend/Order/create.html.twig
```diff
@@ -11,15 +11,15 @@
...
+    {#<h1><span class="glyphicon glyphicon-plus"></span> {{ 'sylius.ui.new_order'|trans|raw }}</h1>#}
+{#</div>#}
+{##}
+{#{{ form_errors(form) }}#}
+{##}
+{#<form action="{{ path('sylius_backend_order_create') }}" method="post" class="form-horizontal" novalidate>#}
+    {#<center>Not implemented yet.</center>#}
+    {#<hr>#}
+    {#{{ create() }}#}
+{#</form>#}
```
michalmarcinkowski: I believe this template and corresponding routing and references can be removed.

src/Sylius/Bundle/CoreBundle/Checkout/Step/FinalizeStep.php
```diff
@@ -77,9 +77,9 @@ protected function completeOrder(OrderInterface $order)
 
         $this->applyTransition(OrderCheckoutTransitions::TRANSITION_COMPLETE, $order);
         $this->get('sm.factory')->get($order, OrderTransitions::GRAPH)->apply(OrderTransitions::SYLIUS_CREATE, true);
-        if ($order->getCurrency() !== $currencyProvider->getBaseCurrency()) {
+        if ($order->getCurrencyCode() !== $currencyProvider->getBaseCurrency()) {
```
michalmarcinkowski: `$currencyProvider->getBaseCurrency()` returns `Currency` object, so using `!==` will always return true. IMO we can remove this `if` and left the logic to always save the exchange rate on `Order`.

pjedrzejewski:
@GSadee Please add note to UPGRADE file.

GSadee:
Done 👍

michalmarcinkowski:
Good job Grzesiu! 👍

PR #5232, opened 2016-06-11
[Resource] Align createPaginator signature on Doctrines methods

cdaguerre:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Related tickets |  |
| License | MIT |
##### Doctrine methods

Align `createPaginator` signature on those of Doctrine's `findBy` methods. 

``` php
public function findBy(array $criteria, array $orderBy = null, $limit = null, $offset = null)
```

``` php
public function findOneBy(array $criteria, array $orderBy = null)
```
##### Fix empty sorting parameter

This route will fail if there is no `sorting` query parameter because sorting cant be null. This PR fixes that.

``` yml
sylius_api_cart_index:
    path: /
    methods: [GET]
    defaults:
        _controller: sylius.controller.cart:indexAction
        _sylius:
            repository:
                method:    createPaginator
                arguments: [$criteria, $sorting]
```

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Doctrine/ORM/EntityRepository.php
+++ b/src/Sylius/Bundle/ResourceBundle/Doctrine/ORM/EntityRepository.php
@@ -47,12 +47,12 @@ class EntityRepository extends BaseEntityRepository implements RepositoryInterfa
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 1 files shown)


loic425:
@cdaguerre It seems to me that there is the same problem with no criteria. Can you check that ?

PR #5233, opened 2016-06-11
[Resource] Check for ResourceInterface during resource registration

cdaguerre:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

This is a suggestion to correct the fact that `SyliusResourceBundle` expects resources to live in the `Sylius\\` namespace, as mentioned [here](https://github.com/Sylius/Sylius/pull/5075#issuecomment-223984424).

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/EventListener/AbstractDoctrineSubscriber.php
+++ b/src/Sylius/Bundle/ResourceBundle/EventListener/AbstractDoctrineSubscriber.php
@@ -14,6 +14,7 @@ namespace Sylius\Bundle\ResourceBundle\EventListener;
 use Doctrine\Common\EventSubscriber;
 use Doctrine\Common\Persistence\Mapping\ClassMetadata;
 use Sylius\Component\Resource\Metadata\RegistryInterface;
+use Sylius\Component\Resource\Model\ResourceInterface;
 
 /**
  * @author Ben Davies <ben.davies@gmail.com>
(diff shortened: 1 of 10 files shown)


src/Sylius/Bundle/ResourceBundle/EventListener/AbstractDoctrineSubscriber.php
```diff
@@ -40,6 +41,10 @@ public function __construct(RegistryInterface $resourceRegistry)
      */
     protected function isSyliusClass(ClassMetadata $metadata)
```
bendavies: `isResource`

nakashu:
@cdaguerre any update o this ?

isometriks:
This has mostly worked for me. The one issue I found was that in the ORM driver (I'm sure the others as well but only using ORM) was that after this line - https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ResourceBundle/EventListener/ORMMappedSuperClassSubscriber.php#L78

You need to do this:

```php
$parentMetadata->wakeupReflection(new \Doctrine\Common\Persistence\Mapping\RuntimeReflectionService());
```

You need to wake up the reflection, because when this gets passed into isSyliusClass then `$metadata->getReflectionClass()` will fail because `getReflectionClass` is just a simple getter. If the reflection hasn't been instantiated then this call will always fail and isSyliusClass will always return false.

There might obviously be a better way to create the metadata and / or move it to the abstract parent to do this, but fixing this makes this work for me for non-sylius base classes.

isometriks:
This can be closed now, this was all put into #5371

PR #5242, opened 2016-06-13
[Shipping] Add translatable description to ShippingMethod

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets |
| License         | MIT

While working on https://github.com/Sylius/Sylius/pull/5220, we've come up with an idea, that it would be nice to have translatable description on ``ShippingMethod`` (as, for example, on ``PaymentMethod``).


pjedrzejewski:
Thank you Mateusz! 👍

PR #5246, opened 2016-06-14
Add missing setter method

sblaut:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

One would expect this method to be present in the interface. In fact, Sylius\Component\Core\Model\Order implements this method and uses an @inheritdoc annotation.


pjedrzejewski:
Hi @sblaut, in this interface we do not want to make an assumption about how coupon is set. We only need to get it if it is there. So I'd say this method should go to OrderInterface if it is missing there.

sblaut:
@pjedrzejewski makes total sense, and the method is in the OrderInterface of the Core component. False alarm.

Issue #5244, opened 2016-06-13
[Grid] Default options / options vaidation.


dantleech:
Currently

``` php
sylius_grid:
    grids:
            fields:
                date:
                    type: datetime
```

Produces a twig error:

```
Expected the key "format" to exist
```

And I can specify unknown keys in the options, for example:

```

                date:
                    type: datetime
                    label: sylius.ui.date
                    options:
                        asd: asd
```

Would it be possible to use the Symfony OptionsResolver here?

``` php
$optionsResolver = new OptionsResolver();
$options->setDefault('format', 'Y-m-d H:m:s');
```

pjedrzejewski:
I used OptionsResolver in my original implementation of grids and I was hoping to move it to new version shortly after merged, but it did not work out. :) Have a look at ``grids`` branch of this repo. I think we should definitely do that.

michalmarcinkowski:
Fixed.

PR #5247, opened 2016-06-14
[Grid] Grid options

dantleech:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | #5244
| License         | MIT

This PR (will) integrate the OptionsResolver with the Grid component `FieldTypeInterface` and `FilterInterface` classes.

TODO:

- [x] FieldTypes
- [x] <strike>Filters</strike> Filters already have form types

### DIFF
--- a/src/Sylius/Bundle/GridBundle/Renderer/TwigGridRenderer.php
+++ b/src/Sylius/Bundle/GridBundle/Renderer/TwigGridRenderer.php
@@ -18,6 +18,7 @@ use Sylius\Component\Grid\Renderer\GridRendererInterface;
 use Sylius\Component\Grid\View\GridView;
 use Sylius\Component\Registry\ServiceRegistryInterface;
 use Symfony\Component\Form\FormFactoryInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * @author Paweł Jędrzejewski <pawel@sylius.org>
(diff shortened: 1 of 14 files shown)


src/Sylius/Bundle/GridBundle/Renderer/TwigGridRenderer.php
```diff
@@ -92,7 +93,12 @@ public function render(GridView $gridView, $template = null)
      */
     public function renderField(GridView $gridView, Field $field, $data)
     {
-        return $this->fieldsRegistry->get($field->getType())->render($field, $data);
+        $fieldType = $this->fieldsRegistry->get($field->getType());
+        $resolver = new OptionsResolver();
+        $fieldType->configureOptions($resolver);
+        $options = $resolver->resolve($field->getOptions());
```
dantleech: I think eventually both the retrieval of the field type service and the resolution of the options could be done at a higher level, and so the method signature for the rednerer could be something like:

```php
public function renderField(GridView $gridView, Field $field, FieldTypeInterface $type, array $options, $data)
```

But nothing to do with this PR.

dantleech:
@pjedrzejewski the Filters have form types, which already have options. So, unless we separate the form options from the filter options then I guess it makes no sense to add additional option configuration (?).

pjedrzejewski:
Good stuff, thanks Dan! Yes, in previous implementation filter forms had to have the same option configuration, even if they did not use these options which sucked, but I think I'd be fine with that, because it is better than having no control over and no default options. WDYT?

PR #5259, opened 2016-06-15
[Shipping] Removed unneeded composite resolver interface

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets |
| License         | MIT

Change proposed in [this comment](https://github.com/Sylius/Sylius/pull/5216#issuecomment-225164083).


src/Sylius/Bundle/ShippingBundle/spec/Form/Type/ShippingMethodChoiceTypeSpec.php
```diff
@@ -13,24 +13,20 @@
...
-use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * @author Arnaud Langlade <arn0d.dev@gamil.com>
  */
 class ShippingMethodChoiceTypeSpec extends ObjectBehavior
 {
     function let(
-        CompositeMethodsResolverInterface $compositeMethodsResolver,
+        MethodsResolverInterface $compositeMethodsResolver,
```
CarwynNelson: This variable is called `$compositeMethodsResolver` but the interface is now `MethodsResolverInterface`. It's just a small thing but this now makes it seem like a composite methods resolver is expected to be injected in, even though that won't necessarily be true.

Similar occurrences elsewhere.

src/Sylius/Component/Shipping/spec/Resolver/CompositeMethodsResolverSpec.php
```diff
@@ -38,7 +37,7 @@ function it_is_initializable()
 
     function it_implements_composite_methods_resolver_interface()
```
CarwynNelson: This function name should really be changed to represent the change in interface away from a composite methods resolver to a methods resolver.

src/Sylius/Component/Shipping/Resolver/CompositeMethodsResolver.php
```diff
@@ -45,4 +45,12 @@ public function getSupportedMethods(ShippingSubjectInterface $shippingSubject)
 
         return [];
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supports(ShippingSubjectInterface $subject)
+    {
+        return true;
```
pamil: It supports a shipping subject only if one of nested resolvers supports it.

pamil: Anyway, it seems like this method is not used anywhere except `CompositeMethodsResolver::getShippingMethods()` and always returns true. We can remove it and then that method will return first not-empty result from nested methods resolver.

CarwynNelson: `ShippingMethodsByZonesAndChannelResolver` seems to be the only class which implements this interface that does not return true. Could there be a possibility that a separate interface is required for that class and `supports()` can be removed from the `MethodResolverInterface`? Not sure if I'm over thinking this as I've only just started diving into Sylius (just looking at some PR's to get the feel of it)

https://github.com/Sylius/Sylius/blob/db5a8dcd3b23fbaea73deaf80ddc6487ea5b60b9/src/Sylius/Component/Core/Resolver/ShippingMethodsByZonesAndChannelResolver.php

pjedrzejewski:
@CarwynNelson Thanks for the review! :)

src/Sylius/Bundle/ShippingBundle/Form/Type/ShippingMethodChoiceType.php
```diff
@@ -52,12 +50,12 @@ class ShippingMethodChoiceType extends AbstractType
...
 
     /**
-     * @param CompositeMethodsResolverInterface $compositeShippingMethodsResolver
+     * @param MethodsResolverInterface $compositeShippingMethodsResolver
      * @param ServiceRegistryInterface $calculators
      * @param RepositoryInterface $repository
      */
     public function __construct(
-        CompositeMethodsResolverInterface $compositeShippingMethodsResolver,
+        MethodsResolverInterface $compositeShippingMethodsResolver,
```
michalmarcinkowski: Then we should also rename the variable.

src/Sylius/Bundle/ShippingBundle/spec/Form/Type/ShippingMethodChoiceTypeSpec.php
michalmarcinkowski: Same as above.

src/Sylius/Component/Shipping/Resolver/CompositeMethodsResolver.php
Zales0123: @pamil @CarwynNelson the idea was, to unified all shipping methods resolvers with one ``MethodsResolverInterface``, to give possibility to inject both specific method resolver and composite resolver, which contains resolvers registry. Due to that fact, we must keep ``supports()`` method in all of them, as they can be used separately. I agree, that returning ``true`` here is nasty, but I can't came up with better option.

pamil:
```php
foreach ($this->resolversRegistry->all() as $resolver) {
     if ($resolver->supports($shippingSubject)) {
         return true;
     }
}

return false;
```

pjedrzejewski:
Thank you Mateusz. :)

PR #5260, opened 2016-06-15
[Shop] Break infinite loop

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/config/routing/partial/product.yml
+++ b/src/Sylius/Bundle/ShopBundle/Resources/config/routing/partial/product.yml
@@ -7,6 +7,7 @@ sylius_shop_partial_product_index_latest:
     defaults:
         _controller: sylius.controller.product:indexAction
         _sylius:
+            permission: false
             template: $template
             repository:
                 method: findLatestByChannel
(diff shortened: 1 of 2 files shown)


CarwynNelson:
What causes this infinite loop to occur? Is there a github issue which provides more details about this bug?

Arminek:
Rendering partial on main shop layout. ``{% render(url('sylius_shop_partial_taxon_index_by_code', {'code': 'category', 'template': '@SyliusShop/Taxon/_horizontalMenu.html.twig'})) %}``

CarwynNelson:
Are there any tests that could be written to prove that an infinite loop does indeed occur under these circumstances? It sounds like it would be quite a tricky one to test, but it would definitely be worth it. I can imagine somebody else later on down the line changing this bit of config and not realising that it causes an infinite loop.

pjedrzejewski:
@CarwynNelson The reason is our RBAC system, which checks for permissions and roles of users. And in our current test environment, it is enabled for tests which cover RBAC, but disabled for rest, for performance reasons. That being said, we should definitely enable it for all scenarios, which would reveal such issues early.

Also, ``permission: false`` should be a default. Because it is more annoying to set it on all unsecured routes than configure it for entire admin.

pjedrzejewski:
Thank you Arek!

PR #5262, opened 2016-06-15
[Shop] Add initial images support

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig
+++ b/src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig
@@ -171,7 +171,7 @@
 
 {% block sylius_image_widget %}
     <div class="ui upload box segment">
-        {% if form.vars.value is null %}
+        {% if form.vars.value.path|default(null) is null %}
             <label for="{{ form.file.vars.id }}" class="ui icon labeled button"><i class="cloud upload icon"></i> {{ 'sylius.ui.choose_file'|trans }}</label>
         {% else %}
             <img class="ui small bordered image" src="{{ form.vars.value.path|imagine_filter('sylius_small') }}" alt="" />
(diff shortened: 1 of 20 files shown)


src/Sylius/Component/Core/Model/Product.php
```diff
@@ -355,4 +355,16 @@ public function getImage()
...
         return $this->getFirstVariant()->getImage();
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getImages()
+    {
+        if (null === $this->getFirstVariant()) {
+            return new ArrayCollection();
```
CarwynNelson: I'm not sure if I'm reading this correctly but from the `ProductInterface` it looks like the `firstVariant()` is supposed to return an `ImageInterface` but in the case where there is no first variant we return an `ArrayCollection`. Is this correct? If so is it possible to convert the result of `getImages()` to an `ArrayCollection`, or possibly return a null image interface.

Eg:

```
if (null === $this->getFirstVariant() {
    return new EmptyImageInterface();
}

// etc ...
```

pjedrzejewski: @CarwynNelson Actually, it returns first ProductVariant from $variants collection, but product can have no variants at all, if not configured properly yet. Anyway, we will be adding images support on Product, so these methods will return images and first image from product itself. :)

src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig
```diff
@@ -171,7 +171,7 @@
 
 {% block sylius_image_widget %}
     <div class="ui upload box segment">
-        {% if form.vars.value is null %}
+        {% if form.vars.value.path|default(null) is null %}
```
michalmarcinkowski: Do we need `default(null)` here?

michalmarcinkowski:
Thanks Paweł!

PR #5219, opened 2016-06-09
[Fixtures] Brand new fixtures bundle

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | documentation in #5286, fixtures implementation in #5285
| License         | MIT

It's really a work in progress, but feel free to review :tada: It will be a standalone bundle, not depending on any of Sylius components or bundles.

PR #5263, opened 2016-06-15
Do not redefine constant

dantleech:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

I am using the fixutres in functional tests, this hack prevents the constant from being redefined (ofc it would be better if it were not required).


dantleech:
Doesn't work as the references are based on IDs, so it *has* to be redefined if it were to work. Would have to be refactored as a `$global` ...

pamil:
Brace yourself, #5219 is coming soon and old fixtures will be deleted! ❄️

PR #5240, opened 2016-06-13
[Shop] New checkout payment step

tuka217:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | -
| License         | MIT

PR #5265, opened 2016-06-15
[Behat][Shop] Seeing shipping details on order summary

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets |
| License         | MIT

Based on https://github.com/Sylius/Sylius/pull/5240

### DIFF
--- a/features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
+++ b/features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
@@ -11,23 +11,23 @@ Feature: Seeing order addresses on order summary page
         And the store allows paying offline
         And I am logged in customer
 
-    @todo
+    @ui
     Scenario: Seeing the same shipping and billing address on order summary
         Given I have product "Lannister Coat" in the cart
-        And I am at the checkout addressing step
-        When I specify the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
+        When I specified the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
         And I complete order with "Free" shipping method and "Offline" payment
-        Then I should be on the checkout summary page
+        Then I should be on the checkout summary step
         And I should see this shipping address as shipping and billing address
 
-    @todo
+    @ui
     Scenario: Seeing different shipping and billing address on order summary
         Given I have product "Lannister Coat" in the cart
         And I am at the checkout addressing step
         When I specify the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
         And I choose the different billing address
         And I specify the billing address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Eddard Stark"
+        And I complete the addressing step
         And I complete order with "Free" shipping method and "Offline" payment
-        Then I should be on the checkout summary page
+        Then I should be on the checkout summary step
         And I should see this shipping address as shipping address
         And I should see this billing address as billing address
--- a/features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
+++ b/features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
@@ -1,0 +1,33 @@
+@checkout
+Feature: Seeing order addresses on order summary page
+    In order be certain about shipping and bulling address
+    As a Customer
+    I want to be able to see addresses on order summary page
+
+    Background:
+        Given the store operates on a single channel in "France"
+        And the store has a product "Lannister Coat" priced at "$19.99"
+        And the store ships everywhere for free
+        And the store allows paying offline
+        And I am logged in customer
+
+    @todo
+    Scenario: Seeing the same shipping and billing address on order summary
+        Given I have product "Lannister Coat" in the cart
+        And I am at the checkout addressing step
+        When I specify the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
+        And I complete order with "Free" shipping method and "Offline" payment
+        Then I should be on the checkout summary page
+        And I should see this shipping address as shipping and billing address
+
+    @todo
+    Scenario: Seeing different shipping and billing address on order summary
+        Given I have product "Lannister Coat" in the cart
+        And I am at the checkout addressing step
+        When I specify the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
+        And I choose the different billing address
+        And I specify the billing address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Eddard Stark"
+        And I complete order with "Free" shipping method and "Offline" payment
+        Then I should be on the checkout summary page
+        And I should see this shipping address as shipping address
+        And I should see this billing address as billing address
--- a/features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
+++ b/features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
@@ -1,33 +1,32 @@
 @checkout
 Feature: Seeing order addresses on order summary page
-    In order be certain about shipping and bulling address
+    In order to be certain about shipping and billing address
     As a Customer
-    I want to be able to see addresses on order summary page
+    I want to be able to see addresses on the order summary page
 
     Background:
         Given the store operates on a single channel in "France"
         And the store has a product "Lannister Coat" priced at "$19.99"
         And the store ships everywhere for free
-        And the store allows paying offline
-        And I am logged in customer
+        And the store allows paying with "Cash on Delivery"
+        And I am a logged in customer
 
     @ui
     Scenario: Seeing the same shipping and billing address on order summary
         Given I have product "Lannister Coat" in the cart
-        When I specified the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
-        And I complete order with "Free" shipping method and "Offline" payment
+        And I specified the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
+        And I proceed order with "Free" shipping method and "Cash on Delivery" payment
         Then I should be on the checkout summary step
-        And I should see this shipping address as shipping and billing address
+        And address to "Jon Snow" should be used for both shipping and billing of my order`
 
     @ui
     Scenario: Seeing different shipping and billing address on order summary
         Given I have product "Lannister Coat" in the cart
         And I am at the checkout addressing step
         When I specify the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
-        And I choose the different billing address
         And I specify the billing address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Eddard Stark"
         And I complete the addressing step
-        And I complete order with "Free" shipping method and "Offline" payment
+        And I proceed order with "Free" shipping method and "Cash on Delivery" payment
         Then I should be on the checkout summary step
-        And I should see this shipping address as shipping address
-        And I should see this billing address as billing address
+        And my order's shipping address should be to "Jon Snow"
+        But my order's billing address should be to "Eddard Stark"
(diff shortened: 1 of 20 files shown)


features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
```diff
@@ -0,0 +1,33 @@
+@checkout
+Feature: Seeing order addresses on order summary page
+    In order be certain about shipping and bulling address
```
GSadee: In order **to** be certain about shipping and **billing** address

features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
```diff
@@ -0,0 +1,33 @@
...
+    As a Customer
+    I want to be able to see addresses on order summary page
```
GSadee: ... on **an** order summary page ?

features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
```diff
@@ -0,0 +1,33 @@
...
+
+    Background:
+        Given the store operates on a single channel in "France"
+        And the store has a product "Targaryen T-Shirt" priced at "$19.99"
+        And the store ships everywhere for free
+        And the store allows paying offline
+        And I am logged in customer
```
tuka217: `a logged in customer`

features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
```diff
@@ -0,0 +1,33 @@
...
+    Scenario: Seeing different shipping and billing address on order summary
+        Given I have product "Lannister Coat" in the cart
+        And I am at the checkout addressing step
+        When I specify the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
+        And I choose the different billing address
+        And I specify the billing address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Eddard Stark"
+        And I complete the addressing step
+        And I complete order with "Free" shipping method and "Offline" payment
+        Then I should be on the checkout summary step
+        And I should see this shipping address as shipping address
```
pjedrzejewski: ???

src/Sylius/Behat/Page/Shop/Checkout/SummaryPage.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * @param string $countryCode
+     *
+     * @return string
+     */
+    private function getCountryName($countryCode)
+    {
+        return strtoupper(Intl::getRegionBundle()->getCountryName($countryCode, 'en'));
```
pjedrzejewski: Can't we use the converter from Addressing component?

features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
Zales0123: Ok, this looks quite confusing 😄  This step should check if typed shipping address is displayed on summary page... How do you think it should be written? ``And I should see typed shipping address as shipping address on order summary page``? ``And I should see typed shipping address on order summary page``?

src/Sylius/Behat/Page/Shop/Checkout/SummaryPage.php
Zales0123: Ok, unfortunately we can't ;/ This is ``CountryNameConverter`` with method ``convertToCode``, but we need country name from country code... I could implement this functionality inside this converter (or a separate one), but I don't know is it needed for now, wdyt?

features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
tuka217: What is wrong with `I should see this shipping address on order summary`?

Arminek: ``I should see this shipping address`` And ``I should see this billing address``. WDYT?

features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
```diff
@@ -0,0 +1,33 @@
...
+        Given the store operates on a single channel in "France"
+        And the store has a product "Lannister Coat" priced at "$19.99"
+        And the store ships everywhere for free
+        And the store allows paying offline
+        And I am a logged in customer
+
+    @ui
+    Scenario: Seeing the same shipping and billing address on order summary
+        Given I have product "Lannister Coat" in the cart
+        When I specified the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
```
michalmarcinkowski: `When I specify` or `Given I specified`.

features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
```diff
@@ -0,0 +1,33 @@
...
+        And I complete order with "Free" shipping method and "Offline" payment
+        Then I should be on the checkout summary step
+        And I should see this shipping address as shipping and billing address
```
michalmarcinkowski: We should avoid  "I should see". IMO the most readable scenario would be with repeated address:
`And shipping and billing address of the order should be "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"`
or we can compromise on:
`And shipping and billing address of the order should be the same as specified in the addressing step`

src/Sylius/Behat/Context/Ui/CheckoutContext.php
```diff
@@ -498,6 +553,17 @@ public function iShouldNotBeAbleToSelectPaymentMethod($paymentMethodName)
     }
 
     /**
+     * @Given I complete order with :shippingMethod shipping method and :paymentMethod payment
```
michalmarcinkowski: `When I complete ...` or `Given I completed ...`

src/Sylius/Behat/Resources/config/suites/ui/checkout/checkout.yml
```diff
@@ -11,13 +11,15 @@ default:
                 - sylius.behat.context.transform.lexical
                 - sylius.behat.context.transform.payment_method
                 - sylius.behat.context.transform.product
-                - sylius.behat.context.transform.shipping_method
                 - sylius.behat.context.transform.zone
+                - sylius.behat.context.transform.shared_storage
+                - sylius.behat.context.transform.shipping_method
```
michalmarcinkowski: Shouldn't we keep alphabetical order?

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/PaymentStepType.php
```diff
@@ -33,7 +32,7 @@ public function __construct(array $validationGroups = [])
     {
         $this->validationGroups = $validationGroups;
     }
-
+
```
michalmarcinkowski: Should be reverted.

src/Sylius/Behat/Page/Shop/Checkout/SummaryPage.php
michalmarcinkowski: Country**Name**Converter suggests it converts the name, so adding a method to convert a code is not a good idea. I wouldn't bother to create a separate class to extract this one line of code.

src/Sylius/Behat/Resources/config/suites/ui/checkout/checkout.yml
Zales0123: We should, unfortunately there is a conflict between ``the`` transformer from``SharedStorageContext`` and ``the rest of the world`` transformer from ``ZoneContext``. If ``sylius.behat.context.transform.shared_storage`` is included before ``sylius.behat.context.transform.zone``, part of step ``the rest of the world`` is identified as **resource from shared context with key "rest_of_the_world"**, not as **"Rest of the World" zone" ;/

features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
Zales0123: The problem is, we must have ``this|the shipping address`` part in the step (or similar), to have access to "shipping_address" resource from shared storage... What do you think about ``And shipping and billing address of the order should be the same as previously (specified shipping address)``?

src/Sylius/Behat/Resources/config/suites/ui/checkout/checkout.yml
michalmarcinkowski: 💥  Looks like we should do something about the regexp then...

Zales0123: Sure we should! However I think we both agree this is something should be done i separate PR ;)

features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature
michalmarcinkowski: Then the better option is to repeat the address rather than depend on object saved under `shipping_address` key.

Zales0123: I'm not sure... Therefore, in [this scenario](https://github.com/Sylius/Sylius/pull/5265/files#diff-377570f6ef22dac4556875c9756df64dR23) we would write address 4 times :| Of course using shared context is always quite dangerous, but it's a risk we took consciously when we implemented this service.

pjedrzejewski: Repeating the address is not good too. Why can't we do something like this:

```Then this address should be used for both shipping and billing```

For other case:

```Then this shipping/billing address should be used for my order```

We can check the rest inside of the context.

Comment (michalmarcinkowski) [inline features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature]:
```
When I specify the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
And I choose the different billing address
And I specify the billing address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Eddard Stark"
And I complete the addressing step
And I complete order with "Free" shipping method and "Offline" payment
Then I should be on the checkout summary step
And this shipping address should be used for my order # what is "this shipping address" ??
And this billing address should be used for my order
```

"this shipping/billing address" is confusing and does not show exactly what would be my final address. We should either be more specific `And the shipping address specified in addressing step should be used for my order` or direct `And "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow" should be used for my order shipping address`.

Comment (michalmarcinkowski) [inline features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature]: Repeating the address is not the best, but it's better to be direct than confusing. The scenarios should be specific examples without need to guess what "this address" means and figuring out that it comes from 5 steps above.

Comment (pjedrzejewski) [inline features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature]: Side notes:

1. "I complete order with X shipping method and Y payment ... is wrong, please replace it with proceed, because the order is not completed when I am on the summary page.
2. Offline is too technical, I think we agreed to use "Cash" payment as an example.

Comment (pjedrzejewski) [inline features/checkout/seeing_order_summary/seeing_addresses_on_order_summary_page.feature]: Another solution for different billing and shipping address is to keep the address in SharedStorage by the fullname of recipient and then do something like:

```
When I specify the shipping address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Jon Snow"
And I specify the billing address as "Ankh Morpork", "Frost Alley", "90210", "France" for "Eddard Stark"
And I complete the addressing step
And I proceed with "Free" shipping method and "Offline" payment
Then I should be on the checkout summary step
And my order's shipping address should be to "Jon Snow"
But my billing should be to "Eddard Stark"
```

We can only check the firstName and lastName on the order or like I said: store addresses per fullName in shared storage and hide the full comparision in the context.

Another side note: Step "I choose different billing address" should not be needed. The action of clicking the checkbox (which is a specific UI implementation) should be a part of the "I specify billing address" page implementation.

michalmarcinkowski: Agree with all.

"And this address should be used for both shipping and billing **of my order**"?

> ... store addresses per fullName in shared storage and hide the full comparison in the context. 

I would go with that approach.

pjedrzejewski:
Thanks Mateusz!

Issue #5261, opened 2016-06-15
[Resource] Do not check for permission by default


pjedrzejewski:
We should configure ``_sylius: permission:`` setting for all admin routes (and generate it for the ones, which are generated automatically) and make it ``false`` by default. It is really annoying to disable permissions for shop routes. See #5260.

inssein:
Iv'e got this most of the way (adjusted RequestConfiguration and the admin routes), but I cannot figure out how to get the generated routes to not check for permission. Any pointers?

pjedrzejewski:
Implemented in #5279. Community FTW!

PR #5279, opened 2016-06-17
[Resource] Do not check for permission by default

inssein:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/5261
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/routing/security.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/routing/security.yml
@@ -4,6 +4,8 @@ sylius_admin_login:
         _controller: sylius.controller.security:loginAction
         _sylius:
             template: SyliusAdminBundle:Security:login.html.twig
+            permission: true
+
 
 sylius_admin_login_check:
     path: /login-check
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/routing/security.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/routing/security.yml
@@ -6,7 +6,6 @@ sylius_admin_login:
             template: SyliusAdminBundle:Security:login.html.twig
             permission: true
 
-
 sylius_admin_login_check:
     path: /login-check
     defaults:
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/routing/taxon.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/routing/taxon.yml
@@ -5,15 +5,13 @@ sylius_admin_taxon:
         templates: SyliusAdminBundle:Taxon
         except: ['show', 'index']
         redirect: update
+        permission: true
         vars:
             all:
                 subheader: sylius.ui.manage_categorization_of_your_products
                 templates:
                     form: SyliusAdminBundle:Taxon:_form.html.twig
     type: sylius.resource
-    defaults:
-        _sylius:
-            permission: true
 
 sylius_admin_taxon_index:
     path: /taxons/
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/routing/taxon.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/routing/taxon.yml
@@ -11,6 +11,9 @@ sylius_admin_taxon:
                 templates:
                     form: SyliusAdminBundle:Taxon:_form.html.twig
     type: sylius.resource
+    defaults:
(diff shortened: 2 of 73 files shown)


inssein:
@pjedrzejewski I just realized that I missed the routes in WebBundle/backend. It seems like the whole WebBundle is going away before the next release, right?

Is there an updated roadmap anywhere?

src/Sylius/Bundle/AdminBundle/Resources/config/routing/security.yml
```diff
@@ -4,6 +4,8 @@ sylius_admin_login:
         _controller: sylius.controller.security:loginAction
         _sylius:
             template: SyliusAdminBundle:Security:login.html.twig
+            permission: true
+
```
michalmarcinkowski: Redundant line

src/Sylius/Bundle/AdminBundle/Resources/config/routing/taxon.yml
```diff
@@ -28,4 +31,6 @@ sylius_admin_taxon_tree:
             template: SyliusAdminBundle:Taxon:index.html.twig
             repository:
                 method: findRootNodes
+            permission: true
+
```
michalmarcinkowski: This line should not be added.

src/Sylius/Bundle/AdminBundle/Resources/config/routing/zone.yml
```diff
@@ -35,5 +38,7 @@ sylius_admin_zone_create:
                 route:
                     parameters:
                         type: $type
+            permission: true
+
```
michalmarcinkowski: Same as above.

src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
```diff
@@ -470,7 +470,7 @@ public function getEvent()
     public function hasPermission()
     {
         if (!$this->parameters->has('permission')) {
-            return true;
+            return false;
         }
 
         return false !== $this->parameters->get('permission');
```
michalmarcinkowski: `return true === $this->parameters->get('permission');` would be safer approach (e.g. `false !== 'false'` - will return true...).

michalmarcinkowski: The method `hasPermission` is confusing and actually it checks if permission is disabled. Maybe a better name for the method would be `isPermissionCheckDisabled`? /cc @pjedrzejewski

michalmarcinkowski: The method `hasPermission` is now used in a wrong context. See [RbacAuthorizationChecker](https://github.com/Sylius/Sylius/blob/147d0169768a91c08979cc6841cf5c1b7e9b5746/src/Sylius/Bundle/RbacBundle/Authorization/RbacAuthorizationChecker.php#L44). Taking into account current logic:

| $this->parameters->has('permission') | $this->parameters->get('permission')  | hasPermission |
| ------------- |------------- |------------
| false      | - | false |
| true      | false | false |
| true      | 'permission' | true |

When I do not set permission

lchrusciel: Why can't we define some default value as a second parameter? Won't it be equal?
```php
if (!$this->parameters->has('permission')) {
    return false;
}
 
return false !== $this->parameters->get('permission');
```
/\
||
\/
```php
return false !== $this->parameters->get('permission', false);
```

inssein:
@michalmarcinkowski @lchrusciel thanks for the code review - made the changes now.

inssein:
@michalmarcinkowski I only have two tests failing - I can't figure out which config that is using. Any pointers?

pjedrzejewski:
These are RBAC scenarios failing for old admin, because after this change all routes in WebBundle do not have any permission set... So we either set it to true on all routes or wait a bit with this merge until SyliusWebBundle is removed.

inssein:
@pjedrzejewski alright if I get some more time I will go ahead and add it to the old routes. Is there a roadmap anywhere? I'd love to be able to roughly plan. My sylius store is currently at 0.18, but working with master on a branch.

pjedrzejewski:
Thank you Hussein! 👍

PR #5292, opened 2016-06-19
[JS] Upgrade a npm package version

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

Proper version in package.json file

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,4 @@
 {
-  "name": "sylius",
-  "version": "0.17.0",
-  "description": "eCommerce for Developers",
-  "main": "Gulpfile.js",
   "dependencies": {
     "jquery": "^2.2.0",
     "semantic-ui-css": "^2.1.8"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "sylius",
-  "version": "0.17.0",
+  "version": "0.18.0",
```
pamil: Isn't it `0.19.0` on master right now? :)

pjedrzejewski:
I am pretty sure we should remove this and all other configurations which may suggest that this is a standalone package on NPM. We should only keep our requirements and that's it, wdyt?

package.json
lchrusciel: Right now we have `0.19.0-dev` so it is not a `0.19.0` directly

lchrusciel:
@pjedrzejewski it makes sense. I have just found a forgotten piece of code. 👍 for removal.

pjedrzejewski:
@lchrusciel Could you do it in this PR? Thanks!

lchrusciel:
sure, no problem

pjedrzejewski:
Thanks Łukasz!

PR #5293, opened 2016-06-19
[JS] Run gulp script

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

It is not always suitable to install a gulp globally. If it is not installed globally, the following command had to be called:
```bash
$ ./node_modules/.bin/gulp
```
This change allows to use more developer friendly command:
```bash
$ npm run gulp
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ If you want to try out new Sylius UI, please run the following commands:
 
 ```bash
 $ npm install
-$ gulp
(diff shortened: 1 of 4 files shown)


pjedrzejewski:
Thanks Łukasz!

PR #5300, opened 2016-06-20
[Core] Cleanup inventory and guest order paramaters

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets |
| License         | MIT

- [x] Removed invalid parameters - backorders and inventory tracking is now [configured in the bundle](https://github.com/michalmarcinkowski/Sylius/blob/386c226db60cc37c7e4cd0bf632caa95ccfad2d8/src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius.yml#L99) and there is no need for invalid parameters in the Core.
- [x] Removed guest checkout parameter from OrderBundle - the parameter from bundle isn't even registered in the container, so it is redundant. The `sylius.order.allow_guest_order` is now used in the `CheckoutStep`, so it can temporary stay, but will be removed together with the old checkout flow.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/parameters.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/parameters.yml
@@ -2,8 +2,6 @@
 # (c) Paweł Jędrzejewski
 
 parameters:
-    sylius.inventory.backorders_enabled: true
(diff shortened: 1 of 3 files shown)


pjedrzejewski:
Thank you Michał!

PR #5312, opened 2016-06-21
[Attrbiute] Use value setter & fix properties visibility

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes (uses setter method, changes visibility)
| Related tickets | -
| License         | MIT

`AttributeValue` is currently using getter to get the value and does not use setter to set the value. 

Moreover, it changes visibility of `get%type%()` methods from public to protected and their properties from protected to private (all modifications should be done through setter methods).


pjedrzejewski:
Thank you Kamil!

Issue #5001, opened 2016-05-11
[RFC] Remove "*.class" parameters


pamil:
We have two types of `*.class` parameters:
 - the ones from ResourceBundle: `sylius.model.resource.class` etc.
 - the ones from any other services

In this RFC I would like to address the second ones, reasoning behind removing these is similar to Symfony (citation from initial symfony/symfony#11881):

> All services in core define their classes as parameters. The convention is to create a foo.class parameter for the class of the foo service. This approach comes with several problems:

> The parameters are dumped like any other one in the compiled class and it makes the file larger for no good reasons (read: it slows down your app for free).

> When someone wants to change a service, just changing the class name is rarely enough: the new service probably have some different constructor arguments, ... Or put another way, overriding a service by just changing the class name of a service is a very rare use case.

> Of course, if several bundles change the value of such a parameter, the "last" definition wins. Classic inheritance vs composition problem.

> For all these reasons, I like to remove all those parameters for Symfony 3.0.

tristanbes:
👍 I do agree with you @pamil

pjedrzejewski:
👍  for services not related to the Resources!

peteward:
> Or put another way, overriding a service by just changing the class name of a service is a very rare use case.

Not that rare in our case Mr. Fabpot, it's actually very convenient:

```
    sylius.availability_checker.default.class: Reiss\Component\Inventory\AvailabilityChecker
    sylius.email_manager.order.class: Reiss\Bundle\StoreBundle\EmailManager\OrderEmailManager
    sylius.callback.order_payment.class: Reiss\Bundle\StoreBundle\StateMachine\Callback\OrderPaymentCallback
    sylius.checkout_scenario.class: Reiss\Bundle\StoreBundle\Checkout\CheckoutProcessScenario
    sylius.checkout_step.finalize.class: Reiss\Bundle\StoreBundle\Checkout\Step\FinalizeStep
    sylius.checkout_step.security.class: Reiss\Bundle\StoreBundle\Checkout\Step\SecurityStep
    sylius.context.currency.class: Reiss\Bundle\StoreBundle\Regionalization\Context\CurrencyContext
    sylius.controller.backend.dashboard.class: Reiss\Bundle\StoreBundle\Controller\DashboardController
    sylius.controller.backend.form.class: Reiss\WebBundle\Controller\Backend\FormController
    sylius.controller.checkout.class: Reiss\Bundle\StoreBundle\Controller\CheckoutController
[...]

PR #5339, opened 2016-06-23
[Theme] Ability to disable assets / templates / translations overwriting support

### LINKED ISSUE #5113: [ThemeBundle] Disable Translator
Hi,

Is there a way to disable anything related to translations in the ThemeBundle when the global symfony configuration indicates:

    framework:
        translator:
            enabled: false

Because for now it gives me this error:

    Warning: array_merge(): Argument #2 is not an array in vendor\sylius\theme-
    bundle\Translation\Provider\CompositeTranslatorResourceProvider.php line 40

Thanks!

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | fixes #5113, removes classes parameters as discussed in #5001
| License         | MIT


pjedrzejewski:
Nice! Although I am not sure about this ``support`` naming. We can and should do this for few more things in other bundles, so let's set some conventions. The most sensible term that comes to my mind now is "Feature Flags", but not perfect too... Any ideas, community? :)

BTW. Interesting bundle: https://github.com/DZunke/FeatureFlagsBundle.

pamil:
Yep, that isn't the best name for that, but it isn't a typical feature flag, we only need to load specific services file or not. I'm open for suggestions though.

pjedrzejewski:
Perfect, thanks!

PR #5351, opened 2016-06-24
[Core] Fix nth order rule checker for new guest customers

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

This is for our new checkout, when promotions are applied before new customer is flushed.


michalmarcinkowski:
Thanks! Will be fixed in #5363.

Issue #4060, opened 2016-02-04
[RFC] Order state machine


pjedrzejewski:
Hello!

Current states and state machine of Order is pretty confusing. This RFC aims to solve following problems:

* Clarifying all states that we have on the order;
* Making the checkout process cleaner;
* Making it easier to identify what is cart and what is order;

I will be talking about the following states and field:

* Order.state
* Order.checkoutState
* Order.shippingState
* Order.paymentState
* Order.completedAt

We currently use ``Order.completedAt`` to decide whether it is a Cart or Order. That's annoying. As suggested in another PR/issue, I think we should rename it to ``orderCheckoutCompletedAt`` and use it for information purposes when exactly the checkout has been completed.

## Checkout state

Now, to the states. Let's start from ``checkoutState``, which is pretty clear and reworked already in #3918. I will paste the configuration overview for reference:

```yaml
winzou_state_machine:
    sylius_order_checkout:
        class: %sylius.model.order.class%
        property_path: checkoutState
        graph: sylius_order_checkout
        state_machine_class: %sylius.state_machine.class%
        states:
            cart: ~
            started: ~
            addressed: ~
            shipping_selected: ~
            payment_selected: ~
            completed: ~
        transitions:
            start:
                from: [cart]
                to: started
            address:
                from: [started]
                to: addressed
            readdress:
                from: [payment_selected, shipping_selected, addressed]
                to: started
            select_shipping:
                from: [addressed]
                to: shipping_selected
            reselect_shipping:
                from: [payment_selected, shipping_selected]
                to: addressed
            select_payment:
                from: [shipping_selected]
                to: payment_selected
            reselect_payment:
                from: [payment_selected]
                to: shipping_selected
            complete:
                from: [payment_selected]
                to: completed
```

This is pretty much self-explanatory and covers the whole checkout process. If you customize it, you will need to rework this state-machine, but this gives us clean way of executing callbacks, recalculations, etc.

## State

Next state is ``Order.state``, this one is a bit tricky, but it should represent the overall state of the order and also we should use it to identify if Order is a cart. Here is the state machine I propose as initial implementation:

```yaml
winzou_state_machine:
    sylius_order:
        class: %sylius.model.order.class%
        property_path: state
        graph: sylius_order
        state_machine_class: %sylius.state_machine.class%
        states:
            cart: ~
            new: ~
            abandoned: ~
            cancelled: ~
            closed: ~
        transitions:
            create:
                from: [cart]
                to: new 
[...]

PR #5377, opened 2016-06-28
[Order] [WIP/RFC] Add state machine to order state resolver

### LINKED ISSUE #5366: [Order] StateResolver to use state machine?
The StateResolver changes the state of the order directly - this means that the state machine is not used, and no event is fired which could be used to notify something that, f.e., the payment has been completed.

Should the SM be used here? If not - what would be the best way to trigger some action when the order has been paid for?

dantleech:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | fixes #5366 
| License         | MIT

This PR adds the state machine to the order payment status, as decided by the StateResolver.

### DIFF
--- a/src/Sylius/Component/Core/OrderProcessing/StateResolver.php
+++ b/src/Sylius/Component/Core/OrderProcessing/StateResolver.php
@@ -15,46 +15,53 @@ use Sylius\Component\Core\Model\OrderInterface;
 use Sylius\Component\Core\Model\OrderShippingStates;
 use Sylius\Component\Core\Model\PaymentInterface;
 use Sylius\Component\Core\Model\ShipmentInterface;
+use SM\Factory\Factory;
+use Sylius\Component\Order\OrderTransitions;
+use Sylius\Component\Core\OrderCheckoutStates;
 
 /**
  * @author Paweł Jędrzejewski <pawel@sylius.org>
  */
 class StateResolver implements StateResolverInterface
 {
+    private $smFactory;
+
+    public function __construct(Factory $smFactory)
+    {
+        $this->smFactory = $smFactory;
+    }
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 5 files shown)


src/Sylius/Component/Core/OrderProcessing/StateResolver.php
```diff
@@ -15,46 +15,53 @@
...
+        // do not recalculate payment state. althoguh the payment may have been
+        // refunded, the payment WAS completed for this order.
+        if ($order->getPaymentState() === OrderInterface::STATE_PAYMENT_COMPLETED) {
+            return;
+        }
+
+        // do not calculate payment state when checkout has not been completed
+        if ($order->getCheckoutState() === OrderCheckoutStates::STATE_COMPLETED) {
+            return;
+        }
```
dantleech: The above two cases assume the COMPLETED is the final *payment state* of the order, which similifies things here, but is probably insufficent.

I guess we should take into account refunded, failed etc also.

```
if (all payments are refunded) state = REFUNDED
if (some payments are refunded) state = PARTIALLY_REFUNDED
if (some payments failed) state = PENDING
if (all paymens failed) state = FAILED
```

?

dantleech:
Lacking tests until we agree on this or a solution.

antonioperic:
ping @pjedrzejewski can you give some feedback on this?

dantleech:
I see that the payment transitions already exist in `CoreBundle\..\state_machine.yml` ..

michalmarcinkowski:
There is a pending discussion on state machine of an Order (https://github.com/Sylius/Sylius/issues/4060). Few things has changed and we will give it more thoughts soon, but any feedback is appreciated.

dantleech:
Closing this, assuming this is going to be handled elsewhere.

PR #5393, opened 2016-06-30
use code to show shipping method options

videni:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| License         | MIT

use code to show shipping method options, because id will change when we generate sample data each time , but  code is fixed ,it  is much easier for api.


pjedrzejewski:
Sounds good to me, although we should make it consistent for entire API if we decide to do it this way, wdyt? (Not asking to do it in this PR ofc)

michalmarcinkowski:
Thanks @videni!

PR #5394, opened 2016-06-30
[Resource] gracefully handle state machine transitions that cannot be applied

bendavies:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | 
| License         | MIT

At the moment, the `applyStateMachineTransitionAction` endpoint will throw a `500` Internal Server Error if a transition cannot be applied. This isn't very nice at all.

Propose returning a `400` Bad Request instead.


pjedrzejewski:
Very cool! Thanks Ben!

PR #5402, opened 2016-07-01
[Behat] Remove *.class parameters to improve readability

pamil:
Related to #5001


pjedrzejewski:
I agree in case of services and maybe contexts, but for pages it is quite useful to have them - You rarely need to change deps of the page, but you often want to provide your own implementation / extends from our page, just to change small thing to make it work with your layout - wdyt?

pamil:
Agreed, reverted those changes on pages.

pjedrzejewski:
Thanks Kamil!

Issue #5384, opened 2016-06-29
[RFC] Naming of subject-aware classes like Option, Attribute or Association


pamil:
Right now, we have:
 - `Sylius\Component\Product\Model\Archetype`
 - `Sylius\Component\Product\Model\Option`
 - `Sylius\Component\Product\Model\Variant`

but:
 - `Sylius\Component\Product\Model\ProductAssociation`

`Sylius\Component\Product\Model\Variant`, but `Sylius\Component\Product\Factory\ProductVariantFactory`.

`Sylius\Bundle\ProductBundle\Form\EventSubscriber\ProductOptionFieldSubscriber` but `Sylius\Bundle\ProductBundle\EventListener\ArchetypeUpdateListener`.

There are a lot inconsistencies for these kind of subject-aware classes and it's definitely the issue we must cope with before v1.0, because it takes way too much time to figure out what is the naming used in some cases.

In my opinion, we should prefix all subject-aware classes with their subject name, eg. there's `Sylius\Component\Variation\Model\Variant` that does not know anything about its subject, but `Sylius\Component\Product\Model\Variant` does, so it should be renamed to `Sylius\Component\Product\Model\ProductVariant`.

What do you think? :camel:

peteward:
yes definitely 👍

pamil:
Moreover, we have `Sylius\Component\Product\Model\OptionInterface` without any additional behaviour (marker interface), but `Sylius\Component\Product\Model\Product` does not enforce its options to implement it anyway. We end up using the one from Variation component and Product component interchangeably.

One more argument for prefixing is the situation, when we have two subjects with options and we want to overwrite them in CoreBundle - a conflict arises. We haven't encountered it yet, nevertheless there is general variant `Sylius\Component\Variation\Model\Variant`, product aware `Sylius\Component\Product\Model\Variant` and another one product aware `Sylius\Component\Core\Model\ProductVariant` with its name prefixed :tada:

pjedrzejewski:
👍

pamil:
We will probably need a list of models to change:
- [x] Product - Option (#6150)
- [x] Product - Variant (#6150)
- [ ] Promotion - Coupon (#6273)
- [ ] Promotion - Rule
- [ ] Promotion - Action

pamil:
Closing this one, extracted the rest of renaming changes as sub-issues.

PR #5406, opened 2016-07-01
[Core] Tweak the default fixtures to be more realistic and add orders

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

- [x] Add "brands" taxonomy
- [x] Initialize RBAC roles for admins


src/Sylius/Bundle/CoreBundle/Fixture/Factory/PaymentMethodExampleFactory.php
```diff
@@ -56,11 +56,14 @@ public function __construct(FactoryInterface $paymentMethodFactory, RepositoryIn
...
         $this->optionsResolver =
             (new OptionsResolver())
+                ->setDefault('code', function (Options $options) {
+                    return StringInflector::nameToCode($options['name']);
+                })
                 ->setDefault('name', function (Options $options) {
                     return $this->faker->words(3, true);
                 })
-                ->setDefault('code', function (Options $options) {
-                    return StringInflector::nameToCode($options['name']);
```
pamil: :-1: for moving this above name configuration, because we generate the code based on the name.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/PaymentMethodExampleFactory.php
```diff
@@ -87,7 +90,8 @@ public function create(array $options = [])
             $paymentMethod->setCurrentLocale($localeCode);
             $paymentMethod->setFallbackLocale($localeCode);
 
-            $paymentMethod->setName(sprintf('[%s] %s', $localeCode, $options['name']));
+            $paymentMethod->setName($options['name']);
```
pamil: This may create a lot of confusion if someone will change his locale and translatables will be the same. We can just omit printing locale code for default locale, wdyt?

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ShippingMethodExampleFactory.php
```diff
@@ -79,11 +82,11 @@ public function __construct(
                 ->setDefault('zone', LazyOption::randomOne($zoneRepository))
                 ->setAllowedTypes('zone', ['null', 'string', ZoneInterface::class])
                 ->setNormalizer('zone', LazyOption::findOneBy($zoneRepository, 'code'))
-                ->setDefault('shipping_category', LazyOption::randomOne($shippingCategoryRepository))
-                ->setAllowedTypes('shipping_category', ['null', 'string', ShippingCategoryInterface::class])
-                ->setNormalizer('shipping_category', LazyOption::findOneBy($shippingCategoryRepository, 'code'))
+                ->setDefined('category')
```
pamil: It's something related to #5384, ShippingMethod has both `taxCategory` and `category` which can be confusing, that's why I decided to use `shipping_category` as name for this option.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/TaxonExampleFactory.php
```diff
@@ -73,14 +79,13 @@ public function __construct(
                 ->setDefault('description', function (Options $options) {
                     return $this->faker->paragraph;
                 })
-                ->setDefault('parent', LazyOption::randomOneOrNull($taxonRepository, 70))
-                ->setAllowedTypes('parent', ['null', 'string', TaxonInterface::class])
-                ->setNormalizer('parent', function (Options $options, $previousValue) use ($taxonManager) {
+                ->setDefault('children', [])
+                ->setAllowedTypes('children', ['array'])
+                ->setNormalizer('children', function (Options $options, $previousValue) use ($taxonManager) {
                     $taxonManager->flush();
```
pamil: Without usage of another normalizer, this can be put somewhere inside `loadResource` method to make options resolver simpler.

pamil: Anyway, with `parent` replaced by `children`, do we even need it?

src/Sylius/Bundle/CoreBundle/Fixture/TshirtProductFixture.php
```diff
@@ -155,21 +162,23 @@ public function load(array $options)
...
                 'name' => sprintf('T-Shirt "%s"', $this->faker->word),
                 'code' => $this->faker->uuid,
-                'main_taxon' => 'TSHIRTS',
+                'main_taxon' => $categoryTaxonCode,
                 'product_archetype' => 'TSHIRT',
-                'taxons' => ['TSHIRTS', 'SUPER-TEES'],
+                'taxons' => [$categoryTaxonCode],
                 'product_attributes' => [
-                    'TSHIRT-BRAND' => $this->faker->randomElement(['Nike', 'Adidas', 'JKM-476 Streetwear', 'Potato', 'Centipede Wear']),
+                    'TSHIRT-brand' => $this->faker->randomElement(['Nike', 'Adidas', 'JKM-476 Streetwear', 'Potato', 'Centipede Wear']),
```
pamil: It looks weird to mix uppercase with lowercase, let's unify it.

src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
```diff
@@ -12,111 +12,108 @@ sylius_fixtures:
                 locale:
                     options:
                         locales:
-                            en_US: ~
-                            en_GB: ~
-                            es_ES: ~
-                            pl_PL: ~
+                            %locale%: ~
```
pamil: We don't even need the configuration here:

```yaml
fixtures:
  locale: ~
```

will do the same

src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
```diff
@@ -12,111 +12,108 @@ sylius_fixtures:
...
 
                 currency:
                     options:
                         currencies:
-                            EUR: ~
-                            USD: ~
-                            GBP: ~
-                            CHF: ~
+                            %currency%: ~
```
pamil: Same as for locale fixture configuration.

src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
```diff
@@ -12,111 +12,108 @@ sylius_fixtures:
...
+                                name: FedEx
 
                 channel:
                     options:
-                        random: 3
+                        custom:
+                            us_web_store:
+                                name: US Web Store
+                                locales:
+                                    - %locale%
```
pamil: Newer versions of Symfony YAML needs this to be surrounded with quotation marks - `- "%locale%"`

src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
```diff
@@ -12,111 +12,108 @@ sylius_fixtures:
...
+                                currencies:
+                                    - %currency%
```
pamil: Same here

src/Sylius/Bundle/FixturesBundle/Command/FixturesLoadCommand.php
```diff
@@ -31,7 +31,7 @@ protected function configure()
         $this
             ->setName('sylius:fixtures:load')
             ->setDescription('Loads fixtures from given suite')
-            ->addArgument('suite', InputArgument::REQUIRED, 'Suite name')
+            ->addArgument('suite', InputArgument::OPTIONAL, 'Suite name', 'default')
```
pamil: I'm not sure, fixtures bundle is not aware of suites configuration, but we may actually recommend `default` suite name to end users.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/PaymentMethodExampleFactory.php
pjedrzejewski: Sounds like an idea, but I'd like to remove it now and bring back as soon as we start working on i18n scenarios for new shop.

pjedrzejewski:
Applied fixes after review, added order fixtures (checking out programmatically, yay) + RBAC. Should be good for final review.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
```diff
@@ -218,15 +211,16 @@ public function create(array $options = [])
         $product->setEnabled($options['enabled']);
         $product->setMainTaxon($options['main_taxon']);
         $product->setArchetype($options['product_archetype']);
-        $product->setShippingCategory($options['shipping_category']);
```
pamil: Why?

src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php
```diff
@@ -0,0 +1,269 @@
...
+     */
+    private $stateMachineFactory;
+
+    /**
+     * @var \Faker\Generator
+     */
+    private $faker;
+
+    /**
+     * OrderFixture constructor.
```
tuka217: Should be removed.

pamil: :(

src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php
```diff
@@ -0,0 +1,269 @@
...
+            $countryCode = $this->faker->randomElement($countries)->getCode();
+
+            $currencyCode = $this->faker->randomElement($channel->getCurrencies()->toArray())->getCode();
+            $localeCode = $this->faker->randomElement($channel->getLocales()->toArray())->getCode();
+
+            $order = $this->orderFactory->createNew();
+            $order->setChannel($channel);
+            $order->setCustomer($customer);
+            $order->setCurrencyCode($currencyCode);
+            // $order->setLocaleCode($localeCode);
```
tuka217: Same here.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/UserExampleFactory.php
```diff
@@ -36,6 +36,11 @@
     private $customerFactory;
 
     /**
+     * @var RepositoryInterface
```
lchrusciel: It is not a part of this PR, but just a short thought about Sylius repository interfaces in general. WDYT about creating a repository interface for each class, that would at least extend a general RepositoryInterface? Even in this class constructor depends on two different repositories with the same typehint. It is super easy to change an order of them by an accident. If we would create these interfaces/markers we would received a message, that something is not properly injected much faster. 

I will probably create a separate RFC for that.

src/Sylius/Bundle/CoreBundle/Fixture/TshirtProductFixture.php
```diff
@@ -108,19 +108,26 @@ public function load(array $options)
...
+                    'name' => 'T-Shirts',
+                    'children' => [
+                        [
+                            'code' => 'mens_t_shirts',
+                            'name' => 'Men',
+                        ],
+                        [
+                            'code' => 'womens_t_shirts',
+                            'name' => 'Women',
+                        ]
```
lchrusciel: Missing comma at the end of line here and in the below lines.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
pjedrzejewski: For now, as we do not support shipping categories and I am not sure yet if we will - bringing this back when we properly support them will be easy.

pjedrzejewski:
Fixed, should be good to go in the morning! :)

src/Sylius/Bundle/CoreBundle/Fixture/TshirtProductFixture.php
```diff
@@ -155,21 +162,23 @@ public function load(array $options)
             [
                 'name' => 'T-Shirt',
                 'code' => 'TSHIRT',
-                'product_attributes' => ['TSHIRT-BRAND', 'TSHIRT-COLLECTION', 'TSHIRT-MATERIAL'],
+                'product_attributes' => ['TSHIRT-brand', 'TSHIRT-COLLECTION', 'TSHIRT-MATERIAL'],
```
michalmarcinkowski: Why `brand` is the only lowercased word?

src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
```diff
@@ -9,119 +9,116 @@ sylius_fixtures:
...
                 tax_rate:
                     options:
-                        random: 10
-
-                taxon:
-                    options:
-                        random: 50
+                        custom:
+                            clothing_tax:
+                                name: Clothing Sales Tax 7%
```
michalmarcinkowski: It will be strange to have tax rate named `Clothing Sales Tax 7%` with random rate from 0 to 40%. Maybe simply remove `7%` from name?

michalmarcinkowski:
Nice! Thank you Paweł!

Let's apply remaining comments in a separate PR.

PR #5409, opened 2016-07-01
[Admin] Initial support for static content

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | based on #5417
| License         | MIT


src/Sylius/Behat/Context/Ui/Admin/ManagingStaticContentsContext.php
```diff
@@ -0,0 +1,214 @@
...
+            (int) $amount,
+            $this->indexPage->countItems(),
+            'Amount of currencies should be equal %s, but was %2$s.'
+        );
+    }
+
+    /**
+     * @Then the static content :title should not be added
+     */
+    public function theCurrencyShouldNotBeAdded($title)
```
pjedrzejewski: Currency? :)

features/cmf/managing_static_contents/deleting_static_content.feature
```diff
@@ -0,0 +1,15 @@
...
+
+    Background:
+        Given I am logged in as an administrator
+
+    @ui
+    Scenario: Deleted taxon should disappear from the registry
+        Given the store has static content "Krzysztof Krawczyk"
+        When I delete static content "Krzysztof Krawczyk"
+        Then I should be notified that it has been successfully deleted
+        And the static content "Krzysztof Krawczyk" should no longer exist in the registry
```
pjedrzejewski: ``in the store`` sounds better than registry, wdyt?

src/Sylius/Behat/Page/Admin/StaticContent/UpdatePage.php
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function getDefinedElements()
+    {
+        return array_merge(parent::getDefinedElements(), [
+            'body' => '#sylius_static_content_body',
+        ]);
+    }
+
```
pjedrzejewski: Extra blank line.

pjedrzejewski:
Thanks Kamil!

src/Sylius/Behat/Context/Ui/Admin/ManagingStaticContentsContext.php
```diff
@@ -0,0 +1,214 @@
...
+
+use Behat\Behat\Context\Context;
+use Sylius\Behat\Page\Admin\Crud\IndexPageInterface;
+use Sylius\Behat\Page\Admin\StaticContent\CreatePageInterface;
+use Sylius\Behat\Page\Admin\StaticContent\UpdatePageInterface;
+use Sylius\Bundle\ContentBundle\Document\StaticContent;
+use Webmozart\Assert\Assert;
+
+/**
+ * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
```
lchrusciel: nice :)

src/Sylius/Behat/Resources/config/services/contexts/setup.xml
```diff
@@ -146,6 +146,13 @@
             <tag name="sylius.behat.context" />
         </service>
 
+        <service id="sylius.behat.context.setup.static_content" class="Sylius\Behat\Context\Setup\StaticContentContext" scope="scenario">
```
lchrusciel: Shouldn't we consider put it in another file? For example split setup to ORM and ODM setup? I think its not a big deal, but still can be helpful if we would like to give an option to toggle phpcr.

lchrusciel:
Won't it be worthy to split this implementation somehow from a default admin? I can imagine many cases, where somebody would like to take our admin, but manage a static content different way.

src/Sylius/Behat/Context/Ui/Admin/ManagingStaticContentsContext.php
pamil: Nothing to be afraid of, that's a piece of really good code! :D

pamil:
I guess we can port the features from the old admin panel to the now one first and then think about making it an optional feature, there's much more to do in order to achieve that.

Issue #5275, opened 2016-06-16
[Documentation] Sylius 1.0 - The Documentation Plan


CoderMaggie:
# Hello Folks! 👋 👌 🙌

As we are currently approaching the final Sylius 1.0 BETA we have started thinking about upgrading our documentation. You've probably noticed a few emerging PRs to the Docs ( #5174, #5195, #5214 ) which are just a beginning of a whole big revolution.

We would love to **_develop understandable, well-organized, comprehensive documentation_**. To achieve that we will be needing **strong assistance of the Sylius community**. Every single PR with the _[Documentation]_ tag will be appreciated a lot. :)
# The Documentation Plan 📖

Together with @pjedrzejewski and @michalmarcinkowski we have sketched a detailed (but still initial) plan for the docs development. 
The docs are divided into chapters and subchapters. Those which are for now the most important ones are marked with a 💥 . 
We are presenting them as a checklist so that everyone will be able to see what has been already done, with a link to relevant PRs.
### The Installation: 💥
- [x] Requirements for running Sylius ( #5364 )
- [x] Installing Sylius via Composer ( #5174 )
### System Architecture:
- [x] Fullstack Symfony ( #5420 )
- [x] Doctrine & Twig ( #5420 )
- [x] Division into Components, Bundles, Platform ( #5451 )
- [x] Division into Core, Admin, Shop, Api ( #5451 )
- [x] Resource Layer 💥  ( #5489 )
- [x] State Machine 💥  ( #5489 )
- [x] Third Party Libraries: Payum, etc. ( #5500 )
### Detailed concepts of Sylius:
- [x] **General**: 💥 
  Channels ( #5511 )
  Currencies ( #5527 )
  Locales ( #5518 )
- [X] **Products, Variants and Options** ( #5581 )
- [X] **Taxons** ( #5630 )
- [X] **Attributes** ( #5608 )
- [X] **Order** ( #5705 )
- [x] Pricing
- [X] Payments
- [X] Shipments
- [X] **Checkout**: ( #5730 )
  Addressing
  Shipping
  Payment
  Summary
- [X] **Addresses**: ( #5920 )
  Countries
  Zones
- [X] **Inventory**: ( #6460 )
- [X] **Taxation**: ( #5962 )
  Categories
  Rates
  Zones
- [X] **Payments**: ( #5816 )
  Methods
  State Machine
- [X] **Shipments**: ( #5816 )
  Methods
  Categories
  Zones
  State Machine
[...]

PR #5414, opened 2016-07-04
[Documentation] Installation via composer

TheMadeleine:
| Q               | A
| --------------- | ---
| Doc fix?        | yes
| New docs?    | no
| BC breaks?      | no
| License         | MIT


docs/contributing/index.rst
```diff
@@ -5,6 +5,55 @@ Contributing
...
+    currency (USD):
+    wkhtmltopdf.bin_path (/usr/bin/wkhtmltopdf):
+    wkhtmltoimage.bin_path (/usr/bin/wkhtmltoimage):
+
+After everything is in place, run the following commands:
+
+.. code-block:: bash
+
+    $ cd sylius # Move to the newly created directory
+    $ php app/console sylius:install
```
lchrusciel: You need also a gulp call to see the shop. We should add a note that the following lines are require to present basic shop: 
```bash
$ npm install
$ npm run gulp
```

BTW should we think about adding it to `sylius:install` command?

docs/book/installation.rst
```diff
@@ -87,6 +38,12 @@ place, run the following commands:
 
 This package has the whole ``sylius/sylius`` package in vendors, so you can easily update it and focus on your custom development.
 
+.. warning::
+
+    During the ``sylius:install`` command you will be asked to provide important information, but also its execution ensures
+    that the default **currency** (USD) and the default **locale** (English - US) are set.
```
pjedrzejewski: This really depend on your parameters.yml configuration, so I'd avoid giving USD/en_US examples here.

pjedrzejewski: OK, the next sentence clarifies it a bit, so let's just add a note that these defaults are configurable in ``parameters.yml``.

docs/contributing/index.rst
pjedrzejewski: +1 for adding these commands, but not sure about including them in sylius:install - we will see. For now let's keep them separate.

michalmarcinkowski: Maybe `npm install` could be run on composer `post-install`?

pjedrzejewski: The problem here is that if someone does not have npm installed, it will break the script.

TheMadeleine: Added below :)

pjedrzejewski:
Thanks Magda!

PR #5274, opened 2016-06-16
[Admin][Shop] New order appears in the admin panel

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | https://github.com/Sylius/Sylius/issues/4060
| License         | MIT

PR #5416, opened 2016-07-04
[Core] Fix Order's state machine

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | #5274
| License         | MIT

After changes in https://github.com/Sylius/Sylius/pull/5274 the inventory is not updated after payment is completed. This PR fixes that and temporary removes "pending" state as it is currently not used anymore. The whole state machine will be reworked soon (see https://github.com/Sylius/Sylius/issues/4060).


pjedrzejewski:
Thanks Michał!

PR #5417, opened 2016-07-04
[Behat] Refactor checkValidationMessageFor to getValidationMessage

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | -
| License         | MIT

There are assertions like:

```php
Assert::true(
    $this->updatePage->checkValidationMessageFor('code', 'This code is already in use.'),
    'Unique code violation message should appear on page, but it does not.'
);
```

We are getting "Unique code violation message should appear on page, but it does not." when the validation message differs from "This code is already in use.". If the validation message gets changed, we will get an exception, that there is no unique code violation on the page, but there will be.

I've changed it to:

```php
Assert::same($this->updatePage->getValidationMessage('code'), 'This code is already in use.');
```

It will throw an exception with the actual message and the asserted message. If there is no validation message found, `ElementNotFoundException` will be thrown inside `getValidationMessage`.

### DIFF
--- a/src/Sylius/Behat/Context/Ui/Admin/ManagingChannelsContext.php
+++ b/src/Sylius/Behat/Context/Ui/Admin/ManagingChannelsContext.php
@@ -178,12 +178,10 @@ final class ManagingChannelsContext implements Context
      */
     public function iShouldBeNotifiedThatAtLeastOneChannelHasToBeDefinedIsRequired()
     {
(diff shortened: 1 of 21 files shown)


pjedrzejewski:
Thanks Kamil! One thing to consider for the future is ... more than 1 message per field. ;)

PR #5422, opened 2016-07-04
[Fixtures] Ability to reuse a fixture in the same suite

### LINKED ISSUE #5401: [FixturesBundle] Allow to use the same fixtures multiple times in one suite
It will allow for the following scenario:
 - add 5 random, enabled admin users
 - add 50 random, 90% enabled / 10% disabled regular users

```yaml
sylius_fixtures:
    suites:
        suite:
            fixtures:
                # Add 5 random, enabled admin users 
                - user:
                    random: 5
                    prototype:
                        admin: true
                        enabled: true

                # Add 50 random, 90% enabled / 10% disabled regular users
                - user:
                    random: 50
                    prototype:
                        admin: false
```

We need to replace configuration from `'fixtures' => ['user' => ...]` with `'fixtures' => [['user' => ...], ['user' => ...]]` and enhance suite factory.

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | fixes #5401
| License         | MIT

TODO:
 - [x] Document the new feature (#5434)

TODO after merge:
 - Create an issue about logging executed suites (they should show fixture aliases, not the actual fixture names)

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
@@ -69,9 +69,14 @@ sylius_fixtures:
                     options:
                         random: 3
 
-                user:
+                regular_user:
+                    name: user
                     options:
                         random: 20
+
+                admin_user:
+                    name: user
+                    options:
                         custom:
                             -
                                 email: sylius@example.com
--- a/src/Sylius/Bundle/FixturesBundle/DependencyInjection/Configuration.php
+++ b/src/Sylius/Bundle/FixturesBundle/DependencyInjection/Configuration.php
@@ -46,6 +46,24 @@ final class Configuration implements ConfigurationInterface
                     ->prototype('array')
         ;
 
+        $suitesNode
+            ->validate()
+                ->ifArray()
+                ->then(function (array $value) {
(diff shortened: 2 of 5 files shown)


src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
```diff
@@ -69,9 +69,14 @@ sylius_fixtures:
                     options:
                         random: 3
 
-                user:
+                regular_user:
+                    name: user
```
pjedrzejewski: Maybe ``type`` or alias?

pamil: I come up with the following naming:
 - `regular_user` is an alias
 - `user` is a fixture name

It is based on `FixtureInterface::getName()` method (and similar to `ListenerInterface::getName()`).

I'm not sure for `type` too, as IMO it assumes that there is some closed list of available choices.

pjedrzejewski: Makes sense!

pjedrzejewski:
Thanks Kamil! 👍

PR #5429, opened 2016-07-05
Only add steps that are active

dantleech:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no/hope not
| Related tickets | 
| License         | MIT

Currently the `isActive` method of the `StepInterface` is not used.

We have a situation where we would like to "remove" a checkout step based on a certain condition.

I assume that this is what the `isActive()` method was intended for, it is here added to the `ProcessBuilder`.

### DIFF
--- a/src/Sylius/Bundle/FlowBundle/Process/Builder/ProcessBuilder.php
+++ b/src/Sylius/Bundle/FlowBundle/Process/Builder/ProcessBuilder.php
@@ -79,6 +79,10 @@ class ProcessBuilder implements ProcessBuilderInterface
             throw new \InvalidArgumentException('Step added via builder must implement "Sylius\Bundle\FlowBundle\Process\Step\StepInterface"');
         }
 
+        if (!$step->isActive()) {
+            return $this;
+        }
+
         if ($step instanceof ContainerAwareInterface) {
             $step->setContainer($this->container);
         }
--- a/src/Sylius/Bundle/FlowBundle/spec/Process/Builder/ProcessBuilderSpec.php
+++ b/src/Sylius/Bundle/FlowBundle/spec/Process/Builder/ProcessBuilderSpec.php
@@ -14,6 +14,9 @@ namespace spec\Sylius\Bundle\FlowBundle\Process\Builder;
 use PhpSpec\ObjectBehavior;
 use Sylius\Bundle\FlowBundle\Process\Builder\ProcessBuilderInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
+use Sylius\Bundle\FlowBundle\Process\Step\StepInterface;
+use Sylius\Bundle\FlowBundle\Process\Scenario\ProcessScenarioInterface;
+use Prophecy\Argument;
 
 class ProcessBuilderSpec extends ObjectBehavior
 {
@@ -30,5 +33,28 @@ class ProcessBuilderSpec extends ObjectBehavior
     function it_is_process_builder()
     {
         $this->shouldImplement(ProcessBuilderInterface::class);
+    }
+
+    function it_should_not_add_inactive_steps(
+        StepInterface $step,
+        ProcessScenarioInterface $scenario
+    ) {
+        $this->build($scenario);
+        $step->isActive()->willReturn(false);
+        $step->setName(Argument::any())->shouldNotBeCalled();
+
+        $this->add('foobar', $step);
+    }
+
+    function it_should_add_active_steps(
+        StepInterface $step,
+        ProcessScenarioInterface $scenario
+    ) {
+        $step->getName()->willReturn(null);
+        $this->build($scenario);
+        $step->isActive()->willReturn(true);
+        $step->setName('foobar')->shouldBeCalled();
+
+        $this->add('foobar', $step);
     }
 }


src/Sylius/Bundle/FlowBundle/Process/Builder/ProcessBuilder.php
```diff
@@ -79,6 +79,10 @@ public function add($name, $step)
             throw new \InvalidArgumentException('Step added via builder must implement "Sylius\Bundle\FlowBundle\Process\Step\StepInterface"');
         }
 
+        if (false === $step->isActive()) {
```
Zales0123: ``if (!$step->isActive())``

src/Sylius/Bundle/FlowBundle/spec/Process/Builder/ProcessBuilderSpec.php
```diff
@@ -31,4 +34,27 @@ function it_is_process_builder()
     {
         $this->shouldImplement(ProcessBuilderInterface::class);
     }
+
+    function it_should_not_add_inactive_steps(
+        StepInterface $step,
+        ProcessScenarioInterface $scenario
+    )
+    {
```
Zales0123: CS

src/Sylius/Bundle/FlowBundle/spec/Process/Builder/ProcessBuilderSpec.php
```diff
@@ -31,4 +34,27 @@ function it_is_process_builder()
...
+        $this->build($scenario);
+        $step->isActive()->willReturn(false);
+        $step->setName(Argument::any())->shouldNotBeCalled();
+        $this->add('foobar', $step);
```
Zales0123: I would add blank line before, to improve readability (as well as in next spec).

src/Sylius/Bundle/FlowBundle/Process/Builder/ProcessBuilder.php
dantleech: Is it OK for `isActive()` to return an empty string, a zero or otherwise a non-boolean value?

Zales0123: Nope, ``isActive()`` returns only bool value.

pamil: `is*()` should always return `bool` value, if it isn't - that's the issue :)

Zales0123: If think, that now both ``isActive()`` implementations ([this](https://github.com/Sylius/Sylius/blob/f5d1b5f7fe845adfcc2fac485d6e19cd1708e03b/src/Sylius/Bundle/FlowBundle/Process/Step/AbstractStep.php#L59) and [this](https://github.com/Sylius/Sylius/blob/f5d1b5f7fe845adfcc2fac485d6e19cd1708e03b/src/Sylius/Bundle/FlowBundle/Process/Step/AbstractControllerStep.php#L60)) returns hardcoded ``true`` 😄

dantleech: I think should = must. If it is returning anything else then it is behaving badly. I really prefer the strict check.

dantleech: > If think, that now both isActive() implementations

Yes they do :) We just implemented a custom step which is conditional.

dantleech:
@pjedrzejewski what do you think? is this how you intended the `isActive()` method to be used?

src/Sylius/Bundle/FlowBundle/Process/Builder/ProcessBuilder.php
dantleech: Does the strict check make sense to you? Or am I just being a CS violator?

pamil: It's not a big deal here, we got a contract that `isActive()` returns `bool` so either `false === isActive()` or `!isActive()` are all right and we're just wasting our time on this discussion which depends on personal preferences.

IMO I'd go for `!isActive()` just for one reason - to make a distinction between functions returning mixed values (like `strpos()` - `bool|int`), where `!strpos()` will be `true` if `strpos() === 0 ` and those which are restricted to return `bool` only.

Zales0123: Well, for me as ``isActive()`` **must** return ``bool``, there is no need for strict check. If it returns something else (even if it can be casted to ``bool``, as empty string or ``0``) it is a problem of bad code implementation, however not here, but in ``isActive()`` method. But it's only my opinion, I'll be happy to read others minds ;)

pamil: Wow, that's quite a nice shitstorm over a bool comparison :tada: Any other programming language would not allow us to even think about it, but PHP is not one of them ❤️

dantleech: ok, in Symfony CMF we do `=== false` generally. It makes more sense to me as it means that we implicitly do not support the subject returning non-boolean falsy values. But I really don't feel strongly about it - and often use the negator myself, its just a matter of setting a precedent for future contributions :) Will change.

dantleech:
Updated.

pjedrzejewski:
Makes sense, thanks Dan!

PR #5415, opened 2016-07-04
[Order][Checkout] Thx page

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

PR #5442, opened 2016-07-06
[Core][Payment] Populate gateway choices based on payum configuration

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/DependencyInjection/SyliusPayumExtension.php
+++ b/src/Sylius/Bundle/PayumBundle/DependencyInjection/SyliusPayumExtension.php
@@ -14,6 +14,7 @@ namespace Sylius\Bundle\PayumBundle\DependencyInjection;
 use Sylius\Bundle\ResourceBundle\DependencyInjection\Extension\AbstractResourceExtension;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 
 /**
@@ -21,7 +22,7 @@ use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
  *
  * @author Maksim Kotlyar
  */
-class SyliusPayumExtension extends AbstractResourceExtension
+class SyliusPayumExtension extends AbstractResourceExtension implements PrependExtensionInterface
 {
     /**
      * {@inheritdoc}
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/PaymentBundle/Form/Type/PaymentGatewayChoiceType.php
```diff
@@ -11,31 +11,27 @@
...
+    protected $gatewaysRegistry;
 
     /**
-     * Constructor.
-     *
-     * @param array $gateways
+     * @param RegistryInterface $gatewaysRegistry
      */
-    public function __construct(array $gateways)
+    public function __construct(RegistryInterface $gatewaysRegistry)
```
makasim: Would be better to use GatewayRegistryInterface https://github.com/Payum/Payum/blob/master/src/Payum/Core/Registry/GatewayRegistryInterface.php

makasim:
Pay attention  to not all payum gateway could be used out of the box, some of them may require a special converter action.

src/Sylius/Component/Payment/composer.json
```diff
@@ -23,7 +23,8 @@
         "php": "^5.6|^7.0",
 
         "sylius/registry": "^0.19",
-        "sylius/resource": "^0.19"
+        "sylius/resource": "^0.19",
+        "payum/payum": "~1.0",
```
makasim: payum/core would be enough

src/Sylius/Bundle/PaymentBundle/spec/Form/Type/PaymentGatewayChoiceTypeSpec.php
```diff
@@ -39,8 +39,8 @@ function it_is_a_form_type()
 
     function it_has_options(
         GatewayInterface $offlineGateway,
-        OptionsResolver $resolver,
-        RegistryInterface $gatewaysRegistry
+        GatewayRegistryInterface $gatewaysRegistry,
```
pamil: As it's the collaborator injected into constructor it should be the first one here.

makasim:
Not sure sylius payment component must have a dependency on payum/core. So maybe we can introduce a new form type, payum specific. one can use it if both sylius/payment and payum/core are installed. The sylius/sylius could use payum specific by default

It would be also better from BC  point of view.

@pjedrzejewski ping

pjedrzejewski:
Definitely should not depend on Payum/Core. So I'd introduce a gateway choice type, that is replaced/populated with appropriate choices in PayumBundle. Ideally, entire glue between Sylius and Payum is in SyliusPayumBundle. And it should not be that hard, because only problem now is in the new Pay / After Pay actions on OrderController in Core. See #5415. Moving that to PayumBundle down the road should not be a problem.

makasim:
Could be in help https://github.com/Payum/Payum/blob/master/src/Payum/Core/Bridge/Symfony/Form/Type/GatewayFactoriesChoiceType.php

Zales0123:
Ok, so finally we end up with something much simpler 😄  We remove gateways configuration from ``sylius_payment`` and prepend ``payum`` gateways with extension. Form type is left as it was, as well as parameter injected in it. @pjedrzejewski @makasim what say you? ;)

src/Sylius/Bundle/PayumBundle/DependencyInjection/SyliusPayumExtension.php
```diff
@@ -44,4 +45,23 @@ public function load(array $config, ContainerBuilder $container)
...
+    /**
+     * {@inheritdoc}
+     */
+    public function prepend(ContainerBuilder $container)
+    {
+        if (!$container->hasExtension('sylius_payment')) {
+            return;
+        }
+
+        $config = $container->getExtensionConfig('payum');
```
makasim: this is not really accurate since some gateways could be added via container tag, or even stored in database.

Zales0123: Sure they can. However this is just a prepend, gateways configuration can be easily overwritten in ``yml`` file, or in extension, or in any other possible way.

pjedrzejewski:
Thanks Mateusz! 👍

PR #5459, opened 2016-07-07
Remove right trim in order list

ping86:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets |
| License         | MIT

When you define a channel color in the backend order list we need add a right space:

```
<spanstyle="color: Green;">default</span>
```


michalmarcinkowski:
👍

Issue #5227, opened 2016-06-10
[Admin] 403 Access denied after updating admin user


michalmarcinkowski:
The following scenario describes the bug:
```
Given I am logged in as an administrator
When I want to edit this user
And I deselect "Customer can login to the store"
But then select it again
And I save my changes
Then I get 403 access denied exception
And I am not an administrator any more :(
```

Arminek:
I have reproduced this bug, but in current master version i didn't get the access denied exception.

michalmarcinkowski:
The following scenario still occurs:
```
Given I am logged in as an administrator
When I change my password to "abc"
And I logout
And I login to the admin panel with password "abc"
Then I get 403 access denied exception
And I am not an administrator any more :(
```

Sicaine:
Same here. It happens as long as you save the user. Doesn't matter what you change.

bwach:
I couldn't reproduce it in current master

michalmarcinkowski:
Fixed in #5466

PR #5466, opened 2016-07-08
[Admin][User] Remove temporarily authorization roles from user form

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | https://github.com/Sylius/Sylius/issues/5227
| License         | MIT

I have temporarily removed authorization roles from user form because of a problem with updating an administrator in admin panel.


features/user/managing_customers/editing_administrator.feature
```diff
@@ -0,0 +1,20 @@
...
+    In order to change information about an administrator
+    As an Administrator
+    I want to be able to edit my information
+
+    Background:
+        Given the store operates on a single channel in "France"
+        And there is an administrator "Gareth Bale" identified by an email "bale@example.com" and a password "bale11"
+        And I am logged in as "bale@example.com"
+
+    @ui @javascript
```
pamil: Why do we need javascript here?

GSadee: Because in admin panel we have dropdown menu which is a little problem, when I want to press logout button.

features/user/managing_customers/editing_administrator.feature
```diff
@@ -0,0 +1,20 @@
...
+    Scenario: Changing password and signing again
+        Given I want to edit the customer of my account
```
lchrusciel: `the customer of my account` seems a little bit weird. Why didn't you use `I want to change my password` or something  similar?

michalmarcinkowski:
Thanks Grzesiu! 👍

PR #5478, opened 2016-07-10
Generic controller events

### LINKED ISSUE #5456: [RFC] [Resource] Dispatch general events, not specific events.
Currently event names use runtime data such as "application name" and "metadata name".

For example:

```
sylius_product_pre_create
sylius_foobar_pre_create
```

This has (at least) three distinct disadvantages:
- We cannot define constants to refer to these events (unless in userland code). Any third-party event subscriber is at risk of the source event name being changed.
- Event names cannot be easily discovered.
- Very hard to make a subscriber that listens to ALL f.e. pre-create events from the resource controller[1]
  
  I would suggest simplifying the process, removing the resource-specific EventDispatcher and issuing events:

```
pre_create
post_create
show
... etc
```

This shifts the burden of responsibility for acting on events to the event listeners themselves (you could also create an event listener which emulates the current behavior).

The only argument I can currently see for this approach is a performance one, but performance impact would be imperceptible, and in anycase, this is not a performance critical path - and this same strategy is used by the Doctrine ORM where performance cost is much more valuable.

[1] My use case is to automatically set the parent on a PHPCR document from the request.

dantleech:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Related tickets | fixes #5456 |

This PR explores dispatching general, constantly named events, from the resource controller instead of events named after the type of resource, application name etc. for reasons explained in the related issue.

Backwards compatiblity is here maintained through a new event subscriber which dispatches the resource-application specific events. From my POV we could remove these events completely, but I don't know the full context of the problem.

<strike>This is not at all tested or code-reviewed - and is just here to see what people think.</strike>

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcher.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcher.php
@@ -14,12 +14,20 @@ namespace Sylius\Bundle\ResourceBundle\Controller;
 use Sylius\Bundle\ResourceBundle\Event\ResourceControllerEvent;
 use Sylius\Component\Resource\Model\ResourceInterface;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface as SymfonyEventDispatcherInterface;
+use Sylius\Bundle\ResourceBundle\Controller\RequestConfiguration;
+use Sylius\Bundle\ResourceBundle\ResourceControllerEvents;
+use Sylius\Component\Resource\ResourceActions;
 
 /**
  * @author Paweł Jędrzejewski <pawel@sylius.org>
+ * @author Daniel Leech <daniel@dantleech.com>
  */
 class EventDispatcher implements EventDispatcherInterface
 {
+    const PRE_PREFIX = 'pre_';
+    const POST_PREFIX = 'post_';
+    const NO_PREFIX = '';
+
     /**
      * @var SymfonyEventDispatcherInterface
      */
@@ -38,50 +46,61 @@ class EventDispatcher implements EventDispatcherInterface
      */
     public function dispatch($eventName, RequestConfiguration $requestConfiguration, ResourceInterface $resource)
     {
-        $eventName = $requestConfiguration->getEvent() ?: $eventName;
-        $metadata = $requestConfiguration->getMetadata();
-        $event = $this->getEvent($resource);
-
-        $this->eventDispatcher->dispatch(sprintf('%s.%s.%s', $metadata->getApplicationName(), $metadata->getName(), $eventName), $event);
+        $event = new ResourceControllerEvent($resource, $requestConfiguration);
+        $this->eventDispatcher->dispatch($eventName, $event);
+        $this->dispatchResourceEvent($eventName, $event);
 
         return $event;
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function dispatchPreEvent($eventName, RequestConfiguration $requestConfiguration, ResourceInterface $resource)
+    private function dispatchResourceEvent($eventName, ResourceControllerEvent $event)
     {
-        $eventName = $requestConfiguration->getEvent() ?: $eventName;
-        $metadata = $requestConfiguration->getMetadata();
-        $event = $this->getEvent($resource);
+        switch ($eventName) {
+            case ResourceControllerEvents::SHOW:
+                $this->doDispatchResourceEvent(self::NO_PREFIX, ResourceActions::SHOW, $event);
+                return;
+            case ResourceControllerEvents::INDEX:
+                $this->doDispatchResourceEvent(self::NO_PREFIX, ResourceActions::INDEX, $event);
+                return;
+            case ResourceControllerEvents::PRE_UPDATE:
+                $this->doDispatchResourceEvent(self::PRE_PREFIX, ResourceActions::UPDATE, $event);
+                return;
+            case ResourceControllerEvents::POST_UPDATE:
+                $this->doDispatchResourceEvent(self::POST_PREFIX, ResourceActions::UPDATE, $event);
+                return;
+            case ResourceControllerEvents::PRE_CREATE:
+                $this->doDispatchResourceEvent(self::PRE_PREFIX, ResourceActions::CREATE, $event);
+                return;
+            case ResourceControllerEvents::POST_CREATE:
+                $this->doDispatchResourceEvent(self::POST_PREFIX, ResourceActions::CREATE, $event);
+                return;
+            case ResourceControllerEvents::PRE_DELETE:
+                $this->doDispatchResourceEvent(self::PRE_PREFIX, ResourceActions::DELETE, $event);
+                return;
+            case ResourceControllerEvents::POST_DELETE:
+                $this->doDispatchResourceEvent(self::POST_PREFIX, ResourceActions::DELETE, $event);
+                return;
+        }
 
-        $this->eventDispatcher->dispatch(sprintf('%s.%s.pre_%s', $metadata->getApplicationName(), $metadata->getName(), $eventName), $event);
-
-        return $event;
+        throw new \RuntimeException(sprintf(
+            'Do not know how to dispatch resource event "%s"', $eventName
+        ));
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function dispatchPostEvent($eventName, RequestConfiguration $requestConfiguration, ResourceInterface $resource)
+    private function doDispatchResourceEvent($prefix, $eventName, ResourceControllerEvent $event)
     {
+        $requestConfiguration = $event->getRequestConfiguration();
         $eventName = $requestConfiguration->getEvent() ?: $eventName;
         $metadata = $requestConfiguration->getMetadata();
-        $event = $this->getEvent($resource);
 
-        $this->eventDispatcher->dispatch(sprintf('%s.%s.post_%s', $metadata->getApplicationName(), $metadata->getName(), $eventName), $event);
+        $this->eventDispatcher->dispatch(sprintf(
+            '%s.%s.%s%s',
+            $metadata->getApplicationName(),
+            $metadata->getName(),
+            $prefix,
+            $eventName
+        ), $event);
 
         return $event;
-    }
-
-    /**
-     * @param ResourceInterface $resource
-     *
-     * @return ResourceControllerEvent
-     */
-    private function getEvent(ResourceInterface $resource)
-    {
-        return new ResourceControllerEvent($resource);
     }
 }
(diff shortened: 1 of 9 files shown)


dantleech:
(sorry submitted the issue early by mistake, update the description now).

src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcher.php
```diff
@@ -1,87 +0,0 @@
...
-namespace Sylius\Bundle\ResourceBundle\Controller;
-
-use Sylius\Bundle\ResourceBundle\Event\ResourceControllerEvent;
-use Sylius\Component\Resource\Model\ResourceInterface;
-use Symfony\Component\EventDispatcher\EventDispatcherInterface as SymfonyEventDispatcherInterface;
-
-/**
- * @author Paweł Jędrzejewski <pawel@sylius.org>
- */
-class EventDispatcher implements EventDispatcherInterface
```
dantleech: With the new logic I don't think this class is justified.

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
```diff
@@ -171,7 +172,7 @@ public function showAction(Request $request)
         $this->isGrantedOr403($configuration, ResourceActions::SHOW);
         $resource = $this->findOr404($configuration);
 
-        $this->eventDispatcher->dispatch(ResourceActions::SHOW, $configuration, $resource);
+        $this->eventDispatcher->dispatch(ResourceControllerEvents::SHOW, new ResourceControllerEvent($resource, $configuration));
```
dantleech: We would now be able to reference resource events using a constant

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
```diff
@@ -102,7 +102,7 @@ protected function addController(ContainerBuilder $container, MetadataInterface
                 new Reference('sylius.resource_controller.redirect_handler'),
                 new Reference('sylius.resource_controller.flash_helper'),
                 new Reference('sylius.resource_controller.authorization_checker'),
-                new Reference('sylius.resource_controller.event_dispatcher'),
+                new Reference('event_dispatcher'),
```
dantleech: Use the standard SF event dispatcher.

src/Sylius/Bundle/ResourceBundle/Event/ResourceControllerEvent.php
```diff
@@ -157,4 +173,14 @@ public function setErrorCode($errorCode)
...
     }
+
+    /**
+     * Return the request configuration.
+     *
+     * @return RequestConfiguration
+     */
+    public function getRequestConfiguration()
+    {
+        return $this->requestConfiguration;
```
dantleech: Event has access to the request configuration, and through it the Request. For my use case this will allow access to a `parent_path` query parameter, allowing documents to be added at explicit points in a tree - this is required when creating a document from a tree browser
treebrowse

pjedrzejewski:
I will let this sink into my head a bit, but at first look it makes sense! We definitely want generic events, only question here is whether to keep the resource-specific ones. Dispatching an event is quite cheap, so I am leaning towards keeping them.

ste93cry:
I think we should just deprecate the resource-specific events and remove them in the next major version. They don't make sense of exists if this PR is merged and the most appropriate way to handle events becames moving the logic that handles resource types to the event listener.

dantleech:
I have updated this, reintroducing the controller-specfic `EventDispatcher` as it at least means we only instantiate the event in one place.

I am now at the point of porting the original dispatcher spec to the event subscriber which issues the old/resource events.

@pjedrzejewski shall we remove the resource-specific events or keep them?

src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
dantleech: Not any more ...

src/Sylius/Bundle/ResourceBundle/spec/EventListener/ResourceEventDistributorSubscriberSpec.php
```diff
@@ -0,0 +1,129 @@
...
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Sylius\Bundle\ResourceBundle\EventListener\ResourceEventDistributorSubscriber;
+
+/**
+ * @mixin EventDispatcher
+ *
+ * @author Paweł Jędrzejewski <pawel@sylius.org>
+ */
+class ResourceEventDistributorSubscriberSpec extends ObjectBehavior
```
dantleech: This is basically the old event dispatcher spec which i need to adapt/.

src/Sylius/Bundle/ResourceBundle/EventListener/ResourceEventDistributorSubscriber.php
```diff
@@ -0,0 +1,86 @@
+<?php
+
+namespace Sylius\Bundle\ResourceBundle\EventListener;
+
+use Sylius\Bundle\ResourceBundle\Controller\ResourceControllerEvents;
+
+class ResourceEventDistributorSubscriber implements EventSubscriberInterface
```
lchrusciel: Lets make it final.

src/Sylius/Bundle/ResourceBundle/ResourceControllerEvents.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Issued after a resource is created.
+     */
+    const PRE_CREATE = 'pre_create';
+
+    /**
+     * Issued after a resource has been created.
+     */
+    const POST_CREATE = 'post_create';
+
```
lchrusciel: Redundant

src/Sylius/Bundle/ResourceBundle/ResourceControllerEvents.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace Sylius\Bundle\ResourceBundle;
+
+/**
+ * Events that are issued by the resource controller.
+ */
+class ResourceControllerEvents
```
lchrusciel: Making it static would be a good idea, but unfortunately [it is not possible](https://wiki.php.net/rfc/abstract_final_class). But still, we are able to make it final, and create a private constructor. Hacky way, but there is no better solution.

lchrusciel:
@dantleech Sorry, didn't notice, that it is just a POC. Don't worry about my comments then :)

dantleech:
So do we think having general events is the way forward and do we keep the per-resource events or not? /cc @pjedrzejewski

bendavies:
vote yes on both counts

pjedrzejewski:
👍  for keeping generic + resource specific events. Sounds good @dantleech?

dantleech:
Good for me (would personally prefer removing resource-specific but not strongly).  Will finish this off tomorrow if I can.

src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcher.php
```diff
@@ -38,50 +46,68 @@ public function __construct(SymfonyEventDispatcherInterface $eventDispatcher)
...
-        $metadata = $requestConfiguration->getMetadata();
-        $event = $this->getEvent($resource);
-
-        $this->eventDispatcher->dispatch(sprintf('%s.%s.pre_%s', $metadata->getApplicationName(), $metadata->getName(), $eventName), $event);
+        switch ($eventName) {
+            case ResourceControllerEvents::SHOW:
+                $this->doDispatchResourceEvent(self::NO_PREFIX, ResourceActions::SHOW, $event);
+                return;
+            case ResourceControllerEvents::INDEX:
+                $this->doDispatchResourceEvent(self::PRE_PREFIX, ResourceActions::INDEX, $event);
```
dantleech: NO_PREFIX

src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcher.php
```diff
@@ -38,50 +46,61 @@ public function __construct(SymfonyEventDispatcherInterface $eventDispatcher)
...
+            case ResourceControllerEvents::PRE_DELETE:
+                $this->doDispatchResourceEvent(self::PRE_PREFIX, ResourceActions::DELETE, $event);
+                return;
+            case ResourceControllerEvents::POST_DELETE:
+                $this->doDispatchResourceEvent(self::POST_PREFIX, ResourceActions::DELETE, $event);
+                return;
+        }
 
-        return $event;
+        throw new \RuntimeException(sprintf(
```
dantleech: Test

src/Sylius/Bundle/ResourceBundle/Event/ResourceControllerEvent.php
```diff
@@ -44,6 +45,21 @@ class ResourceControllerEvent extends GenericEvent
     protected $errorCode = 500;
 
     /**
+     * @var RequestConfiguration
+     */
+    protected $requestConfiguration;
+
+    /**
+     * @param mixed $subject
```
dantleech: object

dantleech:
Updated. Now dispatching the per-resource events through the controller's event dispatcher (rather than a new subscriber).

src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcher.php
```diff
@@ -38,50 +46,61 @@ public function __construct(SymfonyEventDispatcherInterface $eventDispatcher)
...
+                $this->doDispatchResourceEvent(self::PRE_PREFIX, ResourceActions::CREATE, $event);
+                return;
+            case ResourceControllerEvents::POST_CREATE:
+                $this->doDispatchResourceEvent(self::POST_PREFIX, ResourceActions::CREATE, $event);
+                return;
+            case ResourceControllerEvents::PRE_DELETE:
+                $this->doDispatchResourceEvent(self::PRE_PREFIX, ResourceActions::DELETE, $event);
+                return;
+            case ResourceControllerEvents::POST_DELETE:
+                $this->doDispatchResourceEvent(self::POST_PREFIX, ResourceActions::DELETE, $event);
```
dantleech: In good conscience I should probably add specs for each of these ..

dantleech:
Tests failing with:

```
Then I should see that static content # features/cmf/showing_static_content.feature:14
      Could not open the page: "http://localhost:8080/krzysztof-krawczyk". Received an error status code: 404 (Sylius\Behat\Page\UnexpectedPageException)
```

Known issue?

pamil:
@dantleech it's the most common false negative now. I'll try to solve it, but for now the build has to be restarted (either by someone with ability to do it or by you amending and force pushing :))

bendavies:
no movement here?

PR #5483, opened 2016-07-10
[Docs] Added missing bundle in SyliusResourceBundle installation docs

Zales0123:
| Q               | A
| --------------- | ---
| Doc fix?        | yes
| New docs?    | no
| BC breaks?      | no
| Related tickets |
| License         | MIT

``winzouStateMachineBundle`` is also required, as ``ResourceController`` depends on ``sm.factory`` service.


pjedrzejewski:
Thanks Mateusz! 👍

rvanlaak:
Because of the missing `sm.factory` service `v0.19` now is broken if you only install `promotion-bundle`.

It seems that more services are/were missing in `resource-bundle/Resources/config/controller.xml`:

image

PR #5485, opened 2016-07-11
[Registry] Added generic service registry compiler pass

dantleech:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| License | MIT |

This PR adds a generic compiler pass for the registries implementing `ServiceRegistryInterface`.

Currently it replaces the GridBundle passes, but without looking too closely it could replace 16 others.


src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @param string $serviceRegistryId
+     * @param string $serviceTag
+     * @param array $options
+     */
+    public function __construct(
+        $serviceRegistryId,
+        $serviceTag,
+        array $options = []
```
dantleech: Not sure if this is overkill, but I do not like "optional" arguments :/

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
```diff
@@ -0,0 +1,99 @@
...
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+
+namespace Sylius\Component\Registry\DependencyInjection\Compiler;
```
dantleech: I guess it is debateable if there should be Symfony specific couplings in the component, but it would be a shame for this to be in the bundle as my use case here is actually using it in a Symfony CMF project. If we decide to put this in we should add the DI component to `suggests` in composer.json

pamil:
Shouldn't that compiler pass be placed in Registry component (we don't have RegistryBundle yet) instead?

src/Sylius/Bundle/GridBundle/SyliusGridBundle.php
```diff
@@ -29,8 +30,30 @@ public function build(ContainerBuilder $container)
...
         parent::build($container);
 
-        $container->addCompilerPass(new RegisterDriversPass());
-        $container->addCompilerPass(new RegisterFiltersPass());
-        $container->addCompilerPass(new RegisterFieldTypesPass());
+        $container->addCompilerPass(new AliasedServicePass(
+            'sylius.registry.grid_driver',
+            'sylius.grid_driver',
+            [ 
+                'context' => 'Grid Driver'
```
lchrusciel: Missing comma at the end.

src/Sylius/Bundle/GridBundle/SyliusGridBundle.php
```diff
@@ -29,8 +30,30 @@ public function build(ContainerBuilder $container)
...
+            ]
+        ));
+
+        $container->addCompilerPass(new AliasedServicePass(
+            'sylius.registry.grid_field',
+            'sylius.grid_field',
+            [
+                'context' => 'Grid field',
+                'alias_attribute' => 'type'
```
lchrusciel: Missing comma at the end here and in a below lines

dantleech: Argh. Too much JSON configuration :)

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
```diff
@@ -0,0 +1,99 @@
...
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+use Sylius\Component\Registry\ServiceRegistryInterface;
+
+/**
+ * This compiler pass will add services to a registered ServiceRegistry
+ * definition.
+ *
+ * @author Daniel Leech <daniel@dantleech.com>
+ */
+class AliasedServicePass implements CompilerPassInterface
```
lchrusciel: I would vote to make this class final. It is hundred lines long and pretty project specific.

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
```diff
@@ -0,0 +1,99 @@
...
+
+        if (!$reflection->implementsInterface(ServiceRegistryInterface::class)) {
+            throw new InvalidArgumentException(sprintf(
+                'The service registry "%s" must implement the "%s" interface.',
+                $this->serviceRegistryId, ServiceRegistryInterface::class
+            ));
+        }
+
+        foreach ($container->findTaggedServiceIds($this->serviceTag) as $id => $attributes) {
+            if (!isset($attributes[0][$aliasAttribute]))  {
```
pamil: We're probably missing support for multiple tags for one service.

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * This compiler pass will add services to a registered ServiceRegistry
+ * definition.
+ *
+ * @author Daniel Leech <daniel@dantleech.com>
+ */
+class AliasedServicePass implements CompilerPassInterface
+{
+    private $serviceRegistryId;
```
lchrusciel: Missing docblock. Here, and in the below properties.

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
pamil: Why don't we replace it with `string $aliasAttribute` and remove `context` option?

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        $options = $this->resolveOptions();
+        $aliasAttribute = $options['alias_attribute'];
+        $context = $options['context'];
+
+        $registry = $container->findDefinition($this->serviceRegistryId);
+        $reflection = new \ReflectionClass($registry->getClass());
+
+        if (!$reflection->implementsInterface(ServiceRegistryInterface::class)) {
```
lchrusciel: `Assert::implementsInterface`

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
lchrusciel: `Assert::notEmpty`

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
```diff
@@ -0,0 +1,99 @@
...
+                'The service registry "%s" must implement the "%s" interface.',
+                $this->serviceRegistryId, ServiceRegistryInterface::class
+            ));
+        }
+
+        foreach ($container->findTaggedServiceIds($this->serviceTag) as $id => $attributes) {
+            if (!isset($attributes[0][$aliasAttribute]))  {
+                throw new \InvalidArgumentException(sprintf(
+                    '%s service with tag "%s" needs to have the "%s" attribute.',
+                    $context, $this->serviceTag, $aliasAttribute
```
pamil: It's the only place where `$context` is used and it does not even say which service does not have that attribute.

```
Service "foo.bar" with tag "baz" needs to have the "alias" attribute.
```

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
```diff
@@ -0,0 +1,99 @@
...
+                ));
+            }
+
+            $registry->addMethodCall('register', [$attributes[0][$aliasAttribute], new Reference($id)]);
+        }
+    }
+
+    private function resolveOptions()
```
pamil: Missing docblock.

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
pamil: Missing docblocks.

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
pamil: Yup, it should be either in Registry component or in RegistryBundle, which needs to be created.

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
lchrusciel: Im would vote to add them to constructor with some default value. On the other hand, I'm not convinced about defining one of the options, just because it is declared before the option that I want to define...

src/Sylius/Component/Registry/spec/DependencyInjection/Compiler/AliasedServicePassSpec.php
```diff
@@ -0,0 +1,134 @@
...
+
+    function it_is_initializable()
+    {
+        $this->shouldHaveType(AliasedServicePass::class);
+    }
+
+    function it_should_return_early_if_the_container_does_not_have_the_registry_definition(
+        ContainerBuilder $container
+    )
+    {
```
pamil: CS -> `) {`

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
dantleech: Well, if there was only 1 option, but I don't like `..., $aliasAttribute = 'alias', $context = null`. As it means that if you want to set the context you need to redefine the alias. Also if we were to add another option later it would become very ugly.

But its my personal preference, I don't mind (very) much making them arguments.

dantleech: it is also deabatable as to whether `context` is required - it is just making the exception a bit better to read.

src/Sylius/Component/Registry/spec/DependencyInjection/Compiler/AliasedServicePassSpec.php
```diff
@@ -0,0 +1,134 @@
...
+use Sylius\Component\Registry\DependencyInjection\Compiler\AliasedServicePass;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+use Sylius\Component\Registry\ServiceRegistry;
+use Symfony\Component\DependencyInjection\Reference;
+
+/**
+ * @author Daniel Leech <daniel@dantleech.com>
+ */
+class AliasedServicePassSpec extends ObjectBehavior
```
pamil: Hmm, should we add some integration tests to be sure, that Symfony works just like we think it works? Or replace this spec with it?

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
pamil: @dantleech as for the `$context` argument, this is the change I propose: https://github.com/Sylius/Sylius/pull/5485#discussion_r70219505

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
pamil: Won't work here :)

lchrusciel: Of course. My bad. `Assert::true` than. :)

src/Sylius/Component/Registry/spec/DependencyInjection/Compiler/AliasedServicePassSpec.php
dantleech: Possibly in addition to this spec, although failure in Sylius is all or nothing if we replace all the existing service registry compiler passes - and even then chance of failure is more likely at the level of instantiating the pass (e.g. specifying the wrong tag/alias).

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
dantleech: Yeah we should add that, the idea with the context is just to give better DX: 

```
Grid driver service "my_service" with tag "foo.bar" needs ..
```

If we remove the options then I don't mind removing the `$context`.

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
dantleech: @pamil can you explain more? Not sure how you can have multiple tags for one service?

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
dantleech: Not sure if it makes sense to use Assert here - firstly It means not using the DependencyInjection-specific `InvalidArgumentException`, and secondly we still need to format the message with `sprintf`

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
dantleech: > Assert::notEmpty

Again, we lose the dependency-injection specific exception class, and have to format hte message anyway so as not to lose information.

pamil:
```xml
<service id="foo.bar" class="Foo/Bar">
  <tag name="bar.baz" alias="newAlias" />
  <tag name="bar.baz" alias="legacyAlias" />
</service>
```

This way `$container->findTaggedServiceIds('bar.baz')` will return:

```php
[
  'foo.bar' => [
    ['alias' => 'newAlias'],
    ['alias' => 'legacyAlias'],
  ],
]
```

Right now only the first one will be handled correctly.

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
lchrusciel: `sprintf` is not a problem, you can provide your custom message to the webmozart as a third argument. But I haven't noticed, that you are using a `DependencyInjection-specific` error. We should leave it as it is. _Mea culpa_

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
dantleech: Well -- I don't think putting it in a bundle would make much sense really - the bundle would not *do* anything (that I am aware) - not DI config, not templates, not assets. The most appropriate place for this would be, using SF as an example, the `Bridge` namespace. But, the overhead of creating and requiring that ...

dantleech:
Updated, removed the `options`. Still outstanding comments, and the other ~19 passes to replace if this is accepted.

src/Sylius/Component/Registry/DependencyInjection/Compiler/AliasedServicePass.php
dantleech: > you can provide your custom message to the webmozart

You can, but you cannot provide the arguments to sprintf, hence why you would need to do it before passing the message.

> But I haven't noticed, that you are using a DependencyInjection-specific

I was sneaky and put it there a few minutes ago :)

dantleech:
So it turns out this pass is a direct replacement for only a few of the compiler passes. I think I have identified two types of pass: 

- **Alias-only**: As with the pass currently in this PR. Simply adds a named service to the registry.
- **Alias-Labels**: Uses an alias (service name) attribute, but also has a label attribute and sets an assoicative array of labels in the container.

- Alias-Labels: 
    - price calcuator, 
    - rule checker (promotion)
    - promotion actions. 
    - `RegisterRenderersPass`
    - `RegisterCalculatorPass`
    - `RegisterRuleCheckerPass`
    - `RegisterShippingMethodsResolverPass`
    - `RegisterAttributesTypePass`, 
    - `RegisterCalculatorPass` (uses "alias" as value of the label)
- Alias-only: RegisterSchemasPass, Grid passes, RegisterResolversPass, 

So we could add a "Alias-Labels" pass too, but this is quite Sylius-specific.

I think the challenge is really how to reduce the obvious repetition between these classes whilst allowing a degree of flexibility.

PR #5486, opened 2016-07-11
fix SetupCommand 'tax_calculation_strategy' should not be null

nogo:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

When run sylius:install step 3 fails because tax_calculation_strategy is null.

Step 1: It appears that your database already exists. Would you like to reset it? (y/N) y
Step 2: Load sample data? (y/N) N
Step 3: fails because tax_calculation_strategy is null


pjedrzejewski:
Good catch! Thanks!

michalmarcinkowski:
Thanks Danilo!

PR #5487, opened 2016-07-11
Fix showing channel in the collector

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

Right now, it always has `null` as channel.


pamil:
Btw. I think we need some way to test the profilers e2e :)

michalmarcinkowski:
👍

PR #5488, opened 2016-07-11
Make CurrencyContext work correctly without channel

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | -
| License         | MIT

BC breaks:
 - currency contexts made final
 - core currency context no longer extends the one from component
 - constant `STORAGE_KEY` removed from the interface
 - default currency is taken from the `%currency%` parameter, not from the settings


pamil:
It's just a hotfix, the whole currency / locale management needs a solid refactoring.

pjedrzejewski:
Thanks Kamil!

PR #5490, opened 2016-07-11
Windows path compatibility:

caponica:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

Updated driver detection to work with Windows file paths, which have "\" instead of "/".


pjedrzejewski:
Thanks Christian! 👍

Issue #5476, opened 2016-07-08
[Grid] How to filter grid with join entity.


liverbool:
I'm deep looking in Grid/GridBundle but can't find the way to filter by using `join`.

aramalipoor:
:+1:

I think because `DataSource` is thought to be generic it prevents us to access query builder. Maybe introduce a `OrmDataSourceInterface`? Currently I'm using a very hackish way (reflection to access queryBuilder private property) to modify query builder!!

Anyway another problem I encounter is that Pagerfanta has problems with counting and forces you to do [manual counting](https://github.com/KnpLabs/KnpPaginatorBundle/blob/master/Resources/doc/manual_counting.md), and to do that you need to `setHint` on query! Which you don't have access to even with queryBuilder.

So more flexibility in Grid DataSource is crucial for more advanced usages. WDYT Sylius :) ?

pjedrzejewski:
You have access to custom query builder - https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/AdminBundle/Resources/config/grids/order.yml#L9. What we are missing is ``entity`` filter type, that will use ``entity`` form to render the choices, then you can easily filter on joint entity. Should be simple to implement. (and actually implemented in old grids at "grids" branch of this repo)

oatto:
I agree with @aramalipoor to introduce either `OrmDataSourceInterface` or `OrmExpressionBuilderInterface` for accessing to the query builder.

michalmarcinkowski:
Done!

mikemix:
Well this pull request is quite useless ;/ What if I need two joins not one? Simple `getQueryBuilder()` getter would be sufficient and let us to the hard filtering work.

PR #5501, opened 2016-07-12
[GridBundle] Allow filtering by fields from associated objects

aramalipoor:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | mentioned in #5476
| License         | MIT

I also had some issues when filter form was submitting with empty string value so I thought maybe do not restrict DataSource when filer data is empty?!

### DIFF
--- a/src/Sylius/Bundle/GridBundle/Doctrine/ORM/ExpressionBuilder.php
+++ b/src/Sylius/Bundle/GridBundle/Doctrine/ORM/ExpressionBuilder.php
@@ -74,7 +74,7 @@ class ExpressionBuilder implements ExpressionBuilderInterface
      */
     public function equals($field, $value)
     {
-        $this->queryBuilder->setParameter($field, $value);
+        $this->setParameter($field, $value);
 
         return $this->queryBuilder->expr()->eq($this->getFieldName($field), ':'.$field);
     }
@@ -84,7 +84,7 @@ class ExpressionBuilder implements ExpressionBuilderInterface
      */
     public function notEquals($field, $value)
     {
-        $this->queryBuilder->setParameter($field, $value);
+        $this->setParameter($field, $value);
 
         return $this->queryBuilder->expr()->neq($this->getFieldName($field), ':'.$field);
     }
@@ -94,7 +94,9 @@ class ExpressionBuilder implements ExpressionBuilderInterface
      */
     public function lessThan($field, $value)
     {
-        $this->queryBuilder->andWhere($this->getFieldName($field).' < :'.$field)->setParameter($field, $value);
+        $this->setParameter($field, $value);
+
+        $this->queryBuilder->andWhere($this->getFieldName($field).' < :'.$field);
     }
 
     /**
@@ -102,7 +104,9 @@ class ExpressionBuilder implements ExpressionBuilderInterface
      */
     public function lessThanOrEqual($field, $value)
     {
-        $this->queryBuilder->andWhere($this->getFieldName($field).' =< :'.$field)->setParameter($field, $value);
+        $this->setParameter($field, $value);
+
+        $this->queryBuilder->andWhere($this->getFieldName($field).' =< :'.$field);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/GridBundle/Doctrine/ORM/ExpressionBuilder.php
```diff
@@ -94,31 +94,39 @@ public function notEquals($field, $value)
...
 
     /**
      * {@inheritdoc}
      */
     public function lessThanOrEqual($field, $value)
     {
-        $this->queryBuilder->andWhere($this->getFieldName($field).' =< :'.$field)->setParameter($field, $value);
+        $this->setParameter($field, $value);
+
+        $this->queryBuilder->andWhere($this->getFieldName($field).' =< :'.$field);
```
bigfoot90: This should be ```<=```

src/Sylius/Bundle/GridBundle/Doctrine/ORM/ExpressionBuilder.php
```diff
@@ -94,31 +94,39 @@ public function notEquals($field, $value)
...
 
     /**
      * {@inheritdoc}
      */
     public function greaterThanOrEqual($field, $value)
     {
-        $this->queryBuilder->andWhere($this->getFieldName($field).' => :%s'.$field)->setParameter($field, $value);
+        $this->setParameter($field, $value);
+
+        $this->queryBuilder->andWhere($this->getFieldName($field).' => :%s'.$field);
```
bigfoot90: This should be ```>=```

liverbool:
@pjedrzejewski why this PR not merge? What do you think?

src/Sylius/Bundle/GridBundle/Doctrine/ORM/ExpressionBuilder.php
```diff
@@ -74,7 +74,7 @@ public function comparison($field, $operator, $value)
      */
     public function equals($field, $value)
     {
-        $this->queryBuilder->setParameter($field, $value);
+        $this->setParameter($field, $value);
 
         return $this->queryBuilder->expr()->eq($this->getFieldName($field), ':'.$field);
```
liverbool: BUG: `$field` still not replace with `_`

src/Sylius/Bundle/GridBundle/Doctrine/ORM/ExpressionBuilder.php
```diff
@@ -84,7 +84,7 @@ public function equals($field, $value)
      */
     public function notEquals($field, $value)
     {
-        $this->queryBuilder->setParameter($field, $value);
+        $this->setParameter($field, $value);
 
         return $this->queryBuilder->expr()->neq($this->getFieldName($field), ':'.$field);
```
liverbool: same

src/Sylius/Bundle/GridBundle/Doctrine/ORM/ExpressionBuilder.php
```diff
@@ -94,31 +94,39 @@ public function notEquals($field, $value)
      */
     public function lessThan($field, $value)
     {
-        $this->queryBuilder->andWhere($this->getFieldName($field).' < :'.$field)->setParameter($field, $value);
+        $this->setParameter($field, $value);
+
+        $this->queryBuilder->andWhere($this->getFieldName($field).' < :'.$field);
```
liverbool: same :0

src/Sylius/Bundle/GridBundle/Doctrine/ORM/ExpressionBuilder.php
aramalipoor: OOPS!! MY BAD!!

pjedrzejewski:
Build is failing, otherwise looks good! 👍

aramalipoor:
@pjedrzejewski The fail is about `it filters data containing empty strings` which seems legit, and now I don't know what to decide here :D ?

The  problem is I do not want to apply any filter when form input of filter is not filled by user (e.g. filling only another field) because it will automatically only show data with empty strings! But on the other hand what to do if user **really** wants to filter empty strings?

pjedrzejewski:
The problems is that string filter type: empty / non-empty has no value - so your ``if (empty...`` breaks the tests. I guess we should check if filter type is not these 2.

aramalipoor:
@pjedrzejewski Thanks for the feedback :) I think it's ready now

pjedrzejewski:
Thanks Aram! Tests FTW!

PR #5505, opened 2016-07-13
[Order][Sequence] Remove SequenceBundle and move logic of generating numbers to OrderBundle

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

PR #5516, opened 2016-07-15
[WIP][Order] Refactor order state machine

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

Based on #5505


bendavies:
why has sequence/SequenceBundle been removed?
I was going to use this in a project soon. It's pretty good as a standalone.

GSadee:
@bendavies it is a question to https://github.com/Sylius/Sylius/pull/5505. Sequence and SequenceBundle are going to be removed because of its redundancy. We were using it only to generate order number in only one specific case, so this simple logic should be in `OrderBundle`.

bendavies:
That doesn't negate the fact that it's a nice standalone component. demonstrated by the fact i was going to use it.

I think it would be a shame.

src/Sylius/Bundle/OrderBundle/NumberGenerator/OrderNumberGenerator.php
```diff
@@ -0,0 +1,100 @@
...
+namespace Sylius\Bundle\OrderBundle\NumberGenerator;
+
+use Sylius\Component\Order\Model\OrderSequenceInterface;
+use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+/**
+ * @author Grzegorz Sadowski <grzegorz.sadowski@lakion.com>
+ */
+class OrderNumberGenerator implements OrderNumberGeneratorInterface
```
lchrusciel: This class could be final as well

michalmarcinkowski:
@bendavies you can still use v0.19 of this component if you want to. We already have so many packages that we need to reduce it to minimum. We don't want to maintain this component as a generic and standalone, cause it's more overhead than benefit for us.

pjedrzejewski:
Thanks Grzesiu!

src/Sylius/Behat/Context/Setup/OrderContext.php
```diff
@@ -391,17 +391,17 @@ public function thisOrderIsAlreadyPaid(OrderInterface $order)
...
     {
-        $this->stateMachineFactory->get($order, OrderTransitions::GRAPH)->apply(OrderTransitions::SYLIUS_CANCEL);
+        $this->stateMachineFactory->get($order, OrderTransitions::GRAPH)->apply(OrderTransitions::TRANSITION_CANCEL);
         
         $this->objectManager->flush();
     }
 
     /**
-     * @Given /^the customer confirmed (this order)$/
+     * @Given /^the customer fulfilled (this order)$/
```
michalmarcinkowski: This step will not be used.

src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
```diff
@@ -77,7 +77,7 @@ public function releaseInventoryAction(Request $request)
 
         $this->container->get('sm.factory')
             ->get($order, OrderTransitions::GRAPH)
-            ->apply(OrderTransitions::SYLIUS_RELEASE)
+            ->apply(OrderTransitions::TRANSITION_RELEASE)
```
michalmarcinkowski: There is no `TRANSITION_RELEASE` anymore :)

src/Sylius/Bundle/OrderBundle/Form/Type/OrderStateChoiceType.php
```diff
@@ -27,13 +27,8 @@ public function configureOptions(OptionsResolver $resolver)
         $resolver->setDefaults([
             'choices' => [
                 OrderInterface::STATE_CART => 'sylius.order.state.checkout',
```
michalmarcinkowski: sylius.order.state.**cart**

src/Sylius/Bundle/OrderBundle/Form/Type/OrderStateChoiceType.php
```diff
@@ -27,13 +27,8 @@ public function configureOptions(OptionsResolver $resolver)
...
-                OrderInterface::STATE_CART_LOCKED => 'sylius.order.state.cart_locked',
-                OrderInterface::STATE_PENDING => 'sylius.order.state.ordered',
-                OrderInterface::STATE_CONFIRMED => 'sylius.order.state.order_confirmed',
-                OrderInterface::STATE_SHIPPED => 'sylius.order.state.shipped',
-                OrderInterface::STATE_ABANDONED => 'sylius.order.state.abandoned',
                 OrderInterface::STATE_CANCELLED => 'sylius.order.state.cancelled',
-                OrderInterface::STATE_RETURNED => 'sylius.order.state.returned',
+                OrderInterface::STATE_FULFILLED => 'sylius.order.state.fulfilled',
```
michalmarcinkowski: Missing `new` state.

michalmarcinkowski: Btw. where we use this choice type? I think it should be removed 😉

PR #5523, opened 2016-07-17
save refresh token when oauth login

videni:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| License         | MIT


src/Sylius/Component/User/Model/UserOAuthInterface.php
```diff
@@ -60,4 +60,21 @@ public function getAccessToken();
      * @param string $accessToken
      */
     public function setAccessToken($accessToken);
+
+
+    /**
+     * Get OAuth refresh token.
```
lchrusciel: Redundant

videni: it is fixed

michalmarcinkowski:
Thanks @videni!

PR #5535, opened 2016-07-19
[BC Break] Unplug SyliusWebBundle and SyliusSearchBundle

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | 
| License         | MIT


okwinza:
What about SearchBundle, guys? Any plans to bring it back before 1.0?

src/Sylius/Bundle/CoreBundle/Sitemap/Provider/ProductUrlProvider.php
```diff
@@ -66,7 +65,7 @@ public function generate()
 
         foreach ($products as $product) {
             $productUrl = $this->sitemapUrlFactory->createNew();
-            $localization = $this->router->generate($product, [], true);
+            $localization = $this->router->generate('sylius_shop_product_show', ['slug' => $product->getSlug()], true);
```
pjedrzejewski: This couples core with ShopBundle, but I will open a PR, which fixes that later.

pjedrzejewski:
Thanks Grzesiu! 👍

pjedrzejewski:
@okwinza ElasticSearch integration will come in a different form before 1.0, absolutely. But we do not want to support or maintain mysql based search indexing in Sylius itself. So this bundle needs to be removed, as much as I hate to say that. :/

dantleech:
This PR removed the template customization for the checkout: https://github.com/Sylius/Sylius/pull/5535/files#diff-900dfac95dd61a7259c479f8cc58b329L131

Is there another way to do this?

pjedrzejewski:
@dantleech We will be working on checkout customization docs (/cc @TheMadeleine), but new checkout is mostly routing and state machine configuration, so you can easily customize the templates. See ``ShopBundle/Resources/config/routing/checkout.yml``.

dantleech:
sounds good :) do you mean that I should override that route (`sylius_shop_checkout_summary`) ?

dantleech:
so just to be clear - the FlowBundle / CheckoutScenario is now redundant?

pjedrzejewski:
@dantleech Yes... :)

PR #5556, opened 2016-07-20
[Core] Remove LocaleController

pamil:
That's why we should make classes `final` by default and then consider whether they should be extendable :confused:


src/Sylius/Bundle/CoreBundle/Controller/LocaleController.php
```diff
@@ -1,29 +0,0 @@
...
- * @author Aram Alipoor <aram.alipoor@gmail.com>
- */
-class LocaleController extends BaseLocaleController
-{
-    /**
-     * @return ChannelAwareLocaleProvider
-     */
-    protected function getLocaleProvider()
-    {
-        return $this->container->get('sylius.channel_aware_locale_provider');
```
michalmarcinkowski: DI through inheritance 💣 💥 😮

pjedrzejewski:
Thank you Kamil!

PR #5562, opened 2016-07-21
[Order] Fix number generator

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/OrderBundle/NumberGenerator/OrderNumberGeneratorInterface.php
+++ b/src/Sylius/Bundle/OrderBundle/NumberGenerator/OrderNumberGeneratorInterface.php
@@ -11,13 +11,17 @@
 
 namespace Sylius\Bundle\OrderBundle\NumberGenerator;
 
+use Sylius\Component\Order\Model\OrderInterface;
+
 /**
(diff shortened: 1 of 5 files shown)


pjedrzejewski:
👍

src/Sylius/Bundle/OrderBundle/NumberGenerator/OrderNumberGeneratorInterface.php
```diff
@@ -11,13 +11,17 @@
...
-interface SequentialOrderNumberGeneratorInterface
+interface OrderNumberGeneratorInterface
 {
     /**
+     * @param OrderInterface $order
+     *
      * @return string
      */
-    public function generate();
+    public function generate(OrderInterface $order);
```
lchrusciel: Why did you do this? It is just a generator and it does not care about an order itself.

pjedrzejewski: You may want to generate order number based on: channel code, date, customer group... whatever. This gives us access to all info about the order. Otherwise we have 0.

PR #5567, opened 2016-07-21
[Payment] Refactor payment state machine

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | 
| License         | MIT


features/order/managing_orders/cancelling_order_with_payment.feature
```diff
@@ -0,0 +1,22 @@
...
+        And the customer bought a single "PHP T-Shirt"
+        And the customer chose "Free" shipping method to "France" with "Cash on Delivery" payment
+        And I am logged in as an administrator
+
+    @ui
+    Scenario: Checking payment state of a placed order
+        When I view the summary of the order "#00000022"
+        And I cancel this order
+        Then I should be notified that it has been successfully updated
+        Then it should have cancelled payment state
```
michalmarcinkowski: How about adding this single assertion to already existing [cancel order feature](https://github.com/michalmarcinkowski/Sylius/blob/5c266bb1602a16f0862b7ffce373b14fc71d5f65/features/order/managing_orders/cancelling_order.feature#L23-L23)?

We could check all 3 states (order state, order payment state and order shipment state) in one scenario and avoid duplicated scenarios.

features/order/managing_orders/cancelling_order_with_payment.feature
```diff
@@ -0,0 +1,22 @@
...
+        And the store has a product "PHP T-Shirt"
+        And the store ships everywhere for free
+        And the store allows paying with "Cash on Delivery"
+        And there is a customer "john.doe@gmail.com" that placed an order "#00000022"
+        And the customer bought a single "PHP T-Shirt"
+        And the customer chose "Free" shipping method to "France" with "Cash on Delivery" payment
+        And I am logged in as an administrator
+
+    @ui
+    Scenario: Checking payment state of a placed order
```
michalmarcinkowski: After cancelling order its payment state should be cancelled

features/order/managing_orders/placing_order_with_payment.feature
```diff
@@ -0,0 +1,21 @@
...
+        And there is a customer "lucy@teamlucifer.com" that placed an order "#00000666"
+        And the customer bought a single "Angel T-Shirt"
+        And the customer "Lucifer Morningstar" addressed it to "Seaside Fwy", "90802" "Los Angeles" in the "United States" with identical billing address
+        And the customer chose "Free" shipping method with "Cash on Delivery" payment
+        And I am logged in as an administrator
+
+    @ui
+    Scenario: Checking payment state of a placed order
+        Given I view the summary of the order "#00000666"
+        Then it should have new payment state
```
michalmarcinkowski: Same as above, we can verify this state using already [existing scenario](https://github.com/michalmarcinkowski/Sylius/blob/c4661acfc97c53c2d1eb765c4aea91be1f0321f4/features/order/placing_an_order/placing_an_order_as_a_customer.feature#L21-L21).

src/Sylius/Component/Payment/Model/PaymentInterface.php
```diff
@@ -19,16 +19,16 @@
...
     const STATE_NEW = 'new';
-    const STATE_PENDING = 'pending';
     const STATE_PROCESSING = 'processing';
     const STATE_COMPLETED = 'completed';
-    const STATE_AUTHORIZED = 'authorized';
     const STATE_FAILED = 'failed';
     const STATE_CANCELLED = 'cancelled';
     const STATE_VOID = 'void';
     const STATE_REFUNDED = 'refunded';
+
```
michalmarcinkowski: Redundant

features/order/managing_orders/cancelling_order_with_payment.feature
GSadee: I wasn't sure if we should join these checks into one scenario, but I of course agree with you. It would be more readable.

features/order/managing_orders/placing_order_with_payment.feature
```diff
@@ -0,0 +1,21 @@
+@managing_orders
+Feature: Checking payment state of a placed order
+    In order to check payment state after placing an order
+    As an Administrator
+    I want to payment has proper state
```
pjedrzejewski: This does not sounds right

features/order/managing_orders/placing_order_with_payment.feature
```diff
@@ -0,0 +1,21 @@
+@managing_orders
+Feature: Checking payment state of a placed order
```
pjedrzejewski: Feature: Payments are in the state "new" after checkout

Something like that?

features/order/managing_orders/placing_order_with_payment.feature
```diff
@@ -0,0 +1,21 @@
...
+    In order to check payment state after placing an order
```
pjedrzejewski: In order to correctly process customer's payments

features/order/managing_orders/placing_order_with_payment.feature
pjedrzejewski: I want to have new payments after my customer's checkout

pjedrzejewski: Not perfect, but I have no better idea atm.

features/order/managing_orders/placing_order_with_payment.feature
```diff
@@ -0,0 +1,21 @@
...
+        And there is a customer "lucy@teamlucifer.com" that placed an order "#00000666"
+        And the customer bought a single "Angel T-Shirt"
+        And the customer "Lucifer Morningstar" addressed it to "Seaside Fwy", "90802" "Los Angeles" in the "United States" with identical billing address
+        And the customer chose "Free" shipping method with "Cash on Delivery" payment
+        And I am logged in as an administrator
+
+    @ui
+    Scenario: Checking payment state of a placed order
+        When I view the summary of the order "#00000666"
+        Then it should have new payment state
```
pjedrzejewski: Maybe we could be more clear here: Then it should have payment state "new", because "new payment state" means something different.

pjedrzejewski:
Thank you Grzesiu! 👍

src/Sylius/Component/Payment/PaymentTransitions.php
```diff
@@ -11,15 +11,18 @@
 
 namespace Sylius\Component\Payment;
 
+/**
+ * @author Grzegorz Sadowski <grzegorz.sadowski@lakion.com>
+ */
 class PaymentTransitions
```
lchrusciel: We should make this class final. In addition, I would also suggest to add a private constructor.

PR #5582, opened 2016-07-22
[Theme] Support namespaced paths

### LINKED ISSUE #5560: [ThemeBundle] Namespaced paths not overridden
Just playing around with master right now, and instead of overriding the shop bundle templates, I decided to try out the Theme bundle.

I got pretty far until I realized that the Theme bundle doesn't seem to be able to override namespaced paths (http://symfony.com/doc/current/cookbook/templating/namespaced_paths.html).

For example, the shop bundles `views/Homepage/index.html.twig` has `{% extends '@SyliusShop/layout.html.twig' %}`. I cannot override layout.html.twig via a theme until I change to to use `{% extends 'SyliusShopBundle::layout.html.twig' %}`.

Am I missing something, is this a bug, or a limitation? (If it's a limitation, maybe the ShopBundle shouldn't use namespaced paths?)

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #5560
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ThemeBundle/DependencyInjection/SyliusThemeExtension.php
+++ b/src/Sylius/Bundle/ThemeBundle/DependencyInjection/SyliusThemeExtension.php
@@ -69,6 +69,7 @@ final class SyliusThemeExtension extends Extension implements PrependExtensionIn
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config/services'));
 
         $this->prependSyliusSettings($container, $loader);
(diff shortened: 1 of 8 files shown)


pamil:
It will be worth to have a look at the performance metrics. Moreover, we should consider rewriting this whole template name to template path logic as it has become quite a mess.

pjedrzejewski:
Thank you Kamil! 👍

PR #5589, opened 2016-07-24
Use factory to create model resource

BenjaminGoetz:
Use factory instead of repository to create model resource


pjedrzejewski:
Good catch, thanks a lot Benjamin! 👍

PR #5597, opened 2016-07-25
[Locale] Locale refactoring

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | -
| License         | MIT

Basically the same as currency refactoring in #5497, but much simplier :)


pjedrzejewski:
Thanks Kamil! 👍

src/Sylius/Bundle/LocaleBundle/Listener/RequestLocaleSetter.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        $request = $event->getRequest();
+
+        try {
+            $request->setLocale($this->localeContext->getLocaleCode());
+        } catch (LocaleNotFoundException $exception) {}
+
+        try {
+            $request->setDefaultLocale($this->localeProvider->getDefaultLocaleCode());
+        } catch (LocaleNotFoundException $exception) {}
```
michalmarcinkowski: Default locale is mandatory, so we shouldn't silent the exception.

pamil: Had some reasoning about it, but can't recall it. I'll take a look & think how to manage this in #5601.

ylastapis:
Hi @pamil 

Could you tell me now how to set the currentLocale as LocaleContext became ImmutableLocaleContext ? :)

My requested locale came from the _locale parameter of my route (API calls)
Maybe there's a better way to do that...

Thanks !

PR #5601, opened 2016-07-25
Make locales switchable by user & dependent on current channel

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | based on #5597 
| License         | MIT


src/Sylius/Behat/Context/Ui/Shop/LocaleContext.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    /**
+     * @Then I should not be able to shop using the :localeName locale
+     */
+    public function iShouldNotBeAbleToShopUsingTheLocale($localeName)
+    {
+        $this->homePage->open();
+
+        if (in_array($localeName, $this->homePage->getAvailableLocales(), true)) {
```
Arminek: Assert::true?

pamil: There's no `Assert::notOneOf` and `Assert::true` / `Assert::false` seems weird.

src/Sylius/Bundle/CoreBundle/Resources/config/context.xml
```diff
@@ -62,5 +62,33 @@
...
+            <argument type="service" id="sylius.context.channel" />
+            <argument>%locale%</argument>
+        </service>
+
+        <service id="sylius.storage.locale" class="Sylius\Component\Core\Locale\LocaleStorage">
+            <argument type="service" id="sylius.storage.session" />
+        </service>
+
+        <service id="sylius.context.locale.core" class="Sylius\Component\Locale\Context\CompositeLocaleContext" decorates="sylius.context.locale">
+            <call method="addContext">
```
Arminek: How about making an compiler pass for this. WDYT?

pamil: Might do, but I'd rather not make this PR bigger.

pjedrzejewski:
Thank you Kamil! 👍

PR #5602, opened 2016-07-25
Fix typo in shipping method fixture

ylastapis:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Related tickets | - |
| License | MIT |

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ShippingMethodExampleFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ShippingMethodExampleFactory.php
@@ -82,9 +82,9 @@ final class ShippingMethodExampleFactory implements ExampleFactoryInterface
                 ->setDefault('zone', LazyOption::randomOne($zoneRepository))
                 ->setAllowedTypes('zone', ['null', 'string', ZoneInterface::class])
                 ->setNormalizer('zone', LazyOption::findOneBy($zoneRepository, 'code'))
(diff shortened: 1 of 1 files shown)


ylastapis:
Fixes the following error when trying to set a shipping_category on a shipping_method

[Symfony\Component\Config\Definition\Exception\InvalidConfigurationException]  
  Unrecognized option "shipping_category" under "shipping_method.custom.normal"

pamil:
Name's correct, `Sylius\Component\Shipping\Model\ShippingMethodInterface` has methods `getCategory()` and `setCategory()`.

ylastapis:
You're right, but there is still a confusion between "category" from the fixture & "shipping_category" from the factory. I'm just discovering the brand new fixture system :)

With the fix, following code now works:
```
                shipping_method:
                    options:
                        custom:
                            normal:
                                description: normal
                                name: normal
                                code: normal
                                enabled: true
                                zone: EUROPE
                                category: regular
                                calculator:
                                    type: flat_rate
                                    configuration:
                                        amount: 500
                                        min_delay: 2
                                        max_delay: 4
```

pjedrzejewski:
I was in favor of this change too, but I recall there was some reason to keep it that way - @pamil?

pamil:
@pjedrzejewski I don't really know the reason - https://github.com/Sylius/Sylius/pull/5406#discussion_r69284245 :)

Can we do both the fixtures renaming and model renaming as described in #5384?

PR #5621, opened 2016-07-28
Remove duplicated keys

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

Just as a note, there are more keys to delete, when we would remove a WebBundle.


michalmarcinkowski:
Thanks Łukasz! 👍

PR #5628, opened 2016-07-28
[ThemeBundle] fixed assets re-installation

takeit:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

This PR fixes an issue when you want to update your assets once they were already installed at least one time. Let's say I have installed assets, I edit one of the asset files in my theme and run `sylius:theme:assets:install` command again - the updated asset file won't be copied over to the `web` directory, thus I still get the old version.

### DIFF
--- a/src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstaller.php
+++ b/src/Sylius/Bundle/ThemeBundle/Asset/Installer/AssetsInstaller.php
@@ -141,7 +141,7 @@ class AssetsInstaller implements AssetsInstallerInterface
             $targetFile = $targetDir.'/'.$originFile->getRelativePathname();
             $targetFile = $this->pathResolver->resolve($targetFile, $theme);
 
-            if (file_exists($targetFile)) {
+            if (file_exists($targetFile) && AssetsInstallerInterface::HARD_COPY !== $symlinkMask) {
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/ThemeBundle/Tests/Functional/Bundle/TestBundle/Resources/public/theme_asset.txt
```diff
@@ -1 +1 @@
-Vanilla asset
+Theme asset
```
pamil: ?

pamil: These test were passing, "Vanilla asset" is used here to detect issues with theme assets loading. If we're using a theme and "Vanilla asset" shows, then something is wrong with the codebase, not with tests.

pamil: Or just tests were wrong and in fact only the `RELATIVE_SYMLINK` was used so they were passing because of that. If so, `AssetsInstaller` needs a fix.

takeit: hey, I should have made that WIP :) and yeah you are, right.

pjedrzejewski:
Thanks Rafał! :)

PR #5632, opened 2016-07-28
[API] Fix api authenticator

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

BC Break included. Before, access token was available under `/oauth/v2/token` path. Now, it is available under `/api/oauth/v2/token`.


pjedrzejewski:
Thank you Łukasz!

PR #5647, opened 2016-07-29
[Core] Rewrite ChannelBasedCurrencyProvider

pamil:
There's no reason not to use `map` on Collection instead of `array_map` on array (nice chaining included).


pjedrzejewski:
👍

src/Sylius/Component/Core/Provider/ChannelBasedCurrencyProvider.php
```diff
@@ -45,18 +45,16 @@ public function getAvailableCurrenciesCodes()
...
-            $currencies = $channel
+            return $channel
                 ->getCurrencies()
                 ->filter(function (CurrencyInterface $currency) {
                     return $currency->isEnabled();
                 })
+                ->map(function (CurrencyInterface $currency) {
+                    return $currency->getCode();
+                })
                 ->toArray()
```
michalmarcinkowski: Debugging such chains is not easy 🏃💥 💣

PR #5654, opened 2016-07-29
[Locale] Add request based locale context

### LINKED ISSUE #5649: Resources or not shown in the correct locale through the api
Lets say I want to display all taxon in locale nl_BE through the api.
The path would be /api/nl_BE/taxons.
Currently this is not possible because of the locale used for the resources comes from the LocaleContext which uses the session and our current request. So it always shows the default locale. 

What is the best approach to fix this? 

One option could be to create a custom LocaleContext and check if the current route matches as an api request. Then we always use the locale provide by the request and discard our session storage, cause  an api should be sessionless as far as i know.

Any thoughts

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Related tickets | fixes #5649
| License         | MIT


pjedrzejewski:
Thank you Kamil! 👍

PR #5655, opened 2016-07-29
Remove WebBundle

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | -
| License         | MIT

Not used anymore, but still annoying during some refactorings.


loic425:
Is that normal that this PR has commits from #5648 ?

src/Sylius/Bundle/PayumBundle/Resources/views/layout.html.twig
```diff
@@ -1,4 +1,4 @@
-{% extends 'SyliusWebBundle:Frontend:layout.html.twig' %}
+{% extends 'SyliusShopBundle::layout.html.twig' %}
```
michalmarcinkowski: We should not couple PayumBundle with ShopBundle

pamil: :+1: - nevertheless I'm just moving the PayumBundle coupling from WebBundle to ShopBundle, it should be fixed in another PR.

pjedrzejewski:
Thank you Kamil! :)

PR #5662, opened 2016-07-30
Issue#5561: Taxon root is not being checked

### LINKED ISSUE #5561: Taxon root is not being checked in ProductRepository
In [ProductRepository](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php#L41) inside method `createByTaxonPaginator` tree root condition is missing. Because of this, products which are not assigned to the given taxon will still be fetched from db.

I believe this method should be:
```
    public function createByTaxonPaginator(TaxonInterface $taxon, array $criteria = [])
    {
        $queryBuilder = $this->createQueryBuilder('o');
        $queryBuilder
            ->innerJoin('o.taxons', 'taxon')
            ->andWhere($queryBuilder->expr()->orX(
                'taxon = :taxon',
                ':left < taxon.left AND taxon.right < :right'
            ))
            ->andWhere('taxon.root = :tree_root')
            ->setParameter('tree_root', $taxon->isRoot() ? $taxon->getId() : $taxon->getRoot()->getId())
            ->setParameter('taxon', $taxon)
            ->setParameter('left', $taxon->getLeft())
            ->setParameter('right', $taxon->getRight())
        ;

        $this->applyCriteria($queryBuilder, $criteria);

        return $this->getPaginator($queryBuilder);
    }
```
Tell me If I got this wrong, but as far as I can see tree_left and tree_right condition will match taxons even they are not into the given taxon tree.

Thanks!

vcraescu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #5561 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
@@ -29,13 +29,17 @@ class ProductRepository extends BaseProductRepository implements ProductReposito
      */
     public function createByTaxonPaginator(TaxonInterface $taxon, array $criteria = [])
     {
+        $root = $taxon->isRoot() ? $taxon : $taxon->getRoot();
+
         $queryBuilder = $this->createQueryBuilder('o');
         $queryBuilder
             ->innerJoin('o.taxons', 'taxon')
+            ->andWhere($queryBuilder->expr()->eq('o.root', ':root'))
             ->andWhere($queryBuilder->expr()->orX(
                 'taxon = :taxon',
                 ':left < taxon.left AND taxon.right < :right'
             ))
+            ->setParameter('root', $root)
             ->setParameter('taxon', $taxon)
             ->setParameter('left', $taxon->getLeft())
             ->setParameter('right', $taxon->getRight())


pjedrzejewski:
Thank you Viorel! :)

vcraescu:
uw :)

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
```diff
@@ -29,13 +29,17 @@ class ProductRepository extends BaseProductRepository implements ProductReposito
      */
     public function createByTaxonPaginator(TaxonInterface $taxon, array $criteria = [])
     {
+        $root = $taxon->isRoot() ? $taxon : $taxon->getRoot();
+
         $queryBuilder = $this->createQueryBuilder('o');
         $queryBuilder
             ->innerJoin('o.taxons', 'taxon')
+            ->andWhere($queryBuilder->expr()->eq('o.root', ':root'))
```
pamil: I may be missing something, but I cannot find `root` field on any Product mapping.

pamil: Tests may pass as this method is only used in `OrmFinder` in `SearchBundle`.

vcraescu: that's correct. it should be 'taxon' instead of 'o'. my fault!

pamil: Anyway, for nested set tree, left / right conditions should be enough -  https://en.wikipedia.org/wiki/Nested_set_model. Moreover, we can replace this:
```
->andWhere($queryBuilder->expr()->orX(
    'taxon = :taxon',
    ':left < taxon.left AND taxon.right < :right'
))
```

with:

```
->andWhere(':left <= taxon.left AND taxon.right <= :right')
```

But there should be some automatic tests made before (if we need this method at all).

vcraescu: As far as I can tell they are not enough. You can have left and right value which match the condition from above but the node actually have a different root. I came up with this fix because in sylius admin we found some products being listed under a taxon even they were never assigned to it. Of course it's not easy to replicate this issue but it happens.

pamil: If parent's left is always smaller than child's left and parent's right is always bigger than child's left, there is no way that this issue may exist (assuming that nested set tree implementation is correct). 

Which version do you use? I remember that there were some issues when we used both nested set tree and softdeletable back in the days. (This may still exist in newer versions if using the same database then and now).

vcraescu: Tested on sylius 0.18. Also tested on fresh sylius instance.

vcraescu: Anyway, I created a new pr for the error you spot.
https://github.com/Sylius/Sylius/pull/5676/files

vcraescu: [TaxonRepository](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php#L32)

Isn't quite the same thing?

PR #5682, opened 2016-08-02
Fix setup command issues

gorkalaucirica:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | no
| License         | MIT

I faced the following issues while running `sylius:install` command on `master`:
- Duplicate entry exception thrown while runnig `sylius:install:setup` when default currency was added before by the fixtures
- Field cannot be null exception when creating channel without default currency

Steps to reproduce:
- Clone the repo
- Follow [quick installation guide](https://github.com/Sylius/Sylius#quick-installation)

### DIFF
--- a/src/Sylius/Bundle/InstallerBundle/Command/SetupCommand.php
+++ b/src/Sylius/Bundle/InstallerBundle/Command/SetupCommand.php
@@ -157,10 +157,17 @@ EOT
      */
     protected function setupCurrency(InputInterface $input, OutputInterface $output)
     {
+        $currencyRepository = $this->get('sylius.repository.currency');
         $currencyManager = $this->get('sylius.manager.currency');
         $currencyFactory = $this->get('sylius.factory.currency');
 
         $code = trim($this->getContainer()->getParameter('currency'));
+        $this->currency = $currencyRepository->findOneByCode($code);
+
+        if(null !== $this->currency) {
+            return;
+        }
+
         $name = Intl::getCurrencyBundle()->getCurrencyName($code);
         $output->writeln(sprintf('Adding <info>%s</info> currency.', $name));
 
@@ -192,6 +199,7 @@ EOT
         $channel->setName('Default');
         $channel->setTaxCalculationStrategy('order_items_based');
 
+        $channel->setDefaultCurrency($this->currency);
         $channel->addCurrency($this->currency);
         $channel->addLocale($this->locale);
 


src/Sylius/Bundle/InstallerBundle/Command/SetupCommand.php
```diff
@@ -157,10 +157,17 @@ protected function setupLocale(InputInterface $input, OutputInterface $output)
      */
     protected function setupCurrency(InputInterface $input, OutputInterface $output)
     {
+        $currencyRepository = $this->get('sylius.repository.currency');
         $currencyManager = $this->get('sylius.manager.currency');
         $currencyFactory = $this->get('sylius.factory.currency');
 
         $code = trim($this->getContainer()->getParameter('currency'));
+        $this->currency = $currencyRepository->findOneByCode($code);
```
pjedrzejewski: Why not just a variable?

gorkalaucirica: `$this->currency` is used later to setup the channel in `setupChannel()`. Is that what you mean?

lchrusciel:
I have given a 👍 for the fix, but I have one problem with this command. We have used both `$this->currency` and `$this->locale` to set something on a channel, but this variables are set up during some other functions call. This way, we have to call one function before other to have a successful result, otherwise some error can occur. The problem is even bigger, because all of this function can be overridden, as they are defined with `protected` visibility. So, not only changing an `execute` function can create some problem, but also changing a logic of `setupLocale` or `setupCurrency`. Not saying, that both variables which are used by `setupChannel` are private, and if don't call a a parent on `setupLocale` or `setupCurrency` when override these classes would also broke a command. We need a refactoring here for sure.

gorkalaucirica:
@lchrusciel  What if we get from the repo the default currency and locale (taking them from parameters.yml) and if one or both of them haven't been created an exception will be thrown telling that those need to be created first.

There is no way to call only `setupChannel()` with a command directly and if `execute()` is not modified by a child class then we shouldn't have any problem. Otherwise the developer will get an exception that reminds that the order has to be kept if the `execute()` method has been modified.

pjedrzejewski:
Fixed in https://github.com/Sylius/Sylius/pull/5732, thanks for reporting Gorka!

PR #5689, opened 2016-08-03
[Shop][Payment] Prevent selecting payment method not assigned to channel

tuka217:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | -
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/PaymentMethodRepository.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/PaymentMethodRepository.php
@@ -12,8 +12,10 @@
 namespace Sylius\Bundle\CoreBundle\Doctrine\ORM;
 
 use Sylius\Bundle\PaymentBundle\Doctrine\ORM\PaymentMethodRepository as BasePaymentMethodRepository;
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Core\Repository\PaymentMethodRepositoryInterface;
 
-class PaymentMethodRepository extends BasePaymentMethodRepository
+class PaymentMethodRepository extends BasePaymentMethodRepository implements PaymentMethodRepositoryInterface
 {
     /**
      * {@inheritdoc}
@@ -24,6 +26,20 @@ class PaymentMethodRepository extends BasePaymentMethodRepository
             ->createQueryBuilder('o')
             ->addSelect('translation')
             ->leftJoin('o.translations', 'translation')
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function findEnabledForChannel(ChannelInterface $channel)
+    {
+        return $this->createQueryBuilder('o')
+            ->where('o.enabled = true')
+            ->andWhere('o IN (:paymentMethodsInChannel)')
+            ->setParameter('paymentMethodsInChannel', $channel->getPaymentMethods()->toArray())
+            ->getQuery()
+            ->getResult()
         ;
     }
 }
--- a/src/Sylius/Component/Core/Repository/PaymentMethodRepositoryInterface.php
+++ b/src/Sylius/Component/Core/Repository/PaymentMethodRepositoryInterface.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Core\Repository;
+
+use Doctrine\ORM\QueryBuilder;
+use Sylius\Component\Core\Model\ChannelInterface;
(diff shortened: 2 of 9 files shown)


src/Sylius/Bundle/CoreBundle/Doctrine/ORM/PaymentMethodRepository.php
```diff
@@ -26,4 +28,18 @@ public function createListQueryBuilder()
             ->leftJoin('o.translations', 'translation')
         ;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function findEnabledForChannel(ChannelInterface $channel)
```
pamil: Why won't we just use `$channel->getPaymentMethods()` collection? It implements `Doctrine\Common\Collections\Selectable`, so:

```php
$channel->getPaymentsMethods()->matching(new Criteria(new Comparison('enabled', Comparison::EQ, true))
```

Should work even faster (we won't need to hydrate all the payment methods only to pass them to another query).

pamil: Then we would not need this custom repository, we can use it directly in the resolver or implement on the model - see http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/working-with-associations.html#filtering-collections.

src/Sylius/Component/Core/Resolver/ChannelBasedPaymentMethodsResolver.php
```diff
@@ -0,0 +1,58 @@
...
+    {
+        $this->paymentMethodRepository = $paymentMethodRepository;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSupportedMethods(BasePaymentInterface $payment)
+    {
+        Assert::true($this->supports($payment));
```
pamil: Custom message may be nicer than "expected true got false" :)

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/PaymentMethodRepository.php
tuka217: Method 'mathing' is allowed only for `ArrayCollection`. We have here an instance of `Collection` class.

pamil: Of course the typehint is oversimplified now, it's either `ArrayCollection` (if created manually) or `PersistentCollection` (if hydrated by Doctrine), but both of them implements `Collection` and `Selectable`.

tuka217: Yups :)

src/Sylius/Component/Core/Repository/PaymentMethodRepositoryInterface.php
```diff
@@ -0,0 +1,35 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Core\Repository;
+
+use Doctrine\ORM\QueryBuilder;
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+use Sylius\Component\Payment\Repository\PaymentMethodRepositoryInterface as BasePaymentMethodRepositoryInterface;
+/**
```
lchrusciel: Missing blank line

src/Sylius/Component/Core/Resolver/ChannelBasedPaymentMethodsResolver.php
```diff
@@ -0,0 +1,58 @@
...
+use Sylius\Component\Core\Model\PaymentInterface;
+use Sylius\Component\Core\Repository\PaymentMethodRepositoryInterface;
+use Sylius\Component\Payment\Model\PaymentInterface as BasePaymentInterface;
+use Sylius\Component\Payment\Resolver\MethodsResolverInterface;
+use Webmozart\Assert\Assert;
+
+/**
+ * @author Anna Walasek <anna.walasek@lakion.com>
+ */
+class ChannelBasedPaymentMethodsResolver implements MethodsResolverInterface
```
lchrusciel: final

src/Sylius/Component/Core/Resolver/ChannelBasedPaymentMethodsResolver.php
```diff
@@ -0,0 +1,58 @@
...
+{
+    /**
+     * @var PaymentMethodRepositoryInterface
+     */
+    protected $paymentMethodRepository;
```
lchrusciel: private

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/PaymentMethodRepository.php
tuka217: In current implementation it cannot be done in this way.

pjedrzejewski:
Thank you Ania! :)

PR #5692, opened 2016-08-03
Allow null or string on default currency and locale in ChannelExampleFactory

gorkalaucirica:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Related tickets | no |
| License | MIT |

Channel fixtures don't support defining default currency or locale. It throws the following exception when you try to define a current locale for example:

`The option "default_locale" with value "es_ES" is expected to be of type "Sylius\Component\Locale\Model\LocaleInterface", but is of type "string".`

**Steps to reproduce**
- Clone the master branch of sylius/sylius
- Copy [this gist](https://gist.github.com/gorkalaucirica/b50edfbf767123cb370ed05f4af3eddc) and include it in `app/config/config.yml`. The config is the same as found in [Core Bundle fixtures.yml](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml) but adding `default_locale` in [line 51](https://gist.github.com/gorkalaucirica/b50edfbf767123cb370ed05f4af3eddc#file-fixtures-yml-L51). 
- Run the following:

``` bash
$ php app/console sylius:install:database
$ php app/console sylius:fixtures:load this_fails
```

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ChannelExampleFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ChannelExampleFactory.php
@@ -81,7 +81,7 @@ final class ChannelExampleFactory implements ExampleFactoryInterface
                 ->setDefault('default_locale', function (Options $options) {
                     return $this->faker->randomElement($options['locales']);
                 })
-                ->setAllowedTypes('default_locale', ['null', 'string', LocaleInterface::class])
+                ->setAllowedTypes('default_locale', ['string', LocaleInterface::class])
                 ->setNormalizer('default_locale', LazyOption::findOneBy($localeRepository, 'code'))
                 ->setDefault('locales', LazyOption::all($localeRepository))
                 ->setAllowedTypes('locales', 'array')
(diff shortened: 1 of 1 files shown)


pamil:
LGTM :+1:, except that we should not accept `null` there.

pjedrzejewski:
@gorkalaucirica Could you adjust? Thanks!

gorkalaucirica:
@pjedrzejewski I've removed the null as allowed type. This was hiding another issue that makes fixtures crash. `$options['locales']` and `$options['currencies']` need to be defined always in channel fixtures, otherwise, `$faker->randomElement()` fails to select one, as the options array hasn`t been filled with default values at the time that [default_locale](https://github.com/gorkalaucirica/Sylius/blob/4b8affd350a6624897cfe3eb94081715b8510c47/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ChannelExampleFactory.php#L82) and [default_currency](https://github.com/gorkalaucirica/Sylius/blob/4b8affd350a6624897cfe3eb94081715b8510c47/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ChannelExampleFactory.php#L90) are being set.

michalmarcinkowski:
Fixed already.

Issue #4883, opened 2016-04-26
TranslatableFactory design


peteward:
We had to create a couple of `TranslatableFactory`'s this week. 

In both cases we wanted our own create methods - `createForParent()` or `createForType()` etc.

While it's very possible to do this, it's not very clean and I'm wondering if the design is right here:

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Resource/Factory/TranslatableFactory.php

`TranslatableFactory` decorates the underlying component `Factory`. This works fine while you only have a `createNew()` method, but as soon as you have another create method you need to override both, or as we did just use a single factory but we end up copy&pasting the initialisation of the locales.

Thoughts?

peteward:
It feels like an inverse decorator pattern because your own subject methods are on the inside of the decoration.

pamil:
I'll open a RFC soon with a possible solution that brings some flexibility here. The main idea is to use services names instead of FQCNs in resources' `repository`, `factory` and `forms` configuration options and aliasing them (or even entirely delete some of the configuration options that receive FQCNs), but it needs some more thought and a proof of concept to be made.

For now, the following container configuration should work:
```xml
<service id="acme.factory.model" class="Acme/Factory/ModelFactory" decorates="sylius.factory.model">
    <argument type="service" id="acme.factory.model.inner" />
</service>
```

Then your custom factory may use decorated `sylius.factory.model` factory in `createFor*` methods and will receive empty model with initialised locales.

peteward:
Unless the repository that you want to override is a sylius one then to do this you would need a CompilerPass to override the Compiler pass to decorate the decorated service...

I understand the separation between Factory component and Bundle, but we'll just create our base TranslatableFactory here to save the pain and complexity.

Issue #4339, opened 2016-02-28
AdjustmentFactory service decoration - why? 


peteward:
Some recent changes have added the following config:

```
<parameter key="sylius.custom_factory.adjustment.class">Sylius\Component\Order\Factory\AdjustmentFactory</parameter>

<service id="sylius.custom_factory.adjustment" class="%sylius.custom_factory.adjustment.class%" decorates="sylius.factory.adjustment" public="false">
      <argument type="service" id="sylius.custom_factory.adjustment.inner" />
 </service>
```

I'm just wondering what this achieves? From what I understand, service decoration allows you access to the original service, but I don't see the use-case in getting the default Resource Factory?

The downside of the above changes means that the following doesn't work anymore:

```
sylius_order:
    resources:
        adjustment:
            classes:
                model: Reiss\OrderBundle\Entity\Adjustment
                factory: Reiss\OrderBundle\Entity\Factory\AdjustmentFactory
```

As the decorating service definition is overriding my override. 

Sylius config, while wonderfully powerful isn't exactly for the coding novice. When you start having to worry about alternative methods of overriding the services which are overriding your overridden services then you're really down the rabbit-hole...

The solution wasn't exactly complex, just feels ugly to have an inconsistence here unless it's really required:

```
parameters:
    sylius.custom_factory.adjustment.class: Reiss\OrderBundle\Entity\Factory\AdjustmentFactory
```

peteward:
on closer inspection it could be because `Sylius\Component\Order\Factory\AdjustmentFactory` is wrapping the Resource Factory instead of extending it?  Again, I'm not quite sure why.

peteward:
Hey @pjedrzejewski,

Can you give me a quick comment on the thoughts behind this one?

I've got a similar issue with extending `PaymentFactory` where ideally i should be extending the decorated version but in the `SyliusPaymentExtension` I don't seem to have access to the original service to inject into my extended version...:

```
        $factoryDefinition = new Definition(Factory::class, [new Parameter('sylius.model.payment.class')]);
        $paymentFactoryClass = $container->getParameter('sylius.factory.payment.class');
        $decoratedPaymentFactoryDefinition = new Definition($paymentFactoryClass, [$factoryDefinition]);

        $container->setDefinition('sylius.factory.payment', $decoratedPaymentFactoryDefinition);
```

Is my best bet to duplicate this Extension code in our app for now if this pattern is the way we're heading with decoration rather than extension?

[...]

Issue #4226, opened 2016-02-19
[RFC] [DX] Consistent factory decoration


lchrusciel:
Factories has been introduced to Sylius with refactoring of ResourceBundle. It was a must-have feature and it is great that finally we have decoupled this logic. But factories overriding isn't as easy as other resource-related stuff. We have to change theirs definition in extensions of each bundle, because we decide to use decorator pattern. It makes spec much cleaner, but there is one problem. We have a few different convention around Sylius how to override them. And new ones are coming. Right now I have noticed following implementations:
1. [Taxonomy](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/TaxonomyBundle/DependencyInjection/SyliusTaxonomyExtension.php#L49)
2. [Attribute](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/AttributeBundle/DependencyInjection/Compiler/RegisterAttributeFactoryPass.php)
And another 2 in PRs:
1. [Payment] Adding PaymentFactory #4214 
2. [Promotion] Implement custom promotion action factory #4223

When it is obvious, that attribute needs a different approach, because of it genericness, most of bundles extend factories pretty straightforward. I just want to hear your opinion which implementation is the most readable for you and which one should we use.

michalmarcinkowski:
For simple decoration in extensions I see two options there:

1) Creating two new definitions:
```
$baseFactoryDefinition = new Definition(Factory::class, [new Parameter('sylius.model.promotion_action.class')]);
$promotionActionFactoryClass = $container->getParameter('sylius.factory.promotion_action.class');
$decoratedPromotionActionFactoryDefinition = new Definition($promotionActionFactoryClass, [$baseFactoryDefinition]);
$container->setDefinition('sylius.factory.promotion_action', $decoratedPromotionActionFactoryDefinition);

$container->setDefinition('sylius.factory.promotion_action', $decoratedPromotionActionFactoryDefinition);
```
2) Changing class of base definition and creating only one new definition:
```
$couponFactoryDefinition = $container->getDefinition('sylius.factory.promotion_coupon');
$couponFactoryClass = $couponFactoryDefinition->getClass();
$couponFactoryDefinition->setClass(Factory::class);
$decoratedCouponFactoryDefinition = new Definition($couponFactoryClass);

$container->setDefinition('sylius.factory.promotion_coupon', $decoratedCouponFactoryDefinition);
```

Which approach seems better for you guys?

[...]

PR #5701, opened 2016-08-04
[Factory] [RFC] Make factories final

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

It always a good idea to make some services final. Especially, when we decide to use a decorator pattern to override a factories. It does not resolve a problem with factories extending, they still need some refactoring.

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Factory/ZoneFactory.php
+++ b/src/Sylius/Bundle/AddressingBundle/Factory/ZoneFactory.php
@@ -17,7 +17,7 @@ use Sylius\Component\Resource\Factory\FactoryInterface;
 /**
  * @author Jan Góralski <jan.goralski@lakion.com>
  */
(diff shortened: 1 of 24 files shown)


michalmarcinkowski:
Let's decide on https://github.com/Sylius/Sylius/issues/4226 and resolve exiting issues (https://github.com/Sylius/Sylius/issues/4339#issuecomment-190593174, https://github.com/Sylius/Sylius/issues/4883#issuecomment-214784561) before we merge that. Nevertheless 👍

patrick-mcdougle:
Do you think it makes sense to leave a comment about how we use decoration for modifying/adding factory behavior to all these classes as well, or do you think that's overkill?

PR #5703, opened 2016-08-04
[Shop][Admin] Include tax rate in product price

tuka217:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | -
| License         | MIT

### DIFF
--- a/features/taxation/applying_taxes/applying_correct_taxes_for_product_with_tax_rate_included_in_price.feature
+++ b/features/taxation/applying_taxes/applying_correct_taxes_for_product_with_tax_rate_included_in_price.feature
@@ -1,0 +1,20 @@
+@applying_taxes
+Feature: Apply correct taxes for items with tax rate included in price
+    In order to pay proper amount when buying goods with tax rate included in price
+    As a Visitor
+    I want to have correct taxes applied to my order
+
+    Background:
+        Given the store operates on a single channel in "France"
+        And default tax zone is "FR"
+        And the store has included in price "VAT" tax rate of 23% for "Clothes" within "FR" zone
+        And the store has a product "PHP T-Shirt" priced at "€100.00"
+        And it belongs to "Clothes" tax category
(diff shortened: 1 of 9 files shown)


features/taxation/applying_taxes/applying_correct_taxes_for_product_with_tax_rate_included_in_price.feature
```diff
@@ -0,0 +1,18 @@
...
+        And default tax zone is "FR"
+        And the store has "VAT" tax rate of 23% for "Clothes" within "FR" zone included in product price
+        And the store has a product "PHP T-Shirt" priced at "€100.00"
+        And it belongs to "Clothes" tax category
+
+    @todo
+    Scenario: Proper taxes for taxed product
+        When I add product "PHP T-Shirt" to the cart
+        Then my cart total should be "€123.00"
+        And my product price should be "€123.00"
```
gabiudrescu: if the VAT is included in the €100 price of the product, then why is cart total €123?

imho, it should be 100€. 

also, what do you think about adding an explicit "And" line stating the total cart amount without prices (so that developers or users can explicitly understand the logic of adding taxes)?

For example, I'd add:
```
And my cart total without taxes should be €81,30
```

gabiudrescu: from what I see in Sylius/features/taxation/applying_taxes/applying_correct_taxes_on_visitor_cart.feature, there's a method similar to that:
```
And my cart taxes should be "€18.7"
```

tuka217: It will be fixed.

pjedrzejewski:
Thank you Ania! :)

PR #5717, opened 2016-08-05
[Core][Product] Fix query with true value instead of 1

Niiko:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

Fix findEnabledByTaxonCodeAndChannel in ProductRepository with true value instead of 1, for PostgreSQL compatibility.


michalmarcinkowski:
Thanks Nicolas! 👍

Issue #5574, opened 2016-07-22
ResourceController / SingleResourceProvider Exception Messages.


dantleech:
## Record not found

The ResourceController calls `findOr404` and throws a `HttpNotFound` exception with _no message_.

A precise message could be added if the `SingleResourceProvider` threw an exception when the resource could not be found. it could then be wrapped by the 404 exception.

The different messages could be:

```
Could not find record using repository method "%s::%s" with arguments :"%s"
```

```
Could not find record using repository "%s" with ID "%s"
```

```
Could not find record using repository "%s" and criteria "%s"
```

Is there a case where an exception would not be appropriate - i.e. when `null` is anticipated and acceptable?
## Acess Denied

Currently when access is denied, a `AccessDeniedException` is again thrown, but with no message.

The message could be:

```
User must have permission "%s" to view this page
```

dantleech:
Updated to include adding message to the `AccessDenied` exception too.

PR #5724, opened 2016-08-07
[WIP] Not-found handling for single-resource-provider

dantleech:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no (lets find out!!) |
| Related tickets | #5574 |
| License | MIT |

This PR causes the SingleResourceProvider to throw exception messages when the resource cannot be found, the exception messages indicate how the resolution of the resource was attempted.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/SingleResourceProvider.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/SingleResourceProvider.php
@@ -26,6 +26,13 @@ class SingleResourceProvider implements SingleResourceProviderInterface
         if (null !== $repositoryMethod = $requestConfiguration->getRepositoryMethod()) {
             $callable = [$repository, $repositoryMethod];
 
+            if (!method_exists($repository, $repositoryMethod)) {
+                throw new \InvalidArgumentException(sprintf(
+                    'Method "%s" does not exist on repository of class "%s"',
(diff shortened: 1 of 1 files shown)


dantleech:
If the tests generally pass and people have no objections I will spec this out.

src/Sylius/Bundle/ResourceBundle/Controller/SingleResourceProvider.php
```diff
@@ -26,14 +26,31 @@ public function get(RequestConfiguration $requestConfiguration, RepositoryInterf
         if (null !== $repositoryMethod = $requestConfiguration->getRepositoryMethod()) {
             $callable = [$repository, $repositoryMethod];
 
+            if (!method_exists($repository, $repositoryMethod)) {
```
lchrusciel: If we are using an `InvalidArgumentException` anyway, won't it be cleaner to use [webmozart/assert](https://github.com/webmozart/assert)?

dantleech: I don't see the added value, but don't mind changing.

lchrusciel: I just find it more readable. Thats all.

PR #5728, opened 2016-08-08
[API][Zone] Zone API tests

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets |
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/routing/zone.yml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/routing/zone.yml
@@ -10,6 +10,12 @@ sylius_api_zone_index:
             paginate: $limit
             sortable: true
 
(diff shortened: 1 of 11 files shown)


lchrusciel:
👍 Can be added in separate PR, but we can think about some custom validation, like what would happen if we sent some strange type during creation.

pjedrzejewski:
Thanks Mateusz! We need to remove the ``type`` from the endpoints in the next iteration. It should use a form type that has "type" field enabled. Then the type can be passed as part of the request content and not in the url.

Zales0123:
@pjedrzejewski I've thought about it, but it also require some changes about zone members collection (now it takes ``'zone_type' => $zoneType`` option). Nevertheless - 👍

PR #5730, opened 2016-08-08
[Documentation] Concepts: Checkout

TheMadeleine:
| Q               | A
| --------------- | ---
| Doc fix?        | no
| New docs?    | yes
| BC breaks?      | no
| License         | MIT

### DIFF
--- a/docs/book/checkout.rst
+++ b/docs/book/checkout.rst
@@ -92,7 +92,8 @@ The typical **Address** consists of: country, city, street and postcode - to ass
      $order->setBillingAddress($address);
 
 Having the **Customer** and the **Address** set you can apply a state transition to your order.
-Get the StateMachine for the Order via the StateMachineFactory with a proper schema, and apply a transition.
+Get the StateMachine for the Order via the StateMachineFactory with a proper schema, and apply a transition
+and of course flush your order after that via the manager.
 
 .. code-block:: php
 
@@ -101,16 +102,21 @@ Get the StateMachine for the Order via the StateMachineFactory with a proper sch
     $stateMachine = $stateMachineFactory->get($order, OrderCheckoutTransitions::GRAPH)
     $stateMachine->apply(OrderCheckoutTransitions::TRANSITION_ADDRESS);
 
+    $this->container->get('sylius.manager.order')->flush();
+
 **What happens with the transition?**
 
 The method ``process($order)`` of the ``Sylius\Component\Core\OrderProcessing\OrderProcessor`` is run.
 It is responsible for creating new **Shipments** for each OrderItemUnit of your order and a new **Payment** if they do not exist yet.
 Therefore this transition is preparing the order for the two next steps of checkout.
+Also the method ``recalculate($order)`` of the ``Sylius\Component\Core\OrderProcessing\OrderRecalculator`` is run,
+that takes care of the **shipping charges** for the default ShippingMethod, **promotions** that depend on the chosen address
+and **taxes** that depend on this address.
 
 Selecting shipping
 ~~~~~~~~~~~~~~~~~~
 
-It is a step where the customer selects the way their order will be shipped to him.
+It is a step where the customer selects the way their order will be shipped to them.
 Basing on the ShippingMethods configured in the system the options for the Customer are provided together with their prices.
 
 +---------------------------------------+--------------------------------------------------+
@@ -118,6 +124,42 @@ Basing on the ShippingMethods configured in the system the options for the Custo
 +---------------------------------------+--------------------------------------------------+
 | ``addressed``-> ``shipping_selected`` | ``SyliusShopBundle:Checkout:shipping.html.twig`` |
 +---------------------------------------+--------------------------------------------------+
+
+How to perform the Selecting shipping Step programmatically?
+''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+Before approaching this step be sure that you Order is in the ``addressed`` state. In this state your order
+will already have a default ShippingMethod assigned, but in this step you can change it and have everything recalculated automatically.
+
+Firstly either create new (see how in the `Shipments concept </book/shipments`) or retrieve a **ShippingMethod**
+from the repository to assign it to your order's shipment created defaultly in the addressing step.
+
+.. code-block:: php
+
+    // Let's assume you have a method with code 'DHL' that has everything set properly
+    $shippingMethod = $this->container->get('sylius.repository.shipping_method')->findOneByCode('DHL');
+
+    foreach ($order->getShipments() as $shipment) {
+        $shipment->setMethod($shippingMethod);
+    }
+
+After that get the StateMachine for the Order via the StateMachineFactory with a proper schema,
+and apply a proper transition and flush the order via manager.
+
+.. code-block:: php
+
+    $stateMachineFactory = $this->container->get('sm.factory');
+
+    $stateMachine = $stateMachineFactory->get($order, OrderCheckoutTransitions::GRAPH)
+    $stateMachine->apply(OrderCheckoutTransitions::TRANSITION_SELECT_SHIPPING);
+
+    $this->container->get('sylius.manager.order')->flush();
+
+**What happens with the transition?**
+
+The method ``recalculate($order)`` of the ``Sylius\Component\Core\OrderProcessing\OrderRecalculator`` is run.
+This method is responsible for: controlling the **shipping charges** which depend on the chosen ShippingMethod,
+controlling the **promotions** that depend on the shipping method.
 
 Selecting payment
 ~~~~~~~~~~~~~~~~~
--- a/docs/book/checkout.rst
+++ b/docs/book/checkout.rst
@@ -4,18 +4,145 @@
 Checkout
 ========
 
-**Checkout** is a process that begins when the Customer decides to finish his shopping and transform his **Cart** into an **Order**.
-
-Checkout in Sylius is divided into steps:
-
-* addressing - where the customer provides both shipping and billing addresses,
-* selecting shipping - when the customer seletc the way his order will be shipped to him,
-* selecting payment - where the customer chooses how is he willing to pay for his order,
-* finalizing - when the customer gets an order summary
+**Checkout** is a process that begins when the Customer decides to finish their shopping and transform their **Cart** into an **Order**.
 
 Checkout State Machine
 ----------------------
 
-The Order Checkout state machine has 5 states available: ``cart, addressed, shipping_selected, payment_selected, completed``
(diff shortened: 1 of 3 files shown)


docs/book/checkout.rst
```diff
@@ -0,0 +1,97 @@
+.. index::
+   single: Checkout
+
+Checkout
+========
+
+**Checkout** is a process that begins when the Customer decides to finish his shopping and transform his **Cart** into an **Order**.
```
pamil: Except that a Cart is an Order and an Order is a Cart :tada:

docs/book/checkout.rst
```diff
@@ -0,0 +1,97 @@
...
++--------------------------------------+-------------------------------------------------+
+
+Checkout State Machine
+----------------------
+
+The Order Checkout state machine has 5 states available: ``cart, addressed, shipping_selected, payment_selected, completed``
+and a set of defined transitions between them.
+
+Besides the steps of checkout, each of them can be done once more. For instance if the Customer changes his mind
+and after selecting payment he wants to change the shipping addres he has already specified , he can of course go back and readdress.
```
pjedrzejewski: Extra space before comma.

pjedrzejewski:
We need more details about each step:

* What kind of info is provided there
* For addressing especially: how customer is set (guest, existing e-mail, existing account)
* How to perform checkout programmatically: See https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php

docs/book/checkout.rst
dantleech: s/his/their

docs/book/checkout.rst
```diff
@@ -0,0 +1,97 @@
...
++--------------------------+----------------------------------------------------+
+| Transition after step    | Template                                           |
++--------------------------+----------------------------------------------------+
+| ``cart``-> ``addressed`` | ``SyliusShopBundle:Checkout:addressing.html.twig`` |
++--------------------------+----------------------------------------------------+
+
+Selecting shipping
+~~~~~~~~~~~~~~~~~~
+
+It is a step where the customer selects the way his order will be shipped to him.
```
dantleech: s/his/their

docs/book/checkout.rst
```diff
@@ -0,0 +1,97 @@
...
++---------------------------------------+--------------------------------------------------+
+| Transition after step                 | Template                                         |
++---------------------------------------+--------------------------------------------------+
+| ``addressed``-> ``shipping_selected`` | ``SyliusShopBundle:Checkout:shipping.html.twig`` |
++---------------------------------------+--------------------------------------------------+
+
+Selecting payment
+~~~~~~~~~~~~~~~~~
+
+This is a step where the customer chooses how is he willing to pay for his order.
```
dantleech: *they* are willing to pay for *their* order.

dantleech:
Sorry internet connection really bad here. I would use the
geneder-neutral "they" / "their" instead of "he" / "his".

Also would quote each of the states individually (``addressing``,
``shipment``, etc.)

docs/book/checkout.rst
```diff
@@ -0,0 +1,259 @@
...
+| ``cart``-> ``addressed`` | ``SyliusShopBundle:Checkout:addressing.html.twig`` |
++--------------------------+----------------------------------------------------+
+
+How to perform the Addressing Step programmatically?
+''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+Firstly if the **Customer** is not yet set on the Order it will be assigned depending on the case:
+
+* An already logged in **User** - the Customer is set for the Order using the ``Sylius\Bundle\CoreBundle\EventListener\CartBlamerListener``, that determines the user basing on the event.
+* An existent **User** that is not logged in - they are redirected to log in before continuing.
```
Arminek: actually right now the javascript handle this log in action on addressing step.

docs/book/checkout.rst
```diff
@@ -0,0 +1,259 @@
...
+          to: shipping_selected
+      complete:
+          from: [payment_selected]
+          to: completed
+
+Steps of Checkout
+-----------------
+
+Checkout in Sylius is divided into 4 steps. Each of these steps occurs when the Order goes into a certain state.
+See the Checkout state machine in the ``CoreBundle\Resources\config\app\state_machine.yml``
```
lchrusciel: It would be nice if instead of this path - namespaces we would give a direct link to file.

docs/book/checkout.rst
```diff
@@ -0,0 +1,259 @@
...
++--------------------------+----------------------------------------------------+
+
+How to perform the Addressing Step programmatically?
+''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+Firstly if the **Customer** is not yet set on the Order it will be assigned depending on the case:
+
+* An already logged in **User** - the Customer is set for the Order using the ``Sylius\Bundle\CoreBundle\EventListener\CartBlamerListener``, that determines the user basing on the event.
+* An existent **User** that is not logged in - they are redirected to log in before continuing.
+* A **Customer** that was present in the system before (we've got their e-mail) - the Customer instance is updated via cascade, the order is assigned to it.
```
lchrusciel: It is unclear(at least to me), where did we store a customer. One was here before, but how can we determine it?

docs/book/checkout.rst
```diff
@@ -0,0 +1,259 @@
...
+
+.. code-block:: php
+
+     /** @var AddressInterface $address */
+     $address = $this->container->get('sylius.factory.address')->createNew();
+
+     $address->setFirstName('Name');
+     $address->setLastName('Surname');
+     $address->setStreet('Street');
+     $address->setCountryCode('PL');
```
lchrusciel: Why rest of the samples have no meaning, and here we have direct value? Maybe it would be worthy to put sample data to all of fields. I don't know, an address of white house or something similar?

docs/book/checkout.rst
```diff
@@ -0,0 +1,259 @@
...
+     $address->setCity('City');
+     $address->setPostcode('11111');
+
+     $order->setShippingAddress($address);
+     $order->setBillingAddress($address);
```
lchrusciel: Don't we have some listener which will copy a shipping address to a billing one if it is not set?

docs/book/checkout.rst
```diff
@@ -0,0 +1,259 @@
...
+
+    $stateMachine = $stateMachineFactory->get($order, OrderCheckoutTransitions::GRAPH)
+    $stateMachine->apply(OrderCheckoutTransitions::TRANSITION_SELECT_PAYMENT);
+
+    $this->container->get('sylius.manager.order')->flush();
+
+**What happens with the transition?**
+
+The method ``recalculate($order)`` of the ``Sylius\Component\Core\OrderProcessing\OrderRecalculator`` is run.
+Here this method is responsible for controlling the **promotions** that depend on the payment method.
```
lchrusciel: Maybe we should mention something about adjustments, and what type of adjustment is created during checkout?

docs/book/checkout.rst
```diff
@@ -0,0 +1,259 @@
+.. index::
+   single: Checkout
+
+Checkout
+========
+
+**Checkout** is a process that begins when the Customer decides to finish their shopping and transform their **Cart** into an **Order**.
```
NeverResponse: This phrasing is wrong. Customer doesn't want to transform anything, he just wants to buy stuffs. The things that actually happen in the back are not of his concern. Meaning, split this sentence into two. One as the customer's perspective and second with what happens in the kitchen. But that's just my opinion 😈

docs/book/checkout.rst
pamil: Nope nope nope, how do you want to introduce millions of random listeners from the Core here?

docs/book/checkout.rst
```diff
@@ -0,0 +1,259 @@
...
+
+Checkout State Machine
+----------------------
+
+The Order Checkout state machine has 5 states available: ``cart``, ``addressed``, ``shipping_selected``, ``payment_selected``, ``completed``
+and a set of defined transitions between them - which are saved as the **checkoutState** of the **Order**.
```
NeverResponse: Because the subject changes in the middle of the sentence is not clear whether **states** or **transitions** are saved in the **checkoutStates** (despite the obvious naming 😄 ). It might be confusing on someone's first contact with Sylius.

docs/book/checkout.rst
```diff
@@ -0,0 +1,259 @@
...
+
+Besides the steps of checkout, each of them can be done once more. For instance if the Customer changes their mind
```
NeverResponse: `once more` => `more than once`, the first one suggests you can only do those things twice during the same checkout.

docs/book/checkout.rst
```diff
@@ -0,0 +1,259 @@
...
+.. code-block:: php
+
+    $stateMachineFactory = $this->container->get('sm.factory');
+
+    $stateMachine = $stateMachineFactory->get($order, OrderCheckoutTransitions::GRAPH)
+    $stateMachine->apply(OrderCheckoutTransitions::TRANSITION_ADDRESS);
+
+    $this->container->get('sylius.manager.order')->flush();
+
+**What happens with the transition?**
```
NeverResponse: `with` => `on`, `during` ?

docs/book/checkout.rst
NeverResponse: I'd switch everything below a logged in user into a subsection, stating that depending on provided email address the appropriate entity gets retrieved/created; and moved the rest of the cases after this explanation.
Besides there is too much confusion here at the moment etc. 
`An existent **User** that is not logged in - they are redirected to log in before continuing.` 
How exactly does this magic happens?
How do we know a someone's a User before they get asked to log in?

docs/book/checkout.rst
TheMadeleine: Not sure if it is necessary here.

docs/book/checkout.rst
TheMadeleine: Isn't it encouraging laziness in people? ;>

NeverResponse: Yup, but it's also a lot faster than lookin' for it manually. Besides, are you implying that being lazy is somehow wrong :O ?

TheMadeleine: Yes, I do :D when you are a developer that wants to learn sth from the docs... then yes. being lazy is wrong. 🎉

pamil: There are no good developers that aren't lazy.

lchrusciel: We are always a little lazy... At the end it is not even a proper path. Also, it would be annoying to look for each resource manually.

pjedrzejewski:
Thanks Magda, looks good now! :)

PR #5741, opened 2016-08-09
[API][Order] Update OrderType to have customer optional

danut007ro:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

The OrderType in ApiBundle has the optional **customer** field instead of the mandatory and deprecated **user** field.


pjedrzejewski:
Thank you @danut007ro!

Issue #5746, opened 2016-08-10
Design is not coming properly


napestershine:
HI

I just install sylius on my system Ubuntu 16.04, PHP 7.x, MySQL 5.7.x, apache. I am getting following screen when i run 
`$php app/console server:run
`
![image](https://cloud.githubusercontent.com/assets/8603325/17542819/a720558a-5ee9-11e6-8038-59cbc985b89a.png)

bretto36:
`php app/console assets:install`
try running the above command

napestershine:
@bretto36  Thanks for support. I have run that command. But same results.

napestershine:
Console showing following errors:
![image](https://cloud.githubusercontent.com/assets/8603325/17543070/9838714a-5eeb-11e6-881b-d393e3f77635.png)

bretto36:
Oh you'll need to install `npm` and `gulp` then run `npm run gulp`

Install node (follow instructions here): https://github.com/npm/npm
Then install gulp using NPM: npm install gulp -g

Then run npm run gulp in the sylius base dir

napestershine:
@bretto36  Thanks for support. I have installed npm 
```
$sudo apt-get install nodejs
$sudo apt-get install npm
$npm install gulp -g
```

now I am trying to run 
`$npm rum gulp `
in sylius dir. it gives following error:
`root@wmc:/opt/lampp/htdocs/sylius/sylius-standard# npm run gulp
npm ERR! Linux 4.4.0-34-generic
npm ERR! argv "/usr/bin/nodejs" "/usr/bin/npm" "run" "gulp"
npm ERR! node v4.2.6
npm ERR! npm  v3.5.2

npm ERR! missing script: gulp
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /opt/lampp/htdocs/sylius/sylius-standard/npm-debug.log
`

bretto36:
Did gulp install successfully? It seems to be unable to find the gulp command

try just running command `gulp`

ylastapis:
You should just install it locally (```npm install gulp```) and then just run ```gulp```

napestershine:
Hi

Thanks for support.

```
root@wmc:/opt/lampp/htdocs/sylius/sylius-standard# npm install gulp
[...]

PR #5747, opened 2016-08-10
[Documentation] Installation and Architecture fix

TheMadeleine:
| Q               | A
| --------------- | ---
| Doc fix?        | no
| New docs?    | yes
| BC breaks?      | no
| Related tickets | partially #5275, #5746 
| License         | MIT

### DIFF
--- a/docs/book/installation.rst
+++ b/docs/book/installation.rst
@@ -63,6 +63,12 @@ And now you can use gulp for installing views, by just running a simple command:
 
     $ npm run gulp
 
+Although if you have Gulp installed globally then run just:
+
+.. code-block:: bash
+
+    $ gulp
+
 Accessing the Shop
 ------------------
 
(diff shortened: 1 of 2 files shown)


docs/book/installation.rst
```diff
@@ -61,7 +61,7 @@ And now you can use gulp for installing views, by just running a simple command:
 
 .. code-block:: bash
 
-    $ npm run gulp
+    $ gulp
```
pamil: We should still use `npm run gulp`, because `npm install` installs it locally, but not globally.

TheMadeleine: true. :P

lchrusciel: It is not only about where gulp is installed, but also that the previous command was the only way, to run a npm script. otherwise we would have to call a binary from a node_modules or link this to bin folder. But of course you are 💯 % right.

pjedrzejewski:
👍 👍 👍

PR #5756, opened 2016-08-10
[GridBundle] Add elastica driver

### LINKED ISSUE #4659: [RFC] Replacing SearchBundle with Grid Elastica driver
Helloes,

SearchBundle has been great and provided us with search functionality when there was none, but it has some issues:
- MySQL implementation of faceted search is quite complex and not very efficient
- Finder is stateful and not clean
- ProductController indexByTaxon method is quite messy

Goals:

Maintaining SQL implementation of search is a lot of work and most stores use ElasticaBundle and integrate search manually (which is quite simple) or use our not-so-perfect Elastic engine for SearchBundle. I'd like to simplify the whole thing.

Solution:

Product list by category and default product search would be handled by our GridBundle. Because search is really just a grid of products with filtering and sorting. All this can be handled by Grid component now. Default Sylius would ship with Doctrine ORM grid configured, which would provide basic search functionality (no faceted search):

Routing:

```
sylius_shop_product_index:
     path: /products/{taxonPermalink}
     defaults:
         _sylius:
             template: SyliusShopBundle:Product:index.html.twig
             grid: sylius_shop_product
```

Grid (Doctrine/ORM):

```
sylius_grid:
    grids:
        sylius_shop_product:
            driver:
                name: doctrine/orm
                options:
                    class: %sylius.model.product.class%
                    repository:
                        method: createByChannelAndTaxonPermalinkQueryBuilder
                        arguments:
                            - 'expr:service('sylius.context.channel').getChannel()'
                            - $taxonPermalink
            filters:
                search:
                    type: string
                    options:
                        fields: [translation.name, translation.description]
                 ### Other simple filters totally possible, just without facets and with worse perf. compared to ElasticSearch of course.
            sorting:
                translation.name: asc
```

Grid (Elastica)

```
sylius_grid:
    grids:
        sylius_shop_product:
            driver:
                name: elastica
                options:
                    index: sylius
                    type: product
                    terms: # ???
                        channels: 'expr:service('sylius.context.channel').getChannel()'
                        taxons: $taxonPermalink
            filters:
                search:
                    type: ??? string or elastica (depends on how we can implement it)
                 ### Your facets configurations go here
                 tshirt_color:
                     type: sylius_product_attribute
                     options:
                          code: tshirt_color
```

As you can see, using Grids allows us to reuse the whole system of filters and forms, if we implement Elastica driver, we can use the same system and make it work with both Doctrine ORM and Elastica. Sounds crazy but I already did that for Doctrine DBAL and Doctrine ORM. :dancer: 

Pagerfanta supports Elastica and FOSElasticaBundle supports Pagerfanta, so that makes whole thing even easier.

@NeverResponse will be working on Elastica driver for grids. :)

Steps to implement new driver in Sylius grids:
1. Implement ElasticaDriver, which implements DriverInterface, example for ORM: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/GridBundle/Doctrine/ORM/Driver.php
2. Implement DataSource for this driver: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/GridBundle/Doctrine/ORM/DataSource.php
3. Implement ExpressionBuilder: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/GridBundle/Doctrine/ORM/ExpressionBuilder.php
4. Register in container with tag `sylius.grid_driver`: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/GridBundle/Resources/config/drivers.xml
5. Configure in your Grid :dancer: 

Things to note:
1. Grid bundle should not depend on FOSElasticaBundle, the driver should be only enabled if FOSElasticaBundle is in Kernel.
2. **Initial implementation needs to do only one thing: Search by string with pagination (`findPaginated`) method should be helpful here, we do not need facets in the PoC**

Things to figure out in the future:
1. How to pass Aggregates information to filters.
2. How to take over the world with YAML programming.

Helpful links:
- Elastica documentation: http://elastica.io/getting-started/search-documents.html
- FOSElasticaBundle usage guide: https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Resources/doc/usage.md
- Pagerfanta with ElasticaAdapter usage: https://github.com/whiteoctober/Pagerfanta

### LINKED ISSUE #5775: [SearchBundle] Misssing ElasticaProductListener
Because removal of searchBundle in kernel, ElasticaProductListener is not any more loaded.

So it has to be moved in another bundle, CoreBundle or ProductBundle maybe ?

Niiko:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | yes |
| Related tickets | fixes #4659, fixes #5775 |
| License | MIT |

Here is my work, following your guidelines.

It introduce a BC Break with usage of sylius_shop_product grid. In fact, usage of fos elastica config is a pre-requisite to make it working.

Maybe it's time to add a default config in ShopBundle/resources/config/app/config.yml

A limitation is usage of elastic 1.7 syntax, but as mentionned in the issue, soon we can update to latest one.

Another problem is that i can't be able to use a route as proposed "path: /products/{taxonPermalink}" because of sylius_shop_product_show, so i removed taxonPermalink parameter for now.

ShopBundle/Resources/views/Product/_gridList.html.twig has to be fixed, for now 'sylius_shop_partial_product_index_by_taxon_code' does not use the grid, so i change the template in a "hacky way"

There are a lot of missing things
- [x] Channel is not taken in account, as enabled and other properties added previously on "addAvailableProductsVariantFilters" in SearchBundle/Finder/ElasticsearchFinder
- [x] Most of expressions in expression builder are not supported
- [x] Default fos_elastica config is not added
- [x] Fix _gridList.html.twig 
- [x] Specs

If that PR is in the good direction, i will continue in order to finish it.


Niiko:
Ok, so i think i can be reviewed on that one @pjedrzejewski 

Could you help me on the spec, i can't get rid of my error, on my computer, same tests are ok ?

One things about this PR, i had to add an elasticsearch hook for tests, for now it's using a @BeforeStep, i will work on #5775 to move onto @BeforeScenario

src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
```diff
@@ -205,6 +205,9 @@ public function theProductShouldAppearInTheShop($productName)
      */
     public function iWantToBrowseProducts()
     {
+        // Timeout because of elasticsearch update timing, use "curl 'localhost:9200/_cat/pending_tasks'" to check instead
+        sleep(1);
```
Arminek: Maybe ``->waitFor()`` method inside ``indexPage`` would help you. You can create something like ``->waitForSomething()`` in ``indexPage``.

Niiko: Yes, but it's not related to the page itself and i can't rely on the display.

I think instead make an elasticsearchChecker (as the notificationChecker) and inject it into the context with a waitForPendingRequests().

WDYT ?

Arminek:
The reason that your spec failed is that you coupled ``GridBundle`` with elastica.

src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
Arminek: It sounds ok.

Niiko:
Ok, got it, so "friendsofsymfony/elastica-bundle" has to be in require or require-dev  in GridBundle/composer.json if i want that spec pass, correct ?

The only problem is the guideline from RFC that GridBundle should not depend on that bundle.

So, what do you prefer, remove that spec ? Add this dependency in composer (Bundle has not to be not enabled anyway) ? Another way ?

Niiko:
So, i pass to BeforeScenario elasticsearch checker, there is only one test that fail, but can't understand why...Everything seems ok, i probably miss something big.

pjedrzejewski:
Hi Nicolas! Thank you so much for this work, this is just amazing! 👍  Only problem I see with merging that in the current state: we do not want to make Sylius dependent on ElasticSearch - that should be optional. I am not sure how to achieve this yet, but we will definitely give you hand on this one!

Niiko:
No problem, like to help :)

Yes, don't see an easy way to achieve this. The driver is already not loaded if you don't have fos elastica bundle enabled.

Maybe i could make it the same way to load the config (for mappings and grids). WDYT ?

If FOS, elastic config for all things (default mapping, driver added, doctrine grids replaced by elastic one) by overriding where applicable. 
Else, don't do anything.

Niiko:
About this PR, i will wait for v1.0.

I think i can separate mapping into a separate file as well as for grids (it will override configurations of doctrine grids), and add an entry in docs with "how to add elastica" or something like this. Which will tells to : 

- include mapping file (or redefine it)
- include elastic grids file

What about tests ? I don't know if it's possible to re-launch tests with a separate "config file", all scenarii with a @elasticsearch tag will be executed.

Last concern is about composer and fos-elastica bundle dependency.

Let me know in which direction you want to go, or if you have a better solution.

pjedrzejewski:
@Niiko We will help you make it a "plugin bundle" to Sylius after we tag beta, promised! :)

src/Sylius/Behat/Service/ElasticsearchChecker.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+namespace Sylius\Behat\Service;
+
+use FOS\ElasticaBundle\Elastica\Client;
+
+/**
+ * @author Nicolas Adler <nicolas.adler@openizi.com>
+ */
+final class ElasticsearchChecker implements ElasticsearchCheckerInterface
```
Arminek: I would change the name of this service to something like `ElasticSearchRefresher` WDYT?

src/Sylius/Behat/Service/ElasticsearchChecker.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        $this->client = $client;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function refreshIndex()
+    {
+        $this->client->request('_refresh');
```
Arminek:
```php
$response = $this->cilent->refreshAll()

if ($response->hasError()) {
throw new \RuntimeException(sprintf('Refresh action came back with error "%s"', $response->getError()));
}
```

src/Sylius/Bundle/CoreBundle/EventListener/ElasticaProductListener.php
```diff
@@ -0,0 +1,118 @@
...
+use FOS\ElasticaBundle\Persister\ObjectPersister;
+use Sylius\Component\Product\Model\AttributeValueInterface;
+use Sylius\Component\Product\Model\ProductInterface;
+use Sylius\Component\Core\Model\ProductTranslation;
+use Sylius\Component\Variation\Model\VariantInterface;
+
+/**
+ * @author Nicolas Adler <nicolas.adler@openizi.com>
+ */
+class ElasticaProductListener implements EventSubscriber
```
Arminek: This listener is no longer needed.

src/Sylius/Behat/Service/ElasticsearchChecker.php
Niiko: Yes, it would be more accurate

src/Sylius/Bundle/CoreBundle/EventListener/ElasticaProductListener.php
Niiko: ok

pjedrzejewski:
@Niiko Hey Nicolas! We just discussed the PR and how we could migrate it to a plugin, but we realized that we need to take a bit different approach than what you implemented. We need to follow the idea of other drivers and use a query builder + perhaps implement custom filter types for price range, options, attributes, subtaxons. Your code was very helpful in understanding the problem and we really appreciate your work, but cannot really reuse much of it. Hopefully you can contribute to the plugin development when we release the first version! Thanks!

Please follow the development done by @Arminek in https://github.com/Lakion/SyliusElasticSearchBundle.

Niiko:
ok

PR #5768, opened 2016-08-11
Allow friendsofsymfony/rest-bundle 2.0

bendavies:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Related tickets |  |
| License | MIT |


bendavies:
looks like the serialization format of forms has changed.

bendavies:
well, it's possible but it's a bit more work to support both 1.5+ and 2.0+.
do we want to enforce 2.0?

lchrusciel:
Such a beatifull hack:
```php
            // > 2.0
            if (method_exists($view, 'getSerializationContext')) {
                $view->getSerializationContext()->enableMaxDepthChecks();
            }
            // 2.0
            if (method_exists($view, 'getContext') && method_exists($view->getContext(), 'setMaxDepth')) {
                $view->getContext()->setMaxDepth(PHP_INT_MAX);
            }
            // 2.1+
            if (method_exists($view, 'getContext') && method_exists($view->getContext(), 'enableMaxDepth')) {
                $view->getContext()->enableMaxDepth();
            }
```
We have to decide to one of them. Maybe 2.1 would be a wise decision? @bendavies do you know any particular pros or cons of each version?

bendavies:
@lchrusciel not really, apart from some nice new features.
I like to keep ontop of newer versions in my projects.
run `composer outdated` on sylius and you'll see a list which hurts my OCD 💃

lchrusciel:
So I would vote for 2.1 then 👍 . But this nasty BC Break between 2.0 and 2.1 is terrifying. Is a sem.ver such a tough concept?

bendavies:
@lchrusciel thankfully not a BC break, just a deprecation
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Context/Context.php#L180-L197

which means we could still go `^2.0`

lchrusciel:
so `^2.0` then 🎯

bendavies:
alright, i'll work on this!

bendavies:
@pamil i didn't have time to finish this off but maybe a good idea to just bump to the latest FOSRestBundle now sylius is only supporting symfony3

lchrusciel:
@bendavies we want to bump FOSRestBundle to 2.0 at least before beta2. Why did you delete this branch?

lchrusciel:
Ok, I haven't seen #7349. Nvm :)

PR #5771, opened 2016-08-11
[WIP] [Docker] Refresh Docker support

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT


loic425:
@pamil Great !!
I have to try that :)

Do you (or Lakion) use docker instead of vagrant now ? What about behat tests, is it working well with this docker configuration ?

pamil:
@loic425 I tried to make it work, but, to be honest, I've failed tremendously :) Docker is quite nice, but we're all on OS X and I/O overhead here is unacceptable (https://forums.docker.com/t/file-access-in-mounted-volumes-extremely-slow-cpu-bound/8076). 

Anyway, I'm not a Docker master (trying to use it since yesterday :tada:), so it would be great to have someone with greater knowledge to help with this issue.

pamil:
For example, writing 100MB on my local machine takes ~0.25s, but inside the container in shared folder it takes ~42s.

loic425:
@pamil Yep, we have the same problem cause we are on mac too except one of us.

I've found many weeks ago a tutorial (in french unfortunately for you) to use behat/selenium/docker 
http://blog.lepine.pro/environnement-test-docker-behat/ 
The tutorial works. But i've tried to update the current sylius docker without succeed. But this might help you :)

inssein:
@pamil maybe just move the high IO folders to outside of shared folder? (cache + logs)

rawkode:
Yes, osxfs does have some serious problems right now.  Your best bet is to avoid Docker for Mac and use docker-machine with the `--virtualbox-no-share` flag for now (vboxfs is almost as slow as osxfs)

This means you'll need to get your files onto the VM through other means, of which there are many routes. I recommend my own script: https://github.com/rawkode/docker-developer-utils and use the `docker-machine-rsync` command to sync your home directory, or only your project.

There are NFS alternatives if you need two way file sync.

I'll make a PR to this PR to update it to the most recent Docker standards. Although Windows and OSX users need to make some small workarounds, there's no need to not atleast offer them the compose setup, and of course allow Linux users to utilise

loic425:
@pamil Have you try this docker configuration ? What about performance on OS X and I/O ?

pamil:
Closing as I've never got to run Docker properly on macOS, feel free to base on this branch.

rawkode:
@pamil What about all the people that contribute on Linux?

steffenbrem:
@pamil Use https://github.com/codekitchen/dinghy I am using it for several months with big success 👍

PR #5145, opened 2016-05-30
[Migrations] Initialize Sylius migrations and use them on travis and installer

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets |
| License         | MIT

Finally, we can use *Sylius* migrations properly :) I've generated one massive initial migration (also removed two unnecessarily merged) and added migrations usage in **Travis** and **installer**. I'm pretty sure there should be some work done on **Travis** with possible error handling, but I'm certain it should  by done by @pamil 🐫  😄

PR #5783, opened 2016-08-12
One migration to rule them all

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | -
| License         | MIT

Since our database structure is quite stable at the moment, it would be wise to require migrations to stay up-to-date.


loic425:
@pamil what about #5145  ?

pjedrzejewski:
Thanks Kamil! :)

PR #5789, opened 2016-08-14
[Instaltion] Fix setup instalator

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

By default a setup command should create an admin user, not a shop one.

PR #5791, opened 2016-08-14
[RFC] Identifable User abstaction

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Related tickets | 
| License         | MIT

Both ShopUserInterface and AdminUserInterface had same methods. Our logic force to also override one of the user methods too. Because we have missed this override in admin class I had to made #5789. This PR extracts common logic to one abstract class, just to keep SSoT. Both AdminUserInterface and ShopUserInterface have been left as a markers.


lchrusciel:
Not needed anymore

PR #5824, opened 2016-08-17
[API] Added Order and OrderItem apis tests and fixes

danut007ro:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | yes |
| BC breaks? | yes |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/5231 |
| License | MIT |

Added tests for Order and OrderItem
Added to_identifier and to_hidden_identifier for VariationBundle

PR #5829, opened 2016-08-18
[API][Order] Order item API fixes

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets |
| License         | MIT


src/Sylius/Bundle/ApiBundle/Resources/config/routing/order_item.yml
```diff
@@ -19,6 +19,10 @@ sylius_api_order_item_create:
         _controller: sylius.controller.order_item:createAction
         _sylius:
             form: sylius_api_order_item
+            factory:
+                method: createForOrderWithId
+                arguments:
+                    type: $orderId
```
lchrusciel: It would be better to use an expression here, and decouple a factory from the repository. `expr:service('sylius.repository.order').find($orderId)` 

Then also, we should change a name to `createForOrder`

src/Sylius/Bundle/CoreBundle/Form/Type/OrderItemType.php
```diff
@@ -36,23 +34,11 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             ->addEventListener(FormEvents::PRE_SUBMIT, function (FormEvent $event) use ($options) {
                 $data = $event->getData();
                 if (isset($data['variant'])) {
-                    $event->getForm()->add('variant', 'entity_hidden', [
-                        'data_class' => $options['variant_data_class'],
+                    $event->getForm()->add('variant', 'sylius_product_variant_from_identifier', [
+                        'class' => ProductVariant::class,
```
lchrusciel: Not sure, how it would work, if we would override a ProductVariant class

src/Sylius/Bundle/OrderBundle/Resources/config/services.xml
```diff
@@ -51,6 +52,11 @@
             <argument type="service" id="sylius.custom_factory.adjustment.inner" />
         </service>
 
+        <service id="sylius.custom_factory.order_item" class="%sylius.custom_factory.order_item.class%" decorates="sylius.factory.order_item" public="false">
+            <argument type="service" id="sylius.custom_factory.order_item.inner" />
+            <argument type="service" id="sylius.repository.order" />
```
lchrusciel: Remove

src/Sylius/Component/Order/spec/Factory/OrderItemFactorySpec.php
```diff
@@ -0,0 +1,70 @@
...
+use Sylius\Component\Order\Factory\OrderItemFactoryInterface;
+use Sylius\Component\Order\Model\OrderInterface;
+use Sylius\Component\Order\Model\OrderItemInterface;
+use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class OrderItemFactorySpec extends ObjectBehavior
```
lchrusciel: **final**

tests/Responses/Expected/order_item/create_response.json
```diff
@@ -0,0 +1,27 @@
+{
+    "id": @integer@,
+    "quantity": 1,
+    "unit_price": 2000,
+    "total": 2000,
+    "immutable": @boolean@,
```
lchrusciel: I bet, this value can be determined from the beginning, so I would vote to create different create_response with different values declared, rather then creating on file with lot of matcher. By default it would be better if we would keep matcher for things, that can be changed really easily, such as date or id

lchrusciel:
POST-ing to api without data should result in validation error, so 400 only...

Zales0123:
@danut007ro I really didn't notice your PR! If it will be good to go, I will obviously cancel this one, you were first ;)
About 500 error - this is all about [asserting unit price value](https://github.com/Sylius/Sylius/blob/4c9babd9b516dc4907eeaa1211dcf266ae6eaf42/src/Sylius/Component/Order/Model/OrderItem.php#L144) in ``setUnitPrice`` method. Invalid value cannot be set to not pass validation. We should consider if we really should asset type inside model.

src/Sylius/Component/Order/Factory/OrderItemFactory.php
```diff
@@ -0,0 +1,64 @@
...
+namespace Sylius\Component\Order\Factory;
+
+use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+use Webmozart\Assert\Assert;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class OrderItemFactory implements OrderItemFactoryInterface
```
pamil: final?

danut007ro: @pamil please check https://github.com/Sylius/Sylius/pull/5824
It's (almost) the same pull request with this feedback fixed.

PR #5832, opened 2016-08-18
[Behat][Domain] Minor refactor domain contexts

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Domain/ManagingProductsContext.php
+++ b/src/Sylius/Behat/Context/Domain/ManagingProductsContext.php
@@ -24,7 +24,7 @@ use Webmozart\Assert\Assert;
  * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
  * @author Magdalena Banasiak <magdalena.banasiak@lakion.com>
  */
-final class ProductContext implements Context
+final class ManagingProductsContext implements Context
 {
     /**
      * @var SharedStorageInterface
@@ -66,19 +66,12 @@ final class ProductContext implements Context
 
     /**
      * @When /^I delete the ("[^"]+" variant of product "[^"]+")$/
+     * @When /^I try to delete the ("[^"]+" variant of product "[^"]+")$/
      */
     public function iDeleteTheVariantOfProduct(ProductVariantInterface $productVariant)
     {
-        $this->sharedStorage->set('product_variant_id', $productVariant->getId());
-        $this->productVariantRepository->remove($productVariant);
-    }
-
-    /**
-     * @When /^I try to delete the ("[^"]+" variant of product "[^"]+")$/
-     */
-    public function iTryToDeleteTheVariantOfProduct(ProductVariantInterface $productVariant)
-    {
         try {
+            $this->sharedStorage->set('product_variant_id', $productVariant->getId());
             $this->productVariantRepository->remove($productVariant);
         } catch (DBALException $exception) {
             $this->sharedStorage->set('last_exception', $exception);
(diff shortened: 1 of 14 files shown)


src/Sylius/Behat/Context/Domain/OrderContext.php
```diff
@@ -105,7 +99,7 @@ public function iDeleteTheOrder(OrderInterface $order)
     }
 
     /**
-     * @Then /^([^"]+) should not exist in the registry$/
+     * @Then /^(this order) should not exist in the registry$/
      */
     public function orderShouldNotExistInTheRegistry(OrderInterface $order)
```
michalmarcinkowski: We shouldn't use repository in this method also.

michalmarcinkowski:
The class should be renamed to match the corresponding context in UI namespace `ManagingOrdersContext`.

pjedrzejewski:
Thanks Grzesiu!

src/Sylius/Behat/Context/Domain/ManagingProductsContext.php
```diff
@@ -66,31 +66,26 @@ public function __construct(
...
-    {
-        $this->sharedStorage->set('product_variant_id', $productVariant->getId());
-        $this->productVariantRepository->remove($productVariant);
-    }
-
-    /**
      * @When /^I try to delete the ("[^"]+" variant of product "[^"]+")$/
      */
-    public function iTryToDeleteTheVariantOfProduct(ProductVariantInterface $productVariant)
+    public function iDeleteTheVariantOfProduct(ProductVariantInterface $productVariant)
```
michalmarcinkowski: There is a reason for keeping `I delete` and `I try to delete` methods separate. For domain implementation we mute the exception in step `I try to delete` so we can assert it occurred, but for step `I delete` we shouldn't do that and let it be thrown. This change should be reverted.

src/Sylius/Behat/Context/Domain/ManagingProductsContext.php
```diff
@@ -117,39 +112,28 @@ public function productVariantShouldNotExistInTheProductCatalog()
...
+     * @Then this variant should still exist in the product catalog
      */
-    public function productVariantShouldExistInTheProductCatalog(ProductVariantInterface $productVariant)
+    public function productVariantShouldExistInTheProductCatalog()
     {
-        $productVariant = $this->productVariantRepository->find($productVariant->getId());
+        $productVariantId = $this->sharedStorage->get('product_variant_id');
+        $productVariant = $this->productVariantRepository->find($productVariantId);
 
         Assert::notNull($productVariant);
```
michalmarcinkowski: How about the following?
```
/**
 * @Then /^(this variant) should still exist in the product catalog$/
 */
public function productVariantShouldExistInTheProductCatalog(ProductVariantInterface $productVariant = null)
{
    Assert::notNull($productVariant);
} 
```

src/Sylius/Behat/Context/Domain/ManagingPromotionCouponsContext.php
```diff
@@ -45,31 +45,20 @@ public function __construct(SharedStorageInterface $sharedStorage, CouponReposit
...
+     * @When /^I delete ("[^"]+" coupon) related to (this promotion)$/
      * @When /^I try to delete ("[^"]+" coupon) related to (this promotion)$/
      */
     public function iTryToDeleteCoupon(CouponInterface $coupon, PromotionInterface $promotion)
     {
         try {
             $promotion->removeCoupon($coupon);
             $this->couponRepository->remove($coupon);
-
-            throw new \Exception(sprintf('Coupon "%s" has been removed, but it should not.', $coupon->getCode()));
```
michalmarcinkowski: It should not be removed (see my comment above).

PR #5869, opened 2016-08-23
[Rbac] remove rbac bundle and component

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | -
| License         | MIT


okwinza:
I must've missed something big, what's the reason behind this?

pamil:
@okwinza http://sylius.org/blog/sylius-roadmap-update

Niiko:
One question about it, can we rely on rbac bundle instead even if it's in READ-ONLY ? Or it would be preferable to implement own system ?

bendavies:
so sylius won't do any rbac out of the box any more?

harikt:
@pamil one thing I probably missed to understand is does this components / bundle stay at a different repo? I am also looking at the same question on http://disq.us/p/1b5jlhn .

michalmarcinkowski:
The code related to these features will be removed from repository, but we would like to bring some features back to Core in 1.x versions. Nevertheless you're still able to use e.g. RbacBundle v0.19 in your project, but if you would like to use it with Sylius Core in later version than v0.19 you will need to integrate it on your own.

gabiudrescu:
@michalmarcinkowski just to be extra-sure: removing the code from Sylius Core does not imply deleting the single repo - e.g. https://github.com/Sylius/Rbac ?

no need to fork this just to keep it safe?

michalmarcinkowski:
@gabiudrescu you still will be able to download the previous releases, but the repo will be no longer maintained and READ-ONLY, so no new PRs will be merged. @pjedrzejewski correct me if I'm wrong.

gabiudrescu:
oh well, something is better than nothing. thanks for the clarification. I'll mention it on the blog also, so that others may be aware of it, too.

michalmarcinkowski:
@Arminek rebase required

michalmarcinkowski:
Thanks Arek!

PR #5873, opened 2016-08-24
correct arguments name

videni:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| License | MIT |


pjedrzejewski:
This is incorrect integration, it should generate the HWI services as well, otherwise we hardcode it for ``admin`` only, cc @Arminek.

michalmarcinkowski:
Fixed already.

PR #5882, opened 2016-08-24
[Checkout][Inventory] Verify complete checkout

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | #5872
| License         | MIT


features/inventory/validating_product_quantity_on_checkout_complete.feature
```diff
@@ -0,0 +1,35 @@
...
+        Then I should see the thank you page
+
+    @ui
+    Scenario: Product quantity validation fails
+        Given I have 5 products "Iron Maiden T-Shirt" in the cart
+        And I complete addressing step with email "john@example.com" and "France" as shipping country
+        And I select "Free" shipping method
+        And I complete the shipping step
+        And I choose "Offline" payment method
+        When product "Iron Maiden T-Shirt" quantity is changed to 3
```
lchrusciel: I don't get it. This scenario does not have sense to me. You had 5 of 5 t-shirts in a cart, and just before confirmation, you have reduced stock to 3(so 2 of them left) and then you see `I should be notified that this product does not have sufficient stock`

Arminek: This scenario cover something like this when customer is already in the middle of checkout process and for example some other customer has already bought the same product.

lchrusciel: The sentence is really confusing then. 
`But other customer has bough 2 products by this time`?

features/inventory/validating_product_quantity_on_checkout_complete.feature
```diff
@@ -0,0 +1,35 @@
...
+Feature: Validating product quantity on checkout complete
+    In order to be sure that products I buy are available
+    As a Customer
+    I want to have them validated on checkout complete
+
+    Background:
+        Given the store operates on a single channel in "France"
+        And the store has a product "Iron Maiden T-Shirt" priced at "€12.54"
+        And this product is tracked by the inventory
+        And there are 5 items of product "Iron Maiden T-Shirt" available in the inventory
```
michalmarcinkowski: Shouldn't we talk about `units` instead of `items` when it's about the inventory?

features/inventory/validating_product_quantity_on_checkout_complete.feature
```diff
@@ -0,0 +1,35 @@
+@inventory
+Feature: Validating product quantity on checkout complete
```
michalmarcinkowski: Feature: Being unable to buy products that are out of stock

features/inventory/validating_product_quantity_on_checkout_complete.feature
```diff
@@ -0,0 +1,35 @@
...
+    In order to be sure that products I buy are available
+    As a Customer
+    I want to have them validated on checkout complete
```
michalmarcinkowski: I want to be prevented from placing an order with products that are out of stock

features/inventory/validating_product_quantity_on_checkout_complete.feature
```diff
@@ -0,0 +1,35 @@
...
+    Background:
+        Given the store operates on a single channel in "France"
+        And the store has a product "Iron Maiden T-Shirt" priced at "€12.54"
+        And this product is tracked by the inventory
+        And there are 5 items of product "Iron Maiden T-Shirt" available in the inventory
+        And the store ships everywhere for free
+        And the store allows paying offline
+
+    @ui
+    Scenario: Product quantity validation passes
```
michalmarcinkowski: Placing an order with products that have sufficient quantity

features/inventory/validating_product_quantity_on_checkout_complete.feature
```diff
@@ -0,0 +1,35 @@
...
+        Given I have 3 products "Iron Maiden T-Shirt" in the cart
+        And I complete addressing step with email "john@example.com" and "France" as shipping country
+        And I select "Free" shipping method
+        And I complete the shipping step
+        And I choose "Offline" payment method
+        When I confirm my order
+        Then I should see the thank you page
+
+    @ui
+    Scenario: Product quantity validation fails
```
michalmarcinkowski: Being unable to place an order with products that are out of stock

features/inventory/validating_product_quantity_on_checkout_complete.feature
```diff
@@ -0,0 +1,35 @@
...
+        And the store ships everywhere for free
+        And the store allows paying offline
+
+    @ui
+    Scenario: Product quantity validation passes
+        Given I have 3 products "Iron Maiden T-Shirt" in the cart
+        And I complete addressing step with email "john@example.com" and "France" as shipping country
+        And I select "Free" shipping method
+        And I complete the shipping step
+        And I choose "Offline" payment method
```
michalmarcinkowski:
```
Given I added 3 products ...
And I completed addressing step with ...
And I selected "Free" shipping method
And I chose "Offline" payment method
When ...
```

michalmarcinkowski: Btw. don't we have a step where we do the above 3 steps in a single one? Maybe we could think of it since it's not so important for the scenario itself.

features/inventory/validating_product_quantity_on_checkout_complete.feature
```diff
@@ -0,0 +1,35 @@
...
+    @ui
+    Scenario: Product quantity validation fails
+        Given I have 5 products "Iron Maiden T-Shirt" in the cart
+        And I complete addressing step with email "john@example.com" and "France" as shipping country
+        And I select "Free" shipping method
+        And I complete the shipping step
+        And I choose "Offline" payment method
+        When product "Iron Maiden T-Shirt" quantity is changed to 3
+        And I confirm my order
+        Then I should not see the thank you page
```
michalmarcinkowski: How about:
```
And I try to confirm my order
Then I should still be on the order confirmation step
And I should be notified ...
```

src/Sylius/Bundle/CoreBundle/Resources/config/validation.xml
```diff
@@ -127,7 +127,10 @@
     <class name="Sylius\Component\Cart\Model\CartItem">
         <constraint name="Sylius\Bundle\InventoryBundle\Validator\Constraints\InStock">
```
michalmarcinkowski: You should also have a look at this validator and probably refactor it ;)

pjedrzejewski:
Thanks Arek! :)

pjedrzejewski:
As @michalmarcinkowski said, you can have a look at the InStock validation constraint and do a quick refactoring. ;)

PR #5889, opened 2016-08-25
[Rbac] Remove RBAC-related packages from composer "replace" list

coox:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | #5869
| License         | MIT

Now that RBAC is out of core, it should be possible to require `sylius/rbac` and/or `sylius/rbac-bundle` in a project where `sylius/sylius` is already a dependency.

These two lines are preventing it and should be removed.

I don't think `composer.lock` requires an update, please let me know if I'm wrong.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -137,8 +137,6 @@
         "sylius/product-bundle": "self.version",
         "sylius/promotion": "self.version",
         "sylius/promotion-bundle": "self.version",
-        "sylius/rbac": "self.version",
-        "sylius/rbac-bundle": "self.version",
         "sylius/registry": "self.version",
         "sylius/report": "self.version",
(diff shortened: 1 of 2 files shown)


gabiudrescu:
@coox the failing build says that composer.lock and composer.json are not in sync

coox:
I kind of thought this could happen, but wasn't sure how to prevent it.

Just ran `composer update --lock` which, according to the [Composer Doc](), _“Only updates the lock file hash to suppress warning about the lock file being out of date.”_.

Notice that it updated the case of some package author names, too…

michalmarcinkowski:
Thank you Eric! 👍

PR #5896, opened 2016-08-26
[API] Checkout API rework

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Related tickets |
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/routing/checkout.yml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/routing/checkout.yml
@@ -60,7 +60,7 @@ sylius_api_checkout_select_shipping:
             form: sylius_checkout_select_shipping
             repository:
                 method: find
-                arguments: $orderId
+                arguments: [$orderId]
             state_machine:
                 graph: sylius_order_checkout
                 transition: select_shipping
@@ -87,7 +87,7 @@ sylius_api_checkout_select_payment:
             form: sylius_checkout_select_payment
             repository:
                 method: find
-                arguments: $orderId
+                arguments: [$orderId]
             state_machine:
                 graph: sylius_order_checkout
                 transition: select_payment
@@ -114,7 +114,7 @@ sylius_api_checkout_complete:
             form: sylius_checkout_complete
             repository:
                 method: find
-                arguments: $orderId
+                arguments: [$orderId]
             state_machine:
                 graph: sylius_order_checkout
                 transition: complete
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/routing/checkout.yml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/routing/checkout.yml
@@ -51,3 +51,31 @@ sylius_api_checkout_select_shipping:
             state_machine:
                 graph: sylius_order_checkout
                 transition: select_shipping
+
+sylius_api_checkout_select_payment:
+    path: /select-payment/{orderId}
+    methods: [PUT]
+    defaults:
+        _controller: sylius.controller.order:updateAction
+        _sylius:
+            form: sylius_checkout_select_payment
+            repository:
+                method: find
+                arguments: $orderId
+            state_machine:
+                graph: sylius_order_checkout
+                transition: select_payment
+
+sylius_api_checkout_complete:
+    path: /complete/{orderId}
+    methods: [PUT]
+    defaults:
+        _controller: sylius.controller.order:updateAction
+        _sylius:
+            form: sylius_checkout_complete
+            repository:
+                method: find
+                arguments: $orderId
+            state_machine:
+                graph: sylius_order_checkout
+                transition: complete
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/routing/checkout.yml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/routing/checkout.yml
@@ -38,6 +38,19 @@ sylius_api_checkout_addressing:
                 graph: sylius_order_checkout
                 transition: address
 
+sylius_api_checkout_readdress:
+    path: /readdress/{orderId}
+    methods: [PUT]
+    defaults:
+        _controller: sylius.controller.order:applyStateMachineTransitionAction
+        _sylius:
+            repository:
+                method: find
+                arguments: [$orderId]
+            state_machine:
+                graph: sylius_order_checkout
+                transition: readdress
+
 sylius_api_checkout_select_shipping:
     path: /select-shipping/{orderId}
     methods: [PUT]
@@ -52,6 +65,19 @@ sylius_api_checkout_select_shipping:
                 graph: sylius_order_checkout
                 transition: select_shipping
 
+sylius_api_checkout_reselect_shipping:
+    path: /reselect-shipping/{orderId}
+    methods: [PUT]
+    defaults:
+        _controller: sylius.controller.order:applyStateMachineTransitionAction
+        _sylius:
+            repository:
+                method: find
+                arguments: [$orderId]
+            state_machine:
+                graph: sylius_order_checkout
+                transition: reselect_shipping
+
 sylius_api_checkout_select_payment:
     path: /select-payment/{orderId}
     methods: [PUT]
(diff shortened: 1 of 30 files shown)


src/Sylius/Bundle/CoreBundle/Form/EventSubscriber/AddDefaultBillingAddressOnOrderFormSubscriber.php
```diff
@@ -37,7 +37,10 @@ public function preSubmit(FormEvent $event)
     {
         $orderData = $event->getData();
 
-        if (!isset($orderData['differentBillingAddress']) || false === $orderData['differentBillingAddress']) {
+        if (
+            (!isset($orderData['differentBillingAddress']) || false === $orderData['differentBillingAddress']) &&
```
lchrusciel: Can we at least extract this logic to private method?

tests/Controller/CheckoutAddressingApiTest.php
```diff
@@ -0,0 +1,186 @@
...
+ */
+
+namespace Sylius\Tests\Controller;
+
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+class CheckoutAddressingApiTest extends CheckoutApiTestCase
```
Arminek: final?

src/Sylius/Bundle/ApiBundle/Resources/config/routing/checkout.yml
```diff
@@ -15,8 +15,106 @@ sylius_api_checkout_index:
...
+sylius_api_checkout_complete:
+    path: /complete/{orderId}
+    methods: [PUT]
+    defaults:
+        _controller: sylius.controller.order:updateAction
+        _sylius:
+            form: sylius_checkout_complete
+            repository:
+                method: find
+                arguments: $orderId
```
lchrusciel: To use a brackets or not to use them - that is the question!

Three out of seven arguments for find method does not have them ;)

tests/Controller/CheckoutReselectingPaymentApiTest.php
```diff
@@ -0,0 +1,82 @@
...
+        $orderId = $checkoutData['order1']->getId();
+        $this->addressOrder($orderId);
+        $this->selectOrderShippingMethod($orderId, $checkoutData['ups']->getId());
+        $this->selectOrderPaymentMethod($orderId, $checkoutData['cash_on_delivery']->getId());
+
+        $url = sprintf('/api/checkouts/reselect-payment/%d', $orderId);
+        $this->client->request('PUT', $url, [], [], static::$authorizedHeaderWithContentType);
+
+        $response = $this->client->getResponse();
+        $this->assertResponse($response, 'checkout/reselect_payment_response', Response::HTTP_OK);
```
lchrusciel: Why does this endpoint return a `HTTP_OK`? `HTTP_NON_CONTENT` should be returned as it a `PUT` action.

tests/Controller/CheckoutReselectingShippingApiTest.php
```diff
@@ -0,0 +1,81 @@
...
+
+        $orderId = $checkoutData['order1']->getId();
+        $this->addressOrder($orderId);
+        $this->selectOrderShippingMethod($orderId, $checkoutData['ups']->getId());
+
+        $url = sprintf('/api/checkouts/reselect-shipping/%d', $orderId);
+        $this->client->request('PUT', $url, [], [], static::$authorizedHeaderWithContentType);
+
+        $response = $this->client->getResponse();
+        $this->assertResponse($response, 'checkout/reselect_shipping_response', Response::HTTP_OK);
```
lchrusciel: Same as above.

tests/Controller/CheckoutShippingApiTest.php
```diff
@@ -0,0 +1,128 @@
...
+        $this->assertResponse($response, 'checkout/shipping_invalid_order_state', Response::HTTP_INTERNAL_SERVER_ERROR);
+    }
+
+    /**
+     * @test
+     */
+    public function it_does_not_allow_to_select_shipping_for_order_without_specifying_shipping_method()
+    {
+        // TO-DO
+        // feature cannot be tested properly due to bug with cascade shipments validation
```
lchrusciel: Can we create a issue with this one? I'm afraid that we will forget for some time about it, if we don't.

tests/Responses/Expected/checkout/reselect_shipping_response.json
```diff
@@ -0,0 +1,19 @@
+{
+    "id": @integer@,
+    "items": @array@,
```
lchrusciel: I'm not sure about mocking so many things... Cannot we just add basic data such as items or items total to these responses?

lchrusciel:
I'm missing a `GET` requests after each action. It would be useful to check if truly works and, whats probably even more important, will give some example responses for developer. It will also help with documentation.

tests/Responses/Expected/checkout/reselect_shipping_response.json
Zales0123: The problem is, it's ``applyStateMachineTransitionAction`` from ``ResourceController``, so we don't have many options on manipulating the response.

lchrusciel: But we can just instead of matcher use raw value. As I wrote before, it will just help developers to check an api structure.

Zales0123: I'm not sure about it - in case of ``re-`` transitions, the only thing that is relevant for test is checkout state of the order.

tests/Controller/CheckoutReselectingPaymentApiTest.php
Zales0123: Again, it's ``applyStateMachineTransitionAction`` and it always return [HTTP_OK response after successful non-html call](https://github.com/Sylius/Sylius/blob/7653df1267c3d4a6037edacf99e89253779a7111/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php#L412)

tests/Responses/Expected/checkout/reselect_shipping_response.json
lchrusciel: Thats true, but also there is no point in mocking it if we are sure about them. Anyway, it is up to you, so if you prefer to leave them this way, I'm ok with it.

Zales0123:
Fixed and improved 👍

src/Sylius/Bundle/CoreBundle/Form/EventSubscriber/AddDefaultBillingAddressOnOrderFormSubscriber.php
```diff
@@ -37,10 +37,24 @@ public function preSubmit(FormEvent $event)
...
+    /**
+     * @param array $orderData
+     *
+     * @return bool
+     */
+    private function shouldBillingAndShippingAddressBeTheSame(array $orderData)
+    {
+        return
+            (!isset($orderData['differentBillingAddress']) ||
+            false === $orderData['differentBillingAddress']) &&
```
pamil: Can we merge this two first conditions into one line? It would be a little bit cleaner.

tests/Controller/CheckoutAddressingApiTest.php
```diff
@@ -0,0 +1,191 @@
...
+        $customers = $this->loadFixturesFromFile('resources/customers.yml');
+        $checkoutData = $this->loadFixturesFromFile('resources/checkout.yml');
+
+        $data =
+<<<EOT
+        {
+            "shippingAddress": {
+                "firstName": "Hieronim",
+                "lastName": "Bosch",
+                "street": "Surrealism St.",
```
lchrusciel: 👍

lchrusciel:
Ready to merge for me

pjedrzejewski:
Thank you Mateusz! Great work! :)

PR #5914, opened 2016-08-29
[ThemeBundle] make ThemeConfiguration extendable

ahilles107:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | none
| License         | MIT

Remove final from class. 

Why? We have Theme system based on ThemeBundle and it's impossible now without duplication whole configuration class (registering it to container and replacing Sylius one with new one in Processor) to add some new properties to theme.json file.


pamil:
How would that help in customising those settings? The only non-private method there is `getConfigTreeBuilder` which is public, can be decorated, but still it doesn't provide any way to modify the configuration.

PR #5918, opened 2016-08-30
[Account] Subscribe to the newsletter in customer account

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Page/Shop/Account/ProfileUpdatePage.php
+++ b/src/Sylius/Behat/Page/Shop/Account/ProfileUpdatePage.php
@@ -76,6 +76,22 @@ class ProfileUpdatePage extends SymfonyPage implements ProfileUpdatePageInterfac
     /**
      * {@inheritdoc}
      */
+    public function subscribeToTheNewsletter()
+    {
+        $this->getDocument()->checkField('Subscribe to the newsletter');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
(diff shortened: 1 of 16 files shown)


src/Sylius/Behat/Page/Shop/Account/ProfileUpdatePage.php
```diff
@@ -76,6 +76,22 @@ public function saveChanges()
     /**
      * {@inheritdoc}
      */
```
Zales0123: Why do we need a ``{@inheritdoc}`` if there is no docblock in interface?

src/Sylius/Component/Customer/Model/Customer.php
```diff
@@ -266,6 +271,22 @@ public function setPhoneNumber($phoneNumber)
         $this->phoneNumber = $phoneNumber;
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function isNewsletterSubscription()
```
pjedrzejewski: isSubscribedToNewsletter? (we can skip "The")
protected $subscribedToNewsletter = false;

pjedrzejewski:
@GSadee Nice, please tweak the field name and it is good to go!

pjedrzejewski:
Thank you Grzesiu! :)

PR #5919, opened 2016-08-30
[Migrations] Add missing migrations after removing metadata and reports

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT


pjedrzejewski:
Thanks Grzesiu, good catch! :)

PR #5920, opened 2016-08-30
[Documentation] Concepts: Addresses

TheMadeleine:
| Q               | A
| --------------- | ---
| Doc fix?        | yes
| New docs?    | no
| BC breaks?      | no
| License         | MIT

### DIFF
--- a/docs/book/addresses.rst
+++ b/docs/book/addresses.rst
@@ -4,24 +4,47 @@
 Addresses
 =========
 
-Every address in Sylius is represented by *Address* model. Default structure has the following fields:
+Every address in Sylius is represented by the **Address** model.
+It has a few important fields:
 
-* firstname
-* lastname
-* street
-* city
-* postcode
-* reference to *Country*
-* reference to *Province* (optional)
-* createdAt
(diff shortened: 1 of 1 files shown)


docs/book/addresses.rst
```diff
@@ -4,24 +4,49 @@
...
+   $address->setFirstName('Harry');
+   $address->setLastName('Potter');
+   $address->setCompany('Ministry of Magic');
+   $address->setCountryCode('UK');
+   $address->setProvinceCode('UKJ');
+   $address->setCity('Little Whinging');
+   $address->setStreet('4 Privet Drive');
+   $address->setPostcode('000001');
+
+   $this->container->get('sylius.repository.address')->add($address);
```
pjedrzejewski: This is incorrect, address should not be added to the repository "alone", it can be, but shouldn't. :D Example with adding it to an order is enough. (persistence of the order is out of scope of this docs)

TheMadeleine: Okay! :P

pjedrzejewski:
Thank you Magda!

PR #5946, opened 2016-09-01
[Documentation] Concepts: Promotions

TheMadeleine:
| Q               | A
| --------------- | ---
| Doc fix?        | yes
| New docs?    | no
| BC breaks?      | no
| License         | MIT

### DIFF
--- a/docs/book/promotions.rst
+++ b/docs/book/promotions.rst
@@ -4,37 +4,173 @@
 Promotions
 ==========
 
-...
+The system of **Promotions** in **Sylius** is really flexible. It is a combination of promotion rules and actions.
+
+Promotions have a few parameters - a unique ``code``, ``name``, ``usageLimit``,
+the period of time when it works.
+There is a possibility to define **exclusive promotions** (no other can be applied if an exclusive promotion was applied)
+and **priority** that is useful for them, because the exclusive promotion should get the top priority.
+
+How to create a Promotion programmatically?
+-------------------------------------------
+
+Just as usually, use a factory. The promotion needs a ``code`` and a ``name``.
+
+.. code-block:: php
+
+   /** @var PromotionInterface $promotion */
+   $promotion = $this->container->get('sylius.factory.promotion')->createNew();
+
+   $promotion->setCode('simple_promotion_1');
+   $promotion->setName('Simple Promotion');
+
+**Of course an empty promotion would be useless** - it is just a base for adding **Rules** and **Actions**.
+Let's see how to make it functional.
 
 Promotion Rules
 ---------------
 
-...
+The promotion **Rules** restrict in what circumstances a promotion will be applied.
+An appropriate **RuleChecker** (each Rule type has its own RuleChecker) may check if the Order:
+
+* Contains a number of items from a specified taxon (for example: *contains 4 products that are categorized as t-shirts*)
+* Has a specified total price of items from a given taxon (for example: *all mugs in the order cost 20$ in total*)
+* Has total price of at least a defined value (for example: *the orders' items total price is equal at least 50$*)
+
+And many more similar, suitable to your needs.
+
+Rule Types
+''''''''''
+
+The types of rules that are configured in **Sylius** by default are:
+
+* **Cart Quantity** - checks if there is a given amount of items in the cart,
+* **Item Total** - checks if items in the cart cost a given amount of money,
+* **Taxon** - checks if there is at least one item from given taxons in the cart,
+* **Items From Taxon Total** - checks in the cart if items from a given taxon cost a given amount of money,
+* **Contains Taxon** - checks if there is a given amount of items from a given taxon in the cart,
+* **Nth Order** - checks if this is for example the second order made by the customer,
+* **Shipping Country** - checks if the order's shipping address is in a given country.
+
+How to create a new Rule programmatically?
+''''''''''''''''''''''''''''''''''''''''''
+
+Creating a **Rule** is really simple since we have the `RuleFactory <https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Factory/RuleFactory.php>`_.
+It has dedicated methods for creating all types of rules available by default.
+
+In the example you can see how to create a simple Cart Quantity rule. It will check if there are at least 5 items in the cart.
+
+.. code-block:: php
+
+   /** @var RuleFactoryInterface $ruleFactory */
+   $ruleFactory = $this->container->get('sylius.factory.promotion_rule');
+
+   $quantityRule = $ruleFactory->createCartQuantity('5');
+
+   // add your roule to the previously created Promotion
+   $promotion->addRule($quantityRule);
+
+.. note::
+
+   **Rules** are just constraints that have to be fulfilled by an order to make the promotion **eligible**.
+   To make something happen to the order you will need **Actions**.
 
 Promotion Actions
 -----------------
 
-...
+What happens with the Order when the rules of a Promotion are fulfilled - this is an **Action**.
 
-Coupons
--------
+There are a few kinds of actions in **Sylius**:
 
-...
+* fixed discount on the order (for example: -5$ discount)
+* percentage discount on the order (for example: -10% on the whole order)
+* fixed unit discount (for example: -1$ for one specific Mug)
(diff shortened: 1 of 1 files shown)


docs/book/promotions.rst
```diff
@@ -4,37 +4,126 @@
 Promotions
 ==========
 
-...
+The system of **Promotions** in **Sylius** is really flexible. It is a combination of promotion rules and actions.
+
+Promotions are assigned to Orders. They have a few parameters - a unique ``code``, ``name``, ``usageLimit``,
```
Zales0123: ``an unique`` 😄

pjedrzejewski: This is confusing, I would remove "Promotions are assigned to Orders.". We should explain that separately when describing the "applying" logic.

docs/book/promotions.rst
```diff
@@ -4,37 +4,126 @@
...
+
+.. code-block:: php
+
+   /** @var PromotionInterface $promotion */
+   $promotion = $this->container->get('sylius.factory.promotion')->createNew();
+
+   $promotion->setCode('simple_promotion_1');
+   $promotion->setName('Simple Promotion');
+
+   $this->container->get('sylius.repository.promotion')->add($promotion);
```
pjedrzejewski: Would be better to create a simple promotion with action and then add it repo, this code does nothing for the shop currently. ;)

TheMadeleine: Yes, but... adding codes and actions is described later ;>

docs/book/promotions.rst
```diff
@@ -4,37 +4,126 @@
...
+How to create an Action programmatically?
+'''''''''''''''''''''''''''''''''''''''''
 
-...
+In order to create a new Action we can use the dedicated `ActionFactory <https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Factory/ActionFactory.php>`_.
 
-Examples
---------
+It has special methods for creating all types of actions available by default.
+In the example below you can how to create a simple Fixed Discount action, that reduces the total of an order by 10$.
```
NeverResponse: `In the example below you can how to create` I'm sorry, I can what?

docs/book/promotions.rst
```diff
@@ -4,37 +4,173 @@
...
+
+.. code-block:: php
+
+   /** @var PromotionInterface $promotion */
+   $promotion = $this->container->get('sylius.factory.promotion')->createNew();
+
+   $promotion->setCode('simple_promotion_1');
+   $promotion->setName('Simple Promotion');
+
+**Of course an empty promotion would be useless** it is just a base for adding **Rules** and **Actions**.
```
pamil: would be useless **-** it is just

docs/book/promotions.rst
```diff
@@ -4,37 +4,173 @@
...
 Promotion Actions
 -----------------
 
-...
+What happens with the Order when the rules of a Promotion are fulfilled - this is an **Action**.
+
+There are a few kinds of actions in **Sylius**:
+
+* fixed discount on the order (for example: -5$ discount)
+* percentage discount on the order (for example: 10% discount)
```
tuka217: Below you use `-10%`, so maybe here should be minus too?

docs/book/promotions.rst
```diff
@@ -4,37 +4,173 @@
...
+How to create an Action programmatically?
+'''''''''''''''''''''''''''''''''''''''''
 
-Coupons
--------
+In order to create a new Action we can use the dedicated `ActionFactory <https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Factory/ActionFactory.php>`_.
 
-...
+It has special methods for creating all types of actions available by default.
+In the example below you can how to create a simple Fixed Discount action, that reduces the total of an order by 10$.
```
tuka217: (...) you can **learn** hot to (...) or something like that ?

docs/book/promotions.rst
```diff
@@ -4,37 +4,173 @@
...
+   $promotion->setName('Simple Promotion');
+
+**Of course an empty promotion would be useless** it is just a base for adding **Rules** and **Actions**.
+Let's see how to make it functional.
 
 Promotion Rules
 ---------------
 
-...
+The promotion **Rules** are methods of checking if a promotion should be applied.
```
NeverResponse: `The promotion **Rules** restrict in what circumstances a promotion will be applied.`
Checking whether it is, takes place in an appropriate **RuleChecker**.

pjedrzejewski:
Nice work Magda! :) This piece of doc made me think that these custom factory methods are a bit confusing. Maybe we should have a look at them before 1.0. :)

rvanlaak:
After this docs merge I still have problems upgrading to `beta.1`. This probably is because of making the forms `final`. We only use the `PromotionBundle` and did "override" the `RuleType` and `ActionType`, but keep on getting the following exception:

> Promotion action service "item_total" does not exist, available promotion action services: "invite_status_change", "fixed_discount", "generate_coupon", "percentage_discount", "tag", "send_mail"

After diving into it, this seems to have something to do with the form subscriber that adds the items tagged as `sylius.promotion_rule_checker` from the registry. There is [this fixme that's probably related](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/PromotionBundle/Form/EventListener/AbstractConfigurationSubscriber.php#L110).

The weird thing is that `item_total` is a default rule provided with the bundle, but the error says that the action service does not exist. Seems something goes wrong with the registry?

TheMadeleine:
Hi @rvanlaak!

The way of customizing forms has changed lately - have you checked if you are doing it in the new way? See [here](http://docs.sylius.org/en/latest/customization/form.html).

rvanlaak:
@TheMadeleine thanks, yes I did follow that. The weird thing with the error is that somehow actions are confused with rules:

> Promotion action service "item_total" does not exist

`item_total` actually isn't an action, but is a `rule` and is a default one in Sylius. The related classes (`ItemTotalRuleChecker`, `ItemTotalConfigurationType`) do exist and are registered to the container.

rvanlaak:
After debugging this a bit further, it seems the `ServiceRegistry` attached to the `Subscriber` is the wrong one:

```
In AbstractConfigurationSubscriber.php line 114:
array:3 [▼
  0 => "item_total"
  1 => ServiceRegistry {#2714 ▼
    -services: array:6 [▶]
    -interface: "Sylius\Component\Promotion\Action\PromotionActionCommandInterface"
    -context: "promotion action"
  }
  2 => "Sylius\Bundle\PromotionBundle\Form\EventListener\BuildPromotionRuleFormSubscriber"
]
```

In other words, the service definition of the `RuleFormSubscriber` is wrong, and the right registry has to be injected:

```xml
<service id="sylius.form.event_subscriber.build_promotion_rule" class="Sylius\Bundle\PromotionBundle\Form\EventListener\BuildPromotionRuleFormSubscriber">
        <argument type="service" id="sylius.registry_promotion_action" />
        <argument type="service" id="form.factory" />
</service>
```
See [service definition of `BuildPromotionRuleFormSubscriber`](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/PromotionBundle/Resources/config/services/forms.xml#L76). This has to be changed to `sylius.registry_promotion_rule_checker` if I'm correct (ping @pjedrzejewski )

pjedrzejewski:
They are switched indeed, the question is ... how does it work now? :D

rvanlaak:
Don't think this works (or maybe the demo "add rule" page shows the "actions"? :)

see https://github.com/Sylius/Sylius/pull/7082

lchrusciel:
They are redefined in [core](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Resources/config/services/form.xml#L142-L151), thats why they work.

rvanlaak:
Ah yeah, in the `demo` they do work:

image

Instead of redefining, I think core should decorate them right?

rvanlaak:
Next one up in upgrading `promotions` is passing "currency" as option to the form (read: configuration fields). As we don't have configured the channels concept of Sylius, but simply render the `FormType`, how will this work?

It apparently is not as simple as:

```php
$form = $this->createForm(RuleType::class, $rule, [
    'currency' => 'EUR',
]);
```

PR #5948, opened 2016-09-01
[UI] Remove flags near to locale name

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/grids/locale.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/grids/locale.yml
@@ -12,11 +12,8 @@ sylius_grid:
                     type: string
                     label: sylius.ui.code
                 name:
-                    type: twig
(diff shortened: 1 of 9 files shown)


pjedrzejewski:
Why?

GSadee:
Because locales are not always linked with the countries. For example, we have an issue that there was no flag for locale `Africaans`, so we decided that locales shouldn't  have flags.

michalmarcinkowski:
Right now I have second thoughts about that, because we display flags on translation forms and it prettify our UI and is a good UX. Maybe we could leave current logic, so display flags of locales that are connected with country and don't display for locales like `Africaans`. @GSadee @pjedrzejewski WDYT?

GSadee:
So.. we should leave it as it is now 😃

Issue #5222, opened 2016-06-09
[RFC] A way to change response code and/or body of ResourceController::updateAction 


koemeet:
Right now the `updateAction` is forcing a `204` and provides you with an empty `View` object (no data in it). Before you had the resource with it and you could for example override the `ViewHandler` and manipulate the response to be `200` with the serialized resource with it. 

For my API that I build around all Sylius resources I am in need to have the whole resource serialized in the payload when you update it, instead of a `HTTP 204 No Content`. 

Before this change happend, I achieved it by doing the following in a decorated view handler:

``` php
/**
 * {@inheritdoc}
 */
public function handle(RequestConfiguration $requestConfiguration, View $view)
{
    if (Response::HTTP_NO_CONTENT === $view->getStatusCode() && null !== $view->getData()) {
        $view->setStatusCode(Response::HTTP_OK);
    }

    return $this->viewHandler->handle($requestConfiguration, $view);
}
```

It is not the cleanest solution, but it worked for me. Right now I am stuck since the whole data property of the view is `null`.

This problem was introduced by PR https://github.com/Sylius/Sylius/pull/5161.

bendavies:
how about supporting returning responses from the events?
https://github.com/Sylius/Sylius/blob/6fa42871f1e22a777044928cadd2e9c94ede48ae/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php#L314

bendavies:
note that @michalmarcinkowski says at the bottom of #5161 that it could be made configurable in the future.

pjedrzejewski:
@bendavies That would be cool for other reasons too! :)

@steffenbrem What if we make it configurable? I see some usecases for getting 200 with a resource after update, for mobile api responses where number of requests is quite important. Would be perfect to configure it globally and per route. Although I am not too happy about adding more stuff to ``_sylius`` config. :D

bendavies:
@pjedrzejewski sorry, i removed the comment. since you replied, what i said was, what about supporting returning responses from events.

steffenbrem:
@bendavies I think that is the best way to implement it in the ResourceController indeed. Something like a `ResourceResponse` event?

bendavies:
`ResourceControllerEvent` is already used. what about adding an optional response?

steffenbrem:
What do you mean exactly with an optional response?

bendavies:
[...]

PR #5950, opened 2016-09-01
[RFC] [Resource] Change code of API response for state machine transition

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

`applyStateMachineTransitionAction` makes an update on resource, so 204 response code would be much more appropriate.


michalmarcinkowski:
I think that returning updated entity after state machine transition can be useful. Perhaps we could make it configurable later see past discussions https://github.com/Sylius/Sylius/pull/5161, https://github.com/Sylius/Sylius/issues/5222.

lchrusciel:
@michalmarcinkowski ok, closing then

PR #5951, opened 2016-09-01
[BugFix][Core] Added shipping method eligibility checker to core methods resolver

### LINKED ISSUE #5861: Shipping categories ?
Hi
I noticed that in the 1.0 branch, products' shipping categories are not taken into account when checking out the cart — the rules of each shipping categories are not taken into account and all shipping methods are displayed, regardless of the products

Is this normal ? Is there a new configuration / thing to do to make this work as intended ?

thanks

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no|yes
| BC breaks?      | no|yes
| Related tickets | fixes #5861
| License         | MIT

Check for shipping method eligibility when getting shipping methods based on zone and channel.

### DIFF
--- a/src/Sylius/Component/Core/spec/Resolver/ZoneAndChannelBasedShippingMethodsResolverSpec.php
+++ b/src/Sylius/Component/Core/spec/Resolver/ZoneAndChannelBasedShippingMethodsResolverSpec.php
@@ -12,6 +12,7 @@
 namespace spec\Sylius\Component\Core\Resolver;
 
 use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
 use Sylius\Component\Addressing\Matcher\ZoneMatcherInterface;
 use Sylius\Component\Addressing\Model\ZoneInterface;
 use Sylius\Component\Core\Model\AddressInterface;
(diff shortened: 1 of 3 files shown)


src/Sylius/Component/Core/spec/Resolver/ZoneAndChannelBasedShippingMethodsResolverSpec.php
```diff
@@ -49,6 +52,7 @@ function it_implements_shipping_methods_by_zones_and_channel_resolver_interface(
     }
 
     function it_returns_shipping_methods_matched_for_shipment_order_shipping_address_and_order_channel(
+        $eligibilityChecker,
```
Zales0123: Could you please use type hints also for constructor dependencies mocks?

src/Sylius/Component/Core/spec/Resolver/ZoneAndChannelBasedShippingMethodsResolverSpec.php
```diff
@@ -87,9 +94,41 @@ function it_returns_empty_array_if_zone_matcher_could_not_match_any_zone(
 
         $zoneMatcher->matchAll($address)->willReturn([]);
 
+        $eligibilityChecker->isEligible(Argument::any(), Argument::any())->willReturn(true);
```
Zales0123: I think we could use mocks here rather than ``Argument::any()``.

src/Sylius/Component/Core/spec/Resolver/ZoneAndChannelBasedShippingMethodsResolverSpec.php
steffenbrem: Oh I didn't know that was possible, nice!

steffenbrem:
Ready.

pjedrzejewski:
Thank you Steffen! We should add scenarios for this functionality as it has not been migrated from the "legacy" properly.

Issue #5886, opened 2016-08-25
[Theme] Exception tab at symfony profiler doesn't work


lchrusciel:
1. Make a 404 error
2. Open symfony profiler
3. Open exception tab
4. See this:
<img width="1034" alt="zrzut ekranu 2016-08-25 o 14 33 21" src="https://cloud.githubusercontent.com/assets/6213903/17969160/0706a5ec-6ad1-11e6-84bb-95ea63540a42.png">

PR #5958, opened 2016-09-02
[Theme] Twig loader quick fix to avoid profiler exception

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/5886
| License         | MIT


pjedrzejewski:
Nice, thanks Mateusz! :)

tchapi:
This breaks templating ... and **Twig_ExistsLoaderInterface** is deprecated since 1.12.

Apart from the deprecation, wouldn't something like that be a bit better for now ? At least not BC breaking things :
```php
try {
    return stat($this->findTemplate($name)) !== false;
} catch (\Exception $exception) {
    return false;
}
```

ahilles107:
👍  for @tchapi chapi solution. We really need to check if this file exists there. It breaks default fallback for error.html.twig template in error pages.

PR #5969, opened 2016-09-03
5968 coupon flash message fix

### LINKED ISSUE #5968: Promotion Flash Messages
Whenever i add a promotion to the cart i get a number of messages that are unnecessary. On thank you page i'll get 3 flashes saying the same thing 
capture2

When adding the promotion to the cart i seem to get an invalid message for all other coupon based promotions even though i didn't try and add themcapture3

If i remove one of the promotions from the channel i get 1 less error message

bretto36:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #5968 
| License         | MIT

Removed the message that displays when a coupon doesn't belong to an active promotion.


src/Sylius/Component/Promotion/Checker/CouponsEligibilityChecker.php
```diff
@@ -46,8 +46,6 @@ public function isEligible(PromotionSubjectInterface $subject, PromotionInterfac
         }
 
         if (!$this->isCouponEligible($promotion, $subject)) {
-            $this->eventDispatcher->dispatch(SyliusPromotionEvents::COUPON_NOT_ELIGIBLE, new GenericEvent($promotion));
```
bretto36: I feel this is unneeded, as it does not need to display a message when the coupon does not belong to one of the active promotions. Just needs to return false.

bretto36:
Can anyone tell me where i go to change the behat tests?

src/Sylius/Component/Promotion/Checker/CouponsEligibilityChecker.php
gabiudrescu: I think this could be used to issue messages like: in order to use this coupon, you need to comply with these rules: 
- have O(n) product(s) from taxon ZXY
- cart total should be greater than

ofc, this is something I need in my project, most probably as a userLand implementation.

bretto36: This isn't anything to do with rules though. This is just checking to ensure the code you entered is the same as the one attached to the promotion. 

We definitely need a change to the Rules section though to add what you are proposing.

bretto36:
Pretty sure this has been fixed with some of the pull requests done by @pamil. Thank you @pamil

PR #5970, opened 2016-09-03
[RFC] [POC] [WIP] [Cart] Introduce CartModifier service

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Related tickets | 
| License         | MIT

## Concept
Small proof of concept / RFC. in this PR I have extracted logic of adding a cart item from CartItemController to a separate service. This service will be responsible for determining if the new cart item should be added or just increase some of existing one, and then it will recalculate the whole cart. 

## Profit
Such a solution ensure us, that we have only one true way of adding new items, when we are sure that whole cart will be recalculated after each addition. It can be also useful I one would like to add one route with `Order now` feature. 

I would like also to decouple Sylius logic as much as it is possible form symfony events. They are nice, but only for end users. I am pretty sure, that managing this kind of logic with regular service is much better approach.

The problem with the previous implementation is, that order recalculation is handled via Cart event `SyliusCartEvents::CART_CHANGE`.

### TODOs for now:
 - [ ] Extract cart item removal

### Ideas for next iteration
I think that cart clearing logic could be also added to this service.


src/Sylius/Component/Core/Modifier/CartModifierInterface.php
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @author Łukasz Chrusciel <lukasz.chrusciel@lakion.com>
+ */
+interface CartModifierInterface
+{
+    /**
+     * @param OrderInterface $cart
+     * @param CartItemInterface $cartItem
+     */
+    public function addToCart(OrderInterface $cart, CartItemInterface $cartItem);
```
Arminek: ``OrderInterface`` :dancer:

src/Sylius/Component/Cart/Modifier/CartModifier.php
```diff
@@ -0,0 +1,83 @@
...
+    {
+        $cart->removeItem($item);
+        $this->orderProcessor->process($cart);
+    }
+
+    /**
+     * @param CartInterface $cart
+     * @param CartItemInterface $item
+     */
+    private function resolveCartItem(CartInterface $cart, CartItemInterface $item)
```
pamil: Does it only resolve the cart item? Or resolves and merges with the passed one if found? It would probably need some better naming.

lchrusciel: Any ideas?

lchrusciel: As far as we won't came up with a better idea, I think it could be merged and I will create an issue with naming.

src/Sylius/Bundle/CartBundle/Controller/CartItemController.php
```diff
@@ -61,15 +60,15 @@ public function addAction(Request $request)
...
+            $this->getCartModifier()->addToCart($cart, $newResource);
 
             $this->eventDispatcher->dispatchPostEvent(ResourceActions::CREATE, $configuration, $newResource);
 
             if (!$configuration->isHtmlRequest()) {
                 return $this->viewHandler->handle($configuration, View::create($newResource, Response::HTTP_CREATED));
             }
 
-            $this->getEventDispatcher()->dispatch(SyliusCartEvents::CART_CHANGE, new GenericEvent($cart));
+            $this->getEventDispatcher()->dispatch(SyliusCartEvents::ITEM_ADDED_TO_CART, new GenericEvent($cart));
```
michalmarcinkowski: I think it was correct. We already dispatch order item `create` event and that should be enough, but we still need `cart_change` event to be dispatched on each cart modification (also add/remove something) at least for now.

michalmarcinkowski:
Good job Łukasz! 👍

PR #5971, opened 2016-09-03
[RFC] [OrderComponent] OrderProcessorInterface moved to order component.

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

`CompositeOrderProcessor` doesn't have any external dependencies, so it could be useful to keep it inside an `Order` component instead of core. Same with `OrderProcessorInterface`. 

When it'll be merged, `CartModifier` from #5970 could be moved to `CartBundle` from the Core also.


pjedrzejewski:
Thank you Łukasz!

PR #5972, opened 2016-09-03
[Cart] Fix event and flush order in CartItemController

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

I'm pretty sure, that this event should be dispatched also for not html requests.


pjedrzejewski:
Thanks Łukasz!

PR #5976, opened 2016-09-05
[Promotion] Use proportional distribution in order percentage discount

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Core/Promotion/Action/PercentageDiscountAction.php
+++ b/src/Sylius/Component/Core/Promotion/Action/PercentageDiscountAction.php
@@ -11,7 +11,9 @@
 
 namespace Sylius\Component\Core\Promotion\Action;
 
-use Sylius\Component\Core\Distributor\IntegerDistributorInterface;
+use Sylius\Component\Core\Distributor\ProportionalIntegerDistributorInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Model\OrderItemInterface;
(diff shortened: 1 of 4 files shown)


michalmarcinkowski:
We should add more scenarios here. The order percentage discount should also take into account shipping fee.

src/Sylius/Component/Core/Promotion/Action/PercentageDiscountAction.php
```diff
@@ -63,7 +66,12 @@ public function execute(PromotionSubjectInterface $subject, array $configuration
             return;
         }
 
-        $splitPromotion = $this->distributor->distribute($promotionAmount, $subject->countItems());
+        $itemsTotal = [];
+        foreach ($subject->getItems() as $orderItem) {
+            $itemsTotal[] = $orderItem->getTotal();
+        }
```
michalmarcinkowski: There is `getItemsTotal()` method on `Order`

pamil: It returns the sum of items totals, not the array with items totals.

pamil:
@michalmarcinkowski there's "Receiving percentage discount does not affect the shipping fee" scenario, so we should rather discuss on it.

michalmarcinkowski:
@pamil that's great! I recall we simplified that, so you should configure shipping percentage discount separately. Good job then! 👍

PR #5991, opened 2016-09-06
[Settings] Do not inject container to ParameterTransformerListener

pamil:


michalmarcinkowski:
👍

bendavies:
@pamil this is causing an error in my app after updating settings bundle.
```
  [Symfony\Component\DependencyInjection\Exception\ServiceCircularReferenceException]
  Circular reference detected for service "doctrine.dbal.default_connection", path: "doctrine.dbal.default_connection".
```
I imagine this  is why the container was injected in the first place. Are you certain this does not break sylius when integrated? Will con consider reverting?

Thanks

bendavies:
I'm guessing one of my services somewhere must be made lazy...

bendavies:
I thought making my custom settings schema lazy would fix this (it has doctrine repository deps) but it doesn't work. if you ave any idea @pamil, let me know!

pamil:
Ouch, that should be reverted indeed. Tests were passing, because SettingsBundle is no longer used in Sylius.

bendavies:
thanks @pamil what does sylius use for settings now?

PR #5994, opened 2016-09-06
[Core] Improve factories usage in DI

pamil:
`factory-service` / `factory-method` on `service` is deprecated


michalmarcinkowski:
👍

PR #5995, opened 2016-09-06
[Core] Remove unused event listener

pamil:
There is no `customerUpdateEvent` method anywhere


michalmarcinkowski:
But there should be! We should add a scenario for changing password of a registered customer and verify it has changed (I recall we should have scenario for that, but couldn't find it anywhere). Looks like it doesn't work now...

michalmarcinkowski:
Ping @pamil :)

pamil:
@michalmarcinkowski extracting an issue out of this PR.

PR #6000, opened 2016-09-07
[Promotion] Clean up promotions handling

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | -
| License         | MIT


docs/components/Promotion/basic_usage.rst
```diff
@@ -196,7 +196,7 @@ Below you can see how it works:
      * @param ServiceRegistryInterface $registry
      * @param EventDispatcherInterface $dispatcher
      */
-    $checker = new PromotionEligibilityChecker($checkerRegistry, $dispatcher);
+    $checker = new CompositePromotionEligibilityChecker($checkerRegistry, $dispatcher);
```
lchrusciel: `$checker = new CompositePromotionEligibilityChecker($checkerRegistry);`

src/Sylius/Bundle/CoreBundle/Resources/config/services/promotion.xml
```diff
@@ -0,0 +1,79 @@
...
+ (c) Paweł Jędrzejewski
+
+ For the full copyright and license information, please view the LICENSE
+ file that was distributed with this source code.
+
+-->
+
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <service id="sylius.active_promotions_by_channel_provider" class="Sylius\Component\Core\Provider\ActivePromotionsByChannelProvider">
```
lchrusciel: We should start to clean service name conventions. Even in this file we have two different ones. Can we be change some of them to keep some consistency? 

`sylius.promotion.active_promotions_by_channel_provider`?

src/Sylius/Bundle/PromotionBundle/Form/DataTransformer/CouponToCodeTransformer.php
```diff
@@ -20,26 +20,22 @@
...
      * @var ObjectRepository
      */
-    protected $couponRepository;
+    private $couponRepository;
 
     /**
      * @var EventDispatcherInterface
      */
-    protected $dispatcher;
+    private $dispatcher;
```
lchrusciel: Do we still need it?

michalmarcinkowski:
Thanks Kamil! 👍  Please apply comments in a separate PR.

rvanlaak:
`generateUniqueCode` visibility changed from public > private, which is a BC break. 

While I'm okay with making BC breaks as long as beta1 isn't released yet this wasn't documented in [UPGRADE.ml](https://github.com/Sylius/Sylius/blob/master/UPGRADE.md). 

Please document the new way of creating promotion codes by using a `InstructionInterface`.

pamil:
`generateUniqueCode` was in fact changed from `protected` to `private` (but that's still BC).

You can still use any `PromotionCouponGeneratorInterface` instance to generate coupons, the signature is as follows:

```php
/**
 * @param PromotionInterface $promotion
 * @param PromotionCouponGeneratorInstructionInterface $instruction
 *
 * @return array of generated coupons with coupon code as a key
 */
public function generate(PromotionInterface $promotion, PromotionCouponGeneratorInstructionInterface $instruction);
```

rvanlaak:
Oh wow, I came from `0.17` yesterday so before then it still was public: https://github.com/pamil/Sylius/commit/4c9c6a01368defc08dbab809679b070785bd43b6#diff-b933567465782f701b9cadb26f36d5d7L77

Still, it would be nice if such changes are documented, but I think right now too many BC breaks are made to document all of them right? Are there any new docs on how to create coupons? What's Sylius's point of view on this, because I don't want to write libraries on all the Sylius classes I've integrated ;-)

pamil:
`0.17` feels like it was a lifetime ago :)

There were so many BC breaks introduced recently (classes made `final`; removing unused bundles, components and single classes), we got brand new frontend etc., so documenting them was quite tremendous task. I guess that we will improve it since we reach 1.0.0-alpha (which is really, really soon). Nevertheless, if you rely on interfaces to customise the application behaviour, the BC breaks shouldn't be so crucial :)

@TheMadeleine keeps up with extremely good work documenting Sylius so there indeed is a [documentation about generating coupons](http://docs.sylius.org/en/latest/book/coupons.html).

rvanlaak:
Ghehe yes `0.17` actually was ages ago ;) I'm only using the `Promotion` bundle for now so the errors I get quite often are related to missing classes :)

I'm kinda used to follow upgrade logs, but didn't see any mentions about the changes to the `Promotion` component. Actually I did try upgrading a couple of months before, which didn't work out then because I had to do too much maintenance work for it. Over the last days I actually also did take a look in the commits made to the `Promotion` component, but somehow missed this one. So it would be awesome to see a docs hackday ghehe ;)

I'll do a hotfix now, good luck finishing `1.0.0-alpha` in the next few days guys!

PR #6014, opened 2016-09-07
[Core] Remove leftovers from SettingsBundle usage

pamil:


src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml
```diff
@@ -40,7 +40,6 @@ doctrine:
 
 doctrine_cache:
     providers:
-        sylius_settings: "%sylius.cache%"
         sylius_rbac: "%sylius.cache%"
```
michalmarcinkowski: This should also be removed :)

michalmarcinkowski:
👍

Issue #5905, opened 2016-08-26
[ ResourceBundle ] - Criteria not really filtered if the option paginate false is defined


jvahldick:
There is something missing on https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ResourceBundle/Controller/ResourcesResolver.php#L33

If the route has some criteria, it isn't really filtered when the option paginate: false is defined.

bretto36:
You should make a Pull Request for this

jvahldick:
I'll fix this night together with #5857

michalmarcinkowski:
Fixed.

PR #6017, opened 2016-09-07
[ResourceBundle] fixing #5857 #5905 - missing checkage of sortable and filterable items

### LINKED ISSUE #5857: Bug in ResourcesResolver for unpaginated sorting
I'm not sure if you're even using this anymore (might be using the Grid version exclusively), but this doesn't work:

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ResourceBundle/Controller/ResourcesResolver.php#L33

If you have a routing configuration such as:

```
        _sylius:
            template: SyliusWebBundle:Backend/Permission:index.html.twig
            paginate: false
            sorting:
                left: asc
```

The `sorting` will not be applied.

I think the `findAll()` use here isn't required, since this just wraps `findBy()` anyway, the case below should catch everything.

jvahldick:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| Related tickets | fixes #5857 #5905 |
| License | MIT |


src/Sylius/Bundle/ResourceBundle/Controller/ResourcesResolver.php
```diff
@@ -30,11 +30,16 @@ public function getResources(RequestConfiguration $requestConfiguration, Reposit
             return $resources;
         }
 
-        if (!$requestConfiguration->isPaginated() && !$requestConfiguration->isLimited()) {
+        if (
+            ! $requestConfiguration->isPaginated() &&
```
lchrusciel: We should revert a conditions, here and use `OR` instead of `NOT AND`. It will reduce number of require comparisons and reduce complicity. 

Another thing is some extraction. Maybe some private method? But I have problem with naming for it.

src/Sylius/Bundle/ResourceBundle/Controller/ResourcesResolver.php
```diff
@@ -30,11 +30,16 @@ public function getResources(RequestConfiguration $requestConfiguration, Reposit
...
+            ! $requestConfiguration->isLimited() &&
+            ! $requestConfiguration->isFilterable() &&
+            ! $requestConfiguration->isSortable()
+        ) {
             return $repository->findAll();
         }
 
-        if (!$requestConfiguration->isPaginated()) {
+        if ( ! $requestConfiguration->isPaginated()) {
```
pjedrzejewski: This is incorrect according to our CS.

pjedrzejewski:
Hi Jorge! Thank you for this patch, could you fix the small CS issue, please? Then it is good to merge. :) 👍

src/Sylius/Bundle/ResourceBundle/Controller/ResourcesResolver.php
jvahldick: I thought the same.
Perhaps something like:
$this->hasFilterRequirements($requestConfiguration)
or $this->isDefaultResourceRequest($requestConfiguration)

ping @lchrusciel 
please let me know if you have some idea or if one of these names fit for this purpose

lchrusciel: `hasSpecialConditions`?

jvahldick:
I'm not sure about the private method.
I was thinking this method on RequestConfiguration level, probably it will be more useful.

What is your suggestion?

lchrusciel:
I think this one could be merged, and we can rethink about moving or renaming of this function later

michalmarcinkowski:
@jvahldick could you rebase?

jvahldick:
Hey guys,

I double checked it, and I would need to know some business rules to changing something here.

What is the correct behavior if I have:
```yaml
example_index:
   defaults:
      _sylius:
         paginate: false
         filterable: false/true
         criteria:
            enable: true

         sortable: false/true
         sorting:
            id: desc
```

Currently, it is not applying the criteria but it is sorting it.
If we change:
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ResourceBundle/Controller/ResourcesResolver.php#L34

To the code bellow, we can have a better result.
```php
return $repository->findBy($requestConfiguration->getCriteria(), $requestConfiguration->getSorting());
```

Particularly, I don't feel comfortable with these controversial parameters. 
Should the parameters "sortable" and "filterable" be used just for parameters coming from the Request object? Or should we change it also on RequestConfiguration to these parameters deal for both cases?

The current fix (with hasCriteriaRequest method) solves this problem, but I am not sure if it is really what you are expecting.
Let me know which is better for you and I can make the necessary changes.

michalmarcinkowski:
Actually we have two cases:
1. Paginate true
2. Paginate false

But filterable and sortable behavior is common. Currently we don't take into account if filterable is `false` or sortable `false`, so how about the following:

```php
    ...

    $criteria = [];
    if ($requestConfiguration->isFilterable()) {
        $criteria = $requestConfiguration->getCriteria();
    }

    $sorting = [];
    if ($requestConfiguration->isSortable()) {
        $sorting = $requestConfiguration->getSorting();
    }

    if (!$requestConfiguration->isPaginated()) {
            return $repository->findBy($criteria, $sorting, $requestConfiguration->getLimit());
    }

    return $repository->createPaginator($criteria, $sorting);
```

The above snippet will not apply criteria or sorting when disabled.

lchrusciel:
Just a small improvement:
```php
    if ($requestConfiguration->isPaginated()) {
         return $repository->createPaginator($criteria, $sorting);
    }

    return $repository->findBy($criteria, $sorting, $requestConfiguration->getLimit());
```

to avoid reversed condition ;)

jvahldick:
I agree with this approach, it's more clear uses the boolean parameters for both cases.

Congrats for the project, it is really amazing
love you guys hahaha

michalmarcinkowski:
@jvahldick thanks! Could you also cover introduced changes with some specs to prevent from regression?

lchrusciel:
So. This behaviour should be expected. We had bad assumptions. We have to turn on sorting on this endpoint. @jvahldick I have open a PR with a simple fix. Merge it, or just copy to your solution. it is up to you. Anyway, now it is working as expected :) thanks!

michalmarcinkowski:
Thank you Jorge! 👍

PR #6019, opened 2016-09-08
[Core] Remove invalid HwiOAuth integration

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no (already not working :D)
| Related tickets | -
| License         | MIT

Injected services `sylius.factory.admin_user_oauth` and `sylius.repository.admin_user_oauth` do not exist, so we cannot use OAuth User Provider.


pamil:
Looks like it's named `admin_oauth_user`, not `admin_user_oauth`, so I'll try to fix this.

PR #6025, opened 2016-09-08
[Core] Remove services manipulation in the extension in favour of XML definition

pamil:


src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
```diff
@@ -408,11 +408,15 @@
             <tag name="validator.constraint_validator" alias="registered_user_validator" />
         </service>
 
-        <service id="sylius.custom_factory.cart_item" class="Sylius\Component\Core\Factory\CartItemFactory" decorates="sylius.factory.cart_item">
-            <argument type="service" id="sylius.custom_factory.cart_item.inner" />
+        <service id="sylius_core.factory.cart_item" class="Sylius\Component\Core\Factory\CartItemFactory" decorates="sylius.factory.cart_item" public="false">
```
lchrusciel: Why did you introduce another service definition pattern? We have too many of them already

pamil: The `sylius.custom_factory.*` pattern could be good, but it may lead to some conflicts if a service is being decorated in two different places.

lchrusciel: If we have `sylius` as first key word for the whole system, I think we should stay with it, and come up with some better naming after first dot

PR #6036, opened 2016-09-08
[Core] Clean up TaxCalculationStrategy

pamil:


src/Sylius/Bundle/CoreBundle/Taxation/Strategy/TaxCalculationStrategy.php
```diff
@@ -12,19 +12,79 @@
...
+    /**
      * {@inheritdoc}
      */
     public function supports(OrderInterface $order, ZoneInterface $zone)
     {
-        return $order->getChannel()->getTaxCalculationStrategy() === $this->type;
+        $channel = $order->getChannel();
+
+        /** @var ChannelInterface $channel */
+        Assert::isInstanceOf($channel, ChannelInterface::class);
```
lchrusciel: How it is possible that `getChannel` method will return something different then instance of `ChannelInterface`? Maybe `Assert::notNull` would be a better solution.

pjedrzejewski:
Thank you Kamil! :)

src/Sylius/Bundle/CoreBundle/Taxation/Strategy/TaxCalculationStrategy.php
michalmarcinkowski: Order should always be assigned to channel.

pamil: We are using `getTaxCalculationStrategy` which exists only in Core's `ChannelInterface`.

PR #6062, opened 2016-09-12
[Admin][Order] Modifying a customer's shipping address on an order

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | 
| License         | MIT

This PR contains #6122

### DIFF
--- a/features/order/managing_orders/modifying_shipping_address/modifying_shipping_address.feature
+++ b/features/order/managing_orders/modifying_shipping_address/modifying_shipping_address.feature
@@ -1,11 +1,11 @@
-@managing_orders
+@modifying_shipping_address
 Feature: Modifying a customer shipping address after an order has been placed
     In order to ship an order to a correct place
     As an Administrator
     I want to be able to modify a customer's shipping address after an order has been placed
 
     Background:
-        Given the store operates on a single channel in "United States" named "Web"
+        Given the store operates on a single channel in the "United States" named "Web"
         And that channel uses the "USD" currency by default
         And the store ships everywhere for free
         And the store allows paying with "Cash on Delivery"
@@ -19,8 +19,13 @@ Feature: Modifying a customer shipping address after an order has been placed
     @todo
     Scenario: Modifying a customer's shipping address
         When I view the summary of the order "#00000001"
-        And I want to modify a customer's shipping address
-        And I specify their shipping address as "Los Angeles", "Seaside Fwy", "90802", "United States" for "Lucifer Morningstar"
+        And I want to modify a customer's shipping address of this order
+        And I specify the first name as "Lucifer"
+        And I specify the last name as "Morningstar"
+        And I specify the street as "Seaside Fwy"
+        And I choose "United States" as the country
+        And I specify the city as "Los Angeles"
+        And I specify the postcode as "90802"
         And I save my changes
         Then I should be notified that it has been successfully edited
         And this order should be shipped to "Lucifer Morningstar", "Seaside Fwy", "90802", "Los Angeles", "United States"
@@ -29,7 +34,7 @@ Feature: Modifying a customer shipping address after an order has been placed
     Scenario: Modifying a customer's shipping address when a product's price has been changed
         Given the product "Suit" changed its price to "$300.00"
         When I view the summary of the order "#00000001"
-        And I want to modify a customer's shipping address
+        And I want to modify a customer's shipping address of this order
         And I specify their shipping address as "Los Angeles", "Seaside Fwy", "90802", "United States" for "Lucifer Morningstar"
         And I save my changes
         Then I should be notified that it has been successfully edited
@@ -40,7 +45,7 @@ Feature: Modifying a customer shipping address after an order has been placed
     Scenario: Modifying a customer's shipping address when a channel has been disabled
         Given the channel "Web" has been disabled
         When I view the summary of the order "#00000001"
-        And I want to modify a customer's shipping address
+        And I want to modify a customer's shipping address of this order
         And I specify their shipping address as "Los Angeles", "Seaside Fwy", "90802", "United States" for "Lucifer Morningstar"
         And I save my changes
         Then I should be notified that it has been successfully edited
@@ -51,7 +56,7 @@ Feature: Modifying a customer shipping address after an order has been placed
     Scenario: Modifying a customer's shipping address when the currency has been disabled
         Given the currency "USD" has been disabled
         When I view the summary of the order "#00000001"
-        And I want to modify a customer's shipping address
+        And I want to modify a customer's shipping address of this order
         And I specify their shipping address as "Los Angeles", "Seaside Fwy", "90802", "United States" for "Lucifer Morningstar"
         And I save my changes
         Then I should be notified that it has been successfully edited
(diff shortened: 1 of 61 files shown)


features/order/managing_orders/modifying_shipping_address/modifying_shipping_address_on_order_with_promotion.feature
```diff
@@ -0,0 +1,31 @@
...
+        And the promotion gives "$50.00" off if order contains products classified as "Suits"
+        And there is a customer "mike@ross.com" that placed an order "#00000001"
+        And the customer bought a single "Suit"
+        And the customer "Mike Ross" addressed it to "350 5th Ave", "10118" "New York" in the "United States" with identical billing address
+        And the customer chose "Free" shipping method with "Cash on Delivery" payment
+        And I am logged in as an administrator
+
+    @todo
+    Scenario: Modifying a customer's shipping address when the applied promotion is no longer valid
+        Given the promotion was disabled for the channel "Web"
```
gabiudrescu: why is it relevant to disable the promotion from the channel if the purchase was made - and the cart became an order?

GSadee: Because we want to be sure, that changing a shipping address on a placed order won't affect the order totals.

gabiudrescu: so changing something on an order should not trigger the promotion calculator and get the totals changed, right? if so, then the promotion should still be active to testify that promotion calculator is not triggered on a shipping address change from admin.

maybe I'm missing something here.

michalmarcinkowski: For 1.0 we deliver simple shipping address edit feature which will only modify the shipping address and leave the rest of the order untouched. That means the order will not be recalculated and this scenario covers situation when the promotion that was applied to an order is disabled after order was placed. It ensures the promotion is still applied to an order after shipping address modification.

src/Sylius/Behat/Page/Admin/Order/UpdateShippingAddressPage.php
```diff
@@ -0,0 +1,93 @@
...
+    {
+        $this->getDocument()->selectFieldOption('Country', $country);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function specifyShippingAddress($city, $street, $postcode, $country, $firstAndLastName)
+    {
+        $customerName = explode(' ', $firstAndLastName);
```
Zales0123: I would vote for ``$customerNames``? Or sth like this, for sure a plural for as this variable contains an array not a single value.

features/order/managing_orders/modifying_shipping_address/modifying_shipping_address_on_order_with_different_exchange_rate.feature
```diff
@@ -0,0 +1,34 @@
...
+        And the store has a zone "English" with code "EN"
+        And this zone has the "United States" country member
+        And that channel allows to shop using the "USD" currency
+        And that channel allows to shop using the "GBP" currency with exchange rate 3.0
+        And that channel uses the "USD" currency by default
+        And the store allows paying with "Cash on Delivery"
+        And the store has "DHL" shipping method with "$20.00" fee within "EN" zone
+        And the store has a product "Suit" priced at "$400.00"
+        And there is a customer "mike@ross.com" that placed an order "#00000001"
+        And the customer have chosen to order in the "GBP" currency
```
TheMadeleine: ``... has chosen...``

features/order/managing_orders/modifying_shipping_address/modifying_shipping_address_on_order_with_taxes.feature
```diff
@@ -0,0 +1,30 @@
+@modifying_shipping_address
+Feature: Modifying a customer's shipping address on an order with taxes
+    In order to ship an order to a correct place
+    As an Administrator
+    I want to be able to modify a customer's shipping address without changing the order's total
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store ships everything for free within "US" zone
```
TheMadeleine: `` ...within the "US" zone``

features/order/managing_orders/modifying_shipping_address/modifying_shipping_address_on_order_with_taxes.feature
```diff
@@ -0,0 +1,30 @@
...
+        And the store allows paying offline
+        And the store has "VAT" tax rate of 20% for "Suits" within "US" zone
```
TheMadeleine: ``...within the "US" zone``

features/order/managing_orders/modifying_shipping_address/modifying_shipping_address_on_order_with_different_exchange_rate.feature
GSadee: I will fix it in a separate PR 👍

features/order/managing_orders/modifying_shipping_address/modifying_shipping_address_on_order_with_taxes.feature
GSadee: I will fix it in a separate PR 👍

features/order/managing_orders/modifying_shipping_address/modifying_shipping_address_on_order_with_taxes.feature
GSadee: As above

features/order/managing_orders/modifying_shipping_address/modifying_shipping_address.feature
```diff
@@ -0,0 +1,75 @@
+@modifying_shipping_address
+Feature: Modifying a customer shipping address after an order has been placed
+    In order to ship an order to a correct place
+    As an Administrator
+    I want to be able to modify a customer's shipping address after an order has been placed
+
+    Background:
+        Given the store operates on a single channel in the "United States" named "Web"
+        And that channel uses the "USD" currency by default
```
michalmarcinkowski: We should have step `Given the store operates on a single channel in "United States"` which will already create `USD` currency and make it a default currency of this channel.

features/order/managing_orders/modifying_shipping_address/modifying_shipping_address_on_order_with_taxes.feature
```diff
@@ -0,0 +1,30 @@
...
+        And it belongs to "Suits" tax category
+        And there is a customer "mike@ross.com" that placed an order "#00000001"
+        And the customer bought a single "Suit"
+        And the customer "Mike Ross" addressed it to "350 5th Ave", "10118" "New York" in the "United States" with identical billing address
+        And the customer chose "Free" shipping method with "Offline" payment
+        And I am logged in as an administrator
+
+    @ui
+    Scenario: Modifying a customer's shipping address when the applied promotion is no longer valid
+        Given the "VAT" tax rate is of 10%
```
michalmarcinkowski: Given the "VAT" tax rate has changed to 10%

features/order/managing_orders/modifying_shipping_address/modifying_shipping_address.feature
GSadee: Yes, we have, but I wanted to clearly see what is our currency, because we modify it later.

michalmarcinkowski:
Nice work Grzesiu! 👍

PR #6064, opened 2016-09-12
[BugFix][Core] Remove cascade transitions on sylius_shipment_unit

### LINKED ISSUE #5667: [ApiBundle][Checkout] Transition hold does not exist on object Sylius\\Component\\Core\\Model\\OrderItemUnit with graph sylius_shipment_unit
This errors occured when calling for the finalize step with PUT

```PUT /api/checkouts/{{orderId}}```

following the checkout full process here : [http://docs.sylius.org/en/latest/api/checkouts.html](http://docs.sylius.org/en/latest/api/checkouts.html)

Comes from https://github.com/Sylius/Sylius/pull/5606/files which removed the "hold" transition, I suppose...

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no|yes
| BC breaks?      | no|yes
| Related tickets | fixes #5667
| License         | MIT

I looked into issue #5667 and I think the best way to solve it is to remove the whole cascade of the transitions all together. This will prevent some nasty errors that are currently present in core.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine.yml
@@ -145,14 +145,6 @@ winzou_state_machine:
                     do: ["@sm.callback.cascade_transition", "apply"]
                     args: ["object", "event", "'cancel'", "'sylius_order_shipping'"]
 
-    sylius_inventory_unit:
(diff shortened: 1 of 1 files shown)


michalmarcinkowski:
Thanks Steffen! 👍

PR #6066, opened 2016-09-12
Let us to create a custom checkout flow

videni:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| License | MIT |

Because OrderCheckoutTransitions class is final , we have no way to add custom translation , and also  CheckoutResolver is bound to OrderCheckoutTransitions,  we can't create a custom checkout flow.  making CheckoutResolver to get translation graph from request , probably this is the simplest way to let us create a custom checkout flow。

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Checkout/CheckoutResolver.php
+++ b/src/Sylius/Bundle/CoreBundle/Checkout/CheckoutResolver.php
@@ -14,7 +14,6 @@ namespace Sylius\Bundle\CoreBundle\Checkout;
 use SM\Factory\FactoryInterface;
 use Sylius\Component\Cart\Context\CartContextInterface;
 use Sylius\Component\Core\Model\OrderInterface;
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/CoreBundle/Checkout/CheckoutResolver.php
```diff
@@ -71,6 +70,7 @@ public function __construct(
      */
     public function onKernelRequest(GetResponseEvent $event)
     {
+
```
pamil: ?

pjedrzejewski:
Sounds good to me, could you have a look at the failing build?

videni:
@pjedrzejewski , it seems that falling build is not relevant to this summit.

michalmarcinkowski:
@videni thanks for your effort! Merged in #6959.

PR #6077, opened 2016-09-13
[Core] Remove all *.class parameters

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | partially #5001
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/integration/hwi_oauth.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/integration/hwi_oauth.xml
@@ -12,12 +12,8 @@
 -->
 
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
-    <parameters>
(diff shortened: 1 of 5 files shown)


src/Sylius/Bundle/CoreBundle/Form/Type/ImageType.php
```diff
@@ -14,7 +14,7 @@
 use Sylius\Bundle\ResourceBundle\Form\Type\AbstractResourceType;
 use Symfony\Component\Form\FormBuilderInterface;
 
-class ImageType extends AbstractResourceType
+abstract class ImageType extends AbstractResourceType
```
michalmarcinkowski: Why making this abstract?

pamil: I've created new class `ProductVariantImageType` to avoid `TaxonImageType` extending `ImageType` (being the product variant one).

michalmarcinkowski: Can't we have 3 types? `sylius_image`, `sylius_product_variant_image` and `sylius_taxon_image`?

pamil: Removed that abstract, registering this one as `sylius_product_variant_image`. `sylius_image` has no meaning right now as we don't have a general model for images (just `ProductVariantImage`, `TaxonImage`, etc.).

michalmarcinkowski:
👍

PR #6093, opened 2016-09-13
[Ui] Remove .class parameters

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | partially #5001
| License         | MIT


src/Sylius/Bundle/UiBundle/Resources/config/services.xml
```diff
@@ -0,0 +1,21 @@
...
+ (c) Paweł Jędrzejewski
+
+ For the full copyright and license information, please view the LICENSE
+ file that was distributed with this source code.
+
+-->
+
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <imports>
+        <import resource="services/controller.xml" />
```
lchrusciel: Instead of importing different files, it would be enough to place all of services in this file.

pamil: I've just aimed to unify services definitions localisation. This may be a subject for an another discussion (the current organisation encourages to classify the service under one of four categories).

lchrusciel: It is up to you, if you move them. Everything else is fine for me.

michalmarcinkowski:
👍

PR #6100, opened 2016-09-13
[ShopBundle] Locale and currency switcher small improvement

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no|yes
| Related tickets | 
| License         | MIT

I noticed that when you directly switch the locale or currency you get a 500 error since there is no `referer`. These changes will provide a nice fallback for when that happens.


michalmarcinkowski:
Thanks Steffen!

PR #6101, opened 2016-09-13
Move checkout complete listener from ShopBundle to CoreBundle.

### LINKED ISSUE #6097: [ShopBundle] Move CheckoutCompleteListener to CoreBundle
The core bundle itself depends on the session `sylius_order_id` (for the thank you page for example), so it would make sense to move the listener responsible for setting that session to the core bundle, right?

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | no|yes
| New feature?    | no|yes
| BC breaks?      | no|yes
| Related tickets | fixes #6097 
| License         | MIT

This provides a better integration when using `CoreBundle` without `ShopBundle`.


pjedrzejewski:
I think this is specific to shop implementation, not the core itself, hm? :)

michalmarcinkowski:
Closing as this listener will be removed by the end of the week. @steffenbrem look forward to the PR :)

PR #6112, opened 2016-09-14
[OrderItemUnit][ShippingUnit][InventoryUnit] Clean up

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

- remove inventoryState from InventoryUnit
- remove shippingState from ShipmentUnit
- remove inventoryState and shippingState from OrderItemUnit
- remove OrderInventoryListener
- remove OrderItemInventoryListener
- remove InventoryHandler


steffenbrem:
@GSadee This PR may be somehow related to a fix I proposed in #6064.

src/Sylius/Component/Shipping/Model/ShipmentUnit.php
```diff
@@ -62,7 +57,7 @@ public function getId()
     }
 
     /**
-     * {@inheritdoc}
+     * {@inheritd
```
michalmarcinkowski: Whaat?!

michalmarcinkowski:
Good job Grzesiu!

tchapi:
Just a quick question on this PR. Before, in order to tell that an order was shipped programmatically (think of a mp3 download, for instance), I was doing that : 

    // ...
    foreach ($item->getUnits() as $unit) {
            $unit->setShippingState(ShipmentInterface::STATE_SHIPPED);
            $unit->getShipment()->setState(ShipmentInterface::STATE_SHIPPED);
     }
    // ...

Now, I just remove the `$unit->setShippingState()` call, is that right ? And the `setState()` of the **Shipment** will do the rest.

Thanks for your light on this !

steffenbrem:
@tchapi I think so yeah. From what I can tell the shipment now says if it's shipped or not. And I guess when you want to do partial shipping, you just remove some units of a shipment and put it on a new one and ship that.

tchapi:
hmmm ... that makes sense, right. A little bit of testing gave me the same result, so I guess I'll go this way. Thanks again **Steffen** !

PR #6136, opened 2016-09-16
Fix Shipping & Billing mapping to order.

ylastapis:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

When trying to use the same billing address or shipping address for an order, I got an Integrity constraint violation from the unique indexes mapped to billing & shipping address.

Was there a reason for having a one-to-one mapping on order ?

@pamil, @Zales0123  can you check I'm not doing something crazy ?

### DIFF
--- a/app/migrations/Version20160916113304.php
+++ b/app/migrations/Version20160916113304.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Sylius\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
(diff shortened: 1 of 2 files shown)


Zales0123:
In fact, each address attached to order should be a different entity. We should not use address that already exist, as if you change it in one order it would also change in others ;/ Better approach is to clone address entity while setting it to order, to prevent unneeded modifications.

ylastapis:
@Zales0123 Many thanks, it makes sense now :)

PR #6137, opened 2016-09-16
5374 find one by name consistency

### LINKED ISSUE #5374: findOneByName inconsistent behaviour
For all translatable entities `findOneByName` implementation from the repo makes use of `getOneOrNullResult`() method which will throw an exception when more than one record is returned.
On all the other entity (non-translatable) repositories `findOneByName` is implemented using `findOneBy([...])` which does not throw any exception at all.

`findOneByName` should have same behaviour across all the entities otherwise it will lead to errors and confusion.
`getOneOrNullResult` should be replaced by `getResult()` and just first record fetched for all translatable entities. This is the default behaviour of `findOneBy` method from `EntityRepository` and should be maintained.

Thanks!

bretto36:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #5374
| License         | MIT


steffenbrem:
@bretto36 I think you already noticed this but you should ignore your editor specific files. You can do this in your global gitignore file, you can find it at `~/.gitignore_global`.

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
```diff
@@ -116,6 +116,7 @@ public function findOneByName($name)
             ->leftJoin('o.translations', 'translation')
             ->where('translation.name = :name')
             ->setParameter('name', $name)
+            ->setMaxResults(1)
```
lchrusciel: Why have you add this, if we use `getOneOrNullResult` anyway?

.buildpath
```diff
@@ -0,0 +1,199 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
lchrusciel: To revert

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
bretto36: So the issue is, if an entity that's translatable has 2 items with the same name, getOneOrNullResult will actually return an exception. If you read the linked issue it makes more sense

bretto36:
Thanks @steffenbrem. I did not know that a global gitignore was able to be created! You've just changed my world. I'm forever having to remember to ignore it

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
bretto36: See the issue mentioned in pull request.

Zales0123:
@bretto36 it's good for me, although you could squash/fix some commits ;)

bretto36:
You can actually do the commit squashing on merge - https://github.com/blog/2141-squash-your-commits

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
bretto36: Its true that it will, but only when using the find**One**ByName method. It will still display all the items when doing a findBy as it's not trying to find one. All this PR is doing is making the findOneByName method return the correct values that the default findOneByName would return (eg no Exception if multiple entries exist).

michalmarcinkowski:
We should fix this issue by specifying a locale in which we search. Instead of `setMaxResults(1)` we should add `where('translation.locale = :locale')`.

lchrusciel:
@michalmarcinkowski AFAIR it would not resolve same problem... We don't have any constrains on a names, so for example two products can have same. So adding this where statement is a must-have, but it is not enough. Anyway, using method `findOneByName` in this context is a misconception for me, because many resource can  be matched.

bretto36:
For the Product example, the only place where this function is currently called is in the Behat/Context/Transform/ProductContext.php.

lchrusciel:
So it is a wrong usage. We should not define new functions just to use them in test environment.

bretto36:
What would be a better solution for the tests? Only single field indexed on the product translation table is slug. Do we use that?

michalmarcinkowski:
Thanks Brett for your effort. Fixed in #6172.

PR #6140, opened 2016-09-16
[BugFix][Core] Order::isInvoiceAvailable, remove `STATE_RETURNED` constant.

### LINKED ISSUE #6139: [Bug][Core] Order::isInvoiceAvailable uses undefined constant STATE_RETURNED
The `Order::isInvoiceAvailable` still uses the `ShipmentInteface::STATE_RETURNED` constant, which has been removed. See: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Model/Order.php#L494

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no|yes
| BC breaks?      | no|yes
| Related tickets | fixes #6139
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Core/Model/Order.php
+++ b/src/Sylius/Component/Core/Model/Order.php
@@ -491,7 +491,7 @@ class Order extends Cart implements OrderInterface
     public function isInvoiceAvailable()
     {
         if (false !== $lastShipment = $this->getLastShipment()) {
-            return in_array($lastShipment->getState(), [ShipmentInterface::STATE_RETURNED, ShipmentInterface::STATE_SHIPPED]);
+            return in_array($lastShipment->getState(), [ShipmentInterface::STATE_SHIPPED]);
         }
 
         return false;


src/Sylius/Component/Core/Model/Order.php
```diff
@@ -491,7 +491,7 @@ public function getLastShipment()
     public function isInvoiceAvailable()
     {
         if (false !== $lastShipment = $this->getLastShipment()) {
-            return in_array($lastShipment->getState(), [ShipmentInterface::STATE_RETURNED, ShipmentInterface::STATE_SHIPPED]);
+            return in_array($lastShipment->getState(), [ShipmentInterface::STATE_SHIPPED]);
         }
 
         return false;
```
michalmarcinkowski: Not sure if we should hardcode this logic on model, but if so I would make it a single line:
`return OrderShippingStates::STATE_SHIPPED === $this->shippingState`

tchapi:
I don't understand the point of this PR since there are no more invoices in Sylius, so this function can be removed altogether I guess... or am I missing something?

PR #6142, opened 2016-09-17
[WIP][Resource] Make api response codes configurable 

### LINKED ISSUE #5222: [RFC] A way to change response code and/or body of ResourceController::updateAction

koemeet:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | maybe |
| Related tickets | fixes #5222 |
| License | MIT |

> I made one big change and that is using `array_replace_recursive` instead of `array_merge` for the default parameters and the passed in api parameters. This was needed so the default configuration `response_code` would stay intact when overridden inside a single route.

You can basically change the response code of every single action inside the resource controller.
### Usage

You can change the default response codes globally by doing:

``` yml
sylius_resource:
    settings:
        response_code:
            create: 201
            update: 200
```

You can then override specific response codes again on a per-route basis:

``` yml
sylius_api_some_resource:
    #...
    defaults:
        _sylius:
            response_code:
                create: 204 
```

You can override response codes for the following actions:
- create
- update
- delete 
- applyStateMachineTransition
- enable
- disable
- moveUp
- moveDown

Please let me know what you guys think about such a solution! Would love to hear any ideas for improvements on this 😄


src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
```diff
@@ -92,6 +92,23 @@ public function isHtmlRequest()
...
+     * @return integer
+     */
+    public function getResponseCode()
+    {
+        $responseCode = $this->parameters->get('response_code');
+
+        if (is_array($responseCode)) {
+            $controller = $this->request->attributes->get('_controller');
+            if (preg_match('/^.+\:(.+)Action$/i', $controller, $matches)) {
+                return $responseCode[$matches['1']];
```
lchrusciel: Why did you use a number in `'`? Also, why it is always 1?

steffenbrem: It is faster to quote array indexes.

steffenbrem: I will improve this part and also want to cover different cases in phpspec. Not even sure if I want to even use a reg ex.

steffenbrem:
@lchrusciel Yeah indeed, it's still in a wip and wanted to show it early to showcase the configuration idea etc. code-wise this is not finished yet.

src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
lchrusciel: It was my second question, is it needed here. Regex are useful, but usually hard to maintain.

PR #6149, opened 2016-09-19
[Behat] Remove unused steps

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT


michalmarcinkowski:
Please also remove method `thisUserHasProductInTheCart` and `thisUserBought`. Modify scenario that uses it (step `customer "X" placed order ..."` should be used instead).

src/Sylius/Behat/Context/Ui/CheckoutContext.php
```diff
@@ -202,17 +192,6 @@ public function iTryToOpenCheckoutCompletePage()
     }
 
     /**
-     * @Given /^(this user) bought(?: this product| those products)$/
-     */
-    public function thisUserBought(ShopUserInterface $user)
-    {
-        $this->sharedSecurityService->performActionAsShopUser($user, function () {
```
michalmarcinkowski: I think we can also remove `performActionAsShopUser` method from `sharedSecurityService`, it's not used anywhere now.

PS. Sorry for being annoying 😄

michalmarcinkowski:
Thanks Arek!

PR #6154, opened 2016-09-19
[Core] Minor fixes in TaxonInterface model

pamil:


michalmarcinkowski:
Good catch! It was overlooked since at least 2014 😮

PR #6156, opened 2016-09-19
[Admin][Product] Move images from a product variant to a product

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

In this PR I also refactored a little bit `Image` and `TaxonImage` classes.

For future:
- remove relation from `Image` to its owner (`Taxon` or `Product`)
- combine `ProductImage` and `TaxonImage` into one object (`Image`)


src/Sylius/Component/Core/Model/ImageInterface.php
```diff
@@ -14,11 +14,22 @@
 use Sylius\Component\Resource\Model\ResourceInterface;
 use Sylius\Component\Resource\Model\TimestampableInterface;
 
-interface ImageInterface extends
-    ResourceInterface,
-    TimestampableInterface
+/**
+ * @author Grzegorz Sadowski <grzegorz.sadowski@lakion.com>
+ */
+interface ImageInterface extends ResourceInterface, TimestampableInterface
```
michalmarcinkowski: Do we need to implement `TimestampableInterface`? Don't think it's useful for image...

src/Sylius/Bundle/CoreBundle/DependencyInjection/Configuration.php
```diff
@@ -55,15 +54,15 @@ private function addResourcesSection(ArrayNodeDefinition $node)
...
                             ->children()
                                 ->variableNode('options')->end()
                                 ->arrayNode('classes')
                                     ->addDefaultsIfNotSet()
                                     ->children()
-                                        ->scalarNode('model')->defaultValue(ProductVariantImage::class)->cannotBeEmpty()->end()
-                                        ->scalarNode('interface')->defaultValue(ProductVariantImageInterface::class)->cannotBeEmpty()->end()
+                                        ->scalarNode('model')->defaultValue(ProductImage::class)->cannotBeEmpty()->end()
+                                        ->scalarNode('interface')->defaultValue(ImageInterface::class)->cannotBeEmpty()->end()
                                         ->scalarNode('controller')->defaultValue(ResourceController::class)->cannotBeEmpty()->end()
```
michalmarcinkowski: Do we need controllers for images?

src/Sylius/Bundle/CoreBundle/EventListener/ImageUploadListener.php
```diff
@@ -27,86 +27,35 @@ class ImageUploadListener
...
+    private function uploadImages(ImageAwareInterface $subject)
     {
-        $images = $productVariant->getImages();
+        $images = $subject->getImages();
         foreach ($images as $image) {
             if ($image->hasFile()) {
                 $this->uploader->upload($image);
             }
 
-            // Upload failed? Let's remove that image.
```
michalmarcinkowski: That can be useful comment :)

src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/Image.orm.xml
```diff
@@ -17,6 +16,8 @@
         <field name="updatedAt" column="updated_at" type="datetime" nullable="true">
             <gedmo:timestampable on="update"/>
         </field>
+
+        <field name="code" column="code" type="string" unique="false" />
```
michalmarcinkowski: Code should be just after `id`

src/Sylius/Component/Core/Model/Product.php
```diff
@@ -330,24 +336,58 @@ public function getPrice()
...
     {
-        if (null === $this->getFirstVariant()) {
-            return null;
+        return $this->images;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getImageByCode($code)
```
okwinza: Since these methods are common to any entity containing an image i think it makes sense to introduce `ImageOwnerInterface` for DX sake.

okwinza:
IMO moving image relations from variant to product seems fairly questionable as there are many cases in which you want to display an image for one particullar variant. 

I know it is kinda simpler to have images only rely on products but in the real world there should be some kind of connection between ProductVariant and the image. Or at least a way to easily implement it. The chosen approach in this PR only complicates things. 

On the other hand, the base Image model and stuff look pretty good now.

src/Sylius/Component/Core/Model/Product.php
GSadee: And we have already sth like that: `ImageAwareInterface`

src/Sylius/Behat/Page/Admin/Product/UpdateSimpleProductPage.php
```diff
@@ -140,4 +177,43 @@ private function openTaxonBookmarks()
...
+    private function getImageElementByCode($code)
+    {
+        $images = $this->getElement('images');
+        $inputCode = $images->find('css', 'input[value="'.$code.'"]');
+
+        if (null === $inputCode) {
+            return null;
+        }
+
+        return $inputCode->getParent()->getParent()->getParent();
```
tuka217: Maybe find a parent element which contain the input with passed code?

michalmarcinkowski:
@okwinza the reason for that change is that current implementation is bad. Our current logic is based on using images from "default" variant, so there is no "image per variant" logic at all.

I know the per variant image is a basic eCommerce feature, but as a first step to fixing current logic we decided to simplify it. The default shop interface will now work correctly and if we find some time before beta, we will probably add proper logic for displaying images per variant to Core (but can't guarantee that). If you need per variant images you can still easily add it as customization. The only thing that is removed is relation to image, you can bring it back in your custom model and that's it :)

src/Sylius/Behat/Page/Admin/Product/UpdateSimpleProductPage.php
GSadee: Good idea, but.. I have tried without any result, could you tell mi how to do that correctly in this case?

okwinza:
@michalmarcinkowski Happy to hear you're still planning on bring it back, this is really basic and important feature. Could you please also share your thoughs about "proper logic" to have images tied to variants?

And yes, it is still possible to override and implement whatever you need but now it requires fixing the entire stack from models to forms only to bring back what was already there IIUC.

src/Sylius/Behat/Page/Admin/Product/CreateConfigurableProductPage.php
```diff
@@ -42,11 +43,51 @@ public function selectOption($optionName)
...
+
+    /**
+     * @return NodeElement
+     */
+    private function getLastImageElement()
+    {
+        $images = $this->getElement('images');
+        $items = $images->findAll('css', 'div[data-form-collection="item"]');
+
+        return end($items);
```
michalmarcinkowski: `end($items)` can return `false` we should assert it didn't.

src/Sylius/Behat/Page/Admin/Product/CreateSimpleProductPage.php
```diff
@@ -116,11 +134,25 @@ private function waitForFormElement($timeout = 5)
...
+
+    /**
+     * @return NodeElement
+     */
+    private function getLastImageElement()
+    {
+        $images = $this->getElement('images');
+        $items = $images->findAll('css', 'div[data-form-collection="item"]');
+
+        return end($items);
```
michalmarcinkowski: Same as above and should be applied to every other similar case.

src/Sylius/Bundle/AdminBundle/Resources/views/Product/_mainImage.html.twig
```diff
@@ -1,3 +1,3 @@
-{% set path = product.image is not null ? product.image.path|imagine_filter(filter|default('sylius_admin_product_thumbnail')) : 'http://placehold.it/50x50' %}
+{% set path = product.imageByCode('thumbnail') is not null ? product.imageByCode('thumbnail').path|imagine_filter(filter|default('sylius_admin_product_thumbnail')) : 'http://placehold.it/50x50' %}
```
michalmarcinkowski: We should fallback to first image if there is no image with `thumbnail` code. Same for `main`

michalmarcinkowski:
Nice work Grzesiu! 👍

PR #6161, opened 2016-09-20
[Shop] Different variant prices on product details page

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Related tickets |
| License         | MIT

On product details page, we have always one price displayed (from the first/default/etc variant). It's obviously confusing, as we could have different price for each variant - therefore it should change after changing variant/option. In this PR, I will implement some lovely JS script, to handle this logic and make *Sylius* product details page more and more perfect, to reach the sky and more with its perfectness 🌠 🎉 

zrzut ekranu 2016-09-20 o 15 22 58
zrzut ekranu 2016-09-20 o 15 23 03
zrzut ekranu 2016-09-20 o 15 23 26
zrzut ekranu 2016-09-20 o 15 24 09

PR #6165, opened 2016-09-20
[RFC][Resource] Make Resource actions easier to overwrite

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets |
| License         | MIT

As we know, ``ResourceController`` gives us a lot, starting from duplication reduction, on sense of security ending 😄  However, it's pretty hard to use its methods with some extra data or logic - it usually ends up with copying whole action. It's ridiculous, especially when we just want to add some extra data to template (which was my use case in https://github.com/Sylius/Sylius/pull/6161 - I ended up with overwriting the whole damn action).
I suggest to introduce some protected methods returning templates data, that can be overwrote instead of actions. I know it's not the most clean way (we still have some parameters passed to these functions which looks awful for method, that should just return required data), but could be a first step for making ``ResourceController`` easier to use.
I would appreciate any feedback about this, is it the path we want to follow or not. 🐫

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
@@ -182,7 +182,7 @@ class ResourceController extends Controller
             $view
                 ->setTemplate($configuration->getTemplate(ResourceActions::SHOW . '.html'))
                 ->setTemplateVar($this->metadata->getName())
-                ->setData($this->getShowTemplateData($configuration, $resource))
+                ->setData($this->provideShowTemplateData($configuration, $resource))
             ;
         }
 
@@ -207,7 +207,7 @@ class ResourceController extends Controller
             $view
                 ->setTemplate($configuration->getTemplate(ResourceActions::INDEX . '.html'))
                 ->setTemplateVar($this->metadata->getPluralName())
-                ->setData($this->getIndexTemplateData($configuration, $resources))
+                ->setData($this->provideIndexTemplateData($configuration, $resources))
             ;
         }
 
@@ -259,7 +259,7 @@ class ResourceController extends Controller
         }
 
         $view = View::create()
-            ->setData($this->getCreateTemplateData($configuration, $newResource, $form))
+            ->setData($this->provideFormTemplateData($configuration, $newResource, $form))
             ->setTemplate($configuration->getTemplate(ResourceActions::CREATE . '.html'))
         ;
 
@@ -315,7 +315,7 @@ class ResourceController extends Controller
         }
 
         $view = View::create()
-            ->setData($this->getUpdateTemplateData($configuration, $resource, $form))
+            ->setData($this->provideFormTemplateData($configuration, $resource, $form))
             ->setTemplate($configuration->getTemplate(ResourceActions::UPDATE . '.html'))
         ;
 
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
```diff
@@ -550,4 +531,87 @@ protected function findOr404(RequestConfiguration $configuration)
...
+    }
+
+    /**
+     * @param RequestConfiguration $configuration
+     * @param ResourceInterface $resource
+     * @param FormInterface $form
+     *
+     * @return array
+     */
+    protected function getUpdateTemplateData(
```
lchrusciel: We should leave it in one line.

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
```diff
@@ -550,4 +531,87 @@ protected function findOr404(RequestConfiguration $configuration)
...
+    }
+
+    /**
+     * @param RequestConfiguration $configuration
+     * @param ResourceInterface $newResource
+     * @param FormInterface $form
+     *
+     * @return array
+     */
+    protected function getCreateTemplateData(
```
lchrusciel: Is it really needed to different create and update template data provider?

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
```diff
@@ -550,4 +531,87 @@ protected function findOr404(RequestConfiguration $configuration)
...
+            'configuration' => $configuration,
+            'metadata' => $this->metadata,
+            'resource' => $resource,
+            $this->metadata->getName() => $resource,
+        ];
+    }
+
+    /**
+     * @param RequestConfiguration $configuration
+     * @param mixed $resources
```
lchrusciel: Isn't it `ResourceInterface`?

Zales0123: No :( As I noticed, there could be passed array or ``ResourceGridView`` and God knows what else... 😢

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
Arminek: This method name is so confusing.

Zales0123: Tell me more how it should be named, I have no idea 😄

lchrusciel: IMHO we should leave just `provideFormTemplateData`.

Zales0123: Ok, so there are two approaches:
1. ``provideFormTemplateData`` for create and update action
2. ``provideCreate\UpdateFormTemplateData`` which uses ``provideFormTemplateData`` for templates data separation

👍  for 1, 🎉  for 2, ready, set, go! 😄

pjedrzejewski:
From my experience, 99% cases you can obtain the extra data in a much cleaner way. (partial that is cachable separately / twig)

And if you really need custom data from the controller, then resource is not the way to go and much cleaner is to simply add a super simple custom controller. (most of the time not needed though)

So I am against this change, unless we see a really annoying use-case for that. :D I commented in #6161, which is (I assume) the cause for this PR.

PR #6172, opened 2016-09-21
[Behat][Repositories] Refactory find by name repository methods

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | https://github.com/Sylius/Sylius/pull/6137, https://github.com/Sylius/Sylius/issues/5374
| License         | MIT

As entities names are not unique, whole concept of ``findOneByName`` repository methods is incorrect and leads to some potential errors. I've changed each ``findOneByName`` to ``findByName``, therefore giving possibility to return zero or multiple entities.
Moreover, currently these methods are used only in Behat contexts. Instead of asserting ``notNull`` in them (basically in transformer's methods), now we assets ``eq(1,...)``, to be sure we operate on **one** and only **one** object.

### DIFF
--- a/src/Sylius/Behat/Context/Setup/TaxationContext.php
+++ b/src/Sylius/Behat/Context/Setup/TaxationContext.php
@@ -176,7 +176,7 @@ final class TaxationContext implements Context
     {
         $taxCategories = $this->taxCategoryRepository->findByName($taxCategoryName);
         if (empty($taxCategories)) {
-            $taxCategories = [$this->createTaxCategory($taxCategoryName)];
+            return $this->createTaxCategory($taxCategoryName);
         }
 
         Assert::eq(
--- a/src/Sylius/Behat/Context/Setup/TaxationContext.php
+++ b/src/Sylius/Behat/Context/Setup/TaxationContext.php
@@ -22,6 +22,7 @@ use Sylius\Component\Resource\Factory\FactoryInterface;
 use Sylius\Component\Resource\Repository\RepositoryInterface;
 use Sylius\Component\Taxation\Model\TaxCategoryInterface;
 use Sylius\Component\Taxation\Repository\TaxCategoryRepositoryInterface;
+use Webmozart\Assert\Assert;
 
 /**
  * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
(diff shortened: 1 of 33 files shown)


src/Sylius/Bundle/PaymentBundle/Doctrine/ORM/PaymentMethodRepository.php
```diff
@@ -28,25 +28,10 @@ public function findByName(array $names)
...
-        return $this->createQueryBuilder('o')
-            ->addSelect('translation')
-            ->leftJoin('o.translations', 'translation')
-            ->where('translation.name = :name')
             ->setParameter('name', $name)
             ->getQuery()
-            ->getOneOrNullResult()
-        ;
+            ->getResult()
+            ;
```
lchrusciel: Should be reverted

src/Sylius/Behat/Context/Setup/TaxationContext.php
```diff
@@ -173,12 +174,18 @@ public function theTaxRateIsOfAmount(TaxRateInterface $taxRate, $amount)
      */
     private function getOrCreateTaxCategory($taxCategoryName)
     {
-        $taxCategory = $this->taxCategoryRepository->findOneByName($taxCategoryName);
-        if (null === $taxCategory) {
-            $taxCategory = $this->createTaxCategory($taxCategoryName);
+        $taxCategories = $this->taxCategoryRepository->findByName($taxCategoryName);
+        if (empty($taxCategories)) {
```
lchrusciel: I don't like this logic, but it can be case for another PR

src/Sylius/Behat/Context/Setup/TaxationContext.php
```diff
@@ -173,12 +174,18 @@ public function theTaxRateIsOfAmount(TaxRateInterface $taxRate, $amount)
...
+            $taxCategories = [$this->createTaxCategory($taxCategoryName)];
```
michalmarcinkowski: `return $this->createTaxCategory($taxCategoryName)` 😉

src/Sylius/Component/Payment/Repository/PaymentMethodRepositoryInterface.php
```diff
@@ -21,17 +21,10 @@
 interface PaymentMethodRepositoryInterface extends RepositoryInterface
 {
     /**
-     * @param array $names
+     * @param string $names
```
michalmarcinkowski: `name`

michalmarcinkowski:
You should also apply [my comment](https://github.com/Sylius/Sylius/pull/6137#issuecomment-248406620), for translatable resources we should specify a locale in which we would like to search.

Zales0123:
@michalmarcinkowski Done 👍

michalmarcinkowski:
Thanks Mateusz! 👍

src/Sylius/Behat/Context/Transform/ProductContext.php
```diff
@@ -41,14 +41,15 @@ public function __construct(ProductRepositoryInterface $productRepository)
      */
     public function getProductByName($productName)
     {
-        $product = $this->productRepository->findOneByName($productName);
+        $products = $this->productRepository->findByName($productName, 'en_US');
```
Zales0123: For now we use English language in Behat I think, but it obviously should be changed (getting locale from shared storage?)

lchrusciel: Passing locale as a second argument with default value?

pamil: In fact, we should use `%locale%` parameter, but I've got no idea how to pass it here right now.

michalmarcinkowski: @pamil that's the reason for having it hardcoded for now 😉

PR #6176, opened 2016-09-21
[Shop] Preselect only available country on addressing form

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets |
| License         | MIT

### DIFF
--- a/features/checkout/addressing_order/having_only_possible_country_preselected.feature
+++ b/features/checkout/addressing_order/having_only_possible_country_preselected.feature
@@ -1,0 +1,24 @@
+@checkout
+Feature: Having only possible country preselected
+    In order to not be forced to select country when only one country is available
+    As a Customer
+    I want to have only available country preselected
+
+    Background:
+        Given the store operates on a single channel in "United States"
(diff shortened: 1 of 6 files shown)


michalmarcinkowski:
👍

pjedrzejewski:
This is a required field, so we should not allow "empty" option, this looks like a bit of workaround for a simple thing.

michalmarcinkowski:
Do you know a better solution than one `if` statement?

pjedrzejewski:
Yes, using the built-in HTML feature... 😆  https://jsfiddle.net/u7La63ph/

The rendered select is invalid, because it allows empty value, which is incorrect, as this field must be selected. It will work even if there are no countries configured. So the solution is to fix Symfony form to render properly, not this if and "forcing" a value on the field.

michalmarcinkowski:
Welcome back our frontend master! 🎉  😆

pjedrzejewski:
XD

PR #6182, opened 2016-09-22
make PaymentMethodChoiceType to use code as choice option value

videni:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |

We are not able to guess the id of a payment method, so I suggest to use code as choice option, we can replace PaymentMethodChoiceType with ours , but it is not a good practice to create a new class whose function is same as the original one.


pamil:
Seems to be already done in some other PR. Anyway, thanks for the PR @videni! :)

PR #6184, opened 2016-09-22
[Behat] Fix geographical step

tuka217:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Related tickets | - |
| License | MIT |

### DIFF
--- a/features/checkout/addressing_order/choosing_province_for_country.feature
+++ b/features/checkout/addressing_order/choosing_province_for_country.feature
@@ -6,7 +6,7 @@ Feature: Choosing province for country
 
     Background:
         Given the store operates on a single channel in "United States"
-        And country "United States" has the "New York" province with "NY" code
+        And the country "United States" has the "New York" province with "NY" code
         And the store ships to "Nepal"
         And the store has a product "The Dark Knight T-Shirt" priced at "$49.99"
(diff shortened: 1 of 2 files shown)


features/checkout/addressing_order/choosing_province_for_country.feature
```diff
@@ -6,7 +6,7 @@ Feature: Choosing province for country
 
     Background:
         Given the store operates on a single channel in "United States"
-        And country "United States" has the "New York" province with "NY" code
+        And the country "United States" has the "New York" province with "NY" code
```
lchrusciel: 1) And the "United States" country has the "New York" province with "NY" code
2) It is not totally true, I guess. As far as I remember, countries don't have `the` before theirs name. The exception is when the name of a country suggest larger number of smaller pieces that together are country. /cc @TheMadeleine

TheMadeleine: @lchrusciel That is why our fixtures are probably wrong -> "The United States". But here the step is okay.
Since we are describing entities.

pjedrzejewski:
@tuka217 Can you rebase, please? Thank you! :)

PR #6186, opened 2016-09-22
[RFC] [Shop] [Twig] All metatags in matatags block

lchrusciel:
| Q | A |
| --- | --- |
| Bug fix? | maybe |
| New feature? | no |
| BC breaks? | no |
| Related tickets |  |
| License | MIT |

Just an idea. Is there any reason why some of metatags shouldn't be place in matatags block? If no, I think it would be better to keep them there.

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/layout.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/layout.html.twig
@@ -2,14 +2,13 @@
 
 <html lang="{{ app.request.locale }}">
 <head>
-    <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


pamil:
Is there some reason that we should keep them in there? :)

lchrusciel:
Just an idea. If we have a metatags block, it would be nice to keep meta tags inside.

inssein:
Not a big fan of this PR personally. There are some <meta> tags that are for the page, and some for search engines.

I usually keep them separate, it allows me to override SEO related meta tags without having to re-declare things like charset.

PR #6193, opened 2016-09-23
[Product] Refactor product variant generator

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | -
| License         | MIT

- [x] Removed `$isRecursiveStep` argument from public method `build()`
- [x] Removed `SetBuilderInterface` as it is really hard to enforce a contract there (what it takes and what it returns)
- [x] Put `CartesianSetBuilder` directly into the default `ProductVariantGenerator`


src/Sylius/Component/Product/Generator/CartesianSetBuilder.php
```diff
@@ -9,22 +9,35 @@
...
+        return $this->doBuild($setTuples, false);
+    }
+
+    /**
+     * @param array $setTuples
+     * @param bool $isRecursiveStep
+     *
+     * @return array
+     */
+    private function doBuild(array $setTuples, $isRecursiveStep)
```
lchrusciel: Instead of `doBuild` we should name it `buildRecursive`. It will be more meaningful.

Zales0123: DO
DO BUILD
DO BUILD MICH
do-build

src/Sylius/Component/Product/Generator/CartesianSetBuilder.php
```diff
@@ -9,22 +9,35 @@
...
-namespace Sylius\Component\Product\SetBuilder;
+namespace Sylius\Component\Product\Generator;
 
 /**
  * Builds the Cartesian product set from one or more given sets.
  *
  * @author Paweł Jędrzejewski <pawel@sylius.org>
  */
-final class CartesianSetBuilder implements SetBuilderInterface
+final class CartesianSetBuilder
```
lchrusciel: Why did you remove interface? I guess cartesian set can be generated different way, based on a chosen algorithm

lchrusciel: Ok, I get it. Im not sure about that, but ok, description of PR makes sense.

src/Sylius/Component/Product/Generator/CartesianSetBuilder.php
pamil: The first call made in `build()` method is not recursive :)

pjedrzejewski:
Thank you Kamil!

PR #6208, opened 2016-09-26
[Behat] Add assert to getting last element of an array

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Transform/OrderContext.php
+++ b/src/Sylius/Behat/Context/Transform/OrderContext.php
@@ -12,6 +12,7 @@
 namespace Sylius\Behat\Context\Transform;
 
 use Behat\Behat\Context\Context;
+use Sylius\Component\Core\Model\OrderInterface;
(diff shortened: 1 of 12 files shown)


src/Sylius/Behat/Context/Transform/OrderContext.php
```diff
@@ -69,7 +70,10 @@ public function getOrderByCustomer($email)
         $orders = $this->orderRepository->findByCustomer($customer);
         Assert::notEmpty($orders);
 
-        return end($orders);
+        $lastOrder = end($orders);
```
pamil: `Assert::notEmpty($orders)` is enough here.

GSadee:
I agree that `Assert::notEmpty` could be enough.

pjedrzejewski:
Thanks Grzesiu!

PR #6216, opened 2016-09-26
[Behat] Divide behat fixtures into directories

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT


michalmarcinkowski:
Do we really need separate directories? I think it doesn't matter what image we will upload in test environment and we don't need to categorize them 😉

GSadee:
I wanted it to be more transparent, but of course I agree, that it doesn't matter what images and from we are uploading.

PR #6223, opened 2016-09-27
Added raw keyword on User Confirmation Email

bretto36:
| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Related tickets |  |
| License | MIT |

Found the user confirmation email was outputting <br/> in the email content and raw wasn't used

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/views/Email/userConfirmation.html.twig
+++ b/src/Sylius/Bundle/CoreBundle/Resources/views/Email/userConfirmation.html.twig
@@ -3,5 +3,5 @@
 {% endblock %}
 
 {% block body %}
-    {{ 'sylius.email.user_confirmation.content'|trans({'%username%': user.username}) }}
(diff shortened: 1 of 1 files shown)


michalmarcinkowski:
I think this probably should be fixed in all email templates, have you checked them?

/cc @pjedrzejewski

bretto36:
Yes all the others either have raw or use a block tag to escape them

On 27 Sep 2016, at 7:20 PM, Micha? Marcinkowski <notifications@github.com<mailto:notifications@github.com>> wrote:


I think this probably should be fixed in all email templates, have you checked them?

/cc @pjedrzejewski<https://github.com/pjedrzejewski>

-

bretto36:
see  Sylius/src/Sylius/Bundle/CoreBundle/Resources/views/Email/orderConfirmation.html.twig for blocktag

michalmarcinkowski:
After #6880 you should override the template and put whatever you need there.

PR #6235, opened 2016-09-28
[Shop][UI] Fix styles path for the shop

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Gulpfile.js
+++ b/src/Sylius/Bundle/ShopBundle/Gulpfile.js
@@ -23,7 +23,7 @@ var paths = {
         ],
         sass: [
             '../UiBundle/Resources/private/sass/**',
-            'ShopBundle/Resources/private/scss/**'
+            'Resources/private/scss/**'
         ],
         css: [
             '../../../../node_modules/semantic-ui-css/semantic.min.css'


src/Sylius/Bundle/ShopBundle/Gulpfile.js
```diff
@@ -23,7 +23,7 @@ var paths = {
         ],
         sass: [
             '../UiBundle/Resources/private/sass/**',
```
pamil: Will it work? With Sylius installed as single components (not `sylius/sylius` replacing all the other packages), the UiBundle will be at `vendor/sylius/ui-bundle/` and ShopBundle at `vendor/sylius/shop-bundle/`.

src/Sylius/Bundle/ShopBundle/Gulpfile.js
```diff
@@ -23,7 +23,7 @@ var paths = {
...
-            'ShopBundle/Resources/private/scss/**'
+            'Resources/private/scss/**'
```
pamil: Side note: we have `sass` folder in UiBundle and `scss` here in ShopBundle.

pjedrzejewski: Yeah, I will clean this up when finalizing the UI.

src/Sylius/Bundle/ShopBundle/Gulpfile.js
pjedrzejewski: If you use Shop/Admin/Api, which all depend on Core, I think we won't recommend using them as standalone components and maybe we even remove their composer.json and recommend depending on ``sylius/sylius``, which solves this and many other problems. Thoughts?

michalmarcinkowski:
👍

PR #6237, opened 2016-09-28
Cleanup some state machine callbacks.

ylastapis:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

- sylius_create_payment (sylius_order_checkout state machine) is already applyed on sylius_order state machine
- move sylius_request_payment from sylius_order to sylius_order_checkout, to keep consistency with sylius_request_shipping


michalmarcinkowski:
Thank you Yann! 👍

pjedrzejewski:
Actually, it should be the other way around. :) @ylastapis Yann, could you move both callbacks to sylius_order state machine? This is "order" logic, not really related to the checkout, so it should happen when "cart" order becomes "new" order. Sounds good? :) Thanks!

ylastapis:
Sounds good to me ;)

Created a new PR here https://github.com/Sylius/Sylius/pull/6246

Issue #6258, opened 2016-10-01
Payment Bugs


inssein:
I just integrated payments into my Sylius (~1.0) application today, and there are several bugs that are fairly critical. Some of them have open tickets, some do not. These are all stripe related.

1) The credit card page is fairly broken, and can be fixed (https://github.com/Sylius/Sylius/issues/6239).
2) When using stripe there is no action that supports `ResolveNextRoute`. I copied the one from paypal and re-used it.
3) When using multiple currencies with a switcher, the amount applied to a payment is not correct. It set's the right currency, but it sets the wrong amount (it is no converted).
4) If you override the CurrencyConverter (for example, lets say you want to apply rounding to every currency that is not base), the number in the finalize page and after are not consistent. I fixed this by overriding PriceHelper and doing my logic there as well (in the case code and exchange rate are provided)

As for the third issue, I don't currently have a workaround. I see that the OrderPaymentProcessor is responsible for setting the code and the amount. Setting the amount to `$order->getTotal() * $order->getExchangeRate()` seems wrong.

inssein:
#1 is not resolved, but has a duplicate (https://github.com/Sylius/Sylius/issues/6239)
#2 is being addressed by https://github.com/Sylius/Sylius/pull/6361
#3 is already resolved
#4 is not an issue / is an issue depending on how we resolve https://github.com/Sylius/Sylius/issues/6319

PR #6259, opened 2016-10-01
[Payum] Use converted amount for payment

inssein:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | partially https://github.com/Sylius/Sylius/issues/6258
| License         | MIT


pjedrzejewski:
This is incorrect, it is the responsibility of the gateway to convert the amount, have a look at our paypal implementation.

inssein:
Ah I think I missed that, though I did see the PayPal example.

Is there a plan to add a Generic converter?

PR #6265, opened 2016-10-03
[Checkout] Preventing to pay for cancelled order

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | -
| License         | MIT

### DIFF
--- a/features/checkout/paying_for_order/preventing_to_pay_for_cancelled_order.feature
+++ b/features/checkout/paying_for_order/preventing_to_pay_for_cancelled_order.feature
@@ -13,7 +13,7 @@ Feature: Preventing to pay for the cancelled order
         And the store allows paying offline
         And there is a customer "sylius@example.com" that placed an order "#00000022"
 
-    @todo
+    @ui
     Scenario: Not being able to pay for cancelled order
         Given the customer bought 3 "Iron Maiden T-Shirt" products
         And the customer chose "Free" shipping method to "United States" with "offline" payment
--- a/features/checkout/paying_for_order/preventing_to_pay_for_cancelled_order.feature
+++ b/features/checkout/paying_for_order/preventing_to_pay_for_cancelled_order.feature
@@ -1,0 +1,22 @@
+@paying_for_order
+Feature: Preventing to pay for the cancelled order
+    In order to not be able to buy order that is already cancelled
+    As a Customer
+    I want to be prevented from paying for the cancelled order
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "Iron Maiden T-Shirt"
+        And "Iron Maiden T-Shirt" product is tracked by the inventory
+        And there are 5 units of product "Iron Maiden T-Shirt" available in the inventory
+        And the store ships everywhere for free
+        And the store allows paying offline
+        And there is a customer "sylius@example.com" that placed an order "#00000022"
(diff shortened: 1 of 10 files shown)


src/Sylius/Bundle/ShopBundle/Resources/views/Checkout/orderDetails.html.twig
```diff
@@ -4,7 +4,7 @@
     {% set lastNewPayment = order.getLastNewPayment() %}
     {% include '@SyliusShop/Common/Order/_summary.html.twig' %}
 
-    {% if order.paymentState is not same as ('paid') %}
+    {% if order.paymentState is not same as ('paid') and order.paymentState is not same as ('cancelled') %}
```
pamil: `{% if order.paymentState not in ['paid', 'cancelled'] %}`

pamil: Side note: we should have issers for paid / cancelled / etc. states. The current implementation depends heavily on the state machine, but not on the model. Then we can do `{% if not order.paid and not order.cancelled %}`.

src/Sylius/Behat/Resources/config/suites/ui/checkout/paying_for_order.yml
```diff
@@ -8,19 +8,23 @@ default:
...
                 - sylius.behat.context.setup.channel
                 - sylius.behat.context.setup.geographical
                 - sylius.behat.context.setup.currency
+                - sylius.behat.context.setup.order
                 - sylius.behat.context.setup.payment
                 - sylius.behat.context.setup.shop_security
                 - sylius.behat.context.setup.user
                 - sylius.behat.context.setup.product
                 - sylius.behat.context.setup.shipping
+                - sylius.behat.context.transform.shipping_method
```
Zales0123: Why it's not placed together with other transform contexts? :(

features/checkout/paying_for_order/preventing_to_pay_for_cancelled_order.feature
```diff
@@ -0,0 +1,22 @@
...
+        And the store allows paying offline
+        And there is a customer "sylius@example.com" that placed an order "#00000022"
+
+    @ui
+    Scenario: Not being able to pay for cancelled order
+        Given the customer bought 3 "Iron Maiden T-Shirt" products
+        And the customer chose "Free" shipping method to "United States" with "offline" payment
+        And this order was cancelled
+        When I want to browse order details for this order
+        Then I should not be able to pay
```
michalmarcinkowski: The customer is not able to pay for the order when `offline` gateway was chosen regardless of the order state. In order to validate this scenario we should choose other than `offline` gateway.

Arminek: If i choose offline gateway i am still able to change the payment method and try to pay again.

michalmarcinkowski: You should be able to change a payment method, but the `pay` button should be hidden for `offline` gateway. E.g. I select "cash on delivery" payment method and then I press `pay` button? Looks strange, what happens then?

PR #6280, opened 2016-10-04
[Shipping][ShippingBundle] Remove shipping category

tuka217:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | -
| License         | MIT


tchapi:
Hi
I don't understand this PR. If we remove the shipping categories, how do we apply different shipping methods to different products ? If I have "downloadables" (mp3 songs for instance) and physical products (that need shipping), how will the shipping be calculated ?

PR #6288, opened 2016-10-04
[Resource] Remove our ExpressionLanguage

pamil:
There is already a Symfony one that does the same.


isometriks:
It would be useful to have the expression language be it's own service so you could add function providers to it:

https://github.com/symedit/SymEdit/blob/master/src/SymEdit/Bundle/CoreBundle/ExpressionLanguage/Provider/SettingsExpressionLanguageProvider.php

That would let you use expr:setting('namespace.param') in routing so only the routing is tied to settings bundle instead of using it directly in the controller. Using a compiler pass to add the providers via `registerProvider`

pamil:
Good catch @isometriks, will revert it! :)

pjedrzejewski:
Thank you Kamil!

PR #6294, opened 2016-10-05
[Resource] Remove ResourceSelectionToIdentifierCollectionTransformer

pamil:
Seems like it is not used anywhere in our codebase.


pjedrzejewski:
👍 👍 👍

PR #6311, opened 2016-10-05
[Images][Taxon] Validation for unique code during creation/edition

tuka217:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Related tickets | -
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Validator/Constraints/ImageUniqueCodeValidator.php
+++ b/src/Sylius/Bundle/CoreBundle/Validator/Constraints/ImageUniqueCodeValidator.php
@@ -26,17 +26,19 @@ class ImageUniqueCodeValidator extends ConstraintValidator
      */
     public function validate($images, Constraint $constraint)
     {
+        $imagesCodes = [];
+
         /** @var ImageInterface[] $images */
         foreach ($images as $key => $image) {
-            $filteredImages = $images->filter(function(ImageInterface $imageFromObject) use ($image) {
-                return $imageFromObject->getOwner() === $image->getOwner()
-                        && $imageFromObject->getCode() === $image->getCode()
-                        && $imageFromObject !== $image
-                ;
-            });
+            if (!array_key_exists($image->getCode(), $imagesCodes)) {
+                $imagesCodes[$image->getCode()] = $key;
+                continue;
+            }
 
-            if(0 !== count($filteredImages)) {
-                $this->context->addViolationAt(sprintf('[%d].code', $key), $constraint->message);
+            $this->context->addViolationAt(sprintf('[%d].code', $key), $constraint->message);
+            if (false !== $imagesCodes[$image->getCode()]) {
+                $this->context->addViolationAt(sprintf('[%d].code', $imagesCodes[$image->getCode()]), $constraint->message);
+                $imagesCodes[$image->getCode()] = false;
             }
         }
     }
--- a/src/Sylius/Bundle/CoreBundle/Validator/Constraints/ImageUniqueCodeValidator.php
+++ b/src/Sylius/Bundle/CoreBundle/Validator/Constraints/ImageUniqueCodeValidator.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\CoreBundle\Validator\Constraints;
+
+use Sylius\Component\Core\Model\ImageInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+/**
+ * @author Anna Walasek <anna.walasek@lakion.com>
+ */
+class ImageUniqueCodeValidator extends ConstraintValidator
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function validate($images, Constraint $constraint)
+    {
+        /** @var ImageInterface[] $images */
+        foreach ($images as $key => $image) {
+            $filteredImages = $images->filter(function(ImageInterface $imageFromObject) use ($image) {
+                return $imageFromObject->getOwner() === $image->getOwner()
+                        && $imageFromObject->getCode() === $image->getCode()
+                        && $imageFromObject !== $image
+                ;
+            });
+
+            if(0 !== count($filteredImages)) {
+                $this->context->addViolationAt(sprintf('[%d].code', $key), $constraint->message);
+            }
+        }
+    }
+}
--- a/src/Sylius/Bundle/CoreBundle/spec/Validator/Constraints/ImageUniqueCodeValidatorSpec.php
+++ b/src/Sylius/Bundle/CoreBundle/spec/Validator/Constraints/ImageUniqueCodeValidatorSpec.php
@@ -40,35 +40,33 @@ final class ImageUniqueCodeValidatorSpec extends ObjectBehavior
     
     function it_adds_violation_if_there_two_images_with_the_same_owner_which_have_same_codes(
         ImageInterface $firstImage,
-        ImageInterface $secondImage,
-        ImageAwareInterface $owner
+        ImageInterface $secondImage
     ) {
-        $firstImage->getOwner()->willReturn($owner);
-        $secondImage->getOwner()->willReturn($owner);
-
         $firstImage->getCode()->willReturn('car');
         $secondImage->getCode()->willReturn('car');
 
-        $this->bulidViolation('[0].code', Argument::type('string'))->shouldBeCalled();
-        $this->bulidViolation('[1].code', Argument::type('string'))->shouldBeCalled();
+        $this->addViolationAt('[0].code', Argument::type('string'))->shouldBeCalled();
+        $this->addViolationAt('[1].code', Argument::type('string'))->shouldBeCalled();
 
-        $this->validate(new ArrayCollection($firstImage->getWrappedObject(), $secondImage->getWrappedObject()), new ImageUniqueCode());
+        $this->validate(
+            new ArrayCollection($firstImage->getWrappedObject(), $secondImage->getWrappedObject()),
+            new ImageUniqueCode()
+        );
     }
 
     function it_does_not_add_violation_if_there_is_no_duplication_of_a_code(
         ImageInterface $firstImage,
-        ImageInterface $secondImage,
-        ImageAwareInterface $owner
+        ImageInterface $secondImage
     ) {
-        $firstImage->getOwner()->willReturn($owner);
-        $secondImage->getOwner()->willReturn($owner);
-
         $firstImage->getCode()->willReturn('car');
         $secondImage->getCode()->willReturn('wipers');
 
-        $this->bulidViolation('[0].code', Argument::type('string'))->shouldNotBeCalled();
-        $this->bulidViolation('[1].code', Argument::type('string'))->shouldNotBeCalled();
+        $this->addViolationAt('[0].code', Argument::type('string'))->shouldNotBeCalled();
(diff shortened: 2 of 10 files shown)


src/Sylius/Bundle/CoreBundle/Validator/Constraints/ImageUniqueCodeValidator.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->resourceRepository = $resourceRepository;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function validate($images, Constraint $constraint)
+    {
+        /** @var ImageInterface[] $imagesFromDatabase */
+       $imagesFromDatabase =  $this->resourceRepository->findAll();
```
pamil: Getting all of them will be sloooooooooow.

src/Sylius/Bundle/CoreBundle/Validator/Constraints/ImageUniqueCodeValidator.php
```diff
@@ -0,0 +1,65 @@
...
+
+        /** @var ImageInterface[] $images */
+        foreach ($images as $key => $image) {
+            foreach ($imagesFromDatabase as $imageFromDatabase) {
```
pamil: `$images` -> `$newImages`
`$imagesFromDatabase` -> `$persistedImages`
?

src/Sylius/Bundle/CoreBundle/Validator/Constraints/ImageUniqueCodeValidator.php
```diff
@@ -0,0 +1,65 @@
...
+                if ($imageFromDatabase->getOwner() === $image->getOwner() && $imageFromDatabase->getCode() === $image->getCode()) {
+                    $this->context->addViolationAt(sprintf('[%d].code', $key), $constraint->message);
```
pamil: `addViolationAt` is deprecated, we should use `buildViolation($message)->atPath($path)->addViolation()`.

src/Sylius/Bundle/CoreBundle/Validator/Constraints/ImageUniqueCodeValidator.php
pamil: We can filter them by owner.

src/Sylius/Bundle/CoreBundle/Validator/Constraints/ImageUniqueCodeValidator.php
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public function __construct(RepositoryInterface $resourceRepository)
+    {
+        $this->resourceRepository = $resourceRepository;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function validate($images, Constraint $constraint)
```
pamil: Btw. isn't something like that enough? Do we even need to use the repository?

```
$imagesCodesUsage = array_count_values(array_map(function (ImageInterface $image) { 
  return $image->getCode(); 
}, $images));

$duplicatedImages = array_filter($images, function (ImageInterface $image) use ($imagesCodesUsage) {
  return $imagesCodesUsage[$image->getCode()] > 1;
});

foreach (array_keys($duplicatedImages) as $key) {
  $this->context->addViolationAt(sprintf('[%d].code', $key), $constraint->message);
}
```

src/Sylius/Bundle/CoreBundle/spec/Validator/Constraints/ImageUniqueCodeValidatorSpec.php
```diff
@@ -0,0 +1,112 @@
...
+
+        $thirdImage->getCode()->willReturn('abs');
+
+        $firstImage->getOwner()->willReturn($owner);
+        $secondImage->getOwner()->willReturn($owner);
+
+        $firstImage->getCode()->willReturn('car');
+        $secondImage->getCode()->willReturn('car');
+
+        $this->bulidViolation('[0].code', $constraint->message)->shouldBeCalled();
```
pamil: `$constraint->message` is null here.

src/Sylius/Bundle/CoreBundle/spec/Validator/Constraints/ImageUniqueCodeValidatorSpec.php
```diff
@@ -0,0 +1,112 @@
...
+    {
+        $this->shouldHaveType(ConstraintValidator::class);
+    }
+    
+    function it_adds_violation_if_there_two_images_with_the_same_owner_which_have_same_codes(
+        ImageInterface $firstImage,
+        ImageInterface $secondImage,
+        ImageInterface $thirdImage,
+        ImageAwareInterface $owner,
+        ImageUniqueCode $constraint,
```
pamil: We should use `new` to create it inside spec, mocking constraint gives us no profit.

src/Sylius/Bundle/CoreBundle/spec/Validator/Constraints/ImageUniqueCodeValidatorSpec.php
```diff
@@ -0,0 +1,112 @@
...
+        RepositoryInterface $imageRepository
+    ) {
+        $imageRepository->findBy(['owner' => $owner])->willReturn([$firstImage]);
+
+        $secondImage->getOwner()->willReturn($owner);
+
+        $firstImage->getCode()->willReturn('car');
+        $secondImage->getCode()->willReturn('car');
```
pamil: Is second image even used in this spec?

src/Sylius/Bundle/CoreBundle/spec/Validator/Constraints/ImageUniqueCodeValidatorSpec.php
tuka217: And what kind of profit gives creating new constraint? It has only an message.

pamil: Which is equal to null while mocking it :)

tuka217: But message can be any string. In fact it should be argument with type string

pamil: Yup, my bad, mocking does indeed preserve the properties value (but does not run the constructor).

src/Sylius/Bundle/CoreBundle/spec/Validator/Constraints/ImageUniqueCodeValidatorSpec.php
```diff
@@ -0,0 +1,112 @@
...
+        $this->shouldHaveType(ConstraintValidator::class);
+    }
+    
+    function it_adds_violation_if_there_two_images_with_the_same_owner_which_have_same_codes(
+        ImageInterface $firstImage,
+        ImageInterface $secondImage,
+        ImageInterface $thirdImage,
+        ImageAwareInterface $owner,
+        ImageUniqueCode $constraint,
+        RepositoryInterface $imageRepository
```
pamil: Should be the first argument as it's injected to the constructor.

src/Sylius/Bundle/CoreBundle/spec/Validator/Constraints/ImageUniqueCodeValidatorSpec.php
tuka217: And ok, it should be create by new. But not because that the message is null, but more because the kind of instance of ImageUnigeCode does not metter.

src/Sylius/Bundle/CoreBundle/Validator/Constraints/ImageUniqueCodeValidator.php
tuka217: But first step return an warning which stopped a work of validator.

pjedrzejewski:
@tuka217 Is it still a WIP?

tuka217:
Yups, 5 minutes.

src/Sylius/Bundle/CoreBundle/Validator/Constraints/ImageUniqueCodeValidator.php
```diff
@@ -0,0 +1,43 @@
...
+                        && $imageFromObject->getCode() === $image->getCode()
+                        && $imageFromObject !== $image
+                ;
+            });
+
+            if(0 !== count($filteredImages)) {
+                $this->context->addViolationAt(sprintf('[%d].code', $key), $constraint->message);
+            }
+        }
+    }
```
michalmarcinkowski: I think we don't have to check the owner, as we always receive collection of images of a single "owner". The following implementation would be simpler:
```
$imagesCodes = [];

/** @var ImageInterface[] $images */
foreach ($images as $key => $image) {
    if (array_key_exists($imagesCodes[$image->getCode()])) {
        $this->context->addViolationAt(sprintf('[%d].code', $key), $constraint->message);
    }
    $imagesCodes[$image->getCode()] = $image;
}
```

michalmarcinkowski: How about the following?
```
$imagesCodes = [];

/** @var ImageInterface[] $images */
foreach ($images as $key => $image) {
    if (!array_key_exists($imagesCodes[$image->getCode()])) {
        $imagesCodes[$image->getCode()] = $key;
        continue;
    }

    $this->context->addViolationAt(sprintf('[%d].code', $key), $constraint->message);
    if (false !== $imagesCodes[$image->getCode()]) {
        $this->context->addViolationAt(sprintf('[%d].code', $imagesCodes[$image->getCode()]), $constraint->message);
        $imagesCodes[$image->getCode()] = false;
    }
}
```

michalmarcinkowski: Filtering is like `for` loop, so for each image you iterate over all images and the complexity is exponential O(n^2), so for 10 images we have 100 iterations (10*10). With the above code the complexity is linear O(n), so we would have only 10 iterations.

tuka217: Then ok.

pamil:
```php
$imagesCodes = [];

/** @var ImageInterface[] $images */
foreach ($images as $key => $image) {
        $imagesCodes[$image->getCode()][] = $key;
}

foreach ($imagesCodes as $code => $keys) {
    if (count($keys) <= 1) {
        continue;
    }

    foreach ($keys as $key) {
        $this->context->addViolationAt(sprintf('[%d].code', $key), $constraint->message);
    }
}
```

Quite more iterations but is easier to understand IMHO.

tuka217: It isn't so hard to understand what the first solution do. And solution with "filter" has only two loops, so if you want the more readable solution I would chose this.

michalmarcinkowski: Considering only collection of images it should never be so big that the complexity will influence the performance in any noticeable way, so I'm for @pamil's solution as it's clean and easy to understand. When there will be use cases for bigger collections and performance issues we can provide less complex solution.

michalmarcinkowski: @tuka217 the current solution is readable but doesn't assign validation error to the first image with duplicated code.

tuka217: It's assign, but I change it do your solution and please, leave it in this form. It's not so hard to understand what this method do.

pjedrzejewski:
Thank you Ania, good job! ;)

PR #6322, opened 2016-10-06
[Resource] Remove DateExtension

pamil:
It seems like it is not used anywhere in our codebase.


michalmarcinkowski:
Thanks Kamil! 👍

PR #6134, opened 2016-09-16
[Cart][Order] Merge Cart into Order

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets |
| License         | MIT

It's a huge BC Break, I suppose one of the lasts before *Sylius* beta. It completely remove *CartBundle* and move its features into *OrderBundle* (and some to *CoreBundle*).

To-do:
- [x] fix some failing tests
- [x] remove/change documentation
- [x] rebase with ``master`` (it's going to be a nightmare 😄 💀 - EDIT: in fact, not at all)
- [x] some refactoring would be nice
- [ ] consider naming (although we don't have *Cart* bundle and component, there are still some places where *Cart* concept is more adequate than *Order*)

PR #6323, opened 2016-10-06
[OrderBundle] Fix Add-to-cart Pre-create event

### LINKED ISSUE #6318: listeners for sylius.cart_item.pre_create are not called
If I declare a listener like this : 

    my_app.test:
        class: My\Bundle\Service\Listener
        tags:
            - { name: kernel.event_listener, event: sylius.cart_item.pre_create, method: preCreate }

It is never called when I add an item in the cart. `sylius.order_item.pre_create` does not work either.
According to this PR https://github.com/Sylius/Sylius/pull/6134 and the actual master codebase (namely in `OrderItemController::addAction()` ), it looks like the event is dispatched, but the listeners are never called.

Am I missing something ?

tchapi:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #6318
| License         | MIT

Fix `sylius.cart_item.pre_create` not being fired due to incorrect naming. Now, the event we can listen to is `sylius.order_item.pre_create`. A `sylius.order_item.post_create` is now also available.

### DIFF
--- a/src/Sylius/Bundle/OrderBundle/Controller/OrderItemController.php
+++ b/src/Sylius/Bundle/OrderBundle/Controller/OrderItemController.php
@@ -72,7 +72,7 @@ class OrderItemController extends ResourceController
         if ($request->isMethod('POST') && $form->submit($request)->isValid()) {
             $newResource = $form->getData();
 
-            $event = $this->eventDispatcher->dispatchPreEvent('sylius.cart_item.pre_create', $configuration, $newResource);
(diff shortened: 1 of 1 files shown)


michalmarcinkowski:
This bug should be fixed in after review fixes of CartBundle merge (see [my comment](https://github.com/Sylius/Sylius/pull/6134#discussion_r79985794)). @Zales0123 please make sure all review fixes were applied 😉

michalmarcinkowski:
Thank you @tchapi! 👍

PR #6324, opened 2016-10-06
[Payum] Fix PayumBundle -> SyliusPayumBundle -> SyliusPaymentBundle integration

pamil:
Currently it only processes the first `payum` configuration, this fix makes it processing all `payum` configurations.


pjedrzejewski:
Thanks Kamil!

PR #6325, opened 2016-10-06
[Resource] Remove Twig SyliusResourceExtension

pamil:
It seems like it is not used in our codebase.


pjedrzejewski:
👍

jrobeson:
It wasn't used in the sylius codebase, but that doesn't mean it was unused.  I wish this had been documented.

PR #6340, opened 2016-10-06
[BugFix] Apply payment state transition for Payum Notify requests

### LINKED ISSUE #6337: [Bug][Payum] Update payment state extension does not work for notify
Hi,

I noticed that the current implementation of the `PaymentStateExtension` ([source](https://github.com/Sylius/Sylius/blob/4bf30cee02d6d066dd3aff568b54c03c98791475/src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php#L67)) is not taking `notify` requests into account. Some providers will use the notify url to later mark it as paid. As of right now, when a notify request is triggered the payment details are updated properly but not its state. This is because it always expects a top-level `GetStatus` request otherwise the whole payment state check is skipped.

I propose we check for both `GetStatusInterface` and `Notify` so the payment state is updated properly. Also the priority of the `PaymentStateExtension` should be before the `DoctrineStorage` extension, otherwise the payment state isn't even persisted to the DB.

/cc @makasim

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no|yes
| BC breaks?      | no|yes
| Related tickets | fixes #6337
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
+++ b/src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
@@ -15,6 +15,7 @@ use Payum\Core\Extension\Context;
 use Payum\Core\Extension\ExtensionInterface;
 use Payum\Core\Request\Generic;
 use Payum\Core\Request\GetStatusInterface;
+use Payum\Core\Request\Notify;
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
👍

pjedrzejewski:
Thank you Steffen!

PR #6349, opened 2016-10-07
[Composer] Update testing dependencies

pamil:
- Update PHPUnit to ^5.6
 - Update Behat to ^3.2 (killing Behat with ^C works now)
 - Update PHPSpec to ^3.1
 - Update Sylius' composer.lock


pamil:
We need to fix our transformers before we can switch to Behat 3.2:

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Behat/Context/Transform/CountryContext.php#L48
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Behat/Context/Transform/AddressingContext.php#L48
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Behat/Resources/config/suites/ui/account/customer.yml#L11-L12

Closing for now.

PR #6351, opened 2016-10-07
[Content] Clean up a little bit

pamil:


pamil:
Don't want to spam Travis, temporarily closed :)

gabiudrescu:
from what you've seen so far, is this kind of clean up speeding up the build or an installed Sylius instance?

pamil:
Actually the #6365 was opened as I couldn't reopen this one.

The goal of cleanups we're doing now is to unify the code to make it use the same conventions, finalise as many classes as possible, decouple it and remove unused code. It probably won't make Sylius noticably faster (this is the next phase), but more stable and more prepared for alpha & beta releases. The less code to maintain the better :)

gabiudrescu:
I see. well, keep up the good work.

PR #6353, opened 2016-10-07
Remove Storage component

### LINKED ISSUE #3009: Does Cookie storage actually persist?

### LINKED ISSUE #4309: [RFC] PSR-6 and Storage component.
I had an idea that the Storage component could be [PSR-6 compliant](http://www.php-fig.org/psr/psr-6/).

Difference between the two being that PSR-6 is specific to caching whereas Storage is not. So if storage were to be PSR-6 compliant, would it be done by:

Implementing Storage and decorating with PSR-6:

```php
use Psr\Cache\CacheItemPoolInterface;
use Sylius\Component\Storage\StorageInterface;

class Psr6Storage implements StorageInterface
{
    /** @var CacheItemPoolInterface */
    private $pool;

    public function setData($key, $value)
    {
        $item = new CacheItem($key, $value);

        $this->pool->save($item);
    }

    // other methods.....
}
```

Or implementing PSR-6 and decorating with Storage:

```php
use Psr\Cache\CacheItemPoolInterface;
use Psr\Cache\CacheItemInterface;
use Sylius\Component\Storage\StorageInterface;

class StoragePool implements CacheItemPoolInterface
{
    /** @var StorageInterface */
    private $storage;

    public function save(CacheItemInterface $item)
    {
        $this->storage->setData($item->getKey(), $item);
    }

    // other methods.....
}
```

If I've understood the spec correctly (and I must admit without examples, it's tricky to understand) then the second option would be the more compliant one.

### LINKED ISSUE #6345: [Resource] Do not require Storage component
Storage component is not used anywhere, but in services definitions:

```xml
<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
    <services>
        <service id="sylius.storage.cookie" class="Sylius\Component\Storage\CookieStorage" public="false">
            <call method="setRequest">
                <argument type="service" id="request" on-invalid="null" strict="false" />
            </call>
        </service>

        <service id="sylius.storage.session" class="Sylius\Component\Storage\SessionStorage" public="false">
            <argument type="service" id="session" />
        </service>
    </services>
</container>
```

And:

 - `sylius.storage.cookie` is not used anywhere in Sylius
 - `sylius.storage.session` is used only by `sylius.storage.currency` and `sylius.storage.locale`

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | fixes #3009, fixes #3437, fixes #4309, fixes #6345
| License         | MIT


pjedrzejewski:
Thanks Kamil!

PR #6321, opened 2016-10-06
[Order] Automatically cancel unpaid orders

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets |
| License         | MIT

This PR is based on https://github.com/Sylius/Sylius/pull/6291.

PR #6291, opened 2016-10-05
[Order] Automatically delete expired carts

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets |
| License         | MIT

To-do:
- [x] command used to delete expired carts

PR #6356, opened 2016-10-08
[OrderBundle] Bug fix - add flush after cancel expired order/cart.

liverbool:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | #6321 #6291
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/OrderBundle/Command/CancelUnpaidOrdersCommand.php
+++ b/src/Sylius/Bundle/OrderBundle/Command/CancelUnpaidOrdersCommand.php
@@ -43,5 +43,7 @@ class CancelUnpaidOrdersCommand extends ContainerAwareCommand
 
         $unpaidCartsStateUpdater = $this->getContainer()->get('sylius.unpaid_orders_state_updater');
         $unpaidCartsStateUpdater->cancel();
+
+        $this->getContainer()->get('sylius.manager.order')->flush();
     }
 }
--- a/src/Sylius/Bundle/OrderBundle/Command/RemoveExpiredCartsCommand.php
+++ b/src/Sylius/Bundle/OrderBundle/Command/RemoveExpiredCartsCommand.php
@@ -43,5 +43,7 @@ class RemoveExpiredCartsCommand extends ContainerAwareCommand
 
         $expiredCartsRemover = $this->getContainer()->get('sylius.expired_carts_remover');
         $expiredCartsRemover->remove();
+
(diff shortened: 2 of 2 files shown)


michalmarcinkowski:
Thanks Ishmael! 👍

pjedrzejewski:
Actually, the service should call the flush. @Zales0123 How is the scenario passing?

michalmarcinkowski:
@pjedrzejewski not sure about that, now there is no service that does that, if all services would call `flush` we would have too many flushes during single request.

The scenario is passing because the entity is loaded into unity of work and there is no "refresh from database" between service call and assertion. Can we enforce that?

pjedrzejewski:
I am of course against calling flush from inside most of services, but in this case, the service without flush does not make any sense. And you only call this service if you want to *really* cancel/delete these orders.

liverbool:
😶 hard to say.

`flush` inside service make sense if this service doesn't use together with others.

pjedrzejewski:
@liverbool I think in case of services like this, it makes sense. They are meant to be run separately. I just want to avoid situation when someone is surprised that carts are not cleared / cancelled when this is called.

Zales0123:
I think the reason it works now, is because we're calling repository [``remove`` method](https://github.com/Sylius/Sylius/blob/6cb794ef5bbf43d7cea6bb7a9229a60d3e0926a9/src/Sylius/Component/Order/Remover/ExpiredCartsRemover.php#L45) inside each of this services. It implies ``flush`` [after each removal](https://github.com/Sylius/Sylius/blob/4bf30cee02d6d066dd3aff568b54c03c98791475/src/Sylius/Bundle/ResourceBundle/Doctrine/ORM/EntityRepository.php#L43), which also not good, as manager is flushed multiple times :/

pjedrzejewski:
Oh, that's bad - let's add ``remove`` on the manager and do a late flush.

Zales0123:
Good for me 👍  This requires of course injecting a manager, but it's lesser evil, I suppose ;)

liverbool:
i dose't look in https://github.com/Sylius/Sylius/blob/4bf30cee02d6d066dd3aff568b54c03c98791475/src/Sylius/Bundle/ResourceBundle/Doctrine/ORM/EntityRepository.php#L43 that surprised me, why we need `manager.flush()` again !! (in my PR)

Issue #6262, opened 2016-10-02
[User] Forgot Password


inssein:
Sorry for the barrage of issues this weekend, just creating them as I see, and will fix what I get time for.

Forgot Password doesn't work the way it is, and there are a few problems:
1) The email sends the wrong link.
2) Once the link is adjusted, you realize "sylius.user.resetting.token_ttl" is not set anywhere, and it has to be set to a DateTime interval (P1D).
3) Once you do that, the template for resetting is broken because of a bad route

pjedrzejewski:
@inssein Could you include any error logs/messages, please?

inssein:
For each of the issues? If any one of those is not clear, let me know and I'll do my best by including logs and/or exactly where its incorrect.

Issue #5859, opened 2016-08-21
Route 'sylius_user_security_login' does not exist anymore


tchapi:
In **UserBundle\Controller\UserController**, line  [217](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/UserBundle/Controller/UserController.php#L217), there is still a reference to a deleted route : 

```
$redirectRouteName = $request->attributes->get('_sylius[redirect]', 'sylius_user_security_login', true);
```

should be : 

```
$redirectRouteName = $request->attributes->get('_sylius[redirect]', 'sylius_shop_login', true);
```

There are some other occurences throughout the file too.

lchrusciel:
Can be closed /cc @michalmarcinkowski

tchapi:
Well, `sylius_user_security_login` is still present in the SecurityController and this route does not exist ...

lchrusciel:
Oh, sorry. My mistake. Thanks @tchapi

PR #6357, opened 2016-10-08
[Shop] Fix Forgot Password

inssein:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/6262
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/views/Email/passwordReset.html.twig
+++ b/src/Sylius/Bundle/CoreBundle/Resources/views/Email/passwordReset.html.twig
@@ -3,5 +3,5 @@
 {% endblock %}
 
 {% block body %}
-    {{ 'sylius.email.password_reset.content'|trans({'%username%': user.username, '%url%': url('sylius_shop_request_password_reset_token', { 'token' : user.passwordResetToken})})|raw }}
+    {{ 'sylius.email.password_reset.content'|trans({'%username%': user.username, '%url%': url('sylius_shop_password_reset', { 'token' : user.passwordResetToken})})|raw }}
 {% endblock %}
--- a/src/Sylius/Bundle/ShopBundle/Resources/config/routing/security.yml
+++ b/src/Sylius/Bundle/ShopBundle/Resources/config/routing/security.yml
@@ -39,9 +39,14 @@ sylius_shop_request_password_reset_token:
             template: SyliusShopBundle:Account:requestPasswordReset.html.twig
             redirect: sylius_shop_login
 
-sylius_shop_password_resetting:
-    resource: "@SyliusUserBundle/Resources/config/routing/password_reset.yml"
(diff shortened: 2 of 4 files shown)


tchapi:
Password reset still doesn't work because of https://github.com/Sylius/Sylius/issues/5859 : the route 'sylius_user_security_login' does not exist anymore and is called in UserController when resetting the password.

inssein:
@tchapi not true in this case. You can override the route like done in this PR (https://github.com/Sylius/Sylius/pull/6357/files#diff-d64ce4611cdad270fd9f2da7a8749f3dR49).

src/Sylius/Bundle/CoreBundle/Resources/views/Email/passwordReset.html.twig
```diff
@@ -3,5 +3,5 @@
 {% endblock %}
 
 {% block body %}
-    {{ 'sylius.email.password_reset.content'|trans({'%username%': user.username, '%url%': url('sylius_shop_request_password_reset_token', { 'token' : user.passwordResetToken})})|raw }}
+    {{ 'sylius.email.password_reset.content'|trans({'%username%': user.username, '%url%': url('sylius_shop_password_reset', { 'token' : user.passwordResetToken})})|raw }}
```
lchrusciel: This file is placed in CoreBundle but it is coupled to shop-bundle. It should be moved out of here or a route should be declared both in core and shop bundle(with some other name)

tchapi:
@inssein sure but still, a non-existent route in a Core file seems odd to me, no ? We shouldn't have to override anything to make this feature work, right ? (I mean, the global "reset password" feature)

inssein:
@lchrusciel what do you suggest? I saw that problem, but it was already happening since all emails are in Core, for example: (https://github.com/inssein/Sylius/blob/fc4fc48c0a7776b1eee962bf34838bf213c2b13e/src/Sylius/Bundle/CoreBundle/Resources/views/Email/verification.html.twig)

@tchapi its not a non existent route in core, its a route in UserBundle that is supposed to be defined when integrated. I agree it seems silly, but thats how all the login stuff is currently done. I am open to modifying this PR so that it is done a better way.

lchrusciel:
So all of this files are placed in a wrong bundle. CoreBundle should be usable without admin nor shop bundle. But it is not with the current implementation.

inssein:
@lchrusciel having a hard time thinking of where to move them. I initially thought ShopBundle, but then API related actions should also send these emails. CoreBundle is the only thing connecting these bundles.

Maybe there is a way to optionally couple these bundles?

michalmarcinkowski:
Thank you Hussein! 👍 

We will try to decouple the routing in a separate PR.

PR #6361, opened 2016-10-10
[Payment] Sylius stripe integration

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | -
| License         | MIT


composer.json
```diff
@@ -75,7 +75,8 @@
         "willdurand/hateoas-bundle": "^0.4|^1.0",
         "winzou/state-machine-bundle": "~0.2",
         "zendframework/zend-hydrator": "^1.0|^2.0",
-        "zendframework/zend-stdlib": "^2.0|^3.0"
+        "zendframework/zend-stdlib": "^2.0|^3.0",
+        "stripe/stripe-php": "^4.0"
```
pjedrzejewski: That should be in require-dev.

src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml
```diff
@@ -115,6 +115,19 @@ payum:
...
+            username: EDITME
+            password: EDITME
+            signature: EDITME
+            sandbox: true
+        credit_card:
+            factory: stripe_checkout
+            secret_key: EDITME
+            publishable_key: EDITME
+            payum.template.layout: SyliusShopBundle::Checkout/layout.html.twig
+            payum.template.obtain_token: SyliusPayumBundle::Action/obtainCheckoutToken.html.twig
```
pjedrzejewski: Why configuring it in main config? Default config should contain 0 gateways (except offline). These should be configured in test env only and devs should configure these per project for prod/dev.

src/Sylius/Bundle/PayumBundle/Resources/views/Action/obtainCheckoutToken.html.twig
```diff
@@ -0,0 +1,25 @@
...
+    </form>
+{% endblock %}
+
+{% block javascripts %}
+    {{ parent() }}
+
+    <script type="text/javascript">
+        $(function() {$('form .stripe-button-el').click();});
+    </script>
+{% endblock %}
```
lchrusciel: Missing blank line

src/Sylius/Bundle/PayumBundle/composer.json
```diff
@@ -33,7 +33,8 @@
         "sylius/currency": "^1.0",
         "sylius/order-bundle": "^1.0",
         "sylius/payment-bundle": "^1.0",
-        "sylius/resource-bundle": "^1.0"
+        "sylius/resource-bundle": "^1.0",
+        "stripe/stripe-php": "^4.0"
```
lchrusciel: Same as mentioned [above](https://github.com/Sylius/Sylius/pull/6361/files#r82559838).

makasim: it is not supposed to be added here.

inssein:
I mentioned this in one of the tickets, but I will mention it here as well. Sorry if it doesn't apply to this PR necessarily, as I haven't figured out exactly where the bug is.

If you use any part of the Symfony / Twig form rendering bits in a page that could take the credit card information, or in this case, the checkout token, it breaks as it cannot find the base twig form files.

Arminek:
@inssein It is default behaviour of payum in this case it doesn't render form. If you want to render credit card form you need to override Stripe ``CaptureAction`` and execute ``ObtainCreditCardAction``.

pjedrzejewski:
Thank you Arek! :)

PR #6371, opened 2016-10-10
Use variadic instead of call_user_func_array

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

Using variadic arguments instead of `call_user_func_array` is faster (it's the reason of [PerformanceExtension](https://github.com/FriendsOfBehat/PerformanceExtension) too).


pjedrzejewski:
👍 👍 👍

PR #6374, opened 2016-10-10
Fix Customer Group Type and RuleChecker

tchapi:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

Fix the CustomerGroupType and RuleChecker

### DIFF
--- a/src/Sylius/Component/Core/Promotion/Checker/Rule/CustomerGroupRuleChecker.php
+++ b/src/Sylius/Component/Core/Promotion/Checker/Rule/CustomerGroupRuleChecker.php
@@ -39,7 +39,7 @@ class CustomerGroupRuleChecker implements RuleCheckerInterface
             return false;
         }
 
-        if (is_null($customer->getGroup())) {
+        if (null === $customer->getGroup()) {
             return false;
         }
 
--- a/src/Sylius/Component/Core/Promotion/Checker/Rule/CustomerGroupRuleChecker.php
+++ b/src/Sylius/Component/Core/Promotion/Checker/Rule/CustomerGroupRuleChecker.php
@@ -39,7 +39,11 @@ class CustomerGroupRuleChecker implements RuleCheckerInterface
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


src/Sylius/Component/Core/Promotion/Checker/Rule/CustomerGroupRuleChecker.php
```diff
@@ -39,7 +39,11 @@ public function isEligible(PromotionSubjectInterface $subject, array $configurat
             return false;
         }
 
-        if ($configuration['groups'] === $customer->getGroup()->getId()) {
+        if (is_null($customer->getGroup())) {
```
michalmarcinkowski: We have the following convention for checking for `null` value: `if (null === $customer->getGroup()) {`

src/Sylius/Component/Core/Promotion/Checker/Rule/CustomerGroupRuleChecker.php
```diff
@@ -39,7 +39,11 @@ public function isEligible(PromotionSubjectInterface $subject, array $configurat
...
+            return false;
+        }
+
+        if ($configuration['group'] === $customer->getGroup()->getId()) {
```
michalmarcinkowski: We should use `code`, same as we do in `TaxonRuleChecker` (can be done in a separate PR).

src/Sylius/Component/Core/Promotion/Checker/Rule/CustomerGroupRuleChecker.php
tchapi: alright, changed

pjedrzejewski:
Thanks @tchapi! Regarding using the code, we can handle that in a separate PR or when covering this functionality with scenarios.

tchapi:
Yes, that's what I thought too because it's not really related to this PR in particular

PR #6383, opened 2016-10-11
[Channel][Currency][Locale][Money] Clean up components and bundles

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

- [x] Channel
- [x] ChannelBundle
- [x] Currency
- [x] CurrencyBundle
- [x] Locale
- [x] LocaleBundle
- [x] MoneyBundle

### DIFF
--- a/src/Sylius/Behat/Resources/config/services/contexts/setup.xml
+++ b/src/Sylius/Behat/Resources/config/services/contexts/setup.xml
@@ -62,7 +62,7 @@
 
         <service id="sylius.behat.context.setup.locale" class="Sylius\Behat\Context\Setup\LocaleContext" scope="scenario">
             <argument type="service" id="sylius.behat.shared_storage" />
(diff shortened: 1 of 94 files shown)


pjedrzejewski:
Why renaming ``sylius.context.currency`` with ``sylius.currency_context``? I think we can identify a common group of "context" services in Sylius. I think it was good DX to know that you can always look for ``sylius.context.X`` service to get something. I understand it is a bit against my arguments for grouping services (you can't really swap contexts, they do not share an interface), but I have a feeling that was a good naming convention in terms of DX. Any thoughts?

GSadee:
I've been suggested by what we were talking about earlier, but of course I understand your arguments and what is more, I agree with them 😃  So I'll revert these changes.

pjedrzejewski:
Thank you Grzesiu!

PR #6389, opened 2016-10-11
[Installer] Fix getCode on null exception

### LINKED ISSUE #5762: [BUG] Sylius installer can be called only once
It is possible to load sample data only on a fresh database. If it previously contains any data, an installer would throw an exception. 

Step to reproduce: 
1. Clean Sylius installation (without any database).
2. app/console sylius:install
3. app/console sylius:install one more time
4. Say yes for database reset.
5. Say yes for load sample data.
6. See the following result:
zrzut ekranu 2016-08-11 o 09 03 43

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #5762 
| License         | MIT

I don't remember, why did we do this magic with unsetting logger, but at least it is the fastest working solution.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Command/InstallSampleDataCommand.php
+++ b/src/Sylius/Bundle/CoreBundle/Command/InstallSampleDataCommand.php
@@ -60,16 +60,10 @@ EOT
             return 1;
         }
 
-        $doctrineConfiguration = $this->get('doctrine.orm.entity_manager')->getConnection()->getConfiguration();
-        $logger = $doctrineConfiguration->getSQLLogger();
-        $doctrineConfiguration->setSQLLogger(null);
-
         $commands = [
             'sylius:fixtures:load' => ['--no-interaction' => true],
         ];
 
         $this->runCommands($commands, $input, $output);
-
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
We do it to avoid logging all sql queries from fixtures (performance ofc.), are you sure it is a proper fix? Can you compare the perf. gains? Thanks!

lchrusciel:
Tomorrow I will do some profiling.

lchrusciel:
Unfortunately, the difference is noticeable. It consumes 11% more memory
profile

lchrusciel:
I cannot spend more time at the moment on it, but maybe somebody will debug it further.

pix-art:
I'm looking into this issue as we speak

lchrusciel:
Closing in favour of #6397

lchrusciel:
As it is still not resolved I will keep it open.

pjedrzejewski:
Thanks Łukasz! We will look into optimizing this after the tag.

PR #6397, opened 2016-10-12
[WIP][Installer] #5762 fixed by adding a priority to the country fixture

### LINKED ISSUE #5762: [BUG] Sylius installer can be called only once

pix-art:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #5762, mentioned in #6389
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
@@ -13,6 +13,7 @@ sylius_fixtures:
                 currency: ~
 
                 geographical:
+                    priority: 100
                     options:
                         countries:
                             - US


pix-art:
Keep this on hold, i think the root of this problem is found lower down the chain. I'm looking into it right now.

lchrusciel:
Can you add a [WIP] tag to the title then?

PR #6399, opened 2016-10-13
[Shop] Sort product by price

tuka217:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets |-
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Setup/ProductContext.php
+++ b/src/Sylius/Behat/Context/Setup/ProductContext.php
@@ -540,6 +540,15 @@ final class ProductContext implements Context
     }
 
     /**
+     * @Given /^(this product)'s price is "([^"]+)"$/
+     */
+    public function thisProductPriceIs(ProductInterface $product, $price)
+    {
+        $product->getFirstVariant()->setPrice($this->getPriceFromString($price));
+        $this->objectManager->flush($product);
+    }
+
+    /**
      * @param string $type
      * @param string $name
      * @param string $code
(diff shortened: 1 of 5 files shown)


src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -540,6 +540,15 @@ public function thisProductHasAnImageWithACode(ProductInterface $product, $image
     }
 
     /**
+     * @Given /^(this product)'s price is "([^"]+)"$/
+     */
+    public function thisProductPriceIs(ProductInterface $product, $price)
+    {
+        $product->getFirstVariant()->setPrice($this->getPriceFromString($price));
```
michalmarcinkowski: `getFirstVariant()` should not be used anymore, we should always use `defaultVariantResolver` as we do e.g. [here](https://github.com/tuka217/Sylius/blob/fdef0bf31a94eb6b3b34c4394101870fe72af188/src/Sylius/Behat/Context/Setup/ProductContext.php#L497).

PS. We should remove `getFirstVariant` method from `Product` in a separate PR.

michalmarcinkowski:
Thanks Ania! 👍

Zales0123:
Something is wrong with the build, it should fail :/
zrzut ekranu 2016-10-13 o 12 36 51

michalmarcinkowski:
Probably "rerun" overrides the result of the build, but don't take into account that there were undefined steps... :/

/cc @pamil

PR #6401, opened 2016-10-13
[Order][Promotion] Clean up

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | 
| License         | MIT

- [x] Order
- [x] OrderBundle
- [x] Promotion
- [x] PromotionBundle


src/Sylius/Bundle/OrderBundle/Resources/config/routing.yml
```diff
@@ -3,20 +3,25 @@
...
     path: /add
-    defaults: { _controller: sylius.controller.order_item:addAction }
+    defaults:
+        _controller: sylius.controller.order_item:addAction
 
 sylius_cart_item_remove:
     path: /{id}/remove
-    defaults: { _controller: sylius.controller.order_item:removeAction }
+    defaults:
+        _controller: sylius.controller.order_item:removeAction
```
michalmarcinkowski: I think we decided to remove routing from bundles.

src/Sylius/Bundle/OrderBundle/Resources/config/validation/Adjustment.xml
```diff
@@ -1,34 +0,0 @@
...
-                <option name="groups">sylius</option>
-            </constraint>
-            <constraint name="Type">
-                <option name="type">integer</option>
-                <option name="message">sylius.adjustment.amount.integer</option>
-                <option name="groups">sylius</option>
-            </constraint>
-        </property>
-    </class>
-</constraint-mapping>
```
michalmarcinkowski: Why removing validation?

src/Sylius/Bundle/OrderBundle/Resources/translations/flashes.en.yml
```diff
@@ -4,8 +4,5 @@
 sylius:
     cart:
         cart_save_completed: The cart has been successfully updated.
-        cart_clear_completed: The cart has been successfully cleared.
         item_add_completed: Item has been added to cart.
         item_remove_completed: Item has been removed from cart.
```
michalmarcinkowski: These flashes are probably not used in bundle, have we decided where to put them?

src/Sylius/Bundle/OrderBundle/Resources/config/routing.yml
GSadee: Ok, I remove this file

src/Sylius/Bundle/OrderBundle/Resources/config/validation/Adjustment.xml
GSadee: I guess that we don't need validation for an adjustment as much as we don't need form type for this entity. Should I revert this change?

michalmarcinkowski: We have CRUD for adjustments in ApiBundle, but we don't have tests for that, so it probably doesn't work now. Validation works not only for form types, but anytime the object is passed to the validator. Despite the fact that we don't verify this validation anywhere it is so simple and basic that I would leave it (we have such constraints on the database level as well).

michalmarcinkowski:
Good job Grzesiu!

PR #6406, opened 2016-10-13
[Admin] Fix customer gird confirmed label.

liverbool:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | ~
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Customer/Field/confirmed.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Customer/Field/confirmed.html.twig
@@ -1,5 +1,5 @@
 {% if data.user is not null %}
-    {% if data.user.enabled %}
+    {% if data.user.verified %}
         <span class="ui teal label">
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/AdminBundle/Resources/views/Customer/Field/confirmed.html.twig
```diff
@@ -1,5 +1,5 @@
 {% if data.user is not null %}
-    {% if data.user.enabled %}
+    {% if data.user.verified %}
```
michalmarcinkowski: We have 2 concepts right now:
1. Account can be enabled/disabled (can or can't login)
2. Account can be verified/not verified (can login, but the email was/wasn't verified)

I think we can display `verified` in admin right now since we with current logic all user accounts are enabled by default, but we should also rename the field in grid to `verified` instead of `confirmed` to make it clear.

michalmarcinkowski:
Fixed in #6440.

PR #6418, opened 2016-10-14
[Development] Vagrant rework to PHP 7, MySQL 5.6, Apache/FPM, Debian

pix-art:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

I've been fixing some small bugs on sylius for a long time now, and every time that i have to setup an environment I got frustrated that it just didn't work out the box. We've had a vagrant setup with puppet for a while now and it never functions by just running `vagrant up`. 

I've rewritten the whole `etc/vagrant` with a box that does a complete install + composer install + sylius:install --no-interaction

So now if you follow the `etc/vagrant/README.md` you will get exactly what you need, this should help a lot of devs who wanna do a quick bug fix and don't wanna do the hassle of setting up their own environment.

**Setup:**
Apache/2.4.10 (Debian)
PHP 7.0.11-1~dotdeb+8.1
PHP-FPM
MySQL 5.6.32-78.1 Percona Server

**Extra tools:**
Composer 1.2.1
Git 2.1.4
Vim
Zip/Unzip
Curl

screenshot 2016-10-14 15 44 23


pix-art:
Would people be interested maybe in adding Node/Npm so while provisioning we also install our assets with gulp? Because this box creates a working sylius but doesn't install the assets with gulp?

lchrusciel:
@pix-art I think I would be great

pjedrzejewski:
@pix-art Definitely!

pix-art:
There you go, now the readme has been updated, the whole sylius package works instantly when running the vagrant up command no other actions required.

My pleasure :D

pjedrzejewski:
Awesome work Joeri! :) This is great and perfect timing! :)

PR #6429, opened 2016-10-17
Reorganise services definitions localisation

### LINKED ISSUE #4726: [RFC] Managing services definitions across multiple files
Hello folks!

While I was surfing through Sylius, most of our `DependencyInjection\Sylius*Extension` classes loading multiple files are doing it on the PHP side like:

```php
// SyliusMoneyExtension.php

$loader->load('services.xml');
$loader->load('templating.xml');
$loader->load('twig.xml');
```

or

```php
// SyliusCurrencyExtension.php

$configFiles = [
    'services.xml',
    'templating.xml',
    'twig.xml',
];

foreach ($configFiles as $configFile) {
    $loader->load($configFile);
}
```

They're all files that are always loaded no matter what happens, so I came up with an idea to include additional files in the main one (usually `services.xml`) using imports:

```php
$loader->load('services.xml');
```

```xml
<imports>
    <import resource="templating.xml" />
    <import resource="twig.xml" />
</imports>
```

This approach will apply only to files statically included, the dynamic ones like database driver dependent configuration (#4721) or the ones included only in test environment will still be loaded in extension class.

What do you think? :camel: :dromedary_camel:

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #4726
| License         | MIT


michalmarcinkowski:
Thanks Kamil!

PR #6432, opened 2016-10-17
[Behat] Split 'delete' and 'try to delete' methods in Domain contexts

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #5832 
| License         | MIT

As explained in review comment, we shouldn't catch the exception on `When I delete product`


pjedrzejewski:
Thank you Michał!

PR #6437, opened 2016-10-17
[Promotion] Fix per channel promotions availability

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

`ActivePromotionsByChannelProvider` was not used in `Core`

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services/promotion.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services/promotion.xml
@@ -13,7 +13,7 @@
 
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/CoreBundle/Resources/config/services/promotion.xml
```diff
@@ -17,6 +17,12 @@
             <argument type="service" id="sylius.repository.promotion" />
         </service>
 
+        <service id="sylius.promotion_processor" class="Sylius\Component\Promotion\Processor\PromotionProcessor">
```
pamil: Shouldn't we overwrite `sylius.active_promotions_provider` with `ActivePromotionsByChannelProvider` instead of overwriting `sylius.promotion_processor` changing its argument?

michalmarcinkowski: Sounds like a plan! That way we ensure that proper provider is injected everywhere else 👍

michalmarcinkowski: Already fixed, couldn't wait for tomorrow 😆 

Btw. REVIEW COMMENTS ARE MARKED AS OUTDATED AGAIN!! 🎉 🎉

pjedrzejewski:
Thank you Michał! :)

PR #6438, opened 2016-10-17
[Critical][Order] Fix order cancelling issue

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

`OrderPaymentProcessor` should be called instead of `OrderProcessor` on payment fail/cancel to avoid whole order recalculation.

Based on #6437.

### DIFF
--- a/features/order/managing_orders/cancelling_order_with_promotion.feature
+++ b/features/order/managing_orders/cancelling_order_with_promotion.feature
@@ -1,0 +1,27 @@
+@managing_orders
+Feature: Cancelling order with promotion applied
+    In order to maintain correct order history
+    As an Administrator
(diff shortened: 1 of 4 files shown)


features/order/managing_orders/cancelling_order_with_promotion.feature
```diff
@@ -0,0 +1,27 @@
+@managing_orders
+Feature: Cancelling order order with promotion applied
```
tuka217: 'Cancelling order with (...)'

pjedrzejewski:
Thanks Michał!

PR #6442, opened 2016-10-18

SyliusBot:


pjedrzejewski:
@pamil Any idea why removing ``products``, ``taxons``, etc. ? I don't think these translations are gone.

pamil:
@pjedrzejewski nope, can't see any related activity on CrowdIn or in our translations files.

PR #6446, opened 2016-10-18
[RFC] Add PhpSpec templates

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

By adding these templates, PhpSpec will generate classes / interfaces / specifications with license block automatically added.

### DIFF
--- a/.phpspec/class.tpl
+++ b/.phpspec/class.tpl
@@ -1,0 +1,14 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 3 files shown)


Zales0123:
I don't respect this change :( We should base on good code review, not on forcing phpspec templates. I have my own templates implemented in ``.phpspec`` directory and I don't want to be forced this one (as I also have ``@author`` block configured etc.)

lchrusciel:
I have shaken, not stirred feelings about this PR. On the one hand it is really useful, to enforce developers to declare classes as final by default. On the other hand many people has theirs own templates defined in this place, with additional `@author` block for example.

PR #6448, opened 2016-10-18
[CoreBundle][DX] Division of the state_machine.yml file

TheMadeleine:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

In order to make the customization of state machines easier we have divided the giant ``state_machine.yml`` file into several smaller.


src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine/sylius_review.yml
```diff
@@ -0,0 +1,23 @@
...
+        transitions:
+            accept:
+                from: [new]
+                to: accepted
+            reject:
+                from: [new]
+                to: rejected
+        callbacks:
+            after:
+                update_price:
```
michalmarcinkowski: I think it should be `update_rating` :)

TheMadeleine: Totally true. @pjedrzejewski should I change it here?

pjedrzejewski:
Thanks Magda! :)

PR #6449, opened 2016-10-18
[Core] Remove TaxonRuleChecker

### LINKED ISSUE #5355: [Promotions] ContainsTaxonRuleChecker & TaxonRuleChecker
There seems to be duplication here, `ContainsTaxonRuleChecker` and `TaxonRuleChecker` are almost identical, but `ContainsTaxonRuleChecker` supports a quantity option.

I suggest removing `TaxonRuleChecker`.

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | fixes #5355
| License         | MIT

Duplicates `ContainsTaxonRuleChecker` (which is actually used).


pamil:
Actually, their behaviour isn't the same:

 - `TaxonRuleChecker` allows us to check if we have at least one product from at least one given taxon
 - `ContainsTaxonRuleChecker` allows us to check if we have N products from the given taxon

Anyway, we have to think of better naming.

pamil:
Closing PR, opening #6457.

PR #6453, opened 2016-10-18
Remove @mixin annotations in specs

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

Some time ago I started to add `@mixin` annotations in Sylius' specs due to code completion in PhpStorm for `$this->subjectUnderTestMethod()` calls. It was better than nothing.

Right now, there is a PhpStorm EAP which finally adds PhpSpec support. Unfortunately, `@mixin` annotation makes it not working correctly. I guess we can remove them, those who want to still have that SUT code completion (and much more) may use an EAP version, so we can catch and report all the bugs in the PhpSpec support for the greater good! :tada:


pjedrzejewski:
Requires rebasing.

pjedrzejewski:
👍 👍 👍

PR #6455, opened 2016-10-18
[Behat] Cleanup ProductContext

### LINKED ISSUE #5934: [RFC][Behat] Inventory units consistency
Right now in our feature files we use `items` instead of `units` for inventory steps e.g. `there is 1 item of product "PHP T-Shirt" available in the inventory`

In order to have uniform dictionary and make it consistent with implementation, when talking about inventory units (number of units of a product available in the inventory) IMO we should use the word `unit` instead of `item`. I'm not an expert in english eCommerce dictionary, but I think it will be more understandable for everyone. WDYT?

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #5934
| License         | MIT

- Rename `items` to `units` when talking about inventory (#5934)
- Use `DefaultVariantResolver` instead of `getFirstVariant`

### DIFF
--- a/features/cart/shopping_cart/adding_products_with_quantity_to_cart.feature
+++ b/features/cart/shopping_cart/adding_products_with_quantity_to_cart.feature
@@ -10,7 +10,7 @@ Feature: Adding a simple product of given quantity to the cart
 
     @ui
     Scenario: Adding a product with stated quantity to the cart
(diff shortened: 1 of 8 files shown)


pjedrzejewski:
Thank you Michał!

PR #6463, opened 2016-10-19
[Critical][Promotion] Fix per customer usage limit

### LINKED ISSUE #5412: [Promotions] countByCustomerAndCoupon BUG
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php#L201

This leftJoin is not only pointless, but it multiplies the count by the number of items in the order. So it returns 3 uses of a coupon if you had 3 orderItems.

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #5412
| License         | MIT

The `count(o.id)` was counting the same order multiple times (depending on number of items) because of `leftJoin('o.items', 'item')`.

**UPDATE**
Revealed another bug, related to `nth` order scenarios implementation.


pjedrzejewski:
👍 👍 👍

PR #6478, opened 2016-10-20
[Cart] Fix cart summary display

### LINKED ISSUE #6267: [Shop] Wrong total on order item with discounts
zrzut ekranu 2016-10-03 o 13 45 26

I suppose, order item "total" should be $560.00 - it seems as it contains also ``order`` promotions.
Promotion applied - 20% for every "jacket" product (``order_unit`` promotion) and $50 for whole order (``order`` promotion).

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #6267
| License         | MIT

### DIFF
--- a/features/cart/shopping_cart/change_currency_of_cart.feature
+++ b/features/cart/shopping_cart/change_currency_of_cart.feature
@@ -18,6 +18,6 @@ Feature: All of my cart's values get updated to the currency of my choosing
         Given I have product "The Pug Mug" in the cart
         When I switch to the "GBP" currency
         Then my cart total should be "£32.00"
-        And total price of "The Pug Mug" item should be "£22.00"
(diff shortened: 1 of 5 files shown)


pjedrzejewski:
Are you sure? This will be quite confusing. It won't include shipping? Or we have wrong implementation of subtotal. Subtotal should only items, without shipping, etc.

michalmarcinkowski:
It's `OrderItem::subTotal` not `Order::subTotal`. It returns `discountedUnitPrice * quantity` while `OrderItem::total` returns also distributed order promotions which causes the confusing display (see #6267).

pjedrzejewski:
Ahhh, perfect! Thanks Michał! :)

Issue #6190, opened 2016-09-23
Promotion Coupon Bug


bretto36:
Hi All

Found a bug today, If i have 2 different items in my cart, 

Product 1 - 10 quantity @ 100 each, 1000 total
Product 2 - 1 quantity @ 1 each, 1 total. 

Then try and apply a 10% discount, the Total discount should be $100.1. But it tries to split the promotion amount across all the items. And gives 50.05 off the first item. but reduces the next item to 0. So technically

The amount displayed in the discount line in the totals area seems correct, but it's not applied to the items correctly. 

The scenario i'm getting below. Tax is 10%. I have 2 items in the cart, unit price of 35. 11 in total of the items. Therefore cart total of 385. I'm using a 50% discount coupon. Discount of 192.50 which you can see on the right is correct. So the grand total should be 192.50 \* 10% tax = 211.75. 

The distributor is trying to distribute evenly, it's trying to distribute 96.25 to both items. But one of the items isn't more than 96.25. So it 0's that out, and forgets about the rest of the discount. 

![syliusbug](https://cloud.githubusercontent.com/assets/6217994/18775723/258e7f34-81a7-11e6-9c58-ae60648d1e9f.PNG)

michalmarcinkowski:
@bretto36 are you able to write Behat scenario that will reproduce this issue? I tried in #6479, but it looks like everything works fine. Did you do any customizations that could cause this issue in your application?

bretto36:
It's possible. I'll write a behat for it when i figure out how to run them haha

michalmarcinkowski:
@bretto36 are you able to at least provide exact instructions how to reproduce this bug?

michalmarcinkowski:
@bretto36 please reopen if this is still an issue on the newest master.

PR #6479, opened 2016-10-20
[Promotion] Better coverage of percentage promotion distribution

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | mentioned in #6190
| License         | MIT

To make sure that the promotion is applied correctly even if the discount is bigger than the item total.


pjedrzejewski:
Thank you Michał! :)

PR #6496, opened 2016-10-20
[Grid] Improve ArrayToDefinitionConverter

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Grid/Definition/ArrayToDefinitionConverter.php
+++ b/src/Sylius/Component/Grid/Definition/ArrayToDefinitionConverter.php
@@ -28,59 +28,103 @@ final class ArrayToDefinitionConverter implements ArrayToDefinitionConverterInte
         }
 
         foreach ($configuration['fields'] as $name => $fieldConfiguration) {
-            $field = Field::fromNameAndType($name, $fieldConfiguration['type']);
-
-            if (array_key_exists('path', $fieldConfiguration)) {
-                $field->setPath($fieldConfiguration['path']);
-            }
-            if (array_key_exists('label', $fieldConfiguration)) {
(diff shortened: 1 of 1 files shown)


src/Sylius/Component/Grid/Definition/ArrayToDefinitionConverter.php
```diff
@@ -28,59 +28,103 @@ public function convert($code, array $configuration)
...
+     * @param string $name
+     * @param array $configuration
+     *
+     * @return Field
+     */
+    private function convertField($name, array $configuration)
+    {
+        $field = Field::fromNameAndType($name, $configuration['type']);
+
+        if (array_key_exists('path', $configuration)) {
```
lchrusciel: Won't it be better to get keys and then use `PropertyAccessor`?

pamil: Quite too much magic for a simple task IMO.

lchrusciel: Do you think that calling `array_key_exists` 9 times is cleaner?

pjedrzejewski:
Thanks Kamil!

src/Sylius/Component/Grid/Definition/ArrayToDefinitionConverter.php
pamil: Actually, I think that these keys should be always set (configuration parser should be responsible), but since it isn't, calling **known** methods with usage of `PropertyAccessor` rather obfuscates it than clears it.

PR #6480, opened 2016-10-20
[Promotion] UI for filters

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Related tickets | continuation of #6054
| License         | MIT

First part of filters reworking. Only price range filter is implemented. More filter would be added in the next PR.

PR #6489, opened 2016-10-20
[Promotion] UI for price range and taxon filtration 

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | continuation of #6480 and #6054
| License         | MIT

Based on #6480.

PR #6507, opened 2016-10-21
[Promotion] Product filter

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | continuation for #6480 
| License         | MIT

### DIFF
--- a/features/promotion/receiving_discount/receiving_fixed_discount_on_specific_products.feature
+++ b/features/promotion/receiving_discount/receiving_fixed_discount_on_specific_products.feature
@@ -1,27 +1,30 @@
 @receiving_discount
-Feature: Receiving fixed discount on products from specific taxon
-    In order to pay less while buying goods from promoted taxons
+Feature: Receiving fixed discount on specific products
+    In order to buy specific product with a discount
     As a Customer
-    I want to receive discount for my purchase
+    I want to receive discount on each unit of promoted product
 
     Background:
         Given the store operates on a single channel in "United States"
-        And the store classifies its products as "T-Shirts" and "Mugs"
         And the store has a product "PHP T-Shirt" priced at "$100.00"
-        And it belongs to "T-Shirts"
         And the store has a product "PHP Mug" priced at "$20.00"
-        And it belongs to "Mugs"
         And there is a promotion "T-Shirts promotion"
-        And it gives "$10.00" off on every product classified as "T-Shirts"
+        And it gives "$10.00" off on a "PHP T-Shirt" product
 
     @ui
-    Scenario: Receiving fixed discount on a single item from specific taxon
+    Scenario: Receiving fixed discount on a single item
         When I add product "PHP T-Shirt" to the cart
         Then its price should be decreased by "$10.00"
         And my cart total should be "$90.00"
 
     @ui
-    Scenario: Receiving fixed discount on a multiple items from specific taxon
+    Scenario: Not receiving fixed discount on another item
+        When I add product "PHP Mug" to the cart
+        Then product "PHP Mug" price should not be decreased
+        And my cart total should be "$20.00"
+
+    @ui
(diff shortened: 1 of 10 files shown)


features/promotion/receiving_discount/receiving_fixed_discount_on_specific_products.feature
```diff
@@ -0,0 +1,64 @@
+@receiving_discount
+Feature: Receiving fixed discount on specific products
+    In order to pay less while buying promoted goods
+    As a Customer
+    I want to receive discount for my purchase
```
michalmarcinkowski: I want to receive discount on each unit of promoted product

features/promotion/receiving_discount/receiving_fixed_discount_on_specific_products.feature
```diff
@@ -0,0 +1,64 @@
+@receiving_discount
+Feature: Receiving fixed discount on specific products
+    In order to pay less while buying promoted goods
```
michalmarcinkowski: In order to buy specific product with a discount

src/Sylius/Behat/Context/Setup/PromotionContext.php
```diff
@@ -11,6 +11,7 @@
 
 namespace Sylius\Behat\Context\Setup;
 
+use Behat\Behat\Tester\Exception\PendingException;
```
michalmarcinkowski: Redundant.

src/Sylius/Behat/Context/Setup/PromotionContext.php
```diff
@@ -564,6 +565,22 @@ public function thePromotionGivesOffIfOrderContainsProducts(PromotionInterface $
...
+    {
+        $this->createUnitFixedPromotion($promotion, $discount, $this->getProductsFilterConfiguration([$product->getCode()]));
+    }
+
+    /**
+     * @Given /^([^"]+) gives ("[^"]+%") off on a ("[^"]*" product)$/
+     */
+    public function itGivesPercentageDiscountOffOnAProduct(PromotionInterface $promotion, $discount, ProductInterface $product)
+    {
+        $this->createUnitPercentagePromotion($promotion, $discount, $this->getProductsFilterConfiguration([$product->getCode()]));
```
michalmarcinkowski: This promotion will be applied to every order, but not every order will be affected by it's action. We should only apply the promotion to orders that contain this product, so we should add a promotion rule `ContainsProduct` and probably modify the step to "it gives "20%" off on a "PHP T-Shirt" product **if order contains it**"

lchrusciel: But this is not an idea. We have `ContainsProduct` rule already, but this step should create a promotion that will be applied to every order, but only affect specified items.

lchrusciel: But thats a point of this test. This promotion should be assigned to every order, but not every item should be affected. The following scenario checks if filter works properly:
```yml
    Background:
        And the store has a product "PHP T-Shirt" priced at "$100.00"
        And the store has a product "PHP Mug" priced at "$20.00"
        And there is a promotion "T-Shirts promotion"
        And it gives "$10.00" off on a "PHP T-Shirt" product

    @ui
    Scenario: Receiving fixed discount on a single item
        When I add product "PHP T-Shirt" to the cart
        Then its price should be decreased by "$10.00"
        And my cart total should be "$90.00"

    @ui
    Scenario: Not receiving fixed discount on another item
        When I add product "PHP Mug" to the cart
        Then product "PHP Mug" price should not be decreased
        And my cart total should be "$20.00"
```

pjedrzejewski:
Merged in #6489.

pjedrzejewski:
Thank you Łukasz!

Issue #6517, opened 2016-10-22
[Shop] Order summary doesn't show any of the details


inssein:
In one of the recent commits, @pjedrzejewski removed `{% include '@SyliusShop/Common/Order/_summary.html.twig' %}` from the order summary page. Not sure of there was meant to be a different order summary or not.

PR #6518, opened 2016-10-22
[Shop] Re-add details to order summary

inssein:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/6517
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/Order/show.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/Order/show.html.twig
@@ -21,8 +21,9 @@
         </div>
     </h1>
 
-    {% set lastNewPayment = order.getLastNewPayment() %}
+    {% include '@SyliusShop/Common/Order/_summary.html.twig' %}
 
+    {% set lastNewPayment = order.getLastNewPayment() %}
     {% if order.paymentState not in ['paid', 'cancelled'] %}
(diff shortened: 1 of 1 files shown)


michalmarcinkowski:
@pjedrzejewski was it accidental or intentional removal?

pjedrzejewski:
Intentional, to make this page simpler and not duplicate the order summary. Although I am open to discuss if we feel that full order info should be included here. I left number, date and total.

inssein:
@pjedrzejewski as far as I am aware, this is the only "order summary" page for a user who is not registered.

What is currently there seems just like a page header, but maybe thats just me.

michalmarcinkowski:
I don't think we need to display whole order summary for guest customers. Customer receives an email with all order details, so here we can display only brief information like number, date, total, as well as dynamic data like payments and shipments together with its states. 

IMO we should not display sensitive data like addresses (this page is not so secure, anyone with link is able to access this page). Items are also not necessary (can't change and are in the email).

inssein:
@michalmarcinkowski thats fair. I should probably make the token much more unique as I currently need this functionality.

I have added it to my project for now, but can close this PR.

That said, I don't think we send any order details in any of the emails yet. But thats probably a separate issue / feature request.

PR #5734, opened 2016-08-08
[BC Break][Content] Remove Assetic and CMFMenuBundle

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets |
| License         | MIT

This is a completion of https://github.com/Sylius/Sylius/pull/5593

PR #6519, opened 2016-10-24
[Twig] Use the right `date` filter

lologhi:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

Because while running `assetic:dump` command I always have:

```
The template "SyliusResourceBundle::layout.html.twig" contains an
error: Unknown "format_date" filter. Did you mean "format_file" in
"SyliusResourceBundle::layout.html.twig" at line 50
```

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Customer/Show/_content.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Customer/Show/_content.html.twig
@@ -6,7 +6,7 @@
                     {{ customer.fullName|default('sylius.ui.guest_customer'|trans) }}
                 </a>
                 <div class="meta">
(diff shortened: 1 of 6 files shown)


Zales0123:
I think you should not even use ``assetic:dump``, as we dropped support for assetic in [this PR](https://github.com/Sylius/Sylius/pull/5734).

lchrusciel:
It is not a proper fix. `format_date` is a proper filter, but it is provided by SonataIntlBundle. If you receive such an error it means, that we have forgotten to add it to requirements. I will open PR with this solution.

lologhi:
Oh, great! Thx @lchrusciel ! I close this one.

pamil:
Why? As mentioned #6521, there is no reason to require `SonataIntlBundle` just for this one filter, that can be successfully replaced by `date`.

PR #6521, opened 2016-10-24
[Resource] Remove layout.html.twig

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | resolves #6519
| License         | MIT

Based on #6524


lchrusciel:
@pamil I found it more appropriate to add this requirement, because we are using this bundle in several places. Just for consistency. 

Anyway, is there any reason, why we are still keeping templates in ResourceBundle?

pamil:
@lchrusciel still, there's no need for downloading many more packages for someone using only ResourceBundle. I guess the whole `layout.html.twig` template can be deleted.

lchrusciel:
@pamil I can agree with that. Removing layout.html.twig will be the best solution.

pjedrzejewski:
Thank you Łukasz!

PR #6524, opened 2016-10-24
[Promotion] Add missing requirement

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

Without this fix packages check will fail. I'm surprised that it does not occur on current master. 
zrzut ekranu 2016-10-24 o 10 57 05


pjedrzejewski:
Thanks Łukasz!

PR #6530, opened 2016-10-24
[OrderBundle] Fix unpaid order query.

liverbool:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | ~
| License         | MIT

Without this we will get both `fulfilled` or `cancelled`


pjedrzejewski:
Thank you Ishmael! :)

PR #6538, opened 2016-10-25
[Documentation][Cookbook] How to customize Sylius Checkout?

TheMadeleine:
| Q | A |
| --- | --- |
| Doc fix? | no |
| New docs? | yes |
| BC breaks? | no |
| Related tickets | partially #5275 |
| License | MIT |

### DIFF
--- a/docs/cookbook/checkout.rst
+++ b/docs/cookbook/checkout.rst
@@ -184,6 +184,94 @@ and then the ``SyliusShopBundle/Resources/views/Checkout/SelectPayment/_form.htm
         </div>
     </div>
 
+Overwrite routing for Checkout
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Unfortunately there is no better way - you have to overwrite the whole routing for Checkout.
+To do that copy the content of ``ShopBundle/Resources/config/routing/checkout.yml``
+to the ``app/Resources/SyliusShopBundle/config/routing/checkout.yml`` file. **Remove routing** of ``sylius_shop_checkout_select_shipping``
+and change the **redirect route** in ``sylius_shop_checkout_address``. The rest should remain the same.
+
+.. code-block:: yaml
+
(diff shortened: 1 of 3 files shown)


docs/cookbook/checkout.rst
```diff
@@ -0,0 +1,292 @@
+How to customize Sylius Checkout?
+=================================
+
+Why would you override the Checkout process?
+--------------------------------------------
+
+This is a common problem for many Sylius users. Sometimes the checkout process we have designed is not suitable for your custom business needs.
+therefore you need to learn how to modify it, when you will need to for example:
```
Zales0123: ``Therefore``

Zales0123:
Pretty nice for me 👍  But as we've said on today's stand up - we should think about describing some more sophisticated example, as single page checkout, which is quite common use case.

docs/cookbook/checkout.rst
```diff
@@ -0,0 +1,292 @@
...
+                <i class="arrow right icon"></i>
+                {{ 'sylius.ui.next'|trans }}
+            </button>
+        </div>
+    </div>
+
+Overwrite routing for Checkout
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Unfortunately there is no better way - you have to overwrite the whole routing for Checkout.
```
TheMadeleine: @pjedrzejewski, @michalmarcinkowski wdyt?

michalmarcinkowski: After import we would also have `sylius_shop_checkout_select_shipping` route available and that's wrong. I would say that if we customize checkout it's better to copy the whole routing and have full control over it. Same as we do with the state machine.

michalmarcinkowski: It's even a Symfony best practice http://symfony.com/doc/current/bundles/override.html#routing

pjedrzejewski:
Thank you Magda! :)

PR #6541, opened 2016-10-25
[Addressing] Simplify AddressComparator

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes (static calls to non-static method)
| New feature?    | no
| BC breaks?      | no
| Related tickets | related to #6536 
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Addressing/Comparator/AddressComparator.php
+++ b/src/Sylius/Component/Addressing/Comparator/AddressComparator.php
@@ -23,26 +23,29 @@ final class AddressComparator implements AddressComparatorInterface
      */
     public function same(AddressInterface $firstAddress, AddressInterface $secondAddress)
     {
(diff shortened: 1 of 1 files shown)


liverbool:
👍 But my concern, When someone who extends `Address` he need to decorate this service too, Can we have magic and safe  logic?

I thinks about using `ReflectionClass::getMethods` since we need to compare all method except `id`, right?

pamil:
@liverbool that's right, if someone adds a method to `AddressInterface` then this service needs customisation too.

We should not compare neither all methods nor all properties of the object being passed as:
 - it may be an instance of a different class than the other object
 - it may have additional methods, that are not defined on interface, but does not change comparison logic
 - it may have additional properties, that does not change comparison logic

This is rather not a **magic and safe** logic :tada:

pjedrzejewski:
Thank you Kamil!

PR #6543, opened 2016-10-25
[Composer] Add SonataIntlBundle to requirements

### LINKED ISSUE #6525: [SonataIntlBundle] Is it needed?
This bundle is added to main composer file, but it isn't in bundles which use it. 

Nevertheless, `format_date` is used only 8 times(7 after #6521) in whole Sylius project. I haven't checked rest of its filters yet. Do we really need it? If so, we should add it to AttributeBundle(used forth times), AdminBundle(used twice) and ShopBundle(used once)

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #6525
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/composer.json
+++ b/src/Sylius/Bundle/AdminBundle/composer.json
@@ -24,7 +24,8 @@
 
         "sylius/ui-bundle": "^1.0",
         "sylius/core-bundle": "^1.0",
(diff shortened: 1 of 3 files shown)


michalmarcinkowski:
Thanks Łukasz! 👍

PR #6546, opened 2016-10-25
[Documentation][API] Fix API authorization doc

jjanvier:
Hello there :hand: 
It has been a while :)

This PR allows to copy/paste the API commands. Without the `\`, the next lines are not taken into consideration. It can lead to misunderstandings, as the command does not fail, even if the other arguments are not applied.

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | -
| License         | MIT


pjedrzejewski:
Thank you Julien! What a surprise! :)

PR #6553, opened 2016-10-26
[Review] Enhance review factory

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets |
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Review/Factory/ReviewFactoryInterface.php
+++ b/src/Sylius/Component/Review/Factory/ReviewFactoryInterface.php
@@ -12,6 +12,7 @@
 namespace Sylius\Component\Review\Factory;
 
 use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\Component\Review\Model\ReviewableInterface;
 use Sylius\Component\Review\Model\ReviewerInterface;
 use Sylius\Component\Review\Model\ReviewInterface;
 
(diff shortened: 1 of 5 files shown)


src/Sylius/Component/Review/Factory/ReviewFactoryInterface.php
```diff
@@ -21,17 +22,17 @@
 interface ReviewFactoryInterface extends FactoryInterface
 {
     /**
-     * @param mixed $subjectId
+     * @param ReviewableInterface $subject
      *
      * @return ReviewInterface
      */
-    public function createForSubject($subjectId);
+    public function createForSubject(ReviewableInterface $subject);
```
lchrusciel: One question. We have two methods here `createForSubject` and `createForSubjectWithReviewer`. But the second parameter in a second method is optional. Are both of these method needed? Can we remove one of them? Or remove optional requirement in a second one?

Zales0123: 👍  for removing optionality of ``$reviewer`` in the second method - if we're creating review with reviewer we should not allow to create it without reviewer 😄  🐃

michalmarcinkowski: I would do it with pleasure, but right now we use this method in our "magical" yml and when we are creating review as guest we will pass `null` instead of `Customer`, so it must stay like this until we figure out something better (not a concern of this PR).

pjedrzejewski:
Thank you Michał! :)

PR #6554, opened 2016-10-26
[UiBundle][UserBundle] #3448 prevent logged in users from visiting login

### LINKED ISSUE #3448: [UserBundle] Authenticated user can still access the login page
I know that this might be silly, but it's quite an annoying behaviour.

The scenario is:
- I logged in the backend (/administration/login)
- I'm redirected to the dashboard (/administration/dashboard)
- If I try to access the backend login URL in the browser address bar, I'm not being redirected to the dashboard page (which I think it's the page I should be redirected considering I'm already logged in)
- Instead of this, I can still access the login page.

I know that this is related to the firewall configuration. However, shouldn't a user be redirected to another page as the user is already logged in?

Checking this in the loginAction should do the work.

```
```

pix-art:
I've tried creating a listener that would check against your firewall settings, but you can no longer access these settings once your security context is created. This was my best bet after 3 hours of searching. Let me know what you guys think. I don't like that the routes are hard coded, which makes it less scalable.

| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | fixes #3448
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Controller/SecurityController.php
+++ b/src/Sylius/Bundle/UiBundle/Controller/SecurityController.php
@@ -11,10 +11,14 @@
 
 namespace Sylius\Bundle\UiBundle\Controller;
 
+use Symfony\Bundle\FrameworkBundle\Routing\Router;
 use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
 use Symfony\Component\Form\FormFactoryInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage;
+use Symfony\Component\Security\Core\Authorization\AuthorizationChecker;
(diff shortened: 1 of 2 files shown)


pix-art:
Tests are failing but i'm waiting for feedback to see if this would be a viable solution for the issue.

michalmarcinkowski:
This solution couples the `UiBundle` with Shop and Admin. If there is no better way to do it than in a controller then we should have two separate controllers for Shop and Admin with configurable redirect (argument passed be under `_sylius`).

pix-art:
That's probably a better solution, I wanted to create a listener that could parse our firewall and check if the current route was the login path and if you were logged in redirect you to the default_target_path.

```
firewalls:
        shop:
            form_login:
                login_path: sylius_admin_login
                default_target_path: sylius_admin_dashboard
```

But for some reason you can no longer get to your firewall settings once the security context has been build. 

You can just close this pull request then and lets go for a more sensible solution with the argument passed in the router like you suggested.

michalmarcinkowski:
Let's add the redirect logic to `Ui\SecurityController` and just specify the redirect route in a route configuration separately for shop and admin.

PR #6558, opened 2016-10-26
[Behat][DX] Use variadics wherever they can be used

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets |
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Setup/ProductContext.php
+++ b/src/Sylius/Behat/Context/Setup/ProductContext.php
@@ -379,27 +379,17 @@ final class ProductContext implements Context
         $option->setName($optionName);
         $option->setCode(strtoupper($optionName));
 
-        $firstOptionValue = $this->addProductOption($option, $values[0], 'POV1');
-        $secondOptionValue = $this->addProductOption($option, $values[1], 'POV2');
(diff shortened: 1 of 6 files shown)


src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -371,34 +371,30 @@ public function thisProductHasDateTimeAttributeWithDate(ProductInterface $produc
...
         /** @var ProductOptionInterface $option */
         $option = $this->productOptionFactory->createNew();
 
         $option->setName($optionName);
         $option->setCode(strtoupper($optionName));
 
-        $firstOptionValue = $this->addProductOption($option, $firstValue, 'POV1');
-        $secondOptionValue = $this->addProductOption($option, $secondValue, 'POV2');
+        $firstOptionValue = $this->addProductOption($option, $values[0], 'POV1');
+        $secondOptionValue = $this->addProductOption($option, $values[1], 'POV2');
```
lchrusciel: It is possible to use this step with only one value, so it is not a best solution. What with foreach with name generation? `POV` + index or something similar?

michalmarcinkowski:
Thanks Mateusz!

PR #6560, opened 2016-10-26
[API] Use codes instead of IDs in the REST API for simple objects

jjanvier:
Hello there,

I think it could really easier and huge DX bonus to be able to use the API via codes (or business identifiers) instead of MySQL or technical IDs.

(and by the way, it could really help me for what I'm currently doing :p)

This is just an example, I could do it really quick for all others models if you are interested.
## TODO:
- [ ] !!adjustment.yml
- [ ] channel.yml
- [ ] checkout.yml
- [x] country.yml
- [x] currency.yml
- [ ] customer.yml
- [x] locale.yml
- [ ] order.yml
- [ ] order_item.yml => no code or unique property, impossible right now
- [ ] payment.yml
- [ ] payment_method.yml
- [ ] product.yml
- [ ] product_attribute.yml
- [ ] product_option.yml
- [ ] product_variant.yml
- [ ] promotion.yml
- [ ] promotion_coupon.yml
- [x] province.yml
- [ ] setting.yml => no code or unique property, impossible right now
- [ ] shipment.yml => no code or unique property, impossible right now
- [ ] shipping_category.yml
- [x] shipping_method.yml
- [ ] tax_category.yml
- [x] tax_rate.yml
- [x] taxon.yml
- [ ] user.yml
- [x] zone.yml

| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | yes |
| Related tickets | - |
| License | MIT |


jjanvier:
it seems my build is not successful
how can I know which Behat is failing?

okwinza:
@jjanvier red build doesn't seem to be related to your changes. Try to reopen this PR to get it restarted.

michalmarcinkowski:
Thanks @jjanvier! That's one of the reasons for introducing codes in the first place 😉  We planned to do that, but unfortunately we have other higher priority tasks now, so we would be really happy if you could also apply this change to other models!

Regarding the build, looks like we exceeded the shared memory on travis in our JS suite, working on a fix to bring back the stability of the build. In the meantime I restarted your build, should help.

Btw. the usage of codes can be also applied to the relations (see [example](https://github.com/Sylius/Sylius/pull/5393/files))

jjanvier:
thank you @michalmarcinkowski for the feedback
ok, I'll do it for all the models

jjanvier:
@michalmarcinkowski @pjedrzejewski should I change also the [default routing](http://docs.sylius.org/en/latest/bundles/SyliusResourceBundle/routing.html#generating-generic-crud-routing) generation of the ResourceBundle? to replace {id} by {code}?

or maybe you want to keep the default to id to be compatible with all entities?

steffenbrem:
Codes are meant to be globally unique right? Would like to know if I got the same idea behind "codes" as you guys do.

michalmarcinkowski:
@jjanvier I think the default should be `id` as it's the most popular identifier and we shouldn't enforce using codes, however it will be nice to make it configurable, but that's something for a separate PR 😉 

@steffenbrem exactly, the `id` can vary between systems, but the `code` should allow to uniquely identify the entity in a multi-system environment.

michalmarcinkowski:
Great job @jjanvier! We would like to merge this and you if you have time to work on the rest, you can continue in a separate PR. Could you rebase? The build should be stable now, so if everything is fine it should become green 😉

jjanvier:
@michalmarcinkowski rebased. OK to merge this right now. I have difficulties with objects that are linked to others. Like Channel for instance. I'll try to explain it the next PR.

pjedrzejewski:
Thank you Julien! Not sure if it is good, but seems like there is no docs to update. :D

PR #6564, opened 2016-10-27
[Shop][Product] Associated products on product page

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/features/product/viewing_products/viewing_product_associations.feature
+++ b/features/product/viewing_products/viewing_product_associations.feature
@@ -1,8 +1,8 @@
 @viewing_products
 Feature: Viewing product's associations
-    In order to see related products
+    In order to quickly navigate to other products associated with the one I'm currently viewing
     As a Visitor
-    I want to be able to see product's associations
+    I want to see related products when viewing product details
 
     Background:
         Given the store operates on a single channel in "United States"
--- a/features/product/viewing_products/viewing_product_associations.feature
+++ b/features/product/viewing_products/viewing_product_associations.feature
@@ -1,0 +1,21 @@
+@viewing_products
+Feature: Viewing product's associations
+    In order to see related products
+    As a Visitor
+    I want to be able to see product's associations
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "LG G3"
+        And the store has "LG headphones" and "LG earphones" products
+        And the store has also "LG G4" and "LG G5" products
+        And the store has a product association type "Accessories"
+        And the store has also a product association type "Alternatives"
+        And the product "LG G3" has an association "Accessories" with products "LG headphones" and "LG earphones"
+        And the product "LG G3" has also an association "Alternatives" with products "LG G4" and "LG G5"
+
+    @ui @todo
+    Scenario: Viewing a detailed page with product's associations
+        When I view product "LG G3"
+        Then I should see the product association "Accessories" with products "LG headphones" and "LG earphones"
+        And I should also see the product association "Alternatives" with products "LG G4" and "LG G5"
--- a/features/product/viewing_products/viewing_product_associations.feature
+++ b/features/product/viewing_products/viewing_product_associations.feature
@@ -14,7 +14,7 @@ Feature: Viewing product's associations
         And the product "LG G3" has an association "Accessories" with products "LG headphones" and "LG earphones"
         And the product "LG G3" has also an association "Alternatives" with products "LG G4" and "LG G5"
 
-    @ui @todo
+    @ui
     Scenario: Viewing a detailed page with product's associations
         When I view product "LG G3"
         Then I should see the product association "Accessories" with products "LG headphones" and "LG earphones"
--- a/src/Sylius/Behat/Context/Setup/ProductAssociationContext.php
+++ b/src/Sylius/Behat/Context/Setup/ProductAssociationContext.php
@@ -20,7 +20,7 @@ use Sylius\Component\Resource\Repository\RepositoryInterface;
 /**
  * @author Grzegorz Sadowski <grzegorz.sadowski@lakion.com>
  */
-final class ProductAssociationTypeContext implements Context
+final class ProductAssociationContext implements Context
 {
     /**
      * @var SharedStorageInterface
--- a/src/Sylius/Behat/Context/Setup/ProductAssociationContext.php
+++ b/src/Sylius/Behat/Context/Setup/ProductAssociationContext.php
@@ -14,6 +14,8 @@ namespace Sylius\Behat\Context\Setup;
 use Behat\Behat\Context\Context;
 use Sylius\Behat\Service\SharedStorageInterface;
 use Sylius\Component\Association\Model\AssociationTypeInterface;
+use Sylius\Component\Core\Model\ProductInterface;
+use Sylius\Component\Product\Model\ProductAssociationInterface;
 use Sylius\Component\Resource\Factory\FactoryInterface;
 use Sylius\Component\Resource\Repository\RepositoryInterface;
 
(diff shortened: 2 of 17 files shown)


src/Sylius/Behat/Context/Setup/ProductAssociationContext.php
```diff
@@ -81,6 +111,29 @@ private function createProductAssociationType($name, $code = null)
...
+     * @param ProductInterface $product
+     * @param AssociationTypeInterface $productAssociationType
+     * @param ProductInterface $firstProduct
+     * @param ProductInterface|null $secondProduct
+     */
+    private function createProductAssociation(
+        ProductInterface $product,
+        AssociationTypeInterface $productAssociationType,
+        ProductInterface $firstProduct,
+        ProductInterface $secondProduct = null
```
lchrusciel: Instead of nullable value as a second parameter we can use variadics and then use a foreach.

src/Sylius/Behat/Context/Setup/ProductAssociationContext.php
```diff
@@ -81,6 +111,29 @@ private function createProductAssociationType($name, $code = null)
...
+    ) {
+        /** @var ProductAssociationInterface $productAssociation */
+        $productAssociation = $this->productAssociationFactory->createNew();
+        $productAssociation->setType($productAssociationType);
+        $productAssociation->addAssociatedObject($firstProduct);
+        $productAssociation->addAssociatedObject($secondProduct);
```
lchrusciel: Can we add a null value?

src/Sylius/Behat/Context/Transform/ProductContext.php
```diff
@@ -38,6 +38,8 @@ public function __construct(ProductRepositoryInterface $productRepository)
      * @Transform /^"([^"]+)" product(?:|s)$/
      * @Transform /^(?:a|an) "([^"]+)"$/
      * @Transform :product
+     * @Transform :firstProduct
```
lchrusciel: Is it needed?

src/Sylius/Behat/Context/Ui/Shop/ProductContext.php
```diff
@@ -427,4 +429,39 @@ public function iShouldSeeAsItsAverageRating($rating)
...
+
+        $this->assertIsProductIsInAssociation($firstProductName, $productAssociationName);
+        $this->assertIsProductIsInAssociation($secondProductName, $productAssociationName);
+    }
+
+    /**
+     * @param string $productName
+     * @param string $productAssociationName
+     *
+     * @throws /InvalidArgumentException
```
lchrusciel: / -> \

src/Sylius/Behat/Context/Ui/Shop/ProductContext.php
```diff
@@ -12,6 +12,8 @@
 namespace Sylius\Behat\Context\Ui\Shop;
 
 use Behat\Behat\Context\Context;
+use Behat\Behat\Tester\Exception\PendingException;
+use Buzz\Exception\InvalidArgumentException;
```
lchrusciel: It should be removed

src/Sylius/Behat/Context/Ui/Shop/ProductContext.php
```diff
@@ -12,6 +12,8 @@
 namespace Sylius\Behat\Context\Ui\Shop;
 
 use Behat\Behat\Context\Context;
+use Behat\Behat\Tester\Exception\PendingException;
```
lchrusciel: Not needed

src/Sylius/Behat/Context/Ui/Shop/ProductContext.php
Zales0123: What kind of abomination is that? :O 🗡  💀

pamil: 🛬  vs 🛫

features/product/viewing_products/viewing_product_associations.feature
```diff
@@ -0,0 +1,21 @@
+@viewing_products
+Feature: Viewing product's associations
+    In order to see related products
+    As a Visitor
+    I want to be able to see product's associations
```
michalmarcinkowski:
```
In order to quickly navigate to other products associated with the one I'm currently viewing
As a Visitor
I want to see related products when viewing product details
```

src/Sylius/Behat/Page/Shop/Product/ShowPage.php
```diff
@@ -226,9 +226,29 @@ public function isMainImageDisplayed()
...
+        return null !== $this->getElement('association', ['%association-name%' => $productAssociationName]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function hasProductInAssociation($productName, $productAssociationName)
+    {
+        $associationHeader = $this->getElement('association', ['%association-name%' => $productAssociationName]);
+        $associations = $associationHeader->getParent()->find('css', '.four');
```
michalmarcinkowski: `Assert::notNull($associations)` to prevent from calling `->find(` on `null` object

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
```diff
@@ -170,6 +170,22 @@ public function findOneBySlug($slug)
     /**
      * {@inheritdoc}
      */
+    public function findByAssociationId($associationId)
```
michalmarcinkowski: Is it used?

src/Sylius/Bundle/ShopBundle/Resources/views/Product/Show/_associations.html.twig
```diff
@@ -1,2 +1,9 @@
-<h4 class="ui horizontal section divider header">{{ 'sylius.ui.you_may_also_like'|trans }}</h4>
-{{ render(url('sylius_shop_partial_product_index_latest', {'count': 4, 'template': '@SyliusShop/Product/_simpleList.html.twig'})) }}
+{% for association in product.associations %}
+    <div>
+        <h4 class="ui horizontal section divider header">{{ association.type.name }}</h4>
+        {{ render(url('sylius_shop_partial_associated_products', {'id': association.id, 'template': '@SyliusShop/Product/_associations.html.twig'})) }}
```
michalmarcinkowski: Why do we need to call controller? We have all associations already loaded in the template, we can simply replace this line with `{% include '@SyliusShop/Product/_simpleList.html.twig' with {'products': association.associatedObjects} %}`, can't we?

gabiudrescu: I'd also go for a partial for this kind of components of a product page, because I'd use HTTP cache to avoid making the product page heavy to generate.

this is something we do for example with menus that are generated dinamically at bestvalue.eu - we rely on a subrequest that is cached for 24h.

src/Sylius/Behat/Page/Shop/Product/ShowPage.php
```diff
@@ -226,9 +227,32 @@ public function isMainImageDisplayed()
...
+        return null !== $this->getElement('association', ['%association-name%' => $productAssociationName]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function hasProductInAssociation($productName, $productAssociationName)
+    {
+        $associationHeader = $this->getElement('association', ['%association-name%' => $productAssociationName]);
+        Assert::notNull($associationHeader);
```
michalmarcinkowski: `$this->getElement` should never return `null`

src/Sylius/Behat/Page/Shop/Product/ShowPage.php
```diff
@@ -226,9 +227,32 @@ public function isMainImageDisplayed()
     /**
      * {@inheritdoc}
      */
+    public function hasAssociation($productAssociationName)
+    {
+        return null !== $this->getElement('association', ['%association-name%' => $productAssociationName]);
```
michalmarcinkowski: `$this->getElement` will throw exception when element will not be found. We should try-catch to return boolean instead of exception.

michalmarcinkowski:
Good job Grzesiu!

PR #6589, opened 2016-10-29
[ResourceController] Allow state machine transition from createAction

winzou:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | n/a
| License         | MIT

We have state machine transition possible from `updateAction` but not from `createAction`. Can be pretty useful depending on your model.


pjedrzejewski:
Thank you Alexandre! :)

PR #6593, opened 2016-10-31
[Behat][ProductReview] Recalculate average rating after review rate change

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Related tickets | |
| License         | MIT

### DIFF
--- a/features/product/managing_product_reviews/recalculating_product_average_rating.feature
+++ b/features/product/managing_product_reviews/recalculating_product_average_rating.feature
@@ -1,0 +1,20 @@
+@managing_product_reviews
+Feature: Recalculating product average rating
+    In order to have my product's average rating properly calculated
+    As an Administrator
+    I want to have product's average rating recalculated after review rate change
+    
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "Lamborghini Gallardo Model"
+        And this product has a review titled "Awesome" and rated 4 with a comment "Nice product" added by customer "ross@teammike.com"
+        And this product has a review titled "Not bad" and rated 3 with a comment "Not bad car" added by customer "specter@teamharvey.com"
+        And I am logged in as an administrator
+
+    @ui @javascript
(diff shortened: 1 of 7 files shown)


src/Sylius/Behat/Context/Ui/Admin/ManagingProductReviewsContext.php
```diff
@@ -149,9 +149,9 @@ public function iChooseAsItsRating($rating)
     }
 
     /**
-     * @Then I should see the product :productName
+     * @Then I should see the product with name :productName
```
michalmarcinkowski: I think this change is redundant. It was natural as well as grammatically correct, so I don't see any benefit of that change (only disadvantage, more bytes used 😆 ).

features/product/managing_product_reviews/recalculating_product_average_rating.feature
```diff
@@ -0,0 +1,20 @@
...
+        And this product has a review titled "Not bad" and rated 3 with a comment "Not bad car" added by customer "specter@teamharvey.com"
+        And I am logged in as an administrator
+
+    @ui @javascript
+    Scenario: Product's average rating is correctly recalculated after review's rate change
+        When I want to modify the "Awesome" product review
+        And I choose 5 as its rating
+        And I save my changes
+        Then I should be notified that it has been successfully edited
+        And this product average rating should be 4
```
michalmarcinkowski: Let's be specific here and write `And average rating of product "Lamborghini Gallardo Model" should be 4`

src/Sylius/Behat/Context/Ui/Admin/ManagingProductReviewsContext.php
michalmarcinkowski: `Then the review should be for product :productName` or something similar?

michalmarcinkowski:
Thanks Mateusz!

PR #6275, opened 2016-10-04
[Payum] Use converted amount for generic capture

inssein:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | partially https://github.com/Sylius/Sylius/issues/6258
| License         | MIT

PR #6600, opened 2016-11-01
[Critical] Payum uses int for amount, not float

### LINKED ISSUE #6599: Payum > Omnipay > Paypal Express, price divided by 100
Hi

I'm on the latest master.
When using the Omnipay Bridge for Express Checkout, the checkout works perfectly **except** that the price transmitted to Paypal is divided by 100 !

When I look in the payment table, for an omnipay request I see : 

`{"amount":2.05,"currency":"EUR","description":"Payment contains 1 items for a total of 205.00" ... }`

and for Paypal Express checkout directly through Payum : 

`{"PAYMENTREQUEST_0_INVNUM":"151-154","PAYMENTREQUEST_0_CURRENCYCODE":"EUR","PAYMENTREQUEST_0_AMT":"205.00" ... }`

So this is quite critical. I don't know if it's the Payum bundle, or the Omnipay bridge, but well.

tchapi:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | kinda
| Related tickets | fixes #6599 
| License         | MIT

I don't even know how this could work before. **Payum** uses ints internally to represent total amounts (see https://github.com/Payum/Payum/issues/597#issuecomment-255673769), and the **Payum\Core\Model\Payment** class explicitly declares `$totalAmount` as an int so

I hope I'm not overlooking something important here, but to me this looks like something that should be fixed.

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/Action/CapturePaymentAction.php
+++ b/src/Sylius/Bundle/PayumBundle/Action/CapturePaymentAction.php
@@ -101,10 +101,10 @@ final class CapturePaymentAction extends GatewayAwareAction
      * @param int $price
      * @param string $currencyCode
      *
-     * @return float
+     * @return int
      */
     private function convertPrice($price, $currencyCode)
     {
-        return round($this->currencyConverter->convertFromBase($price, $currencyCode) / 100, 2);
+        return $this->currencyConverter->convertFromBase($price, $currencyCode);
     }
 }
--- a/src/Sylius/Bundle/PayumBundle/Action/CapturePaymentAction.php
+++ b/src/Sylius/Bundle/PayumBundle/Action/CapturePaymentAction.php
@@ -67,7 +67,8 @@ final class CapturePaymentAction extends GatewayAwareAction
                 $payumPayment->setClientId($order->getCustomer()->getId());
                 $payumPayment->setDescription(sprintf(
                     'Payment contains %d items for a total of %01.2f',
-                    $order->getItems()->count(), $totalAmount
+                    $order->getItems()->count(),
+                    $totalAmount
                 ));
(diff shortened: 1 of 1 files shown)


michalmarcinkowski:
This was introduced by https://github.com/Sylius/Sylius/pull/6275, which fixed multi currency converts, but unfortunately also changed the amounts to float. 

@tchapi thanks for reporting and fixing! Please correct also the amount in the payment description.

tchapi:
The payment description text is correct, no need to change it

pjedrzejewski:
Thank you @tchapi!

src/Sylius/Bundle/PayumBundle/Action/CapturePaymentAction.php
```diff
@@ -67,7 +67,8 @@ public function execute($request)
                 $payumPayment->setClientId($order->getCustomer()->getId());
                 $payumPayment->setDescription(sprintf(
                     'Payment contains %d items for a total of %01.2f',
-                    $order->getItems()->count(), $totalAmount
+                    $order->getItems()->count(),
+                    $totalAmount
```
michalmarcinkowski: Should be `round($totalAmount / 100, 2)`, right now for amount $12.00 it will display `... for a total of 1200.00` because the total is integer.

inssein:
Strange - I must have my currency converted implemented incorrectly. Sorry about that. I will be double checking this when I get home since I have been accepting money with the previous code =(

gorkalaucirica:
@pjedrzejewski Can we have alpha2 tagged containing this bug fix?, its critical as it is breaking a payment gateway we are using in a project that we are releasing in two weeks. 

We can rely on an specific commit but I guess other people may face the same issues we are facing with this.

Thanks!

PR #6603, opened 2016-11-01
Deleted unused private methods in OrderController

crevillo:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| License         | MIT

These two private methods are not called from any other method in that class, so i guess they can be removed.


michalmarcinkowski:
Thanks Carlos!

PR #6628, opened 2016-11-03
[Promotion] Remove weird "decorating" stuff from PromotionExtension

### LINKED ISSUE #6602: [Promotion] Unable to overwrite Action factory
We've upgraded `0.17` to `0.19`, and now aren't able to manually add promotion actions anymore. After diving into the code I've spotted the problem, decorating the factory is wrong. The related error is:

> Warning: Missing argument 1 for Sylius\Component\Resource\Factory\Factory::__construct(), called in /www/vendor/sylius/resource/Factory/Factory.php on line 39 and defined

The code that is responsible for the error can be found in the [`SyliusPromotionExtension`](https://github.com/Sylius/Sylius/commit/6757589a6d05ecc0441ee172d0de0f33f43401a3#diff-cb1b8612a570d72d1050f43e4dffeb49L81).

What is happening, is that the `Factory::$className` actually is the original `Factory` class itself. So when [`createNew`](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Resource/Factory/Factory.php#L39) is executed, it doesn't create a `sylius.model.promotion_action.class` but a new version of the decorated `Factory`.

I think the overwriteActionFactory should be changed to the following code, but before filing a PR I'd like to get your feedback first.

```php
    /**
     * @param ContainerBuilder $container
     */
    private function overwriteActionFactory(ContainerBuilder $container)
    {
        $promotionActionFactoryClass = $container->getParameter('sylius.factory.promotion_action.class');
        $promotionActionModelClass = $container->getParameter('sylius.model.promotion_action.class');

        $decoratedPromotionActionFactoryDefinition = new Definition($promotionActionFactoryClass, [$promotionActionModelClass]);
        $container->setDefinition('sylius.factory.promotion_action', $decoratedPromotionActionFactoryDefinition);
    }
```

Is this fixed in `v1.0.0-alpha`? Otherwise we'll upgrade to that version instead of filing a PR.

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | fixes #6602
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius/sylius_promotion.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius/sylius_promotion.yml
@@ -21,6 +21,3 @@ sylius_promotion:
                 repository: Sylius\Bundle\PromotionBundle\Doctrine\ORM\PromotionCouponRepository
                 form:
                     default: Sylius\Bundle\CoreBundle\Form\Type\Promotion\PromotionCouponType
-        promotion_action:
-            classes:
-                factory: Sylius\Component\Core\Factory\PromotionActionFactory
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services/promotion.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services/promotion.xml
@@ -13,6 +13,11 @@
 
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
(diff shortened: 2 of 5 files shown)


isometriks:
Wanted to bring this up again if talking about factories: https://github.com/Sylius/Sylius/issues/4226

I had added a comment on the end there about some kind of default factory that just makes the object (like the decorated service you're using). If you needed to change this factory more you'd have to decorate the decorator because changing the config will just change that inner one, and probably break it so I feel like that makes the config misleading. I'm not really sure what the best solution is, if we do that then you have to make sure the service is named perfectly which I guess isn't a huge deal.

pjedrzejewski:
Thanks Kamil!

@isometriks I think we should remove the config for factory class indeed, it is not very useful and only complicates stuff.

rvanlaak:
@pjedrzejewski with these configs you mean the separate functions in the extensions that override the container factory service?

PR #6629, opened 2016-11-04
[BUG] Added Correct Currency Code to Checkout Payments

bretto36:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

When viewing the checkout complete page, it shows the correct value but the wrong currency code.


pjedrzejewski:
I think the bug goes a bit deeper, because this currency should be updated when order's currency has changed.

michalmarcinkowski:
Already fixed, the payment will always have correct currency.

PR #6633, opened 2016-11-04
[Github] Fix default PR description template

Zales0123:
There were problems with ``|`` signs in few rows.

Previous:
zrzut ekranu 2016-11-04 o 10 09 49

Now:
zrzut ekranu 2016-11-04 o 10 45 53

### DIFF
--- a/PULL_REQUEST_TEMPLATE.md
+++ b/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,7 @@
 | Q               | A
 | --------------- | ---
-| Bug fix?        | no|yes
-| New feature?    | no|yes
-| BC breaks?      | no|yes
+| Bug fix?        | no/yes
+| New feature?    | no/yes
+| BC breaks?      | no/yes
 | Related tickets | fixes #X, partially #Y, mentioned in #Z
 | License         | MIT


Zales0123:
👍

pamil:
We can remove `|` at the end of the lines now :tada:

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

Zales0123:
Ups 😄

PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,7 +1,7 @@
 | Q               | A
 | --------------- | ---
-| Bug fix?        | no|yes
-| New feature?    | no|yes
-| BC breaks?      | no|yes
+| Bug fix?        | no\|yes
+| New feature?    | no\|yes
+| BC breaks?      | no\|yes
```
michalmarcinkowski: How about `| BC breaks?      | no/yes`? 😄

pjedrzejewski: @Zales0123 Please adjust the template and it is good to go!

Zales0123: I've done this way, as @pamil suggested previously, to keep this ``|`` signs rather than replace them into something else. So I don't know now, should it be ``|`` or ``/`` 😄

pjedrzejewski: But why? I want to choose "yes" or "no" and that's it, so using "/" is one less backspace.

Zales0123: Sure, but the point was not to change how PR table looks like, but only to fix it :) I have no problem with slash and could change it in a minute ;)
zrzut ekranu 2016-11-07 o 09 38 12

michalmarcinkowski:
👍

PR #6665, opened 2016-11-08
[BC Break] Remove getLastShipment method from Order

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/4638
| License         | MIT

Based on https://github.com/Sylius/Sylius/pull/6662

### DIFF
--- a/src/Sylius/Component/Core/spec/Taxation/Applicator/OrderShipmentTaxesApplicatorSpec.php
+++ b/src/Sylius/Component/Core/spec/Taxation/Applicator/OrderShipmentTaxesApplicatorSpec.php
@@ -11,7 +11,7 @@
 
 namespace spec\Sylius\Component\Core\Taxation\Applicator;
 
-use Doctrine\Common\Collections\Collection;
+use Doctrine\Common\Collections\ArrayCollection;
 use PhpSpec\ObjectBehavior;
 use Prophecy\Argument;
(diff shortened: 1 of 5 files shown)


src/Sylius/Component/Core/spec/Taxation/Applicator/OrderShipmentTaxesApplicatorSpec.php
```diff
@@ -113,49 +105,45 @@ function it_does_nothing_if_there_are_no_shipment_taxes_on_order(
         $this->apply($order, $zone);
     }
 
-    function it_does_nothing_if_order_has_no_shipment(
-        TaxRateResolverInterface $taxRateResolver,
+    function it_throws_exception_if_order_has_no_shipment_but_shipment_total_is_greater_than_0(
```
pamil: Is it even possible?

michalmarcinkowski:
@Zales0123 the problem is that we have a model designed for multiple shipments, but in few places we assume only a single shipment for simplicity because in default implementation we don't do split shipments. There is no concept of default shipping method on an order since we should always perform actions on each shipping separately, but this complicates logic in few places. We shouldn't create a service for that because this will encourage people to using it 🦀 .

pjedrzejewski:
Thanks Michał!

PR #6666, opened 2016-11-08
[Product][Shop] Buy product in correct prices for channel and currency

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | |
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Setup/CurrencyContext.php
+++ b/src/Sylius/Behat/Context/Setup/CurrencyContext.php
@@ -137,7 +137,8 @@ final class CurrencyContext implements Context
         }
 
         foreach ($currenciesCodes as $currencyCode) {
-            $channel->addCurrency($this->provideCurrency($currencyCode));
+            $currency = $this->provideCurrency($currencyCode);
+            $channel->addCurrency($currency);
         }
 
         $this->channelManager->flush();
--- a/src/Sylius/Behat/Context/Setup/CurrencyContext.php
+++ b/src/Sylius/Behat/Context/Setup/CurrencyContext.php
@@ -137,8 +137,7 @@ final class CurrencyContext implements Context
         }
 
         foreach ($currenciesCodes as $currencyCode) {
-            $currency = $this->provideCurrency($currencyCode);
-            $channel->addCurrency($currency);
+            $channel->addCurrency($this->provideCurrency($currencyCode));
         }
 
         $this->channelManager->flush();
--- a/src/Sylius/Behat/Context/Setup/CurrencyContext.php
+++ b/src/Sylius/Behat/Context/Setup/CurrencyContext.php
@@ -119,7 +119,7 @@ final class CurrencyContext implements Context
      */
     public function theStoreHasCurrencyWithExchangeRate($currencyCode, $exchangeRate)
     {
-        $currency = $this->createCurrency($currencyCode, $exchangeRate);
+        $currency = $this->createCurrency($currencyCode, (float) $exchangeRate);
         $currency->setEnabled(true);
 
         $this->saveCurrency($currency);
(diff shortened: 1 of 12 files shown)


Zales0123:
I've opened this PR accidentally (I didn't want to open it yet) - but then I've just realized I can't lose such a beautiful PR number as #6666 😄

Zales0123:
@pjedrzejewski @michalmarcinkowski do we want also to implement displaying product price based on pricing calculator on product's details page? Or is it the case for the separate PR?

src/Sylius/Behat/Context/Setup/CurrencyContext.php
```diff
@@ -150,9 +150,6 @@ public function thatChannelAllowsToShopUsingAndCurrencies(ChannelInterface $chan
     public function itUsesTheCurrencyByDefault(ChannelInterface $channel, $currencyCode)
     {
         $currency = $this->provideCurrency($currencyCode);
-        $currency->setExchangeRate(1.0);
```
michalmarcinkowski: Why removing this?

src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -540,6 +558,33 @@ public function thisProductHasAnImageWithACode(ProductInterface $product, $image
     }
 
     /**
+     * @Given /^(it) has different prices for different channels and currencies$/
+     */
+    public function itHasDifferentPricesForDifferentChannelsAndCurrencies(ProductInterface $product)
+    {
+        $product->getFirstVariant()->setPricingCalculator(Calculators::CHANNEL_AND_CURRENCY_BASED);
```
michalmarcinkowski: `getFirstVariant` method should be removed. `defaultVariantResolver` should be used instead.

src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -540,6 +558,33 @@ public function thisProductHasAnImageWithACode(ProductInterface $product, $image
...
+    /**
+     * @Given /^(it) has price ("[^"]+") for ("[^"]+" channel) and "([^"]+)" currency$/
+     */
+    public function itHasPriceForChannelAndCurrency(
+        ProductInterface $product,
+        $price,
+        ChannelInterface $channel,
+        $currency
+    ) {
+        $variant = $product->getFirstVariant();
```
michalmarcinkowski: Same as above.

src/Sylius/Behat/Resources/config/services/contexts/setup.xml
```diff
@@ -20,6 +20,7 @@
             <argument type="service" id="__symfony__.sylius.factory.channel" />
             <argument type="service" id="__symfony__.sylius.repository.channel" />
             <argument type="service" id="__symfony__.sylius.manager.channel" />
+            <argument>%__behat__.mink.parameters%</argument>
```
michalmarcinkowski: Do we need this?

Zales0123: No 👍  😄

src/Sylius/Behat/Context/Setup/CurrencyContext.php
Zales0123: This step is ``that channel uses the currency by default`` -> it should not set this currency's exchange rate to *1.0*, as exchange rate is not configured per channel, but for whole system. In my case, it caused a mistake when one channel has base currency ``EUR`` and the other one ``GBP``.

Zales0123:
@pamil Not anymore 🐫

src/Sylius/Behat/Context/Setup/CurrencyContext.php
michalmarcinkowski: Ok, so `it` means channel, not store. Totally valid fix, good catch! 👍

michalmarcinkowski:
Thank you Mateusz! Please apply all comments in a separate PR.

Issue #3836, opened 2016-01-08
[Resource / Translation] Sorting by translatable field while listing resources


pamil:
### Expected behaviour

``` gherkin
Feature: Sorting products by name

Background:
    Given there is a product with Polish name "Cebula" and English name "Onion"
    And there is another product with Polish name "Marchew" and English name "Carrot"

Scenario: Sorting products by Polish names
    Given I browse the site using Polish locale
    When I list all products sorted by name
    Then "Cebula" should be before "Marchew"

Scenario: Sorting products by English names
    Given I browse the site using English locale
    When I list all products sorted by name
    Then "Carrot" should be before "Onion"
```
### Actual behaviour

When we're listing a list of any translatable resource that is sorted by translatable field, we cannot be sure in which order they show up (probably depends on the database implementation).

For example, the query for a list of products looks like (in pseudo-DQL): `SELECT DISTINCT p FROM products p LEFT JOIN p.translations t ORDER BY t.name`. If product has more translations that one, the sorting is broken as it may use different locales translations to make the sorting.

Doctrine ORM in 2.5 version provides [`LimitSubqueryWalker::validate()`](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Tools/Pagination/LimitSubqueryWalker.php#L128) that does not allow us to make those kind of Schrodinger-queries and therefore Sylius fails to support it for now.

nakashu:
Is the way how this should be fixed already discussed somewhere ?
If we could have a proposed solution here, someone could make a PR sooner for this.

peteward:
hmmm yeah that's a tricky issue :/

It's Doctrine breaking a perfectly valid query by adding distinct on it for pagination which then makes it technically invalid... 

Is it possible to call one of the other walkers like `WhereInWalker` or `CountWalker`?

nakashu:
Stumbled upon this bug again. Couldn't we just inject or set the `localProvider` to `ORM\translatableResourceRepository`? 

The same way it is done in `Sylius\Bundle\TranslationBundle\Doctrine\ODM\MongoDB\TranslatableResourceRepository`

```php
  /**
     * {@inheritdoc}
     */
    public function setLocaleProvider(LocaleProviderInterface $localeProvider)
    {
        $this->localeProvider = $localeProvider;

        return $this;
    }
    // ...
/**
     * @param string $name
     *
     * @return string
     */
    protected function getPropertyName($name)
    {
[...]

Issue #3775, opened 2015-12-26
@pamil, why did you default DoctrineORMOutputWalker to true?


peteward:
In your recent merged PR, in the Entity Repository:

```
-        return new Pagerfanta(new DoctrineORMAdapter($queryBuilder, true, false));
+        return new Pagerfanta(new DoctrineORMAdapter($queryBuilder, true, true));
```

This reverts a very explicit one-line change I merged a while ago in #3495.

Pagination on a million row table now back up to 60 seconds page load time.

I'm guessing you made this change for some reason, what was it? Is there a use-case where we do want OutputWalkers? I think the default should be false and only set to true where you really need it.

pjedrzejewski:
Ouch, I totally forgot about this when merging the PR. I believe this was changed to the sake of Doctrine ORM 2.5 compatiblity. Right, @pamil?

@peteward For such large collections of resources, I was thinking about using ElasticSearch. With Grids implemented, we will implement ElasticSearch driver for Grid, so you would be able to use it for customers/products backend etc.

peteward:
@pjedrzejewski, that's an interesting idea. I've been toying with the idea of pushing more data to Elasticsearch but it starts to snowball with the content you would end up pushing into it.

For us so far Customers is the main set of data we've migrated from our existing MySQL DB into our Sylius one, but we'll soon have the following all in the millions:

* Orders
* OrderItems
* OrderItemUnits
* Payments
* Shipments
* OrderReturns

Does it really make sense to duplicate all of this kind of data (and some of it gets very large - our current live DB is 15GB) into an alternative storage mechanism that must be synchronised? You're adding overhead of storage costs and an maintaining consistency between them. If you search for an entity through elasticsearch to then change it through MySQL, the changes really need to be reflected immediately when you get back to the grid. FosElastica can do this, but when there are complex relations involved it gets challenging (as we found with our project). I'm guessing you're talking about an alternative engine to MySQL? not introducing it as a requirement?

Anyway, this is slightly away from the point here. 

I can't see anything in DoctrineORM 2.5 that suggests this should be an issue, the only PR related supposedly fixes something that could have been an issue, not introducing one: https://github.com/doctrine/doctrine2/issues/3648

I think paginated queries against any size of dataset shouldn't be run with querywalkers active here, it's just that we can see the results very obviously in our project.

This isn't killing us because we've already worked around this call, but I spotted it whilst upgrading today and it looks wrong.

[...]

PR #6674, opened 2016-11-09
[RFC] Bring back doctrine 2.5 support

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes (kind of)
| New feature?    | no
| BC breaks?      | yes
| Related tickets | unfixes #3836 (in fact, it is not really fixed right now too)
| License         | MIT

I have spent some time in order to find the best solution and bring back doctrine 2.5 support and this is my proposal. 

The second parameter of DoctrineORMAdapter defines if `fetchJoinCollection` variable. By default it is value is equal to true. According to [doctrine documentation](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/pagination.html) it should behave like this:
```
Paginating Doctrine queries is not as simple as you might think in the beginning. If you have complex fetch-join scenarios with one-to-many or many-to-many associations using the “default” LIMIT functionality of database vendors is not sufficient to get the correct results.

By default the pagination extension does the following steps to compute the correct result:

 * Perform a Count query using DISTINCT keyword.
 * Perform a Limit Subquery with DISTINCT to find all ids of the entity in from on the current page.
 * Perform a WHERE IN query to get all results for the current page.

This behavior is only necessary if you actually fetch join a to-many collection. You can disable this behavior by setting the $fetchJoinCollection flag to false; in that case only 2 instead of the 3 queries described are executed. We hope to automate the detection for this in the future.
```

As far as we are using only a default locale to list product in backend, fetch join is not required. But using this solution on a shop side, where we should add a fallback translation, for the resources without current one, we will end up with `Schrodinger-queries` as was mentioned in #3836. :cat: 

Another possible solution is to use DoctrineORMOutputWalker, but as was written in #3775, it has a major impact on performance. Maybe using both of this solution would something considerable(this for admin and output walkers in shop), but it won't as trivial as it looks like.

Some further readings:
 - [Doctrine pagination](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/tutorials/pagination.html)
 - [Probable problem with proposed solution](http://stackoverflow.com/questions/5620771/limiting-a-doctrine-query-with-a-fetch-joined-collection)
 - [Doctrine join explanation](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/dql-doctrine-query-language.html#joins)

 [Expected problem with output walker](https://github.com/doctrine/doctrine2/issues/5622).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "doctrine/doctrine-cache-bundle": "~1.0",
         "doctrine/doctrine-fixtures-bundle": "~2.2",
         "doctrine/doctrine-migrations-bundle": "~1.0",
-        "doctrine/orm": "^2.4.8,<2.5",
+        "doctrine/orm": "^2.4.8",
         "friendsofsymfony/oauth-server-bundle": "~1.4",
         "friendsofsymfony/rest-bundle": "~1.5",
         "fzaninotto/faker": "~1.2",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,8 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "7c910c932db8a42205661c43c4d89584",
-    "content-hash": "615c38d1a2a7b742dc97a6ca4595b41d",
+    "hash": "34547adb12ab53b69858dcca0cfbe8ef",
+    "content-hash": "ddada40493e432d2a74c917f0f238b36",
     "packages": [
         {
             "name": "behat/transliterator",
@@ -1038,28 +1038,31 @@
         },
         {
             "name": "doctrine/orm",
-            "version": "v2.4.8",
+            "version": "v2.5.5",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/doctrine2.git",
-                "reference": "5aedac1e5c5caaeac14798822c70325dc242d467"
+                "reference": "73e4be7c7b3ba26f96b781a40b33feba4dfa6d45"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/doctrine2/zipball/5aedac1e5c5caaeac14798822c70325dc242d467",
-                "reference": "5aedac1e5c5caaeac14798822c70325dc242d467",
+                "url": "https://api.github.com/repos/doctrine/doctrine2/zipball/73e4be7c7b3ba26f96b781a40b33feba4dfa6d45",
+                "reference": "73e4be7c7b3ba26f96b781a40b33feba4dfa6d45",
                 "shasum": ""
             },
             "require": {
-                "doctrine/collections": "~1.1",
-                "doctrine/dbal": "~2.4",
+                "doctrine/cache": "~1.4",
+                "doctrine/collections": "~1.2",
+                "doctrine/common": ">=2.5-dev,<2.7-dev",
+                "doctrine/dbal": ">=2.5-dev,<2.6-dev",
+                "doctrine/instantiator": "~1.0.1",
                 "ext-pdo": "*",
-                "php": ">=5.3.2",
-                "symfony/console": "~2.0"
+                "php": ">=5.4",
+                "symfony/console": "~2.5|~3.0"
             },
             "require-dev": {
-                "satooshi/php-coveralls": "dev-master",
-                "symfony/yaml": "~2.1"
+                "phpunit/phpunit": "~4.0",
+                "symfony/yaml": "~2.3|~3.0"
             },
             "suggest": {
                 "symfony/yaml": "If you want to use YAML Metadata Mapping Driver"
@@ -1071,7 +1074,7 @@
             "type": "library",
             "extra": {
                 "branch-alias": {
-                    "dev-master": "2.4.x-dev"
+                    "dev-master": "2.6.x-dev"
                 }
             },
             "autoload": {
@@ -1107,7 +1110,7 @@
                 "database",
                 "orm"
             ],
-            "time": "2015-08-31 13:19:01"
+            "time": "2016-09-10 18:51:13"
         },
         {
             "name": "friendsofsymfony/oauth-server-bundle",
@@ -2352,7 +2355,7 @@
                     "email": "stof@notk.org"
                 },
                 {
-                    "name": "Knplabs",
+                    "name": "KnpLabs",
                     "homepage": "http://knplabs.com"
                 },
                 {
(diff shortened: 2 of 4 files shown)


pamil:
Rebase needed 🌮

pamil:
Rebase needed once again :taco:

lchrusciel:
@pamil done

pjedrzejewski:
Thank you Łukasz!

Aniade:
Hello @lchrusciel. 
I am wondering why you set fetch join collection to false for pagination in this [commit](https://github.com/Sylius/Sylius/pull/6674/commits/36afcd67c1f77ef4a11b5fd87beabd4d33b69ae2) ? I am writing a custom query builder for an index page and my pagination is broken because of joined collections.
Thank you :)

LaetitiaRiffaud:
Hello @lchrusciel. 
I am wondering why you set fetch join collection to false for pagination in this [commit](https://github.com/Sylius/Sylius/pull/6674/commits/36afcd67c1f77ef4a11b5fd87beabd4d33b69ae2) ? I am writing a custom query builder for an index page and my pagination is broken because of joined collections.
Thank you :)

lchrusciel:
Hello @LaetitiaRiffaud,
I have tried to be as precise as possible when describing this solution. Turning off fetch joins allowed us to build a proper query to the database. Otherwise, we had not enough results on index pages. Turning off the fetch joins allows us to use lazy loading for other translations if needed, but at the same time use properly `setMaxResults` functions. When we were using fetch joins, it was possible to have only 5 results on the index page with limit 10, if each of item had 2 translations.

peteward:
Hi @lchrusciel,

I agree with @LaetitiaRiffaud here. 

We've just upgraded to the latest ResourceBundle and this broke our pagination on various pages with 1-to-many join queries.

The result that we see is incomplete lists. The example in front of me is listing Orders and showing the OrderItems in them.  e.g. If the pagination is set to 10, the first order has 3 items, all other orders have only a single item, the list will only show 7 orders.

You can see this from the crude queries that doctrine creates as a result, doing a simple `COUNT` and then a `LIMIT, OFFSET` combination which just won't work for any 1-to-n selects: 

```
SELECT 
  count(DISTINCT r0_.id) AS sclr_0 
FROM 
  reiss_order r0_ 
  INNER JOIN sylius_order_item s2_ ON r0_.id = s2_.order_id 
  INNER JOIN reiss_product_variant r4_ ON s2_.variant_id = r4_.id 
  INNER JOIN reiss_product r5_ ON r4_.product_id = r5_.id 
WHERE 
  r0_.completed_at >= ?
```
```
SELECT *
FROM 
  reiss_order r0_ 
  INNER JOIN sylius_order_item s2_ ON r0_.id = s2_.order_id 
  INNER JOIN reiss_product_variant r4_ ON s2_.variant_id = r4_.id 
  INNER JOIN reiss_product r5_ ON r4_.product_id = r5_.id 
WHERE 
  r0_.completed_at >= ? 
ORDER BY 
  r0_.completed_at DESC, 
LIMIT 
  10 OFFSET 0
```

Doctrine is going to select 10 database rows and if 3 of those rows correspond to a single order you're only going to get 7 entities hydrated (potentially some of them get split between pages too which is really wrong).   

What does work is setting the fetch join collection to true. This creates the following 3 queries COUNT, DISTINCT, then explicit SELECT by ids (as described in the docs):

```
SELECT 
  count(DISTINCT r0_.id) AS sclr_0 
FROM 
  reiss_order r0_ 
  INNER JOIN sylius_order_item s2_ ON r0_.id = s2_.order_id 
  INNER JOIN reiss_product_variant r4_ ON s2_.variant_id = r4_.id 
  INNER JOIN reiss_product r5_ ON r4_.product_id = r5_.id 
WHERE 
  r0_.completed_at >= ?
```

```
SELECT 
  DISTINCT r0_.id AS id_0, 
  r0_.completed_at AS completed_at_1, 
  r0_.updated_at AS updated_at_2 
FROM 
  reiss_order r0_ 
  INNER JOIN sylius_order_item s2_ ON r0_.id = s2_.order_id 
  INNER JOIN reiss_product_variant r4_ ON s2_.variant_id = r4_.id 
  INNER JOIN reiss_product r5_ ON r4_.product_id = r5_.id 
WHERE 
  r0_.completed_at >= ? 
ORDER BY 
  r0_.completed_at DESC, 
  r0_.updated_at DESC 
LIMIT 
  10 OFFSET 0
```

```
SELECT *
FROM 
  reiss_order r0_ 
  INNER JOIN sylius_order_item s2_ ON r0_.id = s2_.order_id 
  INNER JOIN reiss_product_variant r4_ ON s2_.variant_id = r4_.id 
  INNER JOIN reiss_product r5_ ON r4_.product_id = r5_.id 
WHERE 
  r0_.completed_at >= ? 
  AND r0_.id IN (['5380988', '5380987', '5380983', '5380986', '5380985', '5380984', '5380963', '5380939', '5380891', '5380894', '5380825', '5380807', '5380786', '5380798', '5380750', '5380735', '5380726', '5380705', '5380648', '5380633']) 
ORDER BY 
  r0_.completed_at DESC, 
  r0_.updated_at DESC, 
  s2_.id ASC
```

I've read through this PR a couple of times and it's not clear to me which situation you were trying to solve. I remember that the problem is complicated and there might not be a perfect solution.

However, this change has broken pagination lists that involve selecting data from multiple tables with a 1-to-n resultset. Whilst this is not the most efficient way to write queries (we try to use `PARTIAL`s where necessary to mitigate this), it is probably still more efficient than hammering the database with lots of lazy-load single selects. As soon as you start using lazy-load in large lists of data you're asking for trouble in the number of queries called and can create strain on your DB.

We're going to revert this change in our project, but it would be interesting to know what the thinking was and if there's another approach that could be used.

peteward:
ping

lchrusciel:
Hello @peteward,

It was a while since I have made this change. But AFAIR, the problem was same as you described it but in a matter of locales. If resources had 2 translation on paginated pages we had only half of expected items. After this change, I have checked a web profiler but execution time and a number of queries actually have been reduced. Still, we are testing Sylius usually on a small amount of data, but @pamil has made some performance tests with Gatlling and it was fine. 

I'm open for any other solution, we can also just revert this change and see what will break(if anything). 
I have also discussed this problem briefly with Ocramius, and his recommendation was to create a custom paginator.

peteward:
Hi @lchrusciel,

Reviewing the issues behind this, it does appear to come back to this one: https://github.com/Sylius/Sylius/issues/3836. I think this change inadvertently works around this issue but I don't think it's the correct solution and it has the adverse affect of breaking what I believe to be a more common use-case.

I'm not really sure how this solves the above issue though. It will run 1 less query, but it gets the wrong result with selected joins.

I think this issue needs to be revisited and consider various use-cases.

lchrusciel:
@peteward I have reverted this change in #8038 PR and we can see what will be affected.

I don't have time to dig into it atm, but maybe somebody will be able to tackle it.

peteward:
Thanks @lchrusciel.

Hopefully this is the last we'll hear of this but I suspect not! :)

PR #3185, opened 2015-08-27
[Resource] Fix sorting bugs

Zales0123:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2892, #3004, #3027 
| License       | MIT
| Doc PR        |

Even by setting false, Paginator didn't use output walkers, because of comparing set value to ``null``.

PR #3495, opened 2015-10-27
[Resource] Fix DoctrineORMPaginator instantiation to set OutputWalker to false 

peteward:
I know that the ResourceBundle refactor is going to add this, but it's killing us at the moment with a 60 second page load and with no ETA on that refactor this is a quick fix.

PR #6676, opened 2016-11-09
[RFC] [Doctrine] Default output walkers turn off

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Related tickets | |
| License         | MIT
 
During my work with #6674 I have noticed, that we slightly might missed a proper turn off. The default option for third parameter of constructor [is null](https://github.com/whiteoctober/Pagerfanta/blob/master/src/Pagerfanta/Adapter/DoctrineORMAdapter.php#L35). And the place where it changes the behaviour of doctrine paginator is [here](https://github.com/doctrine/doctrine2/blob/master/lib/Doctrine/ORM/Tools/Pagination/Paginator.php#L210:L212).

It has been fixed previously by #3185.


lchrusciel:
Sorry, didn't notice #3495. It should be false

PR #6677, opened 2016-11-09
[Github] Yet another pull request template fix

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Related tickets | 
| License         | MIT

Without this change by default we would end up with the same result as above. It is similar to first @Zales0123 solution from #6633. Now: 

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |


michalmarcinkowski:
👍

PR #6689, opened 2016-11-10
[AdminBundle] Create checkbox based on form instead of attribute

sweoggy:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

By checking the form type instead of the attribute type it is possible to use the CheckboxAttributeType of Sylius as form type parent and thereby get its looks in the product's attributes view. This is more in line with how using form type parents work in Symfony.


pjedrzejewski:
Thank you Oscar!

PR #6691, opened 2016-11-10
[Core] Use cache_dir parameter to ensure writable cache dir in InstallCommand

venyii:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes? |
| License         | MIT |

This fixes a potential problem if the `cache_dir` is not configured to be the default `app/cache` (e.g. Vagrantfile).

I'm not sure about the BC of this in commands, the `APP_CACHE` constant was exposed.

Unrelated question: [Here](http://docs.sylius.org/en/latest/contributing/code/patches.html#make-a-pull-request) it's described that each BC Break PR must contain an entry in the CHANGELOG/UPGRADE file. I've checked the last BC breaking PRs and none of them had any changes like that. Is this still relevant?


pjedrzejewski:
Thanks @venyii! Unfortunately, due to quite big changes we have decided to use UPGRADE file since beta release (this month).

PR #6692, opened 2016-11-10
[Core] Use dbname from doctrine connection to allow custom values

venyii:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |

This change makes it possible to use the same database name in every environment. Currently, the `InstallDatabaseCommand` will always prepend `_%env%` to the database name, even if I specified `sylius` (no `_%env%`) in my dev config, which won't match the expectations made here.

It will return `NULL` for SQLite and eventually explode since `listDatabases` is not supported anyway.


pjedrzejewski:
Thank you @venyii! 👍

PR #6697, opened 2016-11-11
[Core] Ensure writable web/ directories with absolute paths in the installer

venyii:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |

This fixes a problem when the install command is not called from the Symfony root directory. Since the web paths are relative, the directories will be created wherever I call the installer from.

Additionally, I removed a `no-interaction` check since Symfony's `askConfirmation` method will do this automatically for us in the next step.


pjedrzejewski:
Thank you @venyii!

PR #6700, opened 2016-11-13
[WIP] [RFC] Semantic ui calendar

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes (because of changes in promotion form type) |
| Related tickets |  |
| License         | MIT |

I have recently notice, that semantic ui calendar has been publish as a [third party library](https://github.com/mdehoog/Semantic-UI-Calendar). I have needed it during some work with promotion so I tried to implement it to Sylius. It looks much better, but probably it will be hard to make it translatable. I'm the really last person to touch js, so if anybody is willing to finish it feel free to take my work. 

With semantic calendar:
with semantic calendar

Without semantic calendar:
without semantic calendar

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/private/js/semantic-ui-calendar.js
+++ b/src/Sylius/Bundle/AdminBundle/Resources/private/js/semantic-ui-calendar.js
@@ -1,0 +1,11 @@
+(function($) {
+    $(document).ready(function() {
+        $('.calendar').calendar();
+        $('#rangestart').calendar({
+            endCalendar: $('#rangeend')
+        });
+        $('#rangeend').calendar({
+            startCalendar: $('#rangestart')
+        });
+    });
+})(jQuery);
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Promotion/_form.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Promotion/_form.html.twig
@@ -23,8 +23,8 @@
         <h4 class="ui dividing header">{{ 'sylius.ui.start_date'|trans }} & {{ 'sylius.ui.end_date'|trans }}</h4>
 
         <div class="two fields">
-            {{ form_row(form.startsAt) }}
-            {{ form_row(form.endsAt) }}
+            {{ form_row(form.startsAt, { 'id': 'rangestart'}) }}
+            {{ form_row(form.endsAt, { 'id': 'rangeend'}) }}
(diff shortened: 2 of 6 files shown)


src/Sylius/Bundle/AdminBundle/Resources/views/Promotion/_form.html.twig
```diff
@@ -23,8 +23,8 @@
         <h4 class="ui dividing header">{{ 'sylius.ui.start_date'|trans }} & {{ 'sylius.ui.end_date'|trans }}</h4>
 
         <div class="two fields">
-            {{ form_row(form.startsAt) }}
-            {{ form_row(form.endsAt) }}
+            {{ form_row(form.startsAt, { 'id': 'rangestart'}) }}
+            {{ form_row(form.endsAt, { 'id': 'rangeend'}) }}
```
pjedrzejewski: We should definitely avoid overriding ids of form fields.

src/Sylius/Bundle/AdminBundle/Resources/private/js/semantic-ui-calendar.js
```diff
@@ -0,0 +1,11 @@
+(function($) {
+    $(document).ready(function() {
+        $('.calendar').calendar();
+        $('#rangestart').calendar({
```
pjedrzejewski: Perhaps we can add it to all "date" form types?

src/Sylius/Bundle/AdminBundle/Resources/views/Promotion/_form.html.twig
lchrusciel: The default calendar is rendered without anything. The case of this `id` is to define with which other calendar it should be coupled to present a range of dates. I didn't have better idea for this.

src/Sylius/Bundle/AdminBundle/Resources/private/js/semantic-ui-calendar.js
lchrusciel: I have overridden a default datetime templates to add this calendar to them. I thought about something like `full calendar`, `date calendar` and `time calendar`(I know how it sounds like) classes to determine this UI

pjedrzejewski: Makes sense to me.

PR #6514, opened 2016-10-22
[Admin] Implement string blocks management

pjedrzejewski:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Related tickets | - |
| License | MIT |
- Scenarios for admin
- Admin CRUD
- Fixtures
- Added PHPCR Initializer

PR #5776, opened 2016-08-12
implement slideshow block for new AdminBundle

videni:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | Yes |
| BC breaks? | no |
| License | MIT |

Todo:
- [x] When saving SlideshowBlock and ImagineBlock , automatically set the SlideshowBlock as the parent of ImagineBlock
- [x] Hide parentDocument form field for cmf block
- [x]  Only display filters whose data_loader is cmf_media_doctrine_phpcr for ImagineBlockType
- [ ] Correct Behat feature to test image upload for slideshow block

PR #6717, opened 2016-11-14
Remove ContentBundle

### LINKED ISSUE #316: [TaxonomyBundle] PHPCR Support + Dynamic Taxonomies?
Hi,

One of the things missing from the Symfony CMF project now is Taxonomies, and this bundle looks very similar to the one I had in mind.

It looks like PHPCR support would be easy to add, but I am not sure how you would handle the dependencies efficiently - i.e. we wouldn't want to pull in `doctrine-extensions`.

Also I imagined supporting flat / dynamic taxonomies, e.g. handling "tagging" in a blog post scenario and adding form types etc.

So, I am wondering:
- Would it make sense to support dynamic / flat taxonomies, e.g. handling "tags" in the context of a blog?
- How would you go about managing dependencies for other other persistenece backends in this bundle?

/cc @dbu

### LINKED ISSUE #3178: Locale support in URL for static content
I have found that currently there are no possibility to specify locale in URL. It is important for Search Engine Crawler to index page in different languages.

There are also Google recommendations to properly implement locale support in different way to make crawlable multilingual content:
- [Multi-regional and multilingual sites](https://support.google.com/webmasters/answer/182192?hl=en#1)
  More preferable approaches are:
- Domain-based like _de.example.com_ or _example.ie_
- or URL subdirectory like _example.com/de/_

There is already some Symfony support from the box:
- Fully-supported by the routing system the special [_locale parameter](http://symfony.com/doc/current/book/translation.html#the-locale-and-the-url])

After investigation possible solutions I have few bundles with some pros and cons.
May be most popular is [JMSI18nRoutingBundle](https://github.com/schmittjoh/JMSI18nRoutingBundle).
_Pros:_
- it includes ready solution to prefix static routes with /_locale/ without hardcoding it into routing configuration

_Cons:_
- bundle seems to be not supported very well

Other possible bundle for integration may be [LuneticsLocaleBundle](https://github.com/lunetics/LocaleBundle/).
_Pros:_
- supported by community(latest commit 20 days ago)
- suggested in SymfonyCMF [documentation](http://symfony.com/doc/current/cmf/book/handling_multilang.html#initial-language-choice-lunetics-localebundle)
- provide similar functionality that already implemented in Sylius LocaleBundle

_Cons:_
- do not have solution to prefix static routes

Both bundles do not solve issue with DynamicRouter that routes must be prefixed.

My thoughts about implementation is following:
- Integrate LuneticsLocaleBundle(but it is optional and I guess everything may be achieved with Sylius LocaleBundle)
- Create possibility to prefix static and dynamic routes in Sylius side.

### LINKED ISSUE #3629: [Routing] Sub routes (e.x. /about-us/extra-route)
Currently you can only have routes at the top level, sub routes are not supported. So you could never create a content page and make the URL `/about-us/extra-route` route to it. Is there currently something going on on how we could solve this?

The issue lies with that the PHPCR `Route` document works with parents and children. Should we adjust the form type so you can link a route to an existing route as its parent?

### LINKED ISSUE #3810: [RFC] Assign channels to pages, routes and blocks.
Like products; Pages, routes and blocks should also be able to have channels assigned to them. Now, every channel will have the exact same content structure. I am currently implementing this in my own project. Is this something Sylius wants in it's core too?

I now have an ORM -> PHPCR binding between documents and channels, they are stored as a `long` multivalue in PHPCR:
Screenshot

The Symfony CMF route provider needs to be altered to support this, it needs to filter the routes based on if the current channel id exists for the route document (this last part I am still working on).

What are your thoughts on this @pjedrzejewski ?

### LINKED ISSUE #4139: [RFC] i18n for your own static content
I'm curious about how others handle multi locale Sylius instances in terms of static content. For example, pages [like this](http://demo.sylius.org/administration/content/static-contents//cms/pages/terms-of-service/edit) should be translated in all supported locales.

Do you currently create two versions of this page or rely on some user land implementation?

cc @peteward

### LINKED ISSUE #4399: PHPCR per channel
Hi,
I want to make blocks available per channel.
I think that phpcr_node table must have channel_id column and after this to filter in repository by channel if it is given in criteria.
1. It is good my logic?
2. How can I add channel_id to  phpcr_node?I did't know what entity to extend to implement this logic.
3. What repository should be extended to add filter by channel?

### LINKED ISSUE #4808: [RFC] Proposal for translatable PHPCR documents (pages, blocks etc.)
So we have a lot of translatable models right now, but some very important ones are missing, such as pages and menu nodes. I am currently trying to find a good strategy to handle this. Since we have the `TranslatableInterface` in the resource component, I was thinking about setting up translations in PHPCR this way. You also have multi-lang build in into the PHPCR ODM of Doctrine, but it has some serious drawbacks, such as not being able to load multiple translations of the same document in the same runtime (I think you can work around it by cloning etc, but still, it feels very different from how Sylius handles translations now).

What do you guys think about not using the translations mechanism Doctrine provides for us and simply have a separate document to store translatable fields, just as we do with products etc. We could for example store the translation documents in `/pages/about-us/en`, `/pages/about-us/nl`.

Being able to do the following would be awesome:

``` php
$page = new Page();
$page->setName('about-us');

$translation = new PageTranslation();
$translation->setLocale('en');
$translation->setTitle('About us');
$translation->setBody('Content in English');
$page->addTranslation($translation);

$translation = new PageTranslation();
$translation->setLocale('nl');
$translation->setTitle('Over ons');
$translation->setBody('Content in het Nederlands');
$page->addTranslation($translation);

$dm->persist($page);
$dm->flush();

$page->getTranslations(); // would return "PageTranslation" documents
```

Would love to hear any ideas or feedback on this!

I will probably have this finished for myself on Monday, if this approach is preferred, I will be glad to open source it and provide a PR with the implementation for this.

### LINKED ISSUE #6559: Issue with mapping between routes and static content, also routes are not generated : admin panel(admin side)
issue 1/ when i add a static content and i create a route with that static content  , route does'nt work (no route found exception)
issue 2/when a delete a static content its route does'nt delete too wich cause a (Impossible to access an attribute ("title") on a null variable in @SyliusAdmin\Route\Grid\Field\content.html.twig at line 1.)

### LINKED ISSUE #6622: [documentation] the "content" page is empty
this page is empty : 
http://docs.sylius.org/en/latest/book/content.html

how do you plan to support page création 'CMS-ish'

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | **Related to PHPCR usage in Sylius:** closes #6622, closes #6559, closes #6514, closes #5776, closes #4808, closes #4399, closes #4139, closes #3810, closes #3629, closes #3178, closes #316 |
| License         | MIT |

PHPCR usage blocks us from supporting Symfony 3 - ContentBundle will be rewritten as a stand-alone bundle built on top of Sylius as soon as possible.


Zales0123:
So many issue closed 👍  🐃  💥

gabiudrescu:
RIP ContentBundle. and now his watch has ended. 

it will be missed - as we (heavily) rely on it to power almost every static content on our website.

pamil:
I hope to get it revived after we figure out how to solve issues with PHPCR-related packages releases supporting Symfony 3.

pjedrzejewski:
@gabiudrescu It will be back and even better. :) But we need to bump to Symfony 3.0. Also, content bundle will become a "Sylius Bundle", so you will simply drop it in the kernel together with deps, import 1 config file, 1 routing file and the same functionality we have right now will appear in your admin. :)

Another reason for removing it from the core is that more and more projects use Sylius with eZ, so they do not need the whole SymfonyCMF + Sonata + admin UI.

pjedrzejewski:
Thanks Kamil! :)

gabiudrescu:
@pjedrzejewski that's good news. (Y)

PR #6721, opened 2016-11-15
[ShopBundle] Fix redirect when referer is blank

inssein:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | none |
| License         | MIT |

Currently the currency switch action falls back to `$request->getBaseUrl()`, but unfortunately that is also blank for me. I switched it to use `$request->getSchemeAndHttpHost()`.

By the way, why are controllers final? I was just going to override this controller and change the one action, but can't.


michalmarcinkowski:
Thanks Hussein!

PR #6735, opened 2016-11-15
[Twig] Add missing currency variable

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |


pjedrzejewski:
Thanks Łukasz! I think we should also pass the exchange rate, although we will discuss this whole thing tomorrow anyway. :)

PR #6743, opened 2016-11-16
Latest translatable issue

### LINKED ISSUE #3513: Product translations weird behavior
There is the "Newest products" section in leftsidebar which contains 5 latest products. But I found out that the amount of displayed products varies based on how many translations are filled. If you have 3 languages and you filled in the latest product's title and description in all 3 languages, you will see only 2 products in "Newest products" section.
It looks like it counts every product's translation like one product but doesn't display it. It displays only correct translation. 
So it consumes 3 "places" in sidebar for the latest product and 2 left places for the 2nd latest product. In some combinations of translation counts for last products, you can see only product image without any description or title in "Newest products" section - it loads product (image and everything non-translatable) but not the correct translation.
I found it in my test implementation and also recreated it today on demo.sylius.org.

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #3513 |
| License         | MIT |

The issue #3513 does not exist at the moment. This scenarios would protect us from regression. 

The base problem of an issue is to add select with translation. With the current [function](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php#L74-L86) we do not add any select. If we would add any to-many association, then this query would be broken. Of course, we can consider to have translation fetched from database directly in the query instead of lazy loaded, but before we won't measure its impact on performance it is just a blind shot.

Anyway, if we decide to add some any to-many association there are at least two solutions:
 * Build two separated queries as it is explained [here](http://stackoverflow.com/questions/8554119/set-limit-with-doctrine-2)
 * Make this to-many association to-one, i.e. we can add some where constraint 
 
**Edit**: Last commit fixes admin dashboard with little overhead of number of queries, but not with the time. But to confirm that I would have to do better profiling, what is not a case at the moment.
before: 
before

after:
after

FYI: xdebug was turned on

### DIFF
--- a/src/Sylius/Behat/Context/Setup/ProductContext.php
+++ b/src/Sylius/Behat/Context/Setup/ProductContext.php
@@ -224,6 +224,7 @@ final class ProductContext implements Context
 
     /**
      * @Given /^(this product) is named "([^"]+)" (in the "([^"]+)" locale)$/
+     * @Given /^the (product "[^"]+") is named "([^"]+)" (in the "([^"]+)" locale)$/
      */
     public function thisProductIsNamedIn(ProductInterface $product, $name, $locale)
     {
@@ -269,6 +270,19 @@ final class ProductContext implements Context
     {
         foreach ($productsNames as $productName) {
             $this->saveProduct($this->createProduct($productName));
+        }
+    }
+
+    /**
+     * @Given /^(this channel) has "([^"]+)", "([^"]+)", "([^"]+)" and "([^"]+)" products$/
+     */
+    public function thisChannelHasProducts(ChannelInterface $channel, ...$productsNames)
+    {
+        foreach ($productsNames as $productName) {
+            $product = $this->createProduct($productName);
+            $product->addChannel($channel);
+
+            $this->saveProduct($product);
         }
     }
 
--- a/src/Sylius/Behat/Context/Ui/Shop/HomepageContext.php
+++ b/src/Sylius/Behat/Context/Ui/Shop/HomepageContext.php
@@ -1,0 +1,57 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Behat\Context\Ui\Shop;
+
+use Behat\Behat\Context\Context;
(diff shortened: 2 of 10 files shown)


src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -657,6 +650,10 @@ private function createProduct($productName, $price = 0, $date = null)
         $productVariant->setPrice($price);
         $productVariant->setCode($product->getCode());
 
+        if ($this->sharedStorage->has('channel')) {
+            $product->addChannel($this->sharedStorage->get('channel'));
+        }
```
michalmarcinkowski: Why you moved assigning channels here? It's not a concern of this step. The previous place was correct.

src/Sylius/Behat/Page/Shop/HomePage.php
```diff
@@ -105,6 +105,16 @@ public function switchLocale($localeCode)
     /**
      * {@inheritdoc}
      */
+    public function countProductsItems()
+    {
+        $products = $this->getDocument()->findAll('css','.sylius-product-name');
+
+        return count($products);
```
michalmarcinkowski: `count` can return `false`, we should handle this case.

pamil: `findAll` returns `NodeElement[]` according to its definition, so we shouldn't be bothered by that method returning uncountable value.

pamil: Side note: `getLatestProductsNames()` won't be difficult to implement, but will allow for more flexibility in the future.

src/Sylius/Behat/Context/Ui/Shop/HomepageContext.php
```diff
@@ -0,0 +1,57 @@
...
+     * @Then I should see :numberOfProducts products in the list
+     */
+    public function iShouldSeeProductsInTheList($numberOfProducts)
+    {
+        $foundRows = $this->homePage->countProductsItems();
+
+        Assert::same(
+            (int) $numberOfProducts,
+            $foundRows,
+            '%s rows with products should appear on page, %s rows has been found'
```
pamil: `%s` -> `%d`, they are not strings

pjedrzejewski:
Thanks Łukasz!

Issue #6457, opened 2016-10-19
Better naming for taxon-related rule checkers


pamil:
- `TaxonRuleChecker` allows us to check if we have at least one product from at least one given taxon
- `ContainsTaxonRuleChecker` allows us to check if we have N products from the given taxon

Their names don't say enough I think 🐫

gabiudrescu:
these rules created lots of discussions in our company, too.

TaxonRuleChecker with the default translation "Taxon" makes me think that cart should have all products from the taxons, not at least one from at least one given taxon.

I believe the behavior from TaxonRuleChecker could be achieved by using ContainsTaxonRuleChecker with value 1 which should translate in terms of UI like this:
 - rule "Contains number of items from taxon"
 - first option: "Taxons: Books, Stickers"
 - second option: "1 item"

to make it more semantic, I'd switch the form configuration to make it look like this:
![image](https://cloud.githubusercontent.com/assets/5156054/19660289/cfdf7260-9a37-11e6-95d6-1d3e57836bad.png)

----

if you still want to keep TaxonRuleChecker as a separate rule, I'd go for the following semantics:
 - className: ContainsFromMultipleTaxonsRuleChecker
 - label: Contains at least one from taxons
![image](https://cloud.githubusercontent.com/assets/5156054/19660377/2fa5b858-9a38-11e6-9db4-3d99e5af2fe8.png)

----

Anyway, I still have this question in my mind: why don't you modify ContainsTaxonRuleChecker to accommodate the behavior of multiple select for taxons from TaxonRuleChecker? 

What's the downside?

Or point the user to add multiple "ContainsTaxonRuleChecker" to achieve the same behavior from "TaxonRuleChecker" ?

PR #6744, opened 2016-11-16
[Core][Promotion] Unify taxon rule checkers

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/5355, https://github.com/Sylius/Sylius/issues/6457 |
| License         | MIT |

For now, we decided to keep only one, simpler rule, which checks is there any product from one of picked taxons. The other one (with count) could be easily implemented if needed. I'm also not 100% about new rule checker name, so I'd be glad to see any feedback about that.


features/promotion/complex_promotions.feature
```diff
@@ -26,15 +26,6 @@ Feature: Receiving a discount based on a configured promotion
         And my discount should be "-$10.00"
 
     @ui
-    Scenario: Receiving a discount on products from a specific taxon if an order contains products from an another taxon
```
michalmarcinkowski: That scenario should work with `HasTaxonRule`, we shouldn't to remove it :)

src/Sylius/Bundle/CoreBundle/Resources/translations/messages.en.yml
```diff
@@ -120,9 +120,6 @@ sylius:
         product_association_type:
             name: Name
         promotion_rule:
-            contains_taxon:
-                taxon: Taxon
-                count: Count
             taxon:
```
tuka217: shouldn't it be now `has_taxon`?

src/Sylius/Behat/Context/Ui/Admin/ManagingPromotionsContext.php
```diff
@@ -143,22 +143,12 @@ public function iAddIt()
     }
 
     /**
-     * @Given I add the "Contains number of items from taxon" rule configured with :count :taxonName
+     * @Given I add the "Has at least one from taxons" rule configured with :firstTaxon
+     * @Given I add the "Has at least one from taxons" rule configured with :firstTaxon and :secondTaxon
```
GSadee: Shouldn't these steps start from `@When` ?

michalmarcinkowski:
Thanks Mateusz, please apply all comments in a separate PR.

PR #6755, opened 2016-11-17
[Promotion] Fix product and taxon filter

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

### DIFF
--- a/src/Sylius/Component/Core/Promotion/Filter/ProductFilter.php
+++ b/src/Sylius/Component/Core/Promotion/Filter/ProductFilter.php
@@ -21,7 +21,7 @@ final class ProductFilter implements FilterInterface
      */
     public function filter(array $items, array $configuration)
     {
-        if (!isset($configuration['filters']['products_filter'])) {
+        if (empty($configuration['filters']['products_filter']['products'])) {
             return $items;
         }
 
--- a/src/Sylius/Component/Core/spec/Promotion/Filter/ProductFilterSpec.php
+++ b/src/Sylius/Component/Core/spec/Promotion/Filter/ProductFilterSpec.php
@@ -53,4 +53,9 @@ final class ProductFilterSpec extends ObjectBehavior
     {
         $this->filter([$item], [])->shouldReturn([$item]);
     }
+
+    function it_returns_all_items_if_configuration_is_empty(OrderItemInterface $item)
+    {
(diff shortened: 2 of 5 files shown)


src/Sylius/Component/Core/Promotion/Filter/ProductFilter.php
```diff
@@ -21,7 +21,7 @@
      */
     public function filter(array $items, array $configuration)
     {
-        if (!isset($configuration['filters']['products_filter'])) {
+        if (!isset($configuration['filters']['products_filter']) || empty($configuration['filters']['products_filter']['products'])) {
```
pamil: `!isset($configuration['filters']['products_filter']['products'])` would be enough

src/Sylius/Component/Core/spec/Promotion/Filter/ProductFilterSpec.php
```diff
@@ -53,4 +53,9 @@ function it_returns_all_items_if_configuration_is_invalid(OrderItemInterface $it
     {
         $this->filter([$item], [])->shouldReturn([$item]);
     }
+
+    function it_returns_all_items_if_configuration_is_empty(OrderItemInterface $item)
+    {
+        $this->filter([$item], ['filters' => ['products_filter' => []]])->shouldReturn([$item]);
```
GSadee: More apprioprate would be `['filters' => ['products_filter' => ['products' => []]]]`

src/Sylius/Component/Core/spec/Promotion/Filter/ProductFilterSpec.php
```diff
@@ -53,4 +53,9 @@ function it_returns_all_items_if_configuration_is_invalid(OrderItemInterface $it
     {
         $this->filter([$item], [])->shouldReturn([$item]);
     }
+
+    function it_returns_all_items_if_configuration_is_empty(OrderItemInterface $item)
```
pamil: I wonder whether it is a desired behaviour. There should be no filter at all first if we want to pass through all the items.

lchrusciel: The problem is, that is it the default behaviour of of promotion configuration. The configuration is present, but it is empty, because filter form type is included in an action form type. We need it with a current solution.

michalmarcinkowski:
👍

PR #6752, opened 2016-11-17
[Taxation] Remove useless argument type property

venyii:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

This fixes a 'non-numeric value encountered' warning when running
PHP7.1. PR in Symfony to handle these: https://github.com/symfony/symfony/pull/20539

PR #6757, opened 2016-11-17
[Travis] Enable PHP 7.1 for tests

venyii:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |

Run the travis test suite in PHP 7.1 as well. This might help to spot errors like #6752

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,13 @@ matrix:
     include:
         -
             sudo: required
-            php: 7.0
+            php: 7.1
             env: SYLIUS_SUITE="application"
             services:
                 - memcached
         -
             sudo: false
-            php: 7.0
+            php: 7.1
             env: SYLIUS_SUITE="docs packages"
             addons:
                 apt:
--- a/etc/travis/suites/common/before_install.sh
+++ b/etc/travis/suites/common/before_install.sh
@@ -4,7 +4,7 @@ source "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/../../../bash/common.lib.s
 
 print_header "Customizing the environment" "Sylius"
 run_command "git fetch origin master:refs/remotes/origin/master" || exit $? # Make origin/master available for is_suitable steps
-run_command "phpenv config-rm xdebug.ini" || exit $? # Disable XDebug
+run_command "phpenv config-rm xdebug.ini" # Disable XDebug
 run_command "echo \"memory_limit=4096M\" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini" || exit $? # Increase memory limit to 4GB
 run_command "mkdir -p \"${SYLIUS_CACHE_DIR}\"" || exit $? # Create Sylius cache directory
 
--- a/src/Sylius/Behat/Context/Ui/Admin/ManagingProductVariantsContext.php
+++ b/src/Sylius/Behat/Context/Ui/Admin/ManagingProductVariantsContext.php
@@ -648,7 +648,7 @@ final class ManagingProductVariantsContext implements Context
      */
     public function iSpecifyThereAreVariantsIdentifiedByCode($nthVariant, $code)
     {
-        $this->generatePage->nameCode($nthVariant, $code - 1);
+        $this->generatePage->nameCode($nthVariant - 1, $code);
     }
 
     /**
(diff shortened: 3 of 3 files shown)


pamil:
Unfortunately, Sylius is quite large and this change would require additional ~20 minutes of Travis time per build, so I'd rather not do so. Anyway, as soon as PHP 7.1 gets released, we can replace the existing PHP 7.0 build with PHP 7.1 one.

venyii:
But aren't they running in parallel?

pamil:
There is a limit of 5 concurrent jobs per organization (the current build has 4 jobs, this proposed one has 6 jobs).

venyii:
Okay, how about something like the current change? It combines `application`/`packages` and lets PHP7 do the docs once. That's four builds total 🎉

venyii:
@pamil @pjedrzejewski Is this a viable change?

venyii:
So PHP 7.1 was released yesterday (:tada:), I rebased and excluded `packages` for PHP 5.6 like in master. Event though the build is green, a test is failing (same in master https://travis-ci.org/Sylius/Sylius/jobs/180624628).

michalmarcinkowski:
It will make the build even longer... How about running PHP 7.1 nightly?

alcaeus:
@michalmarcinkowski builds are run in parallel, so build times are not extended. Besides, since the 7.0 and 7.1 builds are about 30% faster than 5.6, maybe that version should be moved to a nightly build?

michalmarcinkowski:
@alcaeus they run in parallel, but still use the container that could be used for builds of another PR which result in longer waiting time for other PRs to pass. It's important to run PHP 5.6 on every PR because there are many backwards incompatible changes that should be resolved before merging, but with PHP 7.1 it's much more likely that everything will pass.

pamil:
What about replacing PHP 7.0 build with PHP 7.1?

venyii:
Do you want to use the new format (in this PR) or the old one (master)?

.travis.yml
```diff
@@ -1,40 +1,35 @@
...
+            - parallel
 env:
     global:
         - SYLIUS_CACHE_DIR=$HOME/.sylius-cache
         - SYLIUS_BUILD_DIR=etc/build
+    matrix:
+        - SYLIUS_SUITE="application packages"
 
 matrix:
+    fast_finish: true
```
pamil: Not needed.

.travis.yml
```diff
@@ -1,40 +1,35 @@
...
 language: php
 
+php:
+    - 5.6
+    - 7.1
+
+services:
+    - memcached
+
+addons:
```
pamil: Should be only included for "packages" build.

.travis.yml
```diff
@@ -1,40 +1,35 @@
+sudo: required
+
 language: php
 
+php:
+    - 5.6
+    - 7.1
+
+services:
```
pamil: Should be only included for "application" build.

.travis.yml
```diff
@@ -1,40 +1,35 @@
+sudo: required
```
pamil: Should be only included for "application" build.

.travis.yml
```diff
@@ -1,40 +1,35 @@
...
+addons:
+    apt:
+        packages:
+            - parallel
 env:
     global:
         - SYLIUS_CACHE_DIR=$HOME/.sylius-cache
         - SYLIUS_BUILD_DIR=etc/build
+    matrix:
+        - SYLIUS_SUITE="application packages"
```
pamil: Merging these two may cause timeouts in case when all packages are reinstalled.

etc/travis/suites/common/before_install.sh
```diff
@@ -4,7 +4,7 @@ source "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/../../../bash/common.lib.s
 
 print_header "Customizing the environment" "Sylius"
 run_command "git fetch origin master:refs/remotes/origin/master" || exit $? # Make origin/master available for is_suitable steps
-run_command "phpenv config-rm xdebug.ini" || exit $? # Disable XDebug
+run_command "phpenv config-rm xdebug.ini" # Disable XDebug
```
pamil: 👍

venyii:
Reverted the `.travis.yml` changes and only replaced the PHP version.

pjedrzejewski:
Thank you @venyii!

PR #6761, opened 2016-11-17
[Behat] Allow to define array of elements

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

According to [doc block](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Behat/Page/Page.php#L145-L149), an array can be returned from defined elements as well.

### DIFF
--- a/src/Sylius/Behat/Page/Page.php
+++ b/src/Sylius/Behat/Page/Page.php
@@ -234,7 +234,7 @@ abstract class Page implements PageInterface
             ));
         }
 
-        $elementSelector = strtr($definedElements[$name], $parameters);
+        $elementSelector = $this->resolveParameters($name, $parameters, $definedElements);
 
         return new NodeElement(
             $this->getSelectorAsXpath($elementSelector, $this->session->getSelectorsHandler()),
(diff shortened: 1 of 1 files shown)


src/Sylius/Behat/Page/Page.php
```diff
@@ -255,4 +255,20 @@ private function getSelectorAsXpath($selector, SelectorsHandler $selectorsHandle
...
     }
+
+    /**
+     * @param string $name
+     * @param array $parameters
+     * @param array$definedElements
+     *
+     * @return string
+     */
+    private function addParameters($name, array $parameters, array $definedElements)
```
pamil: `addParameters` -> `resolveParameters`

src/Sylius/Behat/Page/Page.php
```diff
@@ -255,4 +255,20 @@ private function getSelectorAsXpath($selector, SelectorsHandler $selectorsHandle
...
+    {
+        if (!is_array($definedElements[$name])) {
+            return strtr($definedElements[$name], $parameters);
+        }
+
+        return $definedElements[$name];
```
pamil: If it's an array, than `strtr` should be called for each value.

michalmarcinkowski:
👍

PR #6767, opened 2016-11-17
Remove formtypes specs

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | currently part of #6738 |
| License         | MIT |

Writing specifications for formtypes has no sense as we're mocking objects we don't own, assert only that "a method should be called" and overall copies the form, so changing anything (and formtypes are changed monstrously in #6738) results in big pain updating the specs. They bring no real value.

Proposing it as independent PR to make #6738 cleaner.


pjedrzejewski:
👍 👍 👍

PR #6777, opened 2016-11-18
[Admin][Taxon] Fix setting taxon's slug in multiple languages

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |


src/Sylius/Behat/Context/Setup/TaxonomyContext.php
```diff
@@ -148,6 +170,29 @@ private function createTaxon($name)
...
     /**
+     * @param array $names
+     *
+     * @return TaxonInterface
+     */
+    private function createTaxonInManyLanguages(array $names)
+    {
+        /** @var TaxonInterface $taxon */
+        $taxon = $this->taxonFactory->createNew();
+        $taxon->setCode(StringInflector::nameToCode($names['en_US']));
```
lchrusciel: What if 'en_US' would not be defined?

Zales0123: It will :boom: in a massive explosion that would be visible from the Moon! We could obviously pick there name in first locale.

pjedrzejewski:
Thank you Mateusz!

src/Sylius/Behat/Context/Setup/TaxonomyContext.php
lchrusciel: It could be a nice picture... 🎆

Issue #6641, opened 2016-11-06
Rounding Issues in Cart


bretto36:
I have an order in my system see image. The unit price * quantity doesn't match the total. When i send the order to paypal i get an error that the totals don't match.

![sylius order bug](https://cloud.githubusercontent.com/assets/6217994/20042745/fa0c5164-a4ca-11e6-95b7-5e9d23e41a4b.PNG)

bretto36:
My exchange rate was to 4 decimal places. That could be the issue. But i feel that the prices should be converted on unit price then the total formed from the unit price x quantity

PR #6779, opened 2016-11-18
[WIP] [Currency] Channel base currency

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | partialy #6641  |
| License         | MIT |


inssein:
@lchrusciel any chance you could provide some information with this PR? It's clear that you guys have had an internal discussion regarding some issues with handling of currency.. but it would be nice if sometimes those discussions were in the issues.

michalmarcinkowski:
Shortend story:
If we want to accept payments in multiple currencies we need to perform all operations, as well as save to the database values in this currency in order to prevent from rounding issues. With our current design (single base currency) we can allow for payments only in base currency and display the approximate amount in chosen currency, but  the customer will be charged the value in base currency anyway (if we want to avoid rounding issues #6641). This introduces a limitation that one Sylius instance can only accept payments in one currency... In order to get rid of this limitation we are going to make base currency configurable per channel, this way you will be able to accept payments in different currencies without issues! 🎉 

I know it's quite a big change, but we must resolve this issue...

docs/book/currencies.rst
```diff
@@ -36,26 +36,6 @@ To manage the currently used currency, we use the **CurrencyContext**. You can a
...
-    <?php
-
-    public function fooAction()
-    {
-        // convert 100 of the base currency (for instance 100$ if USD is your base) to PLN.
-        $this->get('sylius.currency_converter')->convertFromBase(100, 'PLN');
-
-        // or the other way - convert 100 PLN to amount in the base currency
-        $this->get('sylius.currency_converter')->convertToBase(100, 'PLN');
-    }
```
michalmarcinkowski: This service will stay, but the method signature will change. I would make it `convert($fromCurrencyCode, $toCurrencyCode, $amount)`

michalmarcinkowski: But I think it's not a concern of this PR, so I think you can leave it untouched 😄

src/Sylius/Behat/Context/Setup/CurrencyContext.php
```diff
@@ -173,16 +173,6 @@ public function thatChannelAllowsToShopUsingCurrency(ChannelInterface $channel,
...
 
     /**
-     * @Given /^the exchange rate for (currency "[^"]+") was changed to ([0-9\.]+)$/
-     * @Given /^the ("[^"]+" currency) has an exchange rate of ([0-9\.]+)$/
-     */
-    public function theExchangeRateForWasChangedTo(CurrencyInterface $currency, $exchangeRate)
-    {
-        $currency->setExchangeRate($exchangeRate);
-        $this->saveCurrency($currency);
-    }
```
michalmarcinkowski: Why removing this? In this PR you should only remove exchange rate from order. Removing it from currency will break a lot of things and should be done in separate PR.

src/Sylius/Behat/Context/Setup/CurrencyContext.php
```diff
@@ -202,7 +192,6 @@ private function createCurrency($currencyCode, $exchangeRate = 1.0)
         /** @var CurrencyInterface $currency */
         $currency = $this->currencyFactory->createNew();
         $currency->setCode($currencyCode);
-        $currency->setExchangeRate($exchangeRate);
```
michalmarcinkowski: Same as above.

src/Sylius/Bundle/CoreBundle/Collector/SyliusCollector.php
```diff
@@ -32,20 +33,18 @@
...
         array $bundles,
-        $defaultCurrencyCode,
         $defaultLocaleCode
     ) {
         $this->shopperContext = $shopperContext;
 
         $this->data = [
             'version' => Kernel::VERSION,
-            'default_currency_code' => $defaultCurrencyCode,
+            'default_currency_code' => null,
```
michalmarcinkowski: Let's rename it to `base_currency_code` (will be renamed in #6766).

src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
```diff
@@ -25,7 +25,6 @@
         <import resource="services/installer_requirements.xml" />
         <import resource="services/listeners.xml" />
         <import resource="services/order_processing.xml" />
-        <import resource="services/pricing.xml" />
```
michalmarcinkowski: No need to remove it.

src/Sylius/Bundle/CoreBundle/Resources/config/services/form.xml
```diff
@@ -61,7 +61,6 @@
         <service id="sylius.form.type.currency" class="%sylius.form.type.currency.class%">
             <argument>%sylius.model.currency.class%</argument>
             <argument>%sylius.validation_groups.currency%</argument>
-            <argument>%currency%</argument>
             <tag name="form.type" alias="sylius_currency" />
         </service>
```
michalmarcinkowski: I think we won't need to override this sevice in Core anymore.

src/Sylius/Bundle/CoreBundle/Resources/config/services/promotion.xml
```diff
@@ -44,15 +44,13 @@
...
             <argument type="service" id="sylius.promotion.units_promotion_adjustments_applicator" />
-            <argument type="service" id="sylius.currency_converter" />
             <tag name="sylius.promotion_action" type="order_fixed_discount" label="Order fixed discount" />
         </service>
         <service id="sylius.promotion_action.unit_fixed_discount" class="Sylius\Component\Core\Promotion\Action\UnitFixedDiscountPromotionActionCommand">
             <argument type="service" id="sylius.factory.adjustment" />
             <argument type="service" id="sylius.promotion_filter.price_range" />
             <argument type="service" id="sylius.promotion_filter.taxon" />
             <argument type="service" id="sylius.promotion_filter.product" />
-            <argument type="service" id="sylius.currency_converter" />
```
michalmarcinkowski: The promotions will be reworked separately, do the minimum you need to do to make scenarios for single channel green. Scenarios for multi currency and multi channel can be marked temporarly `@todo`

lchrusciel:
Closing temporary, to avoid travis spamming with my changes

PR #6782, opened 2016-11-18
[UserBundle] Delete routing and templates from UserBundle

### LINKED ISSUE #5859: Route 'sylius_user_security_login' does not exist anymore

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | fixes #5859 |
| License         | MIT |


tchapi:
Can you provide a little detail as to what this does exactly ? If fixes https://github.com/Sylius/Sylius/issues/5859 as in "we remove the whole functionality so basically this is fixed" ? What happens then with the reset password stuff, do we have to code it ourselves until you decide to put it back again ?

michalmarcinkowski:
@tchapi I don't understand your frustration... 😕  Did you have a look at the changes? This PR removes not really useful templates (that are not even tested) and routing (that is not tested also) and doesn't work correctly because of redirects to non existing routes. It fixes the Controller by removing wrong and hardcoded templates and redirects and forces to configure them in route definition.

I didn't remove any functionality, so please verify your concerns before writing such comments...

tchapi:
The frustration is in the explanations — for every PR that gets merged in master, we need to check and cherry-pick it to see what it does exactly and how we need to adjust our codebase before updating to the current master.

Of course we could stay at a particular commit hash but then, wait ten days and everything has changed and the migration is a nightmare. so we need to stay as close as possible to master to be able to bubble up all the changes that happen in our custom code.

It would be great if you just explained each PR with a few words, _especially_ if you tag it "BC break". Else it's just a game of chasing tails. We're really eager and excited to use Sylius but it's complicated if we have to do this guessing game for every PR that comes by.

michalmarcinkowski:
This week will be really hard for upgrades because we will do all necessary BC breaks before BETA release. After the 1.0.0-beta tag, BC breaks will be done only when really necessary and all will be well documented. 

We're approaching the great milestone that will hopefully end the "upgrade nightmare" for all people using Sylius, just be patient, beta incoming at the end of November, it will be worth!

pjedrzejewski:
Thank you Michał! @tchapi We are really close! :)

PR #6787, opened 2016-11-20
Show name field in product variant form so it wont get lost on updating

### LINKED ISSUE #6786: [Admin] Product variant is lost after editing the variant
In our project we are working in a import script that creates variant. Each variant has it's own name. 

But after, when you try to update the variant from the admin interface, name fields is not being showed in the form. Submitting the form updates the variant setting its name to empty... 

this implies that you can't identify the variant in the product page. see attached pic. 
captura de pantalla de 2016-11-20 13-04-00

crevillo:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #6786 |
| License         | MIT |

Just adds the name field in the product variant edit form.


pjedrzejewski:
Thanks Carlos! :)

PR #6791, opened 2016-11-21
[Behat] Fix base currency step semantics

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | #6766 |
| License         | MIT |

According to upcoming changes with currencies, it will be more appropriate to have base currency set in channel creation step (``United States`` channel has obviously ``USD`` base currency).

### DIFF
--- a/features/account/customer_account/viewing_orders_history/viewing_order_recalculated_by_exchange_rate.feature
+++ b/features/account/customer_account/viewing_orders_history/viewing_order_recalculated_by_exchange_rate.feature
@@ -11,7 +11,7 @@ Feature: Viewing details of an order
         And the store operates on a channel named "Web"
         And that channel allows to shop using the "USD" currency
         And that channel allows to shop using the "GBP" currency with exchange rate 3.0
-        And that channel uses the "USD" currency by default
+        And that channel uses the "USD" currency as base
         And the store allows paying with "Cash on Delivery"
         And the store has "DHL" shipping method with "$20.00" fee within the "EN" zone
         And the store has a product "Angel T-Shirt" priced at "$20.00"
--- a/features/account/customer_account/viewing_orders_history/viewing_order_recalculated_by_exchange_rate.feature
+++ b/features/account/customer_account/viewing_orders_history/viewing_order_recalculated_by_exchange_rate.feature
@@ -8,10 +8,9 @@ Feature: Viewing details of an order
         Given the store ships to "British Virgin Islands"
         And the store has a zone "English" with code "EN"
         And this zone has the "British Virgin Islands" country member
-        And the store operates on a channel named "Web"
+        And the store operates on a channel named "Web" in currency "USD"
         And that channel allows to shop using the "USD" currency
         And that channel allows to shop using the "GBP" currency with exchange rate 3.0
-        And that channel uses the "USD" currency as base
         And the store allows paying with "Cash on Delivery"
         And the store has "DHL" shipping method with "$20.00" fee within the "EN" zone
         And the store has a product "Angel T-Shirt" priced at "$20.00"
--- a/features/account/customer_account/viewing_orders_history/viewing_order_recalculated_by_exchange_rate.feature
+++ b/features/account/customer_account/viewing_orders_history/viewing_order_recalculated_by_exchange_rate.feature
@@ -8,8 +8,7 @@ Feature: Viewing details of an order
         Given the store ships to "British Virgin Islands"
         And the store has a zone "English" with code "EN"
         And this zone has the "British Virgin Islands" country member
-        And the store operates on a channel named "Web" in currency "USD"
-        And that channel allows to shop using the "USD" currency
+        And the store operates on a channel named "Web" in "USD" currency
         And that channel allows to shop using the "GBP" currency with exchange rate 3.0
         And the store allows paying with "Cash on Delivery"
         And the store has "DHL" shipping method with "$20.00" fee within the "EN" zone
--- a/features/cart/shopping_cart/adding_product_to_cart_with_prices_dependent_on_channel_and_currency.feature
+++ b/features/cart/shopping_cart/adding_product_to_cart_with_prices_dependent_on_channel_and_currency.feature
@@ -7,9 +7,9 @@ Feature: Adding a product to cart with prices dependent on a channel and a curre
     Background:
         Given the store has currency "EUR"
         And the store has currency "GBP" with exchange rate 0.7
-        And the store operates on a channel named "Web-EU" in currency "EUR"
+        And the store operates on a channel named "Web-EU" in "EUR" currency
         And that channel allows to shop using "EUR" and "GBP" currencies
-        And the store operates on another channel named "Web-GB"
+        And the store operates on another channel named "Web-GB" in "GBP" currency
         And that channel allows to shop using "EUR" and "GBP" currencies
         And the store has a product "Leprechaun's Gold" priced at "€12.54" available in channel "Web-EU" and channel "Web-GB"
         And it has different prices for different channels and currencies
--- a/features/cart/shopping_cart/adding_product_to_cart_with_prices_dependent_on_channel_and_currency.feature
+++ b/features/cart/shopping_cart/adding_product_to_cart_with_prices_dependent_on_channel_and_currency.feature
@@ -9,10 +9,10 @@ Feature: Adding a product to cart with prices dependent on a channel and a curre
         And the store has currency "GBP" with exchange rate 0.7
         And the store operates on a channel named "Web-EU"
         And that channel allows to shop using "EUR" and "GBP" currencies
-        And that channel uses the "EUR" currency by default
+        And that channel uses the "EUR" currency as base
         And the store operates on another channel named "Web-GB"
         And that channel allows to shop using "EUR" and "GBP" currencies
-        And that channel uses the "GBP" currency by default
+        And that channel uses the "GBP" currency as base
         And the store has a product "Leprechaun's Gold" priced at "€12.54" available in channel "Web-EU" and channel "Web-GB"
         And it has different prices for different channels and currencies
         And it has price "€10.00" for "Web-EU" channel and "EUR" currency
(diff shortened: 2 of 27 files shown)


features/account/customer_account/viewing_orders_history/viewing_order_recalculated_by_exchange_rate.feature
```diff
@@ -11,7 +11,7 @@ Feature: Viewing details of an order
         And the store operates on a channel named "Web"
         And that channel allows to shop using the "USD" currency
         And that channel allows to shop using the "GBP" currency with exchange rate 3.0
-        And that channel uses the "USD" currency by default
+        And that channel uses the "USD" currency as base
```
lchrusciel: I would change a whole sentence... `And that channel operates on "USD" currency`?

Zales0123: Not exactly, as channel can operate on many currencies, but it has only one base currency. Maybe sth like `And that channel's base currency is "USD"`?

lchrusciel: Not exactly. Channel will operate only on a one currency, but prices can be shown in different for your convenience. I would stay with `And that channel operates on "USD" currency`

Zales0123: Not exactly. Channel **will** operate only on a one currency, but for now it can operate on many of them. So for now this sentence would be incorrect.

lchrusciel: Not exactly. Channel can operate only on one currency from the beginning. If we want to be so semantical strict, the sentence should be ` And that channel base display currency is "USD"`. Anyway, with your last argument there is no point in changing it now, because this sentence will be for 100% in one/two days.

Zales0123: Not exactly. There is a [concept](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Model/Channel.php#L29) of ``baseCurrency`` on ``Channel`` entity and it will not be deleted in the nearest (or further) future. So the sentence ``And that channel's base currency is "USD"`` is 100% correct now and in the future. ``And that channel uses the "USD" currency by default`` is not 100% correct neither now nor in the future, IMO the same with ``And that channel operates on "USD" currency``. ``channel's base currency`` is not open to interpretation, it's exactly what we do with channel (especially as it's the setup step).

features/cart/shopping_cart/adding_product_to_cart_with_prices_dependent_on_channel_and_currency.feature
```diff
@@ -7,12 +7,10 @@ Feature: Adding a product to cart with prices dependent on a channel and a curre
     Background:
         Given the store has currency "EUR"
         And the store has currency "GBP" with exchange rate 0.7
-        And the store operates on a channel named "Web-EU"
+        And the store operates on a channel named "Web-EU" in currency "EUR"
         And that channel allows to shop using "EUR" and "GBP" currencies
-        And that channel uses the "EUR" currency as base
         And the store operates on another channel named "Web-GB"
```
michalmarcinkowski: ... another channel named "Web-GB" **in currency "GBP"**

features/account/customer_account/viewing_orders_history/viewing_order_recalculated_by_exchange_rate.feature
```diff
@@ -8,10 +8,9 @@ Feature: Viewing details of an order
         Given the store ships to "British Virgin Islands"
         And the store has a zone "English" with code "EN"
         And this zone has the "British Virgin Islands" country member
-        And the store operates on a channel named "Web"
+        And the store operates on a channel named "Web" in currency "USD"
         And that channel allows to shop using the "USD" currency
```
pamil: Is it still needed?

features/cart/shopping_cart/adding_product_to_cart_with_prices_dependent_on_channel_and_currency.feature
```diff
@@ -7,12 +7,10 @@ Feature: Adding a product to cart with prices dependent on a channel and a curre
     Background:
         Given the store has currency "EUR"
         And the store has currency "GBP" with exchange rate 0.7
-        And the store operates on a channel named "Web-EU"
+        And the store operates on a channel named "Web-EU" in currency "EUR"
         And that channel allows to shop using "EUR" and "GBP" currencies
-        And that channel uses the "EUR" currency by default
         And the store operates on another channel named "Web-GB"
```
michalmarcinkowski: ... another channel named "Web-GB" **in "GBP" currency

features/checkout/seeing_order_summary/seeing_prices_recalculated_by_exchange_rate_on_order_summary.feature
```diff
@@ -8,10 +8,9 @@ Feature: Seeing prices recalculated by exchange rate on order summary
         Given the store ships to "British Virgin Islands"
         And the store has a zone "English" with code "EN"
         And this zone has the "British Virgin Islands" country member
-        And the store operates on a channel named "Web"
+        And the store operates on a channel named "Web" in currency "USD"
         And that channel allows to shop using the "USD" currency
```
pamil: Same here etc.

src/Sylius/Component/Core/Test/Services/DefaultUnitedStatesChannelFactory.php
```diff
@@ -206,14 +199,18 @@ private function createCountry()
     /**
      * @return CurrencyInterface
      */
-    private function provideCurrency()
+    private function provideCurrency($currencyCode = null)
     {
+        if (null === $currencyCode) {
```
pamil: `$currencyCode = $currencyCode ?: self::DEFAULT_CURRENCY_CODE;`

src/Sylius/Behat/Context/Setup/CurrencyContext.php
```diff
@@ -144,10 +144,10 @@ public function thatChannelAllowsToShopUsingAndCurrencies(ChannelInterface $chan
     }
 
     /**
-     * @Given /^(that channel) uses the "([^"]+)" currency by default$/
-     * @Given /^(it) uses the "([^"]+)" currency by default$/
+     * @Given /^(that channel) uses the "([^"]+)" currency as base$/
+     * @Given /^(it) uses the "([^"]+)" currency as base$/
      */
-    public function itUsesTheCurrencyByDefault(ChannelInterface $channel, $currencyCode)
+    public function itUsesTheCurrencyAsBase(ChannelInterface $channel, $currencyCode)
```
michalmarcinkowski: Should be removed.

src/Sylius/Component/Core/Test/Services/DefaultChannelFactory.php
```diff
@@ -127,14 +127,18 @@ public function create($code = null, $name = null)
     /**
      * @return CurrencyInterface
      */
-    private function provideCurrency()
+    private function provideCurrency($currencyCode = null)
     {
+        if (null === $currencyCode) {
```
pamil: `$currencyCode = $currencyCode ?: $this->defaultCurrencyCode;`

michalmarcinkowski:
Which of these two is correct `... in currency "USD"` or `... in "USD" currency`? /cc @TheMadeleine

TheMadeleine:
@michalmarcinkowski both are ok, but  ``in "USD" currency`` sounds better imo.

michalmarcinkowski:
I would vote for `in "USD" currency` also. @Zales0123 could you correct it? :)

src/Sylius/Component/Core/Test/Services/DefaultUnitedStatesChannelFactory.php
```diff
@@ -206,14 +199,16 @@ private function createCountry()
     /**
      * @return CurrencyInterface
      */
-    private function provideCurrency()
+    private function provideCurrency($currencyCode = null)
     {
+        $currencyCode = $currencyCode ?: self::DEFAULT_CURRENCY_CODE;
```
michalmarcinkowski: I think we can always use `USD` in this case, it's "The United States" channel, so the currency is pretty much known and immutable :) In Default channel it makes sense, but for this specific factory it doesn't IMO.

src/Sylius/Behat/Context/Setup/OrderContext.php
```diff
@@ -567,10 +567,9 @@ private function createCart(
 
         $order->setCustomer($customer);
         $order->setChannel((null !== $channel) ? $channel : $this->sharedStorage->get('channel'));
-        $order->setCurrencyCode((null !== $currencyCode) ? $currencyCode : $this->sharedStorage->get('currency')->getCode());
         $order->setLocaleCode((null !== $localeCode) ? $localeCode : $this->sharedStorage->get('locale')->getCode());
 
-        $currencyCode = $currencyCode ? $currencyCode : $this->sharedStorage->get('currency')->getCode();
+        $currencyCode = $currencyCode ? $currencyCode : $order->getChannel()->getBaseCurrency()->getCode();
```
michalmarcinkowski: The order should be always created with base currency of the channel in which it was placed. We should not allow to pass other value here.

src/Sylius/Behat/Context/Setup/ChannelContext.php
```diff
@@ -105,25 +114,28 @@ public function storeOperatesOnASingleChannelInTheUnitedStatesNamed($channelIden
...
      */
-    public function theStoreOperatesOnAChannelNamed($channelIdentifier)
+    public function theStoreOperatesOnAChannelNamed($channelIdentifier, $currencyCode = null)
     {
-        $defaultData = $this->defaultChannelFactory->create($channelIdentifier, $channelIdentifier);
+        $defaultData = $this->defaultChannelFactory->create($channelIdentifier, $channelIdentifier, $currencyCode);
 
         $this->sharedStorage->setClipboard($defaultData);
         $this->sharedStorage->set('channel', $defaultData['channel']);
+        $this->currencyStorage->set($defaultData['channel'], $defaultData['currency']->getCode());
```
michalmarcinkowski: Do we need this?

michalmarcinkowski:
Thanks Mateusz! Please apply my comments in a separate PR.

PR #6794, opened 2016-11-21
[Channel] Remove base currency from parameters and move to channel

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | partialy #6641  |
| License         | MIT |

Reopen of #6779 (force-pushed changes block reopening)

Shortend story based on @michalmarcinkowski [comment](https://github.com/Sylius/Sylius/pull/6779):

> If we want to accept payments in multiple currencies we need to perform all operations, as well as save to the database values in this currency in order to prevent from rounding issues. With our current design (single base currency) we can allow for payments only in base currency and display the approximate amount in chosen currency, but the customer will be charged the value in base currency anyway (if we want to avoid rounding issues #6641). This introduces a limitation that one Sylius instance can only accept payments in one currency... In order to get rid of this limitation we are going to make base currency configurable per channel, this way you will be able to accept payments in different currencies without issues! 🎉 I know it's quite a big change, but we must resolve this issue...


src/Sylius/Component/Core/Pricing/ChannelAndCurrencyBasedCalculator.php
```diff
@@ -49,14 +49,7 @@ public function calculate(PriceableInterface $subject, array $configuration, arr
         $channel = $context['channel'];
         $currencyCode = $context['currency'];
```
tuka217: This variables should be removed.
Is it needed to pass `$configuration` and `$context` if you only return price from `$subject`?

lchrusciel:
Because of many small changes connected with this PR, we will work on separate branch, and merge whole improvement to core at once. Please, help us with reviewing it part by part, because later reviewing could be quite painful.

PR #6813, opened 2016-11-22
[Behat] Fix base channel currency step

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

I open this PR once more time to avoid merge problems on ``currency-party``.

PR #6823, opened 2016-11-23
[Core] Select variant price for each channel in base currency

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | based on #6813  |
| License         | MIT |

This is a huge, but for now necessary change in our pricing logic. Now we're deleting concept of pricing calculators and configuration to achieve one simple goal - allow (and extort) to define product prices in each channel base currency. Therefore, we introduced ``ChannelPricing`` entity, which is used to configure price per variant and channel.
For now, we must select price for each channel (even if product is not available in it), but of course we would to improve it in the nearest future.
The main goal is to completely remove calculating price in proper currency from logic, we want all operation to be done in base currency of channel, the only thing when currency would change is view.

- [x] removed pricing configuration for product variants
- [x] introduced ChannelPricing concept
- [x] implemented default calculator to get price for current channel, together with corresponding twig filters
- [x] adjusted scenarios to handle new logic
- [x] fix displaying different prices for options/variants on product show page
- [x] validate do variant has prices configured for channels
- [x] fix product variant generation


features/product/managing_product_reviews/accepting_product_review.feature
```diff
@@ -5,7 +5,8 @@ Feature: Accepting a product review
     I want to be able to accept a product review
 
     Background:
-        Given the store has customer "Mike Ross" with email "ross@teammike.com"
+        Given the store operates on a single channel in "United States"
+        And the store has customer "Mike Ross" with email "ross@teammike.com"
```
michalmarcinkowski: Do we need channel for admin feature?

features/product/managing_products/product_validation.feature
```diff
@@ -5,14 +5,14 @@ Feature: Products validation
...
+        Given the store operates on a single channel in "United States"
         And I am logged in as an administrator
 
     @ui
     Scenario: Adding a new simple product without specifying its code
-        Given I want to create a new simple product
-        When I name it "Dice Brewing" in "English (United States)"
-        And I set its price to "$10.00"
+        When I want to create a new simple product
+        And I name it "Dice Brewing" in "English (United States)"
```
michalmarcinkowski: Currently we use `Given I want to...` and `When I name it ...` everywhere I think.

features/product/managing_products/removing_product_price_after_channel_deletion.feature
```diff
@@ -0,0 +1,22 @@
...
+        And I am logged in as an administrator
+
+    @ui @todo
+    Scenario: Removing a product's price after corresponding channel deletion
+        When I delete channel "Web-GB"
+        Then product "Dice Brewing" should have price "$10.00" for channel "Web-US"
+        And this product should no longer have price for channel "Web=GB"
+
+
+
```
michalmarcinkowski: Too many blank lines

features/product/managing_products/removing_product_price_after_channel_deletion.feature
```diff
@@ -0,0 +1,22 @@
...
+        And the store operates on a channel named "Web-US" in "USD" currency
+        And the store operates on another channel named "Web-GB" in "GBP" currency
+        And the store has a product "Dice Brewing" priced at "$10.00" for "Web-US" channel and "£5.00" for "Web-GB" channel
+        And I am logged in as an administrator
+
+    @ui @todo
+    Scenario: Removing a product's price after corresponding channel deletion
+        When I delete channel "Web-GB"
+        Then product "Dice Brewing" should have price "$10.00" for channel "Web-US"
+        And this product should no longer have price for channel "Web=GB"
```
michalmarcinkowski: There is no `Web=GB` channel, so this scenario will be always green 😄

src/Sylius/Behat/Context/Ui/Admin/ManagingProductVariantsContext.php
```diff
@@ -161,12 +161,12 @@ public function iEnableItsTracking()
...
     /**
-     * @When /^I set its(?:| default) price to ("(?:-)?(?:€|£|\$)[^"]+")$/
+     * @When /^I set its(?:| default) price to (?:|-)(?:€|£|\$)([^"]+) for "([^"]+)" channel$/
      * @When I do not set its price
      */
-    public function iSetItsPriceTo($price = null)
+    public function iSetItsPriceTo($price = null, $channel = null)
     {
-        $this->createPage->specifyPrice($price);
+        $this->createPage->specifyPrice($channel, $price);
```
michalmarcinkowski: Why different order of the arguments than in the above method?

src/Sylius/Behat/Context/Ui/Admin/ManagingProductVariantsContext.php
```diff
@@ -276,7 +289,7 @@ public function iShouldSeeProductVariantsInTheList($numberOfProductVariants)
     }
 
     /**
-     * @When /^I delete the ("[^"]+" variant of product "[^"]+")$/
+     * @When /^I delete the (źź variant of product "[^"]+")$/
```
michalmarcinkowski: ?? 😕

Zales0123: Probably not needed XDXD

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
```diff
@@ -269,15 +285,32 @@ private function createVariants(ProductInterface $product, array $options)
         /** @var ProductVariantInterface $productVariant */
         foreach ($product->getVariants() as $productVariant) {
             $productVariant->setAvailableOn($this->faker->dateTimeThisYear);
-            $productVariant->setPrice($this->faker->randomNumber(4));
             $productVariant->setCode(sprintf('%s-variant#%d', $options['code'], $i));
             $productVariant->setOnHand($this->faker->randomNumber(1));
 
+            foreach ($this->channelRepository->findAll() as $channel) {
+                $this->createChannelPricings($productVariant, $channel);
+            }
```
michalmarcinkowski: We did the random price because we couldn't create variant without price, now there is no need to create random prices for each channel, because they are probably not used anyway.

src/Sylius/Bundle/CoreBundle/Form/Type/Product/ChannelPricingType.php
```diff
@@ -0,0 +1,63 @@
...
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setDefaultOptions(OptionsResolverInterface $resolver)
+    {
+        $resolver->setDefaults([
+            'data_class' => ChannelPricing::class,
```
michalmarcinkowski: I think this should not be hardcoded, but injected together with validation groups as we do for all resources (extend `AbstractResourceType`)

src/Sylius/Bundle/CoreBundle/Form/Type/Product/ProductVariantType.php
```diff
@@ -88,39 +77,13 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'empty_value' => '---',
                 'label' => 'sylius.form.product_variant.tax_category',
             ])
-            ->add('pricingCalculator', 'sylius_price_calculator_choice', [
-                'label' => 'sylius.form.shipping_method.calculator',
+            ->add('channelPricings', CollectionType::class, [
+                'entry_type' => ChannelPricingType::class,
```
michalmarcinkowski: Do we need this? We should not hardcoded the formtype

src/Sylius/Bundle/CoreBundle/Resources/config/services/form.xml
```diff
@@ -186,5 +179,13 @@
             <argument>%sylius.validation_groups.order_item%</argument>
             <tag name="form.type" alias="sylius_add_to_cart" />
         </service>
+
+        <service id="sylius.form.type.channel_pricing" class="Sylius\Bundle\CoreBundle\Form\Type\Product\ChannelPricingType">
+            <tag name="form.type" alias="sylius_channel_pricing" />
+        </service>
```
michalmarcinkowski: Shouldn't we configure this form type in `Configuration.php` as default?

src/Sylius/Bundle/CoreBundle/Resources/config/validation/ProductVariant.xml
```diff
@@ -13,17 +13,6 @@
...
-            <constraint name="NotBlank">
-                <option name="message">sylius.product_variant.price.not_blank</option>
-                <option name="groups">sylius</option>
-            </constraint>
-            <constraint name="Range">
-                <option name="min">0.00</option>
-                <option name="minMessage">sylius.product_variant.price.min</option>
-                <option name="groups">sylius</option>
-            </constraint>
-        </property>
```
Zales0123: Yup, validation will be added in upcoming commits 👍

src/Sylius/Bundle/CoreBundle/Templating/Helper/ChannelBasedPriceHelper.php
```diff
@@ -0,0 +1,80 @@
...
+        $this->productVariantPriceCalculator = $productVariantPriceCalculator;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriceForCurrentChannel(ProductVariantInterface $productVariant)
+    {
+        /** @var OrderInterface $currentCart */
+        $currentCart = $this->cartContext->getCart();
```
michalmarcinkowski: This helper shouldn't be stateful, we should pass the channel as an argument (in templates it's under `shopperContext.channel`).

src/Sylius/Bundle/CoreBundle/Templating/Helper/ChannelBasedPriceHelper.php
```diff
@@ -0,0 +1,80 @@
...
+
+        try {
+            $price = $this->productVariantPriceCalculator->calculate($productVariant, $currentCart->getChannel());
+        } catch (\InvalidArgumentException $exception) {
+            return null;
```
michalmarcinkowski: We should let the exception go 😄  the price should always be specified for the channel.

src/Sylius/Bundle/ShopBundle/Resources/views/Product/Show/_price.html.twig
```diff
@@ -1,3 +1,3 @@
 <span class="ui huge header" id="product-price">
-    {{ product.price|sylius_price }}
+    {{ product.variants.first()|sylius_channel_variant_price }}
```
michalmarcinkowski: We should use `DefaultVariantsResolver` here.

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculatorInterface.php
```diff
@@ -0,0 +1,31 @@
...
+ */
+interface ProductVariantPriceCalculatorInterface
+{
+    /**
+     * @param ProductVariantInterface $productVariant
+     * @param ChannelInterface $channel
+     *
+     * @return int
+     *
+     * @throws \InvalidArgumentException
```
michalmarcinkowski: "when price for given channel doesn't exist`

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+final class ProductVariantPriceCalculator implements ProductVariantPriceCalculatorInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function calculate(ProductVariantInterface $productVariant, ChannelInterface $channel)
+    {
+        $channelPricing = $productVariant->getChannelPricingForChannel($channel);
+        Assert::notNull($channelPricing);
```
michalmarcinkowski: We should add custom message here

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculatorInterface.php
```diff
@@ -0,0 +1,31 @@
...
+     */
+    public function calculate(ProductVariantInterface $productVariant, ChannelInterface $channel);
```
michalmarcinkowski: We decided on passing `array $context` instead of `ChannelInterface` to allow for different calculation strategies (e.g. per customer pricing).

Zales0123: Yeah, I forgot about that 🐃

src/Sylius/Bundle/CoreBundle/Templating/Helper/ChannelBasedPriceHelper.php
Zales0123: I didn't know about that! Great 👍

src/Sylius/Behat/Context/Ui/Admin/ManagingProductVariantsContext.php
lchrusciel: :D

src/Sylius/Bundle/ShopBundle/Resources/views/Product/Show/_price.html.twig
Zales0123: Yup, but it would require to implement separate filter - which would make more mess in this already messed up PR 😄  For now I would vote to leave it as it is (moreover, this has been done this way before -> by getting price with ``getPrice()`` method) and improve it later.

src/Sylius/Bundle/CoreBundle/Form/Type/Product/ProductVariantType.php
Zales0123: I've done it this way due to upcoming upgrade to Symfony3 - or can it be done differently?

src/Sylius/Bundle/CoreBundle/Form/Type/Product/ChannelPricingType.php
pamil: `$this->dataClass`

src/Sylius/Bundle/CoreBundle/Form/Type/Product/ProductVariantType.php
pamil: I'm loving it <3

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
Zales0123: I think that unfortunately we have two :/ These products are in order fixtures taken to order and due to order process we have to have prices specified for each channel (otherwise it would 💥 )

features/product/managing_product_variants/adding_product_variant.feature
```diff
@@ -14,16 +14,17 @@ Feature: Adding a new product variant
     Scenario: Adding a new product variant
         Given I want to create a new variant of this product
         When I specify its code as "VODKA_WYBOROWA_PREMIUM"
-        And I set its price to "$100.00"
+        And I set its price to $100.00 for "United States" channel
```
michalmarcinkowski: Why removing quotes?

features/product/managing_product_variants/generating_product_variant_validation.feature
```diff
@@ -15,13 +15,13 @@ Feature: Generating product variant generation
         Given I want to generate new variants for this product
         When I specify that the 1st variant is identified by "WYBOROWA_ORANGE" code
         And I try to generate it
-        Then I should be notified that price is required for the 1st variant
+        Then I should be notified that prices in all channels must be defined for the 1st variant
```
michalmarcinkowski: Prices should be required only for channels in which the product is available. What is more when we enable product in a specific channel we should validate that all it's variants have specified prices for this channel.

src/Sylius/Behat/Context/Setup/PaymentContext.php
```diff
@@ -96,7 +96,7 @@ public function storeAllowsPaying($paymentMethodName, $position = null)
      */
     public function storeAllowsPayingForAllChannels($paymentMethodName, array $channels)
     {
-        $paymentMethod = $this->createPaymentMethod($paymentMethodName, 'PM_'.$paymentMethodName, 'Payment method', false);
+        $paymentMethod = $this->createPaymentMethod($paymentMethodName, 'PM_'.$paymentMethodName, 'Payment method');
```
michalmarcinkowski: `'PM_'.$paymentMethodName` - we should use `StringInflector` for that.

lchrusciel: the last parameter should remain `false`

src/Sylius/Bundle/CoreBundle/Templating/Helper/ChannelBasedPriceHelper.php
```diff
@@ -0,0 +1,83 @@
...
+        /** @var OrderInterface $currentCart */
+        $currentCart = $this->cartContext->getCart();
+
+        try {
+            $price = $this
+                ->productVariantPriceCalculator
+                ->calculate($productVariant, ['channel' => $currentCart->getChannel()])
+            ;
+        } catch (\InvalidArgumentException $exception) {
+            return null;
```
michalmarcinkowski: Don't catch the exception :)

src/Sylius/Bundle/CoreBundle/Validator/Constraints/HasAllPricesDefinedValidator.php
```diff
@@ -0,0 +1,57 @@
...
+
+        $enabledChannels = $this->channelRepository->findBy(['enabled' => true]);
+
+        foreach ($enabledChannels as $channel) {
+            /** @var ChannelPricingInterface $channelPricing */
+            $channelPricing = $productVariant->getChannelPricingForChannel($channel);
+            if (null === $channelPricing || null === $channelPricing->getPrice()) {
+                $this->context->addViolationAt('channelPricings', $constraint->getMessage());
+
+                return;
```
michalmarcinkowski: I think we shouldn't return on first error, we should provide validation error for every amount that should be fulfilled.

Zales0123: I agree 👍  But would we be ok with merging this as it is to not lengthen this PR and I will fix this immediately after merge?

michalmarcinkowski:
Awesome job Mateusz! 🎉 💥

PR #6821, opened 2016-11-22
Use FQCNs for form types instead of aliases - first part: basic types

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | kind of |
| Related tickets | based on #6816, extracted from #6738 |
| License         | MIT |

PR #6832, opened 2016-11-23
Use FQCNs for form types instead of aliases - second part: bundles

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | based on #6821, extracted from #6738 |
| License         | MIT |


pjedrzejewski:
Thanks Kamil!

PR #6835, opened 2016-11-23
Use FQCNs for form types instead of aliases - third part: bundles

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | based of #6832 |
| License         | MIT |


src/Sylius/Bundle/CoreBundle/Form/Type/Order/AddToCartType.php
```diff
@@ -12,34 +12,36 @@
...
      * {@inheritdoc}
      */
     public function configureOptions(OptionsResolver $resolver)
     {
-        parent::configureOptions($resolver);
-
         $resolver
+            ->setDefaults([
+                'data_class' => AddToCartCommand::class,
+                'validation_groups' => ['sylius'],
```
michalmarcinkowski: These should be configurable not hardcoded, why not extend `AbstractResourceType`?

pamil: :+1:

pjedrzejewski:
Thank you Kamil! :)

PR #6844, opened 2016-11-24
[Collector] Fix multichannel handling

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Collector/SyliusCollector.php
+++ b/src/Sylius/Bundle/CoreBundle/Collector/SyliusCollector.php
@@ -12,6 +12,8 @@
 namespace Sylius\Bundle\CoreBundle\Collector;
 
 use Sylius\Bundle\CoreBundle\Application\Kernel;
+use Sylius\Component\Channel\Context\ChannelNotFoundException;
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/CoreBundle/Collector/SyliusCollector.php
```diff
@@ -120,7 +122,7 @@ public function collect(Request $request, Response $response, \Exception $except
 
             $this->data['base_currency_code'] = $channel->getBaseCurrency()->getCode();
             $this->data['currency_code'] = $this->shopperContext->getCurrencyCode();
-        } catch (CurrencyNotFoundException $exception) {}
+        } catch (ChannelNotFoundException $exception) {}
```
michalmarcinkowski: I think it would be better to catch both exceptions, `$this->shopperContext->getCurrencyCode()` can also throw `CurrencyNotFoundException` even if there is a channel.

michalmarcinkowski:
👍

PR #6865, opened 2016-11-25
[Core] State machine clean up

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| License         | MIT |


src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine/sylius_order.yml
```diff
@@ -1,8 +1,36 @@
...
+                to: fulfilled
         callbacks:
+            before:
+                sylius_assign_number:
+                    on: "create"
+                    do: ["@sylius.order_number_assigner", "assignNumber"]
+                    args: ["object"]
             after:
-                sylius_order_confirmation_email:
+                sylius_complete_checkout:
```
pjedrzejewski: This should be on the checkout SM and should be named something like: sylius_save_checkout_completion_date? Because that's the purpose of this callback.

Also, we could rename this service, callback is too generic and is part of our "legacy" from old state machines.

pjedrzejewski:
Thank you Arek! :)

PR #6873, opened 2016-11-26
[BC Break] Initial repository cleanup

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |


src/Sylius/Bundle/AddressingBundle/Doctrine/ORM/ZoneRepository.php
```diff
@@ -1,36 +0,0 @@
...
-        return $this->createQueryBuilder('o')
-            ->addSelect('members')
-            ->leftJoin('o.members', 'members')
-            ->where('o.name = :name')
-            ->setParameter('name', $name)
-            ->getQuery()
-            ->getResult()
-        ;
-    }
-}
```
pjedrzejewski: Why removing this? You pretty much always need zone with its members.

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductOptionRepository.php
```diff
@@ -22,30 +22,28 @@ class ProductOptionRepository extends EntityRepository implements ProductOptionR
...
-            ->getQuery()
-            ->getResult()
         ;
     }
 
     /**
      * {@inheritdoc}
      */
-    public function createListQueryBuilder($locale)
+    public function findByName($name, $locale)
```
pjedrzejewski: findOneByName?

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/PaymentMethodRepository.php
```diff
@@ -24,7 +24,6 @@ public function createListQueryBuilder($locale)
     {
         return $this
             ->createQueryBuilder('o')
-            ->addSelect('translation')
```
pjedrzejewski: I am pretty sure this will cause to lazy load translation contents anyway, as they will not be part of the select clause.

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductOptionRepository.php
michalmarcinkowski: The name is not unique, so we can have more than 1 object returned by this method.

src/Sylius/Bundle/AddressingBundle/Doctrine/ORM/ZoneRepository.php
michalmarcinkowski: This method is only used in a Behat context and I'm not sure you will ever use it in your application, because it's more likely that you will use `code` for querying. 

Using "default" joins is a separate topic, if we want to introduce it we should do it for all entities.

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/PaymentMethodRepository.php
michalmarcinkowski: You're right! There are many methods without selects, we will have to add them...

pjedrzejewski:
Thanks Michał!

PR #6875, opened 2016-11-26
[RFC][BC Break] Rename checkout summary event

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |

IMO `complete` is more appropriate name for an event dispatched after the checkout summary step.


pjedrzejewski:
Definitely, all events are matching the transition name. Thanks Michał!

PR #6876, opened 2016-11-26
[BugFix] Nth order promotion should exclude cancelled orders

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
@@ -164,8 +164,14 @@ class OrderRepository extends BaseOrderRepository implements OrderRepositoryInte
      */
     public function countByCustomer(CustomerInterface $customer)
     {
-        $count = $this->createByCustomerQueryBuilder($customer)
+        $queryBuilder = $this->createQueryBuilder('o');
+
+        $count = $queryBuilder
             ->select('count(o.id)')
+            ->andWhere('o.customer = :customer')
+            ->andWhere($queryBuilder->expr()->notIn('o.state', ':states'))
+            ->setParameter('customer', $customer)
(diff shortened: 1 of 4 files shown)


src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
```diff
@@ -164,8 +164,14 @@ public function createCheckoutsPaginator(array $criteria = null, array $sorting
...
     {
-        $count = $this->createByCustomerQueryBuilder($customer)
+        $queryBuilder = $this->createQueryBuilder('o');
+
+        $count = $queryBuilder
             ->select('count(o.id)')
+            ->andWhere('o.customer = :customer')
+            ->andWhere($queryBuilder->expr()->notIn('o.state', ':states'))
+            ->setParameter('customer', $customer)
+            ->setParameter('states', [OrderInterface::STATE_CART, OrderInterface::STATE_CANCELLED])
```
lchrusciel: Not sure if we should define it here. Method `countByCustomer` suggest that it will count all customer method. We should change a name of this method. And consider an additional parameter

pjedrzejewski: Yeah, maybe something like ``countValidOrdersByCustomer``?

lchrusciel: sounds good to me

michalmarcinkowski: That's how we end up having `createQueryBuilderForEnabledByTaxonIdAndChannelAndLocale` 😄 

I agree that current method name is too generic, but let's discuss the names on Monday and standardize them all at once, because there are many more methods that needs renaming.

pjedrzejewski:
Thanks Michał!

PR #6878, opened 2016-11-27
Use FQCNs for form types instead of aliases - sixth part

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | related to #6738 |
| License         | MIT |

The worst bundles to upgrade are done - now only **GridBundle** and **CoreBundle** are remaining for upgrade.


pjedrzejewski:
Thanks Kamil! :)

Issue #6604, opened 2016-11-01
[RFC] (root) Taxon with children taxons behavior in listing


gabiudrescu:
⚠️ 🛣 **This issue is a part of the** [Sylius Public Roadmap](https://github.com/orgs/Sylius/projects/2) 🛣 ⚠️

Preconditions:
 - I have 3 taxons like this:
   - TShirts
       - Men
       - Women
 - I have 4 products where Men taxon is assigned
 - I have 8 products where Women taxon is assigned

Steps to reproduce:
 - go to http://sylius.localhost/taxons/category/t-shirts/men and observe the 4 products 
 - go to http://sylius.localhost/taxons/category/t-shirts/women and observe 8 products
 - go to http://sylius.localhost/taxons/category/t-shirts

What happens: 
 - there is no product

What I think it should happen:
 - see all the 12 products assigned to the children taxons of T-Shirts

----

![image](https://cloud.githubusercontent.com/assets/5156054/19903052/d5ac9378-a075-11e6-81d9-d8e433d9cc66.png)

![image](https://cloud.githubusercontent.com/assets/5156054/19903072/e3893fb4-a075-11e6-9e48-f25d399c464a.png)

----

We have implemented such a behavior here, based on ElasticSearch: 
 - https://bestvalue.eu/listing/parfumuri 
 - https://bestvalue.eu/listing/parfumuri/apa-de-parfum
 - https://bestvalue.eu/listing/parfumuri/apa-de-toaleta

gabiudrescu:
@pjedrzejewski isn't this supposed to be already available for testing on http://demo.sylius.org/ ? how often do you deploy?

michalmarcinkowski:
The demo will be updated tomorrow to the 1.0.0-beta release 🎉 . After that we will deploy on each release.

teohhanhui:
I believe this was not fixed the right way. Any product listing of a parent taxon must automatically include products from all its children. It is counter-intuitive and error prone for the user to have to manually add products from child taxons directly to the parent taxon. And it's especially problematic when you have a more nested hierarchy.

teohhanhui:
And putting the `/` in the taxon slug is a ugly hack that is again very error prone. Imagine you have tens or hundreds of child taxons, and if you rename or change the slug of the parent taxon, you'll have to then make sure to update the slug for each child taxon?

pamil:
[...]

PR #6886, opened 2016-11-27
[BugFix] Adjust taxon slugs and add T-Shirts as parent taxon

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | Fixes https://github.com/Sylius/Sylius/issues/6604 |
| License         | MIT |


pjedrzejewski:
Thanks Michał!

teohhanhui:
I believe this was not fixed the right way. Any product listing of a parent taxon must automatically include products from all its children. It is counter-intuitive and error prone for the user to have to manually add products from child taxons directly to the parent taxon. And it's especially problematic when you have a more nested hierarchy.

PR #6890, opened 2016-11-28
Remove loggable

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Resources/config/doctrine/model/Address.orm.xml
+++ b/src/Sylius/Bundle/AddressingBundle/Resources/config/doctrine/model/Address.orm.xml
@@ -19,28 +19,14 @@
             <generator strategy="AUTO" />
         </id>
 
-        <field name="firstName" column="first_name" type="string">
-            <gedmo:versioned />
-        </field>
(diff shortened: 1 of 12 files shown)


Shine-neko:
Why ?

pjedrzejewski:
@Shine-neko Because we do not support this feature properly, you can easily add the same thing to your app. We will reimplement this for BETA2 release.

mickaelandrieu:
@pjedrzejewski is it worth to make a BC break between beta 1 and beta 2, because it's not perfect ?

pjedrzejewski:
@mickaelandrieu Introducing it back won't cause any major BC break. We do not display these in the admin anywhere.

michalmarcinkowski:
@mickaelandrieu it's better to put a tag on features that are fully working and add new features later when they will be implemented well and working without any issues. This feature is not well thought out, so if we leave it as it is now we will have to do BC breaks after beta anyway to fix it. If you use it now, you can easily add it back in your application.

pjedrzejewski:
Thank you Grzesiu!

PR #6893, opened 2016-11-28
Use FQCNs for form types instead of aliases - seventh part

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | related to #6738 |
| License         | MIT |


bendavies:
why have you removed the form listeners and replaced them with un testable closures?

pamil:
Most of them were bound to specific form types implementation and used only in that form type. In order to test them we should use PhpUnit instead of PhpSpec (to follow "don't mock what you don't own") and test them along with all the event subscribers and listeners like it is shown in Symfony documentation - http://symfony.com/doc/3.1/form/unit_testing.html.

bendavies:
@pamil fair enough, thanks!

pamil:
Right now we have to depend on acceptance tests made by Behat or application API tests, but I'm looking forward to write some tests specifically for form types in the Beta phase (especially for those dynamic, weird ones 🌮 ).

bendavies:
@pamil please add the symfony3 tag. I check https://github.com/Sylius/Sylius/labels/Symfony%203.0 daily 😄

michalmarcinkowski:
👍

PR #6269, opened 2016-10-03
[ThemeBundle] allow symfony 3.x

ahilles107:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | no |
| BC breaks? | no |
| Related tickets | partialy #3564 |
| License | MIT |

PR #6273, opened 2016-10-03
[Promotion] Promotion coupon validation

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | yes
| Related tickets | fixes #6166, fixes #6308
| License         | MIT

PR #6896, opened 2016-11-28
Upgrade to Symfony 3

### LINKED ISSUE #3018: YamlPersister Yaml::parse compatible with Symfony 3.0
To keep compatibility with 3.0, in your class YamlPersister:

line 28 when your parse file, use

$data = Yaml::parse(file_get_contents($his->file));

instead of:

$data = Yaml::parse($this->file);

### LINKED ISSUE #3564: [RFC] Symfony 3.0 Compatibility
Do you plan to make [sylius/resource-bundle](https://packagist.org/packages/sylius/resource-bundle) symfony 3.0 compatible?

What about these changes in composer 

```
{
    //
    "require": {
        //
        "symfony/framework-bundle": "~2.3|3.0",
        //
    }
}

```

According to the good example in [Symfony 3.0 support in FOSUserBundle](https://github.com/1ed/FOSUserBundle/blob/symfony-3-support/composer.json)

### LINKED ISSUE #5590: [ResourceBundle] ResourceFormFactory is not compatible with Symfony 2.8+
I don't know if this can be considered a bug, a feature request or neither of the two, but let me explain the problem first: I'm using the `ResourceBundle` with Symfony 2.8. From that version of Symfony, how a form is declared (and istantiated) was changed. Instead of using an alias name for each form type, the FCQN of the type class is used ([#15079](https://github.com/symfony/symfony/pull/15079)). If a FCQN is used as the form type in the resource config, the default `ResourceFormFactory` tries to instantiate the form type by doing a simple `new $className()`, and then pass the instance to the underlying form factory. This causes two problems: the first is a deprecation message triggered by the `createNamedBuilder` method that doesn't accept anymore an instance of the form type, and the second is that not using an alias there is no way to have a form that takes arguments in the constructor (e.g. injected services). My suggestion is to remove the check for the FCQN in the Sylius form factory, and always let the underyling Symfony form factory to instantiate the form by using the specified type, whatsever it is.

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | closes #6738, closes #6269, closes #5590, closes #3571, closes #3564, closes #3018 |
| License         | MIT |


composer.json
```diff
@@ -20,76 +20,77 @@
...
-        "zendframework/zend-hydrator": "^1.0|^2.0",
-        "zendframework/zend-stdlib": "^2.0|^3.0"
+        "stof/doctrine-extensions-bundle": "^1.2",
+        "swiftmailer/swiftmailer": "^5.4",
+        "symfony/monolog-bundle": "^3.0",
+        "symfony/polyfill-iconv": "^1.3",
+        "symfony/polyfill-intl-icu": "^1.3",
+        "symfony/polyfill-mbstring": "^1.3",
+        "symfony/swiftmailer-bundle": "^2.4",
+        "symfony/symfony": "^3.2@rc",
```
pamil: Stable version will be released this week, so I think it's worth to aim for Symfony 3.2

pamil: Especially that it features priority settings for compiler passes and forms types extensions.

pamil:
Btw. it's still needed to update packages' `composer.json`, but I'll take care of it in another PR (there are some issues with testing their interoperability while changing so much requirements). This one brings support for Symfony 3 on application level.

pamil:
Aaaand let's take a look at some stats (Behat running non-JS scenarios):

**PHP 7 & Symfony 2.8** 5m54.83s (678.72Mb) **vs** 4m38.20s (52.28Mb) **PHP 7 & Symfony 3.2**
**PHP 5.6 & Symfony 2.8** 13m6.17s (1.61Gb) **vs** 10m34.61s (96.04Mb) **PHP 5.6 & Symfony 3.2**

bendavies:
noticed the time reduction and thought, cool. then noticed the memory!

pamil:
Getting rid of https://github.com/symfony/symfony/pull/15945#r86547326 from our codebase might have been one of the reasons.

rvanlaak:
These performance improvements really look crazy! 🎉 https://github.com/Sylius/Sylius/pull/6896#issuecomment-263283595

pjedrzejewski:
Thank you Kamil, great work!

rvanlaak:
Nice work guys, christmas is early this year 🚀

pamil:
@rvanlaak just wait for #6910🕴 :)

rvanlaak:
Was looking at it now, no remarks. That one also impressive, think it is the biggest major version bumping PR I've ever seen :)

rvanlaak:
My early findings on upgrading from `v0.19` to latest master:

* Had to rename `@sylius.generator.promotion_coupon` to `@sylius.promotion_coupon_generator`
* Had to prepend `Promotion` to the related `Coupon` and `CouponInterface` references
* Had to make my `Order` entity implement the `ResourceInterface`
* Saw coupon validation was reworked in https://github.com/Sylius/Sylius/pull/6273, so `CouponToCodeTransformer` is no service anymore but `PromotionCouponToCodeType` is final and is the transformer itself. How can I adjust the `PromotionCouponToCodeType` to my own needs?

Would be awesome to have some notes about these things in `UPGRADE.md` 👍

pamil:
@rvanlaak we're working on updating our customisation guides :)

In order to extend a form type you only need to create [form type extension](http://symfony.com/doc/current/form/create_form_type_extension.html). You can take a look how it is done for Sylius in [`src/Sylius/Bundle/CoreBundle/Form/Extension`](https://github.com/Sylius/Sylius/tree/master/src/Sylius/Bundle/CoreBundle/Form/Extension).

rvanlaak:
Ah @pamil yes great, that's way better than extending it because that could lead to `EventListeners` that are added twice in certain cases. Had that problem before. 

Still have to get used to the Symfony 3 way of doing such stuff. Are all `Types` final now?

pamil:
I'm proposing to make them all final in #6925. If we come into any problems, final can be always dropped without any BC breaks.

ilukac:
Just to mentioned that we are entering a period with no Symfony LTS version until the 3.4 version will be released.

rvanlaak:
Symfony community once discussed whether 3.0 should be a LTS version too next to 2.8, but because of the strict BC promise upgrading minor versions to `3.4` will not require any code changes.

I've heard @pjedrzejewski about a similar LTS mechanism for Sylius in [Sound of Symfony 13](http://www.soundofsymfony.com/episode/episode-13/) so he might be able to elaborate on that.

pamil:
Anyway, there's a plenty of time between the first beta release and the stable one :)

PR #6900, opened 2016-11-28
[Behat] Fix exception msg for Toggles checkbox assertion

venyii:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

**Before:**
```
Behat\Testwork\Call\Exception\FatalThrowableError: Fatal error: Wrong parameters for RuntimeException([string $message [, long $code [, Throwable $previous = NULL]]]) in vendor/sylius/sylius/src/Sylius/Behat/Behaviour/Toggles.php:57
```

**After:**
```
RuntimeException: Toggleable element state is 'false' but expected 'true'. in vendor/sylius/sylius/src/Sylius/Behat/Behaviour/Toggles.php:57
```


michalmarcinkowski:
Thanks @venyii!

PR #6904, opened 2016-11-28
[Core] Maintain previous cart after logging in

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

This PR solves two issues:
- when we had some products in the cart (or even some checkout steps already passed) and logged out, cart was still available for the not logged in user
- after logging out with unfinished order, it was impossible to continue shopping after session expired/on different browser etc.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Context/CustomerAndChannelBasedCartContext.php
+++ b/src/Sylius/Bundle/CoreBundle/Context/CustomerAndChannelBasedCartContext.php
@@ -62,7 +62,7 @@ final class CustomerAndChannelBasedCartContext implements CartContextInterface
         try {
             $channel = $this->channelContext->getChannel();
         } catch (ChannelNotFoundException $exception) {
-            return;
+            throw new CartNotFoundException('Sylius was not able to find the cart, as there is no current channel.');
         }
 
         $customer = $this->customerContext->getCustomer();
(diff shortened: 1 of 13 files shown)


src/Sylius/Bundle/CoreBundle/Context/CustomerAndChannelBasedCartContext.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCart()
+    {
+        try {
+            $channel = $this->channelContext->getChannel();
+        } catch (ChannelNotFoundException $exception) {
+            return;
```
michalmarcinkowski: `throw new CartNotFoundException`?

src/Sylius/Bundle/CoreBundle/Context/CustomerAndChannelBasedCartContext.php
```diff
@@ -0,0 +1,80 @@
...
+        } catch (ChannelNotFoundException $exception) {
+            throw new CartNotFoundException('Sylius was not able to find the cart, as there is no current channel.');
+        }
+
+        $customer = $this->customerContext->getCustomer();
+        if (null === $customer) {
+            throw new CartNotFoundException('Sylius was not able to find the cart, as there is no logged in user.');
+        }
+
+        $cart = $this->orderRepository->findLatestCartByChannelAndCustomer($channel, $customer);
```
Zales0123: Well, sure we can!

pjedrzejewski:
Thank you Mateusz!

PR #6933, opened 2016-11-29
[Core][Promotion] Apply promotion only when any action has been applied

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | |
| License         | MIT |

### DIFF
--- a/src/Sylius/Component/Core/Promotion/Action/UnitFixedDiscountPromotionActionCommand.php
+++ b/src/Sylius/Component/Core/Promotion/Action/UnitFixedDiscountPromotionActionCommand.php
@@ -74,13 +74,12 @@ final class UnitFixedDiscountPromotionActionCommand extends UnitDiscountPromotio
 
         $channelCode = $subject->getChannel()->getCode();
         if (!isset($configuration[$channelCode])) {
-            return;
+            return false;
         }
 
         $amount = $configuration[$channelCode]['amount'];
-
         if (0 === $amount) {
-            return;
+            return false;
         }
 
         $filteredItems = $this->priceRangeFilter->filter(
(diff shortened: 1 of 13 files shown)


src/Sylius/Component/Core/Promotion/Action/UnitFixedDiscountPromotionActionCommand.php
```diff
@@ -93,6 +92,8 @@ public function execute(PromotionSubjectInterface $subject, array $configuration
         foreach ($filteredItems as $item) {
```
michalmarcinkowski: And before that loop we can do `if(empty($filteredItems)) { return false; }` and have the logic working for filters also 🎉

src/Sylius/Component/Promotion/Action/PromotionApplicator.php
```diff
@@ -38,11 +38,15 @@ public function __construct(ServiceRegistryInterface $registry)
      */
     public function apply(PromotionSubjectInterface $subject, PromotionInterface $promotion)
     {
+        $applyPromotion = false;
         foreach ($promotion->getActions() as $action) {
-            $this->getActionCommandByType($action->getType())->execute($subject, $action->getConfiguration(), $promotion);
+            $result = $this->getActionCommandByType($action->getType())->execute($subject, $action->getConfiguration(), $promotion);
+            $applyPromotion = $result ? true : false;
```
michalmarcinkowski: Should be `$applyPromotion |= $result;`

We're missing one example:
1. First action was applied
2. Second was not applied
3. The promotion should be applied

pjedrzejewski:
Thank you Mateusz!

PR #6940, opened 2016-11-29
POC - Use trans choice instead of multiple message keys

gabiudrescu:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | n/a |
| License         | MIT |

currently, adding a product to cart will render "1 items". instead of using ifs, why not using transchoice?

it looks like this:
image

image

image

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/Cart/_widget.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/Cart/_widget.html.twig
@@ -2,7 +2,11 @@
 {% set cart = sylius_cart_get() %}
 
 <div id="sylius-cart-button" class="ui circular cart button">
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
Nice! Thank you Gabi! Just in time. ;)

gabiudrescu:
Glad I could help. I haven't researched the app much, but are there many similar situations in the codebase?

michalmarcinkowski:
There are some for sure, just need to find them.

gabiudrescu:
So basically, it should be a project standard from now on, right? Transchoice everywhere?

pjedrzejewski:
@gabiudrescu I think so, only thing to figure out is a convention and how using the dot ``foo.bar`` key under ``sylius.ui.`` goes with our conventions. But it seems to work, so I think ``.choice`` suffix is good enough. 👍

PR #6941, opened 2016-11-29
[BugFix] Enable translations at default

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

With this change the translation services will be loaded always, unless intentionally disabled.


pjedrzejewski:
👍

Issue #6961, opened 2016-12-01
[Bug][Critical] Bug on dashboard after fresh installation


Zales0123:
![zrzut ekranu 2016-12-01 o 08 40 02](https://cloud.githubusercontent.com/assets/6212718/20785704/e0f7df88-b7a1-11e6-9200-ce728013dc9c.png)

Related: https://github.com/Sylius/Sylius/pull/6958

Zales0123:
Working on fixing this ASAP

Zales0123:
What was the problem with previous implementation?
```yml
sylius_shop_partial_cart_summary:
    path: /summary
    methods: [GET]
    defaults:
        _controller: sylius.controller.order:summaryAction
        _sylius:
            template: $template
            form: Sylius\Bundle\OrderBundle\Form\Type\CartType
```

pjedrzejewski:
Please add a new widgetAction to order controller that does the same as summaryAction except the form creation. Use it for the widget partial, I will explain rationale in couple of minutes. :)

michalmarcinkowski:
It works on travis because in PROD env the exception in subrequest is muted, we must change that!

PR #6958, opened 2016-12-01
Fix cart widget (broke cart summary page)

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | - |
| License         | MIT |

PR #6962, opened 2016-12-01
[Critical][Shop] Revert changes that "fixed" broken cart widget

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/6961, related to https://github.com/Sylius/Sylius/pull/6958 |
| License         | MIT |

The problem is, in ``showAction`` repository uses newly created cart as argument to ``find`` method - but it does not have an id, and that's why it fails.

Again two questions:
- what was the problem with previous implementation?
- how is it possible it does not break the build on Travis? It fails every single shop scenario on my local machine


pjedrzejewski:
Thanks Mateusz! :)

PR #6964, opened 2016-12-01
[Documentation][Customization Guide] Forms update

### LINKED ISSUE #6894: Customizing sylius_customer_registration
Is there any recipe?

I need to make Phone Number field required.

I'm trying to override it following the "Customizing Forms" topic, but it appears that things are not that simple with this form type.

The only way that I can specify something in config.yml looks like this:
```yml
sylius_user:
    resources:
        whatever: # need to specify it here, because it uses attribute name as key in array prototype configuration
            user:
                classes:
                    form:
                        registration: AppBundle\Form\Type\CustomerRegistrationType
```

It ends with `sylius.form.type.customer_user_registration` form type in `debug:container`

But routing describes `sylius_shop_register` with `sylius_customer_registration` form type…

It appears that I can't specify something for neither `sylius_customer` nor `sylius_core`.
Customer bundle doesn't have a registration form in config. Core bundle also doesn't have anything like this.

I'm out of ideas 😓

TheMadeleine:
| Q               | A
| --------------- | ---
| Doc fix?        | yes |
| New feature?    | no |
| Related tickets | changes according to #6943, fixes #6894 |
| License         | MIT |

UPDATE: **Works now!**

### DIFF
--- a/docs/customization/form.rst
+++ b/docs/customization/form.rst
@@ -1,11 +1,8 @@
 Customizing Forms
 =================
 
-The forms in Sylius are placed in the ``Sylius\Bundle\*BundleName*\Form\Type`` namespaces.
-
-.. warning::
-    Many forms in Sylius are **extended in the CoreBundle**.
-    If the form you are willing to override exists in the ``CoreBundle`` your should be extending the one from Core, not the base form from the bundle.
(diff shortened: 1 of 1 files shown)


michalmarcinkowski:
@TheMadeleine how about adding `contactHours` to `Channel`?

michalmarcinkowski:
:+1:

PR #6991, opened 2016-12-02
Change app/console to bin/console

frak:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

Change the path in the docs to match where the console command is.


pjedrzejewski:
Thanks Michael, good catch! :)

PR #7004, opened 2016-12-05
[Core][Addressing][Channel][Payment][Product] Work on factories overriding vol.1

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | partially https://github.com/Sylius/Sylius/issues/4226 |
| License         | MIT |

This is the first step to make factories overriding consistent and get rid of billions of ways to decorate/override them which we use in Sylius. In this PR I just moved factory decoration from extensions/compiler passes to xml's. So for now, in ``Configuration`` classes there is always ``Factory`` or ``TranslatableFactory`` as default, which are overrode in xml if necessary. There are 3 exceptions:
- ``AttributeFactory`` - generic bundle, probably it would stay in compiler pass for now
- ``ReviewFactory`` - the same
- ``OrderItemUnitFactory`` - requires ``OrderItem`` in constructor, I will process on this factory in further PR

We should also finally decide, is it really necessary to configure resources' factories in ``Configuration`` files, as it causes more harm than good.


pjedrzejewski:
Thank you Mateusz!

Issue #7003, opened 2016-12-05
"Create" transition is never triggered in sylius_payment state_machine


gorkalaucirica:
I want to listen to payment creations using [sylius_payment state machine](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/PaymentBundle/Resources/config/app/state_machine.yml) but "create" transition is never triggered. 

Currently the state machine also has a "cart" state that a payment never has, [payments are created with "new" state by default](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Payment/Model/Payment.php#L48) and no transition new -> cart is available.

What is the expected behavior here?

gorkalaucirica:
I found that create is called in [sylius_order state maching on order create](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine/sylius_order.yml#L45).

The issue now is that in case a payment is cancelled and a user retries the payment, no "create" is triggered

pjedrzejewski:
Default payment state should be "cart", then the transition shall be executed. Thanks for reporting!

pjedrzejewski:
Thanks for reporting Gorka, let us know if it is fixed now. :)

PR #7010, opened 2016-12-05
[Payment] Fix default payment state

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/7003 |
| License         | MIT |

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -18,7 +18,7 @@
   To use it replace name of constraint in constraint mapping file from: `Sylius\Bundle\CoreBundle\Validator\Constraints\ImageUniqueCode`
   to: `Sylius\Bundle\ResourceBundle\Validator\Constraints\UniqueWithinCollectionConstraint`
 
-* Renamed ``getLastNewPayment()`` on ``OrderInterface`` to ``getLastPayment($state)``, where ``$state`` is target last payment state.
+* Renamed ``getLastNewPayment()`` on ``OrderInterface`` to ``getLastPayment($state)``, where ``$state`` is target last payment state. Every ``getLastNewPayment()`` method should be replaced with ``getLastPayment(PaymentInterface::STATE_NEW)``.
 
 ### Currency / CurrencyBundle
 
@@ -40,7 +40,7 @@
 
 ### Payment / PaymentBundle
 
-* Changed default ``Payment`` ``$state`` from *new* to *cart*.
+* Changed default ``Payment::$state`` from *new* to *cart*.
 
 ### PayumBundle
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -18,7 +18,7 @@
   To use it replace name of constraint in constraint mapping file from: `Sylius\Bundle\CoreBundle\Validator\Constraints\ImageUniqueCode`
   to: `Sylius\Bundle\ResourceBundle\Validator\Constraints\UniqueWithinCollectionConstraint`
 
-* Renamed ``getLastNewPayment()`` on ``OrderInterface`` to ``getLastCartPayment()`` due to ``Payment`` default state change.
+* Renamed ``getLastNewPayment()`` on ``OrderInterface`` to ``getLastPayment($state)``, where ``$state`` is target last payment state.
 
 ### Currency / CurrencyBundle
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -18,6 +18,8 @@
   To use it replace name of constraint in constraint mapping file from: `Sylius\Bundle\CoreBundle\Validator\Constraints\ImageUniqueCode`
   to: `Sylius\Bundle\ResourceBundle\Validator\Constraints\UniqueWithinCollectionConstraint`
 
+* Renamed ``getLastNewPayment()`` on ``OrderInterface`` to ``getLastCartPayment()`` due to ``Payment`` default state change.
+
 ### Currency / CurrencyBundle
 
 ### Customer / CustomerBundle
@@ -37,6 +39,8 @@
 ### Order / OrderBundle
 
 ### Payment / PaymentBundle
+
+* Changed default ``Payment`` ``$state`` from *new* to *cart*.
 
 ### PayumBundle
 
--- a/src/Sylius/Behat/Context/Setup/OrderContext.php
+++ b/src/Sylius/Behat/Context/Setup/OrderContext.php
@@ -698,7 +698,7 @@ final class OrderContext implements Context
         }
         $this->applyTransitionOnOrderCheckout($order, OrderCheckoutTransitions::TRANSITION_SELECT_SHIPPING);
 
-        $payment = $order->getLastNewPayment();
+        $payment = $order->getLastCartPayment();
         $payment->setMethod($paymentMethod);
 
         $this->applyTransitionOnOrderCheckout($order, OrderCheckoutTransitions::TRANSITION_SELECT_PAYMENT);
--- a/src/Sylius/Behat/Context/Setup/OrderContext.php
+++ b/src/Sylius/Behat/Context/Setup/OrderContext.php
@@ -698,7 +698,7 @@ final class OrderContext implements Context
         }
         $this->applyTransitionOnOrderCheckout($order, OrderCheckoutTransitions::TRANSITION_SELECT_SHIPPING);
 
-        $payment = $order->getLastCartPayment();
+        $payment = $order->getLastPayment(PaymentInterface::STATE_CART);
         $payment->setMethod($paymentMethod);
 
         $this->applyTransitionOnOrderCheckout($order, OrderCheckoutTransitions::TRANSITION_SELECT_PAYMENT);
--- a/src/Sylius/Bundle/PayumBundle/Controller/PayumController.php
+++ b/src/Sylius/Bundle/PayumBundle/Controller/PayumController.php
@@ -20,6 +20,7 @@ use Sylius\Bundle\PayumBundle\Request\ResolveNextRoute;
 use Sylius\Bundle\ResourceBundle\Controller\FlashHelperInterface;
 use Sylius\Bundle\ResourceBundle\Controller\RequestConfigurationFactoryInterface;
 use Sylius\Bundle\ResourceBundle\Controller\ViewHandlerInterface;
+use Sylius\Component\Core\Model\OrderInterface;
 use Sylius\Component\Order\Repository\OrderRepositoryInterface;
 use Sylius\Component\Payment\Model\PaymentInterface;
 use Sylius\Component\Resource\Metadata\MetadataInterface;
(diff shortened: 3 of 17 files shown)


src/Sylius/Component/Core/Model/Order.php
```diff
@@ -302,12 +302,20 @@ public function hasPayment(BasePaymentInterface $payment)
      */
     public function getLastNewPayment()
     {
+        return $this->getLastCartPayment();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getLastCartPayment()
```
michalmarcinkowski: We shouldn't add this method, even better, we should get rid of `getLastNewPayment()` also if possible.

src/Sylius/Component/Core/OrderProcessing/OrderPaymentProcessor.php
```diff
@@ -62,7 +62,7 @@ public function process(BaseOrderInterface $order)
             return;
         }
 
-        $newPayment = $order->getLastNewPayment();
+        $newPayment = $order->getLastCartPayment();
```
michalmarcinkowski: The logic is not good here... after the order is placed, we should no longer create payments in `cart` state, this needs to be well-thought

Zales0123: But is ``OrderPaymentProcessor`` called after placing order? For now, it would be called on each checkout step (which is the same as we had it before) and it's also called on ``fail`` and ``cancel`` transition on order payment.

src/Sylius/Component/Core/OrderProcessing/OrderPaymentProcessor.php
```diff
@@ -93,6 +104,12 @@ private function createNewPayment(BaseOrderInterface $order)
         }
 
         $payment->setMethod($paymentMethod);
+
+        if (OrderCheckoutStates::STATE_COMPLETED === $order->getCheckoutState()) {
+            $stateMachine = $this->stateMachineFactory->get($payment, 'sylius_payment');
```
gorkalaucirica: Maybe we can replace `sylius_payment` with `PaymentTransitions::GRAPH`

src/Sylius/Component/Core/OrderProcessing/OrderPaymentProcessor.php
michalmarcinkowski: `fail` and `cancel` transition on payment is triggered after the order was placed...

Zales0123: Yup 😄  and that's why if orders is ``completed``, payment has already ``create`` transition applied. This is of course a temporary solution - we should have two processors, for checkout and fail/cancel callbacks. Or should I implement them in this PR?

src/Sylius/Component/Core/OrderProcessing/OrderPaymentProcessor.php
pjedrzejewski: This should not be done here, there is already a cascade transition on "complete" transition IIRC, can you check @Zales0123?

UPGRADE.md
```diff
@@ -0,0 +1,12 @@
+UPGRADE
+=======
+
+## From 1.0.0-beta.1 to 1.0.0-beta.2
+
+### Core and CoreBundle
+
+* Renamed ``getLastNewPayment()`` on ``OrderInterface`` to ``getLastCartPayment()`` due to ``Payment`` default state change.
```
pjedrzejewski: The method is still there.

src/Sylius/Component/Core/Model/Order.php
```diff
@@ -298,16 +298,22 @@ public function hasPayment(BasePaymentInterface $payment)
...
+    public function getLastPayment($state = null)
     {
         if ($this->payments->isEmpty()) {
             return null;
         }
 
-        $payment = $this->payments->filter(function (BasePaymentInterface $payment) {
-            return $payment->getState() === BasePaymentInterface::STATE_NEW;
+        if (null === $state) {
+            return $this->payments->last();
```
michalmarcinkowski: Can return `false`

UPGRADE.md
```diff
@@ -10,6 +10,8 @@
 
 ### Core / CoreBundle
 
+* Renamed ``getLastNewPayment()`` on ``OrderInterface`` to ``getLastPayment($state)``, where ``$state`` is target last payment state.
```
pamil: You can also add how exactly can I replace `getLastNewPayment()` in my userland code

UPGRADE.md
```diff
@@ -26,6 +28,8 @@
 
 ### Payment / PaymentBundle
 
+* Changed default ``Payment`` ``$state`` from *new* to *cart*.
```
pamil: ``Payment::$state``?

src/Sylius/Behat/Context/Setup/OrderContext.php
```diff
@@ -698,7 +698,7 @@ private function proceedSelectingShippingAndPaymentMethod(OrderInterface $order,
         }
         $this->applyTransitionOnOrderCheckout($order, OrderCheckoutTransitions::TRANSITION_SELECT_SHIPPING);
 
-        $payment = $order->getLastNewPayment();
+        $payment = $order->getLastPayment(PaymentInterface::STATE_CART);
```
pamil: Hmm, why we're using `PaymentInterface::STATE_CART` here?

src/Sylius/Bundle/PayumBundle/Controller/PayumController.php
```diff
@@ -97,7 +99,7 @@ public function prepareCaptureAction(Request $request, $tokenValue)
         $request->getSession()->set('sylius_order_id', $order->getId());
         $options = $configuration->getParameters()->get('redirect');
 
-        $payment = $order->getLastNewPayment();
+        $payment = $order->getLastPayment(PaymentInterface::STATE_NEW);
```
pamil: And `PaymentInterface::STATE_NEW` here?

src/Sylius/Component/Core/Model/Order.php
```diff
@@ -298,16 +298,27 @@ public function hasPayment(BasePaymentInterface $payment)
...
+            $lastPayment = $this->payments->last();
+            if (false === $lastPayment) {
+                return null;
+            }
+
+            return $lastPayment;
+        }
+
+        $payment = $this->payments->filter(function (BasePaymentInterface $payment) use ($state) {
+            return $payment->getState() === $state;
```
pamil: `return null === $state || $payment->getState() === $state;` and the whole previous `if` can be deleted

src/Sylius/Component/Core/Model/Order.php
```diff
@@ -298,16 +298,27 @@ public function hasPayment(BasePaymentInterface $payment)
...
         })->last();
 
         return $payment !== false ? $payment : null;
```
pamil: We've already checked that payments aren't empty, so last element will never be `false`.

src/Sylius/Component/Core/Payment/Provider/OrderPaymentProviderInterface.php
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface OrderPaymentProviderInterface
+{
+    /**
+     * @param OrderInterface $order
+     * @param string $targetState
+     *
+     * @return PaymentInterface
```
pamil: `PaymentInterface|null` actually

src/Sylius/Component/Core/Payment/Provider/OrderPaymentProvider.php
```diff
@@ -0,0 +1,140 @@
...
+
+        $paymentMethod = $this->getDefaultPaymentMethod($payment, $order);
+        $lastPayment = $this->getLastPayment($order);
+
+        if (null !== $lastPayment) {
+            $paymentMethod = $lastPayment->getMethod();
+        }
+
+        if (null === $paymentMethod) {
+            return null;
```
pamil: We should throw an exception, if I run `$provider->provideOrderPayment(...)` then I do not expect `null`.

src/Sylius/Bundle/PayumBundle/Controller/PayumController.php
Zales0123: As it has been done previously, each operation connected with payment is done on last payment of order. Previously it was always ``NEW`` payment, as it was a default (wrong) state. Now, the default state is ``CART`` and in checkout we should always operate on last ``CART`` payment. However, when checkout is over, we then always operate on ``NEW`` payment - this is required state for payum and also payments with this state are created after previous payment fail/cancel.

pamil: It's something that definitely should go to `UPGRADE.MD`.

UPGRADE.md
```diff
@@ -10,6 +10,8 @@
 
 ### Core / CoreBundle
 
+* Renamed ``getLastNewPayment()`` on ``OrderInterface`` to ``getLastPayment($state)``, where ``$state`` is target last payment state. Every ``getLastNewPayment()`` method should be replaced with ``getLastNewPayment(PaymentInterface::STATE_NEW)``.
```
pjedrzejewski: getLastPayment(PaymentInterface::STATE_NEW)

Zales0123: damn it 😄

pjedrzejewski:
I am not sure I understand the idea behind "OrderPaymentProvider" service, perhaps this is just naming issue, but could you elaborate a bit about these changes?

Zales0123:
Of course ;)
Previous implementation of ``OrderPaymentProcessor`` made it responsible of:
- getting last **new** payment of order and setting proper currency code and amount on it
- if there is no new payment creating new order payment
    - this payment method was set the same as last cancelled payment
    - if there is no last cancelled payment it was set the same as last failed payment
    - if there was no last failed payment method was provided from ``DefaultPaymentMethodResolver``
- new payment was also adding to order

Because of payment default state, we should also add some logic here to determine, if new payment should be created with **cart** state (for checkout) or immediately proceeded to **new** state (for after-checkout payment fail/cancel callbacks).
It was just too much, so I extracted some logic to ``OrderPaymentProvider`` - we can now be sure in ``OrderPaymentProcessor`` that new payment would be created with good method and good state and add it to the order if it has no new payment at all. Now the only thing ``OrderPaymentProcessor`` does is processing last **new** payment of order or new payment provided from provider.

pjedrzejewski:
Thank you Mateusz! :)

PR #7013, opened 2016-12-05
[UnpaidOrdersStateUpdater] Use class constant

tchapi:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

Just a small fix — using a defined class constant instead of a string


michalmarcinkowski:
Thanks! 👍

PR #7008, opened 2016-12-05
[Documentation][Customization Guide] Grids

TheMadeleine:
| Q               | A
| --------------- | ---
| Doc fix?        | no |
| New docs?    | yes |
| Related tickets | |
| License         | MIT |

Before having a full customization guide to grids we need to fix that behaviour of variableNodes in yaml configs -> #6637 .

PR #7015, opened 2016-12-06
[Grid] Field, Action & Filter position in Grids

TheMadeleine:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | related to #7008 |
| License         | MIT |

By changing the value of `position` on field/filter/action we can organize them in a grid. They are now rendered sorted by position ASC. **Position defaults to `100`** to ensure that default will be rendered after the configured.


src/Sylius/Component/Grid/spec/Definition/FieldSpec.php
```diff
@@ -91,4 +91,12 @@ function it_can_have_options()
         $this->setOptions(['template' => 'SyliusUiBundle:Grid/Field:_status.html.twig']);
         $this->getOptions()->shouldReturn(['template' => 'SyliusUiBundle:Grid/Field:_status.html.twig']);
     }
+
+    function it_has_position_no_position_by_default()
```
tuka217: `it_has_no_position_by_default`. And there should be two methods: `no_position_by_default` and `mutable_position`

src/Sylius/Bundle/UiBundle/Resources/views/Macro/table.html.twig
```diff
@@ -1,7 +1,7 @@
 {% macro headers(grid, definition, requestAttributes) %}
     {% import 'SyliusUiBundle:Macro:sorting.html.twig' as sorting %}
 
-    {% for field in definition.fields %}
+    {% for field in definition.fields|sort_by('position', 'DESC') %}
```
pjedrzejewski: I think it should be ASC, hm?

TheMadeleine: Then what value should be default on the Field model?

src/Sylius/Component/Grid/spec/Definition/FieldSpec.php
TheMadeleine: Yeah, moths of docs writing and you forget how to spec. 😞

tuka217: :(

src/Sylius/Bundle/UiBundle/Resources/views/Macro/table.html.twig
pjedrzejewski: 0 is OK.

src/Sylius/Bundle/UiBundle/Resources/views/Macro/table.html.twig
```diff
@@ -1,7 +1,7 @@
 {% macro headers(grid, definition, requestAttributes) %}
     {% import 'SyliusUiBundle:Macro:sorting.html.twig' as sorting %}
 
-    {% for field in definition.fields %}
+    {% for field in definition.fields|sort_by('position', 'ASC') %}
```
GSadee: By default is `ASC`, you can remove second parameters

TheMadeleine: That's true!

pjedrzejewski:
Thanks Magda!

PR #7026, opened 2016-12-06
[UI][JS] Make attribute choice listener generic

sweoggy:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | n/a |
| License         | MIT |

So that it now also works on other attributes than product attribute. Such as when enabling attributes for product variants.

Without this PR, no attributes will be returned by the 
`$attributes = $form->getData();`
in the "renderAttributeValueFormsAction" function of an attribute enabled model.

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-product-attributes.js
+++ b/src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-product-attributes.js
@@ -66,14 +66,16 @@
     }
 
     function setAttributeChoiceListener() {
-        $('#attributeChoice button').on('click', function(event) {
+        var $attributeChoice = $('#attributeChoice');
+        $attributeChoice.find('button').on('click', function(event) {
             var $attributesContainer = $('#attributesContainer');
             event.preventDefault();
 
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
👍  I have restarted the failing build.

pjedrzejewski:
Thank you Oscar!

gabiudrescu:
do you plan to link product attributes with product variants?

if so, I'd be (very) interested in to know why and how.

pjedrzejewski:
@gabiudrescu Actually, you would use another set of attributes of variants. AttributeBundle is generic, it can attach attr to any entity. We do not support this in core, but yeah, you can totally customize it. We hope to document this soon, but you can have a look already at how products implement that.

gabiudrescu:
@pjedrzejewski  thanks. now I understand how. 

but I'd be interested to know more about the useCase. why do you want to do that. how does this impacts the standard implementation where attributes are linked to products?

should some attributes be related to variants and other related to products?

Whenever I think of attributes of a variant, I think of options - the term used in Sylius. options or attributes that create variation.

PR #7033, opened 2016-12-08
[Shop] Custom 404 error page

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

zrzut ekranu 2016-12-08 o 09 01 14

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/error404.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/error404.html.twig
@@ -1,0 +1,10 @@
+{% extends '@SyliusShop/layout.html.twig' %}
+
+{% block content %}
+    <h2 class="ui center aligned icon header">
+        <img src="{{ asset('assets/shop/img/logo.png') }}" alt="Sylius logo" class="ui small image" />
+        <br/><br/>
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/ShopBundle/Resources/views/error404.html.twig
```diff
@@ -0,0 +1,13 @@
+{% extends '@SyliusShop/layout.html.twig' %}
+
+{% block content %}
+    <h2 class="ui center aligned icon header">
+        <img src="{{ asset('assets/shop/img/logo.png') }}" alt="Sylius logo" class="ui small image" />
+        <br/><br/>
+        {{ 'sylius.ui.page_not_found.header'|trans }}
```
pjedrzejewski: Why this new convention for translation keys ... ?

``sylius.ui.not_found`` will be enough. As for text, I'd suggest something a bit more neutral: ``sylius.ui.the_page_you_are_looking_for_does_not_exist`` and that's it.

pjedrzejewski:
Thank you Mateusz!

pjedrzejewski:
Could you also update Sylius-Standard with that?

PR #7034, opened 2016-12-08
Delete sylius-attribute as it does not look to be used anymore

crevillo:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no|
| Related tickets | |
| License         | MIT |

This js file doesn't look to be used anywere, so i propose to delete it to avoid noise with the `sylius-product-attribute` one. 

Please discard this if the file can't be deleted.


pjedrzejewski:
Thank you Carlos!

PR #7038, opened 2016-12-08
[Grid][Admin][Order] Entity filter on orders' list

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/grids/order.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/grids/order.yml
@@ -85,6 +85,11 @@ sylius_grid:
                     label: sylius.ui.date
                     options:
                         field: checkoutCompletedAt
+                channel:
+                    type: entity
+                    label: sylius.ui.channel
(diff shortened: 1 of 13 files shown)


src/Sylius/Bundle/AdminBundle/Resources/config/grids/order.yml
```diff
@@ -85,6 +85,11 @@ sylius_grid:
                     label: sylius.ui.date
                     options:
                         field: checkoutCompletedAt
+                channel:
+                    type: entity
+                    label: sylius.ui.channel
+                    options:
+                        class: Sylius\Component\Core\Model\Channel
```
pjedrzejewski: We should use a parameter here ``%sylius.model.channel.class%``, this way it will still work after someone overrides channel model. ;)

pjedrzejewski:
Thank you Grzesiu! 👍

PR #7054, opened 2016-12-09
[Core] Refactor default payment method resolver

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | |
| License         | MIT |

`DefaultPaymentMethodResolver` from *Payment* component is not used in *Sylius*, better - it's not even registered. Is there any reason to share interface between this class and `DefaultPaymentMethodResolver` from *Core*? It only forces us to do some weird stuff like passing `Payment` to resolver, necessarily with set `Order` to get channel from it rather than just passing an `Order` object.

### DIFF
--- a/src/Sylius/Component/Core/Resolver/DefaultPaymentMethodResolverInterface.php
+++ b/src/Sylius/Component/Core/Resolver/DefaultPaymentMethodResolverInterface.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Core\Resolver;
+
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Core\Model\PaymentMethodInterface;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface DefaultPaymentMethodResolverInterface
+{
+    /**
+     * @param ChannelInterface $channel
+     *
+     * @return PaymentMethodInterface
+     */
+    public function getDefaultPaymentMethodByChannel(ChannelInterface $channel);
+}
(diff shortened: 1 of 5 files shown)


src/Sylius/Component/Core/Resolver/DefaultPaymentMethodResolverInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface DefaultPaymentMethodResolverInterface
+{
+    /**
+     * @param ChannelInterface $channel
+     *
+     * @return PaymentMethodInterface
+     */
+    public function getDefaultPaymentMethodByChannel(ChannelInterface $channel);
```
pamil: `getForChannel(ChannelInterface $channel)`?

src/Sylius/Component/Core/Resolver/DefaultPaymentMethodResolverInterface.php
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface DefaultPaymentMethodResolverInterface
+{
+    /**
+     * @param ChannelInterface $channel
+     *
+     * @return PaymentMethodInterface
```
pamil: + `@throws Unreso...`

pjedrzejewski:
ByChannel sounds like a concrete implementation, I preferred the previous interface, which was more generic. You may want to resolve payment methods based on something else than channel. (from order)

michalmarcinkowski:
Agree with @pjedrzejewski. The concept of this resolver is the following: "give me a default payment method for this payment" and in the specific implementation we can resolve the payment method based on whatever we want. The same concept works for shipping and provide good abstraction to any requirements.

Zales0123:
Ok, I just understood this as "give me default payment method" and that's it - that's why it was weird to me to pass payment to this method, as it looked for like some workaround ;) Closing then.

PR #7057, opened 2016-12-09
[Core][Fixture] Promotions fixtures

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |


pjedrzejewski:
I don't like that these are hardcoded to the actions and rules provided by core. We could add much more flexiblity by allowing to configure any type of action/rule with any configuration. The helper factory methods are only doing setType, setConfiguration... You know what I mean? :)

GSadee:
@pjedrzejewski Yes, I know what you mean. I wanted that configuration to be simpler and in some way more mistake-proof, but of course, you are right that we are limited by this approach. I will change it 👍

src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
```diff
@@ -150,6 +150,36 @@ sylius_fixtures:
...
+                                starts_at: "now"
+                                ends_at: "7 day"
+                                channels:
+                                    - "US_WEB"
+                                rules:
+                                    -
+                                        type: "item_total"
+                                        configuration:
+                                            US_WEB:
+                                                amount: 100
```
michalmarcinkowski: Maybe we should write `100.00` to show that the value is not integer and means `$100`?

src/Sylius/Bundle/CoreBundle/Fixture/Factory/PromotionRuleExampleFactory.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver
+            ->setDefault('type', CartQuantityRuleChecker::TYPE)
+            ->setAllowedValues('type', [
```
pjedrzejewski: This effectively locks us to built-in rules, same as for actions. I'd love to use these fixtures to configure my custom promotions too. (as we do not have validation of the config, this is not a benefit at all)

src/Sylius/Bundle/CoreBundle/Fixture/Factory/PromotionActionExampleFactory.php
```diff
@@ -0,0 +1,102 @@
...
+        return $promotionAction;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver
+            ->setDefault('type', PercentageDiscountPromotionActionCommand::TYPE)
```
Zales0123: Why this is a default type? In my opinion the **most basic** type of promotion action is ``FixedDiscountPromotionActionCommand``.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/PromotionExampleFactory.php
```diff
@@ -0,0 +1,166 @@
...
+            ->setAllowedTypes('starts_at', ['null', 'string'])
+            ->setDefault('ends_at', null)
+            ->setAllowedTypes('ends_at', ['null', 'string'])
+            ->setDefault('channels', LazyOption::all($this->channelRepository))
+            ->setAllowedTypes('channels', 'array')
+            ->setNormalizer('channels', LazyOption::findBy($this->channelRepository, 'code'))
+            ->setDefined('rules')
+            ->setNormalizer('rules', function (Options $options, array $rules) {
+                if (empty($rules)) {
+                    return [0 => []];
```
Zales0123: ``return [[]];`` would be sufficient.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/PromotionExampleFactory.php
```diff
@@ -0,0 +1,166 @@
...
+                }
+
+                return $rules;
+            })
+            ->setDefined('actions')
+            ->setNormalizer('actions', function (Options $options, array $actions) {
+                if (empty($actions)) {
+                    return [0 => []];
```
Zales0123: Same here.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/PromotionActionExampleFactory.php
GSadee: Yup, but this type is simpler in configuration because it is independent of the channel.

pjedrzejewski:
Thank you Grzesiu!

PR #7061, opened 2016-12-11
[RFC] [BC Break] [Addressing] Zone matcher refactor

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |

Small refactoring of ZoneMatcher as too many things happen there.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,9 +4,10 @@
 
 ### Addressing / AddressingBundle
 
-* `Sylius\Component\Addressing\Matcher\ZoneMatcher` has been made final, and you can no longer extend this class. Also, `Sylius\Component\Addressing\Resolver\AddressZoneResolver` service has been introduced.
- If one has been extending a `Sylius\Component\Addressing\Matcher\ZoneMatcher` class please use appropriate pattern or provide your custom implementation. 
- `Sylius\Component\Addressing\Resolver\AddressZoneResolver` should make it easier to provide custom implementation with less overhead. 
+* `Sylius\Component\Addressing\Matcher\ZoneMatcher` has been renamed to `Sylius\Component\Addressing\Resolver\ZoneResolver` made final, so you can no longer extend this class. 
+Also, `Sylius\Component\Addressing\Matcher\AddressZoneMatcher` service has been introduced.
+ If one has been extending a `Sylius\Component\Addressing\Resolver\ZoneResolver` class please use appropriate pattern or provide your custom implementation. 
+ `Sylius\Component\Addressing\Matcher\AddressZoneMatcher` should make it easier to provide custom implementation with less overhead. 
  If the class has been overridden because of zones priority order, then no overriding would be required and priority can be defined in a yaml file as follows:
  ```yml
     sylius_addressing:
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,6 +4,15 @@
 
 ### Addressing / AddressingBundle
 
+* `Sylius\Component\Addressing\Matcher\ZoneMatcher` has been made final, and you can no longer extend this class. Also, `Sylius\Component\Addressing\Resolver\AddressZoneResolver` service has been introduced.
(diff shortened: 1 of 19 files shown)


src/Sylius/Component/Addressing/spec/Resolver/AddressZoneResolverSpec.php
```diff
@@ -0,0 +1,106 @@
...
+    {
+        $this->beConstructedWith($repository);
+    }
+
+    function it_is_initializable()
+    {
+        $this->shouldHaveType(AddressZoneResolver::class);
+    }
+
+    function it_implements_zone_matcher_interface()
```
Zales0123: I don't think so 😄

src/Sylius/Component/Addressing/spec/Resolver/AddressZoneResolverSpec.php
```diff
@@ -0,0 +1,106 @@
...
+use Sylius\Component\Addressing\Model\CountryInterface;
+use Sylius\Component\Addressing\Model\ProvinceInterface;
+use Sylius\Component\Addressing\Model\ZoneInterface;
+use Sylius\Component\Addressing\Model\ZoneMemberInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+/**
+ * @author Saša Stamenković <umpirsky@gmail.com>
+ * @author Gonzalo Vilaseca <gvilaseca@reiss.co.uk>
+ * @author Jan Góralski <jan.goralski@lakion.com>
```
Zales0123: Why are you not here?

michalmarcinkowski:
I think we can get rid of this nasty PHP code by using a better database query, have you tried this approach?

src/Sylius/Component/Addressing/Matcher/ZoneMatcher.php
```diff
@@ -54,7 +63,7 @@ public function match(AddressInterface $address, $scope = null)
 
         /* @var ZoneInterface $zone */
         foreach ($availableZones = $this->getZones($scope) as $zone) {
```
pamil: `$availableZones = $this->getZones($scope) as $zone` say whaaaaat

lchrusciel: hey makarena 💃

UPGRADE.md
```diff
@@ -4,6 +4,15 @@
 
 ### Addressing / AddressingBundle
 
+* `Sylius\Component\Addressing\Matcher\ZoneMatcher` has been made final, and you can no longer extend this class. Also, `Sylius\Component\Addressing\Resolver\AddressZoneResolver` service has been introduced.
+ If one has been extending a `Sylius\Component\Addressing\Matcher\ZoneMatcher` class please use appropriate pattern or provide your custom implementation. 
+ `Sylius\Component\Addressing\Resolver\AddressZoneResolver` should make it easier to provide custom implementation with less overhead. 
+ If class has been override because of zones priority order, than no overriding would be required and priority can be defined in a yaml file gias follows:
```
lchrusciel: thanks!

michalmarcinkowski:
Doing a BC break just to make class smaller by 50 lines of code is not worth it, especially if we need to also rename the class. It's pretty late for such refactoring and no major benefit.

lchrusciel:
It is not just 50 lines smaller class, but classes which have been made final, to make our lives easier later in a later phase. Each class like this will bound our hands. Also this class violates SOLID principles as it does not do only one thing. Only reason, why this class is still not final is to have an access to `$priorities` variable. 
1. Benefits: 
- final classes with private properties and methods
- smaller class
- decoupled logic
- single responsible principle is kept
2. Tradeoff:
- bc break

I just though that it is still suitable to do such a changes. Nevertheless, closing than.

PR #6936, opened 2016-11-29
[Resource] Auto complete choice type

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| License         | MIT |

This autocomplete is initialized on `MainTaxon` and `ProductTaxons` in `ProductTypeExtension`. Rest of initialization was introduced by #7070.

PR #7070, opened 2016-12-13
[Promotion] Use one autocomplete type to rule them all

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| License         | MIT |

Resource auto complete configuration reference:
```php
->add('products', ResourceAutocompleteChoiceType::class, [
                'label' => 'sylius.form.promotion_filter.products',
                'multiple' => true,
                'required' => false,
                'choice_value' => 'id',
                'choice_name' => 'name',
                'resource' => 'sylius.product',
])
```
```twig
{{form(form, { 'remote_url': path('sylius_admin_ajax_product_index'), 'remote_criteria_type': 'contains', 'remote_criteria_name': 'search' })}}
```
And that is it generic javascript will catch this and do the rest
Blocked by #6936

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig
+++ b/src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig
@@ -196,7 +196,11 @@
     {{- form_label(form) -}}
     <div
             class="sylius-autocomplete ui fluid search selection dropdown {% if multiple %}multiple{% endif %}"
-            data-url="{{ remote_url }}"
+            {% if remote_url is defined %}
+                data-url="{{ remote_url }}"
+            {% elseif remote_route is defined %}
+                data-url="{{ path(remote_route, remote_route_parameters) }}"
+            {% endif %}
             data-choice-name="{{ choice_name }}"
             data-choice-value="{{ choice_value }}"
             data-criteria-type="{{ remote_criteria_type }}"
--- a/src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
+++ b/src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
@@ -50,5 +50,11 @@
 
         $(document).productSlugGenerator();
         $(document).taxonSlugGenerator();
+
+        $(document).on('collection-form-update', function (event, choice) {
+            $.each($('.sylius-autocomplete'), function (index, element) {
+                $(element).autoComplete();
+            });
+        });
     });
 })(jQuery);
--- a/src/Sylius/Bundle/CoreBundle/Form/Type/Promotion/Filter/ProductFilterConfigurationType.php
+++ b/src/Sylius/Bundle/CoreBundle/Form/Type/Promotion/Filter/ProductFilterConfigurationType.php
@@ -11,8 +11,7 @@
 
 namespace Sylius\Bundle\CoreBundle\Form\Type\Promotion\Filter;
 
-use Sylius\Bundle\ProductBundle\Form\Type\ProductChoiceType;
-use Sylius\Component\Core\Repository\ProductRepositoryInterface;
+use Sylius\Bundle\ResourceBundle\Form\Type\ResourceAutocompleteChoiceType;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\DataTransformerInterface;
 use Symfony\Component\Form\FormBuilderInterface;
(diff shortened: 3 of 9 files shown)


src/Sylius/Bundle/ResourceBundle/Form/DataTransformer/ResourceToIdentifierTransformer.php
```diff
@@ -49,7 +49,7 @@ public function __construct(RepositoryInterface $repository, $identifier = 'id')
     public function transform($value)
     {
         if (null === $value) {
-            return '';
+            return null;
```
michalmarcinkowski: Are you sure about that change? I recall that according to [symfony docs](http://symfony.com/doc/current/form/data_transformers.html#creating-the-transformer) we should return empty string here.

src/Sylius/Bundle/CoreBundle/Form/DataTransformer/ProductTaxonToTaxonTransformer.php
```diff
@@ -0,0 +1,68 @@
...
+        $this->productTaxonFactory = $productTaxonFactory;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function transform($productTaxon)
+    {
+        if (null === $productTaxon) {
+            return null;
```
michalmarcinkowski: `return '';`?

src/Sylius/Bundle/ResourceBundle/Form/DataTransformer/ResourceToIdentifierTransformer.php
Arminek: This is ``ResourceToIdentifier`` IMO it should not change type.

src/Sylius/Bundle/CoreBundle/spec/Form/DataTransformer/ProductTaxonToTaxonTransformerSpec.php
```diff
@@ -0,0 +1,76 @@
...
+    function it_transforms_product_taxon_to_taxon(
+        ProductTaxonInterface $productTaxon,
+        TaxonInterface $taxon
+    ) {
+        $productTaxon->getTaxon()->willReturn($taxon);
+
+        $this->transform($productTaxon)->shouldReturn($taxon);
+    }
+
+    function it_returns_null_during_transform()
```
Zales0123: ``it_returns_null_during_transforming_null_value``? As it seems like it does it every time 😄

src/Sylius/Bundle/CoreBundle/spec/Form/DataTransformer/ProductTaxonToTaxonTransformerSpec.php
```diff
@@ -0,0 +1,76 @@
...
+        TaxonInterface $taxon,
+        FactoryInterface $productTaxonFactory
+    ) {
+        $productTaxonFactory->createNew()->willReturn($productTaxon);
+        $productTaxon->setTaxon($taxon)->shouldBeCalled();
+
+        $this->reverseTransform($taxon)->shouldReturn($productTaxon);
+    }
+
+    function it_returns_null_during_reverse_transform()
```
Zales0123: ``it_returns_null_during_reverse_transforming_of_null_value``

src/Sylius/Bundle/CoreBundle/spec/Form/DataTransformer/ProductTaxonToTaxonTransformerSpec.php
```diff
@@ -0,0 +1,76 @@
...
+    {
+        $this->reverseTransform(null)->shouldReturn(null);
+    }
+
+    function it_throws_invalid_argument_exception_during_transforms()
```
Zales0123: ``it_throws_invalid_argument_exception_during_transformation_if_passed_value_is_invalid``

src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig
```diff
@@ -191,3 +191,20 @@
...
+{% block sylius_resource_autocomplete_choice_row %}
+    {{- form_label(form) -}}
+    <div
+            class="sylius-autocomplete ui fluid search selection dropdown {% if form.vars.multiple %}multiple{% endif %}"
+            data-url="{{ path(form.vars.remote_route) }}"
+            data-choice-name="{{ form.vars.choice_name }}"
+            data-choice-value="{{ form.vars.choice_value }}"
+            data-criteria-type="{{ form.vars.remote_criteria_type }}"
+            data-criteria-name="{{ form.vars.remote_criteria_name }}"
+    >
```
Zales0123: Formatting doesn't look good here 😕

src/Sylius/Bundle/ResourceBundle/Form/DataTransformer/ResourceToIdentifierTransformer.php
michalmarcinkowski: I don't like it also, but remember that Symfony recommended that.

From `DataTransformerInterface`:
```
     * This method must be able to deal with empty values. Usually this will
     * be NULL, but depending on your implementation other empty values are
     * possible as well (such as empty strings). The reasoning behind this is
     * that value transformers must be chainable. If the transform() method
     * of the first value transformer outputs NULL, the second value transformer
     * must be able to process that value.
     *
     * By convention, transform() should return an empty string if NULL is
     * passed.
```

By convention it should return empty string, but not sure why 😄

pamil: There are some exceptions even in Symfony codebase:
 - `BooleanToStringTransformer`
 - `ChoiceToValueTransformer`

I guess the convention was made to overcome this bug: https://github.com/symfony/symfony/pull/18357

PR #7073, opened 2016-12-13
[Admin][ExchangeRate] Filtering exchange rates by a specific currency

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/grids/exchange_rate.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/grids/exchange_rate.yml
@@ -28,6 +28,15 @@ sylius_grid:
                     type: string
                     label: sylius.ui.ratio
                     sortable: ~
+            filters:
+                currency:
+                    type: entity
+                    label: sylius.ui.currency
+                    options:
+                        class: "%sylius.model.currency.class%"
+                        choiceLabel: name
+                        fields: [sourceCurrency, targetCurrency]
+
             actions:
                 main:
                     create:
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/grids/exchange_rate.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/grids/exchange_rate.yml
@@ -34,7 +34,7 @@ sylius_grid:
                     label: sylius.ui.currency
                     options:
                         class: "%sylius.model.currency.class%"
-                        choiceLabel: name
+                        choice_label: name
                         fields: [sourceCurrency, targetCurrency]
 
             actions:
--- a/src/Sylius/Bundle/GridBundle/Form/Type/Filter/EntityFilterType.php
+++ b/src/Sylius/Bundle/GridBundle/Form/Type/Filter/EntityFilterType.php
@@ -32,8 +32,8 @@ final class EntityFilterType extends AbstractType
             'placeholder' => 'sylius.ui.all',
         ];
 
-        if (null !== $options['choiceLabel']) {
-            $idOptions['choice_label'] = $options['choiceLabel'];
+        if (null !== $options['choice_label']) {
+            $idOptions['choice_label'] = $options['choice_label'];
         }
 
         $builder
@@ -48,7 +48,7 @@ final class EntityFilterType extends AbstractType
     {
         $resolver
             ->setDefaults([
-                'choiceLabel' => null,
+                'choice_label' => null,
                 'class' => null,
             ])
             ->setRequired('class')
--- a/src/Sylius/Bundle/GridBundle/Form/Type/Filter/EntityFilterType.php
+++ b/src/Sylius/Bundle/GridBundle/Form/Type/Filter/EntityFilterType.php
@@ -24,32 +24,13 @@ final class EntityFilterType extends AbstractType
     /**
      * {@inheritdoc}
      */
-    public function buildForm(FormBuilderInterface $builder, array $options)
-    {
-        $idOptions = [
-            'class' => $options['class'],
-            'label' => false,
-            'placeholder' => 'sylius.ui.all',
-        ];
-
-        if (null !== $options['choice_label']) {
-            $idOptions['choice_label'] = $options['choice_label'];
-        }
-
-        $builder
-            ->add('id', EntityType::class, $idOptions)
-        ;
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function configureOptions(OptionsResolver $resolver)
     {
         $resolver
             ->setDefaults([
-                'choice_label' => null,
                 'class' => null,
+                'label' => false,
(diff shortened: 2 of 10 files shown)


src/Sylius/Bundle/AdminBundle/Resources/config/grids/exchange_rate.yml
```diff
@@ -28,6 +28,15 @@ sylius_grid:
                     type: string
                     label: sylius.ui.ratio
                     sortable: ~
+            filters:
+                currency:
+                    type: entity
+                    label: sylius.ui.currency
+                    options:
+                        class: "%sylius.model.currency.class%"
+                        choiceLabel: name
```
pjedrzejewski: should be ``choice_label``.

src/Sylius/Bundle/GridBundle/Form/Type/Filter/EntityFilterType.php
```diff
@@ -42,11 +48,14 @@ public function configureOptions(OptionsResolver $resolver)
     {
         $resolver
             ->setDefaults([
+                'choice_label' => null,
                 'class' => null,
-                'data_class' => null,
             ])
-            ->setDefined('class')
             ->setRequired('class')
+            ->setDefined('field')
```
pamil: It's not used anywhere.

src/Sylius/Bundle/GridBundle/Form/Type/Filter/EntityFilterType.php
```diff
@@ -42,11 +48,14 @@ public function configureOptions(OptionsResolver $resolver)
...
+            ->setAllowedTypes('field', 'string')
+            ->setDefined('fields')
```
pamil: It's not used anywhere.

src/Sylius/Bundle/GridBundle/Form/Type/Filter/EntityFilterType.php
```diff
@@ -26,12 +26,18 @@
...
+            $idOptions['choice_label'] = $options['choice_label'];
+        }
+
         $builder
-            ->add('id', EntityType::class, [
-                'class' => $options['class'],
-                'label' => false,
-                'placeholder' => 'sylius.ui.all',
-            ])
+            ->add('id', EntityType::class, $idOptions)
```
pamil: Why can't we just set `EntityType` as its parent and then only overwrite `label` and `placeholder` options in `configureOptions`?

src/Sylius/Component/Grid/Filter/EntityFilter.php
```diff
@@ -29,8 +29,16 @@ public function apply(DataSourceInterface $dataSource, $name, $data, array $opti
...
 
         $field = isset($options['field']) ? $options['field'] : $name;
+        $fields = isset($options['fields']) ? $options['fields'] : [$field];
 
-        $dataSource->restrict($dataSource->getExpressionBuilder()->equals($field, $data['id']));
+        $expressionBuilder = $dataSource->getExpressionBuilder();
+
+        $expressions = [];
+        foreach ($fields as $field) {
+            $expressions[] = $expressionBuilder->equals($field, $data['id']);;
```
pamil: If we follow my suggestion from above to make `EntityFilterType` a child of `EntityType`, we'll just need `$data` instead of `$data['id']` here.

src/Sylius/Bundle/GridBundle/Form/Type/Filter/EntityFilterType.php
GSadee: It is used in `EntityFilter`

src/Sylius/Bundle/GridBundle/Form/Type/Filter/EntityFilterType.php
GSadee: It is used in `EntityFilter`

src/Sylius/Bundle/GridBundle/Form/Type/Filter/EntityFilterType.php
pamil: ♋️

src/Sylius/Component/Grid/Filter/EntityFilter.php
```diff
@@ -29,8 +29,16 @@ public function apply(DataSourceInterface $dataSource, $name, $data, array $opti
         }
 
         $field = isset($options['field']) ? $options['field'] : $name;
+        $fields = isset($options['fields']) ? $options['fields'] : [$field];
```
pamil: Anyway, if we're adding `fields`, let's just remove `field`.

GSadee: I guess that we should leave both version, because `field` could be a more common use case

pamil: Then some sanity check that allows only for one of them to be set (`isset($options['field']) xor isset($options['fields']) === true`) would be nice, as setting them both would result in unexpected behaviour.

GSadee: Unexpected? If both are set, this implementation uses only fields as it is more complex setting and ignores a field.

pjedrzejewski:
Thank you Grzesiu!

src/Sylius/Component/Grid/Filter/EntityFilter.php
pamil: Yup, let's say I have the following grid configuration included from thrid-party vendor:

```yaml
sylius_grid:
  my_grid:
    # ...
    filters:
      field:
        type: entity
        options:
          fields: [first_field, second_field]
```

Then I want to overwrite the filtering to use the first field only:

```
sylius_grid:
  my_grid:
    # ...
    filters:
      field:
        type: entity
        options:
          field: first_field
```

This merges into:

```yaml
sylius_grid:
  my_grid:
    # ...
    filters:
      field:
        type: entity
        options:
          field: first_field
          fields: [first_field, second_field]
```

And I am not expecting, that GridBundle will still look for that value in `second_field`.

pamil: Anyway, even if `field` option is a more common use-case, replacing it with `fields` is pretty straightforward, we just change:

```yaml
sylius_grid:
  my_grid:
    # ...
    filters:
      field:
        type: entity
        options:
          field: field
```

Into:

```yaml
sylius_grid:
  my_grid:
    # ...
    filters:
      field:
        type: entity
        options:
          fields: [field]
```

Three characters more in the config, but lots of nasty logic removed from code.

pjedrzejewski:
@GSadee Could you apply changes suggested by @pamil, please? Thank you!

PR #7076, opened 2016-12-13
[Admin][ShippingMethod][TaxRate] Null zone validation

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |


pjedrzejewski:
I understand the need for scenario, but this is a BC break and I am not sure if it is worth it or even makes sense. We are changing the signature of the method in order to test validation. But we actually never want to setZone(null) to happen.

Zales0123:
We're actually changing method signature to make validation work 😕  previous implementation threw fatal error as it tried to null was passed as ``public function setZone(ZoneInterface $zone)`` method argument.

pjedrzejewski:
I see now, thank you Mateusz! :)

PR #7078, opened 2016-12-13
[Checkout][Core] Fix disappearing email after going back to addressing step

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

Proceeding checkout as guest and then going back to addressing step caused vanishing of email field. That was because we only checked is there `customer` set on cart and if yes, then corresponding field was not added. After this change, we're also checking if there is also `user` on this cart customer - which means is there user logged in or not. Maybe it's not the best solution but imo it would be sufficient for now.


michalmarcinkowski:
Missing Behat scenario for that. "Going back to addressing and changing email address".

pjedrzejewski:
Thank you Mateusz! 👍

PR #7082, opened 2016-12-14
[Promotion] Inject correct registry in subscriber

rvanlaak:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | see https://github.com/Sylius/Sylius/pull/5946#issuecomment-266992768 |
| License         | MIT |

During our upgrade process to `beta.1` the forms were not able to render, as the `action` registry was injected in the `rule` subscriber (and the other way around).


pjedrzejewski:
Thanks Richard!

PR #7085, opened 2016-12-14
[Promotion] Make subscriber abstract, rename variables

rvanlaak:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | comment https://github.com/Sylius/Sylius/pull/5946#issuecomment-266992768 |
| License         | MIT |

As the `AbstractConfigurationSubscriber` should be abstract, it should not mention rule or action. Renamed these internal variables for readability, as it both can be `Rules` and `Actions` (or anything implementing the interfaces).

### DIFF
--- a/src/Sylius/Bundle/PromotionBundle/Form/EventListener/AbstractConfigurationSubscriber.php
+++ b/src/Sylius/Bundle/PromotionBundle/Form/EventListener/AbstractConfigurationSubscriber.php
@@ -63,13 +63,13 @@ abstract class AbstractConfigurationSubscriber implements EventSubscriberInterfa
      */
     public function preSetData(FormEvent $event)
     {
-        $action = $event->getData();
+        $data = $event->getData();
 
-        if (null === $type = $this->getRegistryIdentifier($action, $event->getForm())) {
+        if (null === $type = $this->getRegistryIdentifier($data, $event->getForm())) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


rvanlaak:
Actually, next problem with the `AbstractSubscriber` is that it doesn't allow you to pass options, as `addConfigurationFields` does add the fields to the form (see https://github.com/Sylius/Sylius/pull/5946#issuecomment-266998716).

Probably this can be resolved via a `FormExtension` I guess? But should the DX be that hard? :)

rvanlaak:
Having an extension to set the default currency solved it, but feels kinda harsh to do this. What about making `currency` a configuration parameter (with a default), which "core" thereafter makes this channel-specific?

```php
class ItemTotalRuleConfigurationExtension extends AbstractTypeExtension
{

    {
        $resolver
            ->setDefault('currency', 'EUR')
        ;
    }

    public function getExtendedType()
    {
        return ItemTotalConfigurationType::class;
    }

}
```

Service definition:

```yml
promotion.item_total_rule_configuration_extension:
    class: AppBundle\Promotion\Type\ItemTotalRuleConfigurationExtension
    tags:        
        - { name: form.type_extension, extended_type: Sylius\Bundle\PromotionBundle\Form\Type\Rule\ItemTotalConfigurationType }
```

michalmarcinkowski:
Thanks Richard!

Issue #7079, opened 2016-12-13
[RFC] Bring back original price in product variants


gorkalaucirica:
Hi, 

Is this feature in the backlog? If so, lets discuss it and **we will send a PR**.

In #6774 this feature was removed in order to introduce [ChannelPricing](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Model/ChannelPricing.php)

**Specs** that were deleted and we need  to recover:
- it_does_not_have_original_price_by_default
- its_original_price_should_be_mutable
- its_original_price_should_accept_only_integer

**Scenarios:**
```gherkin
# features/product/managing_products/editing_product.feature
Scenario: Changing a simple product price
        Given I want to modify the "Dice Brewing" product
        When I change its original price to "$25.00" and price to "$15.00" for "United States" channel
        And I save my changes
        Then I should be notified that it has been successfully edited
        And it should be priced at $15.00 with $10.00 discount included for channel "United States"
```

```gherkin
# features/product/viewing_products/viewing_product_price_on_products_list.feature
Scenario: Viewing products with discounted price on list
        Given the store has a product "T-shirt watermelon" priced at "$19.00" with "$10.00" included
        And the store classifies its products as "T-Shirts"
        And this product belongs to "T-Shirts"
        When I browse products from taxon "T-Shirts"
        Then I should see the product "T-shirt watermelon" with price "$19.00" and "$29.00" should have a strikethrough
```

```gherkin
# features/product/viewing_products/viewing_product.feature
Scenario: Viewing a detailed page with a discounted product price 
        Given the store has a product "T-shirt watermelon" priced at "$19.00" with "$10.00" discount
        When I check this product's details
        Then I should see the product "T-shirt watermelon" with price "$19.00" and "$29.00" should have a strikethrough
```

```gherkin
# features/cart/shopping_cart/adding_product_to_cart_with_discounted_prices_dependent_on_channel.feature
@shopping_cart
Feature: Adding a product to cart with prices dependent on a channel and currency
    In order to buy products in correct discounted prices
    As a Customer
    I want to add products to my cart with discounted prices dependent on visited channel and currency
    
    Background:
        Given the store has currency "EUR"
        And the store has currency "GBP"
        And the store operates on a channel named "Web-EU" in "EUR" currency
        And that channel allows to shop using "EUR" and "GBP" currencies
        And the store operates on another channel named "Web-GB" in "GBP" currency
[...]

PR #7087, opened 2016-12-14
[WIP] Bring back original price in product variants

gorkalaucirica:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | #7079  |
| License         | MIT |

WIP.

- Admin scenarios are completed (I guess)
- I had an issue with form labels, cannot translate them. Symfony does not support passing variables, any workaround?
- I will really appreciate a code review with some feedback before I add the required features to ShopBundle

Thanks in advance!!


gorkalaucirica:
I broke some tests... Will try to fix them

src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
```diff
@@ -492,6 +501,14 @@ public function iChangeItsPriceTo($price, $channelName)
     }
 
     /**
+     * @When /^I change its original price to (?:€|£|\$)([^"]+) for "([^"]+)" channel$/
```
pamil: We usually put prices inside quotation marks, so `(?:€|£|\$)([^"]+)` -> `"(?:€|£|\$)([^"]+)"`

src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
```diff
@@ -1004,6 +1021,20 @@ public function itShouldBePricedAtForChannel(ProductInterface $product, $price,
     }
 
     /**
+     * @Then /^(its|this products) original price should be (?:€|£|\$)([^"]+) for channel "([^"]+)"$/
```
pamil: We usually put prices inside quotation marks, so `(?:€|£|\$)([^"]+)` -> `"(?:€|£|\$)([^"]+)"`

src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
```diff
@@ -1004,6 +1021,20 @@ public function itShouldBePricedAtForChannel(ProductInterface $product, $price,
...
+     * @Then /^(product "[^"]+") original price should be (?:€|£|\$)([^"]+) for channel "([^"]+)"$/
```
pamil: We usually put prices inside quotation marks, so `(?:€|£|\$)([^"]+)` -> `"(?:€|£|\$)([^"]+)"`

src/Sylius/Behat/Page/Admin/ProductVariant/CreatePageInterface.php
```diff
@@ -27,13 +28,19 @@
     public function specifyPrice($price, $channel);
 
     /**
+     * @param int $originalPrice
+     * @param string $channel
+     */
+    public function specifyOriginalPrice($originalPrice, $channel);
```
pamil: `$channel` -> `$channelName`?

src/Sylius/Behat/Page/Admin/Product/CreateSimpleProductPage.php
```diff
@@ -68,6 +69,14 @@ public function specifyPrice($channelName, $price)
     /**
      * {@inheritdoc}
      */
+    public function specifyOriginalPrice($channelName, $originalPrice)
+    {
+        $this->getElement('original_price', ['%channel%' => $channelName])->setValue($originalPrice);
```
Arminek: `%channelName%` WDYT?

src/Sylius/Bundle/AdminBundle/Resources/views/ProductVariant/Tab/_details.html.twig
```diff
@@ -10,7 +10,12 @@
             <div class="two fields">
                 {{ form_row(form.shippingCategory) }}
             </div>
-            {{ form_row(form.channelPricings) }}
+            {% for pricing_form in form.channelPricings %}
```
pamil: `pricing_form` -> `pricingForm`?

src/Sylius/Component/Core/spec/Model/ChannelPricingSpec.php
```diff
@@ -49,4 +50,41 @@ function its_price_is_mutable()
...
+        $this->getOriginalPrice()->shouldReturn(null);
+    }
+
+    function its_original_price_should_be_mutable()
+    {
+        $this->setOriginalPrice(2000);
+        $this->getOriginalPrice()->shouldReturn(2000);
+    }
+
+    function its_original_price_should_not_be_lower_than_price()
```
tuka217: Shouldn't you check somehow that that the original price is not lower than the price? Or maybe this function is outdated?

src/Sylius/Behat/Context/Ui/Admin/ManagingProductVariantsContext.php
```diff
@@ -170,6 +171,14 @@ public function iSetItsPriceTo($price = null, $channel = null)
     }
 
     /**
+     * @When /^I set its original price to "(?:€|£|\$)([^"]+)" for "([^"]+)" channel$/
+     */
+    public function iSetItsOriginalPriceTo($originalPrice = null, $channel = null)
+    {
+        $this->createPage->specifyOriginalPrice($originalPrice, (null === $channel) ? $this->sharedStorage->get('channel') :$channel);
```
Zales0123: 1. We should definitely pass a name of channel two that function, not an object (it would require ``__toString`` method)
2. With this step definition it's not even possible to have ``$originalPrice``` and ``$channel`` as null (as it won't be matched ;))

src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
Zales0123: Is this step used anywhere? We should not define steps in advance.

src/Sylius/Behat/Page/Admin/Product/CreateSimpleProductPage.php
```diff
@@ -241,7 +250,8 @@ protected function getDefinedElements()
             'images' => '#sylius_product_images',
             'language_tab' => '[data-locale="%locale%"] .title',
             'name' => '#sylius_product_translations_%locale%_name',
-            'price' => '#sylius_product_variant_channelPricings [data-form-collection="item"]:contains("%channel%") input',
+            'price' => '#sylius_product_variant_channelPricings [data-form-collection="item"]:contains("%channel%") input[name$="[price]"]',
+            'original_price' => '#sylius_product_variant_channelPricings [data-form-collection="item"]:contains("%channel%") input[name$="[originalPrice]"]',
```
Zales0123: I would keep alphabetical order here ;)

src/Sylius/Behat/Page/Admin/Product/UpdateSimpleProductPage.php
```diff
@@ -427,7 +447,8 @@ protected function getDefinedElements()
             'images' => '#sylius_product_images',
             'language_tab' => '[data-locale="%locale%"] .title',
             'name' => '#sylius_product_translations_%locale%_name',
-            'price' => '#sylius_product_variant_channelPricings [data-form-collection="item"]:contains("%channel%") input',
+            'price' => '#sylius_product_variant_channelPricings [data-form-collection="item"]:contains("%channel%") input[name$="[price]"]',
+            'original_price' => '#sylius_product_variant_channelPricings [data-form-collection="item"]:contains("%channel%") input[name$="[originalPrice]"]',
```
Zales0123: And here.

src/Sylius/Bundle/CoreBundle/Form/Type/Product/ChannelPricingType.php
```diff
@@ -35,7 +36,12 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 $form = $event->getForm();
 
                 $form->add('price', MoneyType::class, [
-                    'label' => $channel->getName(),
+                    'label' => 'Price for ' . $channel->getName(),
```
Zales0123: 1. What is the reason of this change? I know the UI is not perfect for now, but is it really required? (provide some screenshots if you don't mind 😄 )
2. If we need this, there should be a translation key.

src/Sylius/Behat/Context/Ui/Admin/ManagingProductVariantsContext.php
Arminek: Maybe try to use transformer. WDYT? Right now `$channel` has mixed type.

src/Sylius/Bundle/CoreBundle/Form/Type/Product/ChannelPricingType.php
pamil: We should look for some way to allow translating these labels.

src/Sylius/Component/Core/Model/ChannelPricingInterface.php
```diff
@@ -1,12 +1,22 @@
 <?php
 
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
```
Zales0123: 👍

src/Sylius/Component/Core/spec/Model/ChannelPricingSpec.php
```diff
@@ -49,4 +50,41 @@ function its_price_is_mutable()
         $this->setPrice(1000);
         $this->getPrice()->shouldReturn(1000);
     }
+
+    function it_does_not_have_original_price_by_default()
+    {
+        $this->getOriginalPrice()->shouldReturn(null);
+    }
+
+    function its_original_price_should_be_mutable()
```
Zales0123: Avoid word ``should`` in spec names -> ``its_original_price_is_mutable`` would be sufficient

src/Sylius/Component/Core/spec/Model/ChannelPricingSpec.php
```diff
@@ -49,4 +50,41 @@ function its_price_is_mutable()
...
+        $this->getOriginalPrice()->shouldReturn(2000);
+    }
+
+    function its_original_price_should_not_be_lower_than_price()
+    {
+        $this->setPrice(2000);
+        $this->setOriginalPrice(1000);
+    }
+
+    function its_price_is_reduced()
```
Zales0123: What about ``its_price_can_be_reduced``? It's not reduced all the time 😄

src/Sylius/Behat/Page/Admin/ProductVariant/CreatePage.php
```diff
@@ -29,7 +30,15 @@ class CreatePage extends BaseCreatePage implements CreatePageInterface
      */
     public function specifyPrice($price, $channel)
     {
-        $this->getDocument()->fillField($channel, $price);
+        $this->getDocument()->fillField('Price for ' . $channel, $price);
```
Arminek: `specifyPrice($price, ChannelInterface $channel)`. Instead of concatenation try to use `sprintf('Price for %s', $channel->getName())`.

src/Sylius/Component/Core/spec/Model/ChannelPricingSpec.php
Arminek: It seams that `ChannelPricing` does not have this behaviour.

src/Sylius/Component/Core/spec/Model/ChannelPricingSpec.php
Arminek: `its_checks_if_price_has_been_reduced` WDYT?

src/Sylius/Behat/Page/Admin/Product/CreateSimpleProductPage.php
gorkalaucirica: It's copied from above, can change both

src/Sylius/Behat/Page/Admin/ProductVariant/CreatePageInterface.php
gorkalaucirica: It's copied from above, can change both

src/Sylius/Bundle/CoreBundle/Form/Type/Product/ChannelPricingType.php
gorkalaucirica: As i commented in the PR details I don't know how to add a translation to a label that supports a variable.
Translation string could be Price for %channel% in english but cannot pass to the translator that variable

src/Sylius/Component/Core/spec/Model/ChannelPricingSpec.php
gorkalaucirica: Whatever you want

src/Sylius/Component/Core/spec/Model/ChannelPricingSpec.php
gorkalaucirica: Should we have it? I copied as it was from the PR where the feature was removed, and definitely the test is not correct

gorkalaucirica:
I need to know how to approach to this two things:

- Original price validation, where should be done and constraints to be applied
- Form rendering, any idea to improve it? Injecting the translator to translate the label its a solution that works but maybe is not the best one

app/migrations/Version20161214101054.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_channel_pricing ADD originalPrice INT');
```
inssein: I think the convention is for the field to be be named `original_price`

src/Sylius/Bundle/CoreBundle/Form/Type/Product/ChannelPricingType.php
gabiudrescu: Maybe this can help: 
http://stackoverflow.com/questions/19593851/symfony2-form-use-translator-for-attr-attribute

Although, it seems this feature is not provided out of the box by symfony: https://github.com/symfony/symfony/pull/10125#issuecomment-112468612

gabiudrescu:
I see originalPrice as following:

 - Default NOT null
 - if it is non existent, then it should be equal to sellingPrice
 - always equal or higher than selling price

This should be implemented as low level as possible. Schema, application level validation, you name it.

I believe this mostly because of separation of comcerns. By having prices as described above, the display layer should always check if those two are equal, then show both or not.

Any other new conditions breaks the separation of concern principle and it rings an alarm bell in my head.

If original price would have been lower than selling price, what would I, as a customer, understand from this? How would that motivate me to buy that product?

Once again, I am telling this because my perception about originalPrice is that we have it because this is the price that gets written on the page with a strikethrough font decoration (and a badge with blinking red saying discount, buy now)

pjedrzejewski:
I agree with @gabiudrescu, original price should be only for displaying purposes. I did not review the code yet, but we definitely want to keep existing logic and only add this as a simple feature.

gorkalaucirica:
@pjedrzejewski So its enough if we leave original_price without validation and we use `isPriceReduced()` to check if a promotion is available in ShopBundle and painting the original page with a strikethrough according to that?

crevillo:
i don't understand the funwction of the original_price combined with the price then. I mean, if we're going to strikethrough the original price and print the price depending on promotions why we need the price itself? price or original price should be enough. 

I would undertand better the original price thing if "price" will be threated as a discount, just like magento does in some ways. something like ~$20.00~ $10.00 from x to y (or something like that)

Actually i feel is not very easy to add different discounts to a lot products, btw...

pjedrzejewski:
@gorkalaucirica Exactly. Catalog wide discounts is a separate feature that we will need to tackle as a separate plugin or maybe it would be possible to implement them without BC breaks. Otherwise things are getting a bit too complicated, while not really solving the problem of catalog wide promotions. (you would still need to edit these manually)

gorkalaucirica:
@pjedrzejewski So we can leave the model and the admin as it is and start working on the shop?

src/Sylius/Bundle/CoreBundle/Form/Type/Product/ChannelPricingType.php
gorkalaucirica: Yes, and the don`t plan to support this, so injecting the translator seems the only way to do it

pamil: As for translation issues:

We can set label for `price` field to "sylius.ui.price", label for `originalPrice` field to "sylius.ui.original_price" and then pass `$channel` to form view to allow for custom rendering. Then we can overwrite Twig block for channel pricing (after #7664 is merged) to something like:

```html
<fieldset>
  <legend>{{ channel.name }}</legend>
  {{ form_row(channelPricing.price) }}
  {{ form_row(channelPricing.originalPrice) }}
</fieldset>
```

If you can't find time to dig into it further, tell me so I'll make a PR to your branch.

app/migrations/Version20161214101054.php
pamil: 👍

gorkalaucirica:
@pamil I will rebase and try to apply your suggested changes. Hope this gets into beta2. We've been using it for a couple of months and works great. I think it's a must have

gorkalaucirica:
Rebased and changes made. 

Still to do:

* Form string translation. Maybe @pamil can help here and somehow blocked by #7664 
* Implement this feature in ShopBundle (can be done in another PR)

Anything else?

gorkalaucirica:
Rebased after #7664

pamil:
Form translations made in https://github.com/gorkalaucirica/Sylius/pull/1 :)

gorkalaucirica:
@pamil merged

pamil:
https://github.com/gorkalaucirica/Sylius/pull/2 should make tests passing once again 🎉

gorkalaucirica:
@pamil Merged 😉

pjedrzejewski:
Thanks a lot Gorka! And everyone for review & help. @gorkalaucirica Would you find a moment to have a look at the shop part of the implementation?

gorkalaucirica:
@pjedrzejewski I will give a try this weekend if I find some time.

PR #7088, opened 2016-12-14
[Shop] Sorting products by price

TheMadeleine:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

Together with @Zales0123 ;)


pjedrzejewski:
We should add a scenario about products with multiple variants and make sure we sort by the price of lowest variant.

src/Sylius/Bundle/ShopBundle/Resources/views/Product/Index/_main.html.twig
```diff
@@ -8,7 +8,7 @@
     <div class="ui three column stackable grid" id="products">
         {% for product in resources.data %}
             <div class="column">
-                {% include '@SyliusShop/Product/_box.html.twig' %}
+                {% include '@SyliusShop/Product/_box.html.twig' with { 'product': product[0] } %}
```
michalmarcinkowski: Do we need this?

TheMadeleine: @Zales0123 do we need this? 😆

pjedrzejewski:
Thank you Magda and Mateusz! Nice work! :)

TheMadeleine:
Hi @inspiran It may fail unfortunately :( This query does not use any aggregating function. I didn't have this issue locally since I'm using mysql 5.6.26.

@michalmarcinkowski @pjedrzejewski What do you suggest doing?

marcoslibre:
This is a SQL standard. Omitting unaggregated columns leads to indeterminate behavior.
MySQL (prior 5.7) sloppy implementation of this is more like a bug than a feature.

PR #7091, opened 2016-12-14
[Product] Translatable product variant name

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | |
| License         | MIT |

### DIFF
--- a/features/product/viewing_products/viewing_product_variants_names.feature
+++ b/features/product/viewing_products/viewing_product_variants_names.feature
@@ -5,19 +5,19 @@ Feature: Viewing product's variants names
     I want to be aware of product's variants names
     
     Background:
-        Given the store operates on a single channel in "United States"
-        And the store is available in "English (United States)"
-        And the store is also available in "Polish (Poland)"
+        Given the store operates on a channel named "Web-US" in "USD" currency
+        And that channel allows to shop using "English (United States)" and "Polish (Poland)" locales
         And the store has a "Die Hard Movie" configurable product
         And it has variant named "Die Hard - Extended Cut" in "English (United States)" locale and "Szklana Pułapka - Wersja Reżyserska" in "Polish (Poland)" locale
+        And this product has also variant named "Die Hard - Theatrical Cut" in "English (United States)" locale and "Szklana Pułapka - Wersja Kinowa" in "Polish (Poland)" locale
 
     @ui
     Scenario: Seeing variant's name in default locale
-        When I view product "Wyborowa Vodka"
+        When I view product "Die Hard Movie"
         Then its current variant should be named "Die Hard - Extended Cut"
 
     @ui
     Scenario: Seeing proper variant's name after locale change
         When I switch to the "Polish (Poland)" locale
-        And I view product "Wyborowa Vodka"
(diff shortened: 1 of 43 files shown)


pjedrzejewski:
* Migration should preserve the data saved already (I'd suggest adding it to the default locale - should be safe assumption)
* UPGRADE file should be updated!

features/product/viewing_products/viewing_product_variants_names.feature
```diff
@@ -0,0 +1,23 @@
...
+    @ui
+    Scenario: Seeing variant's name in default locale
+        When I view product "Die Hard Movie"
+        Then its current variant should be named "Die Hard - Extended Cut"
+
+    @ui
+    Scenario: Seeing proper variant's name after locale change
+        When I switch to the "Polish (Poland)" locale
+        And I view product "Die Hard Movie"
+        Then its current variant should be named "Szklana Pułapka - Wersja Reżyserska"
```
michalmarcinkowski: Selecting other variant should change the displayed name, does it work right now or will it be a new feature?

src/Sylius/Bundle/AdminBundle/Resources/views/ProductVariant/Tab/_details.html.twig
```diff
@@ -2,8 +2,18 @@
...
+                <div data-locale="{{ locale }}">
+                    <div class="title{% if 0 == loop.index0 %} active{% endif %}">
+                        <i class="dropdown icon"></i>
+                        <i class="{{ locale|slice(3, 2)|lower }} flag"></i> {{ locale|sylius_locale_name }}
+                    </div>
+                    <div class="ui content{% if 0 == loop.index0 %} active{% endif %}">
+                        {{ form_row(translationForm.name) }}
+                    </div>
+                </div>
+            {% endfor %}
```
michalmarcinkowski: I noticed that almost all translation forms rendering look the same, maybe we could extract it to a macro?

Zales0123: Really good idea! 👍

features/product/viewing_products/viewing_product_variants_names.feature
Zales0123: I think it will be a new feature - for now if we have variant selection method chosen as ``choice``, then variant names are displayed in variants table. If its is ``match`` then, yes, it's not changed anywhere :/ I think we could implement changing product name or event add some subheader with variant name under product name? To be considered, but totally for separate PR ;)

gorkalaucirica:
Maybe add a link to the PR where the BC break was done will help when the UPGRADE needs to be done to check what has been changed, especially a review the PR to find if any relevant changes have been applied to the the themes

WDYT?

pjedrzejewski:
@gorkalaucirica Makes perfect sense, thanks for the suggestion! @Zales0123 Let's do this. :)

pjedrzejewski:
Thank you Mateusz! :) Nice work!

PR #7096, opened 2016-12-14
Fix #7093 #7094

### LINKED ISSUE #7093: [SyliusUserBundle] "The helper "dialog" is not defined."
Hi,

I downloaded a fresh dev-master of sylius-standard and I wanted to play with the Sylius API.
So, after creating my first admin user account, when interacting with `sylius:install`, I wanted to call `sylius:user:promote` to add a `ROLE_API_ACCESS` to my account.

Then I got the following exception:
```
  [Symfony\Component\Console\Exception\InvalidArgumentException]
  The helper "dialog" is not defined.
```

Indeed the code relies on `$this->getHelper('dialog')` to interact with the human behind the terminal, which is deprecated since Symfony 2.5 and removed since Symfony 3.0:
http://symfony.com/doc/current/components/console/helpers/dialoghelper.html

A minor refactoring should be done to use the new Question helper.

### LINKED ISSUE #7094: [SyliusUserBundle] Promoting / Demoting only work on "shop" users
After fixing #7093 on my side, I wanted to promote a `ROLE_API_ACCESS` to my user account, in order to use the API.

Then the console told me there was no account registered with the email address I entered.

This is because the promote / demote actions only rely on `sylius.repository.shop_user`, although my account is an admin account.

I suggest asking the human to specify which type of account it should retrieve before applying any modification.

bpolaszek:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | maybe* |
| Related tickets | fixes #7093 #7094  |
| License         | MIT |

## Promoting / Demoting user roles
[SyliusUserBundle] Refactored question helper - fixes #7093 
[SyliusUserBundle] Added a new interaction to specify the user type (shop / admin) - fixes #7094 

## * BC breaks 
* I don't know if this can be called a "BC break" since the code was already broken since Symfony 3.0, and the patch should work on Symfony 2.5+.
* The protected function `Sylius\Bundle\UserBundle\Command\AbstractRoleCommand::executeRoleCommand` has a new 1st argument: `InputInterface`. This indeed leads to a BC break to possible custom inheritances of `Sylius\Bundle\UserBundle\Command\AbstractRoleCommand`, but the probability to drink a beer with an alien on the Eiffel Tower one day is more important than facing that kind of issue, IMHO.

### DIFF
--- a/src/Sylius/Bundle/UserBundle/Command/AbstractRoleCommand.php
+++ b/src/Sylius/Bundle/UserBundle/Command/AbstractRoleCommand.php
@@ -17,42 +17,55 @@ use Sylius\Component\User\Repository\UserRepositoryInterface;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Console\Question\Question;
 
 /**
  * @author Loïc Frémont <loic@mobizel.com>
  */
 abstract class AbstractRoleCommand extends ContainerAwareCommand
 {
+
+    const ADMIN = 'admin';
+    const SHOP = 'shop';
+
     /**
      * {@inheritdoc}
      */
     protected function interact(InputInterface $input, OutputInterface $output)
     {
         if (!$input->getArgument('email')) {
-            $email = $this->getHelper('dialog')->askAndValidate(
-                $output,
-                'Please enter an email:',
-                function ($email) {
-                    if (empty($email)) {
-                        throw new \Exception('Email can not be empty');
-                    }
-
-                    return $email;
+            $question = new Question('Please enter an email:');
+            $question->setValidator(function ($email) {
+                if (!filter_var($email, FILTER_VALIDATE_EMAIL)) {
+                    throw new \RuntimeException("The email you entered is invalid.");
                 }
-            );
-
+                return $email;
+            });
+            $email = $this->getHelper('question')->ask($input, $output, $question);
             $input->setArgument('email', $email);
         }
 
         if (!$input->getArgument('roles')) {
-            $roles = $this->getHelper('dialog')->ask(
-                $output,
-                'Please enter user\'s roles (separated by space):'
-            );
+            $question = new Question('Please enter user\'s roles (separated by space):');
+            $question->setValidator(function ($roles) {
+                if (strlen($roles) < 1) {
+                    throw new \RuntimeException("The value cannot be blank.");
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/UserBundle/Command/AbstractRoleCommand.php
```diff
@@ -17,43 +17,56 @@
...
+use Symfony\Component\Console\Question\ChoiceQuestion;
+use Symfony\Component\Console\Question\Question;
 
 /**
  * @author Loïc Frémont <loic@mobizel.com>
  */
 abstract class AbstractRoleCommand extends ContainerAwareCommand
 {
+
+    const ADMIN = 'admin';
```
Arminek: All user related commands should be in ``CoreBundle``.

pjedrzejewski:
@bpolaszek Nice work! The only problem (as outlined by @Arminek) is that "shop/admin" is choice specific to Sylius Core. This bundles is not Sylius-only, so we should probably just ask about the type without mentioning shop/admin or get the types from configuration. (each type needs to have a class configured in yml, so we know all types configured)

pjedrzejewski:
@bpolaszek Hi Benoit! Do you plan to work on this PR anytime soon? Thank you for any info! :)

bpolaszek:
Hi @pjedrzejewski, @Arminek !

Sorry for the delay. And happy new year, btw :)
Alright, I thought SyliusUserBundle, unlike other Sylius bundles, was tightly coupled to Sylius Standard.

When I look at the [configuration reference](http://docs.sylius.org/en/latest/bundles/SyliusUserBundle/summary.html), 3 types of users are used in Sylius:
* admin
* shop
* oauth

Semantically, how would you name a variable containing one of these values? `$userType` ? `$userRepositoryType`? None of them convince me.

Besides, I think that, when using SyliusUserBundle as a standalone library, there's no notion of admin / shop / oauth. The user is free to configure its own repository with the correct names according to its application (it could be values like "member", "affiliate", ... instead of "shop", "admin" etc). 

As a consequence, I think we should be able to list exactly which types are already configured and suggest them when using the promote / demote commands.

What do you think about it?

bpolaszek:
Hello again,

After digging into the code I found the answers to all my questions by myself :)

1. Let's use "user type" to name this, unless you prefer something else
2. The list of user types and their respective configurations can be retrieved from `$this->getContainer()->getParameter('sylius.user.users')`
3. This list can be filtered when the model class implements `Sylius\Component\User\Model\UserInterface`
4. When this list is empty, an exception should be thrown because there's no promotable / demotable users
5. When this list contains only 1 item, the developer using the CLI should not be prompted to specify the user type, it should be filled automatically.

New pull request on its way :)

PR #7080, opened 2016-12-13
[Paypal] Mock http client interface instead of concrete class

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

PR #7102, opened 2016-12-15
Use php-http/guzzle6-adapter explicitly

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | related to #7080 |
| License         | MIT |

Right now, if you remove `composer.lock` and `vendor`, `composer update` will fail.

We still have `php-http/guzzle6-adapter` in our `composer.lock`, but there's no information about that in `composer.json`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,6 +44,7 @@
         "ocramius/proxy-manager": "^1.0",
         "payum/payum": "^1.3",
         "payum/payum-bundle": "^2.1",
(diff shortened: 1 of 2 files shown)


Arminek:
@pamil good catch from some reason composer remove does not update ``composer.lock`` properly.

pjedrzejewski:
Thank you Kamil!

PR #7104, opened 2016-12-15
Remove final from ChannelPricing

### LINKED ISSUE #7103: ChannelPricing model is final an cannot be extended
According to [model customization guide](http://docs.sylius.org/en/latest/customization/model.html) extending a model is allowed in Sylius but it's not true for ChannelPricing.

Is there any reason for this class to be different from others?

pamil:
Fixes #7103.


pjedrzejewski:
Thank you Kamil!

PR #7107, opened 2016-12-15
Remove drivers configuration where possible

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | - |
| License         | MIT |


pjedrzejewski:
Why? IIRC it can cause some issues when disabling admin/shop separately on Sylius-Standard.

pamil:
It has happened back then when `doctrine/phpcr-odm` was used along with `doctrine/orm`. Having single `doctrine/orm` is the default option, so it's not needed to declare it both on shop and admin.

pjedrzejewski:
Right, thanks! :)

PR #7109, opened 2016-12-15
Channels should be protected

mathielen:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

If extended and the extended class is configured to be used, there is an doctrine error thrown that the property doesnt exist.


pjedrzejewski:
Thank you Markus! :)

Issue #7112, opened 2016-12-16
PromotionCouponPerCustomerUsageLimitEligibilityChecker - Binding to customer that hasn't been persisted


inssein:
I have been getting a lot of errors that say `Binding entities to query parameters only allowed for entities that have an identifier.` and I have finally figured out why.

The `PromotionCouponPerCustomerUsageLimitEligibilityChecker` tries to `$this->orderRepository->countByCustomerAndCoupon($customer, $promotionCoupon);` which throws an exception because often times the `$customer` entity is constructed by the form logic, and doesn't yet exist in the database.

PR #7113, opened 2016-12-16
[Core] Ensure customer exists before checking for eligibility

inssein:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/7112 |
| License         | MIT |

I hope this is good enough. The better way to do this would be to actually check if this entity is managed by the entity manager to infer existence, but this seemed simpler.


pjedrzejewski:
Can we add a scenario covering this bug? I think you should be able to write it with existing steps. :)

inssein:
done @pjedrzejewski

pjedrzejewski:
Thank you Hussein! Would be great to add a Behat scenario too, but for now this is a good fix and issue is quite critical. 👍

PR #7127, opened 2016-12-19
[Product] Do not show code validation message during update

### LINKED ISSUE #7005: [Bug][Admin][Product] Unique code validation error for simple product
When we create new simple product there is also create one variant for this product. Unfortunately they have the same variant and there is an constraint which says that code of product must be unique among products and variants defined in Sylius. 
We don't get information about this error during creating a product, but only when we make different mistake during editing simple product. 
screen shot 2016-12-05 at 11 59 40

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7005  |
| License         | MIT |

### DIFF
--- a/features/product/managing_products/product_validation.feature
+++ b/features/product/managing_products/product_validation.feature
@@ -107,3 +107,12 @@ Feature: Products validation
         And I try to save my changes
         Then I should be notified that name is required
         And this product should still be named "Dice Brewing"
+
+    @ui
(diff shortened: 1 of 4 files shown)


michalmarcinkowski:
Thanks Łukasz!

PR #7135, opened 2016-12-20
[Core] Not specified price error polishing

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

zrzut ekranu 2016-12-20 o 16 03 45

### DIFF
--- a/src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
+++ b/src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
@@ -212,7 +212,7 @@ final class ManagingProductsContext implements Context
      */
     public function iAssignItToChannel($channel)
     {
-        // Temporary solution until we we'll make current page resolver work with product pages
+        // Temporary solution until we will make current page resolver work with product pages
         $this->updateConfigurableProductPage->checkChannel($channel);
     }
 
--- a/src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
+++ b/src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
@@ -208,6 +208,15 @@ final class ManagingProductsContext implements Context
     }
 
     /**
+     * @When I assign it to channel :channel
+     */
+    public function iAssignItToChannel($channel)
+    {
+        // Temporary solution until we we'll make current page resolver work with product pages
+        $this->updateConfigurableProductPage->checkChannel($channel);
+    }
+
+    /**
      * @When I choose :calculatorName calculator
      */
     public function iChooseCalculator($calculatorName)
@@ -1017,6 +1026,17 @@ final class ManagingProductsContext implements Context
         Assert::same(
             $this->updateSimpleProductPage->getValidationMessageForImageAtPosition((int) $imageNumber - 1),
             'Image code must be unique within this product.'
+        );
+    }
+
+    /**
+     * @Then I should be notified that I have to define product variants' prices for newly assigned channels first
+     */
+    public function iShouldBeNotifiedThatIHaveToDefineProductVariantsPricesForNewlyAssignedChannelsFirst()
+    {
+        Assert::same(
+            $this->updateConfigurableProductPage->getValidationMessage('channels'),
+            'You have to define product variants\' prices for newly assigned channels first.'
         );
     }
 
--- a/src/Sylius/Bundle/CoreBundle/Validator/Constraints/HasAllVariantPricesDefinedValidator.php
+++ b/src/Sylius/Bundle/CoreBundle/Validator/Constraints/HasAllVariantPricesDefinedValidator.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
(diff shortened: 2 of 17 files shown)


src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
```diff
@@ -210,6 +210,15 @@ public function iMakeItAvailableInChannel($channel)
     }
 
     /**
+     * @When I assign it to channel :channel
+     */
+    public function iAssignItToChannel($channel)
+    {
+        // Temporary solution until we we'll make current page resolver work with product pages
```
lchrusciel: 😢

src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
```diff
@@ -1050,6 +1059,17 @@ public function iShouldBeNotifiedThatTheFirstImageShouldHaveAnUniqueCode($imageN
     }
 
     /**
+     * @Then I should be notified that I have to define product variants' prices for newly assigned channels first
+     */
+    public function iShouldBeNotifiedThatIHaveToDefineProductVariantsPricesForNewlyAssignedChannelsFirst()
+    {
+        Assert::same(
+            $this->updateConfigurableProductPage->getValidationMessage('channels'),
```
lchrusciel: 😢

src/Sylius/Bundle/CoreBundle/Validator/Constraints/HasAllVariantPricesDefinedValidator.php
```diff
@@ -0,0 +1,56 @@
...
+        if ($product->isSimple()) {
+            return;
+        }
+
+        $channels = $product->getChannels();
+
+        /** @var ProductVariantInterface $variant */
+        foreach ($product->getVariants() as $productVariant) {
+            /** @var ChannelInterface $channel */
+            foreach ($channels as $channel) {
```
lchrusciel: Can we extract it to some private method?

src/Sylius/Bundle/CoreBundle/Validator/Constraints/HasAllVariantPricesDefinedValidator.php
```diff
@@ -0,0 +1,56 @@
...
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+final class HasAllVariantPricesDefinedValidator extends ConstraintValidator
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function validate($product, Constraint $constraint)
+    {
+        Assert::isInstanceOf($constraint, HasAllVariantPricesDefined::class);
```
lchrusciel: Won't it be always true? Shouldn't we rather assert that a `$product` is instance of *ProductInterface*?

src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
Zales0123: I don't like these methods to, however as they're used in billions of places, they should be refactored all at once... Or maybe we don't care about consistency now and head for splendor, it's also a good way 😄

src/Sylius/Bundle/CoreBundle/Validator/Constraints/HasAllVariantPricesDefinedValidator.php
Zales0123: I would love to, however as we use `return` to stop validation on first invalid value, it would require to throw exception and catching it, so probably two more methods to keep one level of indentation - I don't know is it worth for such a simple case :(

src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
pamil: we we'll

michalmarcinkowski:
We should add one more scenario `Enable product in channel when all its variants already have specified price in this channel`.

Zales0123:
@michalmarcinkowski done 👍

src/Sylius/Bundle/ProductBundle/Resources/config/validation/Product.xml
```diff
@@ -33,7 +33,6 @@
             </constraint>
         </property>
         <getter property="variants">
-            <constraint name="Valid" />
```
pjedrzejewski: Are sure about removing it? What is the reason?

Zales0123: I moved this constraint to [here](https://github.com/Sylius/Sylius/pull/7135/files#diff-64aa754ad20c13ec011448205027e452R54), as it's only needed for simple product (even more - it destroys the universe if it's validated on configurable one 🐃 )

pjedrzejewski:
Thank you Mateusz!

michalmarcinkowski:
👍

PR #7136, opened 2016-12-20
[Core][Promotion][Taxonomy] Unify decorated services ids

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services/promotion.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services/promotion.xml
@@ -13,11 +13,11 @@
 
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
-        <service id="sylius.factory.promotion_action.decorated" class="Sylius\Component\Core\Factory\PromotionActionFactory" decorates="sylius.factory.promotion_action" public="false">
-            <argument type="service" id="sylius.factory.promotion_action.decorated.inner" />
+        <service id="sylius.custom_factory.promotion_action" class="Sylius\Component\Core\Factory\PromotionActionFactory" decorates="sylius.factory.promotion_action" public="false">
+            <argument type="service" id="sylius.custom_factory.promotion_action.inner" />
         </service>
-        <service id="sylius.factory.promotion_rule.decorated" class="Sylius\Component\Core\Factory\PromotionRuleFactory" decorates="sylius.factory.promotion_rule" public="false">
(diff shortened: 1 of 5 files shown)


pjedrzejewski:
Thank you Mateusz!

src/Sylius/Bundle/CoreBundle/Resources/config/services/promotion.xml
```diff
@@ -13,11 +13,11 @@
...
     <services>
-        <service id="sylius.factory.promotion_action.decorated" class="Sylius\Component\Core\Factory\PromotionActionFactory" decorates="sylius.factory.promotion_action" public="false">
-            <argument type="service" id="sylius.factory.promotion_action.decorated.inner" />
+        <service id="sylius.custom_factory.promotion_action" class="Sylius\Component\Core\Factory\PromotionActionFactory" decorates="sylius.factory.promotion_action" public="false">
+            <argument type="service" id="sylius.custom_factory.promotion_action.inner" />
         </service>
-        <service id="sylius.factory.promotion_rule.decorated" class="Sylius\Component\Core\Factory\PromotionRuleFactory" decorates="sylius.factory.promotion_rule" public="false">
-            <argument type="service" id="sylius.factory.promotion_rule.decorated.inner" />
+        <service id="sylius.custom_factory.promotion_rule" class="Sylius\Component\Core\Factory\PromotionRuleFactory" decorates="sylius.factory.promotion_rule" public="false">
```
michalmarcinkowski: But why "**custom**_factory"? Wouldn't it be better to call it more specifically e.g. "**core**_factory"?

lchrusciel: I would vote for "**decorated**_factory"

michalmarcinkowski: For me `sylius.factory.promotion_rule.decorated` was fine.

Zales0123: According to [Symfony documentation](http://symfony.com/doc/current/service_container/service_decoration.html) we should keep convention like ``decorating_factory``. I've changed them to ``custom_factory`` as it was done this way in most cases.
However ``decorating_factory`` or ``custom_factory`` doesn't says too much about the service - maybe in fact we should think about something more sophisticated.

lchrusciel: Why? For me `sylius.decorating_factory.promotion_rule` seems to be sophisticated enough.

pjedrzejewski: I'd follow Symfony's convention. 👍  I did not know there is any.

PR #7142, opened 2016-12-21
[Core] [DX] Extract payment target transition resolver

### LINKED ISSUE #7137: [Core][DX] Make private properties and methods in StateResolvers protected
Hello,
The provided StateResolvers like https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/StateResolver/OrderPaymentStateResolver.php have only private and public properties and methods. 

This make more difficult to extend them as you almost need to duplicate all code in the classes. And  extending this is probably needed if you add more states in place. 

I suggest to change those private to protected. ¿Thoughts?

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7137  |
| License         | MIT |

The only question is should we introduce same concept for [shipping](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/StateResolver/OrderShippingStateResolver.php#L53-L59) and [order](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/StateResolver/OrderStateResolver.php#L56-L62) state resolvers?

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services/state_resolvers.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services/state_resolvers.xml
@@ -18,9 +18,11 @@
         </service>
         <service id="sylius.state_resolver.order_payment" class="Sylius\Component\Core\StateResolver\OrderPaymentStateResolver">
             <argument type="service" id="sm.factory" />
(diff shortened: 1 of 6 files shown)


lchrusciel:
Not needed after beta2

PR #7134, opened 2016-12-20
[ProductAssociationType] Translatable product association type's name

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |

PR #7143, opened 2016-12-21
[ProductAssociationType] Fix filtering product association types by name

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | fixes after #7134 |
| License         | MIT |

### DIFF
--- a/features/product/managing_product_association_types/filtering_product_association_types.feature
+++ b/features/product/managing_product_association_types/filtering_product_association_types.feature
@@ -1,0 +1,24 @@
+@managing_product_association_types
+Feature: Filtering product association types
+    In order to filter product association types
+    As an Administrator
+    I want to be able to filter product association types on the list
+
+    Background:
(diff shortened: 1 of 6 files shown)


features/product/managing_product_association_types/filtering_product_association_types.feature
```diff
@@ -0,0 +1,24 @@
+@managing_product_association_types
+Feature: Filtering product association types
+    In order to filter product association types by name or code
```
lchrusciel: Do we have to be such precise? If new feature would be added should be change also description?
`In order to filter product association types`

GSadee: IMO new feature =~ new file but in this case I agree with you and I will change that

pjedrzejewski:
👍 👍 👍 

Conclusion, everything needs to be convered! 🔥

PR #7147, opened 2016-12-21
[ProductOption] Sorting product options by position

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |

### DIFF
--- a/app/migrations/Version20161221133514.php
+++ b/app/migrations/Version20161221133514.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Sylius\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * @author Grzegorz Sadowski <grzegorz.sadowski@lakion.com>
+ */
+class Version20161221133514 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_product_option ADD position INT NOT NULL');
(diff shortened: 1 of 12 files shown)


pjedrzejewski:
The migration should fill positions on existing product options, otherwise it will break ones database if it contains anything.

GSadee:
@pjedrzejewski Updated, please look if it is a correct solution.

app/migrations/Version20161221133514.php
```diff
@@ -0,0 +1,34 @@
...
+class Version20161221133514 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_product_option ADD position INT NOT NULL');
```
pjedrzejewski: I am pretty sure the migration will fail here, because of not null. Load fixtures on master and switch to this branch, run the migrations and see if it works. :)

GSadee: I of course tested that in this way and it was working. The default position which is set after adding a column is 0.

src/Sylius/Behat/Context/Setup/ProductOptionContext.php
```diff
@@ -98,17 +103,49 @@ public function thisProductOptionHasTheOptionValueWithCode(
...
     }
+
+    /**
+     * @param string $name
+     *
+     * @return string
+     */
+    private function generateCodeFromName($name)
+    {
+        return StringInflector::nameToCode($name);
```
Zales0123: `$this->generateCodeFromName($name)` is even longer than `StringInflector::nameToCode($name)`, why extracting this?

pjedrzejewski:
Thank you Grzesiu!

PR #7150, opened 2016-12-22
[Shop] Sorting products by name

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | nox |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

I add sorting products by name in default locale for a current channel. 

However we have a problem with sorting and also displaying entities in grids in different than default locale, because our queryBuilders choose entities for only one locale. And when we switch to another locale in shop, the products on the list are not translated because queryBuilders do not select other translations.


pjedrzejewski:
Thank you Grzesiu!

PR #7151, opened 2016-12-22
[Resource] Allow to inject IdentityTranslator to flash helper

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

A default [symfony translator](https://github.com/symfony/translation/blob/master/IdentityTranslator.php#L19) implementation doesn't implement *TranslatorBagInterface* but it is require by FlashHelper. This fix will allow to handle all symfony translator classes

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/FlashHelper.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/FlashHelper.php
@@ -15,6 +15,7 @@ use Sylius\Bundle\ResourceBundle\Event\ResourceControllerEvent;
 use Sylius\Component\Resource\Model\ResourceInterface;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Translation\TranslatorBagInterface;
+use Symfony\Component\Translation\TranslatorInterface;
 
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
Thank you Łukasz!

src/Sylius/Bundle/ResourceBundle/Controller/FlashHelper.php
```diff
@@ -129,8 +130,12 @@ private function getResourceMessage($actionName)
...
     {
-        $defaultCatalogue = $this->translator->getCatalogue($locale);
+        if ($this->translator instanceof TranslatorBagInterface) {
+            $defaultCatalogue = $this->translator->getCatalogue($locale);
 
-        return $defaultCatalogue->has($message, 'flashes');
+            return $defaultCatalogue->has($message, 'flashes');
+        }
+
+        return false;
```
pamil: Shouldn't we try to translate the message and compare it with the result?

```php
return $this->translator->translate($message, 'flashes') !== $message;
```

PR #7152, opened 2016-12-22
[Admin][Resource] Create and add another feature

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | |
| License         | MIT |

zrzut ekranu 2016-12-22 o 15 54 49

To achieve "create and add another" functionality, I've changed slightly generation of routes in *ResourceBundle*. 🐃 
Previously redirect action name has been passed for both **create** and **update**: `redirect: index`. After my changes, both of them have separate redirect parameter:
```yml
redirect:
    create: index
    update: update
```

It's also possible to set a variable as redirect parameter:
```yml
redirect:
    create: $redirect
    update: update
```
Thanks two that, I could modify create form buttons and add one more "Create and add another". Each of them passes different value to route `redirect` parameter, e.g.:
```yml
{{ path('sylius_admin_promotion_create', {'redirect': 'referer'}) }}
{{ path('sylius_admin_promotion_create', {'redirect': 'sylius_admin_promotion_update'}) }}
```
Of course, to use this feature `$redirect` variable must be set as **create** action redirect parameter in routing generation.

If it's needed to avoid new "Create and add another" button in create form, following configuration should be passed to `vars` parameter in routing generation.
```yml
create:
    add_another: false
```

I've added some scenarios for this feature, but only of promotions, product and variants - there is just no need to test it in all places, as it works exactly the same.

To-do:
- [x] describe changes in UPDATE.md
- [x] update documentation
- [ ] think about better way to handle redirects with parameters routes (now it's done with additional routes, only 2, but still smells really bad :/)


src/Sylius/Behat/Page/Admin/Product/CreateSimpleProductPage.php
```diff
@@ -209,6 +209,7 @@ public function selectShippingCategory($shippingCategoryName)
     }
 
 
+
```
Zales0123: 🎉 🌮

src/Sylius/Bundle/AdminBundle/Resources/config/routing/product_variant.yml
```diff
@@ -42,6 +42,29 @@ sylius_admin_product_variant_create:
                     form: "@SyliusAdmin/ProductVariant/_form.html.twig"
                     breadcrumb: "@SyliusAdmin/ProductVariant/Create/_breadcrumb.html.twig"
 
+sylius_admin_product_variant_create_and_add_another:
+    path: /another-new
```
Zales0123: I know, it looks like really rotten :potato:, but for now I have no better idea to handle parametrized redirects with parameters :/ Hope somebody can help me with that 😄

gabiudrescu:
if this brings a BC on the table, is this really necessary?

michalmarcinkowski:
@gabiudrescu it's quite handy feature. I know it's annoying since probably requires changes in many places, but don't you think it's worth this small BC break?

gabiudrescu:
I don't see it handy as I am looking at it from the perspective of an eCommerce business with some history. We do not create products manually, but rather import them from other sources - e.g. ERP, product feeds etc.

Imho, product creation is usually a bulk action done via import from other systems, rather than clicking new over and over again. Even so, by being redirected to "index" after creation, you can easily click 

My vote is against mainly because: 
 - I don't find it that useful for our use cases
 - the focus should be on making the platform stable, rather than introducing new features; I expect beta releases to adhere to a feature freeze, especially if it brings BC.
 - in a MoSCoW, I would put this into Could Have - right at the bottom.

But that's just me :grin:

crevillo:
While agree that more important thing now is to make the platform stable, i don't agree is that usual to have bulk imports specially if thinking in very small commerce that lacks of erps and just want something to go online and sell things...

gabiudrescu:
@crevillo yeah, I hope I put enough emphasis on my comment above that's about my opinion, that I do not think it's general available.. but is worth sharing so that we discuss it with others.

anyway, based on previous work experiences, usually products were created in bulk by some kind of import. either via CSV, crafted in Excel by working bees from content management, or by import from ERPs, where someone else was creating the basic product entity (price, tax, basic title) required for orders and accounting. 

to make it clear: I think that if this feature is ready to go live now and the BC is small, the deed is done - it shouldn't be left without merge to master. it would be even more wasteful.

michalmarcinkowski:
Both of you are right about this feature and every opinion is valuable, so @gabiudrescu and @crevillo thanks for your feedback 👍  

Since this feature is already done, we will merge it to the master. We plan to build backlog more transparently to the community to have a chance to gather feedback and decide, as well as prioritise it based on community needs. I hope we share more about that soon, any suggestions about the process and tools are welcome 😉

pjedrzejewski:
@gabiudrescu You are absolutely right, I was hoping we can do this improvement at low cost and with no BC breaks, but we tried to figure it out with @Zales0123 and this is simplest solution we have found. As this is not critical, I will wait with the merge. Maybe someone gets enlightened and provides much simpler solution. (I would put a bet on JS being the key)

Thank you so much for your feedback all!

Zales0123:
@pjedrzejewski @pamil are we willing to introduce such a feature into next versions of **Sylius**? Of course we should think about how to not introduce BC breaks, but maybe it's worth to think about it. I could rewrite/rebase it in my free time :)

gabiudrescu:
@Zales0123 may I propose you some other stuff for that free time :grimacing: ?

Zales0123:
Ping @pjedrzejewski @pamil, my question is still valid 🚀 
@gabiudrescu unfortunately, it seems that free time become very luxury resource for me last time 😄

PR #7158, opened 2016-12-22
[RFC] [Core] [HotFix] Add core bundle root dir parameter

### LINKED ISSUE #7156: Path not good when using Sylius as a vendor package
This config line here seems to break the whole app when using Sylius as a dependency : https://github.com/Sylius/Sylius/blob/70e078142bf015fd3f85be43fc15de7e8248894d/src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml#L36

The following error occurs : 
``` 
Specified non-existing directory "/my/project/app/../src/Sylius/Bundle/ResourceBundle/Model" as Doctrine mapping source.
``` 

Should we override this setting or is there a better way ?

Edit : I'm on the latest master and this never happened before the last `composer update`

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7156  |
| License         | MIT |

It just a proposal and a hot fix. I didn't find a better one. But without it every Sylius based project is broken so it is still better than nothing.

Another possible solution is to this config dynamically in extension. Or extract this config to main app folder and prepare a documentation how to add it. 

Another thing. Should we extract this config to resource bundle?

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/DependencyInjection/SyliusCoreExtension.php
+++ b/src/Sylius/Bundle/CoreBundle/DependencyInjection/SyliusCoreExtension.php
@@ -64,6 +64,8 @@ final class SyliusCoreExtension extends AbstractResourceExtension implements Pre
         if ('test' === $env || 'test_cached' === $env) {
             $loader->load('test_services.xml');
         }
+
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
Thank you Łukasz!

PR #7164, opened 2016-12-27

SyliusBot:


pamil:
Looks like CrowdIn is messing with apostrophes 🌮

michalmarcinkowski:
@pamil can we do something about that?

PR #7171, opened 2016-12-29
set max per page on grid view

### LINKED ISSUE #4697: [Grid] Paging is not supported
@pjedrzejewski was this an accidental omission in your implementing?

loic425:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #4697 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
@@ -14,6 +14,7 @@ namespace Sylius\Bundle\ResourceBundle\Controller;
 use Hateoas\Configuration\Route;
 use Hateoas\Representation\Factory\PagerfantaFactory;
 use Pagerfanta\Pagerfanta;
+use Sylius\Bundle\ResourceBundle\Grid\View\ResourceGridView;
 use Sylius\Component\Resource\Repository\RepositoryInterface;
 
 /**
@@ -48,13 +49,20 @@ final class ResourcesCollectionProvider implements ResourcesCollectionProviderIn
     {
         $resources = $this->resourcesResolver->getResources($requestConfiguration, $repository);
 
-        if ($resources instanceof Pagerfanta) {
+        if ($resources instanceof ResourceGridView or $resources instanceof Pagerfanta) {
             $request = $requestConfiguration->getRequest();
-            $resources->setMaxPerPage($requestConfiguration->getPaginationMaxPerPage());
-            $resources->setCurrentPage($request->query->get('page', 1));
+
+            if ($resources instanceof ResourceGridView) {
+                $data = $resources->getData();
+            } else {
+                $data = $resources;
+            }
+
+            $data->setMaxPerPage($requestConfiguration->getPaginationMaxPerPage());
+            $data->setCurrentPage($request->query->get('page', 1));
 
             // This prevents Pagerfanta from querying database from a template
-            $resources->getCurrentPageResults();
+            $data->getCurrentPageResults();
 
             if (!$requestConfiguration->isHtmlRequest()) {
                 $route = new Route($request->attributes->get('_route'), array_merge($request->attributes->get('_route_params'), $request->query->all()));
--- a/src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourcesCollectionProviderSpec.php
+++ b/src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourcesCollectionProviderSpec.php
@@ -21,6 +21,7 @@ use Sylius\Bundle\ResourceBundle\Controller\RequestConfiguration;
 use Sylius\Bundle\ResourceBundle\Controller\ResourcesCollectionProvider;
 use Sylius\Bundle\ResourceBundle\Controller\ResourcesCollectionProviderInterface;
 use Sylius\Bundle\ResourceBundle\Controller\ResourcesResolverInterface;
+use Sylius\Bundle\ResourceBundle\Grid\View\ResourceGridView;
 use Sylius\Component\Resource\Model\ResourceInterface;
 use Sylius\Component\Resource\Repository\RepositoryInterface;
 use Symfony\Component\HttpFoundation\ParameterBag;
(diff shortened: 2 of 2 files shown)


src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
```diff
@@ -48,13 +49,20 @@ public function get(RequestConfiguration $requestConfiguration, RepositoryInterf
     {
         $resources = $this->resourcesResolver->getResources($requestConfiguration, $repository);
 
-        if ($resources instanceof Pagerfanta) {
+        if ($resources instanceof ResourceGridView or $resources instanceof Pagerfanta) {
```
Zales0123: You should use `||` instead of `or` in condition.

src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
```diff
@@ -48,13 +49,20 @@ public function get(RequestConfiguration $requestConfiguration, RepositoryInterf
...
             $request = $requestConfiguration->getRequest();
-            $resources->setMaxPerPage($requestConfiguration->getPaginationMaxPerPage());
-            $resources->setCurrentPage($request->query->get('page', 1));
+
+            if ($resources instanceof ResourceGridView) {
+                $data = $resources->getData();
+            } else {
+                $data = $resources;
+            }
```
Zales0123: `$data = ($resources instanceof ResourceGridView) ? $resources->getData() : $resources;` wdyt?

src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourcesCollectionProviderSpec.php
```diff
@@ -116,4 +117,30 @@ function it_creates_a_paginated_representation_for_pagerfanta_for_non_html_reque
 
         $this->get($requestConfiguration, $repository)->shouldReturn($paginatedRepresentation);
     }
+
+    function it_handles_ResourceGridView(
```
Zales0123: `it_handles_resource_grid_view`

src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
loic425: yes, i'ts more concise.
What do you think about renaming $data into $paginator ?

src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
pamil: Why don't we just add:

```php
if ($resources instanceof ResourceGridView) {
    $resources = $resources->getData();
}
```

before this if?

src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
```diff
@@ -48,13 +49,20 @@ public function get(RequestConfiguration $requestConfiguration, RepositoryInterf
...
         $resources = $this->resourcesResolver->getResources($requestConfiguration, $repository);
 
-        if ($resources instanceof Pagerfanta) {
+        if ($resources instanceof ResourceGridView or $resources instanceof Pagerfanta) {
             $request = $requestConfiguration->getRequest();
-            $resources->setMaxPerPage($requestConfiguration->getPaginationMaxPerPage());
-            $resources->setCurrentPage($request->query->get('page', 1));
+
+            if ($resources instanceof ResourceGridView) {
+                $data = $resources->getData();
```
pamil: Is there any guarantee that `ResourceGridView::getData()` will return `Pagerfanta` object?

pamil: Applying my earlier comment will solve this issue.

src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
loic425: yes seems to be good ! Moreover, we can ensure $resources is an instance of paginator below.

loic425: We cannot do that, the method need to return ResourceGridView.
I'll propose something in a minute.

src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
```diff
@@ -48,18 +49,20 @@ public function get(RequestConfiguration $requestConfiguration, RepositoryInterf
     {
         $resources = $this->resourcesResolver->getResources($requestConfiguration, $repository);
 
-        if ($resources instanceof Pagerfanta) {
+        $data = $resources instanceof ResourceGridView ? $resources->getData() : $resources;
```
pamil: IMHO `$resources` was more specific variable name than `$data`

loic425: yep, but the "get" method needs to return ResourceGridView when $resources is an instance of ResourceGridView. So if I overwrite $resources var, the method would return Pagerfanta instance.

loic425: the alternative is to rename $data into $paginator

lchrusciel: `$paginator` seems to be the most reasonable

loic425: @pamil what do you think about that ?

pamil: `$paginator` is misleading as it may be just an `iterable` instead of paginator.

I think we can leave it now as it is, but the real issue is `$this->resourcesResolver->getResources()` returning `mixed`.

pjedrzejewski:
Thank you @loic425!

PR #7186, opened 2017-01-04
[Core] ChannelPricing configuration fix

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes, but previous configuration doesn't work either |
| Related tickets | fixes a bug from slack support channel |
| License         | MIT |

An issue posted on Slack. According to @ivanmatas:
guys, when I try to override form for channelPricing I get error that it expects array but got string
```yml
sylius_core:
    driver: doctrine/orm
    resources:
        channel_pricing:
            classes:
                model: AppBundle\Entity\ChannelPricing
                form: AppBundle\Form\Type\Product\ChannelPricingType
```
and if I use 
form:
   default: 
then it's not overriding it...could someone explain is it a bug or?


pjedrzejewski:
Thank you Łukasz!

PR #7189, opened 2017-01-04
Fix product add/remove attribute

Pitoune:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

### DIFF
--- a/src/Sylius/Component/Product/Model/Product.php
+++ b/src/Sylius/Component/Product/Model/Product.php
@@ -248,7 +248,7 @@ class Product implements ProductInterface
     public function addAttribute(AttributeValueInterface $attribute)
     {
         if (!$this->hasAttribute($attribute)) {
-            $attribute->setProduct($this);
+            $attribute->setSubject($this);
             $this->attributes->add($attribute);
         }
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
@Pitoune I recall that there is an alias method for both setter and getter. Did you check? Also, what is the problem here, just naming or you experience some bug?

Pitoune:
My bad, I did not see `Sylius\Component\Product\Model\ProductAttributeValue` which fixes my problem.

David-Guillot:
@pjedrzejewski there's still an issue there: `setProduct` is defined in ProductAttributeValueInterface (which extends AttributeValueInterface), but only AttributeValueInterface is required as a parameter. So there are 2 possible fixes:
* Either what @Pitoune did (but you should have also changed [these unit tests](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Product/spec/Model/ProductSpec.php#L123-L142))
* Or change the required interface, which would be more logical IMHO

PR #7196, opened 2017-01-05
Fixed some Product model methods requirements

David-Guillot:
In Product model, methods interacting with Attributes require Sylius\Component\Attribute\Model\AttributeValueInterface, instead of Sylius\Component\Product\Model\ProductAttributeValueInterface, which exists just for that and is the one used [by the ORM config](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ProductBundle/Resources/config/doctrine/model/Product.orm.xml#L27) and [by the tests](https://github.com/Sylius/Sylius/blob/570d66b1b064fc92d6dd115486077f13b3a14195/src/Sylius/Component/Product/spec/Model/ProductSpec.php#L123-L142).

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7189 |
| License         | MIT |


pjedrzejewski:
Could you add note to the UPGRADE file, please?

David-Guillot:
Sure @pjedrzejewski , done. Thanks for noticing this PR was indeed a BC break so requiring an UPGRADE note ;)

michalmarcinkowski:
These methods are from `Sylius\Component\Attribute\Model\AttributeSubjectInterface` which signature has `Sylius\Component\Attribute\Model\AttributeValueInterface` as argument, this change will not work.

I think that the only thing we can do is `assertInstanceOf` in the method.

David-Guillot:
Ah dammit, good call @michalmarcinkowski , I did not realize that. Closing this PR, I'll make another one when I'll have some time.

PR #7203, opened 2017-01-05
[Promotion] Fix promotion edit page

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

Fix for following issue:
```gherkin
Given this promotion gives "$10.00" discount to every order
When the store also operates on another channel named "EU-WEB"
Then I should be able to modify a "Christmas sale" promotion
```

resulted in: 
zrzut ekranu 2017-01-05 o 15 57 59


pjedrzejewski:
Great catch, thanks Łukasz! :)

PR #7031, opened 2016-12-07
[Shop] Free order 

tuka217:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | yes |
| BC breaks?      | no|
| Related tickets | fixes #5961, #4066 |
| License         | MIT |

PR #7210, opened 2017-01-07
[Free Order] Add missing constant state for checkout

tchapi:
| Q               | A
| --------------- | ---
| Bug fix?        | kindof |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | mentioned in #7031 |
| License         | MIT |

OrderCheckoutStates' constants should reflect the state machine — one constant was missing after the mentioned PR, this adds it back in.


pjedrzejewski:
Thank you @tchapi!

PR #7216, opened 2017-01-09
add a way to make object manager name consistent

videni:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| License         | MIT |

the object manager name created in AbstractResourceBundle is not consistent with the one created in AbstractResourceExtension, which makes us to add extra code to register model namespace.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/AbstractResourceBundle.php
+++ b/src/Sylius/Bundle/ResourceBundle/AbstractResourceBundle.php
@@ -54,7 +54,7 @@ abstract class AbstractResourceBundle extends Bundle implements ResourceBundleIn
                         case ResourceBundleInterface::MAPPING_YAML:
                             $container->addCompilerPass($compilerPassClassName::$compilerPassMethod(
                                 [$this->getConfigFilesPath() => $this->getModelNamespace()],
-                                [sprintf('%s.object_manager', $this->getBundlePrefix())],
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/ResourceBundle/AbstractResourceBundle.php
```diff
@@ -147,4 +147,12 @@ protected function getConfigFilesPath()
             strtolower($this->getDoctrineMappingDirectory())
         );
     }
+
+    /**
+     * @return string
+     */
+    protected function getObjectManagerParameter()
+    {
+      return  sprintf('%s.object_manager', $this->getBundlePrefix());
```
lchrusciel: There are 2 spaces instead of 4 🐋

pamil: There are 2 spaces instead of 1 🐋

pjedrzejewski: @videni Could you please fix that little CS and we are good to merge, thanks! :)

videni:
@pjedrzejewski, I think  it is done.

pjedrzejewski:
Thanks @videni!

PR #7224, opened 2017-01-09
[Shop][Product] Sort product options by position

jurv:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | Complementary to PR #7147  |
| License         | MIT |

Products options aren't sorted by position when displayed on product page. The PR #7147 fixed their order in the admin, but it seems logical that they should be displayed in the same order in the shop.

It seems to me like adding a "order-by" in the product mapping fixes this and doesn't have any other implications.


pjedrzejewski:
This is a correct fix, thank you so much Julien! :)

PR #7228, opened 2017-01-10
stop UserLastLoginSubscriber from executing several times when there …

videni:
…are multiple user entities

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

UserLastLoginSubscriber will execute as many time as the number of user resources, it's not  a problem if we use SyliusUserBundle in one database connection, but when we use it in two database connection, it would throw an error, because the orm mapping  of the two entity manager is different.

### DIFF
--- a/src/Sylius/Bundle/UserBundle/DependencyInjection/SyliusUserExtension.php
+++ b/src/Sylius/Bundle/UserBundle/DependencyInjection/SyliusUserExtension.php
@@ -81,10 +81,12 @@ final class SyliusUserExtension extends AbstractResourceExtension
     private function createServices(array $resources, ContainerBuilder $container)
     {
         foreach ($resources as $userType => $config) {
+            $userClass = $config['user']['classes']['model'];
+
             $this->createTokenGenerators($userType, $config['user'], $container);
             $this->createReloaders($userType, $container);
-            $this->createLastLoginListeners($userType, $container);
-            $this->createProviders($userType, $config['user']['classes']['model'], $container);
+            $this->createLastLoginListeners($userType, $userClass, $container);
+            $this->createProviders($userType, $userClass, $container);
             $this->createUserDeleteListeners($userType, $container);
         }
     }
@@ -203,15 +205,16 @@ final class SyliusUserExtension extends AbstractResourceExtension
 
     /**
      * @param string $userType
+     * @param string $userClass
      * @param ContainerBuilder $container
      */
-    private function createLastLoginListeners($userType, ContainerBuilder $container)
+    private function createLastLoginListeners($userType, $userClass, ContainerBuilder $container)
     {
         $managerServiceId = sprintf('sylius.manager.%s_user', $userType);
         $lastLoginListenerServiceId = sprintf('sylius.listener.%s_user_last_login', $userType);
 
         $lastLoginListenerDefinition = new Definition(UserLastLoginSubscriber::class);
-        $lastLoginListenerDefinition->addArgument(new Reference($managerServiceId));
+        $lastLoginListenerDefinition->setArguments([new Reference($managerServiceId), $userClass]);
         $lastLoginListenerDefinition->addTag('kernel.event_subscriber');
         $container->setDefinition($lastLoginListenerServiceId, $lastLoginListenerDefinition);
     }
--- a/src/Sylius/Bundle/UserBundle/EventListener/UserLastLoginSubscriber.php
+++ b/src/Sylius/Bundle/UserBundle/EventListener/UserLastLoginSubscriber.php
@@ -27,11 +27,18 @@ class UserLastLoginSubscriber implements EventSubscriberInterface
     protected $userManager;
 
     /**
-     * @param ObjectManager $userManager
+     * @var string
      */
-    public function __construct(ObjectManager $userManager)
+    private $userClass;
+
+    /**
+     * @param ObjectManager $userManager
+     * @param string $userClass
+     */
+    public function __construct(ObjectManager $userManager, $userClass)
     {
         $this->userManager = $userManager;
+        $this->userClass = $userClass;
     }
 
     /**
@@ -51,10 +58,7 @@ class UserLastLoginSubscriber implements EventSubscriberInterface
     public function onSecurityInteractiveLogin(InteractiveLoginEvent $event)
     {
         $user = $event->getAuthenticationToken()->getUser();
-
-        if ($user instanceof UserInterface) {
-            $this->updateUserLastLogin($user);
-        }
+        $this->updateUserLastLogin($user);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/Sylius/Bundle/UserBundle/DependencyInjection/SyliusUserExtension.php
```diff
@@ -81,10 +81,12 @@ private function resolveResources(array $resources, ContainerBuilder $container)
     private function createServices(array $resources, ContainerBuilder $container)
     {
         foreach ($resources as $userType => $config) {
+            $userClass=$config['user']['classes']['model'];
```
lchrusciel: CS

src/Sylius/Bundle/UserBundle/DependencyInjection/SyliusUserExtension.php
```diff
@@ -202,16 +204,22 @@ private function createReloaders($userType, ContainerBuilder $container)
     }
 
     /**
-     * @param string $userType
+     * @param $userType
```
lchrusciel: missing typehints

src/Sylius/Bundle/UserBundle/DependencyInjection/SyliusUserExtension.php
```diff
@@ -202,16 +204,22 @@ private function createReloaders($userType, ContainerBuilder $container)
...
      */
-    private function createLastLoginListeners($userType, ContainerBuilder $container)
+    private function createLastLoginListeners($userType, $userClass, ContainerBuilder $container)
     {
         $managerServiceId = sprintf('sylius.manager.%s_user', $userType);
         $lastLoginListenerServiceId = sprintf('sylius.listener.%s_user_last_login', $userType);
 
         $lastLoginListenerDefinition = new Definition(UserLastLoginSubscriber::class);
-        $lastLoginListenerDefinition->addArgument(new Reference($managerServiceId));
+        $lastLoginListenerDefinition->setArguments(
```
lchrusciel: Would fit one line

src/Sylius/Bundle/UserBundle/DependencyInjection/SyliusUserExtension.php
```diff
@@ -202,16 +204,22 @@ private function createReloaders($userType, ContainerBuilder $container)
...
+            [
+                new Reference($managerServiceId),
+                $userClass
```
lchrusciel: But if you insist on multiline declaration, there should be comma at the end

src/Sylius/Bundle/UserBundle/EventListener/UserLastLoginSubscriber.php
```diff
@@ -27,11 +27,16 @@ class UserLastLoginSubscriber implements EventSubscriberInterface
     protected $userManager;
 
     /**
+     * @var string
+     */
+    protected $userClass;
```
lchrusciel: CS

src/Sylius/Bundle/UserBundle/EventListener/UserLastLoginSubscriber.php
```diff
@@ -27,11 +27,16 @@ class UserLastLoginSubscriber implements EventSubscriberInterface
...
+    /**
      * @param ObjectManager $userManager
      */
-    public function __construct(ObjectManager $userManager)
+    public function __construct(ObjectManager $userManager, $userClass)
```
lchrusciel: Missing docblock declaration

videni:
@lchrusciel , the `UserLastLoginSubscriberSpec` will fail , I don't  know how to test this line `$user instanceof $this->userClass`

```

    /**
     * @param InteractiveLoginEvent $event
     */
    public function onSecurityInteractiveLogin(InteractiveLoginEvent $event)
    {
        $user = $event->getAuthenticationToken()->getUser();

        if ($user instanceof $this->userClass) {
            $this->updateUserLastLogin($user);
        }
    }
```

can you show me how to do it?

lchrusciel:
@videni let me know, if my snippet would not be enough for test, we will figure out something more ;)

videni:
@lchrusciel, I think it should be constructed with specific user class, such as AdminUser, ShopUser.  I have changed the code, the test is passed,  but it should also cover the case when the user  is not instance of that specific user class.

src/Sylius/Bundle/UserBundle/spec/EventListener/UserLastLoginSubscriberSpec.php
```diff
@@ -17,7 +17,7 @@
 use Sylius\Bundle\UserBundle\Event\UserEvent;
 use Sylius\Bundle\UserBundle\EventListener\UserLastLoginSubscriber;
 use Sylius\Bundle\UserBundle\UserEvents;
-use Sylius\Component\User\Model\UserInterface;
+use Sylius\Component\Core\Model\ShopUser;
```
lchrusciel: This class is part of User bundle, so we cannot couple it with core. For the basic tests we should use `Sylius\Component\User\Model\User` class. And then, add another test, something like:
```php
     function it_updates_user_last_login_on_security_interactive_login_only_for_class_defined_in_constructor(
          ObjectManager $userManager,
          InteractiveLoginEvent $event,
          TokenInterface $token,
      ) {
          $event->getAuthenticationToken()->willReturn($token);
          $token->getUser()->willReturn(new \stdClass());
 
         $user->setLastLogin(Argument::type(\DateTime::class))->shouldNotBeCalled();
 
         $userManager->persist($user)->shouldNotBeCalled();
         $userManager->flush()->shouldNotBeCalled();
 
         $this->onSecurityInteractiveLogin($event);
     }
```

lchrusciel:
@videni we cannot check it with shop or admin user, because we work in narrowed scope of user bundle only. But the logic in general is simple: 'Do some things if class match. Otherwise nothing should happen'. We can assert it with only User model and stdClass for example

videni:
@lchrusciel,thanks , I changed the code, please consider merge this PR

pamil:
@lchrusciel there is no `$user` variable available.

videni:
@lchrusciel ,  your test is not correct,  we need to feed a specific user class[ShopUser] to `UserLastLoginSubscriber`  constructor ,  and use another concrete user class [AdminUser] to test that case.  since `UserLastLoginSubscriberSpec` is in SyliusUserBundle, we don't have ShopUser and AdminUser, maybe the ` phpspec double ` can save this situation, I don't know how to write this test, can you help me do that?

pjedrzejewski:
Thank you @videni, nice work! :)

lchrusciel:
@videni you are right, but I was almost correct. I have prepared a PR with missing spec

PR #7235, opened 2017-01-11
[User] Finalize listeners

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |

During review of #7228, I have been wondering why are these classes not final. I know that it is a bc break, but we people should provide their own implementation of any of these classes if they want to change a logic. And classes are max 100 lines long.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -141,6 +141,8 @@ After:
 
 ### User / UserBundle
 
+* All classes under `Sylius\Bundle\UserBundle\EventListener` have been made final. Provide your custom implementation if you need to change theirs logic.
+
 ## Application:
 
 ### Configuration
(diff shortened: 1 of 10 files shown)


UPGRADE.md
```diff
@@ -141,6 +141,8 @@ After:
 
 ### User / UserBundle
 
+* All classes under `Sylius\Bundle\UseBundle\EventListener` have been made final. Provide your custom implementation if you need to change theirs logic.
```
pamil: Use**r**

michalmarcinkowski:
I see some major DX drawbacks of this approach. We no longer have a single listener responsible for a single thing. In order to use e.g. different `PasswordUpdater` I have to change it for two services. What is more, I have to know that there are two `PasswordUpdaterListener`s and that I need to modify both of them. What is more I can't immediately see on what events the password is updated, because there are two listeners attached to a different events (similar problem, know about both of them and modify them both).

Not sure if benefit of `final` is bigger than introducing confusion and bc break, WDYT?

PR #7243, opened 2017-01-12
[API] Versioning for serialization

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |


src/Sylius/Bundle/ResourceBundle/test/src/Tests/Responses/Expected/comic-books/versioned_index_response.json
```diff
@@ -0,0 +1,34 @@
...
+            "href": "\/v1.2\/comic-books\/?page=1&limit=10"
+        },
+        "last": {
+            "href": "\/v1.2\/comic-books\/?page=1&limit=10"
+        }
+    },
+    "_embedded": {
+      "items": [
+        {
+          "id":@integer@,
```
pamil: Shouldn't there be a space between `:` and `@`?

src/Sylius/Bundle/ResourceBundle/test/src/Tests/Controller/ComicBookApiTest.php
```diff
@@ -95,31 +116,55 @@ public function it_allows_showing_a_comic_book()
     {
         $objects = $this->loadFixturesFromFile('comic_books.yml');
 
-        $this->client->request('GET', '/comic-books/'. $objects["comic-book1"]->getId());
+        $this->client->request('GET', 'v1/comic-books/'. $objects["comic-book1"]->getId());
```
pamil: After adding version to the paths, there are absolute paths `/v1/comic-books/` and relative paths `v1/comic-books/` all over the API tests, we should unify that (probably to absolute ones).

pjedrzejewski:
Awesome, thanks Łukasz!

app/config/security.yml
```diff
@@ -28,7 +28,7 @@ security:
             anonymous: true
 
         oauth_token:
-            pattern:  ^/api/oauth/v2/token
+            pattern:  ^/api/v1/oauth/v2/token
```
michalmarcinkowski: Should be reverted. The oauth is a standard so the response won't change until we change the authorization method, but then the path will also change since `/oauth/v2` will no longer be appropriate.

docs/api/authorization.rst
```diff
@@ -52,7 +52,7 @@ Definition
 
 .. code-block:: text
 
-    GET /api/oauth/v2/token
+    GET /api/v1/oauth/v2/token
```
michalmarcinkowski: As above.

app/config/security.yml
lchrusciel: Fixed

PR #7244, opened 2017-01-12
[User] Add missing test for UserLastLoginSubscriber

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | missing test from #7228  |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/UserBundle/spec/EventListener/UserLastLoginSubscriberSpec.php
+++ b/src/Sylius/Bundle/UserBundle/spec/EventListener/UserLastLoginSubscriberSpec.php
@@ -79,4 +79,20 @@ final class UserLastLoginSubscriberSpec extends ObjectBehavior
 
         $this->onImplicitLogin($event);
     }
+
+    function it_updates_only_user_specified_in_constructor(
+        ObjectManager $userManager,
+        UserEvent $event,
+        UserInterface $user
+    ) {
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/UserBundle/spec/EventListener/UserLastLoginSubscriberSpec.php
```diff
@@ -79,4 +79,20 @@ function it_updates_user_last_login_on_implicit_login(
 
         $this->onImplicitLogin($event);
     }
+
+    function it_updates_only_user_specified_in_constructor(
+        ObjectManager $userManager,
+        UserEvent $event,
+        UserInterface $user
+    ) {
+        $this->beConstructedWith($userManager, 'Sylius\Component\User\Model\User');
```
pamil: This brings me a huuuuge WTF, we create a mock of `Sylius\Component\User\Model\UserInterface` and set the user class to `Sylius\Component\User\Model\User` (which implements `Sylius\Component\User\Model\UserInterface`) so it looks very confusing.

Can we use any other classname? Like `AppBundle\Model\CustomUser`?

videni:
😀,nice

pjedrzejewski:
👍 👍 👍

Issue #7221, opened 2017-01-09
shouldn't hardcode the  'doctrine.dbal.default_connection' service


videni:
src/Sylius/Bundle/ResourceBundle/Resources/config/services/integrations/doctrine/orm.xmll

```
 <service id="sylius.grid_driver.doctrine.dbal" class="Sylius\Bundle\GridBundle\Doctrine\DBAL\Driver">
            <argument type="service" id="doctrine.dbal.default_connection" />
            <tag name="sylius.grid_driver" alias="doctrine/dbal" />
        </service>
```
In our case , the 'default' connection is used for other purpose, it is not sylius database, I think this should be configurable.

so does this service

src/Sylius/Bundle/ResourceBundle/Resources/config/services/integrations/translation.xml
```
        <service id="sylius.translation.translatable_listener.doctrine.orm" class="Sylius\Bundle\ResourceBundle\EventListener\ORMTranslatableListener">
            <argument type="service" id="sylius.resource_registry" />
            <argument type="service" id="service_container" />
            <tag name="doctrine.event_subscriber" connection="default" priority="99" />
        </service>
```

PR #7250, opened 2017-01-13
make sylius.average_rating_updater service use sylius.manager.product…

videni:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

the `sylius.average_rating_updater ` should use ` sylius.manager.product_review` instead of the default entity manager service.

### DIFF
--- a/src/Sylius/Bundle/ReviewBundle/Resources/config/services.xml
+++ b/src/Sylius/Bundle/ReviewBundle/Resources/config/services.xml
@@ -27,7 +27,7 @@
         <service id="sylius.average_rating_calculator" class="Sylius\Component\Review\Calculator\AverageRatingCalculator" />
         <service id="sylius.average_rating_updater" class="Sylius\Bundle\ReviewBundle\Updater\AverageRatingUpdater">
             <argument type="service" id="sylius.average_rating_calculator" />
-            <argument type="service" id="doctrine.orm.default_entity_manager" />
+            <argument type="service" id="sylius.manager.product_review" />
         </service>
     </services>
 </container>


src/Sylius/Bundle/ReviewBundle/Resources/config/services.xml
```diff
@@ -27,7 +27,7 @@
         <service id="sylius.average_rating_calculator" class="Sylius\Component\Review\Calculator\AverageRatingCalculator" />
         <service id="sylius.average_rating_updater" class="Sylius\Bundle\ReviewBundle\Updater\AverageRatingUpdater">
             <argument type="service" id="sylius.average_rating_calculator" />
-            <argument type="service" id="doctrine.orm.default_entity_manager" />
+            <argument type="service" id="sylius.manager.product_review" />
```
lchrusciel: Thanks God, to have Travis CI, I have approved this changes, but I just have checked the failing build. And we cannot do this change, because it would couple ReviewBundle with ProductBundle....

videni: I also noticed that, i am trying to figure out a way to do that , do you have any suggestion please?

videni: @lchrusciel, there are several places hard-coded default entity manager connection, #7221, we can let developer to configure it at configuration files. besides this, do you have a easy way?

lchrusciel: The answer was too long, so I have prepared a PR with it. The question is if my solution is clean enough... Or maybe we have some better one.

PR #7254, opened 2017-01-13
Fix root option on TaxonChoiceType

Pitoune:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
+++ b/src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
@@ -12,7 +12,6 @@
 namespace Sylius\Bundle\TaxonomyBundle\Doctrine\ORM;
 
 use Sylius\Bundle\ResourceBundle\Doctrine\ORM\EntityRepository;
-use Sylius\Component\Taxonomy\Model\TaxonInterface;
 use Sylius\Component\Taxonomy\Repository\TaxonRepositoryInterface;
 
 /**
@@ -87,7 +86,7 @@ class TaxonRepository extends EntityRepository implements TaxonRepositoryInterfa
     /**
      * {@inheritdoc}
      */
-    public function findNodesTreeSorted()
+    public function findNodesTreeSorted($rootCode = null)
     {
         return $this->createQueryBuilder('o')
             ->addOrderBy('o.root')
@@ -96,6 +95,16 @@ class TaxonRepository extends EntityRepository implements TaxonRepositoryInterfa
             ->getQuery()
             ->getResult()
         ;
+
+        if (null !== $rootCode) {
+            $queryBuilder
+                ->join('o.root', 'root')
+                ->andWhere('root.code = :rootCode')
+                ->setParameter('rootCode', $rootCode)
+            ;
+        }
+
+        return $queryBuilder->getQuery()->getResult();
     }
 
     /**
--- a/src/Sylius/Bundle/TaxonomyBundle/Form/Type/TaxonChoiceType.php
+++ b/src/Sylius/Bundle/TaxonomyBundle/Form/Type/TaxonChoiceType.php
@@ -16,7 +16,6 @@ use Sylius\Component\Taxonomy\Repository\TaxonRepositoryInterface;
 use Symfony\Bridge\Doctrine\Form\DataTransformer\CollectionToArrayTransformer;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\ChoiceList\View\ChoiceView;
-use Symfony\Component\Form\Extension\Core\ChoiceList\ObjectChoiceList;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Form\FormInterface;
@@ -81,21 +80,15 @@ final class TaxonChoiceType extends AbstractType
         $resolver
             ->setDefaults([
                 'choices' => function (Options $options) {
-                    $taxons = $this->taxonRepository->findNodesTreeSorted();
-
-                    if (null !== $options['filter']) {
-                        $taxons = array_filter($taxons, $options['filter']);
-                    }
-
-                    return $taxons;
+                    return $this->getTaxons($options['root_code'], $options['filter']);
                 },
                 'choice_value' => 'id',
                 'choice_label' => 'name',
                 'choice_translation_domain' => false,
-                'root' => null,
+                'root_code' => null,
                 'filter' => null,
             ])
-            ->setAllowedTypes('root', [TaxonInterface::class, 'string', 'null'])
+            ->setAllowedTypes('root_code', ['string', 'null'])
             ->setAllowedTypes('filter', ['callable', 'null'])
         ;
     }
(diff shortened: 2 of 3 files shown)


src/Sylius/Bundle/TaxonomyBundle/Form/Type/TaxonChoiceType.php
```diff
@@ -81,7 +80,14 @@ public function configureOptions(OptionsResolver $resolver)
         $resolver
             ->setDefaults([
                 'choices' => function (Options $options) {
-                    $taxons = $this->taxonRepository->findNodesTreeSorted();
+                    if (null !== $options['root']) {
+                        if ($options['root'] instanceof TaxonInterface) {
```
lchrusciel: Isn't it too much magic? Why don't we stay with only taxon code in `root`?

src/Sylius/Bundle/TaxonomyBundle/Form/Type/TaxonChoiceType.php
```diff
@@ -81,7 +80,14 @@ public function configureOptions(OptionsResolver $resolver)
         $resolver
             ->setDefaults([
                 'choices' => function (Options $options) {
-                    $taxons = $this->taxonRepository->findNodesTreeSorted();
+                    if (null !== $options['root']) {
```
lchrusciel: Could be extracted to private method

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
```diff
@@ -43,6 +42,24 @@ public function findChildren($parentCode)
     /**
      * {@inheritdoc}
      */
+    public function findChildrenByRootCode($rootCode)
+    {
+        $queryBuilder = $this->createQueryBuilder('o');
```
lchrusciel: Redundant

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
```diff
@@ -23,20 +22,30 @@ class TaxonRepository extends EntityRepository implements TaxonRepositoryInterfa
...
-            ->setParameter('parentCode', $parentCode)
+            ->setParameter('code', $parentCode)
         ;
 
+        if ($isParentRoot) {
+            $queryBuilder
+                ->leftJoin('o.root', 'root')
+                ->andWhere('root.code = :code')
+            ;
+        } else {
```
Zales0123: We should avoid this nasty `else` statement. Would be a good to extract this to private method. But also there is a lot of logic in this `if`... So maybe two different methods, `findChildren` and `findChildrenForRootParent`?

src/Sylius/Bundle/TaxonomyBundle/Form/Type/TaxonChoiceType.php
```diff
@@ -107,4 +100,25 @@ public function getBlockPrefix()
     {
         return 'sylius_taxon_choice';
     }
+
+    /**
+     * @param null|string   $root
+     * @param null|callable $filter
```
Zales0123: `string|null` and `callable|null`. Also we should not align parameters vertically.

src/Sylius/Bundle/TaxonomyBundle/Form/Type/TaxonChoiceType.php
```diff
@@ -107,4 +100,25 @@ public function getBlockPrefix()
...
+     *
+     * @return array|TaxonInterface[]
```
Zales0123: `TaxonInterface[]` would be sufficient, `array|` indicates as there could be something else ;)

src/Sylius/Bundle/TaxonomyBundle/Form/Type/TaxonChoiceType.php
```diff
@@ -107,4 +100,25 @@ public function getBlockPrefix()
...
+     */
+    private function getTaxons($root = null, $filter = null)
+    {
+        if (null !== $root) {
+            $taxons = $this->taxonRepository->findChildren($root, true);
+        } else {
```
Zales0123: I know that another extraction would not be a good option (two methods for a simple case), but `else` does not look good too :/ How could we make it shiny and perfect?

src/Sylius/Component/Taxonomy/Repository/TaxonRepositoryInterface.php
```diff
@@ -25,10 +25,11 @@
 {
     /**
      * @param string $parentCode
+     * @param bool   $isParentRoot
```
Zales0123: Again, do not indent parameters vertically.

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
Pitoune: That's what I did in the first place but @lchrusciel said the method `findChildrenForRootParent` was redundant with the `findChildren` one.

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
```diff
@@ -23,20 +22,30 @@ class TaxonRepository extends EntityRepository implements TaxonRepositoryInterfa
     /**
      * {@inheritdoc}
      */
-    public function findChildren($parentCode)
+    public function findChildren($parentCode, $isParentRoot = false)
```
pamil: Why do we need this parameter?

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
lchrusciel: Sorry, I wasn't clear enough... I meant that the variable assignment is redundant... Sorry about that :)

```php
                ->addSelect('child')
                ->leftJoin('o.children', 'child')
                ->leftJoin('o.parent', 'parent')
                ->andWhere('parent.code = :code')
;
```

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
pamil: `findChildren('CODE', true)` shouldn't return grandchildren etc.
It should be rather `findChildren($parentCode)` and `findDescendants($rootCode)`.

src/Sylius/Bundle/TaxonomyBundle/Form/Type/TaxonChoiceType.php
Pitoune: What about adding a nullable argument `$rootCode` to method `findNodesTreeSorted` ? It means putting some logic in the repository but we only need an `if` without `else`.

src/Sylius/Bundle/TaxonomyBundle/Form/Type/TaxonChoiceType.php
```diff
@@ -81,21 +80,15 @@ public function configureOptions(OptionsResolver $resolver)
...
             ->setDefaults([
                 'choices' => function (Options $options) {
-                    $taxons = $this->taxonRepository->findNodesTreeSorted();
-
-                    if (null !== $options['filter']) {
-                        $taxons = array_filter($taxons, $options['filter']);
-                    }
-
-                    return $taxons;
+                    return $this->getTaxons($options['root'], $options['filter']);
```
michalmarcinkowski: I would rename this option to `root_code` to avoid confusion.

Pitoune:
@michalmarcinkowski did you have time to have a look on my last commit?

michalmarcinkowski:
I'm not a fan of optional arguments, but creating a separate method for that doesn't seem a better choice. Thanks Pierre! 👍

PR #7255, opened 2017-01-13
Fix #7209: prevent adding unavailable variant

### LINKED ISSUE #7209: Exception when trying to add to cart a non found variant
Here are the steps to reproduce this issue. 
 - Do a fresh install of beta1
 - Go to the admin backend
 - Create a new configurable product with "Mug Type" as only option and "Options Matching" as Variant Selected Method. 
 - Create a only variant for this product. Set "Medium mug" as "Mug Type" option
 - Now go to the frontend page of the product you have created
 - You will see the possibly option values for the mug type.
 - If you try to add "medium mug" everything works as expected
 - If you try to add the "monster mug" a exception is thrown that can be seen in the symfony profiler. it reads `Expected argument of type "lius\Component\Core\Model\OrderItem::setVariant() must implement interface Sylius\Component\Core\Model\ProductVariantInterface", "NULL" given`

Looks to me that we need to control if we have found a match before trying to add the item to the cart.

David-Guillot:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7209  |
| License         | MIT |

For now only the web UI part is done ; I'm not sure where to act on PHP part, because the exception is thrown very deeply in the Symfony Form handling code.

### DIFF
--- a/features/product/viewing_products/viewing_different_price_for_different_product_variants_selected_with_options.feature
+++ b/features/product/viewing_products/viewing_different_price_for_different_product_variants_selected_with_options.feature
@@ -27,3 +27,4 @@ Feature: Viewing different price for different product variants selected with op
         When I view product "Wyborowa Vodka"
         And I set its volume to "1L"
         Then the product price should be "Unavailable"
+        Then I should be unable to add it to the cart
--- a/src/Sylius/Behat/Context/Ui/Shop/ProductContext.php
+++ b/src/Sylius/Behat/Context/Ui/Shop/ProductContext.php
@@ -293,7 +293,7 @@ final class ProductContext implements Context
     {
         Assert::false(
             $this->showPage->hasAddToCartButton(),
-            'Add to cart button should not be visible.'
+            'Add to cart button should be absent or disabled.'
         );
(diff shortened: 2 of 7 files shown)


David-Guillot:
@michalmarcinkowski PR completed and rebased on upstream/master. FYI:
* The fix on PHP side is to throw a proper exception within the Symfony Form DataTransformer in charge of matching Product Option to Product Variant, according to [the Symfony docs](http://symfony.com/doc/current/form/data_transformers.html#creating-the-transformer)
* I tested the PHP fix on the web UI by disabling my first "cosmetic" workaround, and it works fine: a 400 error is returned, the JS handles it correctly and shows a clean "this value is not valid" message to the end-user ; of course with the "cosmetic" workaround, nobody will see this message because nobody will be able to validate the form, but even if one does, there will be no 500 error
* The API is not affected by this bug as it does not use the "Product Option Code to Product Variant" matching feature

michalmarcinkowski:
Thank you David! Great job! 👍

PR #7272, opened 2017-01-17
[Product] Add groupBy missing parameters for Postgresql/Mysql

Niiko:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | - |
| License         | MIT |

I have an issue with groupBy queries with PostgreSQL

In fact, PostgreSQL requires all columns selected in groupBy clause, or at least, when it's about join selected columns, the id of joined entity.

Maybe there are more in code, i will fix them as i spot them.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
@@ -68,7 +68,7 @@ class ProductRepository extends BaseProductRepository implements ProductReposito
             ->andWhere('channel = :channel')
             ->andWhere('o.enabled = true')
             ->andWhere('channelPricing.channel = :channel')
-            ->groupBy('o.id')
+            ->groupBy('o.id, translation.id, productTaxon.position, channelPricing.price')
             ->setParameter('locale', $locale)
             ->setParameter('taxonSlug', $taxonSlug)
             ->setParameter('channel', $channel)


lchrusciel:
Won't it resolve the issue with mysql full group by also?

Niiko:
Just test it and yes, it will fix it

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
```diff
@@ -68,7 +68,7 @@ public function createQueryBuilderByChannelAndTaxonSlug(ChannelInterface $channe
             ->andWhere('channel = :channel')
             ->andWhere('o.enabled = true')
             ->andWhere('channelPricing.channel = :channel')
-            ->groupBy('o.id')
+            ->groupBy('o.id, translation.id, productTaxon.position')
```
lchrusciel: `->groupBy('o.id, translation.id, productTaxon.position', 'channelPricing.price')`

Niiko: That sorting crash tests because values tested are not the right ones.

Are you sure ? The original groupBy has his own "sorting". Usually, sorty is managed in orderBy no ?

Also, grouping on price may produces non unique results (one per channel)

lchrusciel: I've just checked available options on current master... We also have a sorting by price.

michalmarcinkowski:
The build is failing, could you have a look?

Niiko:
I checked, there is an issue on query, orderBy is not taken in account.

I had to remove channelPricing.price in groupBy AND disable ONLY_FULL_GROUP_BY in mysql to make it works.

I think this behavior, especially for mysql configuration is too much, no?

lchrusciel:
The problem is a little bit more complex. The problem lies somewhere in between query and our scenario.

The following products are available in our store:  
```behat
        And the store has a "Wyborowa Vodka" configurable product
        And the product "Wyborowa Vodka" has "Wyborowa Vodka Exquisite" variant priced at "$40.00"
        And the product "Wyborowa Vodka" has "Wyborowa Apple" variant priced at "$12.55"
        And the store has a "Jack Daniel's" configurable product
        And the product "Jack Daniel's" has "Jack Daniel's Old No.7" variant priced at "$30.00"
        And the product "Jack Daniel's" has "Gentleman Jack" variant priced at "$50.00"
        And the store has a "Johnnie Walker" configurable product
        And the product "Johnnie Walker" has "Johnnie Walker Red Label" variant priced at "$20.00"
        And the product "Johnnie Walker" has "Johnnie Walker Black Label" variant priced at "$25.00"
```
So if we sort them by price, they will look like this:

| Product | Price  |
| ------------- | -----:|
| Wyborowa Apple | $12.55 |
| Johnnie Walker Red Label | $20 |
| Johnnie Walker Black Label | $25 |
| Jack Daniel's Old No.7 | $30 |
| Wyborowa Vodka Exquisite | $40 |
| Gentleman Jack | $50 |

If we look for the cheapest product it would a `Wyborowa Vodka`, because of the Apple variant. And the most expensive one is `Jack Daniel's`. But as we present a price of the first variant we expect something like this: 

| Product | Price  |
| ------------- | -----:|
| Johnnie Walker Red Label | $20 |
| Jack Daniel's Old No.7 | $30 |
| Wyborowa Vodka Exquisite | $40 |

And we covers this case in our feature files:
```behat
    Scenario: Sorting products by price of their first variant with ascending order
        And the first product on the list should have name "Johnnie Walker"
        And the last product on the list should have name "Wyborowa Vodka"
    Scenario: Sorting products by price of their first variant with descending order
        And the first product on the list should have name "Wyborowa Vodka"
        And the last product on the list should have name "Johnnie Walker"
```
It seems that the tests are just wrong and we should adjust them to have a proper behaviour. And change our variant resolver to choose a cheapest variant (for displaying purposes especially). But on the other hand there will be a problem with this behaviour. With presenting the cheapest price on a product list we can confuse user with descending price filter.. But it could be probably adjusted in a second iteration. 

Anyway, we need this fix ASAP, because on the newest MySQL version product list looks like this: 
sylius exception

lchrusciel:
One more thing, we need to cover a scenario with multichannel multivariant case. And sort with the proper value for each of channels.

Issue #7215, opened 2017-01-09
[RFC] Skipping the shipping checkout step (virtual products)


pjedrzejewski:
Hey folks!

While handling virtual/downloadable products is definitely out of scope for the core of Sylius (at least in v1), I think we could make our life (and life of the awesome guys that will implement virtual products bundle for Sylius) a bit easier, by introducing a very basic foundation for such feature.

I have did some research on other eCommerce platforms and most have a simple checkbox on the product, which indicates whether it is a virtual product or not.

As we can see, there is some interest in that already:

* https://github.com/Sylius/Sylius/issues/7180
* https://github.com/Sylius/Sylius/issues/7017
* https://github.com/Sylius/Sylius/issues/2150
* https://github.com/Sylius/Sylius/issues/1272

We have already figured out how to skip the payment step if payment is not required: #7031. Doing so for the shipping step would be very similar:

* We add ``protected $requiresShipping = true`` to the ProductVariant (so that we can have physical and audio book on a single product)
* We create a shipment only for products that require shipping
* We skip the shipping selection step if there are no shipments (just like for payment)
* We display an indication on the order summary that product is "virtual" or does not require shipping

That's it. BETA2, which is coming very soon will be an absolute feature freeze, so this is the last chance to squeeze this in and I believe it would be a huge improvement for the future, achieved at low cost. Otherwise people will need to make their ways around skipping the shipping step, for whatever reason.

Concerns:

* Our "default" address in the checkout address screen is "shipping", looks like we would need to make a switch, because the billing address is always required and needs to be filled, shipping can be the same or non existent. This change would make sense regardless of the changes described above.

Let me hear your thoughts about this! :)

FatalError404:
Hello pjedrzejewski,

did you check the state machine configuration : http://docs.sylius.org/en/latest/customization/state_machine.html ?

Carefull about some process, in shipping step, there are some adjustments, (shipping method is updated according to selected country/province, played thanks to sylius ajax request; AND shipping method is part of the final object which defines the final price).

I guess shipping is part of the model, and currently, it seems to be difficult to disable it (not impossible ;)).

Regards,

JMO.

starspire:
Hello pjedrzejewski,
I'd also dealt with it. In my case I created "Shipment via internet" method and after successful payment in event listener updated shipment status. Because I used API (I used 0.18,where the API was really poor, so I writen some own methods) I didn't have to fill the step with shipment address. 
[...]

PR #7276, opened 2017-01-17
[Checkout] Skipping shipping checkout step

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/7215 |
| License         | MIT |

Skipping shipping step is done similarly to [skipping payment step](https://github.com/Sylius/Sylius/pull/7031) - but based on new property ``$shippingRequired``. 

For now, there is a problem with skipping both **shipping** and **payment** step together, we should also think about how to prevent from invalid checkout operations (e.g. returning to shipping step after shipping step had been skipped).

zrzut ekranu 2017-02-09 o 10 32 34


src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -365,6 +365,25 @@ public function theProductHasVariantPricedAt(
...
 
     /**
+     * @Given /^(this product) has "([^"]+)" variant priced at ("[^"]+") which does not require shipping$/
+     */
+    public function theProductHasVariantWhichDoesNotRequireShipping(
+        ProductInterface $product,
+        $productVariantName,
+        $price
+    ) {
+        $this->createProductVariant(
```
lchrusciel: Can we use some example factory. This class is way to big...

Zales0123: There is no example factory for product variant strictly :/ Do you think it would be worth to implement this? But I think it's a case for separate PR.

src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -338,7 +338,7 @@ public function theProductHasVariantPricedAt(
         $price,
         ChannelInterface $channel = null
     ) {
-        $this->createProductVariant(
+        return $this->createProductVariant(
```
michalmarcinkowski: Why?

Zales0123: 🦅 👁  :D

src/Sylius/Bundle/ShopBundle/Resources/views/Checkout/_steps.html.twig
```diff
@@ -8,24 +8,34 @@
     {% set steps = {'address': 'completed', 'select_shipping': 'completed', 'select_payment': 'completed', 'complete': 'active'} %}
 {% endif %}
 
-{% set is_order_total_zero = (order.total == 0) %}
+{% set order_requires_payment = (order.total != 0) %}
+{% set order_requires_shipping = (order.shipments|length > 0) %}
 
-<div class="ui {{ is_order_total_zero ? 'three' : 'four' }} steps">
+{% set stepsCount = 'four' %}
```
pjedrzejewski: steps_count

Zales0123:
@pjedrzejewski that was my first shot, however there were some problems with this approach. The main was that there is no `Request` on `ResourceControllerEvent`, so we cannot use `RequestMatcher` to determine is it an event from checkout request. Because of that, we can't also allow custom redirections defined in routing (with my solution id redirection is defined it's taken as more important - crucial for **payum** functionality).
I know it affects every request, but I think that (at least for now) it's not so bad. I was thinking about giving possiblity to define custom redirect handler service in routing definition (wdyt about such a feature?) but this is definitely not something we want to do in a PR like this one.

pjedrzejewski:
@Zales0123 Any reason not to use request stack in the listener to get the currently handled request? :)

Zales0123:
@pjedrzejewski good point 😄  I'll check it out and come back with new implementation/new thoughts 🐃

pjedrzejewski:
Nice work Mateusz! 👍 

One (after merge) thought: callback should be ``sylius_resolve_checkout_state`` or something like that, wdyt?

PR #7282, opened 2017-01-18
[Security] Clean up

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/private/js/sylius-variants-prices.js
+++ b/src/Sylius/Bundle/ShopBundle/Resources/private/js/sylius-variants-prices.js
@@ -34,10 +34,10 @@ function handleProductOptionsChange() {
 
         if ($price !== undefined) {
             $('#product-price').text($price);
-            $('#sylius-product-adding-to-cart').enable();
+            $('button[type=submit]').removeAttr('disabled');
         } else {
             $('#product-price').text($('#sylius-variants-pricing').attr('data-unavailable-text'));
-            $('#sylius-product-adding-to-cart').disable();
+            $('button[type=submit]').attr('disabled', 'disabled');
         }
     });
 }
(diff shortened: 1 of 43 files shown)


pjedrzejewski:
@Arminek This is great work, but also a BC break, could you document the changes needed in the UPGRADE? (config files and maybe note about how to upgrade routing)

src/Sylius/Bundle/ShopBundle/Resources/private/js/sylius-variants-prices.js
```diff
@@ -33,11 +33,13 @@ function handleProductOptionsChange() {
...
 
         if ($price !== undefined) {
+            console.log('elo');
             $('#product-price').text($price);
-            $('#sylius-product-adding-to-cart').enable();
+            $('button[type=submit]').removeAttr('disabled');
         } else {
             $('#product-price').text($('#sylius-variants-pricing').attr('data-unavailable-text'));
-            $('#sylius-product-adding-to-cart').disable();
+            console.log('nope');
```
tuka217: Is it needed?

pjedrzejewski: :D

Arminek: 🌮

pjedrzejewski:
Great, thank you Arek! :)

smashou:
Could you, please, just explain or link the documentation which explain why  the locale switch and currency switch require now POST and not GET ? 
Just for my comprehension.

michalmarcinkowski:
@smashou the idea was to protect it from XSS, but it's an overkill so we reverted that change in https://github.com/Sylius/Sylius/pull/7337.

PR #7303, opened 2017-01-20
[Payment] Remove credit card

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| License         | MIT |


pjedrzejewski:
@Arminek You should post some explanation with such PR. :)

We are removing this because it is not really used in Sylius and there is no recommended way for using it - we may introduce such entity in the future but for now it is useless.

pjedrzejewski:
Thanks Arek!

PR #7307, opened 2017-01-23
[Admin][Security] Use csrf token in update position

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Controller/ProductTaxonController.php
+++ b/src/Sylius/Bundle/CoreBundle/Controller/ProductTaxonController.php
@@ -29,13 +29,18 @@ class ProductTaxonController extends ResourceController
 {
     /**
      * @param Request $request
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse
+     *
+     * @return Response
      */
     public function updatePositionsAction(Request $request)
     {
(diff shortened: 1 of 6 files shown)


src/Sylius/Bundle/CoreBundle/Controller/ProductTaxonController.php
```diff
@@ -29,14 +29,19 @@ class ProductTaxonController extends ResourceController
...
+     * @return Response
      */
     public function updatePositionsAction(Request $request)
     {
         $configuration = $this->requestConfigurationFactory->create($this->metadata, $request);
         $this->isGrantedOr403($configuration, ResourceActions::UPDATE);
         $productTaxons = $request->get('productTaxons');
 
+        if ($configuration->isCsrfProtectionEnabled() && !$this->isCsrfTokenValid('update-product-taxon-position', $request->request->get('_csrf_token'))) {
+            throw new HttpException(Response::HTTP_FORBIDDEN, 'Invalid csrf token.');
```
lchrusciel: We could use `throw new AccessDeniedException('Invalid csrf token.')` instead.

pjedrzejewski:
Thanks Arek!

PR #7311, opened 2017-01-23
[Admin][Taxon] Use csrf in move taxon

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |


src/Sylius/Bundle/ResourceBundle/Routing/ResourceLoader.php
```diff
@@ -150,6 +150,9 @@ private function createRoute(MetadataInterface $metadata, array $configuration,
         if ($isApi && in_array($actionName, ['show', 'create', 'update'], true)) {
             $defaults['_sylius']['serialization_groups'] = ['Default', 'Detailed'];
         }
+        if ($isApi && in_array($actionName, ['create', 'update'])) {
```
pamil: Can you join it with the `if` below?

PR #7313, opened 2017-01-23
[DOC] Added string filter option.type configuration

starspire:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

I added info about string filter configuration not mentioned in a docs. I want to write my own filter without choosing matching possibilities but I saw, that this is already implemented but anything in docs about it.


pjedrzejewski:
Thanks Krzysztof, good catch! :)

PR #7320, opened 2017-01-23
Clean up & speed up core repositories

pamil:


pjedrzejewski:
Thank you Kamil! Would be cool to get some quick comparison screenshots for this kind of PRs! WDYT? ;)

pamil:
This one is more for correctness, safety and coding standards than performance, but the next ones will have it (#7325).

PR #7325, opened 2017-01-24
Make product browsing fast again

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes (see `UPGRADE.md`) |
| Related tickets | - |
| License         | MIT |

Before:
screen shot 2017-01-24 at 10 53 21

After:
screen shot 2017-01-24 at 10 52 23

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
@@ -13,6 +13,7 @@ namespace Sylius\Bundle\CoreBundle\Doctrine\ORM;
 
 use Sylius\Bundle\ProductBundle\Doctrine\ORM\ProductRepository as BaseProductRepository;
 use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Core\Model\TaxonInterface;
 use Sylius\Component\Core\Repository\ProductRepositoryInterface;
 
 /**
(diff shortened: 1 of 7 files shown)


src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
```diff
@@ -46,24 +47,18 @@ public function createListQueryBuilder($locale, $taxonId = null)
...
      * {@inheritdoc}
      */
-    public function createQueryBuilderByChannelAndTaxonSlug(ChannelInterface $channel, $taxonSlug, $locale)
+    public function createQueryBuilderByChannelAndTaxon(ChannelInterface $channel, TaxonInterface $taxon, $locale)
     {
         return $this->createQueryBuilder('o')
             ->addSelect('translation')
             ->leftJoin('o.translations', 'translation', 'WITH', 'translation.locale = :locale')
-            ->innerJoin('o.variants', 'variant')
-            ->innerJoin('variant.channelPricings', 'channelPricing')
```
pamil: Unfortunately it's required by grid to sort by prices. Any ideas how to overcome it / join only if there's a need?

pjedrzejewski: We can pass ``$criteria`` (so criteria array from URL) to this method and see if ``sorting[price]`` was set, if not - skip the joins.

pjedrzejewski:
1. Why changing the convention? ``createQueryBuilderForGrid`` tells us less than previous method name.
2. You are testing with a single product? (that's what query count suggest me) This change generates extra query for each product in the list. With a lot of variants, this can be worse than old query perf. wise. Can you try to test both versions with our Gatling setup? This would give us more insight on a large catalogue.

Anyway, I am not sure how much we should invest into optimizing this page, for large stores, it will be handled by ElasticSearch anyway. (https://github.com/Lakion/SyliusElasticSearchBundle)

pamil:
1. Since it's got `array $sorting` as a last argument, it feels like it should be used only in that grid. `createQueryBuilderForChannelAndTaxonAndLocaleSortedBy()`? :(

2. It's the list of products in the taxon, each taxon having 3000-4000 products, with 1-125 variants each (avg 12). 

Before:
screen shot 2017-01-27 at 10 08 30

After:
screen shot 2017-01-27 at 10 08 38

lchrusciel:
`createQueryBuilderForChannelAndTaxon`? Sorting could be optional, as we need it only to decide if we need join or not

pamil:
@lchrusciel there's still locale passed, which is used mostly for grids to allow for correct sorting by name. So there are two arguments not specific for grid and two arguments specific for grid.

pjedrzejewski:
Thanks Kamil!

PR #7328, opened 2017-01-24
Doctrine caching in test_cached environment

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | - |
| License         | MIT |

Before:
screen shot 2017-01-24 at 13 57 40

After:
screen shot 2017-01-24 at 13 57 25

Result cache not even used yet! 🎉

### DIFF
--- a/app/config/config_test_cached.yml
+++ b/app/config/config_test_cached.yml
@@ -8,6 +8,14 @@ doctrine:
     orm:
         entity_managers:
             default:
+                result_cache_driver:
+                    type: memcached
+                    host: localhost
+                    port: 11211
+                query_cache_driver:
+                    type: memcached
+                    host: localhost
+                    port: 11211
                 metadata_cache_driver:
                     type: memcached
                     host: localhost


inssein:
@pamil you have any numbers against the default symfony recommendation (APC)?

```YAML
doctrine:
    orm:
        metadata_cache_driver: apc
        result_cache_driver: apc
        query_cache_driver: apc
```

pjedrzejewski:
It is ok to use it for test_cached, but for production I'd add recommended production settings to the documentation. You may want to use apc/redis/memcached, not sure if we should enforce any, only recommend. WDYT?

pamil:
@inssein I'll make a benchmark tomorrow, seems easier to setup! :)

@pjedrzejewski we might include it in Sylius' `app/config/config_prod.yml` as Sylius-Standard has its own file, so it is not enforced.

pjedrzejewski:
@pamil What's the point of it then? Contributors do not really use prod env and this can cause some confusion - why is it different from Sylius-Standard. I'd say let's remove it, agree?

pamil:
@pjedrzejewski agreed 👍

pjedrzejewski:
Great, thanks Kamil!

PR #7330, opened 2017-01-24
[API] [Documentation] Default oauth fixture

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

It will make our lives easier.

TODO:
 - [x] Document current changes 
 - [x] Adjust all current api examples to use access token from fixtures 

**Edit:**
It would be worthy to split this fixtures and create a new `api` suite. But the problem is, that we are not supporting many fixtures suites. Each fixture load will purge database. Thus I will have to copy whole default one and add the lines, I have added here... Should I duplicate this fixture suite? 

Probably, if we remove ApiBundle from kernel, the default fixture suite will be broken.

### DIFF
--- a/docs/api/authorization.rst
+++ b/docs/api/authorization.rst
@@ -181,16 +181,16 @@ Default values in dev environment
 In a developer environment there are default client id, client secret and default access token provided to allow you to test our API just out-of-the-box.
 In order to access them, please use the following values:
 
-+---------------+---------------------------------------------------------------------------------------------------+
-| Parameter     | Value                                                                                             |
-+===============+===================================================================================================+
-| client_id     | demo_client                                                                                       |
-+---------------+---------------------------------------------------------------------------------------------------+
-| client_secret | secret_demo_client                                                                                |
-+---------------+---------------------------------------------------------------------------------------------------+
-| grant_type    | password                                                                                          |
-+---------------+---------------------------------------------------------------------------------------------------+
-| access_token  | SampleTokenMWExMWM0NzE1NmUyZDgyZDJiMjEzMmFlMjQ4MzgwMmE4ZTkxYzM0YjdlN2U2YzliNDIyMTk1ZDhlNDYxYWE4Ng |
-+---------------+---------------------------------------------------------------------------------------------------+
++---------------+--------------------+
+| Parameter     | Value              |
++===============+====================+
+| client_id     | demo_client        |
++---------------+--------------------+
+| client_secret | secret_demo_client |
++---------------+--------------------+
+| grant_type    | password           |
++---------------+--------------------+
+| access_token  | SampleToken        |
++---------------+--------------------+
 
 These values will be used later on to make it easier for you to check, how our API works.
--- a/docs/api/authorization.rst
+++ b/docs/api/authorization.rst
@@ -174,3 +174,23 @@ You can now use new token to send requests
         "scope": null,
         "refresh_token": "MWI4NzVkNThjZDc2Y2M1N2JiNzBmOTQ0MDFmY2U0YzVjYzllMDE1OTU5OWFiMzJiZTY5NGU4NzYyODU1N2ZjYQ"
     }
+
+Default values in dev environment
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In a developer environment there are default client id, client secret and default access token provided to allow you to test our API just out-of-the-box.
+In order to access them, please use the following values:
+
++---------------+---------------------------------------------------------------------------------------------------+
+| Parameter     | Value                                                                                             |
++===============+===================================================================================================+
+| client_id     | demo_client                                                                                       |
++---------------+---------------------------------------------------------------------------------------------------+
+| client_secret | secret_demo_client                                                                                |
++---------------+---------------------------------------------------------------------------------------------------+
+| grant_type    | password                                                                                          |
++---------------+---------------------------------------------------------------------------------------------------+
+| access_token  | SampleTokenMWExMWM0NzE1NmUyZDgyZDJiMjEzMmFlMjQ4MzgwMmE4ZTkxYzM0YjdlN2U2YzliNDIyMTk1ZDhlNDYxYWE4Ng |
++---------------+---------------------------------------------------------------------------------------------------+
+
+These values will be used later on to make it easier for you to check, how our API works.
--- a/src/Sylius/Bundle/CoreBundle/Fixture/ApiClientFixture.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/ApiClientFixture.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\CoreBundle\Fixture;
+
+use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
+
+/**
+ * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
+ */
+class ApiClientFixture extends AbstractResourceFixture
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return 'api_client';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configureResourceNode(ArrayNodeDefinition $resourceNode)
+    {
+        $resourceNode
+            ->children()
+                ->scalarNode('random_id')->cannotBeEmpty()->end()
+                ->scalarNode('secret')->cannotBeEmpty()->end()
+                ->variableNode('allowed_grant_types')->cannotBeEmpty()->defaultValue([])->end()
+        ;
+    }
+}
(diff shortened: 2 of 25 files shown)


docs/api/authorization.rst
```diff
@@ -174,3 +174,23 @@ You can now use new token to send requests
         "scope": null,
         "refresh_token": "MWI4NzVkNThjZDc2Y2M1N2JiNzBmOTQ0MDFmY2U0YzVjYzllMDE1OTU5OWFiMzJiZTY5NGU4NzYyODU1N2ZjYQ"
     }
+
+Default values in dev environment
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In a developer environment there are default client id, client secret and default access token provided to allow you test our API just out-of-the-box.
```
TheMadeleine: to allow you **to**

docs/api/authorization.rst
```diff
@@ -174,3 +174,23 @@ You can now use new token to send requests
...
+In order to access them, please use following values:
```
TheMadeleine: **the** following values:

docs/api/authorization.rst
```diff
@@ -174,3 +174,23 @@ You can now use new token to send requests
...
+| client_id     | demo_client                                                                                       |
++---------------+---------------------------------------------------------------------------------------------------+
+| client_secret | secret_demo_client                                                                                |
++---------------+---------------------------------------------------------------------------------------------------+
+| grant_type    | password                                                                                          |
++---------------+---------------------------------------------------------------------------------------------------+
+| access_token  | SampleTokenMWExMWM0NzE1NmUyZDgyZDJiMjEzMmFlMjQ4MzgwMmE4ZTkxYzM0YjdlN2U2YzliNDIyMTk1ZDhlNDYxYWE4Ng |
++---------------+---------------------------------------------------------------------------------------------------+
+
+This values will be used in rest of API chapters to make it easier for you to check, how our API works.
```
TheMadeleine: **These** values will be used **later on** to make it easier...

pjedrzejewski:
1. I am not sure we should provide default access token. In real life, you need to obtain it and I think it is not that difficult but also illustrates the proper workflow in the docs. Sample client is fine.
2. Is it possible that the suite configurations will simply merge when we add the client fixtures in the API bundle? Or there is performNoDeepMerging on suites? /cc @pamil

lchrusciel:
@pjedrzejewski with this fixtures we will be able to provide a copy&paste access to our demo from terminal. This token will never expire. So if anyone will be willing to check how our api works one can just try a curl without any previous action.

I also have a doubts about security, but on the other hand fixtures should be used only on dev environment and we are providing a default users with credentials anyway. It is just +10 for DX.

TheMadeleine:
What about these API examples? ;> http://docs.sylius.org/en/latest/cookbook/api.html

src/Sylius/Bundle/CoreBundle/Fixture/ApiClientFixture.php
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function configureResourceNode(ArrayNodeDefinition $resourceNode)
+    {
+        $resourceNode
+            ->children()
+                ->scalarNode('random_id')->cannotBeEmpty()->end()
+                ->scalarNode('secret')->cannotBeEmpty()->end()
+                ->variableNode('allowed_grant_types')->cannotBeEmpty()->defaultValue([])->end()
```
pamil: Is it an array? `->arrayNode('allowed_grant_types')->prototype('scalar')`?

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ApiAccessTokenExampleFactory.php
```diff
@@ -0,0 +1,115 @@
...
+        $options = $this->optionsResolver->resolve($options);
+
+        /** @var AccessTokenInterface $accessToken */
+        $accessToken = $this->accessTokenFactory->createNew();
+
+        $accessToken->setClient($options['client']);
+        $accessToken->setToken($options['token']);
+        $accessToken->setUser($options['user']);
+
+        if (isset($options['expires_at'])) {
```
pamil: There's no `expires_at` defined in `configureOptions` so it'll be always false.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ApiClientExampleFactory.php
```diff
@@ -0,0 +1,89 @@
...
+     * {@inheritdoc}
+     */
+    public function create(array $options = [])
+    {
+        $options = $this->optionsResolver->resolve($options);
+
+        /** @var ClientInterface $client */
+        $client = $this->clientManager->createClient();
+
+        if (isset($options['random_id'])) {
```
pamil: It's always set (`configureOptions` has `setDefault` for this key).

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ApiClientExampleFactory.php
```diff
@@ -0,0 +1,89 @@
...
+            $client->setRandomId($options['random_id']);
+        }
+        if (isset($options['secret'])) {
```
pamil: It's always set (`configureOptions` has `setDefault` for this key).

src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
```diff
@@ -210,3 +210,19 @@ sylius_fixtures:
...
+                                secret: "secret_demo_client"
+                                allowed_grant_types: ['password']
+
+                access_token:
+                    options:
+                        custom:
+                            default_access_token:
+                                client: "demo_client"
+                                user: "api@example.com"
+                                token: "SampleTokenMWExMWM0NzE1NmUyZDgyZDJiMjEzMmFlMjQ4MzgwMmE4ZTkxYzM0YjdlN2U2YzliNDIyMTk1ZDhlNDYxYWE4Ng"
```
pamil: If we are simplifying things, can we use just "SampleToken" as token instead?

src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml
```diff
@@ -210,3 +210,19 @@ sylius_fixtures:
...
                     options:
                         random: 10
+
+                api_client:
+                    options:
+                        custom:
+                            default_api_client:
+                                random_id: "demo_client"
+                                secret: "secret_demo_client"
+                                allowed_grant_types: ['password']
```
pamil: We're usually using `"` instead of `'`, it's the only exception in this file.

src/Sylius/Bundle/CoreBundle/Tests/Fixture/AccessTokenFixtureTest.php
```diff
@@ -0,0 +1,85 @@
...
+        $this->assertConfigurationIsValid([['random' => 4]], 'random');
+    }
+
+    /**
+     * @test
+     */
+    public function access_token_can_be_created_with_custom_random_id()
+    {
+        $this->assertConfigurationIsValid([['custom' => [[
+                'client' => 'some_client',
```
pamil: Too deep alignment.

src/Sylius/Bundle/CoreBundle/Tests/Fixture/AccessTokenFixtureTest.php
```diff
@@ -0,0 +1,85 @@
...
+        ]]]], 'custom.*.client');
+    }
+
+    /**
+     * @test
+     */
+    public function access_token_can_be_created_with_custom_secret()
+    {
+        $this->assertConfigurationIsValid([['custom' => [[
+                'user' => 'api@example.com'
```
pamil: Too deep alignment.

src/Sylius/Bundle/CoreBundle/Tests/Fixture/ApiCredentialsFixtureTest.php
```diff
@@ -0,0 +1,76 @@
...
+        $this->assertConfigurationIsValid([['random' => 4]], 'random');
+    }
+
+    /**
+     * @test
+     */
+    public function oauth_credentials_can_be_created_with_custom_random_id()
+    {
+        $this->assertConfigurationIsValid([['custom' => [[
+                'random_id' => 'totally_random',
```
pamil: Too deep alignment.

src/Sylius/Bundle/CoreBundle/Tests/Fixture/ApiCredentialsFixtureTest.php
```diff
@@ -0,0 +1,76 @@
...
+        ]]]], 'custom.*.random_id');
+    }
+
+    /**
+     * @test
+     */
+    public function oauth_credentials_can_be_created_with_custom_secret()
+    {
+        $this->assertConfigurationIsValid([['custom' => [[
+                'secret' => 'threeCanKeepSecretIfTwoAreDead'
```
pamil: Too deep alignment.

pamil:
@pjedrzejewski it's supported behaviour (see https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/FixturesBundle/Tests/DependencyInjection/ConfigurationTest.php#L85-L169)

lchrusciel:
@pamil was right. Right now we have everything in ApiBundle 👍

lchrusciel:
@TheMadeleine updated!

pjedrzejewski:
@lchrusciel Good point, I agree! :) Thanks Łukasz!

PR #7338, opened 2017-01-25
[Documentation]  SyliusResourceBundle » Forms precision

PyRowMan:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

The service definition of the form wasn't given, it could be a good idea to describe it.

If you don't specify the service the forms construct is not filled and end up with an error.

### DIFF
--- a/docs/bundles/SyliusResourceBundle/forms.rst
+++ b/docs/bundles/SyliusResourceBundle/forms.rst
@@ -41,7 +41,7 @@ Create the form's service :
         class: AppBundle\Form\Type\BookType
         tags:
             - { name: form.type }
-        arguments: ['%sylius.model.order.class%']
+        arguments: ['%app.model.book.class%']
 
 Now, configure it under ``sylius_resource``:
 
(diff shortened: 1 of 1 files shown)


docs/bundles/SyliusResourceBundle/forms.rst
```diff
@@ -33,6 +33,16 @@ You need to create a simple class:
...
 
+Create the form's service :
+
+.. code-block:: yaml
+
+    app.book.form.type:
+        class: AppBundle\Form\Type\BookType
+        tags:
+            - { name: form.type }
+        arguments: ['%sylius.model.order.class%']
```
lchrusciel: I'm pretty sure you should use `app.model.book.class`

PyRowMan: That's a good point, let me verify this in my environnement

PyRowMan: @lchrusciel Yes, you were right, i think that it would have crash when you had submitted the form.


I have updated the PR

michalmarcinkowski:
Thanks @PyRowMan! 👍

PyRowMan:
You are very welcomed :)

PR #7344, opened 2017-01-25
Remove "Test" namespace from Core & CoreBundle

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | - |
| License         | MIT |

`Sylius\Bundle\CoreBundle\Test` and `Sylius\Component\Core\Test` were a namespaces for helper services used in tests, so they should not be a part of Core / CoreBundle in stable release as they aren't really needed there.


pjedrzejewski:
I believe we had this discussion some time ago. These service are not coupled to Behat and can be useful for any kind of testing with any tool. What is the problem with them in core?

pamil:
@pjedrzejewski ideally we would have `Sylius\Tests\Behat`and `Sylius\Tests\Whatever` namespaces (which in fact we can achieve quite easily).

I just don't want them in `Test` namespace in bundles / components as they would follow strict BC policy after Sylius goes stable and they aren't necessary there (more code to maintain). It also generates some kind of "trash bin" (e.g. `RandomInvoiceNumberGenerator` was there instead of `Sylius\Component\Core\Payment` with the other generators).

I haven't put them in `tests` directory (`Sylius\Tests`) namespace as every Behat dependency needs to be autoloaded by Composer in order to reuse our context in extension like Lakion/SyliusCmsBundle etc.

pamil:
I've opened #7355 for these classes which were not moved to `Sylius\Behat`.

PR #7349, opened 2017-01-25
Upgrade fos rest bundle

bendavies:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | no |
| BC breaks?      | maybe? |
| Related tickets |  |
| License         | MIT |

Some dependencies are falling behind a little.
This upgrades fos rest bundle to 2.1.


jetsuit:
Most of people use FOS rest bundle and I meet this problem today in my symfony 3 app when tried to set up this bundle. I think this dependency should be updated to use (FOSRestBundle 1.x is no longer maintained, 1.8 only receives security fixes. Please upgrade to FOSRestBundle 2.x as soon as possible.)

pamil:
This bundle is causing some major memory leak - our Behat (no JS) test suite usually takes ~50MB of memory, after applying this PR it takes 4GB and fails.

bendavies:
Yep, I noticed...That sucks. I'll take a look to see if I can spot anything obvious.

pamil:
Most of it should be handled by Garbage Collector & our own cleaner in [TestAppKernel](https://github.com/Sylius/Sylius/blob/master/app/TestAppKernel.php#L47-L75) :<

app/config/config_test.yml
```diff
@@ -51,5 +51,6 @@ sylius_theme:
     sources:
         test: ~
 
-twig:
-    exception_controller: "Sylius\\Bundle\\CoreBundle\\Controller\\VerboseExceptionController::showAction"
+fos_rest:
+    exception:
+        debug: true
```
pjedrzejewski: If I am correct, debug is true for env ``test`` and default fos_rest config takes it from the kernel's debug setting. We should add this to ``config_test_cached``, which has disabled the debug mode, thus creating the need for VerboseExceptionController in the past. WDYT?

bendavies: Done

pjedrzejewski:
Thank you Ben, nice work! :)

PR #7354, opened 2017-01-26
[Resource] Fix internal server error when apply sorting or filtering

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Doctrine/ORM/EntityRepository.php
+++ b/src/Sylius/Bundle/ResourceBundle/Doctrine/ORM/EntityRepository.php
@@ -86,6 +86,10 @@ class EntityRepository extends BaseEntityRepository implements RepositoryInterfa
     {
         foreach ($criteria as $property => $value) {
             $name = $this->getPropertyName($property);
+
+            if (!in_array($name, $this->_class->getFieldNames())) {
+                return;
+            }
             if (null === $value) {
                 $queryBuilder->andWhere($queryBuilder->expr()->isNull($name));
             } elseif (is_array($value)) {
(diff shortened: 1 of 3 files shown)


pjedrzejewski:
Shouldn't we check this inside of the repository? To avoid this error also when someone uses it via php code, not yml (code, hehe)?

src/Sylius/Bundle/ResourceBundle/Controller/ResourcesResolver.php
```diff
@@ -46,4 +60,42 @@ public function getResources(RequestConfiguration $requestConfiguration, Reposit
...
         return $repository->findBy($criteria, $sorting, $requestConfiguration->getLimit());
     }
+
+    /**
+     * @param array $sorting
+     * @param RepositoryInterface $repository
+     *
+     * @return array
+     */
+    private function unsetUnexistSortingFields(array $sorting, RepositoryInterface $repository)
```
lchrusciel: The only difference between these methods is first argument name... We can merge them

pamil:
This will crash sorting / filtering on properties of joined objects (sort by price on products list on taxon page).

bendavies:
This is wrong IMO. This will break where users have implemented custom criteria filtering inside their repository inside `applyCriteria`. A use might have a custom criteria name that is not a field name on the entity. So, i agree with @pjedrzejewski, this should be done inside the repository.

Arminek:
This yaml coding style makes me sick, of course you are right 💃 .

Arminek:
This solution wont work with joined properties.

pjedrzejewski:
Nice work Arek! Regarding joined fields, ``indexAction`` is generic and should be used wisely. For proper sorting and filtering we recommend using grids.

PR #7357, opened 2017-01-26
[OrderSequence] Resolve race conditions on an order number generation

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
+++ b/src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
@@ -57,7 +57,7 @@ class OrderController extends BaseOrderController
 
                 $this->manager->flush();
             } catch (OptimisticLockException $exception) {
-                $this->addFlash('error', 'sylius.checkout.complete_exception');
+                $this->addFlash('error', 'sylius.checkout.complete_error');
 
                 return $this->redirectHandler->redirectToRoute($configuration, 'sylius_shop_homepage');
             }
--- a/src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
+++ b/src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
@@ -59,7 +59,7 @@ class OrderController extends BaseOrderController
             } catch (OptimisticLockException $exception) {
                 $this->addFlash('error', 'sylius.checkout.complete_error');
 
-                return $this->redirectHandler->redirectToRoute($configuration, 'sylius_shop_homepage');
+                return $this->redirectHandler->redirectToRoute($configuration, 'sylius_shop_checkout_complete');
             }
 
             $this->eventDispatcher->dispatchPostEvent(ResourceActions::UPDATE, $configuration, $order);
--- a/src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
+++ b/src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
@@ -14,6 +14,7 @@ namespace Sylius\Bundle\CoreBundle\Controller;
 use Doctrine\ORM\OptimisticLockException;
 use FOS\RestBundle\View\View;
 use Sylius\Bundle\OrderBundle\Controller\OrderController as BaseOrderController;
+use Sylius\Bundle\ResourceBundle\Event\ResourceControllerEvent;
 use Sylius\Component\Resource\ResourceActions;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -39,6 +40,7 @@ class OrderController extends BaseOrderController
         if (in_array($request->getMethod(), ['POST', 'PUT', 'PATCH'], true) && $form->handleRequest($request)->isValid()) {
             $order = $form->getData();
 
+            /** @var ResourceControllerEvent $event */
             $event = $this->eventDispatcher->dispatchPreEvent(ResourceActions::UPDATE, $configuration, $order);
 
             if ($event->isStopped() && !$configuration->isHtmlRequest()) {
@@ -46,6 +48,10 @@ class OrderController extends BaseOrderController
             }
             if ($event->isStopped()) {
                 $this->flashHelper->addFlashFromEvent($configuration, $event);
+
+                if ($event->hasResponse()) {
+                    return $event->getResponse();
+                }
 
                 return $this->redirectHandler->redirectToResource($configuration, $order);
             }
--- a/src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
+++ b/src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
@@ -11,14 +11,85 @@
 
 namespace Sylius\Bundle\CoreBundle\Controller;
 
+use Doctrine\ORM\OptimisticLockException;
 use FOS\RestBundle\View\View;
 use Sylius\Bundle\OrderBundle\Controller\OrderController as BaseOrderController;
+use Sylius\Component\Resource\ResourceActions;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\HttpException;
 use Webmozart\Assert\Assert;
 
 class OrderController extends BaseOrderController
 {
+    /**
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function completeAction(Request $request)
+    {
+        $configuration = $this->requestConfigurationFactory->create($this->metadata, $request);
+
+        $this->isGrantedOr403($configuration, ResourceActions::UPDATE);
+        $order = $this->findOr404($configuration);
+
+        $form = $this->resourceFormFactory->create($configuration, $order);
+
+        if (in_array($request->getMethod(), ['POST', 'PUT', 'PATCH'], true) && $form->handleRequest($request)->isValid()) {
+            $order = $form->getData();
+
+            $event = $this->eventDispatcher->dispatchPreEvent(ResourceActions::UPDATE, $configuration, $order);
+
+            if ($event->isStopped() && !$configuration->isHtmlRequest()) {
+                throw new HttpException($event->getErrorCode(), $event->getMessage());
+            }
+            if ($event->isStopped()) {
+                $this->flashHelper->addFlashFromEvent($configuration, $event);
+
+                return $this->redirectHandler->redirectToResource($configuration, $order);
+            }
+
+            try {
+                if ($configuration->hasStateMachine()) {
+                    $this->stateMachine->apply($configuration, $order);
+                }
+
+                $this->manager->flush();
+            } catch (OptimisticLockException $exception) {
+                $this->addFlash('error', 'sylius.checkout.complete_exception');
+
+                return $this->redirectHandler->redirectToRoute($configuration, 'sylius_shop_homepage');
+            }
+
+            $this->eventDispatcher->dispatchPostEvent(ResourceActions::UPDATE, $configuration, $order);
+
+            if (!$configuration->isHtmlRequest()) {
+                return $this->viewHandler->handle($configuration, View::create(null, Response::HTTP_NO_CONTENT));
+            }
+
+            $this->flashHelper->addSuccessFlash($configuration, ResourceActions::UPDATE, $order);
+
+            return $this->redirectHandler->redirectToResource($configuration, $order);
+        }
+
+        if (!$configuration->isHtmlRequest()) {
+            return $this->viewHandler->handle($configuration, View::create($form, Response::HTTP_BAD_REQUEST));
+        }
+
+        $view = View::create()
+            ->setData([
+                'configuration' => $configuration,
+                'metadata' => $this->metadata,
+                'order' => $order,
+                'form' => $form->createView(),
+            ])
+            ->setTemplate($configuration->getTemplate(ResourceActions::UPDATE . '.html'))
+        ;
+
+        return $this->viewHandler->handle($configuration, $view);
+    }
+
     /**
      * @param Request $request
      *
--- a/src/Sylius/Bundle/CoreBundle/Order/NumberGenerator/SequentialOrderNumberGenerator.php
+++ b/src/Sylius/Bundle/CoreBundle/Order/NumberGenerator/SequentialOrderNumberGenerator.php
@@ -1,0 +1,117 @@
+<?php
+
+/*
+ * This file is a part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\CoreBundle\Order\NumberGenerator;
+
+use Doctrine\DBAL\LockMode;
+use Doctrine\ORM\EntityManagerInterface;
+use Sylius\Bundle\OrderBundle\NumberGenerator\OrderNumberGeneratorInterface;
+use Sylius\Component\Order\Model\OrderInterface;
+use Sylius\Component\Order\Model\OrderSequenceInterface;
+use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+/**
+ * @author Grzegorz Sadowski <grzegorz.sadowski@lakion.com>
+ */
+final class SequentialOrderNumberGenerator implements OrderNumberGeneratorInterface
+{
+    /**
+     * @var RepositoryInterface
+     */
+    private $sequenceRepository;
+
+    /**
+     * @var FactoryInterface
+     */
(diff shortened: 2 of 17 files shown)


src/Sylius/Bundle/OrderBundle/Controller/OrderController.php
```diff
@@ -183,6 +180,73 @@ public function clearAction(Request $request)
...
+                $this->flashHelper->addFlashFromEvent($configuration, $event);
+
+                return $this->redirectHandler->redirectToResource($configuration, $resource);
+            }
+
+            if ($configuration->hasStateMachine()) {
+                $this->stateMachine->apply($configuration, $resource);
+            }
+
+            if (!$this->manager->isOpen()) {
```
pjedrzejewski: Have you considered incorporating this behavior to updateAction itself?

src/Sylius/Bundle/OrderBundle/NumberGenerator/SequentialOrderNumberGenerator.php
```diff
@@ -92,11 +123,23 @@ private function getSequence()
...
+        $sequence = $this->sequenceFactory->createNew();
+        $this->sequenceManager->persist($sequence);
+
         return $sequence;
     }
+
+    private function openSequenceManagerIfClosed()
+    {
+        if (!$this->sequenceManager->isOpen()) {
+            $this->sequenceManager = EntityManager::create(
```
pjedrzejewski: This smells a bit... we are creating new entity manager here, not really opening the existing one.

src/Sylius/Component/Order/Model/OrderSequence.php
```diff
@@ -24,6 +24,11 @@ class OrderSequence implements OrderSequenceInterface
     /**
      * @var int
      */
+    protected $version = 1;
```
pjedrzejewski: Trait? Just like we do for other types of "behaviours".

src/Sylius/Bundle/OrderBundle/Controller/OrderController.php
GSadee: Of course, but I wasn't sure that we want to have this in resource, however.. in normal situation, when doctrine doesn't throw an exception, the entity manager is open, so nothing would be changed.

src/Sylius/Bundle/OrderBundle/NumberGenerator/SequentialOrderNumberGenerator.php
GSadee: Yup, but doctrine doesn't give us possibility to reopen entity manager, we can only creating a new one, which is a solution recommended by doctrine documentation. Even `resetManager()` method on doctrine `Registry` creates a new manager. 
I will change the method name.

src/Sylius/Component/Order/Model/OrderSequence.php
GSadee: I wanted to avoid the trait 😄  We don't have traits for e.g. `SlugAwareInterface` or `CodeAwareInterface` so I thought that we did't want to create another traits. What do you think @pjedrzejewski ?

src/Sylius/Bundle/OrderBundle/Controller/OrderController.php
GSadee: Another reason why I put this here is that it combines `ResourceBundle` with `ORM`, so I am not sure that I could do that in `ResourceBundle`

src/Sylius/Bundle/OrderBundle/NumberGenerator/SequentialOrderNumberGenerator.php
```diff
@@ -64,11 +76,30 @@ public function __construct(
...
-        $sequence = $this->getSequence();
-        
-        $number = $this->generateNumber($sequence->getIndex());
-        $sequence->incrementIndex();
-        
+        $number = null;
+        $counter = 0;
+        $done = false;
+        while (!$done) {
+            if (++$counter > 100) {
```
Arminek: Why do you need this `$counter` i don't get it. I suppose that is some kind of mechanism to break the loop but still i don't get it. It looks like `sleep`. Maybe something like this will be more readable:
```php
$counter = 0;
$done = false;
do {
stuff....

} while (!$done || ++$counter < 100);
```

src/Sylius/Bundle/OrderBundle/Controller/OrderController.php
```diff
@@ -183,6 +180,73 @@ public function clearAction(Request $request)
...
+                throw new HttpException($event->getErrorCode(), $event->getMessage());
+            }
+            if ($event->isStopped()) {
+                $this->flashHelper->addFlashFromEvent($configuration, $event);
+
+                return $this->redirectHandler->redirectToResource($configuration, $resource);
+            }
+
+            if ($configuration->hasStateMachine()) {
+                $this->stateMachine->apply($configuration, $resource);
```
michalmarcinkowski: @Arminek is right, we should remove whole unused logic.

src/Sylius/Bundle/OrderBundle/NumberGenerator/SequentialOrderNumberGenerator.php
```diff
@@ -64,11 +76,30 @@ public function __construct(
...
     {
-        $sequence = $this->getSequence();
-        
-        $number = $this->generateNumber($sequence->getIndex());
-        $sequence->incrementIndex();
-        
+        $number = null;
+        $counter = 0;
+        $done = false;
+        while (!$done) {
```
lchrusciel: We should extract it to separate method. And than:
```php
        for ($i = 0; $i < 100; $i++) {
//            Your stuff 
//            If succeed, early return
        }

        return 0;
```
No extra variable needed. (despite `$i`)

src/Sylius/Bundle/OrderBundle/NumberGenerator/SequentialOrderNumberGenerator.php
```diff
@@ -64,11 +76,30 @@ public function __construct(
...
+                $sequence->incrementIndex();
+
+                $this->sequenceManager->flush();
+                $done = true;
+            } catch (OptimisticLockException $exception) {
+                $this->resetSequenceManagerIfClosed();
+            }
+        }
+
         return $number;
```
lchrusciel: Should we really return a null value if number could not be generated? Maybe an appropriate will be a wise decision?

src/Sylius/Bundle/OrderBundle/NumberGenerator/SequentialOrderNumberGenerator.php
lchrusciel: Maybe we should pass 100 as a argument to constructor? Or add a constant for that?

src/Sylius/Bundle/OrderBundle/NumberGenerator/SequentialOrderNumberGenerator.php
michalmarcinkowski: We should not allow to place order without number, we probably need to throw an exception here.

makasim:
[SequentialOrderNumberGenerator.php](https://github.com/GSadee/Sylius/blob/c870a8e07585a892527efa3569f6ff0c314dbe43/src/Sylius/Bundle/OrderBundle/NumberGenerator/SequentialOrderNumberGenerator.php) looks a bit over complicated. Extra manager, reseting, locking  and so on. 

Consider doing something like this https://gist.github.com/makasim/846f1820d3b1154c71a58eddb8f3402b

Take the idea and adopt for your needs

src/Sylius/Bundle/OrderBundle/Controller/OrderController.php
```diff
@@ -183,6 +180,73 @@ public function clearAction(Request $request)
     }
 
     /**
+     * @param Request $request
+     *
+     * @return Response
+     */
+    public function completeAction(Request $request)
```
GSadee: IMO.. not, because this method is for an order entity and I do not think that there is a need to create another controller only for one method. What is more, we have in this controller also methods for an order in state cart, so if we want to extract methods, we could also create `CartController`, but I guess that we do not want that.

src/Sylius/Bundle/OrderBundle/Controller/OrderController.php
GSadee: Of course, I could do that, but if we leave that as it is now, we could define e.q. state machine or redirect in the same way as in another checkout steps. I am afraid that customizing this method and changing routing could become unclear.

src/Sylius/Bundle/OrderBundle/Controller/OrderController.php
```diff
@@ -183,6 +180,80 @@ public function clearAction(Request $request)
...
+                    throw new HttpException($event->getErrorCode(), $event->getMessage());
+                }
+                if ($event->isStopped()) {
+                    $this->flashHelper->addFlashFromEvent($configuration, $event);
+
+                    return $this->redirectHandler->redirectToResource($configuration, $resource);
+                }
+
+                if ($configuration->hasStateMachine()) {
+                    $this->stateMachine->apply($configuration, $resource);
```
lchrusciel: Didn't we want to hardcode the transition here?

GSadee: Ou, my comment was hidden after last changes. 
So, of course, I could do that, but if we leave that as it is now, we could define e.q. state machine or redirect in the same way as in another checkout steps. I am afraid that customizing this method (hardcoding transition and route) and changing routing could become unclear.

src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
```diff
@@ -24,6 +28,76 @@ class OrderController extends BaseOrderController
...
+                $this->manager->flush();
+            } catch (OptimisticLockException $exception) {
+                $this->addFlash('error', 'sylius.checkout.complete_error');
+
+                return $this->redirectHandler->redirectToRoute($configuration, 'sylius_shop_homepage');
+            }
+
+            $this->eventDispatcher->dispatchPostEvent(ResourceActions::UPDATE, $configuration, $order);
+
+            $this->getOrderEmailManager()->sendConfirmationEmail($order);
```
michalmarcinkowski: I would remove it and relay on state machine call as it was previously. We would only need to clear the email queue when `OptimisticLockException` thrown.

src/Sylius/Bundle/CoreBundle/Order/NumberGenerator/SequentialOrderNumberGenerator.php
```diff
@@ -0,0 +1,117 @@
...
+use Sylius\Bundle\OrderBundle\NumberGenerator\OrderNumberGeneratorInterface;
+use Sylius\Component\Order\Model\OrderInterface;
+use Sylius\Component\Order\Model\OrderSequenceInterface;
+use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+/**
+ * @author Grzegorz Sadowski <grzegorz.sadowski@lakion.com>
+ */
+final class SequentialOrderNumberGenerator implements OrderNumberGeneratorInterface
```
michalmarcinkowski: Why creating new class in `Core`? I think we should modify the class in `OrderBundle`

src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
```diff
@@ -24,6 +28,76 @@ class OrderController extends BaseOrderController
...
+            try {
+                if ($configuration->hasStateMachine()) {
+                    $this->stateMachine->apply($configuration, $order);
+                }
+
+                $this->manager->flush();
+            } catch (OptimisticLockException $exception) {
+                $this->addFlash('error', 'sylius.checkout.complete_error');
+
+                return $this->redirectHandler->redirectToRoute($configuration, 'sylius_shop_homepage');
```
michalmarcinkowski: Redirect to `checkout_summary` route instead of homepage.

src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
GSadee: I was looking for this and also checked when emails are sent, but there are sent immediately after call `send` method, so I couldn't clear the email queue. 
In our test environment we are using this approach: http://symfony.com/doc/current/email/spool.html
but in prod env it doesn't make sense.

src/Sylius/Bundle/CoreBundle/Order/NumberGenerator/SequentialOrderNumberGenerator.php
GSadee: Because `OrderSequence` has a variable `version` in Core, so I shouldn't locking this entity in OrderBundle

GSadee: The whole logic of optimistic locking is in `Core`

pjedrzejewski: It should be in OrderBundle and Order component, why not? Everyone using the bundle will have the issue, not only core users.

GSadee: Yesterday I asked about that @michalmarcinkowski and I moved this logic to `Core`
Now, I'm a bit confused... 😞

pjedrzejewski: @GSadee @michalmarcinkowski What was the reason?

GSadee: Coupling with ORM? I guess that better in `Core` that in bundles. 
And fact, that in #7425 I could handle versioned `ProductVariant` only in `Core`.

pjedrzejewski: 👍

src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
michalmarcinkowski: I know it is tempting to do it here, but then someone will need to override the controller action if he just want to disable it...

The emails should not be sent before flush, so I'm wondering if using state machine callbacks to send them is a good idea... I know it's convenient but perhaps we should use listeners on `post_action` for that and apply it for all emails, @pjedrzejewski?

pjedrzejewski: Moving them to listeners brings us the problem that they won't be sent out in programmatic checkout and you will have to remember to hook in the listener for new events. Using Spool in production **does make sense**, because you return the response to the user without the email transport communication delay. On the other hand, we will need to clear the spool if something goes wrong, which is extra logic... I'd say let's give it a try though, because that would be best solution.

michalmarcinkowski: With emails on state machine callbacks I see the following issues:
1. We can't be sure some will not configure the environment to not use Spool and it shouldn't be a requirement for using Sylius
2. State machine calls are made before flush, so every single operation that uses state machine will have to have additional logic for handling errors. There are many exceptions that can occur on `flush` (not only locking), then we check for them and clear the spool in every place that uses state machine...
3. Notifying external system/user should be done only after the state of the application actually changed, so it's natural that it should happen always after the changes are saved in the system, so after the `flush`

I would say we should move it to listeners.

michalmarcinkowski:
@GSadee rebase required.

GSadee:
Done 👍

michalmarcinkowski:
Good job Grzesiu! 👍

pjedrzejewski:
1. I think that overriding entire controller action to add this try catch block is bad idea and I raised my concerns before. Couldn't we encapsulate this logic to (yes, yet another) collaborator in resource controller? Especially that in #7425 we override more actions, which is unacceptable. Eventually, any resource may need db locking.

2. This change affects architecture and introduces a new concept, I'd like to have the final review on such changes before they are merged @michalmarcinkowski. Especially we have been discussing this issue and last update was that you guys will experiment with the solution a bit more.

michalmarcinkowski:
@pjedrzejewski this change just overrides a single controller action and doesn't affect architecture. Almost nothing changed after your last review and during our last conversation I had a feeling that we agreed on not introducing yet another collaborator to resource controller, so there is no simpler way to do it that it's done in this PR.

I haven't seen #7425 before, but the solution to override all these actions is unacceptable, so perhaps we can think of some more generic solution. This is just a single use case and I believe we should have at least 2 use cases to implement a generic solution, so it's better to iterate on that in another PR (#7425).

PR #7365, opened 2017-01-30
Add missing translation keys

ylastapis:
introduced by Sylius\Bundle\UserBundle\Form\Model\PasswordResetRequest

| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/translations/validators.en.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/translations/validators.en.yml
@@ -86,6 +86,9 @@ sylius:
     user:
         email:
             unique: This email is already used.
+            min: Minimal number of caracteres not reached.
+            max: Maximal number of caracteres reached.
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/CoreBundle/Resources/translations/validators.en.yml
```diff
@@ -86,6 +86,9 @@ sylius:
     user:
         email:
             unique: This email is already used.
+            min: Minimal number of caracteres not reached.
+            max: Maximal number of caracteres reached.
```
Zales0123: I would vote for something like "Email must have at least X characters." etc.

pamil: caracters -> characters ;)

pjedrzejewski:
Thanks Yann!

PR #7366, opened 2017-01-30
Allow POST on sylius_shop_order_after_pay route

psihius:
There are payment providers that do a POST based redirect back to the store instead of the GET.

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | none |
| License         | MIT |


PyRowMan:
I think that it's not BC Break and it's not a new feature too :)

psihius:
@PyRowMan fixed :)

pjedrzejewski:
Thank you Arvids!

PR #7367, opened 2017-01-30
[ResourceBundle] Prevent overriding existing association mappings

venyii:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

If the metadata already has existing association mappings for the given field, we won't override anything and leave it as is.

**Example:**
In our bundle, we define our own product entity and want to add an `order-by` to the association mapping of an existing field. So we copy the field definition from `Product.orm.xml` to our own configuration and modify it to our needs.

The current code will always override our custom association mappings for a given field. So the fix is to check if the field has custom association mappings and only apply them if none are configured.

Also renamed the `hasRelation` method since it's really checking if it **is** a relation.


pjedrzejewski:
Thanks @venyii, this is a great improvement! :)

PR #7372, opened 2017-01-30
[API] Remove snake_case support for symfony forms

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

During development it appears, that body listener breaks our translation support. We have to revert this changes. With updated tests for product we should be protected from regression.


tests/Controller/CheckoutCompleteApiTest.php
```diff
@@ -64,7 +64,7 @@ public function it_does_not_allow_to_complete_order_that_is_not_addressed_and_ha
     /**
      * @test
      */
-    public function it_allows_to_complete_order_that_is_addressed_and_has_shipping_and_payment_method_selected()
+    public function it_allows_to_complete_order_that_isAddressed_and_has_shipping_and_payment_method_selected()
```
pamil: `isAddressed` -> `is_addressed`

pjedrzejewski:
As discussed in the office, we should try to solve the problem with translation locale's key normalizing. ;)

PR #7375, opened 2017-01-31
[RFC][Documentation] The Book - Image Tiles

TheMadeleine:
| Q               | A
| --------------- | ---
| Doc fix?        | kind of |
| New docs?    | no |
| Related tickets |  |
| License         | MIT |

While being on our Slack #support channel, I must admit that many of the problems come from the misunderstanding of the docs' structure. I think we need to work on the **visability of information.**

That's why I come with an idea of adding visual "Tiles" (yeah, windows style, I know :P) of The Book chapters to the "landing page" of our docs.

👀  The tiles should attract more eyes, and let people know that The Book should be the first place to go if you want to learn sth.

❗️ **Design has changed** ❗️ 

image

⚠️ One problem with them is the `:target:` which simply cannot be ```:doc:`</book/installation>` ``` but a real url of the doc page.

⚠️ And of course! We can make them more visually attractive. This is just an RFC :D

❓ All in all, **what's your opinion**? Maybe some other ideas to improve the docs readability?

Cheers 👋 , Maggie.


pjedrzejewski:
Awesome, I'd just change the fonts to match the rest of documentation. The subtitle looks comic-sans like. Excellent idea @TheMadeleine!

lchrusciel:
Can we capitalize a first letter of each word on this tiles?

docs/index.rst
```diff
@@ -19,12 +19,46 @@ The Book
...
+   :target: `Themes`_
+   :scale: 55%
+
 .. toctree::
    :hidden:
 
    book/index
 
-.. include:: /book/map.rst.inc
+|
```
tuka217: Should be just `|`?

TheMadeleine: ? I don't understand.

`|` is in sphinx a line break :)

pjedrzejewski:
Woo, nice work Maggie! :)

PR #7376, opened 2017-01-31
[API] Fix normalization of locale codes

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

During development it appears, that the default body listener breaks our translation support. It changed locale codes from `en_US` to `enUS` and therefore, we weren't able to update resource translation via API. I have added a custom listener, which is locale codes aware, so it will just skip these keys. With updated tests for product we should be protected from regression.

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Normalizer/LocaleAwareCamelKeysNormalizer.php
+++ b/src/Sylius/Bundle/ApiBundle/Normalizer/LocaleAwareCamelKeysNormalizer.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Bundle\ApiBundle\Normalizer;
+
+use FOS\RestBundle\Normalizer\CamelKeysNormalizer;
+
+/**
+ * Normalizes the array by changing its keys from underscore to camel case, but does not perform this change if the key is an locale code.
+ *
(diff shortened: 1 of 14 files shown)


src/Sylius/Bundle/ApiBundle/Normalizer/LocaleAwareCamelKeysNormalizer.php
```diff
@@ -0,0 +1,43 @@
...
+namespace Sylius\Bundle\ApiBundle\Normalizer;
+
+use FOS\RestBundle\Normalizer\CamelKeysNormalizer;
+
+/**
+ * Normalizes the array by changing its keys from underscore to camel case, but does not perform this change if the key is an locale code.
+ *
+ * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
+ */
+final class LocaleAwareCamelKeysNormalizer extends CamelKeysNormalizer
```
pamil: No unit tests 😿

lchrusciel: There is end-to-end test for that. Do we need another one?

lchrusciel: On it! 🐋

lchrusciel: @pamil [fair enough](https://github.com/Sylius/Sylius/pull/7376/files#diff-c9592b3945ff51f7fecd53b74d396916R19)?

pjedrzejewski:
Thanks Łukasz!

PR #7384, opened 2017-01-31
[Core] [Addressing] [Currency] [Order] [Product] [Review] Repositories cleanup

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes(new ORMDriver dependencies) |
| Related tickets |  |
| License         | MIT |

I have removed a repository definition from configuration tree to decouple bundles from ORM. Unfortunately, the ORM Driver didn't resolved properly repository classes. 

My proposal is to gather repository from repository registry, but we need an alias for that, which can be taken from Metadata registry based on class name. The drawback of this solution is that it will couple GridBundle with ResourceBundle...

Other possible solutions would be refactoring of repository registry to use models FQCN  as a key, or specify or specify resource alias instead of class in grid configuration.

I have chosen this solution, as it affects the smallest part of our code base, but if you think that there is a better one, I'm open for proposals.


lchrusciel:
Closing in order to find a better solution

PR #7387, opened 2017-01-31
[BugFix] Do not clear the object manage between fixture loads

### LINKED ISSUE #6051: Identity map issues with new sylius fixture loader
The `AbstractResourceFixture::load` clears the entire entity manager, this will clear and detach any existing references that happen outside the current fixture. In my case I have a listener that sets a relationship on a product when it is pre persisted. But because the entity manager is cleared the reference is gone and a `A new entity was found through the relationship` error is triggered. Is it really necessary to clear the entity manager for fixtures?

koemeet:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #6051  |
| License         | MIT |

We should not clear the object manager by default since this can have nasty side effects. I think it should be removed and later made configurable in some way.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/AbstractResourceFixture.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/AbstractResourceFixture.php
@@ -83,12 +83,10 @@ abstract class AbstractResourceFixture implements FixtureInterface
 
             if (0 === ($i % 10)) {
                 $this->objectManager->flush();
-                $this->objectManager->clear();
             }
         }
 
         $this->objectManager->flush();
-        $this->objectManager->clear();
     }
 
     /**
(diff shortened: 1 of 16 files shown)


pjedrzejewski:
Definitely we should make it configurable, as it can get pretty sluggish with a lot of entities.

michalmarcinkowski:
@steffenbrem what's the status of this PR? It would be good to merge it before 1.0.0-beta.2 tag.

michalmarcinkowski:
@steffenbrem will you have time to work on this?

steffenbrem:
I've made the first attempt to solve this. I couldn't find any other clean way other than injecting the `Suite` into the `FixtureInterface::loader` method. This way you have access to the suite in every fixture. Is this an acceptable implementation?

pjedrzejewski:
@pamil WDYT?

Zales0123:
It's been a long time... @pamil could you take a look at this? I wonder do we need to iterate over it ⚖️

lchrusciel:
I suppose it should be part of FixtureBundle + it seems to not be the best solution either

pamil:
I guess the overall idea is good (passing Suite to the Fixture), but we need to think on a generic solution that will not require to change every fixture and is more specific to FixturesBundle than CoreBundle.

PR #7397, opened 2017-02-01
[BugFix] Remove final keyword from some Behat pages

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | none |
| License         | MIT |

This should not be final, since you should be able to extend these pages. These are the only behat pages that are final, probably by accident.


steffenbrem:
Shall I recommit something, since the build itself failed for an invalid reason.

steffenbrem:
Ready 👍

pjedrzejewski:
Thanks Steffen!

Issue #6117, opened 2016-09-14
[Api] Creating product attribute error: "`createNew()` not supported, use `createdType()` instead"


koemeet:
The type should either be passed in the url path, so `POST /product-attributes/text` for example. Or we should maybe enable `createNew()` so the `type` could be passed in the actual payload, which is in my opinion nicer. We could use a doctrine event listener to set the `storageType` automatically if needed or only do it on the form in a `POST_SUBMIT` listener.

pjedrzejewski:
I am in favor of including the type in the payload, the only reason not to do this in UI was to simplify the forms (no need to dynamic JS), in API it should work like a charm when sent in payload. 👍

pjedrzejewski:
It is passed as "type" via request now and we keep it this way in 1.0. :)

PR #7398, opened 2017-02-02
Product attribute creation must be typed

gruberro:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | #6117 |
| License         | MIT |

This PR aims to fix product attribute creation via api as the configured factory does not allows creating untyped attributes (https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Attribute/Factory/AttributeFactory.php#L48-L53).

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/routing/product_attribute.yml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/routing/product_attribute.yml
@@ -20,6 +20,10 @@ sylius_api_product_attribute_create:
         _controller: sylius.controller.product_attribute:createAction
         _sylius:
             serialization_version: $version
+            factory:
+                method: createTyped
+                arguments:
+                    type: $type
 
 sylius_api_product_attribute_update:
     path: /{id}


pjedrzejewski:
I think in case of API we can simply send the type together with the configuration, etc. So "type" should be the part of the payload. Factory can be changed if needed. WDYT?

gruberro:
That would be fine in my mind too. Didn't saw a better solution than adding the type this way as the factory can only be invoked containing the type...

gruberro:
Did some tests with `$type` as simple payload param... works as expected too 👍

michalmarcinkowski:
Thank you Robert! 👍

pjedrzejewski:
The reason why I have not merged this is that we need tests... @gruberro Could you write them? It is really straightforward, you can have a look at our existing API tests in ``tests`` directory of this repo.

michalmarcinkowski:
I decided to merge it since we don't have tests for it right now anyway, so it's better to have it fixed than broken, but @pjedrzejewski you're right, I should have asked for tests!

Issue #7373, opened 2017-01-30
Error when updating cart from different browser windows


gonzalovilaseca:
It's not a common case, but we've had a few, the way to reproduce it in http://demo.sylius.org is:
1) Add any 3 items to cart (`window1`)
2) Open a new browser window and navigate to cart page (`window2`).
3) Go back to `window1`, remove 2 items from cart.
4) Goto `window2`, change any item quantity and save cart, you should get "500 Internal Server Error".

Somehow symfony form transformation gets messed up.

michalmarcinkowski:
Fixed in #7406.

Issue #7212, opened 2017-01-08
Changing taxes amount does not change the amount of the carts


gabiudrescu:
Initially reported in #7167:

Preconditions:
 - you have a product with a 10% tax
 - you have that product in a cart (unfinished order)

Steps to reproduce:
 - load the cart and go to the last step of the checkout process
 - change to tax to 20%
 - refresh the cart

What happens:
 - the tax amount remains unchanged
 - you can place an order with the old tax amount

What should happen:
 - the taxes should be recalculated immediately after they were changed (e.g. refreshing a cart step)
 - the taxes should be recalculated when the order is placed (e.g. after pressing place order button)
 - the customer should be notified the cart amount has changed

----

Question: any idea how the customer should be notified about this change in a manner that does not affect conversion rate :) ?

pjedrzejewski:
Maybe it is stupid, but I remember we did that on purpose (to avoid surprising the customer with a different order than he saw on the summary). I think we should probably still recalculate it, but if anything has changed, we shall display an info flash and show the updated summary. Anyway, we have this issue in every step of the checkout.

We should display a flash message if cart is updated due to some configuration/inventory change at any point of the customer's visit in the shop, no matter if it is last step. Hopefully, we will take a look at this before BETA2 release.

gabiudrescu:
issuing a flash whenever cart is updated sounds good - if you are on any checkout step.

michalmarcinkowski:
The cart is recalculated on each refresh and through the whole checkout process. Right now if the order has changed just before we placed it, we will display a flash message that something changed and the customer will be able to review the order once again before placing. We are not be able to place an invalid order anymore, so I assume this can be closed.

PR #7406, opened 2017-02-02
[Checkout] Order shipping method integrity validation

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| Related tickets | #7212, #7373|
| License         | MIT |

Currently in sylius we do not check integrity of the cart it causes scenarios like this possible:
```yaml
Background:
        And the store has "Heavy stuff" shipping category
        And the store has "Small stuff" shipping category
        And the store allows paying offline
        And the store allows shipping with "DHL"
        And this shipping method requires that all units match to "Heavy stuff" shipping category
        And the store allows shipping with "FedEx"
        And this shipping method requires at least one unit matches to "Small stuff" shipping category
        And the store has a product "WestWorld host" priced at "$12.54"
        And this product belongs to "Heavy stuff" shipping category
        And the store has a product "T-shirt Breaking Bad" priced at "$12.54"
        And this product belongs to "Small stuff" shipping category

    @ui
    Scenario: Validate shipping method after adding item which does not fit shipping method requirements
        Given I added product "WestWorld host" to the cart
        And I have completed addressing step with email "guest@example.com" and "United States" based shipping address
        And I have proceeded order with "DHL" shipping method and "Offline" payment
        When I add product "T-shirt Breaking Bad" to the cart
        And I complete checkout
        Then I should see the thank you page
```

### DIFF
--- a/features/checkout/order_promotion/order_promotion_integrity_validation.feature
+++ b/features/checkout/order_promotion/order_promotion_integrity_validation.feature
@@ -1,0 +1,24 @@
+@checkout
+Feature: Order promotions integrity
+    In order to have valid promotions applied on my order
+    As a Customer
+    I want to have information about promotion changes on my order
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store allows paying offline
+        And the store ships everywhere for free
+        And the store has a product "PHP T-Shirt" priced at "$100.00"
+        And there is a promotion "Christmas sale"
+        And this promotion gives "$10.00" discount to every order
+        And this promotion expires tomorrow
+        And I am a logged in customer
+
+    @ui
+    Scenario: I proceed to checkout summary page with a promotion, which expires in 2 hours, but I complete my checkout next day
+        Given I added product "PHP T-Shirt" to the cart
+        And I have proceeded selecting "Offline" payment method
+        And this promotion has already expired
+        When I confirm my order
+        Then I should be informed that this promotion is no longer applied
+        And I should not see the thank you page
--- a/features/checkout/order_promotion/order_promotion_integrity_validation.feature
+++ b/features/checkout/order_promotion/order_promotion_integrity_validation.feature
@@ -15,7 +15,7 @@ Feature: Order promotions integrity
         And I am a logged in customer
 
     @ui
-    Scenario: I proceed to checkout summary page with a promotion, which expires in 2 hours, but I complete my checkout next day
+    Scenario: Preventing customer from completing checkout with already expired promotion
         Given I added product "PHP T-Shirt" to the cart
         And I have proceeded selecting "Offline" payment method
         And this promotion has already expired
--- a/features/checkout/paying_for_order/payment_method_integrity_validation.feature
+++ b/features/checkout/paying_for_order/payment_method_integrity_validation.feature
@@ -12,7 +12,7 @@ Feature: Order payment method integrity
         And I am a logged in customer
 
     @ui
-    Scenario: Validate payment method which was disabled by administrator
+    Scenario: Preventing customer from completing checkout with no longer available payment method
         Given I have product "PHP T-Shirt" in the cart
         And I have proceeded selecting "Offline" payment method
         But this payment method has been disabled
--- a/features/checkout/paying_for_order/payment_method_integrity_validation.feature
+++ b/features/checkout/paying_for_order/payment_method_integrity_validation.feature
@@ -1,0 +1,21 @@
+@paying_for_order
+Feature: Order payment method integrity
+    In order to have valid payment method
+    As a Customer
+    I want to have valid order
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "PHP T-Shirt" priced at "$19.99"
+        And the store ships everywhere for free
+        And the store allows paying offline
+        And I am a logged in customer
+
+    @ui
+    Scenario: Validate payment method which was disabled by administrator
+        Given I have product "PHP T-Shirt" in the cart
+        And I have proceeded selecting "Offline" payment method
+        But this payment method has been disabled
+        When I confirm my order
+        Then I should be informed that this payment method has been disabled
+        And I should not see the thank you page
--- a/features/checkout/product_integrity_validation.feature
+++ b/features/checkout/product_integrity_validation.feature
@@ -1,0 +1,21 @@
+@paying_for_order
+Feature: Order products integrity
+    In order to have valid products
+    As a Customer
+    I want to have enabled products in my order
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "PHP T-Shirt" priced at "$19.99"
+        And the store ships everywhere for free
+        And the store allows paying offline
+        And I am a logged in customer
+
+    @ui
+    Scenario: Validate that products in order are not disabled
+        Given I have product "PHP T-Shirt" in the cart
+        And I have proceeded selecting "Offline" payment method
+        But this product has been disabled
+        When I confirm my order
+        Then I should be informed that this product has been disabled
+        And I should not see the thank you page
--- a/features/checkout/product_integrity_validation.feature
+++ b/features/checkout/product_integrity_validation.feature
@@ -12,7 +12,7 @@ Feature: Order products integrity
         And I am a logged in customer
 
     @ui
-    Scenario: Validate that products in order are not disabled
+    Scenario: Preventing customer from completing checkout with no longer available products
         Given I have product "PHP T-Shirt" in the cart
         And I have proceeded selecting "Offline" payment method
         But this product has been disabled
--- a/features/checkout/shipping_order/shipping_method_integrity_validation.feature
+++ b/features/checkout/shipping_order/shipping_method_integrity_validation.feature
@@ -44,7 +44,7 @@ Feature: Order shipping method integrity
         Given I added product "Westworld host" to the cart
         And I have completed addressing step with email "guest@example.com" and "United States" based shipping address
         And I have proceeded order with "DHL" shipping method and "Offline" payment
-        And product "Westworld host" has its shipping category changed to "Small stuff"
+        And product "Westworld host" shipping category has been changed to "Small stuff"
         When I want to complete checkout
         Then I should not be able to confirm order because products does not fit "DHL" requirements
         And I should not see the thank you page
--- a/features/checkout/shipping_order/shipping_method_integrity_validation.feature
+++ b/features/checkout/shipping_order/shipping_method_integrity_validation.feature
@@ -1,0 +1,50 @@
+@checkout
+Feature: Order shipping method integrity
+    In order to have valid shipping method on my order
+    As a Visitor
+    I want to have valid order
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has "Heavy stuff" shipping category
+        And the store has "Small stuff" shipping category
+        And the store allows paying offline
+        And the store allows shipping with "DHL"
+        And this shipping method requires that all units match to "Heavy stuff" shipping category
+        And the store allows shipping with "FedEx"
+        And this shipping method requires at least one unit matches to "Small stuff" shipping category
+        And the store has a product "WestWorld host" priced at "$12.54"
+        And this product belongs to "Heavy stuff" shipping category
+        And the store has a product "T-shirt Breaking Bad" priced at "$12.54"
+        And this product belongs to "Small stuff" shipping category
+
+    @ui
+    Scenario: Validate shipping method after adding item which does not fit shipping method requirements
+        Given I added product "WestWorld host" to the cart
+        And I have completed addressing step with email "guest@example.com" and "United States" based shipping address
+        And I have proceeded order with "DHL" shipping method and "Offline" payment
+        When I add product "T-shirt Breaking Bad" to the cart
+        And I want to complete checkout
+        Then I should not be able to confirm order because product "T-shirt Breaking Bad" does not belong to "Heavy stuff" shipping category
+        And I should not see the thank you page
+
+    @ui @javascript
+    Scenario: Validate shipping method after removing item which fits shipping method requirements
+        Given I added product "T-shirt Breaking Bad" to the cart
+        And I added product "WestWorld host" to the cart
+        And I have completed addressing step with email "guest@example.com" and "United States" based shipping address
+        And I have proceeded order with "FedEx" shipping method and "Offline" payment
+        When I remove product "T-shirt Breaking Bad" from the cart
+        And I want to complete checkout
+        Then I should not be able to confirm order because product "WestWorld host" does not belong to "Small stuff" shipping category
+        And I should not see the thank you page
+
+    @ui
+    Scenario: Validate shipping method after administrator changes in shipping method requirements
+        Given I added product "WestWorld host" to the cart
+        And I have completed addressing step with email "guest@example.com" and "United States" based shipping address
+        And I have proceeded order with "DHL" shipping method and "Offline" payment
+        When administrator change shipping category to "Small stuff" for product "WestWorld host"
+        And I want to complete checkout
+        Then I should not be able to confirm order because product "WestWorld host" does not belong to "Heavy stuff" shipping category
+        And I should not see the thank you page
--- a/features/checkout/shipping_order/shipping_method_integrity_validation.feature
+++ b/features/checkout/shipping_order/shipping_method_integrity_validation.feature
@@ -25,7 +25,7 @@ Feature: Order shipping method integrity
         And I have proceeded order with "DHL" shipping method and "Offline" payment
         When I add product "T-shirt Breaking Bad" to the cart
         And I want to complete checkout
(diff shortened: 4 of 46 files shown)


features/checkout/shipping_order/shipping_method_integrity_validation.feature
```diff
@@ -0,0 +1,50 @@
...
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has "Heavy stuff" shipping category
+        And the store has "Small stuff" shipping category
+        And the store allows paying offline
+        And the store allows shipping with "DHL"
+        And this shipping method requires that all units match to "Heavy stuff" shipping category
+        And the store allows shipping with "FedEx"
+        And this shipping method requires at least one unit matches to "Small stuff" shipping category
+        And the store has a product "WestWorld host" priced at "$12.54"
```
pamil: WestWorld -> Westworld

src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -675,6 +675,14 @@ public function thisProductBelongsToShippingCategory(ProductInterface $product,
     }
 
     /**
+     * @When administrator change shipping category to :shippingCategory for product :product
```
pamil: administrator change**s**

or even

product "Westworld host" has/gets its shipping category changed to "Small stuff"

src/Sylius/Behat/Page/Shop/Checkout/CompletePage.php
```diff
@@ -291,6 +297,25 @@ public function getShippingPromotionDiscount($promotionName)
...
      */
+    public function tryToOpen(array $urlParameters = [])
+    {
+        if ($this->getDriver() instanceof Selenium2Driver) {
+            $start = microtime(true);
+            $end = $start + 15;
+            do {
+                parent::tryToOpen($urlParameters);
+                sleep(1);
+            } while(!$this->isOpen() && microtime(true) < $end);
```
pamil: That has no sense, if a page isn't open after one second, it will be reopened once again etc. so maximum loading time is always 1s but repeated 15 times

src/Sylius/Behat/Page/Shop/Checkout/CompletePageInterface.php
```diff
@@ -178,4 +178,9 @@ public function getBaseCurrencyOrderTotal();
      * @return string
      */
     public function getShippingPromotionDiscount($promotionName);
+
+    /**
+     * @return mixed
```
pamil: string?

src/Sylius/Behat/Page/Shop/Checkout/CompletePage.php
Arminek: It will try to open page then it will wait 1s. If is not opened and actual time is less than timeout. It will repeat.

src/Sylius/Bundle/CoreBundle/EventListener/OrderPromotionIntegrityChecker.php
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * @param ResourceControllerEvent $event
+     */
+    public function check(ResourceControllerEvent $event)
+    {
+        /** @var OrderInterface $order */
+        $order = $event->getSubject();
+
+        if (!$order instanceof OrderInterface) {
```
Zales0123: Why not use `Assert::isInstanceOf()`?

src/Sylius/Bundle/CoreBundle/Resources/translations/validators.en.yml
```diff
@@ -30,6 +30,9 @@ sylius:
     cart_item:
         not_available: '%itemName% does not have sufficient stock.'
     order:
+        product_eligibility: 'This product %productName% has been disabled.'
+        payment_method_eligibility: 'This payment method %paymentMethodName% has been disabled. Please reselect your payment method.'
+        shipping_method_eligibility: 'Product does not fit requirements for %shippingMethodName% shipping method. Please reselect your shipping method.'
```
Zales0123: Alphabetical order is always a better choice :(

src/Sylius/Bundle/CoreBundle/Validator/Constraints/OrderPaymentMethodEligibilityValidator.php
```diff
@@ -0,0 +1,51 @@
...
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+
+/**
+ * @author Arkadiusz Krakowiak <arkadiusz.krakowiak@lakion.com>
+ */
+final class OrderPaymentMethodEligibilityValidator extends ConstraintValidator
+{
+    /**
+     * @param OrderInterface $value
```
Zales0123: Why should we type-hint like this? It's not necessarily an `OrderInterface`, that's why it's checked few lines below. And the variable name does not match ;)

src/Sylius/Bundle/CoreBundle/Validator/Constraints/OrderPaymentMethodEligibilityValidator.php
```diff
@@ -0,0 +1,51 @@
...
+                    OrderInterface::class,
+                    get_class($order)
+                )
+            );
+        }
+
+        $payments = $order->getPayments();
+
+        foreach ($payments as $payment) {
+            if(!$payment->getMethod()->isEnabled()) {
```
Zales0123: Missing space after `if`.

src/Sylius/Component/Order/spec/Model/OrderSpec.php
```diff
@@ -482,4 +482,11 @@ function it_has_notes()
         $this->setNotes('something squishy');
         $this->getNotes()->shouldReturn('something squishy');
     }
+
+    function it_can_copy_it_self()
```
Zales0123: `it_can_copy_itself`

src/Sylius/Behat/Page/Shop/Checkout/CompletePage.php
pamil: So if it takes over 1s to load that page, it will repeat it 15 times and always fail instead of waiting a second or two more to get the page loaded.

features/checkout/order_promotion/order_promotion_integrity_validation.feature
```diff
@@ -0,0 +1,24 @@
...
+        And the store allows paying offline
+        And the store ships everywhere for free
+        And the store has a product "PHP T-Shirt" priced at "$100.00"
+        And there is a promotion "Christmas sale"
+        And this promotion gives "$10.00" discount to every order
+        And this promotion expires tomorrow
+        And I am a logged in customer
+
+    @ui
+    Scenario: I proceed to checkout summary page with a promotion, which expires in 2 hours, but I complete my checkout next day
```
pamil: Preventing customer from completing checkout with already expired promotion?

features/checkout/paying_for_order/payment_method_integrity_validation.feature
```diff
@@ -0,0 +1,21 @@
...
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "PHP T-Shirt" priced at "$19.99"
+        And the store ships everywhere for free
+        And the store allows paying offline
+        And I am a logged in customer
+
+    @ui
+    Scenario: Validate payment method which was disabled by administrator
```
pamil: Preventing customer from completing checkout with no longer available payment method?

features/checkout/product_integrity_validation.feature
```diff
@@ -0,0 +1,21 @@
...
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "PHP T-Shirt" priced at "$19.99"
+        And the store ships everywhere for free
+        And the store allows paying offline
+        And I am a logged in customer
+
+    @ui
+    Scenario: Validate that products in order are not disabled
```
pamil: Preventing customer from completing checkout with no longer available products?

src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -651,6 +652,7 @@ public function thisProductHasAnImageWithACode(ProductInterface $product, $image
 
     /**
      * @Given /^(this product) belongs to ("([^"]+)" shipping category)$/
+     * @Given product :product has its shipping category changed to :shippingCategory
```
pamil: product :product shipping category has been changed to :shippingCategory (to be consistent with step added above)?

src/Sylius/Behat/Page/Shop/Checkout/CompletePage.php
pamil: Anyway, it's not a logic specific for `CompletePage`.

src/Sylius/Bundle/CoreBundle/EventListener/OrderTotalIntegrityChecker.php
```diff
@@ -0,0 +1,71 @@
...
+                sprintf(
+                    'This checker can only work with "%s", but got "%s".',
+                    OrderInterface::class,
+                    get_class($originalOrder)
+                )
+            );
+        }
+
+        $copiedOrder = $originalOrder->getCopy();
+        $this->orderProcessors->process($copiedOrder);
```
pamil: Shouldn't we process & flush the original one and save the copied for comparison?

src/Sylius/Component/Order/Model/Order.php
```diff
@@ -412,6 +412,14 @@ public function recalculateAdjustmentsTotal()
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function getCopy()
+    {
+        return clone $this;
```
pamil: Isn't custom `__clone` implementation needed?

src/Sylius/Bundle/CoreBundle/Resources/config/services/listeners.xml
```diff
@@ -91,6 +91,7 @@
         <service id="sylius.listener.order_total_integrity_checker" class="Sylius\Bundle\CoreBundle\EventListener\OrderTotalIntegrityChecker">
             <argument type="service" id="sylius.order_processing.order_processor" />
             <argument type="service" id="router" />
+            <argument type="service" id="doctrine.orm.entity_manager" />
```
pamil: `sylius.manager.order`

src/Sylius/Component/Core/Model/OrderItem.php
```diff
@@ -53,8 +53,7 @@ public function setVariant(ProductVariantInterface $variant)
      */
     public function equals(BaseOrderItemInterface $item)
     {
-        return parent::equals($item) || ($item instanceof self && $item->getVariant() === $this->variant)
-        ;
+        return parent::equals($item) || ($item instanceof self && $item->getVariant() === $this->variant);
```
pamil: `self` -> `static`, in case will extend `OrderItem`

src/Sylius/Bundle/CoreBundle/Resources/config/services/listeners.xml
```diff
@@ -81,5 +81,18 @@
...
+            <argument type="service" id="sylius.promotion_eligibility_checker" />
+            <argument type="service" id="event_dispatcher" />
+            <argument type="service" id="router" />
+            <tag name="kernel.event_listener" event="sylius.order.pre_complete" method="check" />
+        </service>
+
+        <service id="sylius.listener.order_total_integrity_checker" class="Sylius\Bundle\CoreBundle\EventListener\OrderTotalIntegrityChecker">
+            <argument type="service" id="sylius.order_processing.order_processor" />
+            <argument type="service" id="router" />
+            <argument type="service" id="doctrine.orm.entity_manager" />
```
pamil: `sylius.manager.order`

src/Sylius/Component/Order/spec/Model/OrderSpec.php
```diff
@@ -482,4 +482,11 @@ function it_has_notes()
         $this->setNotes('something squishy');
         $this->getNotes()->shouldReturn('something squishy');
     }
+
+    function it_can_copy_itself()
+    {
+        $order = $this->getWrappedObject();
+
+        $this->getCopy()->shouldBeLike($order);
```
pamil: Missing `$this->getCopy()->shouldNotReturn($order);`

src/Sylius/Component/Order/Model/OrderInterface.php
```diff
@@ -141,4 +141,9 @@ public function getAdjustmentsTotalRecursively($type = null);
      * @param string|null $type
      */
     public function removeAdjustmentsRecursively($type = null);
+
+    /**
+     * @return self
```
pamil: `self` -> `static`

pjedrzejewski:
Thanks Arek, nice work! :)

src/Sylius/Bundle/CoreBundle/Validator/Constraints/OrderShippingMethodEligibilityValidator.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        Assert::isInstanceOf($order, OrderInterface::class);
+
+        $shipments = $order->getShipments();
+        if ($shipments->isEmpty()) {
+            return;
+        }
+
+        foreach ($shipments as $shipment) {
+            if (!$this->methodEligibilityChecker->isEligible($shipment, $shipment->getMethod())) {
```
michalmarcinkowski: We should also check if the shipping method is not disabled as well as if it's enabled for order's channel.

src/Sylius/Bundle/CoreBundle/EventListener/OrderTotalIntegrityChecker.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    public function check(ResourceControllerEvent $event)
+    {
+        /** @var OrderInterface $order */
+        $order = $event->getSubject();
+
+        Assert::isInstanceOf($order, OrderInterface::class);
+
+        $oldTotal = $order->getTotal();
+        $this->orderProcessors->process($order);
```
michalmarcinkowski: I think we should recalculate the "real" order only after there is at least one error. IMO for this check we should only simulate the recalculation on a cloned object.

pjedrzejewski: Cloning of order is not possible, we discussed this with Arek in the office.

src/Sylius/Bundle/CoreBundle/EventListener/OrderTotalIntegrityChecker.php
```diff
@@ -0,0 +1,78 @@
...
+
+        if ($order->getTotal() !== $oldTotal) {
+            $event->stop('sylius.order.total_integrity', ResourceControllerEvent::TYPE_ERROR);
+            $event->setResponse(new RedirectResponse($this->router->generate('sylius_shop_checkout_complete')));
+
+            $this->manager->persist($order);
+            $this->manager->flush();
```
pjedrzejewski: Where should we flush then? I do not like it either, but that's something that works for now.

michalmarcinkowski: I think the workflow should be the following:

1. Submit the "place order" form
2. Validate the order
2a) If order passed integrity checks, just proceed with default logic
2b) If order did not pass integrity checks, recalculate the order, save it and go back to order summary page

After #7357, we will have a custom `complete` action, so perhaps we can add some logic to the controller. It's for sure better solution that current one.

src/Sylius/Bundle/CoreBundle/EventListener/OrderTotalIntegrityChecker.php
michalmarcinkowski: We would need to perform deep clone to ensure we do not modify anything from the real order. For now we can only clone payments, shipments, order items and order item units of every order item. The adjustments are always removed and new ones added, so can be skipped for the first iteration and the rest of the entities is practically read only.

The following snippet should work for `Order`:
```php
public function __clone()
    {
        $paymentsToClone = null === $this->payments ? new ArrayCollection() : $this->payments;
        $this->payments = new ArrayCollection();
        foreach ($paymentsToClone as $payment) {
            $this->payments->add(clone $payment);
        }

        $shipmentsToClone = null === $this->shipments ? new ArrayCollection() : $this->shipments;
        $this->shipments = new ArrayCollection();
        foreach ($shipmentsToClone as $shipment) {
            $this->shipments->add(clone $shipment);
        }

        $itemsToClone = null === $this->items ? new ArrayCollection() : $this->items;
        $this->items = new ArrayCollection();
        foreach ($itemsToClone as $item) {
            $this->items->add(clone $item);
        }
```

Similar thing for `OrderItem`. It's risky that we can miss something and recalculate only part of the real order and that can cause tragic effect, so I would just remove the `flush` from there for now (as mentioned below) and introduce this change as soon as we will have time to implement it properly (probably after 1.0.0-beta.2 tag). Actually it will not be a BC break.

PR #7410, opened 2017-02-03
[Behat] Improve assertions usage in contexts

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | - |
| License         | MIT |

Changes:
 - [x] `Assert::same($expected, $actual)` changed to `Assert::same($actual, $expected)`, so we can get friendly error messages by default, which leads to removing unnecessary custom messages, therefore making most of them oneliners
 - [x] removed obvious custom messages like `I should see validation error` in `I should be notified about bad credentials` step or `Order should be cancelled, but its not.` for `this order should be automatically cancelled`
 - [x] minor contexts cleanup

### DIFF
--- a/src/Sylius/Behat/Context/Domain/CartContext.php
+++ b/src/Sylius/Behat/Context/Domain/CartContext.php
@@ -60,10 +60,7 @@ final class CartContext implements Context
     {
         $this->expiredCartsRemover->remove();
 
-        Assert::null(
-            $cart->getId(),
-            'This cart should not exist in registry but it does.'
-        );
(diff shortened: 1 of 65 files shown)


pamil:
Or the most annoying one - message `Tax calculation strategy FooBar should be selected, but it is not` for step `the tax calculation strategy for the "loremipsum" channel should be "FooBar"` 🌵

pjedrzejewski:
Thank you Kamil!

PR #7412, opened 2017-02-03
Locale in URLs

### LINKED ISSUE #5694: [RFC][Shop] Locale management in shop
According to [Google's Multi-regional and multilingual sites](https://support.google.com/webmasters/answer/182192?hl=en#1):

> Keep the content for each language on separate URLs. Don’t use cookies to show translated versions of the page. Consider cross-linking each language version of a page. That way, a French user who lands on the German version of your page can get to the right language version with a single click.
> 
> Avoid automatic redirection based on the user’s perceived language. These redirections could prevent users (and search engines) from viewing all the versions of your site.

Right now in ShopBundle the `/switch-language` endpoint is used to change user's language. This  strategy causes mainly SEO issues as only one language can be indexed

Is there any reason to do it in that way?

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | based on storage added in #7486, documentation in #7507, fixes #5694 |
| License         | MIT |

Done:
 - [x] Cookbook how to bring back previous behaviour (editing routing file, security & a few services) - #7507


steffenbrem:
Really nice to see this. Do you plan on making the locale prefix optional for single locale channels?

pamil:
@steffenbrem not right now for sure, it looks quite difficult because every channel might have different default locale. However if you have only one channel or one default locale, you can just import shop routing with `default: { _locale: pt_PT }` and it should work well (+ some security access control enhancements).

pjedrzejewski:
@pamil I think this one definitely deserves a note in UPGRADE. :D

pamil:
@pjedrzejewski done in #7450 🎉

liverbool:
after `beta.2` pls. 😷

pjedrzejewski:
@liverbool Will it cause you a lot of problems?

tchapi:
For us too, I would prefer that you tag the **beta2** before, because we have a routing based on the channel, which I think makes a lot more sense that the one you propose on `locale`. So we'll have a very hard time untangling the whole thing...

tchapi:
Moreover it's a huge BC break and I though that only very necessary breaks were due to happen before the beta... I would definitely consider this to be a kind of option. Will this deactivable altogether ?

psihius:
The other side of this thing is - "I need this YESTERDAY!"
My clients SEO people are going nuts to have language in the URL - we have 3 languages in the store, probably will have even more. Translated product and taxon slugs can take you only so far - it's not the whole store.

liverbool:
I try to keep up-to-date my projects with Sylius latest version, I've few projects go live with current route structure, I think this is big one BC break!! we have a lot of works to do.

pjedrzejewski:
OK, so I guess the solution is - make this behavior optional. Or make reverting to single-locale much simpler. I think it should be possible without much magic. @pamil Can we provide both implementations? That would be most-awesome scenario. :D Only problem I see is testing them.

pamil:
I'm currently working on being able to revert to previous behaviour without too much hassle (routing, security and storage service).

pjedrzejewski:
Thanks Kamil, awesome work!

kpitn:
This is a great feature, but in term of seo there is a problem. (Duplicate content)

My shop has 2 locals (fr_FR and en_US)

All this url works :
http://localhost:8000/fr/products/tractopelle-jcb-4cx-aec-15610
http://localhost:8000/fr_FR/products/tractopelle-jcb-4cx-aec-15610
http://localhost:8000/fu_CK/products/tractopelle-jcb-4cx-aec-15610

I think you need to keep the possiblity to have 2 letters locale.

gorkalaucirica:
@kpitn It depends on the use case, some people may want to have the full locale on the url. In case you don't need it, you can create some rules in `.htaccess` to avoid SEO issues with content duplication redirecting as required.

I guess you can also select enabled languages by channel, but didn't test that with this new feature

liverbool:
I thinks this

```
http://sylius.shop/products <= default locale
http://sylius.shop/pl_PL/products <= alternative locale
```

kpitn:
@gorkalaucirica

The main problem of my message is that even with two locale set in backoffice.
You can set any string matching /[a-z]{2}\_[A-Z]{2}/ the page respond with code 200.

I think you need to check if locale exist, otherwise throw a 404.

pjedrzejewski:
@kpitn Thanks for reporting, we will fix that! :)

PR #7419, opened 2017-02-04
Payment state update for NotifyNullAction

psihius:
Payum provides a mechanism to use a NotifyNullAction for gateways that do not have a per-payment notification URL capability to call the "payum_notify_do_unsafe" route for the URL "/payment/notify/unsafe/{gateway}"

The issue happens because in this scenario, the NotifyNullAction is the last item in the Payum processing stack, as it is the first action that has been called, and due to it's stateless nature, it does not have a PaymentInterface model (getFirstModel returns "null"), meaning the extension can't update the payment status.

This means that Payum's built-in Be2Bill should fail too, but I have my own local gateway provider that has the same workflow. It took me quite a while to figure it out and I have done extensive testing on this.
This is the best idea I came up with how to handle it.

In essence, I look up the next action in the stack and if I see a class names "NotifyNullAction" - I'm not doing a "return" as it done originally.

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | none |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
+++ b/src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
@@ -56,11 +56,13 @@ final class UpdatePaymentStateExtension implements ExtensionInterface
     public function onPostExecute(Context $context)
     {
         $previousStack = $context->getPrevious();
-        $stackSize = count($previousStack);
+        $previousStackSize = count($previousStack);
         
-        if ($stackSize > 1) {
+        if ($previousStackSize > 1) {
             return;
-        } elseif ($stackSize === 1) {
+        } 
+        
+        if ($previousStackSize === 1) {
             $previousActionClassName = get_class($previousStack[0]->getAction());
             if (false === stripos($previousActionClassName, 'NotifyNullAction')) {
                 return;
--- a/src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
+++ b/src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
@@ -55,8 +55,16 @@ final class UpdatePaymentStateExtension implements ExtensionInterface
      */
     public function onPostExecute(Context $context)
     {
-        if ($context->getPrevious()) {
+        $previousStack = $context->getPrevious();
+        $stackSize = count($previousStack);
+        
+        if ($stackSize > 1) {
             return;
+        } elseif ($stackSize === 1) {
+            $previousActionClassName = get_class($previousStack[0]->getAction());
+            if (false === stripos($previousActionClassName, 'NotifyNullAction')) {
+                return;
+            }
         }
 
         /** @var Generic $request */


pjedrzejewski:
@Arminek thoughts?

src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
```diff
@@ -55,8 +55,16 @@ public function onExecute(Context $context)
...
     {
-        if ($context->getPrevious()) {
+        $previousStack = $context->getPrevious();
+        $stackSize = count($previousStack);
+        
+        if ($stackSize > 1) {
             return;
+        } elseif ($stackSize === 1) {
+            $previousActionClassName = get_class($previousStack[0]->getAction());
+            if (false === stripos($previousActionClassName, 'NotifyNullAction')) {
```
pamil: Is there any reason why FQCN is not used?

src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
```diff
@@ -55,8 +55,16 @@ public function onExecute(Context $context)
      */
     public function onPostExecute(Context $context)
     {
-        if ($context->getPrevious()) {
+        $previousStack = $context->getPrevious();
+        $stackSize = count($previousStack);
+        
+        if ($stackSize > 1) {
             return;
+        } elseif ($stackSize === 1) {
```
pamil: Coding style: we can replace this `elseif` with a regular `if`, these conditions are not coupled to each other.

src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
pamil: Like:
```php
$action = $previousStack[0]->getAction();
if (!$action instanceof NotifyNullAction) // ...
```

psihius: @pamil Because this is gateway-specific implementation and Payum does not have a special interface for these - for every gateway that has to implement this, it will sit inside that gateway's extension namespace. That's why we can't use instanceof - those will be different types all the time.

src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
```diff
@@ -55,8 +55,16 @@ public function onExecute(Context $context)
      */
     public function onPostExecute(Context $context)
     {
-        if ($context->getPrevious()) {
+        $previousStack = $context->getPrevious();
+        $stackSize = count($previousStack);
```
psihius: Yep, makes sense

src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
psihius: Updated

Arminek:
@psihius Imo you should override `sylius.payum_extension.update_payment_state` service and do anything you want.

psihius:
@Arminek But this is a general issue with the Payum integration in general on Sylius side, not a shop-specific issue.
This will affect all gateways in Payum, that implement NotifyNullAction action class - without this fix they will not update payment / order states at all when notified by the gateway. And there are quite a few that use NotifyNullAction, including a gateway Be2Bill that is in Payum's core - per this if anyone uses Be2Bill gateway, it will not work out of the box.

pjedrzejewski:
Thanks Arvids!

psihius:
Y'r welcome

PR #7424, opened 2017-02-06
Allow easier decoration for 3rd party extensions

### LINKED ISSUE #7403: [DOCS] Customizing Factories -> decoration pattern problem
Hi,
following the [docs](http://docs.sylius.org/en/latest/customization/factory.html) I wanted to customize the ProductFactory. I did everything like in the docs and got an error that parameter 1 of constructor was TranslableFactory not one implementing the ProductFactoryInterface. So I changed my service definition to:
```
services:
    app.factory.product:
        class: SyliusExtensionBundle\Factory\ProductFactory
        decorates: sylius.custom_factory.product
        arguments: ['@app.factory.product.inner']
        public: false
```
I decorated `sylius.custom_factory.product` not the` sylius.factory.product` itself. This workarounds works fine, but as I check in configuration `sylius.custom_factory.product` is a decorator of `sylius.factory.product`.  I read in symfony [docs](http://symfony.com/doc/current/service_container/service_decoration.html), there can be multiple decorators "stacked" each other like here:
```
foo:
    class: Foo

bar:
    class: Bar
    public: false
    decorates: foo
    decoration_priority: 5
    arguments: ['@bar.inner']

baz:
    class: Baz
    public: false
    decorates: foo
    decoration_priority: 1
    arguments: ['@baz.inner']
```

And the decorator should be private and referenced throw the base one (`foo` in our example) - following this pattern I shouldn't decorates `sylius.custom_factory.product`. Could you tell me if I do right or there is some bug (in my code or sylius) why I can decorate `sylius.factoy.product`?

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7403 |
| License         | MIT |

This ensures that custom, 3rd party decorations without specified priority are made after the Sylius ones.

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Resources/config/services.xml
+++ b/src/Sylius/Bundle/AddressingBundle/Resources/config/services.xml
@@ -17,7 +17,7 @@
     </imports>
 
     <services>
-        <service id="sylius.custom_factory.zone" class="Sylius\Component\Addressing\Factory\ZoneFactory" decorates="sylius.factory.zone" public="false">
+        <service id="sylius.custom_factory.zone" class="Sylius\Component\Addressing\Factory\ZoneFactory" decorates="sylius.factory.zone" decoration-priority="256" public="false">
             <argument type="service" id="sylius.custom_factory.zone.inner" />
             <argument type="service" id="sylius.factory.zone_member" />
         </service>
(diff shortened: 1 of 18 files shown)


pjedrzejewski:
Thanks Kamil!

PR #7448, opened 2017-02-08
[Resource] Add possibility to define custom criteria for generated routing

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

During work on API it start to be convenient to use a generated routing, just with codes instead of id as an identifier. With this change it will be possible with no so much additional code to maintain. 

I have also change a routing criteria for shipping category to see if it works.

Some small docs and test refactor included.


pjedrzejewski:
Thanks Łukasz! Could you also document this new functionality? :)

lchrusciel:
Of course. I will do it today.

PR #7462, opened 2017-02-09
Fix redirect to referer with hash.

liverbool:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | ~ |
| License         | MIT |

Fix missing `hash` when redirect to referer.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
@@ -195,14 +195,14 @@ class RequestConfiguration
     /**
      * Get url hash fragment (#text) which is you configured.
      *
-     * @return null|string
+     * @return string
      */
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
```diff
@@ -195,14 +195,14 @@ public function getRedirectRoute($name)
...
-     * @return null|string
+     * @return string
      */
     public function getRedirectHash()
     {
         $redirect = $this->parameters->get('redirect');
 
         if (!is_array($redirect) || empty($redirect['hash'])) {
-            return null;
+            return;
```
pamil: `return '';`

pjedrzejewski:
👍 👍 👍  Thanks Ishmael!

PR #7464, opened 2017-02-09
[HWIOAuthBundle] OAuth tokens aren't updated on reconnexion

### LINKED ISSUE #7463: OAuth tokens aren't updated on reconnection
Hello,

OAuth tokens aren't updated on reconnexion, 
I've debugged it and will post soon a pull request.

bruno-ds:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7463 |
| License         | MIT |

Hello, please find a patch fixing this issue : *OAuth tokens aren't updated on reconnexion*

I've tried to add a new phpspec entry `UserProviderSpec::it_should_update_token_when_oauth_succeed` but I'm not competent enough with phpsec so the test fail, sorry

*edit : I noted the PR has "BC breaks" by error in the synthesis table*

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
+++ b/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
@@ -99,7 +99,7 @@ class UserProvider extends BaseUserProvider implements AccountConnectorInterface
         ]);
 
         if ($oauth instanceof UserOAuthInterface) {
-            return $oauth->getUser();
+            return $this->updateUserByOAuthUserResponse($oauth->getUser(), $response);
         }
 
         if (null !== $response->getEmail()) {
@@ -167,9 +167,12 @@ class UserProvider extends BaseUserProvider implements AccountConnectorInterface
      */
     private function updateUserByOAuthUserResponse(UserInterface $user, UserResponseInterface $response)
     {
-        $oauth = $this->oauthFactory->createNew();
-        $oauth->setIdentifier($response->getUsername());
-        $oauth->setProvider($response->getResourceOwner()->getName());
+
+        if (null == $oauth = $user->getOAuthAccount($response->getResourceOwner()->getName())) {
+            $oauth = $this->oauthFactory->createNew();
+            $oauth->setIdentifier($response->getUsername());
+            $oauth->setProvider($response->getResourceOwner()->getName());
+        }
         $oauth->setAccessToken($response->getAccessToken());
         $oauth->setRefreshToken($response->getRefreshToken());
 
--- a/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
+++ b/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
@@ -167,12 +167,12 @@ class UserProvider extends BaseUserProvider implements AccountConnectorInterface
      */
     private function updateUserByOAuthUserResponse(UserInterface $user, UserResponseInterface $response)
     {
-
         if (null == $oauth = $user->getOAuthAccount($response->getResourceOwner()->getName())) {
             $oauth = $this->oauthFactory->createNew();
             $oauth->setIdentifier($response->getUsername());
             $oauth->setProvider($response->getResourceOwner()->getName());
         }
+
         $oauth->setAccessToken($response->getAccessToken());
         $oauth->setRefreshToken($response->getRefreshToken());
 
--- a/src/Sylius/Bundle/CoreBundle/spec/OAuth/UserProviderSpec.php
+++ b/src/Sylius/Bundle/CoreBundle/spec/OAuth/UserProviderSpec.php
@@ -187,4 +187,31 @@ final class UserProviderSpec extends ObjectBehavior
 
         $this->loadUserByOAuthUserResponse($response)->shouldReturn($user);
     }
+
+    function it_should_update_token_when_oauth_succeed(
+        $oauthRepository,
+        ShopUserInterface $user,
+        UserOAuthInterface $oauth,
+        UserResponseInterface $response,
+        ResourceOwnerInterface $resourceOwner
+    ) {
+        $resourceOwner->getName()->willReturn('google');
+
+        $response->getUsername()->willReturn('username');
+        $response->getAccessToken()->willReturn('new_access_token');
+        $response->getRefreshToken()->willReturn('new_refresh_token');
+        $response->getResourceOwner()->willReturn($resourceOwner);
+
+        $oauthRepository->findOneBy(['provider' => 'google', 'identifier' => 'username'])->willReturn($oauth);
+
+        $user->getOAuthAccount('google')->willReturn($oauth);
+
+        $oauth->getUser()->willReturn($user);
+        
+        $oauth->setAccessToken('new_access_token')->shouldBeCalled();
+        $oauth->setRefreshToken('new_refresh_token')->shouldBeCalled();
+
+        $this->loadUserByOAuthUserResponse($response);
+    }
+
 }


src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -167,9 +167,12 @@ private function createUserByOAuthUserResponse(UserResponseInterface $response)
      */
     private function updateUserByOAuthUserResponse(UserInterface $user, UserResponseInterface $response)
     {
-        $oauth = $this->oauthFactory->createNew();
-        $oauth->setIdentifier($response->getUsername());
-        $oauth->setProvider($response->getResourceOwner()->getName());
+
```
lchrusciel: Redundant blank line

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -167,9 +167,12 @@ private function createUserByOAuthUserResponse(UserResponseInterface $response)
...
+        if (null == $oauth = $user->getOAuthAccount($response->getResourceOwner()->getName())) {
+            $oauth = $this->oauthFactory->createNew();
+            $oauth->setIdentifier($response->getUsername());
+            $oauth->setProvider($response->getResourceOwner()->getName());
+        }
```
lchrusciel: Missing blank line

src/Sylius/Bundle/CoreBundle/spec/OAuth/UserProviderSpec.php
```diff
@@ -187,4 +187,34 @@ function it_should_create_new_user_when_none_was_found(
 
         $this->loadUserByOAuthUserResponse($response)->shouldReturn($user);
     }
+
```
lchrusciel: Too many blank lines(one will be enough)

src/Sylius/Bundle/CoreBundle/spec/OAuth/UserProviderSpec.php
```diff
@@ -187,4 +187,34 @@ function it_should_create_new_user_when_none_was_found(
...
+        $response->getAccessToken()->willReturn('new_access_token');
+        $response->getRefreshToken()->willReturn('new_refresh_token');
+        $response->getResourceOwner()->willReturn($resourceOwner);
+
+        $oauthRepository->findOneBy(['provider' => 'google', 'identifier' => 'username'])->willReturn($oauth);
+
+        $user->getOAuthAccount('google')->willReturn($oauth);
+
+        $oauth->getUser()->willReturn($user);
+
```
lchrusciel: Too many blank lines

lchrusciel:
Can we also add a test for the whole workflow(in `tests/Controller/OauthTokenApiTest.php`)?

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
brunocombodo: done

src/Sylius/Bundle/CoreBundle/spec/OAuth/UserProviderSpec.php
brunocombodo: done

brunocombodo:
@lchrusciel : My changes in `UserProviderSpec` aren't working, I'm not experienced enough to make them work, so I don't think I'll be able to add the whole workflow(in `tests/Controller/OauthTokenApiTest.php`), sorry.
BTW, would someone helps me complete the spec modification?

lchrusciel:
@brunocombodo sure, but I need to understand a use case. Can you present some custom flow, when I could reproduce this bug? Then I will write a test and contribute to your solution :)

brunocombodo:
@lchrusciel I've just read `tests/Controller/OauthTokenApiTest.php`, if I'm not wrong, it test the sylius' own OAuth serveur, it is not concerned by my bug report. The bug is in the client side, not the server side.
More precisely it is in the integration of [HWIOAuthBundle](http://docs.sylius.org/en/latest/cookbook/facebook-login.html), the sylius' user provider has a bug.

in the database, the `sylius_user_oauth`.`access_token` isn't updated.

lchrusciel:
Can you describe your setup and steps for this issue? I would like to reproduce the bug and then help you with this PR, but I don't fully understand your use case.

src/Sylius/Bundle/CoreBundle/spec/OAuth/UserProviderSpec.php
```diff
@@ -187,4 +187,31 @@ function it_should_create_new_user_when_none_was_found(
 
         $this->loadUserByOAuthUserResponse($response)->shouldReturn($user);
     }
+
+    function it_should_update_token_when_oauth_succeed(
```
tuka217: `it_updates`

src/Sylius/Bundle/CoreBundle/spec/OAuth/UserProviderSpec.php
```diff
@@ -187,4 +187,31 @@ function it_should_create_new_user_when_none_was_found(
...
+        $user->getOAuthAccount('google')->willReturn($oauth);
+
+        $oauth->getUser()->willReturn($user);
+        
+        $oauth->setAccessToken('new_access_token')->shouldBeCalled();
+        $oauth->setRefreshToken('new_refresh_token')->shouldBeCalled();
+
+        $this->loadUserByOAuthUserResponse($response);
+    }
+
```
tuka217: Not needed blank line

lchrusciel:
I have tried to reproduce the bug, but I have failed. Everything works fine. Can you explain briefly what is it about? Why should we update access token on UserLoad? Also, can you describe steps how can the problem be reproduced? 

I have also tracked changes, and it appears, that we had this behaviour for a really long time. It was introduced by @kayue in #1628.

brunocombodo:
@lchrusciel : this `UserProvider` class is called when you connect to an OAuth provider.
On the first connection, the `loadUserByOAuthUserResponse` method try to find a previously created relation between the user and the provider by calling : 
```php
        $oauth = $this->oauthRepository->findOneBy([
            'provider' => $response->getResourceOwner()->getName(),
            'identifier' => $response->getUsername(),
        ]);
```
this will fail, so it will enter this section of code, wich in every case create the association between the user and the provider : 
```php
        if (null !== $response->getEmail()) {
            $user = $this->userRepository->findOneByEmail($response->getEmail());
            if (null !== $user) {
                return $this->updateUserByOAuthUserResponse($user, $response);
            }
        }
        return $this->createUserByOAuthUserResponse($response);
```

when creating this association, it will store the tokens (the access_token and the refresh_token).

until then everything works well, but what if the association already exists? 
then this call
```php
        $oauth = $this->oauthRepository->findOneBy([
            'provider' => $response->getResourceOwner()->getName(),
            'identifier' => $response->getUsername(),
        ]);
```

will positively enter there : 
```
        if ($oauth instanceof UserOAuthInterface) {
            return $oauth->getUser();
        }
```

and the code will return the user without having updated the tokens.

In much case I think tis is ok, but what if in the sylius side you want to call the provider's API ? Without updated token, you can't.


I hope this is much clear with this explanation.

bruno-ds:
My bad, I just found out why this bug wasn't reported before : because there is a smarter way to find the OAuth tokens : 

the Symfony's `TokenStorage` `method getToken()` return a `OAuthToken` wich give access to the access token and the refresh token.

example code : 
```php
        /** @var \Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage $tokenStorage*/
         /** @var \HWI\Bundle\OAuthBundle\Security\Core\Authentication\Token\OAuthToken $token */
        $token = $tokenStorage->getToken();
        $ccessToken = $token->getAccessToken();
```

so I let you see if you want to solve this bug since it is not blocking at all

liverbool:
This's a critical bug!

lchrusciel:
@liverbool are you willing to pick it up? Without BC preferable, please, and I will merge it ASAP

liverbool:
I can wait, just notice to someone ;) i see long time ago pass.

bruno-ds:
I close the PR since the proposed code is out of date

PR #7465, opened 2017-02-09
[Bug][CoreBundle] Add missing controller configuration options.

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no|
| BC breaks?      | no|
| Related tickets |  |
| License         | MIT |

I noticed that the core bundle doesn't register controller services for important resources such as `product_image` etc. Since most people will need to interact with those resources, through an API for example, it is wise to let them come with a controller by default.


pjedrzejewski:
Great, thanks Steffen!

PR #7469, opened 2017-02-09
[Order][Shipment] Move sending email logic from state machines to listeners

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | contains #7478 and associated with #7357 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services/listeners.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services/listeners.xml
@@ -81,6 +81,10 @@
             <argument type="service" id="sylius.email_manager.order" />
             <tag name="kernel.event_listener" event="sylius.order.post_complete" method="sendConfirmationEmail" />
         </service>
+        <service id="sylius.listener.shipment_ship" class="Sylius\Bundle\CoreBundle\EventListener\ShipmentShipListener">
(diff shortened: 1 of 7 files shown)


src/Sylius/Bundle/CoreBundle/Resources/config/services/listeners.xml
```diff
@@ -77,6 +77,14 @@
             <argument type="service" id="request_stack" />
             <tag name="kernel.event_listener" event="sylius.order.pre_complete" method="assignCustomerIpToOrder" />
         </service>
+        <service id="sylius.listener.order_complete" class="Sylius\Bundle\CoreBundle\EventListener\OrderCompleteListener">
+            <argument type="service" id="sylius.email_manager.order" />
+            <tag name="kernel.event_listener" event="sylius.order.post_complete" method="sendConfirmationEmail" />
```
pjedrzejewski: This is a wrong event. It should be checkout event, hm?

GSadee: I guess that we have checkout events with order prefix: http://docs.sylius.org/en/latest/book/orders/checkout.html#checkout-related-events

pjedrzejewski:
Thanks Grzesiu!

PR #7483, opened 2017-02-10
[Resource] Return 204 HTTP code for state machine transitions

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |

We used to return a whole serialized service for `applyStateMachineTransitionAction`. But it creates a cohesion problem when working with many transitions together. For example, when you work with checkout in API some actions returns HTTP 204, and some HTTP 200 and whole object. At the end we would have the  following actions on Order:
 * complete -> return 204 if succeed 
 * cancel -> return 200 and order object if succeed 
 * ship -> return 204 if succeed
 * complete payment -> return 200 and payment object and  if succeed 
when all of this actions mainly are making just transitions on the same order object.

With this fix, every update will return 204, and one will have to call endpoint with the `GET` method to receive 200 and object.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
@@ -422,7 +422,7 @@ class ResourceController extends Controller
         $this->eventDispatcher->dispatchPostEvent(ResourceActions::UPDATE, $configuration, $resource);
 
         if (!$configuration->isHtmlRequest()) {
-            return $this->viewHandler->handle($configuration, View::create($resource, Response::HTTP_OK));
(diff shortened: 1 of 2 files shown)


michalmarcinkowski:
The inconsistency is because we use either `updateAction` or `applyStateMachineTransitionAction`. Not sure about the proper solution.

Related discussion explaining why it's implemented in this way https://github.com/Sylius/Sylius/pull/5161 https://github.com/Sylius/Sylius/issues/5222

/cc @steffenbrem @bendavies

bendavies:
@lchrusciel previously said he wanted to make it configurable, i think.

lchrusciel:
@bendavies sure, I have forgotten about it. It will stay as it is, because it will require to big changes right now.

PR #7486, opened 2017-02-10
Add support for storing currency in a cookie instead of session

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | upgrade file in #7503 |
| License         | MIT |

DONE:
 - [x] Storage subcomponent in Resource
 - [x] Rework of currency context internals (removed currency provider concept, reduce responsibility duplication)
 - [x] Customise `StorageBasedCurrencyContext` to use any of the Resource's storages
 - [x] Upgrade file (see #7503)


src/Sylius/Bundle/ResourceBundle/Storage/CookieStorage.php
```diff
@@ -0,0 +1,124 @@
...
+use Symfony\Component\HttpFoundation\Cookie;
+use Symfony\Component\HttpFoundation\ParameterBag;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\HttpKernel\Event\GetResponseEvent;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+final class CookieStorage implements StorageInterface, EventSubscriberInterface
```
lchrusciel: Can we split this logic? I think it would be better to have CookieStorage and CookieStorageUpdater which will use it.

src/Sylius/Component/Core/Currency/CurrencyStorage.php
```diff
@@ -0,0 +1,94 @@
...
+use Sylius\Component\Currency\Model\CurrencyInterface;
+use Sylius\Component\Resource\Storage\StorageInterface;
+
+/**
+ * @author Kamil Kokot <kamil.kokot@lakion.com>
+ */
+final class CurrencyStorage implements CurrencyStorageInterface
+{
+    /**
+     * @var \Sylius\Component\Resource\Storage\StorageInterface
```
lchrusciel: Could be reduced to `StorageInterface` only

src/Sylius/Bundle/ResourceBundle/Storage/CookieStorage.php
pamil: Why? They don't have any sense without each other.

lchrusciel: Becuase we have to totally different logics in one class. Both of them could be replaced separately. Storage is a storage and could be used anywhere. And listener which just updates it is a different concept.

pamil: The one and only responsibility of cookie storage is to allow for accessing and storing values in browser's cookies - which we can get after the request is sent and set just before response is sent back.

Dividing this class into two separate ones will result in implementation leaks (we only set these cookies that aren't already set and store the existing ones separately, it's not "**just** updating"), difficulties in connecting these two together (as the cookie listener is **only** applicable to cookie storage). All of them can be used separately but it would not work correctly until they are used together.

lchrusciel: But there are two reasons why could we change this class, so for me, it violates the SRP. 
Another thing, is that cookie storage is totally valid as a separate class. We use it in currency context. And this part is just a simple storage, which could be easly replaced. But the `though` logic should be excluded from this storage and could be placed in the cookie listener. And the cookie listener is only applicable to cookie storage, but implementation of cookie storage could be replaced.

pamil: Cookie storage as a separate class fails at being cookie storage and it's an array storage (using only response cookies). There isn't any reason to replace the storage part with another implementation. 

If you want to use an external listener, you should add some additional behaviours to the storage itself, making it something more than regular `StorageInterface`:
 - populating the request cookies from the request
 - getting only the response cookies

And that's where implementation is leaking, how would you change storages then? Creating `CookieStorageInterface` extending `StorageInterface` and specifying the communication for that particular reason gives us no advantage. 

The current implementation is self-contained and has *one* responsibility - managing cookies. Feel free to contribute and propose your changes as code, not plain words.

lchrusciel: As a reviewer, I have provided you a logic reasons, why do I think it should be placed in the sperate class. It is not my duty to write it for you. You can agree with them or not. I am also not responsible for merging it or not.
Saying, that this class has _one_ responsibility is like putting all the code into the controller and saying, that this class has also only _one_ responsibility: generate the response. This class updates a storage on each request as well as providing a cookie storage behaviour (I agree, that it is not a typical one).
Anyway, I don't care about the color of your bikeshed. It is working. If it will be merged I'm ok with that. But for me, storage should be a separate class.

pamil: > It is not my duty to write it for you.

Yeah, but I really would prefer to see a real code instead of some generic concepts being thrown. Talking about code quality and proving that some approach is cleaner than the another one without hard evidence is not enough.

I'd like to see the code for the reasons mentioned before, there is a lot of glue needed between these two parts of cookie storage so I'm interested how would you deal with it.

> Saying, that this class has one responsibility is like putting all the code into the controller and saying, that this class has also only one responsibility: generate the response.

That's true at a given level of abstraction and we all rely on that abstraction successfully since the first days of Symfony or other request-response frameworks.

> This class updates a storage on each request as well as providing a cookie storage behaviour (I agree, that it is not a typical one).

Updating storage on master request and updating the response with new values are one of the parts of providing cookie storage behaviour, it would not be fulfilled without them.

pjedrzejewski:
Looks good to me, could you add it to the UPGRADE file and we are done here. :)

pamil:
@pjedrzejewski done in #7503, all green and ready to go 🎉

pjedrzejewski:
Thank you Kamil, nice work!

PR #7485, opened 2017-02-10
[Resource] [API] [Product] Resource criteria improvment

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

**UPDATED**
Another improvement to routing generator. I have used this new feature to generate cart item routing to see if it works properly.

PR #7487, opened 2017-02-11
Product test refactor

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

During my work on #7485 I have realized that we have missed proper serialization groups on product variant. I have fixed that and also provide some small refactor of the test suite.

### DIFF
--- a/docs/api/product_variants.rst
+++ b/docs/api/product_variants.rst
@@ -5,24 +5,28 @@ These endpoints will allow you to easily manage product variants. Base URI is `/
 
 When you get a collection of resources, "Default" serialization group will be used and the following fields will be exposed:
 
-+-------------------+---------------------------------------------------------------------------+
-| Field             | Description                                                               |
-+===================+===========================================================================+
-| id                | Id of product variant                                                     |
-+-------------------+---------------------------------------------------------------------------+
-| code              | Unique product variant's identifier                                       |
-+-------------------+---------------------------------------------------------------------------+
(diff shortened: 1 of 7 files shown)


docs/api/product_variants.rst
```diff
@@ -134,36 +134,13 @@ Example Response
...
                     "id": 1,
                     "code": "319bf720-e91d-36f5-aa8e-8c674e3861fb-variant-0",
-                    "option_values": [
-                        {
-                            "name": "Mug type",
-                            "code": "mug_type_medium"
-                        }
-                    ],
                     "position": 0,
-                    "translations": {
```
pjedrzejewski: Why?

docs/api/product_variants.rst
```diff
@@ -134,36 +134,13 @@ Example Response
                 {
                     "id": 1,
                     "code": "319bf720-e91d-36f5-aa8e-8c674e3861fb-variant-0",
-                    "option_values": [
```
pjedrzejewski: I think this is very useful info, why removing it? :)

lchrusciel: It is just the matter of product serialization. Right now, I just adjusted tests to the current implementation. It is not a problem to bring them back, but I will have to adjust the docs also.

pjedrzejewski:
Thanks Łukasz!

PR #7488, opened 2017-02-13
[Checkout] Skip shipping step if only one shipping method is available

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | based on https://github.com/Sylius/Sylius/pull/7276 |
| License         | MIT |

Of course, we should finish iterating over parent PR before merging it, I just open it to have it already on the PRs list.

To-do:
 - [x] solve problems with failing checkout scenarios due to single shipping method setup

### DIFF
--- a/app/migrations/Version20170214104908.php
+++ b/app/migrations/Version20170214104908.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Sylius\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20170214104908 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_channel ADD skipping_shipping_step_allowed TINYINT(1) NOT NULL');
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_channel DROP skipping_shipping_step_allowed');
+    }
+}
--- a/app/migrations/Version20170214104908.php
+++ b/app/migrations/Version20170214104908.php
@@ -19,6 +19,7 @@ class Version20170214104908 extends AbstractMigration
         $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
 
         $this->addSql('ALTER TABLE sylius_channel ADD skipping_shipping_step_allowed TINYINT(1) NOT NULL');
+        $this->addSql('UPDATE sylius_channel SET skipping_shipping_step_allowed = 0');
     }
 
     /**
--- a/features/channel/managing_channels/adding_channel.feature
+++ b/features/channel/managing_channels/adding_channel.feature
@@ -31,6 +31,7 @@ Feature: Adding a new channel
         And I define its color as "blue"
         And I choose "Euro" as the base currency
         And I choose "English (United States)" as a default locale
+        And I allow to skip shipping step if only one shipping method is available
         And I add it
         Then I should be notified that it has been successfully created
         And the channel "Mobile channel" should appear in the registry
--- a/features/checkout/skipping_shipping_step_when_only_one_shipping_method_is_available.feature
+++ b/features/checkout/skipping_shipping_step_when_only_one_shipping_method_is_available.feature
@@ -7,19 +7,22 @@ Feature: Skipping shipping step when only one shipping method is available
     Background:
         Given the store operates on a single channel in "United States"
         And the store has a product "Guards! Guards!" priced at "$20.00"
+        And the store allows paying with "Paypal Express Checkout"
         And I am a logged in customer
 
-    @todo
+    @ui
     Scenario: Seeing checkout payment page after addressing if only one shipping method is available
         Given the store has "DHL" shipping method with "$5.00" fee
         And I have product "Guards! Guards!" in the cart
         And I am at the checkout addressing step
         When I specify the shipping address as "Ankh Morpork", "Frost Alley", "90210", "United States" for "Jon Snow"
         And I complete the addressing step
-        Then I should be on the checkout payment step
+        And I select "Paypal Express Checkout" payment method
+        And I complete the payment step
+        Then I should be on the checkout complete step
         And my order's shipping method should be "DHL"
 
-    @todo
+    @ui
     Scenario: Seeing checkout payment page after addressing if only one shipping method is available for current channel
         Given the store has "DHL" shipping method with "$5.00" fee
         Given the store has "FedEx" shipping method with "$15.00" fee not assigned to any channel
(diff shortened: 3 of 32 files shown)


Zales0123:
There is one big issue with this feature - it breaks 90% of our checkout scenarios 😄  The problem is, that in most features background setup consist of only one shipping method (`And the store ships everywhere for free`). Therefore, due to our new logic, shipping step is skipped and all steps that:
- check if current page is shipping step
- select the only available method
- pass whole checkout with some payment and shipping methods

are teared down in a massive explosion 💥 

I see 3 possibilities how we can handle this issue:
1. remove all shipping selection related steps/part of the steps in scenarios that has only one shipping method configured in setup.
2. add second, unused in scenario, shipping method for each scenario (to have two of them and don't allow to skip shipping step if we don't want to)
3. modify `And the store ships everywhere for free` step to create 2 free shipping methods rather than one (the same reason as in 2).

Personally, I think the 3 option would be the best solution. As skipping shipping method is a feature not a default behaviour, our default shop setup should consist of at least **2** shipping methods (as for example we have in the fixtures). The only thing I'm not totally ok with, is creating two shipping methods in step called `And the store ships everywhere for free`. It would be probably better to rename it to something more appropriate.

Please tell me what is the best solution in your opinion ;)

tchapi:
Why would you want to do that ? How is the user going to know what shipping method he's paying for if we just skip the step where he can see all the info about it ?

TheMadeleine:
@tchapi I think you still have the info on the order summary page, where the customer needs to confirm all the order's data with the `Place the order` button. :)

Zales0123:
@tchapi whole feature is just for speeding up a checkout process if it's possible. But we can think about displaying a flash message about skipped shipping step and added costs. I just don't know if it's not redundant - as @TheMadeleine have written, we have a shipping costs information on order summary page.

tchapi:
I can't really think of a web store that does skip the shipping process (_except maybe when it's a kind of downloadable product_). My concern is that the clients are going to land on the summary page with a shipping line and say 'hell, where do I change the shipping method ? I did miss a step, how can I go back ?' and it's going to create frustration again.

Skipping payment when it's free is a totally different process, skipping something that is _not free_ is definitely a bad idea and a user experience anti-pattern I think.

pjedrzejewski:
I agree with @tchapi and this should be optional behavior for stores that sell virtual goods or if they really have a single shipping method and make it obvious for customer. How about we add a setting on Channel entity that defines if this behavior applies or not? We can make it ``false`` by default. WDYT?

psihius:
In my case there can be a single shipping method, but user has to select the shipping destination from a list of places as in Baltics the self-pickup boxes are the prefered shipping method by everyone - I'd say like 90% of small packages is shipped via this method (and it's considerably cheaper).
So just flat-out skipping a shipping method by default is not a good idea. Either make it optional, or (probably a better option), add to the payment method a flag that defines if this shipping method can be skipped at all or requires user interaction.

Zales0123:
Thank you all for the feedback ;) Eventually, I've implemented a flag on `Channel` that @pjedrzejewski proposed, I think it's a good compromise in this case ;) 🐃

src/Sylius/Bundle/CoreBundle/Form/Extension/ChannelTypeExtension.php
```diff
@@ -68,6 +69,10 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'label' => 'sylius.form.channel.contact_email',
                 'required' => false,
             ])
+            ->add('skippingShippingStepAllowed', CheckboxType::class, [
+                'label' => 'sylius.form.channel.skipping_shipping_step_allowed',
```
pjedrzejewski: Bro, do you even allow skipping your shipping? XD

src/Sylius/Bundle/CoreBundle/Templating/Helper/CheckoutStepsHelper.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    /**
+     * @param OrderInterface $order
+     *
+     * @return bool
+     */
+    public function isPaymentRequired(OrderInterface $order)
+    {
+        return 0 < $order->getTotal();
```
pjedrzejewski: Shouldn't we have a service here too?

Zales0123: I was thinking about it too, but I finally decided to not introduce a separate service for such a simple case... However, it would of course be consistent with whole logic, so I can naturally do it.

features/checkout/skipping_shipping_step_when_only_one_shipping_method_is_available.feature
```diff
@@ -0,0 +1,50 @@
...
+        And I complete the addressing step
+        And I select "Paypal Express Checkout" payment method
+        And I complete the payment step
+        Then I should be on the checkout complete step
+        And my order's shipping method should be "DHL"
+
+    @ui
+    Scenario: Seeing checkout payment page after addressing if only one shipping method is available for current channel
+        Given the store has "DHL" shipping method with "$5.00" fee
+        Given the store has "FedEx" shipping method with "$15.00" fee not assigned to any channel
```
GSadee: And

features/checkout/skipping_shipping_step_when_only_one_shipping_method_is_available.feature
```diff
@@ -0,0 +1,50 @@
...
+        And I complete the addressing step
+        And I select "Paypal Express Checkout" payment method
+        And I complete the payment step
+        Then I should be on the checkout complete step
+        And my order's shipping method should be "DHL"
+
+    @ui
+    Scenario: Seeing checkout payment page after addressing if only one shipping method is enabled for current channel
+        Given the store has "DHL" shipping method with "$5.00" fee
+        Given the store has disabled "FedEx" shipping method with "$15.00" fee
```
GSadee: And

app/migrations/Version20170214104908.php
```diff
@@ -0,0 +1,34 @@
...
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_channel ADD skipping_shipping_step_allowed TINYINT(1) NOT NULL');
```
pjedrzejewski: Please set default value as false for existing method, because that's the current behaviour.

features/channel/managing_channels/adding_channel.feature
```diff
@@ -31,6 +31,7 @@ Feature: Adding a new channel
         And I define its color as "blue"
         And I choose "Euro" as the base currency
         And I choose "English (United States)" as a default locale
+        And I allow to skip shipping step if only one shipping method is available
```
michalmarcinkowski: Is it required?

features/checkout/skipping_shipping_step_when_only_one_shipping_method_is_available.feature
```diff
@@ -0,0 +1,50 @@
...
+        And I complete the addressing step
+        And I select "Paypal Express Checkout" payment method
+        And I complete the payment step
+        Then I should be on the checkout complete step
+        And my order's shipping method should be "DHL"
+
+    @ui
+    Scenario: Seeing checkout payment page after addressing if only one shipping method is enabled for current channel
+        Given the store has "DHL" shipping method with "$5.00" fee
+        And the store has disabled "FedEx" shipping method with "$15.00" fee
```
michalmarcinkowski: Can we swap the order of the above steps to ensure we don't simply get the first shipping method from the database?

features/channel/managing_channels/adding_channel.feature
Zales0123: Well, it ensures us that during channel creation we can set this property as true, so I think it provides some measurable value.

michalmarcinkowski: We don't validate if this property is set, but at least check if we can select it, so it's :+1:

pjedrzejewski:
Thanks Mateusz, nice work!

features/channel/managing_channels/adding_channel.feature
Zales0123: Yes, that's what I meant 😄

Issue #7471, opened 2017-02-09
createReviewerMapping: wrong array key


krisElphenor:
When overriding the customer entity mapping, especially the id field. The new value isn't 'seen' by the mapping of the review entity.

For example, let's say, we override the mapping of customer entity and we change 'id' to 'customer_id', we then expect the author_id of the review entity, which references the customer id, to reference the new field 'customer_id'. Unfortunately, that's not the case and author_id still point at customer.id (which doesn't exist anymore) which result in an exception.

The problem seems to be in the file [Sylius/Bundle/ReviewBundle/Doctrine/ORM/Subscriber/LoadMetadataSubscriber.php](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ReviewBundle/Doctrine/ORM/Subscriber/LoadMetadataSubscriber.php)

... in the createReviewerMapping method, in the returned array, I believe the key 'joinColum' should be 'joinColumns' and the expected value should be an array of array, so...

...instead of

```
private function createReviewerMapping($reviewerEntity, ClassMetadata $reviewerEntityMetadata)
    {
        return [
            'fieldName' => 'author',
            'targetEntity' => $reviewerEntity,
            'joinColumn' => [
                'name' => 'author_id',
                'referencedColumnName' => $reviewerEntityMetadata->fieldMappings['id']['columnName'],
                'nullable' => false,
                'onDelete' => 'CASCADE',
            ],
            'cascade' => ['persist'],
        ];
    }
```

...we should have...

```
private function createReviewerMapping($reviewerEntity, ClassMetadata $reviewerEntityMetadata)
    {
        return [
            'fieldName' => 'author',
            'targetEntity' => $reviewerEntity,
            'joinColumns' => [
                [
                    'name' => 'author_id',
                    'referencedColumnName' => $reviewerEntityMetadata->fieldMappings['id']['columnName'],
                    'nullable' => false,
                    'onDelete' => 'CASCADE',
                ]
            ],
            'cascade' => ['persist'],
        ];
    }
```

Doing so, allows the field to be properly overwritten by Doctrine

pjedrzejewski:
@krisElphenor Could you submit a PR with the fix, please? That would be super helpful! :)

krisElphenor:
@pjedrzejewski Yes, of course !

https://github.com/Sylius/Sylius/pull/7490

PR #7490, opened 2017-02-13
createReviewerMapping: impossible override due to wrong array key

krisElphenor:
Please see issue https://github.com/Sylius/Sylius/issues/7471

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  mentioned in #7471 |
| License         | MIT |


src/Sylius/Bundle/ReviewBundle/Doctrine/ORM/Subscriber/LoadMetadataSubscriber.php
```diff
@@ -106,11 +106,13 @@ private function createReviewerMapping($reviewerEntity, ClassMetadata $reviewerE
         return [
             'fieldName' => 'author',
             'targetEntity' => $reviewerEntity,
-            'joinColumn' => [
-                'name' => 'author_id',
-                'referencedColumnName' => $reviewerEntityMetadata->fieldMappings['id']['columnName'],
-                'nullable' => false,
-                'onDelete' => 'CASCADE',
+            'joinColumns' => [
```
lchrusciel: We used to put both arrays openings in the same line (check [LoadMetadataSubscriber in attribute bundle](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/AttributeBundle/Doctrine/ORM/Subscriber/LoadMetadataSubscriber.php#L85-L90)).

krisElphenor: Ok, no worries. But, in that case, to be consistent, we should also update the `createSubjectMapping` method in that same file, as it has the same formatting (see https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ReviewBundle/Doctrine/ORM/Subscriber/LoadMetadataSubscriber.php#L87).

If you wish, I can update this PR to include that changes for both method.

lchrusciel: If you could. Anyway, you need to generate a migration file and update our specs (that's why test suites are failing)

lchrusciel:
And we still need a doctrine migration 
```bash
$ php app/console doctrine:migration:migrate
```

krisElphenor:
I've added the migrations and updated the format and the specs. Sorry for the delay and the multiple commit, I'll try to make it more smooth next time :)

krisElphenor:
There seems to be an issue with the scenario, I'll check it...

krisElphenor:
The problem comes from the condition `And this product has one review` which is satisfied by the method `productHasAReview` defined here: [ProductReviewContext](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Behat/Context/Setup/ProductReviewContext.php#L70)

In this method, the call to `createProductReview` is allowed to set the author of the review as null, which is not allowed by the mapping. 

So, I guess the question is: do we allow anonymous reviews or not ?

src/Sylius/Bundle/ReviewBundle/Doctrine/ORM/Subscriber/LoadMetadataSubscriber.php
```diff
@@ -106,12 +104,12 @@ private function createReviewerMapping($reviewerEntity, ClassMetadata $reviewerE
         return [
             'fieldName' => 'author',
             'targetEntity' => $reviewerEntity,
-            'joinColumn' => [
+            'joinColumns' => [[
                 'name' => 'author_id',
                 'referencedColumnName' => $reviewerEntityMetadata->fieldMappings['id']['columnName'],
-                'nullable' => false,
+                'nullable' => true,
```
pamil: Why?

krisElphenor: Because of my previous comment. 

> In this method, the call to createProductReview is allowed to set the author of the review as null, which is not allowed by the mapping.

So, I've updated the mapping to follow the code expectation, but if a review should always have an author, then this should, indeed, be set to non-nullable and the code need to be updated to reflect that as well.

pamil: /cc @Zales0123 @GSadee

michalmarcinkowski:
@krisElphenor we should not allow for `null` value as author and the step  `And this product has one review` should be renamed to `And this product has one review from "john@doe.com"` so we can set the author.

src/Sylius/Bundle/ReviewBundle/Doctrine/ORM/Subscriber/LoadMetadataSubscriber.php
Zales0123: Yes, review definitely should have an author in all cases, so it should not be nullable.

app/migrations/Version20170215115804.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Sylius\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20170215115804 extends AbstractMigration
```
Zales0123: Could you please merge this two migrations in one?

krisElphenor:
I have done the changes needed. Unfortunately, it seems to fail on 'adding a product with association'. 
I'm quite new to behat, so, i'm sorry for all those little steps and if i'm missing something obvious, but I don't see the link between the associations and reviews. So what do you think I'm missing ?

features/product/managing_products/deleting_product.feature
```diff
@@ -8,7 +8,7 @@ Feature: Deleting a product
         Given the store operates on a single channel in "United States"
         And the store has a product "Toyota GT86 model"
         And this product has "1:43" variant priced at "$15.00"
-        And this product has one review
+        And this product has one review from "john@doe.com"
```
lchrusciel: you have missed `customer`.

And this product has one review from **customer** "john@doe.com"

michalmarcinkowski:
@krisElphenor could you try the fix proposed by @lchrusciel?

features/product/managing_products/deleting_product.feature
krisElphenor: Done ! Thanks for the help @lchrusciel :)

pjedrzejewski:
Thanks @krisElphenor, nice work!

krisElphenor:
No worries, but more importantly, thanks to all of you for the great work you do on Sylius.

PR #7500, opened 2017-02-14
[API] [Checkout] Remove email from address type

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes but only in API |
| Related tickets |  |
| License         | MIT |

I have removed an email field from AddressType, because email is required to create a car anyway. There is no need to send it one more time. I have also refactored the whole checkout test suite. Now, to start checkout the cart is built from the scratch, a new item is added, and then the checkout can proceed.


pjedrzejewski:
Thanks Łukasz!

PR #7504, opened 2017-02-14
[Resource] [DX] Add short message to not found exception

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

Small dx improvement to have at least some message in returned exception.
Before:
```json
{"code":404,"message":""}
```
After:
```json
{"code":404,"message":"The \"order\" has not been found"}
```
or in UI:
zrzut ekranu 2017-02-14 o 11 38 26

Another thing is the translation of this error, but the same problem lays on the every exception in Sylius.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
@@ -463,7 +463,7 @@ class ResourceController extends Controller
     protected function findOr404(RequestConfiguration $configuration)
     {
         if (null === $resource = $this->singleResourceProvider->get($configuration, $this->repository)) {
-            throw new NotFoundHttpException();
+            throw new NotFoundHttpException(sprintf('The "%s" has not been found', $this->metadata->getHumanizedName()));
         }
 
         return $resource;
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
```diff
@@ -463,7 +463,9 @@ protected function isGrantedOr403(RequestConfiguration $configuration, $permissi
     protected function findOr404(RequestConfiguration $configuration)
     {
         if (null === $resource = $this->singleResourceProvider->get($configuration, $this->repository)) {
-            throw new NotFoundHttpException();
+            $message = sprintf('The "%s" has not been found', $this->metadata->getHumanizedName());
+
+            throw new NotFoundHttpException($message);
```
pamil: Why don't we just inject the result of `sprintf` here instead of creating temporary variable just to inject it in the next line? #universityStyle 🌮

pjedrzejewski:
Thanks Łukasz!

PR #7502, opened 2017-02-14
[Documentation][Customization] Non-resource controllers - update

TheMadeleine:
| Q               | A
| --------------- | ---
| Doc fix?        | yes |
| New docs?    | no |
| Related tickets |  |
| License         | MIT |

PR #7506, opened 2017-02-14
[ShopBundle] Make HomepageController final

TheMadeleine:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | related to #7502 |
| License         | MIT |

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -232,6 +232,9 @@ These services will be generated automatically based on subject name.
   * `@SyliusShop/_currencySwitch.html.twig` -> `@SyliusShop/Menu/_currencySwitch.html.twig`
   * `@SyliusShop/_localeSwitch.html.twig` -> `@SyliusShop/Menu/_localeSwitch.html.twig`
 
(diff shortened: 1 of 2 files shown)


TheMadeleine:
Why @pamil? Many ppl extend it in order to add stuff to Homepage :)

pamil:
Let's force them to make their own controller, so we won't introduce future BC breaks eg. by adding an extra argument to this controller.

If they are beginners it's even better for them to extend Symfony's `Controller` instead of fighting with injecting dependencies. At worst they would copy 20 lines of code from this one and customise it.

pamil:
They still need to register their custom controller as a service with the same id as extended controller anyway (https://github.com/Sylius/Sylius/pull/7502#discussion_r101021992).

TheMadeleine:
@pamil, done :)

pamil:
Don't want to be annoying, but the commit message is quite outdated now :D

TheMadeleine:
:+1: @pamil 🎉

pjedrzejewski:
Thank you Magda!

PR #7520, opened 2017-02-15
[Grid] Add dispatching events in grid definition converter

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

### DIFF
--- a/docs/customization/grid.rst
+++ b/docs/customization/grid.rst
@@ -127,6 +127,70 @@ See an example of fields order modification on the `sylius_admin_product_review`
                     author:
                         position: 4
 
+Customize grids by events
+-------------------------
+
+There is also another way to customize grids: **via events**.
+Every grid configuration dispatches an event when its definition is being converted.
+
+For example, **sylius_admin_product** grid dispatches such an event:
+
+.. code-block:: php
+
+    sylius.grid.admin_product # For the grid of products in admin
+
+To show you an example of a grid customization using events, we will remove a field from a grid using that method.
+Here are the steps, that you need to take:
+
+**1.** In order to remove fields from the product grid in **Sylius** you have to create a ``AppBundle\Grid\AdminProductsGridListener`` class.
+
+In the example below we are removing the ``images`` field from the ``sylius_admin_product`` grid.
+
+.. code-block:: php
+
+    <?php
+
+    namespace AppBundle\Grid;
+
+    use Sylius\Bundle\GridBundle\Event\GridDefinitionConverterEvent;
+
+    final class AdminProductsGridListener
+    {
+        /**
+         * @param GridDefinitionConverterEvent $event
+         */
+        public function removeImageField(GridDefinitionConverterEvent $event)
+        {
+            $grid = $event->getGrid();
+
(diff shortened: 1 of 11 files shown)


pjedrzejewski:
I am not sure it should be in the converter. I had a quick look and there is no obvious place, but we should add it to the place where grid definition is taken from the provider. Otherwise it works only for ArrayGridProvider. WDYT?

docs/customization/grid.rst
```diff
@@ -127,6 +127,67 @@ See an example of fields order modification on the `sylius_admin_product_review`
                     author:
                         position: 4
 
+Customize grids by events
+-------------------------
+
+There is also a second way to customize grids: **via events**.
```
TheMadeleine: `There is also another way`

docs/customization/grid.rst
```diff
@@ -127,6 +127,67 @@ See an example of fields order modification on the `sylius_admin_product_review`
...
+Every grid configuration dispatches an event when its definiton is converting.
```
TheMadeleine: `when its definition is being converted.`

docs/customization/grid.rst
```diff
@@ -127,6 +127,67 @@ See an example of fields order modification on the `sylius_admin_product_review`
...
+
+For example, **sylius_admin_product** grid dispatches an event:
```
TheMadeleine: `dispatches such an event:`

docs/customization/grid.rst
```diff
@@ -127,6 +127,67 @@ See an example of fields order modification on the `sylius_admin_product_review`
...
+
+.. code-block:: php
+
+    sylius.grid.admin_product # For the grid of products in admin
+
+**1.** In order to remove fields from the product grid in **Sylius** you have to create a ``AppBundle\Grid\AdminProductsGridListener`` class.
```
TheMadeleine: Before presenting the tutorial we should write what are we going to do:

`To show you and example of a grid customization using events, 
we will remove a field from a grid using that method. Here are the steps, that you need to take:`

docs/customization/grid.rst
```diff
@@ -127,6 +127,67 @@ See an example of fields order modification on the `sylius_admin_product_review`
...
+
+In the example below we are removing an **images** field from ``sylius_admin_product`` grid.
```
TheMadeleine: ```the ``images`` field from the ``sylius_admin_product`` grid. ```

docs/customization/grid.rst
```diff
@@ -127,6 +127,67 @@ See an example of fields order modification on the `sylius_admin_product_review`
...
+
+.. code-block:: yaml
+
+    # app/config/config.yml
+    imports:
+        - { resource: "services.yml" }
+
+**3.** Result:
+
+After these two steps your admin product grid should not have image field.
```
TheMadeleine: ``` the ``image`` field. ```

pjedrzejewski:
Thanks Grzesiu!

PR #7523, opened 2017-02-15
Make `ProductTaxon` a proxy for `Taxon`.

steffenbrem:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | none |
| License         | MIT |

I proposed this setup in Slack and @pjedrzejewski agreed on letting me make a PR. Themes and templates could look a lot cleaner when using these patterns.

With `ProductTaxon` being a proxy of `Taxon` we can do this:

```twig
{# proposal: we change `productTaxons` to `taxons` #}
{% for taxon in product.taxons %}
    {{taxon.name}} {# will get it from the `taxon` internally #}
    {{taxon.position}} {# we still work with a ProductTaxon, so this works :D #}
{% endfor %}
```

I would like to know if it would be preferred to change `Product::productTaxons` to `Product::taxons` if we would introduce this idea. To me it feels a lot better to use just `taxons`, since you basically have a `Taxon`.

### DIFF
--- a/src/Sylius/Component/Core/Model/ProductTaxon.php
+++ b/src/Sylius/Component/Core/Model/ProductTaxon.php
@@ -11,10 +11,13 @@
 
 namespace Sylius\Component\Core\Model;
 
+use Sylius\Component\Resource\Model\TranslationInterface;
+use Sylius\Component\Taxonomy\Model\TaxonInterface as BaseTaxonInterface;
+
 /**
  * @author Anna Walasek <anna.walasek@lakion.com>
  */
-class ProductTaxon implements ProductTaxonInterface
+class ProductTaxon implements ProductTaxonInterface, TaxonInterface
 {
     /**
      * @var mixed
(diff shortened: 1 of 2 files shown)


okwinza:
IMO along with this approach we should also add a method to assign product to a taxon in one go.
Then ProductTaxon will be (almost) pure internal and will not affect DX.

tchapi:
If you can make this backward compatible, I'm all for it ;)

steffenbrem:
@okwinza It is indeed a nice way to hide the product taxon all together this way. But I wonder how many changes are allowed to be made in beta?

@tchapi If needed we could keep `Product::getProductTaxons`, which will call `Product::getTaxons`. Nothing would break from this.

okwinza:
We should decide if we want to make `ProductTaxon` concept as internal as possible, otherwise it would be a huge headache for newcomers. Even i had few WTF moments when i first discover it.

In order to keep things simple, we have to encapsulate everything related to `ProductTaxon` while exposing only Taxon-related stuff. This means introducing some minor BC breaks as we have to rename `Product::getProductTaxons` to `Product::getTaxons` and so on.

I'll open a PR soon to illustrate this.

steffenbrem:
@pjedrzejewski Could you give me your thoughts on how deep you want to go with this? The things @okwinza mentioned are good points and I think it is a good idea to go that way.

steffenbrem:
I think we should implement the proxies first while keeping everything as it is. So we can later iteratively remove certain internal classes, such as the `ProductTaxon`. Any pointers on how to proceed on this would be great 😄 ! Shall I leave this PR as it is, nice and small. Or shall I expand further and try to hide the whole `ProductTaxon` away from end-developers.

pjedrzejewski:
I am not sure about ProductTaxon implementing TaxonInterface, this would make ``$product->setMainTaxon($productTaxonInstance);`` work, but then Doctrine would complain, so it can lead to some confusion. How about we simply add the getters that are useful, code, name, images. Just like we do with ProductOptionValue & ProductAttributeValue. This makes templates cleaner, but does not confuse interfaces. WDYT?

pamil:
:+1: for @pjedrzejewski, we can copy most of the `TaxonInterface` methods definitions, but `ProductTaxonInterface` should never extend `TaxonInterface`.

pjedrzejewski:
@steffenbrem Would you find some time to update the PR if you agree? :)

pamil:
`$product->getTaxons()` returning all taxons looks like the best way to solve it with minimal amount of code, wdyt?

PR #7525, opened 2017-02-15
[Shop] [Checkout] Rework locale assigner listener

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

It appears that we have made our checkout stateful. This PR fix it and moves a logic that depends on context to the listener which listen on shop event.


pjedrzejewski:
Thanks Łukasz!

PR #7534, opened 2017-02-16
[RFC] [Core] [Shop] [Admin] Move listeners to the proper bundle

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |

I have noticed that some of the listeners listen to the custom events from Admin or Shop. Probably it will be cleaner to place them in the appropriate bundle. It is just a proposal, so I'm open for votes should we move them or not. If yes, I will update an UPGRADE file as well.


pjedrzejewski:
Thank you Łukasz!

PR #7547, opened 2017-02-17
[RFC][WIP] Expose product option information in API

alcaeus:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | yes |
| License         | MIT |

The serializer configuration for product options currently does not correctly expose values: options only expose their codes, no names, no values. The `name` property is no longer used, instead this should be a virtual `getName`. The same goes for option values, affecting the `value` attribute.

Unfortunately, exposing option values triggers an error when creating a new option. The `ProductOptionValueType` creates a new `ProductOptionValue` object through the forms `empty_data` closure. However, the resulting object has no locale information, so when the result of the create is serialized, the following error occurs:
> No locale has been set and current locale is undefined.

Obviously, the forms should create new objects not through the default `empty_data` function, but rather through the respective resource factory. Since such a change affects all form classes and might be considered a BC break (it involves adding a required `factory` argument to `AbstractResourceType::__construct`), I wanted to get some input on the issue before going out and changing all forms and services. What do you think?

### DIFF
--- a/src/Sylius/Bundle/ProductBundle/Resources/config/serializer/Model.ProductOption.yml
+++ b/src/Sylius/Bundle/ProductBundle/Resources/config/serializer/Model.ProductOption.yml
@@ -3,8 +3,10 @@ Sylius\Component\Product\Model\ProductOption:
     xml_root_name: option
     properties:
         code:
-            exopse: true
-            type: string
-        name:
             expose: true
             type: string
+        values:
+            expose: true
+    virtual_properties:
+        getName:
+            serialized_name: name
--- a/src/Sylius/Bundle/ProductBundle/Resources/config/serializer/Model.ProductOptionValue.yml
+++ b/src/Sylius/Bundle/ProductBundle/Resources/config/serializer/Model.ProductOptionValue.yml
@@ -5,9 +5,8 @@ Sylius\Component\Product\Model\ProductOptionValue:
         code:
             expose: true
             type: string
-        value:
-            expose: true
-            type: string
     virtual_properties:
         getName:
             serialized_name: name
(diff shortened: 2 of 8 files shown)


src/Sylius/Bundle/ProductBundle/Resources/config/serializer/Model.ProductOption.yml
```diff
@@ -3,8 +3,10 @@ Sylius\Component\Product\Model\ProductOption:
...
         code:
-            exopse: true
-            type: string
-        name:
             expose: true
             type: string
+        values:
+            expose: true
+    virtual_properties:
+        getName:
```
lchrusciel: Instead of fetching the name with virtual property, we should expose all product option translations

src/Sylius/Bundle/ProductBundle/Resources/config/serializer/Model.ProductOptionValue.yml
```diff
@@ -5,9 +5,8 @@ Sylius\Component\Product\Model\ProductOptionValue:
         code:
             expose: true
             type: string
-        value:
-            expose: true
-            type: string
     virtual_properties:
         getName:
             serialized_name: name
+        getValue:
```
lchrusciel: Same as above. Both virtual properties should be removed and we should expose all translations

alcaeus:
Thanks for the quick reply, Łukasz. I'll hold off until #7482 is merged and then rebase and continue working on this.

alcaeus:
Superseded by #7811.

PR #7569, opened 2017-02-21
Add Yarn lock file

### LINKED ISSUE #7565: no style and js error after a new install of sylius
I've followed the installation guide from [https://github.com/Sylius/Sylius#installation](https://github.com/Sylius/Sylius#installation). Sylius run by using `server:run`

When i go to the homepage, i get a page with no style (style are loaded in 200).
I got javascript error in this page 

```
Uncaught TypeError: $(...).first(...).sidebar is not a function
    at HTMLDocument.<anonymous> (app.js:528)
    at i (app.js:2)
    at Object.fireWith [as resolveWith] (app.js:2)
    at Function.ready (app.js:2)
    at HTMLDocument.J (app.js:2)
```

I tried using the latest beta version and the dev-master version

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7565 |
| License         | MIT |

`yarn.lock` is just like `composer.lock` 🌮 

TODO:
 - [x] Update documentation and readmes to use Yarn instead of NPM


pamil:
We should consider moving from NPM to Yarn, it generates the lock file on every packages changes like Composer.

https://www.sitepoint.com/yarn-vs-npm/

venyii:
Before adding this horrible lock file, please replace it with Yarn. We started to use shrinkwrap and had multiple problems and many inconsistencies between different users. I'd die a little inside if you used this. Bad memories.

👍  for Yarn, we're using this for a while now and while it has its problems sometimes (it's still very young), it's working much better and much faster.

gruberro:
Full agreement with @venyii. Shrinkwrap caused a lot of issues and workarounds we had to deal with. Especially updating an existing package was terrible!

pamil:
@venyii @gruberro thanks for feedback, felt quite the same while I was adding that shrinkwrap, Yarn's lock management seems much more similar to Composer. Updated PR to use Yarn then.

pjedrzejewski:
As Yarn has been suggested multiple times and we constantly have trust issues with NPM, I think that's the way to go! @pamil Can you update all references to npm?

pamil:
@pjedrzejewski done! :)

michalmarcinkowski:
@pamil the build is failing, looks like an issue on travis, please have a look.

pamil:
No idea how to get Yarn working on Travis, help would be highly appreciated! :)

etc/travis/suites/application/before_script.sh
```diff
@@ -9,4 +9,4 @@ run_command "bin/console doctrine:migrations:migrate --no-interaction --env=test
 
 print_header "Setting the web assets up" "sylius"
 run_command "bin/console assets:install --env=test_cached --no-debug -vvv" || exit $?
-run_command "npm run gulp" || exit $?
+run_command "npm run yarn run gulp" || exit $?
```
venyii: ~~npm run~~ yarn run gulp

etc/travis/suites/application/install.sh
```diff
@@ -4,7 +4,7 @@ source "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/../../../bash/common.lib.s
 
 print_header "Installing dependencies" "Sylius"
 run_command "composer install --no-interaction --no-scripts --prefer-dist" || exit $?
-run_command "npm install" || exit $?
+run_command "npm run yarn install" || exit $?
```
venyii: ~~npm run~~ yarn install

etc/travis/suites/application/before_script.sh
pamil: There's no global `yarn` command available on Travis.

venyii: Hmm I'm not sure, what if you install it locally (not `-g`) and call `./node_modules/.bin/yarn`?

Also, did you see [this](https://github.com/travis-ci/travis-ci/issues/6720#issuecomment-254988120)?

venyii: Don't know what exactly Travis does, but node, while available in `before_install`, is missing in `install` for example.

Sourcing `nvm.sh` works but can't be a viable solution...
https://github.com/venyii/Sylius/commit/0140c2e37e34fa392e9996676bf00117c0821001
https://travis-ci.org/venyii/Sylius/jobs/207348842

pamil: Wow thanks! I'll give it a try :)

pamil:
I give up, setting Yarn on Travis is too complicated -.- @venyii if you have some spare time and willingness feel free to take it over, I'd be very grateful :)

venyii:
Sure, I'll have a look :)

PR #7572, opened 2017-02-21
Require Semantic UI CSS maximum at 2.2.7 because min JS missing

### LINKED ISSUE #7571: Semantic.min.js missing from semantic-ui-css:2.2.8
Today we ran into some issues with the Semantic JS minified file not loading and therefore breaking JS functionality on the basic shop. It seems that in the last release of Semantic UI CSS ('css only' split) (2.2.8) the minified file got left out, which is breaking things:

https://github.com/Semantic-Org/Semantic-UI-CSS/commit/b6e97c923ab35886b93af0b2d65f93aea3bcc57f

Either we can limit it currently at version 2.2.7, because in there the minified file seems available, or we should include the non-minified version & minify it ourselves.

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7571 (temporary fix I think) |
| License         | MIT |

Semantic UI CSS minified JS file is missing in version 2.2.8, which is required by the ShopBundle (Gulpfile.js). This PR fixes it temporary by requiring maximum 2.2.7, until the issue might be fixed on Semantic UI CSS side.

Another option could be to load the JS version if the min.js version is missing and minify it ourselves. 

I think we should keep an eye on https://github.com/Semantic-Org/Semantic-UI-CSS/issues/29 how it will be solved on their side and adjust based on this. For now this fix will make sure people installing Sylius from master get atleast a working version out of the box.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "dependencies": {
     "jquery": "^2.2.0",
     "lightbox2": "^2.9.0",
-    "semantic-ui-css": "^2.2.0"
+    "semantic-ui-css": "<=2.2.7"
   },
   "devDependencies": {
     "gulp": "^3.9.0",


kurakin-oleksandr:
I vote to use non-minified version of css and js

mbabker:
FWIW 2.2.9 is posted on the repo now.

pjedrzejewski:
@stefandoorn This is no longer needed, right?

stefandoorn:
Indeed. Works again!

Issue #7314, opened 2017-01-23
Can't add product to cart in a new channel created


c-revillo:
Here are the steps to reproduce this issue:
 - Install sylius-master as usual
 - Create a new channel (i set the code for it to "new_channel")
 - Add a product to this channel 
 - Go to your new channel and try to edit that product to your cart. 

A exception happens in this case with the following error
```
Notice: Undefined index: new_cahnnel

500 Internal Server Error - ContextErrorException
```

The error happens in the Shipping FlatRate Calculator
```
in vendor/sylius/sylius/src/Sylius/Component/Core/Shipping/Calculator/FlatRateCalculator.php at line 33   -
        $channelCode = $subject->getOrder()->getChannel()->getCode();
        return (int) $configuration[$channelCode]['amount'];
    }
```

I exposed this in the #support slack channel and we concluded that we could return `0`  if that index is not set. 

My worry here is that you can maybe forget to add a proper flat rate and this could lead to unwanted behaviour like always having `0` as rate...

lchrusciel:
We need to add constraint to shipping method, that will not allow us to turn it on if price for given channel is not defined. Similar to product variant. @c-revillo will you try to fix it?

c-revillo:
i'll try, but anyway, shouldn't we check also that "a least a shipping method" needs to be enabled for that channel?

pjedrzejewski:
This is my guess, but problem may be in the DefaultShippingMethodProvider/Resolver. If you create a new channel, you need to fill in the shipping rate before enabling that shipping method for given channel. So in theory, it is impossible to get this kind of configuration. I tried it and can't save shipping method without filling the cost for newly added channel.

c-revillo:
that's interesting because.. why would i need to add a price for a shipping method in a channel if i don't plan to use that shipping method in that channel?

I mean, by default you have 3 shipping methods, right? you add a channel, and you have not even edited any shipping method to make it available for that new channel... so... the check would need to be something like `at least one shipping method needs to have the amount configured for this channel`?

marmotz:
In fact, it's a bug in DefaultShippingMethodResolver::getDefaultShippingMethod():

Return the first enabled shipping method does not work if the first shipping method was not in current channel:
error in ajax/cart/add "{code: 500, message: "Notice: Undefined index: xxx"} caused by an undefined index error in Sylius\Component\Core\Shipping\Calculator\FlatRateCalculator:calculate() line 33

[...]

PR #7576, opened 2017-02-22
[Core][Shipping] Add missing default shipping method resolver service definition

### LINKED ISSUE #7231: Can't add a product to cart after removing shipping step and creating an other channel
after removing the shipping step ,i've created an other channel with no shipping methods ,  with code 'tn_store' and i've added a product to it , and when i try to add the product to the cart and exception appears in the ajax call :
the url : web/app_dev.php/ajax/cart/add?productId=62
the exeption : Notice: Undefined index: tn_store / 500 Internal Server Error - ContextErrorException
stack trace : 
capture

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7231 and #7314 |
| License         | MIT |

Quick fix for a problem with adding a product to a new channel without a shipping method.


pjedrzejewski:
Thanks Grzesiu!

PR #7578, opened 2017-02-22
[Core] Fix default shipping method resolver

michalmarcinkowski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

The service from `Core` was not used and even not registered causing invalid shipping method to be assigned to a cart (even not enabled for current channel).


michalmarcinkowski:
Fixed it offline on the way and now realized that it's already merged https://github.com/Sylius/Sylius/pull/7576 😄

PR #7587, opened 2017-02-23
Manage variant context relies on implementation instead of interface

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

I'm injecting a different implementation as the default resolver in my project. I think it should rely on an interface instead of a concrete class.


pjedrzejewski:
Thanks Stefan!

PR #7592, opened 2017-02-23
Decrease number of queries made when accessing not-joined translation's fields

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | - |
| License         | MIT |

`PersistentCollection::get($key)` does not store its result, so the following code makes three queries:

```php
$product = $productRepository->find(42);
$product->getName(); // $product->getTranslation()->getName()
$product->getSlug(); // $product->getTranslation()->getSlug()
```


pjedrzejewski:
Thanks Kamil!

PR #7627, opened 2017-02-27
[Ui][Admin] Put grid filters on index pages in accordion

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

[EDIT]
zrzut ekranu 2017-03-01 o 14 31 31

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/views/Grid/_default.html.twig
+++ b/src/Sylius/Bundle/UiBundle/Resources/views/Grid/_default.html.twig
@@ -9,21 +9,28 @@
 {% set path = path(app.request.attributes.get('_route'), app.request.attributes.get('_route_params')) %}
 
 {% if definition.filters|length > 0 %}
-    <div class="ui segment">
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
This actually does it for all filters. I am wondering if this is useful for all cases, sometimes having the search field already visible is better UX than none. Same here, extra filters should be hidden, but basic search (by customer/number) should be visible. Akeneo for example: Has configurable filters = you can check/uncheck filters you prefer. I'd hold this change for now, a small tweak to the filters UI should not cause to many headaches during the beta2 phase.

GSadee:
I understand and IMO, for now, we can add this feature with active accordion for filters. It could be quite useful, because users could hide filters - still better than it is now. And of course, in the near future we will think about some minor features to make filters configurable. What do you think @pjedrzejewski ?

pjedrzejewski:
Thanks Grzesiu!

PR #7644, opened 2017-03-01
[Admin] [UI] Conditional link action

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

I have added a condition to `Manage coupons` action on promotion index. Otherwise, clicking the `Create` button results with `HTTP 500` error if promotion was not coupon based. 

Before:
zrzut ekranu 2017-03-01 o 09 46 31
After:
zrzut ekranu 2017-03-01 o 09 46 45

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/views/Grid/Action/links.html.twig
+++ b/src/Sylius/Bundle/UiBundle/Resources/views/Grid/Action/links.html.twig
@@ -1,16 +1,20 @@
-<div class="ui{% if options.icon is defined %} labeled icon{% endif %} floating dropdown {{ options.class|default('') }} link button">
-    {% if options.icon is defined %}<i class="{{ options.icon }} icon"></i>{% endif %}
-    <span class="text">{{ action.label|trans }}</span>
-    <div class="menu">
-        {% if options.header is defined %}
-        <div class="header">
-            {% if options.header.icon is defined %}<i class="{{ options.header.icon }} icon"></i>{% endif %}
-            {{ options.header.label|trans }}
+{% set visible = options.visible is defined ? options.visible : true %}
+
+{% if visible %}
+    <div class="ui{% if options.icon is defined %} labeled icon{% endif %} floating dropdown {{ options.class|default('') }} link button">
+        {% if options.icon is defined %}<i class="{{ options.icon }} icon"></i>{% endif %}
+        <span class="text">{{ action.label|trans }}</span>
+        <div class="menu">
+            {% if options.header is defined %}
+            <div class="header">
+                {% if options.header.icon is defined %}<i class="{{ options.header.icon }} icon"></i>{% endif %}
+                {{ options.header.label|trans }}
+            </div>
+            <div class="divider"></div>
+            {% endif %}
+            {% for link in options.links %}
+                {% include '@SyliusUi/Grid/Action/_link.html.twig' %}
+            {% endfor %}
         </div>
-        <div class="divider"></div>
-        {% endif %}
-        {% for link in options.links %}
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/UiBundle/Resources/views/Grid/Action/links.html.twig
```diff
@@ -1,16 +1,18 @@
...
+            {% endfor %}
         </div>
-        <div class="divider"></div>
-        {% endif %}
-        {% for link in options.links %}
-            {% include '@SyliusUi/Grid/Action/_link.html.twig' %}
-        {% endfor %}
     </div>
-</div>
+{% endif %}
```
tuka217: Missing blank line

src/Sylius/Bundle/UiBundle/Resources/views/Grid/Action/links.html.twig
```diff
@@ -1,16 +1,18 @@
-<div class="ui{% if options.icon is defined %} labeled icon{% endif %} floating dropdown {{ options.class|default('') }} link button">
-    {% if options.icon is defined %}<i class="{{ options.icon }} icon"></i>{% endif %}
-    <span class="text">{{ action.label|trans }}</span>
-    <div class="menu">
-        {% if options.header is defined %}
-        <div class="header">
-            {% if options.header.icon is defined %}<i class="{{ options.header.icon }} icon"></i>{% endif %}
-            {{ options.header.label|trans }}
+{% if options.visible is not defined or options.visible %}
```
pamil: `{% if options.visible|default(true) %}`

src/Sylius/Bundle/UiBundle/Resources/views/Grid/Action/links.html.twig
```diff
@@ -1,16 +1,18 @@
-<div class="ui{% if options.icon is defined %} labeled icon{% endif %} floating dropdown {{ options.class|default('') }} link button">
-    {% if options.icon is defined %}<i class="{{ options.icon }} icon"></i>{% endif %}
-    <span class="text">{{ action.label|trans }}</span>
-    <div class="menu">
-        {% if options.header is defined %}
-        <div class="header">
-            {% if options.header.icon is defined %}<i class="{{ options.header.icon }} icon"></i>{% endif %}
-            {{ options.header.label|trans }}
+{% if options.visible|default(true) %}
```
GSadee: I guess that we don't support `visible` property in options, instead of that we have `enabled` field on action object. So IMO you should remove this if statement.

lchrusciel: It is the feature of `links.html.twig` only, and we have declared this variable already [here](https://github.com/Sylius/Sylius/pull/7644/files#diff-49971219e0092c4c2fb3635aa788e80fR58) so I assume we are supporting it. This is also a clue of this PR

GSadee: IMO, you should use `enabled: resource.couponBased` under `coupons:`

GSadee: Nevermind, your solution fixes my earlier mistake 👍

lchrusciel: Anyway, my previous solution didn't fix anything. So 👍 for pointing it

pjedrzejewski:
Thank you Łukasz!

PR #7649, opened 2017-03-01
[Resource] Throw 404 if expression is invalid

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

With current implementation, if expression (for example in grid) returns `null` value it's passed further and results in some nasty errors.

Example for unexisting taxon slug on shop:
zrzut ekranu 2017-03-01 o 14 54 05

With this simple change, we can display pretty 404 page if anything gone bad.

zrzut ekranu 2017-03-01 o 14 51 30

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/routing/partial/order.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/routing/partial/order.yml
@@ -19,5 +19,5 @@ sylius_admin_partial_order_latest_in_channel:
                 method: findLatestInChannel
                 arguments:
                     count: $count
-                    channel: expr:service('sylius.repository.channel').findOneByCode($channelCode)
+                    channel: expr:notFoundOnNull(service('sylius.repository.channel').findOneByCode($channelCode))
             template: $template
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/routing/product_variant.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/routing/product_variant.yml
@@ -26,7 +26,7 @@ sylius_admin_product_variant_create:
             factory:
                 method: createForProduct
                 arguments:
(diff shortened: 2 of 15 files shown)


src/Sylius/Bundle/ResourceBundle/ExpressionLanguage/NotNullResultExpressionLanguage.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * {@inheritdoc}
+     *
+     * @throws NotFoundHttpException if expression returns null value
+     */
+    public function evaluate($expression, $values = array())
+    {
+        $result = parent::evaluate($expression, $values);
+        if (null === $result) {
```
pamil: What if I *do* expect it to be `null`?

pamil: IMHO it should be a function `throwNotFoundExceptionOnNull(value)` which returns value if `value != null` or throws an exception if `value == null`, wdyt?

pjedrzejewski:
@Zales0123 What about the function providers we had a look at?

Zales0123:
@pjedrzejewski I explored the topic a little bit and it's not something we can use in this case. With expression function provider, we can override expression language functions like `service` or `parameter` but they're only responsible for the way of getting service/parameter from the container. Functions calls (as `findOneBySlug(...)` for example) are called in `evaluate` method of `ExpressionLanguage` class, co we must create a new class to assert the result of the whole operation.

Zales0123:
OK, I understand now what you meant, we just need another function that would wrap the result of the whole evaluation 👍  on it ;)

pjedrzejewski:
Thanks Mateusz!

PR #7652, opened 2017-03-01
[Shop][Ui] Div tags are not allowed as a child to h2

sweoggy:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | N/A |
| License         | MIT |

Let's use span instead, identical looks thanks to ".sub.header".


pjedrzejewski:
Thanks Oscar!

PR #7607, opened 2017-02-24
[Admin][Shop] Prevent from leaving page with unsaved form data

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

It's one of the most important UX feature. From now, leaving the page with form that was not submitted will require a confirmation (for now it's just a browser default alert box).

Protected forms:
 - all forms on Admin panel (except filters - I think it's unnecessary for them)
 - **personal info** form on my account page in Shop
 - **address** forms on my account page in Shop

zrzut ekranu 2017-02-24 o 15 06 56
zrzut ekranu 2017-02-24 o 15 08 49

PR #7657, opened 2017-03-02
[Shop][Admin] Revert problematic JS feature

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | introduced in https://github.com/Sylius/Sylius/pull/7607 |
| License         | MIT |

It's a really nice UX feature, but for now it makes more harm than good. It results in frequent fails of js scenarios on Travis and as soon as we have no good idea to fix it, it must be removed to not cause more frustration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,6 @@
 {
   "dependencies": {
     "jquery": "^2.2.0",
-    "jquery.dirtyforms": "^2.0.0",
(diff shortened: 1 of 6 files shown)


pjedrzejewski:
Thanks Mateusz!

PR #7667, opened 2017-03-03
[Documentation][Book] Promotion - Rules and Actions configuration reference

### LINKED ISSUE #7661: Document promotion rule/action configurations
We need:

1. Example of creating a promotion rule/action programmatically.
2. Reference of all configurations available for both rules & actions.

TheMadeleine:
| Q               | A
| --------------- | ---
| Doc fix?        | no |
| New docs?    | yes |
| Related tickets | fixes #7661 |
| License         | MIT |


src/Sylius/Component/Promotion/Model/PromotionRuleInterface.php
```diff
@@ -1,3 +1,4 @@
+
```
tuka217: Should be reverted

pjedrzejewski:
Thanks Magda!

PR #7677, opened 2017-03-04
Update images.rst

pearlwithbanana:
| Q               | A
| --------------- | ---
| Bug fix?        | no/yes |
| New feature?    | no/yes |
| BC breaks?      | no/yes |
| Related tickets | fixes #X, partially #Y, mentioned in #Z |
| License         | MIT |

### DIFF
--- a/docs/cookbook/images.rst
+++ b/docs/cookbook/images.rst
@@ -49,7 +49,8 @@ The ``imagine_filter('name')`` is a twig filter. This is how you would get an im
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


pearlwithbanana:
It's worth mention that the liip_imagine config in CoreBundle is "%kernel.root_dir%/../web/media/image" so in case somebody tries to filter the image with asset([path])|imagine_filter([filter_name]) it won't work.

pjedrzejewski:
Thank you @pearlwithbanana!

PR #7679, opened 2017-03-05
Ordered CountryChoiceType

### LINKED ISSUE #7232: [UX] CountryChoiceType should order countries by name
We are running a multilingual store and users are having issues finding the country in the dropdown list in checkout addressing page.

**English** (Correct order, is ordered that way in the DB)
captura de pantalla 2017-01-11 a las 10 22 17

**Spanish** (Same order as in english but country names are translated and in some cases hard to find)
captura de pantalla 2017-01-11 a las 10 21 44
For example Spain -> España is not with the countries starting with "E" in Spanish

**Hotfix we made**

I was unable to figure it out how the labels are generated but I found that `CountryChoiceType` fetches the countries without any ordering.

I've created an extensions with the following code and works properly (for sure is not the best way to fix it in the repo but may be useful for people until this is fixed)

```php
namespace AppBundle\Form\Extension\Addressing;

use Sylius\Bundle\AddressingBundle\Form\Type\CountryCodeChoiceType;
use Symfony\Component\Form\AbstractTypeExtension;
use Symfony\Component\Form\FormInterface;
use Symfony\Component\Form\FormView;

class OrderedCountryCodeChoiceTypeExtension extends AbstractTypeExtension
{
    public function buildView(FormView $view, FormInterface $form, array $options)
    {
        usort($view->vars['choices'], function($a, $b) {
            return $a->label < $b->label ? -1 : 1;
        });
    }

    public function getExtendedType()
    {
        return CountryCodeChoiceType::class;
    }

}
```

gorkalaucirica:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7232 |
| License         | MIT |

As mentioned in #7232, country list was mantaining the order from the DB. When changing between languages it results in bad UX.

I decided to use `usort` but maybe this has to be done at repository level (not sure about this).

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
+++ b/src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
@@ -44,10 +44,16 @@ final class CountryChoiceType extends AbstractType
             ->setDefaults([
                 'choices' => function (Options $options) {
                     if (null === $options['enabled']) {
-                        return $this->countryRepository->findAll();
+                        $countries = $this->countryRepository->findAll();
+                    } else {
+                        $countries = $this->countryRepository->findBy(['enabled' => $options['enabled']]);
                     }
 
-                    return $this->countryRepository->findBy(['enabled' => $options['enabled']]);
+                    usort($countries, function($a, $b) {
+                        return $a->getName() < $b->getName() ? -1 : 1;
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
Thank you Gorka!

PR #7681, opened 2017-03-05
Double check promotions are eligible in PromotionProcessor

### LINKED ISSUE #7584: PromotionProcessor should take into account previously applied promotions
Don't now if this is a bug or expected feature.

**Steps to reproduce**
1. Create two promotions, a 50% discount coupon (priority 1) and a free shipping (priority 0) only for orders over 60€
2. Go to shop and add two items with a total amount of 70€
3. Apply the coupon to get 50% discount
4. See the order total is 35€
5. See that you get free shipping (customer should be charged shipping)

**The issue**

[PromotionProcessor](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Promotion/Processor/PromotionProcessor.php) is responsible of applying adjustment to the order according to promotion rules. It checks for elegible promotions but it does not apply them until the end. This causes issues with promotions that need to take into account previously applied promotion such use an order total amount reduction.

**Our workaround (for beta1)**
https://gist.github.com/gorkalaucirica/51158691e84439bfa0d154989097617e

gorkalaucirica:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7584 |
| License         | MIT |

As mentioned in #7584:

>[PromotionProcessor](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Promotion/Processor/PromotionProcessor.php) is responsible of applying adjustment to the order according to promotion rules. It checks for elegible promotions but it does not apply them until the end. This causes issues with promotions that need to take into account previously applied promotion such use an order total amount reduction.

**The fix**

We first try to find an exclusive promotion and apply it if exists. In case it doesn't, we iterate and start applying "normal" promotions. This way we always have the correct `$subject` context to validate if promotion is eligible. It this loop we only check for non exclusive promotion as exclusive ones should have been applied before. In cases when a exclusive promotion is eligible after applying a promotion before they have to be ignored.

### DIFF
--- a/src/Sylius/Behat/Context/Setup/PromotionContext.php
+++ b/src/Sylius/Behat/Context/Setup/PromotionContext.php
@@ -680,8 +680,7 @@ final class PromotionContext implements Context
         PromotionInterface $promotion,
         $discount,
         $itemTotal
-    )
-    {
+    ) {
         $channelCode = $this->sharedStorage->get('channel')->getCode();
         $rule = $this->ruleFactory->createItemTotal($channelCode, $itemTotal);
         $action = $this->actionFactory->createShippingPercentageDiscount($discount);
--- a/src/Sylius/Behat/Context/Setup/PromotionContext.php
+++ b/src/Sylius/Behat/Context/Setup/PromotionContext.php
@@ -674,6 +674,30 @@ final class PromotionContext implements Context
     }
 
     /**
+     * @Given /^([^"]+) gives ("[^"]+%") discount on shipping to every order over ("(?:€|£|\$)[^"]+")$/
+     */
+    public function itGivesDiscountOnShippingToEveryOrderOver(
+        PromotionInterface $promotion,
+        $discount,
+        $itemTotal
+    )
+    {
+        $channelCode = $this->sharedStorage->get('channel')->getCode();
+        $rule = $this->ruleFactory->createItemTotal($channelCode, $itemTotal);
+        $action = $this->actionFactory->createShippingPercentageDiscount($discount);
+
+        $this->persistPromotion($promotion, $action, [], $rule);
+    }
+
+    /**
+     * @Given /^([^"]+) gives free shipping to every order over ("(?:€|£|\$)[^"]+")$/
+     */
+    public function itGivesFreeShippingToEveryOrderOver(PromotionInterface $promotion, $itemTotal)
+    {
+        $this->itGivesDiscountOnShippingToEveryOrderOver($promotion, 1, $itemTotal);
+    }
+
+    /**
      * @param array $taxonCodes
      *
      * @return array
--- a/src/Sylius/Component/Promotion/Processor/PromotionProcessor.php
+++ b/src/Sylius/Component/Promotion/Processor/PromotionProcessor.php
@@ -60,9 +60,9 @@ final class PromotionProcessor implements PromotionProcessorInterface
             $this->promotionApplicator->revert($subject, $promotion);
         }
 
-        $eligiblePromotions = [];
+        $preQualifiedPromotions = $this->preQualifiedPromotionsProvider->getPromotions($subject);
 
-        foreach ($this->preQualifiedPromotionsProvider->getPromotions($subject) as $promotion) {
+        foreach ($preQualifiedPromotions as $promotion) {
             if (!$this->promotionEligibilityChecker->isEligible($subject, $promotion)) {
                 continue;
             }
@@ -72,12 +72,12 @@ final class PromotionProcessor implements PromotionProcessorInterface
 
                 return;
             }
-
-            $eligiblePromotions[] = $promotion;
         }
 
-        foreach ($eligiblePromotions as $promotion) {
-            $this->promotionApplicator->apply($subject, $promotion);
+        foreach ($preQualifiedPromotions as $promotion) {
+            if($this->promotionEligibilityChecker->isEligible($subject, $promotion) && !$promotion->isExclusive()) {
+                $this->promotionApplicator->apply($subject, $promotion);
+            }
         }
     }
 }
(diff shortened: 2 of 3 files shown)


src/Sylius/Component/Promotion/Processor/PromotionProcessor.php
```diff
@@ -60,9 +60,9 @@ public function process(PromotionSubjectInterface $subject)
             $this->promotionApplicator->revert($subject, $promotion);
         }
 
-        $eligiblePromotions = [];
+        $preQualifiedPromotions = $this->preQualifiedPromotionsProvider->getPromotions($subject);
 
-        foreach ($this->preQualifiedPromotionsProvider->getPromotions($subject) as $promotion) {
+        foreach ($preQualifiedPromotions as $promotion) {
             if (!$this->promotionEligibilityChecker->isEligible($subject, $promotion)) {
```
lchrusciel: In this case, this condition could be merged with the condition below

gorkalaucirica: Do you mean merging both for each clauses?

lchrusciel: We just can get rid of `continiue`: 
```php
if ($this->promotionEligibilityChecker->isEligible($subject, $promotion) && $promotion->isExclusive()) {
    $this->promotionApplicator->apply($subject, $promotion);
}
```

gorkalaucirica: I see... 👌

gorkalaucirica: Changed

michalmarcinkowski:
Looks good! @gorkalaucirica are you able to write a Behat scenarios for that?

gorkalaucirica:
@michalmarcinkowski hope I have some time this afternoon.

gorkalaucirica:
Added scenarios, old PromotionProcessor fails, the new one passes

src/Sylius/Component/Promotion/Processor/PromotionProcessor.php
```diff
@@ -60,24 +60,20 @@ public function process(PromotionSubjectInterface $subject)
...
+        $preQualifiedPromotions = $this->preQualifiedPromotionsProvider->getPromotions($subject);
 
-        foreach ($this->preQualifiedPromotionsProvider->getPromotions($subject) as $promotion) {
-            if (!$this->promotionEligibilityChecker->isEligible($subject, $promotion)) {
-                continue;
-            }
-
-            if ($promotion->isExclusive()) {
+        foreach ($preQualifiedPromotions as $promotion) {
+            if ($this->promotionEligibilityChecker->isEligible($subject, $promotion) && $promotion->isExclusive()) {
```
pamil: `$promotion->isExclusive()` costs less than `$this->promotionEligibilityChecker->isEligible($subject, $promotion)` so it should be executed first.

src/Sylius/Component/Promotion/Processor/PromotionProcessor.php
```diff
@@ -60,24 +60,20 @@ public function process(PromotionSubjectInterface $subject)
...
             }
-
-            $eligiblePromotions[] = $promotion;
         }
 
-        foreach ($eligiblePromotions as $promotion) {
-            $this->promotionApplicator->apply($subject, $promotion);
+        foreach ($preQualifiedPromotions as $promotion) {
+            if($this->promotionEligibilityChecker->isEligible($subject, $promotion) && !$promotion->isExclusive()) {
+                $this->promotionApplicator->apply($subject, $promotion);
```
pamil: Same here.

src/Sylius/Component/Promotion/Processor/PromotionProcessor.php
```diff
@@ -60,24 +60,20 @@ public function process(PromotionSubjectInterface $subject)
...
                 return;
             }
-
-            $eligiblePromotions[] = $promotion;
         }
 
-        foreach ($eligiblePromotions as $promotion) {
-            $this->promotionApplicator->apply($subject, $promotion);
+        foreach ($preQualifiedPromotions as $promotion) {
+            if($this->promotionEligibilityChecker->isEligible($subject, $promotion) && !$promotion->isExclusive()) {
```
pamil: CS (space before parentheses)

src/Sylius/Behat/Context/Setup/PromotionContext.php
```diff
@@ -674,6 +674,30 @@ public function thePromotionGivesOffTheOrderForCustomersFromGroup(
     }
 
     /**
+     * @Given /^([^"]+) gives ("[^"]+%") discount on shipping to every order over ("(?:€|£|\$)[^"]+")$/
+     */
+    public function itGivesDiscountOnShippingToEveryOrderOver(
+        PromotionInterface $promotion,
+        $discount,
+        $itemTotal
+    )
```
pamil: `) {` :)

src/Sylius/Behat/Context/Setup/PromotionContext.php
```diff
@@ -674,6 +674,30 @@ public function thePromotionGivesOffTheOrderForCustomersFromGroup(
     }
 
     /**
+     * @Given /^([^"]+) gives ("[^"]+%") discount on shipping to every order over ("(?:€|£|\$)[^"]+")$/
```
pamil: `/^(it) gives gives ("[^"]+%") discount on shipping to every order over ("(?:€|£|\$)[^"]+")$/`, `([^"]+)` looks too general

pamil:
Anyway, Behat is failing due to wrong indentation in feature file, use fours spaces instead of two and it will pass 🎉

src/Sylius/Behat/Context/Setup/PromotionContext.php
gorkalaucirica: Any alternative?

gorkalaucirica:
@pamil fixed

src/Sylius/Behat/Context/Setup/PromotionContext.php
pamil: `([^"]+)` -> `(it)` :)

PR #7437, opened 2017-02-07
Fix taxon repository

Pitoune:
| Q               | A
| --------------- | ---
| Bug fix?        |yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7390 |
| License         | MIT |

PR #7686, opened 2017-03-06
[RFC] Improve tests organisation, part one

### LINKED ISSUE #6484: Make running tests of single packages easier
Running packages tests using `phpspec` and/or `phpunit` could be make easier if we alias it with `composer test`.

Example for ThemeBundle - instead of:

``` sh
$ bin/phpspec run
$ bin/phpunit
```

we would just do:

```
$ composer test
```

This will also simplify logic in `etc/bin/test-package`.

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | mentioned in #7437, fixes #6484 |
| License         | MIT |

After I tried to write integration tests for #7437 I came into quite a mess in out PHPUnit tests organisation.

Changes (only TaxonomyBundle, ThemeBundle, UiBundle, UserBundle for now):
 - **ability to run packages tests with `composer test`**
 - cache / logs directories moved outside the bundle (no more forgetting to remove them)
 - using MySQL database (making sure integration tests work well)
 - removed number of files in tests
 - moved application setup to `Tests/Functional` (like in Symfony tests cases)
 - fixed autoloading (PHPStan works like a charm now :taco:)
 - removed dev dependency on `polishsymfonycommunity/symfony-mocker-container`
 - removed other unused dev dependencies

Let me know if it looks all right, then I can apply these changes for all our bundles.


pjedrzejewski:
Thanks Kamil! Let's do this for other bundles, wdyt?

src/Sylius/Bundle/TaxonomyBundle/Tests/Functional/app/AppKernel.php
```diff
@@ -42,18 +41,22 @@ public function registerBundles()
...
     /**
      * {@inheritdoc}
      */
-    protected function getContainerBaseClass()
+    public function getCacheDir()
     {
-        if ('test' === $this->environment) {
-            return MockerContainer::class;
-        }
+        return sys_get_temp_dir() . '/SyliusTaxonomyBundle/cache/' . $this->getEnvironment();
```
lchrusciel: Won't it be better to use some common folder for that? Something like `/Sylius/TaxonomyBundle/cache/` will be easier to find and maintain. (For example, when somebody will be looking for logs)

pamil:
@pjedrzejewski that's the plan! :)

PR #7688, opened 2017-03-06
[Grid] Add enabled items twig extension

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

This PR fixes a problem with customizing grids, because now, for example when we disable all filters, that there will remain empty filter block above table.

Before:
zrzut ekranu 2017-03-06 o 14 46 35

After:
zrzut ekranu 2017-03-06 o 14 44 44

### DIFF
--- a/src/Sylius/Bundle/GridBundle/spec/Templating/Helper/EnabledItemsHelperSpec.php
+++ b/src/Sylius/Bundle/GridBundle/spec/Templating/Helper/EnabledItemsHelperSpec.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace spec\Sylius\Bundle\GridBundle\Templating\Helper;
+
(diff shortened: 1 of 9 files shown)


pjedrzejewski:
Or perhaps we should have something like... ``getEnabledXs`` method on Grid definition? Sounds a bit nicer than new twig extension. WDYT?

src/Sylius/Bundle/GridBundle/spec/Templating/Helper/EnabledItemsHelperSpec.php
```diff
@@ -0,0 +1,63 @@
...
+    {
+        $this->shouldHaveType(EnabledItemsHelper::class);
+    }
+
+    function it_is_a_symfony_templating_helper()
+    {
+        $this->shouldImplement(HelperInterface::class);
+    }
+
+    function it_extends_a_base_symfony_templating_helper()
```
TheMadeleine: **the**_base_symfony_templating_helper

GSadee:
I don't know why, but twig extension was my first idea 😄  Your solution sounds good, probably even nicer, I'll change that 👍

GSadee:
Closed in favour of #7690

PR #7690, opened 2017-03-06
[Grid] Add methods for getting only enabled fields, actions and filters

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

This PR fixes a problem with customizing grids, because now, for example when we disable all filters, that there will remain empty filter block above table.

Before:
zrzut ekranu 2017-03-06 o 14 46 35

After:
zrzut ekranu 2017-03-06 o 14 44 44


pjedrzejewski:
Thank you Grzesiu!

PR #7697, opened 2017-03-07
Translatable fields are no longer used

gonzalovilaseca:
Translatable fields were removed from resource bundle config, `setTranslatableFields` and `setLocaleProvider` aren't called from anywhere.

| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Doctrine/ODM/MongoDB/TranslatableRepository.php
+++ b/src/Sylius/Bundle/ResourceBundle/Doctrine/ODM/MongoDB/TranslatableRepository.php
@@ -23,36 +23,6 @@ use Sylius\Component\Resource\Repository\TranslatableRepositoryInterface;
 class TranslatableRepository extends DocumentRepository implements TranslatableRepositoryInterface
 {
     /**
-     * @var TranslationLocaleProviderInterface
-     */
-    protected $localeProvider;
-
-    /**
-     * @var array
-     */
-    protected $translatableFields = [];
-
-    /**
-     * {@inheritdoc}
-     */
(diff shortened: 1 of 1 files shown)


pamil:
This means `TranslatableRepository` is not needed at all :)

gonzalovilaseca:
That's what I thought when I started doing this PR, but `applyCriteria` in this repo is slightly different than the one in `DocumentRepository`:
```
        if (null === $criteria) {
            return;
        }

        foreach ($criteria as $property => $value) {
            if (is_array($value)) {
                $queryBuilder
                    ->field($this->getPropertyName($property))->in($value)
                ;
            } elseif ('' !== $value) {
                $queryBuilder
                    ->field($this->getPropertyName($property))->equals($value)
                ;
            }
        }
```
vs
```
        foreach ($criteria as $property => $value) {
            $queryBuilder->field($property)->equals($value);
        }
```

WDYT?

pamil:
In my opinion we should reflect the behaviour from ORM in `DocumentRepository`, being:

```php
if (null === $value) {
    $queryBuilder->andWhere($queryBuilder->expr()->isNull($name));
} elseif (is_array($value)) {
    $queryBuilder->andWhere($queryBuilder->expr()->in($name, $value));
} elseif ('' !== $value) {
    $parameter = str_replace('.', '_', $property);
    $queryBuilder
        ->andWhere($queryBuilder->expr()->eq($name, ':'.$parameter))
        ->setParameter($parameter, $value)
    ;
}
```

gonzalovilaseca:
Ok, will update the PR then.

pjedrzejewski:
Thank you Gonzalo!

PR #7695, opened 2017-03-07
[Promotion] Promotion position should not be null

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7694 |
| License         | MIT |

PR #7698, opened 2017-03-07
[Promotion] Promotion with priority equals to null should be last 

### LINKED ISSUE #7694: [Promotion] Bug with nulling the priority
Steps to reproduce:
```gherkin
    @ui
    Scenario: Trying to remove priority from existing promotion
        Given there is a promotion "Christmas sale"
        And I want to modify this promotion
        When I remove its priority
        And I try to save my changes
        Then I should be notified that priority is required
```
Instead of notification the `HTTP 500 Error` is thrown. 

The question is how should it be fixed. Not null validation on a field? But during creation null is a valid value... Do you have any idea?

tuka217:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no|
| BC breaks?      | no|
| Related tickets | fixes #7694, based on #7695 |
| License         | MIT |


src/Sylius/Bundle/PromotionBundle/Form/EventListener/SetPriorityFormSubscriber.php
```diff
@@ -0,0 +1,49 @@
...
+use Sylius\Component\Promotion\Model\PromotionInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Form\FormEvent;
+use Symfony\Component\Form\FormEvents;
+use Webmozart\Assert\Assert;
+
+/**
+ * @author Anna Walasek <anna.walasek@lakion.com>
+ */
+class SetPriorityFormSubscriber implements EventSubscriberInterface
```
pamil: Shouldn't we inject this behaviour directly to the entity? If position is set to null, use `-1` instead? It isn't anything specific to forms.

src/Sylius/Behat/Context/Ui/Admin/ManagingPromotionsContext.php
```diff
@@ -118,6 +118,14 @@ public function iNameIt($name = null)
     }
 
     /**
+     * @When I remove its priority
+     */
+    public function iSetItsPriority()
```
pamil: iRemoveItsPriority

src/Sylius/Behat/Page/Admin/Promotion/UpdatePageInterface.php
```diff
@@ -19,6 +19,17 @@
  */
 interface UpdatePageInterface extends BaseUpdatePageInterface
 {
+
+    /**
+     * @param int $priority
+     */
+    public function setItsPriority($priority = null);
```
pamil: I'd rather not use default values here, being explicit looks a lot better in this case.

`$page->setItsPriority()` vs `$page->setItsPriority(null)`

pjedrzejewski:
And... why not just making it NotBlank? :)

tuka217:
It was the second option, but when you create an object and not set the position of this object it will be the last one on the list. So, shouldn't the behavior be same when we update the position?

src/Sylius/Behat/Context/Ui/Admin/ManagingPromotionsContext.php
```diff
@@ -625,4 +633,14 @@ private function assertIfFieldIsTrue(PromotionInterface $promotion, $field)
 
         Assert::true($this->updatePage->hasResourceValues([$field => 1]));
     }
+
+    /**
+     * @Given the :promotion promotion should have priority :priority
+     */
+    public function thePromotionsShouldHavePriority($promotion, $priority)
```
pamil: `$promotion` -> `PromotionInterface $promotion`

src/Sylius/Behat/Page/Admin/Promotion/UpdatePage.php
```diff
@@ -27,6 +27,22 @@ class UpdatePage extends BaseUpdatePage implements UpdatePageInterface
     /**
      * {@inheritdoc}
      */
+    public function setItsPriority($priority)
```
pamil: `setItsPriority()` -> `setPriority()`?

src/Sylius/Behat/Page/Admin/Promotion/UpdatePage.php
```diff
@@ -27,6 +27,22 @@ class UpdatePage extends BaseUpdatePage implements UpdatePageInterface
...
+    {
+        $this->getDocument()->fillField('Priority', $priority);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getItsPriority()
```
pamil: `getItsPriority()` -> `getPriority()`?

src/Sylius/Component/Promotion/Model/Promotion.php
```diff
@@ -175,6 +175,10 @@ public function getPriority()
     public function setPriority($priority)
     {
         $this->priority = $priority;
+
+        if(null === $priority) {
```
pamil: CS

pamil: Anyway:

```php
$this->priority = null === $priority ? -1 : $priority;
```

WDYT?

pjedrzejewski:
Thank you Ania!

PR #7701, opened 2017-03-07
[Core] Add missing channel configuration exception

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

### DIFF
--- a/src/Sylius/Component/Core/spec/Exception/ChannelNotDefinedExceptionSpec.php
+++ b/src/Sylius/Component/Core/spec/Exception/ChannelNotDefinedExceptionSpec.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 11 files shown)


src/Sylius/Component/Core/spec/Exception/ChannelNotDefinedExceptionSpec.php
```diff
@@ -0,0 +1,36 @@
...
+use Sylius\Component\Core\Exception\ChannelNotDefinedException;
+
+/**
+ * @author Grzegorz Sadowski <grzegorz.sadowski@lakion.com>
+ */
+final class ChannelNotDefinedExceptionSpec extends ObjectBehavior
+{
+    function let()
+    {
+        $this->beConstructedWith(ChannelNotDefinedException::class);
```
pamil: ?

michalmarcinkowski:
I think we should also validate configuration in admin panel, so that you must specify the amount for channels in which the shipping method is enabled.

GSadee:
@michalmarcinkowski I've checked that and we already have this validation in admin panel, so this is just an additional security.

pjedrzejewski:
Thank you Grzesiu!

PR #7704, opened 2017-03-08
Foundation for using template events

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | - |
| License         | MIT |

See https://sonata-project.org/bundles/block/master/doc/reference/events.html.

It allows to add a template "@Foo/bar.html.twig" for event "foo.bar" just in four lines of code:

```xml
<service id="xdddd" class="Sylius\Bundle\UiBundle\Block\BlockEventListener">
    <argument>@Foo/bar.html.twig</argument>
    <tag name="kernel.event_listener" event="sonata.block.event.foo.bar" method="onBlockEvent" />
</service>
```

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -3,7 +3,6 @@
 
 imports:
     - { resource: "@SyliusCoreBundle/Resources/config/app/config.yml" }
-    - { resource: "@SyliusUiBundle/Resources/config/app/config.yml" }
     - { resource: "@SyliusAdminBundle/Resources/config/app/config.yml" }
     - { resource: "@SyliusShopBundle/Resources/config/app/config.yml" }
     - { resource: "@SyliusApiBundle/Resources/config/app/config.yml" }
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -3,6 +3,7 @@
 
 imports:
     - { resource: "@SyliusCoreBundle/Resources/config/app/config.yml" }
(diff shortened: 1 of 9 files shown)


src/Sylius/Bundle/UiBundle/Resources/config/app/config.yml
```diff
@@ -0,0 +1,5 @@
+# This file is part of the Sylius package.
+# (c) Paweł Jędrzejewsk
```
TheMadeleine: :tada:

pamil: 🎉

pamil: FREE COMMIT!

app/config/config.yml
```diff
@@ -3,6 +3,7 @@
 
 imports:
     - { resource: "@SyliusCoreBundle/Resources/config/app/config.yml" }
+    - { resource: "@SyliusUiBundle/Resources/config/app/config.yml" }
```
pjedrzejewski: Can't we configure it in admin & shop? This is a small but annoying BC break.

pamil: Should we duplicate the configuration?

pjedrzejewski: I think it would be ok to duplicate it.

pjedrzejewski:
Thanks Kamil!

PR #7705, opened 2017-03-08
[Core][Shop][Admin] Move emails

### LINKED ISSUE #7700: Default order confirmation e-mail is coupled with the Shop bundle
Turning off the Shop bundle makes the "create" order transition to fail with an exception:

> An exception has been thrown during the rendering of a template ("Unable to generate a URL for the named route "sylius_shop_order_show" as such route does not exist.") in "SyliusCoreBundle:Email:orderConfirmation.html.twig" at line 6.

I guess Core template shouldn't use the route from the Shop bundle.

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | fixes #7700 |
| License         | MIT |


pjedrzejewski:
@GSadee I think this one deserves entry in UPGRADE, as all e-mail templates are broken by this. :/

UPGRADE.md
```diff
@@ -175,6 +175,31 @@ You can bring back previous configuration by overriding current routing with you
...
+  * `SyliusShopBundle:Email:verification.html.twig`
+
+* The following classes were moved to `AdminBundle`:
+
+  * `Sylius\Bundle\CoreBundle\EmailManager\ShipmentEmailManager`
+  * `Sylius\Bundle\CoreBundle\EmailManager\ShipmentEmailManagerInterface`
+
+* The following email template was moved to `AdminBundle`:
+
+  * `SyliusShopBundle:Email:shipmentConfirmation.html.twig`
```
lchrusciel: We should add something like `please use the related bundle or provide your custom implementation`

pjedrzejewski:
Thanks Grzesiu!

PR #7710, opened 2017-03-08
Make test methods naming consistent

### LINKED ISSUE #4506: [RFC] [DX] Inconsistent method naming for PHPUnit.
Some of our tests have use close to phpspec snake_case convention like [here](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ThemeBundle/Tests/Translation/DependencyInjection/Compiler/TranslatorLoaderProviderPassTest.php) or [here](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Tests/Controller/CustomerApiTest.php), but others use typical phpunit camelCase convention like [here](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Tests/Controller/CountryApiTest.php) or [here](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Tests/Controller/ShippingCategoryApiTest.php). I would prefer to use the first one, but it is not only my call. WDYT? Which tests should be adjusted?

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #4506 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Tests/Command/DebugResourceCommandTest.php
+++ b/src/Sylius/Bundle/ResourceBundle/Tests/Command/DebugResourceCommandTest.php
@@ -40,9 +40,9 @@ final class DebugResourceCommandTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
-     * It should list all resources if no argument is given.
+     * @test
      */
-    public function testListAll()
+    public function it_lists_all_resources_if_no_argument_is_given()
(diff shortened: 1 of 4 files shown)


pjedrzejewski:
👍

PR #7715, opened 2017-03-09
Fixed typo in email template upgrade instructions

kronosnhz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | ~ |
| License         | MIT |

PR #7718, opened 2017-03-09
[Core] Fix UPGRADE file after moving emails

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes after #7705, related with #7715 |
| License         | MIT |

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -175,30 +175,24 @@ You can bring back previous configuration by overriding current routing with you
 
 * `Channel` relation was removed from `ChannelPricing` model. `Channel::$code` should be used instead.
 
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -177,28 +177,28 @@ You can bring back previous configuration by overriding current routing with you
 
 * The following classes were moved to `ShopBundle`:
```
pamil: Should we skip "to `ShopBundle`" part if we got froms and tos below?

pjedrzejewski:
Thanks Grzesiu!

PR #7724, opened 2017-03-10
[Grid] Add possibility to set visible parameter for each link in links action

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

It fixes a problem with displaying a generating variants button for product without options on product list in admin panel.

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/views/Grid/Action/links.html.twig
+++ b/src/Sylius/Bundle/UiBundle/Resources/views/Grid/Action/links.html.twig
@@ -13,7 +13,11 @@
             <div class="divider"></div>
             {% endif %}
             {% for link in options.links %}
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/UiBundle/Resources/views/Grid/Action/links.html.twig
```diff
@@ -13,7 +13,11 @@
             <div class="divider"></div>
             {% endif %}
             {% for link in options.links %}
-                {% include '@SyliusUi/Grid/Action/_link.html.twig' %}
+                {% set isLinkVisible = link.visible is defined ? link.visible : true %}
```
pjedrzejewski: Twig CS.

pjedrzejewski:
Thanks Grzesiu!

PR #7731, opened 2017-03-10
[RFC] [Core] Introduce Plugin concept

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

# Hello, Folks! 👋 
As you probably know, one of the key concepts for Sylius is to introduce an easy way of creating external plugins which will extend the core logic and don't mess with the core itself. It can be useful to somehow differentiate them from regular Symfony bundles. And this is the main topic of this PR. This abstract class will slightly change the default behaviour of SymfonyBundle and will require creating AppPlugin instead of AppBundle. Also, it will change the way, how the config files should be included. 

## Changes
Let's assume that we have created `SylusSomeBundle` as a plugin with routing, grids and templates inside and want to take advantage of this change

### Routing:
Before:
```yml
some_plugin:
    resource: "@SyliusSomeBundle/Resources/config/app/routing.yml"
```
After:
```yml
some_plugin:
    resource: "@SyliusSomePlugin/Resources/config/app/routing.yml"
```
### Config:
Before:
```yml
imports:
    - { resource: "@SyliusSomeBundle/Resources/config/app/grid.yml" }
```
After:
```yml
imports:
    - { resource: "@SyliusSomePlugin/Resources/config/app/grid.yml" }
```
### Templates:
Before:
```yml
form: "@SyliusSome/_form.html.twig"
```
After:
```yml
form: "@SyliusSomePlugin/_form.html.twig"
```
### DI
Convention for DependencyInjection stay the same, so we still have `SyliusSomeExtension`

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Application/SyliusPluginTrait.php
+++ b/src/Sylius/Bundle/CoreBundle/Application/SyliusPluginTrait.php
@@ -13,16 +13,23 @@ namespace Sylius\Bundle\CoreBundle\Application;
 
 use Symfony\Component\DependencyInjection\Container;
 use Symfony\Component\DependencyInjection\Extension\ExtensionInterface;
-use Symfony\Component\HttpKernel\Bundle\Bundle;
 
 /**
+ * @mixin \Symfony\Component\HttpKernel\Bundle\Bundle
+ * @see \Symfony\Component\HttpKernel\Bundle\Bundle
+ *
  * Provides a common logic for Sylius Plugins.
- * Each of a plugins should be created with Plugin instead of Bundle suffix for the root class and inherit from this class.
+ * Each of a plugins should be created with Plugin instead of Bundle suffix for the root class.
  *
  * @author Łukasz Chruściel <lukasz.chrusciel@lakion.com>
  */
-abstract class Plugin extends Bundle
+trait SyliusPluginTrait
 {
+    /**
+     * @var ExtensionInterface|bool
+     */
+    private $containerExtension;
+
     /**
      * Returns the plugin's container extension.
      *
@@ -32,12 +39,12 @@ abstract class Plugin extends Bundle
      */
     public function getContainerExtension()
     {
-        if (null === $this->extension) {
+        if (null === $this->containerExtension) {
             $extension = $this->createContainerExtension();
 
             if (null !== $extension) {
                 if (!$extension instanceof ExtensionInterface) {
-                    throw new \LogicException(sprintf('Extension %s must implement Symfony\Component\DependencyInjection\Extension\ExtensionInterface.', get_class($extension)));
+                    throw new \LogicException(sprintf('Extension %s must implement %s.', get_class($extension), ExtensionInterface::class));
                 }
 
                 // check naming convention for Sylius Plugins
@@ -51,16 +58,37 @@ abstract class Plugin extends Bundle
                     ));
                 }
 
-                $this->extension = $extension;
+                $this->containerExtension = $extension;
             } else {
-                $this->extension = false;
+                $this->containerExtension = false;
             }
         }
 
-        if ($this->extension) {
-            return $this->extension;
+        if ($this->containerExtension) {
+            return $this->containerExtension;
         }
     }
+
+    /**
+     * Creates the bundle's container extension.
+     *
+     * @return ExtensionInterface|null
+     */
+    abstract protected function createContainerExtension();
+
+    /**
+     * Returns the bundle name (the class short name).
+     *
+     * @return string The Bundle name
+     */
+    abstract protected function getName();
+
+    /**
+     * Gets the Bundle namespace.
+     *
+     * @return string The Bundle namespace
+     */
+    abstract protected function getNamespace();
 
     /**
      * Returns the plugin's container extension class.
--- a/src/Sylius/Bundle/CoreBundle/Application/Plugin.php
+++ b/src/Sylius/Bundle/CoreBundle/Application/Plugin.php
@@ -1,0 +1,76 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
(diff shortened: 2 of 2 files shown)


mbabker:
How would this work with something like https://github.com/BabDev/SyliusSupplierBundle/blob/master/src/BabDevSyliusSupplierBundle.php which is extending `AbstractResourceBundle` versus Symfony's `Bundle`?

kurakin-oleksandr:
What are the benefits?

lchrusciel:
@mbabker That's a good point. It wouldn't. But we can also use the Trait and you will be able to use any other parent class. This trait won't make any contract, it just overrides two functions in order to change a regexp.

@kurakin-oleksandr it is mostly a business related profit. It is sometimes hard to explain a concept of `Bundle` to the business. The `Plugin` will sound more familiar to them. Also, a lot of people at our community are not really familiar with `Symfony`. Also, it will be easier to find `Sylius` related extensions. 

Sylius became a pretty big project and each of the Plugins which will use a `whole` SyliusCore as a base could be named `Plugin` to different it from other Bundles, with more Symfony generic logic. 

If we would go this way, what would you like use? An abstract class or trait?

kurakin-oleksandr:
I'm not sure if we should add this just for clear business explanation. You can still name it as "plugin" or "extension" for business now.
Anyway if you want to go forward with this I'd vote for trait.

@lchrusciel I'd like to announce GUI plugin system for Sylius later this month.

pamil:
I'm not sure about that as well, we can still use Symfony nomenclature in code and advertise bundles as plugins / extensions. Having `Sylius[Name]Bundle` convention looks fair enough to me.

src/Sylius/Bundle/CoreBundle/Application/SyliusPluginTrait.php
```diff
@@ -0,0 +1,77 @@
...
+     * Returns the plugin's container extension.
+     *
+     * @return ExtensionInterface|null The container extension
+     *
+     * @throws \LogicException
+     */
+    public function getContainerExtension()
+    {
+        if (null === $this->extension) {
+            $extension = $this->createContainerExtension();
```
pamil: There's no such method here.

src/Sylius/Bundle/CoreBundle/Application/SyliusPluginTrait.php
```diff
@@ -0,0 +1,77 @@
...
+
+            if (null !== $extension) {
+                if (!$extension instanceof ExtensionInterface) {
+                    throw new \LogicException(sprintf('Extension %s must implement Symfony\Component\DependencyInjection\Extension\ExtensionInterface.', get_class($extension)));
```
pamil:
```
throw new \LogicException(sprintf('Extension %s must implement Symfony\Component\DependencyInjection\Extension\ExtensionInterface.', get_class($extension)));

// vs

throw new \LogicException(sprintf('Extension %s must implement %s.', get_class($extension), ExtensionInterface::class));
```

src/Sylius/Bundle/CoreBundle/Application/SyliusPluginTrait.php
```diff
@@ -0,0 +1,77 @@
...
+                }
+
+                // check naming convention for Sylius Plugins
+                $basename = preg_replace('/Plugin$/', '', $this->getName());
```
pamil: There's no such method here (`$this->getName()`).

src/Sylius/Bundle/CoreBundle/Application/SyliusPluginTrait.php
```diff
@@ -0,0 +1,77 @@
...
+                $expectedAlias = Container::underscore($basename);
+
+                if ($expectedAlias != $extension->getAlias()) {
+                    throw new \LogicException(sprintf(
+                        'Users will expect the alias of the default extension of a plugin to be the underscored version of the plugin name ("%s"). You can override "Bundle::getContainerExtension()" if you want to use "%s" or another alias.',
+                        $expectedAlias, $extension->getAlias()
+                    ));
+                }
+
+                $this->extension = $extension;
```
pamil: There's no such property here (`$this->extension`).

src/Sylius/Bundle/CoreBundle/Application/SyliusPluginTrait.php
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Returns the plugin's container extension class.
+     *
+     * @return string
+     */
+    protected function getContainerExtensionClass()
+    {
+        $basename = preg_replace('/Plugin$/', '', $this->getName());
+
+        return $this->getNamespace().'\\DependencyInjection\\'.$basename.'Extension';
```
pamil: There's no such method here (`$this->getNamespace()`).

lchrusciel: Done 👍

src/Sylius/Bundle/CoreBundle/Application/SyliusPluginTrait.php
```diff
@@ -0,0 +1,109 @@
...
+     * Gets the Bundle namespace.
+     *
+     * @return string The Bundle namespace
+     */
+    abstract protected function getNamespace();
+
+    /**
+     * @return ExtensionInterface|bool
+     */
+    abstract protected function getExtension();
```
pamil: There should be no setters / getters for extension like this. Outside code should always use `getContainerExtension()` and these setters / getters should be replaced with a private property.

pjedrzejewski:
Thanks Łukasz!

pjedrzejewski:
@TheMadeleine Can you update the extension docs and mention this? ;)

PR #7742, opened 2017-03-13
[Core] [Promotion] Move action and rule checkers to bundle

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |

All of these classes has a `getConfigurationType` method which uses FQCN of types from bundles. With current implementation we have coupled these components with its bundles.


gorkalaucirica:
I think moving all the promotions actions and rules to a bundle will make the promotions component almost useless for those using it standalone.

Maybe we can think in creating a small bridge that holds the `getConfigurationFormType()` method and its located in the Bundle and keep the business logic in the component.

**The workaround**

1. Create two classes one for actions and other for rules that receive from constructor a class implementing a `PromotionActionCommandInterface` or a `RuleCheckerInterface` and the form type to be used.

2. Declare all rules and actions as a service in PromotionBundle

3. Define one service for each rule or action using the corresponding class created in step 1, and passing the service created in step 2 along with the form type that should be using

Maybe its a big refactor to be done but in the long term I think it will pay off.

WDYT?

lchrusciel:
@gorkalaucirica thanks for the feedback. After short rethinking, we can do it better, similar to ShippingCalculators we can use form type registry. But I don't have enough time for this refactor... Is somebody willing to do it? /cc @pamil

pjedrzejewski:
@pamil Can you fix this with form registry? :)

pamil:
I'll look into it :)

pamil:
Further development in #7780.

PR #7743, opened 2017-03-13
replace method findByLogicalName by findOneByName

kermorgant:
Hello,

Just installed sylius theme bundle and had to fix that method to get it up and running.


pjedrzejewski:
Thanks Mikael! 👍

PR #7746, opened 2017-03-13
Rename ApiBundle to AdminApiBundle

### LINKED ISSUE #7730: Rename ApiBundle to RestApiBundle / AdminApiBundle?
Hi!

This has been following me for a long time and seems like it is last minute to do this. Sylius currently provide RESTful API for the most important entities and operations, like checkout & cart. That being said, they are done as AdminUser, not the Customer. So it is an Admin API.

I plan to introduce a ShopAPI that will provide a different type of API, suitable for SPAs & mobile devices, etc. I am not sure if it will be in core (1.1) or a separate Plugin. That being said, I want to avoid confusion in a situation where we introduce ShopApiBundle and we end up with ApiBundle.

Options:

1. RestApiBundle
2. AdminApiBundle
3. Keep ApiBundle

BC Breaks:

* Routing: from sylius_api_taxon_create to sylius_rest_api_taxon_create/sylius_admin_api_taxon_create
* Some firewall names, etc, but mostly regexp replacable things.

There was no strong BC promise on the API, so I believe we can still do this. Let me know what do you think!

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | resolves #7730 |
| License         | MIT |

TODO:
 - [x] Upgrade instructions


pjedrzejewski:
Thanks Kamil!

PR #7755, opened 2017-03-14
Double check promotions are eligible in PromotionProcessor (duplicate of #7681)

### LINKED ISSUE #7584: PromotionProcessor should take into account previously applied promotions

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7681, fixes #7584 |
| License         | MIT |

Just a rebased version of #7681, investigating the PHP 5.6 job failure on Travis.

### DIFF
--- a/features/promotion/receiving_discount/receiving_stacked_promotions_with_changing_context.feature
+++ b/features/promotion/receiving_discount/receiving_stacked_promotions_with_changing_context.feature
@@ -1,28 +1,28 @@
 @receiving_discount
 Feature: Receiving stacked promotion with changing context
-  In order to pay proper amount while buying promoted goods
-  As a Customer
-  I want to receive discount for my purchase
+    In order to pay proper amount while buying promoted goods
+    As a Customer
+    I want to receive discount for my purchase
 
-  Background:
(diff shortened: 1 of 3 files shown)


pjedrzejewski:
Thanks Gorka, nice work! 👍

PR #7756, opened 2017-03-14
Don't show shipments segment when no shipments

DmitriyTrt:
In case of digital goods the shipments segment is always empty on the order info page in admin UI, as shipments processor responsible for creating shipments may be disabled.

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |


pjedrzejewski:
@DmitriyTrt I think this should be done on the main template = include the template you modified only if ``order.shipments|length > 0``.

DmitriyTrt:
Sure, see #7757

PR #7757, opened 2017-03-14
Don't show shipments segment when no shipments

DmitriyTrt:
Another variant with the condition in a main template.

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | alternative solution of #7756 |
| License         | MIT |


pjedrzejewski:
Thanks Dmitriy! 👍

PR #7762, opened 2017-03-14
[Core] Remove default username orm listener

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| License         | MIT |

We are using `sylius.shop_user_provider.email_or_name_based ` to provide users. Imo if you want for example use only email as user identifier you should use coresponding provider:
`sylius.shop_user_provider.email_based ` or `sylius.shop_user_provider.name_based`.


pjedrzejewski:
In favor of removing this, but seems like it is needed. (see the failing build)

michalmarcinkowski:
We just need to adjust the API test (remove username from expected response) and it should be fine.

pamil:
It looks to be a bigger issue that we thought at first (no `username` / `email` fields on `ShopUserType`) and will introduce too much confusion (right now there's `Customer::$email` which is always the same as `User::$username` but also there's independent `User::$email`, so it's all complicated). I would like to either clean it all or leave it for now.

PR #7766, opened 2017-03-15
Fix issue-7765. Allow ordering payment methods by gateway config name

### LINKED ISSUE #7765: Error trying to order payment methods by gatewayname
Hi. In the Payment Methods grid in the admin, it looks you can order those by gateway name, but clicking that action throws a exception 
```
[Semantical Error] line 0, col 140 near 'gateway desc': Error: Class Sylius\Component\Core\Model\PaymentMethod has no field or association named gateway
```
We can choose between not allowing ordering by that field or modify the thing a little bit to allow this kind of order. i'm working in a PR for the second option.

c-revillo:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7765  |
| License         | MIT |

See the link for the error. in the fix i join with the gateway config table and then set the correct property to do the order.

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/grids/payment_method.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/grids/payment_method.yml
@@ -29,7 +29,7 @@ sylius_grid:
                     type: string
                     path: gatewayConfig.gatewayName
                     label: sylius.ui.gateway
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
Thank you Carlos!

PR #7773, opened 2017-03-16
Add 'skipping_shipping_step_allowed' setting to Channel Fixture

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | yes |
| BC breaks?      | no |
| License         | MIT |

PR #7776, opened 2017-03-16
Do not apply transition in order fixtures when shipping step is disabled

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | related to #7773 |
| License         | MIT |

When using my earlier change (#7773) to let the fixture set the 'skip shipping method' option, I run into the following error when using the order fixture:

````
  [SM\SMException]                                                             
  Transition "select_shipping" cannot be applied on state "shipping_skipped"   
  of object "Sylius\Component\Core\Model\Order" with graph "sylius_order_checkout"    
````

I think this change should make it possible to combine them, by not applying the specific transition when the setting is enabled.


pjedrzejewski:
The fix is not correct, because this setting decides if you allow skipping, not if skipping is actually possible for this order. There is a separate service for checking that: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Checker/OrderShippingMethodSelectionRequirementChecker.php. :)

stefandoorn:
Ok. So if I understand correctly, using that service to check on the Order should return TRUE when that transformation should be called, right? I would like to adjust the fix, but would like to understand properly first :)

Does this also mean the reason this happens is not due to the previous mentioned change?

pjedrzejewski:
@stefandoorn You are correct. :) The reason it happens is that the method is already selected automatically, on the previous transition. So it skips this step automatically for you. :)

stefandoorn:
Alright. I will try to make the adjustment asap, but of course feel free to take over if it's needed earlier :)

pjedrzejewski:
@stefandoorn Take your time! Also, this will be exactly the same scenario with the payment step, could you adjust it too?

stefandoorn:
@pjedrzejewski I hope this is the change you described :)

pjedrzejewski:
Thanks Stefan, that is correct! :)

PR #7790, opened 2017-03-21
[Core][RFC] Images related semantic fixes

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | |
| License         | MIT |

While woking on a project I encountered some problems with semantics of images-related interfaces and classes. I started from being mislead by `ImageUploaderListener` which in fact listens for **images** rather than **image** uploading. And then my story began :smile:
Proposed changes:

1. Change `ImageAwareInterface` to `ImagesAwareInterface`, as it contains methods for images collection. It's also consistent with other `*AwareInterface`s like `CurrenciesAwareInterface`, `ChannelsAwareInterface`, `ChannelAwareInterface` and `LocalesAwareInterface`.
2. Change `ImageUploadListener` to `ImagesUploadListener` and its method `uploadImage` to `uploadImages` (the same reason).
3. Introduce `ImageOwnerInterface` to base on this abstraction in `Image` class rather than on `ImagesAwareInterface`.

I'm not 100% sure about the last change (also about naming) but I think it would be useful to not couple `Image` with `ImagesAwareInterface`. If, for example, someone would like to create entity with only **one** image (it was my case), it would be easy to create `ImageAwareInterface` which extends the `ImageOwnerInterface` and use *Sylius* `Image` class rather than being forced to implement their own.

Let me know what do you think about this change 🐃  ;)

### DIFF
--- a/src/Sylius/Component/Core/Model/ImageOwnerInterface.php
+++ b/src/Sylius/Component/Core/Model/ImageOwnerInterface.php
@@ -1,0 +1,19 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Core\Model;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface ImageOwnerInterface
+{
+}
--- a/src/Sylius/Bundle/CoreBundle/EventListener/ImagesUploadListener.php
+++ b/src/Sylius/Bundle/CoreBundle/EventListener/ImagesUploadListener.php
@@ -11,7 +11,7 @@
 
 namespace Sylius\Bundle\CoreBundle\EventListener;
 
-use Sylius\Component\Core\Model\ImageAwareInterface;
+use Sylius\Component\Core\Model\ImagesAwareInterface;
 use Sylius\Component\Core\Uploader\ImageUploaderInterface;
 use Symfony\Component\EventDispatcher\GenericEvent;
 use Webmozart\Assert\Assert;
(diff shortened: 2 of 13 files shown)


stefandoorn:
I think the proposed changes sound good, as now it's mainly centered around Collections. Next to this, can we also keep an ImageAwareInterface / ImageType to support easy development when you only need a single image? The cookbook has a section about adding images to entities, and in there it makes also sense to have just one images sometimes (let's say a supplier entity with a logo image).

src/Sylius/Component/Core/Model/ImageOwnerInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sylius\Component\Core\Model;
+
+/**
+ * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
+ */
+interface ImageOwnerInterface
```
pamil: From my POV introducing `ImageOwnerInterface` does not bring us any real benefits, but only complicates things. It gives me nothing to know that `Image::getOwner()` returns an object implementing `ImageOwnerInterface` because I can't do anything with that knowledge.

Zales0123: OK, so how would you solve coupling `Image` with `ImagesAwareInterface`? We cannot assume that owner has multiple images, as it results in problems described above.

pamil: We just don't, `ImageOwnerInterface` typehint is equal to `object` typehint in this case.

Zales0123: I agree, however I still don't see any other way to not couple `Image` with `ImagesAwareInterface` which is not desirable.

Zales0123: /cc @pjedrzejewski @michalmarcinkowski I'd be glad to see your opinion about that ;)

pamil: Hmm, just remove those `ImageOwnerInterface` typehints?

PR #7800, opened 2017-03-21
[Resource] Implement find method in InMemoryRepository

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| License         | MIT |

Can be useful 🎉

### DIFF
--- a/src/Sylius/Component/Resource/Repository/InMemoryRepository.php
+++ b/src/Sylius/Component/Resource/Repository/InMemoryRepository.php
@@ -98,9 +98,9 @@ class InMemoryRepository implements RepositoryInterface
      *
      * @throws UnsupportedMethodException
      */
-    public function find($id = null)
+    public function find($id)
     {
-        throw new UnsupportedMethodException('find');
+        return $this->findOneBy(['id' => $id]);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Sylius/Component/Resource/Repository/InMemoryRepository.php
```diff
@@ -98,9 +98,9 @@ public function remove(ResourceInterface $resource)
      *
      * @throws UnsupportedMethodException
      */
-    public function find($id = null)
+    public function find($id)
     {
-        throw new UnsupportedMethodException('find');
```
pamil: @NeverResponse can you tell why it wasn't implemented in the first place?

pamil: Hmm, it was https://github.com/Sylius/Sylius/pull/3617#r45731850.

pamil:
It was unimplemented in the first place due to IDs being generated on flush for our Doctrine entities, but that should not be the reason not to allow for that (especially that any other project with different ID strategy might use it then).

pamil:
Thanks @Arminek!

PR #7806, opened 2017-03-22
[DX] Create product autocomplete form type

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

This change will just increase a DX for developers which would like to take an advantage of product autocomplete feature. Right now, it is not possible without creating a new block definition. With this change, it would be enough to just use `ProductAutocompleteChoiceType`


pjedrzejewski:
Thank you Łukasz! 👍

PR #7819, opened 2017-03-23
Do not validate order when changing payment method after it is placed

### LINKED ISSUE #7707: Cannot retry payment due to out of stock due to my own order
**Steps to reproduce**

1. Add to cart a product variant with 1 unit left on stock
2. Proceed to checkout and select Paypal for example, as a payment method
3. Cancel paypal payment and return to order show page
4. Retry with paypal.
5. Get "Product does not have sufficient stock" error

**Investigation**

We see that after checkout is completed that product is marked as "onHold" and therefore is out of stock as we have 1 on hand and 1 on hold.

Somehow the InStockValidator needs to check if the products we are trying to pay is on hold in the same order

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7707 |
| License         | MIT |


pjedrzejewski:
Thank you Kamil! 👍

PR #7830, opened 2017-03-24
Make product variant match selection method working correctly again

### LINKED ISSUE #7813: Cannot add variant to cart except the first one
Since my last update on master, I cannot add any product variant to cart except the first one in the option list. Form is showing me up that the value is invalid. I tried to debug, but couldn't find any answer for that. :-\

UPDATE: Also in the admin if I edit a product variant having more than one option, the pre-selected options are not correct. If I create a new Variant.

UPDATE: I found in the admin form the contents of 2 option types are swiched. Means I have option A containing the value of option B and vice versa. I think its happening wenn the sorting of `sylius_product_option` is not the same like it is in `sylius_product_variant_option_value`.

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7813 |
| License         | MIT |

### DIFF
--- a/features/cart/shopping_cart/adding_product_with_options_to_cart.feature
+++ b/features/cart/shopping_cart/adding_product_with_options_to_cart.feature
@@ -8,13 +8,13 @@ Feature: Adding a product with selected option to the cart
         Given the store operates on a single channel in "United States"
 
     @ui
-    Scenario: Adding a product with multiple options to the cart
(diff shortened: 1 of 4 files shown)


pjedrzejewski:
Thanks Kamil!

PR #7840, opened 2017-03-27
Fix product slug field in Admin

Pitoune:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

When you want to add a translation to a Product in admin, the slug field is locked when you have a slug for the default locale. I think that the slug field should not be locked by default if there is no value.


pjedrzejewski:
Thanks Pierre!

PR #7721, opened 2017-03-10
[Core] Configure encrypting for payment gateway configuration

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | #7582 |
| License         | MIT |

PR #7855, opened 2017-03-29
Store encrypted gateway configurations in the database

### LINKED ISSUE #7582: [PayumBundle] Gateway Config
I've been looking at gateway configuration so far, and now that it is web based, it is way less powerful, and I don't know if overriding each of the Form type's and adding the information in the Admin UI is necessarily the right way to go.

For example, for Stripe Checkout, I override two main parameters: `payum.template.obtain_token` (to customize look and feel) and `payum.action.convert_payment` (to perform hackery and grab the customer email to auto populate).

These parameters don't belong in a UI, in my opinion. They should belong in code, but I just can't see where that could happen in the current configuration.

Edit: From further testing, where are configuration like `sandbox` for `paypal_express_checkout` set? They are defaulted to `true` in `PaypalExpressCheckoutGatewayFactory`.

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | closes #7721, fixes #7582 |
| License         | MIT |


app/config/parameters.yml.dist
```diff
@@ -14,4 +14,5 @@ parameters:
     mailer_password: ~
 
     secret: EDITME
+    defuse_secret: def000003f125cbd07bfce7748bf501fa3f5ebb6b8a4d785ddb7b328948ea4e88973d5c1cdf79a5a2784b723ad2fb9d802111d73ec98be7ef55e47966a89e9b1fcfa35b0
```
makasim: I am not sure we should provide any default secrets. It is not secure.

pamil: There are three aproaches here:
 - do not provide default defuse secret (so app will crash when trying to use `EDITME` defuse secret key)
 - provide default defuse secret key (so it would change nothing from the current situation)
 - do not encrypt gateway configs by default and provide a cookbook (so less people will even know about it)

makasim: Let's say I used a default key for some time and then realized I had to change it. Now I have data encrypted with one key and I have to somehow decrypt it first and then crypt with a new secret. It could be tricky to do.

pamil: That makes sense 👍  I'll make a cookbook then.

pjedrzejewski:
Great, thanks Kamil and Maksim! 👍

PR #7876, opened 2017-04-01
[Checkout] Creating Payment Method skip channel option, effective if …

ktzouno:
…there is only 1 Payment Method available

| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| License         | MIT |

Implements a switch in Admin Channel page, to allow skipping the Payment Step of the checkout if there is only 1 available payment method.

It has been inspired and works the same way as https://github.com/Sylius/Sylius/pull/7488

### DIFF
--- a/app/migrations/Version20170401200415.php
+++ b/app/migrations/Version20170401200415.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Sylius\Migrations;
+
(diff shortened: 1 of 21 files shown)


pjedrzejewski:
Thanks @ktzouno! Great work! 👍

PR #7881, opened 2017-04-03
Set default username on existing customer binding

DmitriyTrt:
Existing customer may be kept untouched on user creation because the relation is stored on the user table. So, listener should react to both shop user and customer flush, in order to guarantee the username is set.

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/EventListener/DefaultUsernameORMListener.php
+++ b/src/Sylius/Bundle/CoreBundle/EventListener/DefaultUsernameORMListener.php
@@ -16,6 +16,7 @@ use Doctrine\ORM\Event\OnFlushEventArgs;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\UnitOfWork;
 use Sylius\Component\Core\Model\CustomerInterface;
+use Sylius\Component\Core\Model\ShopUserInterface;
 
 /**
  * Keeps user's username synchronized with email.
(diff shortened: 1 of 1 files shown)


DmitriyTrt:
Test fails look irrelevant to the changes.

lchrusciel:
The build should be restarted indeed. /cc @pjedrzejewski 
Or you can "restart" it by closing and reopening this PR @DmitriyTrt

DmitriyTrt:
>> The job exceeded the maximum time limit for jobs, and has been terminated.

Failed on some downloads this time...

src/Sylius/Bundle/CoreBundle/EventListener/DefaultUsernameORMListener.php
```diff
@@ -42,13 +43,20 @@ public function onFlush(OnFlushEventArgs $onFlushEventArgs)
...
-            $user = $customer->getUser();
-            if (null === $user) {
+        foreach ($entities as $entity) {
+            if ($entity instanceof CustomerInterface) {
+                $customer = $entity;
+                $user = $customer->getUser();
+                if (null === $user) {
+                    continue;
+                }
+            } elseif ($entity instanceof ShopUserInterface) {
```
lchrusciel: I'm not sure about this solution. If we want to keep consistency between username and email, we should prevent from changing it. If we allow for this change, then we will receive a bug report that when I change shop users username, it doesn't change... 

Also, we should avoid `elseif` construction. If we want it in core, it should be refactored and extracted to the private method.

Zales0123:
Closing this for now, as there is clearly no desire to have it in the core (over one year is enough :D). If anyone thinks it still should be considered (@DmitriyTrt or @Sylius/core-team), feel welcomed to reopen :)

PR #7885, opened 2017-04-04
Add sandbox to paypal gateway configuration form

### LINKED ISSUE #7860: Gateway configuration needs more extensability
I know Sylius is in a crunch to get beta2 out, but there are currently a few problems with payment gateway configuration that need to be addressed first (IMO).

1. `PaypalGatewayConfigurationType` is missing the  `sandbox` parameter
2. `StripeGatewayConfigurationType` is missing the `test_mode` parameter
3. There needs to be an easy way to override not only template, but other gateway configurations (for example `payum.action.convert_payment`) via code. 

I've brought up the last item a few items (in past tickets, slack) but none of the solutions seem like a good idea. Is the solution extending each of the gateway factories?

inssein:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | partially fixes #7860  |
| License         | MIT |


src/Sylius/Bundle/PayumBundle/Resources/translations/messages.en.yml
```diff
@@ -7,9 +7,11 @@ sylius:
                 password: Password
                 signature: Signature
                 username: Username
+                sandbox: Sandbox
```
Zales0123: Could you please keep the alphabetical order of keys? 😄

pjedrzejewski:
👍  I have restarted the build.

inssein:
Actually after upgrading to master, I realized test_mode is actually never used. It was used with the other stripe gateway initially, and not actually used by express checkout. I will be removing it shortly.

inssein:
build needs another restart =(

venyii:
💡 You can also close/reopen the PR to get another build started

pjedrzejewski:
@inssein It seems like it's a problem with the changes from this PR, not the build, could you have a look?

pjedrzejewski:
Added the setting for PayPal.

PR #7909, opened 2017-04-06
Add 'configuration' node in 'product_attribute' fixture

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

Adds the 'configuration' node to the 'product_attribute' fixture. It was now missing, as it's only done inside fixtures like 'mug_product'. Simply allows an array with configuration options, as there are many possibilities based on each type.


pjedrzejewski:
Thank you Stefan!

PR #7920, opened 2017-04-10
Remove unused private __construct in Core/Formatter/StringInflector

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

### DIFF
--- a/src/Sylius/Component/Core/Formatter/StringInflector.php
+++ b/src/Sylius/Component/Core/Formatter/StringInflector.php
@@ -45,8 +45,4 @@ final class StringInflector
     {
         return strtoupper(self::nameToCode($value));
     }
-
-    private function __construct()
-    {
-    }
 }


src/Sylius/Component/Core/Formatter/StringInflector.php
```diff
@@ -45,8 +45,4 @@ public static function nameToUppercaseCode($value)
     {
         return strtoupper(self::nameToCode($value));
     }
-
-    private function __construct()
```
venyii: Is it really unused? It prevents instantiating a purely static class?

stefandoorn: Hm, not sure about that if that can be fixed using a private constructor. Besides it doesn't make a lot sense to instantiate this class anyway :)

pjedrzejewski:
This is done on purpose to prevent creation of the object. This is just a static helper class. ;)

PR #7926, opened 2017-04-11
[API] [Product] Add path to default serialization

### LINKED ISSUE #7898: Product API : images path missing within the products collection
My goal is to list products with its default image on the front office of my ecommerce website.

Following the doc http://docs.sylius.org/en/latest/api/products.html, the GET method (collections of products) does not permit to get the images path (only the id and name... Which are not really useful)

So, right now, I have to iterate on each product to GET it via the API /api/v1/products/{code}... Dirty.

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7898 |
| License         | MIT |

The path will be included in default serialization. I have also added a new fixture to avoid regression. The question for later is, should also send an information about related variants.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/serializer/Model.Image.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/serializer/Model.Image.yml
@@ -14,7 +14,7 @@ Sylius\Component\Core\Model\Image:
         path:
             expose: true
             type: string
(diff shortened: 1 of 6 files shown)


pjedrzejewski:
Thank you Łukasz!

PR #7927, opened 2017-04-11
Removed "GROUP BY" clause in createShopListQueryBuilder

marcoslibre:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

The query for retrieving the products list in the shop does not work in PostgreSQL :
```
SQLSTATE[42803]: Grouping error: 7 ERROR: column "s1_.name" must appear in the GROUP BY clause or be used in an aggregate function
```
The problem is, that in PostgreSQL you need to add all the columns you want to have in your select to your group by.
This is not specific to PostgreSQL, this is required by the SQL standard. 
The query works in MySQL, but MySQL's sloppy implementation of this is more like a bug than a feature.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
@@ -74,7 +74,6 @@ class ProductRepository extends BaseProductRepository implements ProductReposito
             ->andWhere('productTaxon.taxon = :taxon')
             ->andWhere(':channel MEMBER OF o.channels')
             ->andWhere('o.enabled = true')
-            ->addGroupBy('o.id')
             ->setParameter('locale', $locale)
             ->setParameter('taxon', $taxon)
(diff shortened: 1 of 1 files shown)


psihius:
This is a problem also since some 5.7.x (5.7.6?) where they enabled the strict GROUP BY by default. So any default installation on Ubuntu 16.04 LTS is going to throw the error.

lchrusciel:
Yes, exactly. We had problems with setting up travis with given version of MySQL, so this bug is still not tested properly and such problems can occur. We have to fix this one as well, but removing this one line will not be enough...

marcoslibre:
I have tried to add all unaggregated columns in the group by clause instead of removing the line, but it still breaks the tests...

FanchTheSystem:
Please check https://github.com/Sylius/Sylius/pull/7978

marcoslibre:
#7978 should fix it

PR #7935, opened 2017-04-13
[ResourceBundle] Allow nested parameter parser.

liverbool:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    |yes |
| BC breaks?      | no |
| Related tickets | ~ |
| License         | MIT |

```php
// it's spec
function it_parses_nested_parameters()
{
    $request = new Request();
    $request->request->set('array', ['foo' => 'bar']);

    $this
        ->parseRequestValues(['nested' => ['value' => '$array.foo']], $request)
        ->shouldReturn(['nested' => ['value' => 'bar']])
     ;
}

```


src/Sylius/Bundle/ResourceBundle/Controller/ParametersParser.php
```diff
@@ -75,15 +75,45 @@ private function parseRequestValue($parameter, Request $request)
     }
 
     /**
+     * @param string $parameter
+     * @param Request $request
+     *
+     * @return mixed
+     */
+    private function getRequestValue($parameter, Request $request)
```
pamil: I think we should use PropertyAccessor instead of operating on arrays.

liverbool:
be useful

liverbool:
outdated!

PR #7964, opened 2017-04-20
[ProductVariant] Add getImages to product variant

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |

The simplest implementation for providing related images from variant. It is not an unusual use-case so it will slightly improve DX, but on the other hand, it is not a pretty solution and also it is a BC-break. I'm not sure is it worthy to merge to the master at this stage, but I had to implement it one way or another, so you can decide :)

Also, usage of this function will be really expensive from performance point of view if a lot of images will be defined on a product.

### DIFF
--- a/src/Sylius/Component/Core/Model/ProductVariant.php
+++ b/src/Sylius/Component/Core/Model/ProductVariant.php
@@ -401,4 +401,22 @@ class ProductVariant extends BaseVariant implements ProductVariantInterface
     {
         $this->shippingRequired = $shippingRequired;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getImages()
+    {
+        $images = [];
+
+        foreach ($this->product->getImages() as $image) {
+            foreach ($image->getProductVariants() as $productVariant) {
+                if ($this === $productVariant){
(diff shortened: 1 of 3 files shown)


Snake-Tn:
As you mentioned, the are loading all the product's images while you need only a sub-set.

I think the best solution is to add `images` attribute to `ProductVariant` entity. then you setup the mapping and let doctrine do the rest
ORM Mapping:

`Sylius/Component/Core/Model/ProductVariant:
    manyToMany:
            images:
                targetEntity: Sylius\Component\Core\Model\ProductImageInterface
                mappedBy: productVariants`

pamil:
Related to #7993.

lchrusciel:
At some point, we have decided that keeping images on a product only makes sense. I don't remember tradeoffs, thus I will wait for others with a decision.

Snake-Tn:
Related to [#8110](https://github.com/Sylius/Sylius/pull/8110)

pjedrzejewski:
Implemented in #8110.

PR #7972, opened 2017-04-21
Allow .woff2 files in platform.sh config

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

Sylius admin tries to load a .woff2 file (icons.woff2) but the platform.sh config does not allow it which results in a 404.


pjedrzejewski:
Thanks Stefan!

PR #7982, opened 2017-04-25
Update redirect after email verification

4oXdev:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

Usually, after successful registration, users will immediately check the post and follow the link in the email and thinking that have already entered, but it is not. This changes will redirect user to login page or to account dashboard.


pjedrzejewski:
👍  Thanks @4oXdev!

PR #7988, opened 2017-04-25
Update label for street field in address form

4oXdev:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

Usually street field is called 'Street address'. Because it's not just the name of the street.


pjedrzejewski:
👍

PR #7992, opened 2017-04-26
[Shop] [Admin] Move liip imagine routing to the main app

### LINKED ISSUE #7984: [Dependencies] [RFC] Liip images routing is defined only for ShopBundle
Routing for Liip images is defined in shop bundle, but we are using it in admin also if I'm correct. Probably should be moved to the main routing file(similar to payum). It will also remove locale prefix in the image path.

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7984 |
| License         | MIT |

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -181,6 +181,13 @@
 
 * `ContactController` has been made final, use decoration instead of extending it.
 
+* `_liip_imagine` routing import has been removed and move to the main routing of Sylius app. If you are not importing `app/config/routing.yml` you need to add this import by your own:
+
+    ```yml
+        _liip_imagine:
+            resource: "@LiipImagineBundle/Resources/config/routing.xml"
+     ```
+ 
 ### Shipping / ShippingBundle
 
 * `UnresolvedDefaultShippingMethodException` has been made final, use decoration instead of extending it.
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -23,3 +23,6 @@ sylius_shop_default_locale:
     methods: [GET]
     defaults:
         _controller: sylius.controller.shop.locale_switch:switchAction
+
+_liip_imagine:
+    resource: "@LiipImagineBundle/Resources/config/routing.xml"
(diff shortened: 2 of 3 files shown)


lchrusciel:
The benefit is that you don't have to carry about adding these lines to your setup when you plug in shop or admin. Otherwise, it will probably break. It doesn't matter if both are turned on. But I can agree that putting it in `app/config/routing.yml` is probably the cleanest solution.

pamil:
> It doesn't matter if both are turned on.

If there are the same two routes, the last wins. Currently there is only one route imported by ShopBundle routing, this PR introduces another one in AdminBundle, but changes nothing, since admin routing is imported *before* shop routing. 

Moving it to `app/config/routing.yml` will make sure we won't have any issues with image urls being behind admin firewall if routing loading order changes. Also, since we've introduced localised urls for shop, there are multiple urls leading to the same resource now (eg. `en_US/image.jpg`, 'de_DE/image.jpg`) and it complicates caching them for sure.

lchrusciel:
@pamil updated.

pjedrzejewski:
@lchrusciel This is a BC break, we should mention it in the upgrade file and also upgrade Sylius-Standard.

lchrusciel:
I will resolve it this weekend

UPGRADE.md
```diff
@@ -181,6 +181,13 @@
 
 * `ContactController` has been made final, use decoration instead of extending it.
 
+* `_liip_imagine` routing import has been removed and move to the main routing of Sylius app. If you are not importing `app/config/routing.yml` you need to add this import by your own:
+
+```yml
+_liip_imagine:
+    resource: "@LiipImagineBundle/Resources/config/routing.xml"
+ ```
```
pamil: Extra space here

lchrusciel: done

Zales0123:
Thank you Łukasz! ⚙️

PR #7996, opened 2017-04-26
Do not overwrite user changes after PRE_EMAIL_VERIFICATION

aistis-:
Current event dispatching positioning does not allow to keep user disabled after the registration. `$user->disable()` would be overwritten with `$user->enable()` in this case.

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |


aistis-:
Will it reach v1? :)

lchrusciel:
Thanks Aistis :)

PR #7997, opened 2017-04-26
Added missing macro call parameters

Persata:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

Hi all,

I noticed that when adding an attribute with child fields (e.g. `datetime` or `date`) to a product, the generated markup from `ProductAttributeController::renderAttributeValueFormsAction` is incorrect, which leads to the attribute & data not being saved against the product.

The generated markup is this:

```html
    <div class="ui label">Attribute Name</div>

    <select id="attribute_name" name="_[attributes][0][value][date][year]" data-name="_[attributes][0][value][date][year]">
        <option value="2012">2012</option>
        <option value="2013">2013</option>
        <option value="2014">2014</option>
        <option value="2015">2015</option>
        <option value="2016">2016</option>
        <option value="2017">2017</option>
        <option value="2018">2018</option>
        <option value="2019">2019</option>
        <option value="2020">2020</option>
        <option value="2021">2021</option>
        <option value="2022">2022</option>
    </select>
```

Note the missing values around the `_` in `name` and `data-name` attributes. The expected (and working) markup should be:

```html
    <div class="ui label">Attribute Name</div>

    <select id="attribute_name" name="sylius_product[attributes][0][value][date][year]" data-name="sylius_product[attributes][0][value][date][year]">
        <option value="2012">2012</option>
        <option value="2013">2013</option>
        <option value="2014">2014</option>
        <option value="2015">2015</option>
        <option value="2016">2016</option>
        <option value="2017">2017</option>
        <option value="2018">2018</option>
        <option value="2019">2019</option>
        <option value="2020">2020</option>
        <option value="2021">2021</option>
        <option value="2022">2022</option>
    </select>
```

which has the correct `sylius_product` designation.

I've tracked the problem to https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/AdminBundle/Resources/views/Product/Attribute/attributeValues.html.twig#L28 where the `subject` and `applicationName` parameters are not passed back into the recursive macro when rendering the children.

I've tested this fix on the `date` and `datetime` attributes, and AFAIK these are the only ones affected by this issue, since no other attributes have child elements in their forms.

If you foresee any problems, let me know and I'll delve into it further.

Thanks!

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Product/Attribute/attributeValues.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Product/Attribute/attributeValues.html.twig
@@ -25,7 +25,7 @@
     {% if item.children|length > 0 %}
         {% set prefix = prefix~'_'~item.vars.name %}
         {% for child in item.children %}
-            {{ _self.formField(child, count, id, prefix) }}
+            {{ _self.formField(child, count, id, prefix, subject, applicationName) }}
         {% endfor %}
     {% elseif item.vars.name != '_token' %}
         {% set namePrefix = prefix|replace({'_': ']['}) %}


sweoggy:
@pamil well, thanks to that flexibility I am able to re-use these templates for attributes to product variants. The flexibility is one of the strong points with Sylius as it saves a lot of time. It is worth keeping if issues require as small changes as this one in my opinion.

Persata:
@pamil Thanks!

I had originally considered removing those parameters and hard-coding the `sylius_product` section of the name, but like @sweoggy has pointed out, that could break BC in some cases if people are using the template elsewhere.

Is it better / more straightforward to keep the parameters in place and apply this fix, then look at revisiting this in future and putting a relevant entry into UPGRADE.md?

pjedrzejewski:
Thank you Ross! :) I think it can be this way and if it causes us some trouble, we might consider hardcoding the appropriate values.

PR #7999, opened 2017-04-27
Add layout events for head & body

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| License         | MIT |

Makes it easier for plugin development to add things to head / end of body etc.

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/layout.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/layout.html.twig
@@ -2,6 +2,8 @@
 
 <html lang="{{ app.request.locale|slice(0, 2) }}">
 <head>
+    {{ sonata_block_render_event('sylius.shop.layout.before_head') }}
+
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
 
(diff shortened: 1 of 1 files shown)


stefandoorn:
I'm using this e.g. in https://github.com/stefandoorn/google-tag-manager-plugin to add the tracking codes to the appropriate locations.

lchrusciel:
Thanks Stefan :)

psren:
Is the block "sylius.shop.layout.before_head" dangerous?
IMO nothing should appear in the html-source before the ```charset```? In the definition of w3.org the charset should appear in die first 1024 chars of the content. If plugins add content to that block this limit could be reached.

stefandoorn:
Basically it's inside the <head>. I tried to follow naming conventions used on several places, where `before` most often means at the beginning of something (before other things are rendered in the head in this case), while after means just after the content of it (after other things inside the head, so basically before `</head>`. 

Of course we can still adjust the naming of it, to make it more declarative.

stefandoorn:
@pjedrzejewski @lchrusciel What are your thoughts on this naming thing?

stefandoorn:
And now on topic, sorry @psren. Maybe we can put the charset higher, or this hook lower to prevent this issue you mentioned.

lchrusciel:
Thanks @psren for pointing out, we should fix that of course. 

@stefandoorn after a while, I am thinking that only `sylius.admin.layout.head` should be placed just before `</head>`. Is there any particular to have to inject points there?

PR #8000, opened 2017-04-27
[Core] [Shop] Move shop session depend listeners

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |


pamil:
#8000, congrats! 🎉

sweoggy:
Is this really a "critical bug" which makes it valid to be introduced this late in the beta process?

lchrusciel:
We could skip it, but without this bug fix, you could have problems with auto starting sessions when using a core bundle only for your hypothetical stateless app. If we move this service a session will be a ShopBundle part only.

There is a workaround for this - one can always unregister this service. But I would prefer to move it.

Zales0123:
@lchrusciel there a change need in phpspec and conflict in `UPGRADE.md` to be resolved. Then, I think it's ready to be merged 👍

lchrusciel:
Will be done in nearest days

lchrusciel:
@Zales0123 done ;)

Zales0123:
Still some issues with phpspec :/

lchrusciel:
@Zales0123 fixed

src/Sylius/Bundle/ShopBundle/Resources/config/services/listeners.xml
```diff
@@ -52,5 +52,11 @@
             <argument type="service" id="sylius.context.locale" />
             <tag name="kernel.event_listener" event="sylius.order.pre_complete" method="assignLocale" />
         </service>
+
+        <service id="sylius.listener.session_cart" class="Sylius\Bundle\ShopBundle\EventListener\SessionCartSubscriber">
+            <argument type="service" id="sylius.context.cart" />
+            <argument>_sylius.cart</argument>
```
pamil: After further investigation, we use it in three services definitions and `ShopUserLogoutHandler` has it hardcoded, we can extract it to some parameter in another PR.

UPGRADE.md
```diff
@@ -7,10 +7,14 @@
 * In order to be compatibile with Doctrine ORM 2.6+ and be more consistent 
   `OrderRepositoryInterface::count()` signature was changed to `OrderRepositoryInterface::countPlacedOrders()`.
 
+* The service definition of `sylius.context.cart.session_based` has been removed. Declare it by your own if you want to use `SessionBasedCartContext`
+
 ### Core / CoreBundle
 
 * `SessionCartSubscriber` and `ShopUserLogoutHandler` has been moved to ShopBundle. If you used them, you need to add ShopBundle to your Kernel or define this services by your own.
 
+* The service definition of `session_and_channel_based` has been moved to ShopBundle. If you used it, you need to add ShopBundle to your Kernel or define this services by your own.
```
pamil: `session_and_channel_based` -> `sylius.context.cart.session_and_channel_based`

and

`The service definition of foo` -> `The foo service definition`?

UPGRADE.md
```diff
@@ -7,10 +7,14 @@
 * In order to be compatibile with Doctrine ORM 2.6+ and be more consistent 
   `OrderRepositoryInterface::count()` signature was changed to `OrderRepositoryInterface::countPlacedOrders()`.
 
+* The service definition of `sylius.context.cart.session_based` has been removed. Declare it by your own if you want to use `SessionBasedCartContext`
```
pamil: Can we also include the deleted code here?

src/Sylius/Bundle/ShopBundle/ShopSession.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ShopBundle;
+
+final class ShopSession
+{
+    public const SESSION_NAME = '_sylius.cart';
```
pamil: In fact it isn't a session name, but a key under a cart id is stored in session. There are a few more places that use `_sylius.cart`, let's change either all of them or none.

Another thought: why wouldn't we make it a service container parameter? This way end users will be able to change it in their apps.

pjedrzejewski:
Thanks Łukasz!

PR #8004, opened 2017-04-28
Update RandomnessGenerator.php

### LINKED ISSUE #7983: Problem with a dot in verification email token
Email verification token could generated with a dots. I think that this is basically not good.
Case: If we have a token with a dot at the end, this causes a problem with recognition of the link in the verification email, because the raw link is used.

jspizziri:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #7983|
| License         | MIT |


pjedrzejewski:
Thank you Jacob!

PR #8016, opened 2017-05-03
[Docs] Update the Slack invitation link

### LINKED ISSUE #7922: Sylius Slack invitation link no longer works
I know this is probably not the right place to report this, but I don't know where else I can report it.

---

On Symfony Slack, a user told us that `slackinvite.me/to/sylius-devs` is down. Indeed, `slackinvite.me` seems to be down, that's why we removed it for the Symfony Slack.

A simple alternative solution is to self-host an application like https://github.com/rauchg/slackin  but at the end we decided to use a "shared invite" which is a link with auto sign-up. The number of people it can use it is limited, but you can renew it. It should be available at `https://sylius-devs.slack.com/admin/shared_invites`

kochen:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Doc fix?           | yes |
| Related tickets | resolved #7922 |
| License         | MIT |

### DIFF
--- a/docs/support/index.rst
+++ b/docs/support/index.rst
@@ -20,7 +20,7 @@ We have created following rooms so far:
 
 If you have any suggestions, please let us know on Slack!
 
-Slack requires inviting new members, but this can be done automatically, just go to `slackinvite.me/to/sylius-devs <https://slackinvite.me/to/sylius-devs>`_,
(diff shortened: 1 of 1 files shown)


Arminek:
👍

PR #8034, opened 2017-05-09
Move the before head hook lower due to conflict with W3 specs

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no|
| BC breaks?      | no |
| Related tickets | mentioned in https://github.com/Sylius/Sylius/pull/7999#issuecomment-300183360 |
| License         | MIT |

As @psren pointed out in the comment linked above, it might be better to move the 'before_head' listener a bit lower based on this information: https://www.w3.org/International/questions/qa-html-encoding-declarations.

Also it seems that the X-UA-Compatible meta tag should be high too: https://msdn.microsoft.com/en-us/library/jj676915%28v=vs.85%29.aspx.

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/layout.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/layout.html.twig
@@ -2,8 +2,6 @@
 
 <html lang="{{ app.request.locale|slice(0, 2) }}">
 <head>
-    {{ sonata_block_render_event('sylius.shop.layout.before_head') }}
-
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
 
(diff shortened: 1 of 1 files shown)


lchrusciel:
Can you remove this tag and rename `after_head` to `head`?

stefandoorn:
Sure. I'm just in doubt whether we want to be able to put things before the CSS and after the CSS. But maybe I'm trying to configure too much, as maybe a single entry point in the head might be enough. WDYT? The CSS might prevent something to load early, but it should be minimal impact if servers are configured properly I guess.

lchrusciel:
We always have a possibility to provide custom metatag block before css. One injection point should be enough :)

stefandoorn:
Ok :) Then it's now good to go I think after the tests.

lchrusciel:
Nice. Thanks, Stefan :)

psren:
Holy... that was fast. Nice job 👍

PR #8038, opened 2017-05-10
[Doctrine] [RFC] Set fetch join collection to true for pagination

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | [This thread with everything related](https://github.com/Sylius/Sylius/pull/6674#issuecomment-298580606) and mainly [this issue](https://github.com/Sylius/Sylius/issues/3836). |
| License         | MIT |

It is not a fix, but it will show what will break and maybe somebody will be able to fix properly.

### DIFF
--- a/src/Sylius/Bundle/GridBundle/Doctrine/ORM/DataSource.php
+++ b/src/Sylius/Bundle/GridBundle/Doctrine/ORM/DataSource.php
@@ -71,7 +71,7 @@ final class DataSource implements DataSourceInterface
     public function getData(Parameters $parameters)
     {
         // Use output walkers option in DoctrineORMAdapter should be false as it affects performance greatly. (see #3775)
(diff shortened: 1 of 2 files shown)


lchrusciel:
Broken builds were expected. The question is, how can we fix them. Also, it would be really nice to see a benchmark for this, but I don't have time for that atm. :(

dannyvw:
I have the same issues with the pagination. "fetchJoinCollection: true" fixes these issues. But if you add a sort by name (which is a translation) then you will get the following error message "Cannot select distinct identifiers from query with LIMIT and ORDER BY on a column from a fetch joined to-many association. Use output walkers." "useOutputWalkers: true" solves this issue. But not sure if this is the right way.

PR #8058, opened 2017-05-16
Remove count constraint on option values collection

alcaeus:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | / |
| License         | MIT |

Currently, a product option must have at least two values for it to be valid. While it may not make a lot of sense to have product options with only one value, there is also no reason for those options not to exist.

In our case, a product may be available in one or more colors. Since colors in a clothing store range from your usual colors (red, blue, green) to weird combinations (striped red and blue) we create a new option for each product using the API. However, if a product were only available in a single color, that color would have to be added as an attribute, which makes looking for all products that are available in a blueish (= color field contains "blue") color difficult.

Unfortunately, removing this limit is not possible without re-implementing the `ProductOption` entity completely, which we'd like to avoid (validation constraints can only be added, not removed in an extending class).

### DIFF
--- a/features/product/managing_product_options/adding_product_option.feature
+++ b/features/product/managing_product_options/adding_product_option.feature
@@ -18,3 +18,23 @@ Feature: Adding a new product option
         And I add it
         Then I should be notified that it has been successfully created
         And the product option "T-Shirt size" should appear in the registry
+
+    @ui
+    Scenario: Adding a new product option without any option values
+        Given I want to create a new product option
+        When I name it "T-Shirt size" in "English (United States)"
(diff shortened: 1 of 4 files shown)


alcaeus:
@pamil done!

pamil:
That was an instant reply, thanks! 🎉

pamil:
@Sylius/core-team, any comments? :)

lchrusciel:
Totally makes sense for me. 

Thanks, Andreas! :)

PR #8062, opened 2017-05-17
Fixes for shipment processor

Persata:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

Hi all,

We've been working on a store with a mix of products where some require shipping and some do not.

I stumbled upon an issue related to skipping the shipping method step, where an order is automatically assigned a shipment & shipping method even though one is not required, which can add a cost to the order, even though shipping should be free.

I've updated the `OrderShipmentProcessor` class to check if the order has any items that require shipping before trying to create & assign a shipment, as well as clearing any existing shipments if this is the case (for example, 2 items in cart, one requires shipping, one does not, if the item that requires shipping is removed, so the shipment should be removed too).

I've updated the `skipping_shipping_step_when_order_does_not_require_any_shipping.feature` Behat feature with the checks and behaviour I'd expect, which then fails without this fix in place - with the fix, the tests pass.

Any thoughts or issues?

Thanks!


features/checkout/skipping_shipping_step_when_order_does_not_require_any_shipping.feature
```diff
@@ -24,6 +25,7 @@ Feature: Skipping shipping step when order does not require any shipping
     Scenario: Seeing checkout shipping page after addressing if at least one of order items require shipping
         Given I have "Guards! Guards! - ebook" variant of product "Guards! Guards!" in the cart
         And I have "Guards! Guards! - book" variant of product "Guards! Guards!" in the cart
+        Then my cart shipping total should be "$5.00"
```
Zales0123: The flow of the Behat scenario should be: `Given` -> `When` -> `Then` (with some `And`s between them). Therefore, we should not check the result of some assumption and then make more assumptions for further check. The best choice would be to create two new scenarios that would just check cart shipping total based one some `Given`s. Like `Seeing free shipping total when none of order items require shipping` and `Seeing not free shipping total when at least one of order items require shipping`? They could be even places in separate feature, I think ;)

src/Sylius/Component/Core/OrderProcessing/OrderShipmentProcessor.php
```diff
@@ -37,7 +38,7 @@
 
     /**
      * @param DefaultShippingMethodResolverInterface $defaultShippingMethodResolver
-     * @param FactoryInterface $shipmentFactory
+     * @param FactoryInterface                       $shipmentFactory
```
Zales0123: Should be reverted.

src/Sylius/Component/Core/OrderProcessing/OrderShipmentProcessor.php
```diff
@@ -55,7 +56,7 @@ public function process(BaseOrderInterface $order)
         /** @var OrderInterface $order */
         Assert::isInstanceOf($order, OrderInterface::class);
 
-        if ($order->isEmpty()) {
+        if ($order->isEmpty() || ! $this->ifItemsRequireShipping($order)) {
```
Zales0123: Extra space after `!`.

src/Sylius/Component/Core/OrderProcessing/OrderShipmentProcessor.php
```diff
@@ -12,6 +12,7 @@
 namespace Sylius\Component\Core\OrderProcessing;
 
 use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Model\OrderItemInterface;
```
Zales0123: I think it's not needed.

src/Sylius/Component/Core/OrderProcessing/OrderShipmentProcessor.php
```diff
@@ -75,7 +76,7 @@ public function process(BaseOrderInterface $order)
     }
 
     /**
-     * @param BaseOrderInterface $order
+     * @param BaseOrderInterface | OrderInterface $order
```
Zales0123: Should also be reverted.

src/Sylius/Component/Core/spec/OrderProcessing/OrderShipmentProcessorSpec.php
```diff
@@ -44,19 +46,34 @@ function it_is_an_order_processor()
...
+
+        $productVariant->isShippingRequired()->willReturn(true);
+
+        $orderItems->getIterator()->willReturn($iterator);
+
+        $iterator->rewind()->shouldBeCalled();
+        $iterator->valid()->willReturn(true);
+        $iterator->current()->willReturn($orderItem);
+
+        $order->getItems()->willReturn($orderItems);
```
Zales0123: It can be `$order->getItems()->willReturn(new \ArrayIterator([$orderItem->getWrappedObject()]))`. Then, the whole part:
```
$orderItems->getIterator()->willReturn($iterator);

$iterator->rewind()->shouldBeCalled();
$iterator->valid()->willReturn(true);
$iterator->current()->willReturn($orderItem);
```
can be removed.

src/Sylius/Component/Core/spec/OrderProcessing/OrderShipmentProcessorSpec.php
```diff
@@ -71,15 +88,65 @@ function it_creates_a_single_shipment_with_default_shipping_method_and_assigns_a
...
+        $productVariant->isShippingRequired()->willReturn(false);
+
+        $orderItems->getIterator()->willReturn($iterator);
+
+        $iterator->rewind()->shouldBeCalled();
+        $iterator->valid()->willReturn(true);
+        $iterator->current()->willReturn($orderItem);
+        $iterator->valid()->willReturn(false);
+
+        $order->getItems()->willReturn($orderItems);
```
Zales0123: The same here.

src/Sylius/Component/Core/spec/OrderProcessing/OrderShipmentProcessorSpec.php
```diff
@@ -71,15 +88,65 @@ function it_creates_a_single_shipment_with_default_shipping_method_and_assigns_a
...
+        $orderItem->getVariant()->willReturn($productVariant);
+
+        $productVariant->isShippingRequired()->willReturn(true);
+        $orderItems->getIterator()->willReturn($iterator);
+
+        $iterator->rewind()->shouldBeCalled();
+        $iterator->valid()->willReturn(true);
+        $iterator->current()->willReturn($orderItem);
+
+        $order->getItems()->willReturn($orderItems);
```
Zales0123: And here.

src/Sylius/Component/Core/spec/OrderProcessing/OrderShipmentProcessorSpec.php
lchrusciel: The definition of `getItems` also allows us to return just an array, so maybe: `$orderItems->getIterator()->willReturn([$orderItem]);`

src/Sylius/Component/Core/OrderProcessing/OrderShipmentProcessor.php
Persata: I added this because the method uses the `hasShipments` / `getShipments` / `addShipment` functionality of the order model, but those are not part of `Sylius\Component\Order\Model\OrderInterface`.

Can the definition of

```php
private function getOrderShipment(BaseOrderInterface $order)
```

be changed to

```php
private function getOrderShipment(OrderInterface $order)
```

so that the interface expectations align?

src/Sylius/Component/Core/spec/OrderProcessing/OrderShipmentProcessorSpec.php
Persata: Will change these - originally had `[$orderItem]` but saw the use of `$iterator` elsewhere in the tests - I'll change it back :smile:

src/Sylius/Component/Core/OrderProcessing/OrderShipmentProcessor.php
lchrusciel: We have an Assert [at the beginning of process](https://github.com/Sylius/Sylius/pull/8062/files/4caa5f5f99e9fbecefaec5b1501a9bcc9a7ac24d#diff-f20198fd413e52074801c922434669d5R57) so I am ok with an `OrderInterface` typehint in private methods

Zales0123: 👍  ok, but it should definitely be `BaseOrderInterface|OrderInterface $order`, without spaces around `|` 😄

Zales0123: Or we can event remove `BaseOrderInterface` typehint, I think.

Zales0123:
@Persata are you willing to continue works on this PR? There are just small fixes there to do ;)

Persata:
@Zales0123 yes! Sorry, have been so busy with other project work, I'll try and get it re-submitted sometime in the next few days if that's ok?

Persata:
@Zales0123 I've made the requested changes and created a new feature for testing the fix. Can you check them over and let me know if anything else needs done? Thanks!

src/Sylius/Component/Core/OrderProcessing/OrderShipmentProcessor.php
```diff
@@ -104,4 +104,20 @@ private function getOrderShipment(BaseOrderInterface $order)
             return null;
         }
     }
+
+    /**
+     * @param OrderInterface $order
+     *
+     * @return bool
+     */
+    private function ifItemsRequireShipping(OrderInterface $order)
```
pjedrzejewski: This method is duplicated [here](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Checker/OrderShippingMethodSelectionRequirementChecker.php#L71), perhaps we extract it to a service to respect SRP?

Can be done as a separate PR.

pjedrzejewski:
Thank you Ross, awesome work, especially you included Behat scenarios. :) I will open a separate issue for extracting the common logic.

PR #8069, opened 2017-05-18
[RFC] Fix migrations

alcaeus:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | possibly |
| Related tickets | - |
| License         | MIT |

Found this while testing a few things with database migrations. Turns out that migrating the database yields different results than recreating it from scratch.

To verify, it should be possible to set up the database, migrate down to an older version and migrate back to the current version:
```
./bin/console doctrine:database:drop --env=test --force && ./bin/console doctrine:database:create --env=test && ./bin/console doctrine:schema:create --env=test
./bin/console doctrine:migrations:version --add --all --env=test --no-interaction
./bin/console doctrine:migrations:migrate --env=test 20161219160441 --no-interaction
./bin/console doctrine:migrations:migrate --env=test --no-interaction
```

However, this will not work:
```
Migration 20170313125424 failed during Execution. Error An exception occurred while executing 'ALTER TABLE sylius_admin_api_access_token DROP FOREIGN KEY FK_7D83AA7F19EB6921':

SQLSTATE[42000]: Syntax error or access violation: 1091 Can't DROP 'FK_7D83AA7F19EB6921'; check that column/key exists
```

The reason this pull request is marked as RFC is because we might need a solution for people who have already run this migration, as their tables contain wrong names for foreign key relationships. This will inevitably cause trouble when the next migration needs to modify these relationships.


lchrusciel:
I'm not sure about changing existing migration. We should leave it as it and either document that this one should be reverted, or just provide a next one which will revert the state. And of course, create the last one with a proper migration.

The second option is a little bit confusing, but provide better DX, cuz nobody needs to event read this thread or docs. I would vote for that solution.

alcaeus:
There is no good way to fix this as the key names differ depending on how the tables were created:
* People who set up their database (using the `doctrine:schema:create` command) after the change was introduced will have a foreign key with name `FK_2AA4915D19EB6921`.
* People who set up their database before the change was introduced and migrated the database to the current state will have a foreign key with name `FK_7D83AA7F19EB6921 `.

Creating a migration to fix this would cause errors for anyone who set up their database after the change was introduced. This migration would have to be conditional and I'm not sure that's even possible in doctrine-migrations.

alcaeus:
Any suggestions on how to proceed with this?

pjedrzejewski:
@pamil What do you think?

wadjeroudi:
@alcaeus the condition should be possible by running a "SHOW CREATE TABLE sylius_admin_api_access_token;" and compare result string with the two names for the foreign key ?

pamil:
Anyway, do we know why these generated foreign keys are different? If we don't solve the root cause and just cure the symptoms, then either we will have a tough time monitoring and testing these irregularities or it will misalign again soon.

The most stable solution for now would be to exclusively use migrations to set up the schema (this would need a change in our installer). What do you think?

alcaeus:
@pamil the most stable solution would be to fix the database for all people that have wrongly named keys. This came up after I migrated the database and ran `doctrine:migrations:diff` on the migrated database. Changing the installer to run the migrations will still produce an undesirable result: the migrated database differs from a fresh installation. Ideally, installing an old database version and running migrations produces the same schema as if the current database was installed.

IMO, the only fix is to conditionally apply the migration if the foreign keys have the wrong name. I'll check whether this could be done by querying the database as @wadjeroudi suggested.

alcaeus:
@pamil Created a conditional migration. The migration in question checks whether a single correct foreign key name appears in one of the affected tables. If it does, the migration is skipped, marking it as executed without executing any queries.
If the key in question is not found, it assumes the database was migrated to the current state rendering invalid key names and corrects them.

Since the original migration was fixed, the down path does not change anything. However, if you'd rather not change the original migration, the wrong keynames could be reapplied in this migration as well. However, I strongly suggest fixing the original migration and conditionally fixing the wrong keynames.

pjedrzejewski:
Looks good to me, I will do some additional testing before tagging the release. Thank you so much @alcaeus! Nice work! 👍

PR #8070, opened 2017-05-18
Fix address log entry column names

alcaeus:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | - |
| License         | MIT |

This uses snake_cased column names for the AddressLogEntry entity, which was missing `column` declaration in mappings. This can cause issues when people change the default naming strategy for ORM and is the only entity that didn't have `column` properties in field mappings.

### DIFF
--- a/app/migrations/Version20170518123056.php
+++ b/app/migrations/Version20170518123056.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Sylius\Migrations;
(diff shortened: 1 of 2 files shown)


pamil:
Thanks Andreas!

PR #8075, opened 2017-05-19
Add missing checkout complete validation group

gruberro:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | ~ |
| License         | MIT |

This fix introduces the missing checkout complete validation group to the finalize step. This is to ensure event data passed through the admin api is validated according to the application requirements.

### DIFF
--- a/tests/Responses/Expected/checkout/complete_validation_failed_disabled_product.json
+++ b/tests/Responses/Expected/checkout/complete_validation_failed_disabled_product.json
@@ -1,0 +1,12 @@
+{
+    "code": 400,
+    "message": "Validation Failed",
+    "errors": {
+        "errors": [
+            "This product Mug has been disabled."
+        ],
+        "children": {
+            "notes": {}
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


lchrusciel:
@gruberro Thanks for this fix, it is must have for us :) I'm just thinking about some tests for that to avoid regression. Can you add a test in `CheckoutCompleteApiTest` where you would break the checkout? For example, you could go through the whole checkout, and before `complete` step you could disable a product that you are trying to buy(Either through a repository or by API call). Then, it should not be possible to finish checkout, but it will be without this `validation_group`.

gruberro:
@lchrusciel I'll add some tests! Did quite the same already on my current project 😄

gruberro:
@lchrusciel added some tests regarding disabled products and insufficient stock. Anything else that I've missed?

tests/Responses/Expected/checkout/complete_validation_failed_disabled_product.json
```diff
@@ -0,0 +1,12 @@
+{
+    "code": 400,
+    "message": "Validation Failed",
+    "errors": {
+        "errors": [
+            "This product Mug has been disabled."
```
lchrusciel: We should iterate over this error message :)

lchrusciel:
@gruberro It would be enough. Thanks, Robert! :)

PR #8081, opened 2017-05-22
[Ui] Re-add @SyliusUi/Grid/Field/money.html.twig with corrected filter

### LINKED ISSUE #8080: @SyliusUi/Grid/Field/money.html.twig uses non existing twig filter
I wanted to use a formatted money number in a custom grid by configuring:
```yml
sylius_grid:
    grids:
        app_admin_my_grid:
            driver:
                name: doctrine/orm
                options:
                    class: "%app.model.entity.class%"
            filters:
                search:
                    type: string
                    label: sylius.ui.search
                    options:
                        fields: [someField]
            fields:
                someFieldThatRepresentsMoney:
                    type: twig
                    label: Label
                    position: 1
                    options:
                        template: "@SyliusUi/Grid/Field/money.html.twig"
            actions:
                main:
                    create:
                        type: create
                item:
                    update:
                        type: update
                    delete:
                        type: delete
```

Then I got the error: `Unknown "sylius_money" filter.` Because apparently this twig filter is replaced for sylius_format_money. So '@SyliusUi/Grid/Field/money.html.twig' needs to be corrected

Shall I make a PR to correct this?

reyostallenberg:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | depends (exists in latest beta, removed in master) |
| BC breaks?      | no |
| Related tickets | fixes #8080 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/views/Grid/Field/money.html.twig
+++ b/src/Sylius/Bundle/UiBundle/Resources/views/Grid/Field/money.html.twig
@@ -1,0 +1,1 @@
+{{ data|sylius_format_money(sylius.currencyCode, sylius.localeCode) }}


Zales0123:
It seems nice, but there is one big question for me - is it used anywhere? If not, then it's also not tested and it's just a dead code, which we should no maintain 😕

lchrusciel:
I guess twig validation is not working properly. I cannot find `Validating (Twig templates)` header in travis log. Also, this PR should fail. /cc @pamil

pamil:
Why should it fail? Anyway, this PR was made before we've introduced Twig validation.

reyostallenberg:
@Zales0123 
> It seems nice, but there is one big question for me - is it used anywhere? If not, then it's also not tested and it's just a dead code, which we should no maintain :confused:

I understand that. And it's fine with me to close the PR. (I have done my own implementation in my app)

I was working on a specific implementation in which I needed money represented in a grid. So I included this template, which was in the codebase of Sylius at that time, that's why I stated 'New feature: depends (exists in latest beta, removed in master)'.

When I included it it gave me an error and I fixed this. At the time of creating the PR I noticed it was gone in master.

I think it is nice to have such utility for a developer wanting to add money representation in the grid. I understand the con of having dead code (but that's from the point of Sylius, not from the point of DX). 
 Maybe we could add a documentation note about this?

lchrusciel:
@pamil AFAIR lint will fail if we will have an unused template.

Anyway, I'm voting for some small example in docs rather than dead code. Or another plugin with an own testing system and a bunch of useful helpers(if we have anything else than this one).

PR #8090, opened 2017-05-23
Add alt attribute to main product image

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no|
| License         | MIT |

ALT tag was empty, but we can easily insert the product name.


lchrusciel:
Thanks, Stefan! :)

teohhanhui:
`alt` is supposed to represent what the image is about. Putting the wrong `alt` would only cause confusion.

PR #8093, opened 2017-05-23
[Order] Fixed sylius:remove-expired-carts help

sweoggy:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

Will surely save some frustration :)

### DIFF
--- a/src/Sylius/Bundle/OrderBundle/Command/RemoveExpiredCartsCommand.php
+++ b/src/Sylius/Bundle/OrderBundle/Command/RemoveExpiredCartsCommand.php
@@ -26,7 +26,8 @@ class RemoveExpiredCartsCommand extends ContainerAwareCommand
     {
         $this
             ->setName('sylius:remove-expired-carts')
-            ->setDescription('Removes carts that have been idle for a configured period. Configuration parameter - sylius_order.cart_expires_after.');
+            ->setDescription('Removes carts that have been idle for a period set in `sylius_order.expiration.cart` configuration key.')
+        ;
     }
 
     /**


src/Sylius/Bundle/OrderBundle/Command/RemoveExpiredCartsCommand.php
```diff
@@ -27,7 +27,7 @@ protected function configure()
     {
         $this
             ->setName('sylius:remove-expired-carts')
-            ->setDescription('Removes carts that have been idle for a configured period. Configuration parameter - sylius_order.cart_expires_after.');
+            ->setDescription('Removes carts that have been idle for a configured period. Configuration parameter - sylius_order.cart_expiration_period.');
```
lchrusciel: I think it would be better to point to the configuration structure where it can be defined: 
```yml
sylius_order:
    expiration:
        cart: 1 day
```

So `sylius_order.expiration.cart`

sweoggy: Well I guess that's a matter of taste. I prefer to refer to the parameter as it is a much quicker way (1 line instead of 3) to set the same thing.

If we were to refer to the configuration structure we should also change CancelUnpaidOrdersCommand.

pamil: @sweoggy in the end, only the bundle configuration would be the one under BC promise

pamil:
Thanks, Oscar! 🎉

Issue #7838, opened 2017-03-27
[BC Break] Unify naming of tag attributes


mvar:
Current Sylius configuration:

```yml
services:
    foo.grid.filter.bar:
        class: FooBundle\Grid\Filter\BarFilter
        tags:
            -
                name: sylius.grid_filter
                type: bar
                form-type: 'FooBundle\Form\Type\Filter\BarType'
```

Please note `form-type`.

In Symfony all attributes follows underscore naming convention. As Sylius is highly dependent on Symfony, it would be nice to see the same coding style too. I suggest to find and replace all similar places before first RC.

mvar:
@pjedrzejewski won't happen?

lchrusciel:
@mvar are you willing to open a PR with this fix?

mvar:
@lchrusciel I have opened PR with updated code and docs. I haven't added any notes to UPGRADE doc, I'm not sure if I should.

pamil:
Merged in #8096. :)

PR #8096, opened 2017-05-24
Use underscores for tag attribute names

mvar:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | mentioned in #7838 |
| License         | MIT |

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,11 @@
 # UPGRADE FROM 1.0.0-beta.2 to 1.0.0
 
+* The following tag attributes were renamed in order to keep consistency:
+
+  * from `form-type` to `form_type`.
+  * from `attribute-type` to `attribute_type`.
+  * from `configuration-form-type` to `configuration_form_type`.
+
 ## Packages:
 
 ### PayumBundle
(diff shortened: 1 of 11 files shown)


lchrusciel:
Please, add also a note in UPGRADE.md

mvar:
Updated README.md. Added info to the main block as it effect 3 or 4 bundles.

mvar:
@lchrusciel it's OK to use dashes in XML (and looks more XML'ish). Then you use dashes you get both versions in the compiler pass. `form-type` and `form_type` will be set as attribute with the same value.

lchrusciel:
What kind of sorcery is that :D I'm not sure about it either. Simple example: newcomer would like to check where this definition is parsed and will look for `form-type` but he will not find anything, cuz compiler pass operates on `form_type`. Anyway, I'm ok with that if others will agree. /cc @pamil @Zales0123 @Arminek

pamil:
Thanks @mvar! 🎉

pjedrzejewski:
👍 👍 👍

PR #8097, opened 2017-05-24
Use codes of option value & option instead of ID in variant generator

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | no|
| BC breaks?      | no|
| License         | MIT |

Use case: I wanted to use it in an unit test, but it's based around the entities ID's and as I didn't want to persist them, there is no way (as I see it) to assign an ID to easily. That makes it harder to use it in stand-alone situations. 

The `code` field of the entities is also a required field, which should be there if there is also an ID, but using `code` allows to use setters to use this generator stand-alone for such use-cases.


pjedrzejewski:
Thanks Stefan!

PR #8117, opened 2017-05-29
Drop support for PHP 5.6 / 7.0

### LINKED ISSUE #8114: Dropping support for PHP 5.6 / 7.0
Hi folks!

After recent comments seen in issues (like #8107), core team discussions and many open-source projects requiring PHP 7+ in their next major releases, we came to the conclusions that it might be worth to consider dropping support for PHP 5.6 / 7.0.

Facts:
 - **PHP 5.6**: no active support, security support ends on December 2018
 - **PHP 7.0**: active support ends on December 2017, security support ends on December 2018
 - **PHP 7.1**: active support ends on December 2018, security support ends on December 2019
 - **Symfony 4.0**: to be released on November this year, meant to support PHP 7.1
 - **Doctrine 3.0**: meant to support PHP 7.2
 - **Sylius 1.0**: to be released this year, next major release in 2019, LTS versions with 4 years of security support (as #8107 mentions).

In addition to these, PHP 5.6 is a lot slower and makes quite a bit of weird false-positives on Travis CI. By introducing PHP 7.1 we could not only make use of new features (like `iterable` typehint, return typehints, scalar typehints) but also align Sylius with other open-source projects and libraries, while having active support for PHP issues.

After this RFC is settled we should also consider introducing some kind of process for adapting the new PHP releases.

We can vote using reactions on this message, 👍  for supporting PHP 7.1 only, 😆  for supporting PHP 7.0+, 👎  for supporting PHP 5.6+. Comments are appreciated for sharing your own experiences and thoughts! :)

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | kind of |
| Related tickets | closes #8114 |
| License         | MIT |

As everyone who voted in #8114 is currently for supporting PHP 7.1 only (let's wait a few more days anyway), here's a bare minimum to get us started.


src/Sylius/Bundle/UiBundle/Twig/SortByExtension.php
```diff
@@ -48,8 +48,7 @@ public function sortBy($array, $field, $order = 'ASC')
             return $array;
         }
 
-        /** "@usort" so it won't explode on php 5.6 */
-        @usort($array, function ($firstElement, $secondElement) use ($field, $order) {
```
pamil: 🎉

FanchTheSystem:
Hello! 

Is this will be merged this summer ?

stefandoorn:
As PHP 7.2 is getting close, should it also be tested against PHP 7.2 already and maybe even nightlies for PHP 7.3? 

This run is e.g. already on PHP 7.2: https://travis-ci.org/doctrine/common/builds/251602714

pamil:
@stefandoorn sure, we just need to get this one passing first 🎉

pjedrzejewski:
Awesome, thanks Kamil!

alcaeus:
@pamil out of curiosity, what's the motivation behind locking doctrine/orm to `~2.5.0? 2.6 will be a backwards compatible update and `dev-master` is only eligible if you specify a `minimum-stability` of `dev` in the root composer.json.

pamil:
When a Sylius package requires another Sylius package, we need both of them not to be retrieved from Packagist, but from the current codebase. That's why there's `../../*/*` path repository is defined and `minimum-stability` set to `dev`. If there's any better way to achieve it, let me know :)

As for the reason of this change, we already have `count()` method in `OrderRepository` which extends `EntityRepository` which introduces `count(array $criteria)` in 2.6.

alcaeus:
> That's why there's `../../*/*` path repository is defined and `minimum-stability` set to `dev`.

I didn't see that, thanks for explaining 👍 

As for `count()`, you could change the signature to `count(array $criteria = [])` to be forwards compatible to 2.6 and have a minor BC break before 1.0.0 😉

pamil:
Trying to add that optional parameter but actually having a trouble with it as we need to filter out orders with state `OrderInterface::STATE_CART` (`order.state != OrderInterface::STATE_CART`). It would be easy check for *equals*, but seems more complicated when it comes to *not equals*.

The other solution might be to rename this method to something like `countPlacedOrders()`.

alcaeus:
`countPlacedOrders` would be even better than an arbitrary `count`

PR #8119, opened 2017-05-30
Fix wrong assertion which prevented default admin user creation

stevemoreau:
Fix wrong assertion which prevented default admin user creation (sylius@example.com / sylius) when installing with '--no-interaction' option

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |


lchrusciel:
Thanks, @stevemoreau! :)

PR #8121, opened 2017-05-30
[BugFix] Make sure query parameters are retained properly when filtering.

koemeet:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

Without these changes, some filter states are lost due to improper merging of the query parameters. If you would for example filter on 50 products per page and change the order to display the products, the limit is for example lost. These changes fix this.


koemeet:
@lchrusciel I am very low on free time in coming weeks/months, so I don't think I am able to build a test scenario around this any time soon 😢 .

pamil:
I'm for merging it even without tests, as the current state is *buggy behaviour and no tests* and this PR makes it *correct behaviour and no tests* ;)

pjedrzejewski:
@pamil Agreed, thank you Steffen!

Issue #8152, opened 2017-06-12
Non-realtime province name has wrong label


silverbackdan:
in Sylius\Bundle\AddressingBundle\Form\EventListener\BuildAddressFormSubscriber

The 'createNamed' method is missing the label and placeholder options:

Options should be the same as in Sylius\Bundle\AddressingBundle\Controller\ProvinceController
```php
/**
     * @param CountryInterface $country
     *
     * @return FormInterface
     */
    protected function createProvinceChoiceForm(CountryInterface $country)
    {
        return $this->get('form.factory')->createNamed('sylius_address_province', ProvinceCodeChoiceType::class, null, [
            'country' => $country,
            'label' => 'sylius.form.address.province',
            'placeholder' => 'sylius.form.province.select',
        ]);
    }

    /**
     * @return FormInterface
     */
    protected function createProvinceTextForm()
    {
        return $this->get('form.factory')->createNamed('sylius_address_province', TextType::class, null, [
            'required' => false,
            'label' => 'sylius.form.address.province',
        ]);
    }
```

PR #8153, opened 2017-06-12
Use appropriate labels and placeholders

silverbackdan:
When the form is loaded without ajax on the form post, the label should be correct immediately. See https://github.com/Sylius/Sylius/issues/8152

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/8152, partially #Y, mentioned in #Z |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Form/EventListener/BuildAddressFormSubscriber.php
+++ b/src/Sylius/Bundle/AddressingBundle/Form/EventListener/BuildAddressFormSubscriber.php
@@ -137,6 +137,8 @@ final class BuildAddressFormSubscriber implements EventSubscriberInterface
         return $this->formFactory->createNamed('provinceCode', ProvinceCodeChoiceType::class, $provinceCode, [
             'country' => $country,
             'auto_initialize' => false,
+            'label' => 'sylius.form.address.province',
+            'placeholder' => 'sylius.form.province.select',
         ]);
     }
 
@@ -150,6 +152,7 @@ final class BuildAddressFormSubscriber implements EventSubscriberInterface
         return $this->formFactory->createNamed('provinceName', TextType::class, $provinceName, [
             'required' => false,
             'auto_initialize' => false,
+            'label' => 'sylius.form.address.province',
         ]);
     }
 }


src/Sylius/Bundle/AddressingBundle/Form/EventListener/BuildAddressFormSubscriber.php
```diff
@@ -137,6 +137,8 @@ private function createProvinceCodeChoiceForm(CountryInterface $country, $provin
         return $this->formFactory->createNamed('provinceCode', ProvinceCodeChoiceType::class, $provinceCode, [
             'country' => $country,
             'auto_initialize' => false,
+            'label' => 'sylius.form.address.province',
```
pamil: Are these really needed? These are the default values, both for `label` and `placeholder` (ProvinceCodeChoiceType is a child of ProvinceChoiceType defining those).

silverbackdan: When I define sylius.form.address.province to 'County' in a messages.yml, it will populate the ajax loaded province input, but the non-ajax loaded input still is labelled as 'Province Name'

Are you saying there's somewhere else I should be defining the labels instead of in the translation files?

The only reason I put this PR in is because these are the options on the inputs when loaded via ajax.

Perhaps it isn't needed for the province dropdown, my issue was only with the text input so it may be the problem is best fixed elsewhere.

pamil: @silverbackdan I guess the issue you encountered is fixed by adding label for `provinceName` field created in `createProvinceNameTextForm` (which is also done in this PR). Can you check if it is still fixed without this changes here?

silverbackdan: @pamil I had only extended it to the dropdown as well because I noticed they hadn't been defined in there either when they had been in the class handling realtime province field updates. The dropdown part is untested, only assumed. My issue is indeed fixed just by adjusting `createProvinceNameTextForm`

pamil:
Thanks @silverbackdan! 🎉

PR #8156, opened 2017-06-12
Add missing permission flag to the admin routes for the authorization check to actually work

DmitriyTrt:
While implementing the `Sylius\Bundle\ResourceBundle\Controller\AuthorizationCheckerInterface` we found out that many routes from the admin bundle miss the `permission` flag and so the authorization check is never called for these routes ([see source](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php#L484)). This is kind of security issue when implementing privileges separation for different roles of admins.

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |


lchrusciel:
Thanks, @DmitriyTrt! :)

PR #8170, opened 2017-06-16
Validate API url with pattern ^/api/.* not just ^/api

krafas:
Now if we want to have url which starts with `/api` we have access denied exception due to /api firewall. Also where is similar issue with FOS FormatListener, etc.

Before 

| url | http code |
|---|----|
| /api/something | 401 |
| /apiary | 401 |

After

| url | http code |
|---|----|
| /api/something | 401 |
| /apiary | 200 |

### DIFF
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -4,7 +4,7 @@
 parameters:
     sylius.security.admin_regex: "^/admin"
     sylius.security.api_regex: "^/api"
-    sylius.security.shop_regex: "^/(?!admin|api)[^/]++"
+    sylius.security.shop_regex: "^/(?!admin|api/.*|api$)[^/]++"
 
 security:
     providers:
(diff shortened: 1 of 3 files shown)


pjedrzejewski:
Looks good to me.

app/config/security.yml
```diff
@@ -4,7 +4,7 @@
 parameters:
     sylius.security.admin_regex: "^/admin"
     sylius.security.api_regex: "^/api"
-    sylius.security.shop_regex: "^/(?!admin|api)[^/]++"
+    sylius.security.shop_regex: "^/(?!admin|api/.*|api$)[^/]++"
```
lchrusciel: Do we need this third condition(`api$`)?

lchrusciel:
Thanks Giedrius!

PR #8175, opened 2017-06-17
serializer config for relations moved to AdminApiBundle

### LINKED ISSUE #8171: AddressingBundle depends on AdminApiBundle
In our custom installation we only extend CoreBundle, having all frontend bundles custom (admin, shop). Default serializer tries to serialize Order resource, which boils down to serializing Zone model, which has hardcoded route sylius_admin_api_zone_show for Hateoas relation.

I propose moving relations section of serializer configuration from core bundles to bundle which defines the route, in this case AdminApiBundle. I'd be glad to submit PR upon approval.

MichaelKubovic:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8171 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/AdminApiBundle/Resources/config/app/config.yml
+++ b/src/Sylius/Bundle/AdminApiBundle/Resources/config/app/config.yml
@@ -24,3 +24,12 @@ jms_serializer:
             sylius-core:
                 namespace_prefix: "Sylius\\Component\\Core"
                 path: "@SyliusAdminApiBundle/Resources/config/serializer"
+            sylius-addressing:
+                namespace_prefix: "Sylius\\Component\\Addressing"
+                path: "@SyliusAdminApiBundle/Resources/config/serializer"
+            sylius-currency:
+                namespace_prefix: "Sylius\\Component\\Currency"
+                path: "@SyliusAdminApiBundle/Resources/config/serializer"
+            sylius-locale:
+                namespace_prefix: "Sylius\\Component\\Locale"
+                path: "@SyliusAdminApiBundle/Resources/config/serializer"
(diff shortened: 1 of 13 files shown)


pjedrzejewski:
@lchrusciel Wasn't that fixed in your PR? :)

src/Sylius/Bundle/AdminApiBundle/Resources/config/app/config.yml
```diff
@@ -24,3 +24,12 @@ jms_serializer:
             sylius-core:
                 namespace_prefix: "Sylius\\Component\\Core"
                 path: "@SyliusAdminApiBundle/Resources/config/serializer"
+            sylius-addressing:
+                namespace_prefix: "Sylius\\Component\\Addressing"
```
lchrusciel: This solution will not work because you have overridden the previous serialization config. We need to either move the whole mapping for these classes to AdminApiBundle, create children class, just to add some mapping(which sucks) or you can check if it will work if you will change a key for this config(`sylius-addressing` -> `sylius-admin-api-addressing` or something similar)

Zales0123:
@lchrusciel @MichaelKubovic what is the status of this PR? Is it still relevant and need to fixed and merged?

MichaelKubovic:
I honestly don't remember what I was trying to fix here and I don't have any active sylius project at the moment where I could test @lchrusciel 's suggestions.

Zales0123:
@lchrusciel do you think we can do something with this PR?

lchrusciel:
Hello Michael!

Thanks a lot for your effort and I'm sorry this PR was not merged. It was proposed a long time ago and with the current refactoring of the whole API approach, I'm not sure, if there is a point to keep it open. Feel free to open it again, if you find it appropriate.

PR #8186, opened 2017-06-20
Disable aliasing serializer service to jms_serializer.serializer

### LINKED ISSUE #8163: enable_short_alias: false with JMS Serializer
Are there any issues with setting `enable_short_alias` option to `false` in JMS Serializer?

When enabled, this option aliases `@serializer` service to `@jms_serializer.serializer`, but the interfaces are not compatible between each other, so our custom app which uses `@serializer` crashes, since the incompatible implementation is being injected into our classes.

If Sylius does not depend on `enable_short_alias: true`, how would you feel about a pull request setting it to `false` by default?

emodric:
| Q               | A
| --------------- | ---
| Bug fix?        | kinda |
| New feature?    | no |
| BC breaks?      | no ? |
| Related tickets | fixes #8163 |
| License         | MIT |

As per instructions by @lchrusciel in #8163, opening this to allow the community to give their opinion :)

### DIFF
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -47,6 +47,11 @@ swiftmailer:
     username: "%mailer_user%"
     password: "%mailer_password%"
 
+# Temporary fix to disable aliasing @serializer service to JMS variant
+# TODO: Remove when upgrading to JMS Serializer Bundle 2.0
+jms_serializer:
+    enable_short_alias: false
+
 fos_rest:
     exception: ~
     view:


mbabker:
FWIW JMSSerializerBundle also removed this in their 2.0 release - https://github.com/schmittjoh/JMSSerializerBundle/issues/558

lchrusciel:
I guess that actually resolves our problem 👍

lchrusciel:
@emodric can you upgrade jms bundle to 2.0?

emodric:
@lchrusciel Done!

emodric:
@lchrusciel Looks like this doesn't work. Upgrading JMS Serializer Bundle to 2.0 required bumping Serializer lib itself to 1.7, which might be the reason for the failures?

lchrusciel:
Ok, it appears that it will require more work. Something doesn't work with old serialization... Maybe we can work with your previous fix and some `TODO` comment and open an issue with the upgrade.

emodric:
I've restored the original patch and added a TODO comment. I'll open a new issue after this is merged :)

emodric:
@lchrusciel Looks okay now (except PHP 5.6, but as I understand this is a known issue).

lchrusciel:
I have restarted the build. Let's wait for the green and it is good to go.

lchrusciel:
Thanks, Edi!

emodric:
Thank you Łukasz!

I've opened an issue for upgrading JMS Serializer Bundle to 2.0 (#8193)

PR #8190, opened 2017-06-21
[ThemeBundle] Fix support for custom Twig namespaces

### LINKED ISSUE #8189: [ThemeBundle] Exception when using custom Twig namespaces
When I use a custom Twig namespace `@my_app` (as per http://symfony.com/doc/current/templating/namespaced_paths.html#registering-your-own-namespaces), I get an exception:

`Bundle "my_appBundle" does not exist or it is not enabled. Maybe you forgot to add it in the registerBundles() method of your AppKernel.php file?`

`TemplateNameParser` from Sylius Theme Bundle is the culprit here, assuming that `@my_app` Twig namespace must always correspond to a bundle.

The issue also comes to light with eZ Platform integration and recent addition of "virtual" Twig namespace called `@ezdesign` to eZ Platform.

emodric:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8189 |
| License         | MIT |

Fixes support for custom Twig namespaces (http://symfony.com/doc/current/templating/namespaced_paths.html#registering-your-own-namespaces)

I don't know if this should be done this way, but the patch works fine.

### DIFF
--- a/src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php
+++ b/src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php
@@ -76,7 +76,7 @@ final class TemplateNameParser implements TemplateNameParserInterface
             try {
                 $this->kernel->getBundle($template->get('bundle'));
             } catch (\Exception $e) {
-                throw new \InvalidArgumentException(sprintf('Template name "%s" is not valid.', $name), 0, $e);
+                return $this->decoratedParser->parse($name);
             }
         }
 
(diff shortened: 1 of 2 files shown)


emodric:
@pamil I've updated the test too, but since I've never worked with PhpSpec, do tell if something is missing :)

pamil:
@emodric looks good to me! PHP 5.6 is just randomly failing, restarted the build.

emodric:
@pamil Everything looks good, going in?

lchrusciel:
Thanks, Edi! :)

emodric:
Thank you :)

pjedrzejewski:
👍

PR #8191, opened 2017-06-21
[UX][Admin] Changed admin titles to include Sylius as postfix

sweoggy:
Instead of prefix, this is better UX. Knowing where you are in the admin
panel is more important to users than knowing what is powering it.

| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no (Well yes if someone relied on page titles...) |
| Related tickets |  |
| License         | MIT |


sweoggy:
Re-open to trigger build (PHP 5.6 was failing as usual)

lchrusciel:
Seems legit. Thanks, Oscar! :)

PR #8200, opened 2017-06-25
[Behat][AddressBook] Added missing assertion

Tykzz:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | none |
| License         | MIT |

Assertion for containing a string in an address was missing which caused the step `it should contain` to always pass.


lchrusciel:
Thanks, Krzysztof!

PR #8215, opened 2017-06-28
[Checkout] decoupled empty order logic from ShopBundle

### LINKED ISSUE #8185: [Core] Checkout logic is coupled to shop bundle
In CheckoutResolver at line 85 we are referring to [the shop path](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Checkout/CheckoutResolver.php#L85). The easiest solution for 3rd party users would be to just create this route or override this class. But we should fix it anyway.

bitbager:
| Q               | A
| --------------- | ---
| Bug fix?        |yes |
| New feature?    | no|
| BC breaks?      | no|
| Related tickets | fixes #8185  |
| License         | MIT |


lchrusciel:
Pretty nice first contribution! 🎉  Thanks, Mikołaj!

bitbager:
You're welcome! Hope I will have more time for other PRs :)

PR #8221, opened 2017-06-29
[Taxonomy] remove locale dependency for admin panel in TaxonRepository

### LINKED ISSUE #8130: Can't add a new product in a taxon
Hi,

I have been trying to add a new product in a category , but I can't see the sub categories under category.

On the sylius demo its looks as below:

sylius1

In my back-end its look like this:

sylius2

I'm using sylius 1.0.0-beta.2

bitbager:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8130 |
| License         | MIT |

### DIFF
--- a/features/product/managing_products/select_taxons_for_product.feature
+++ b/features/product/managing_products/select_taxons_for_product.feature
@@ -10,6 +10,7 @@ Feature: Select taxon for a product
         And the store has a "T-Shirt Banana" configurable product
         And the store has a product "T-Shirt Batman"
         And I am logged in as an administrator
+        And I am using "English (GB)" locale for my panel
 
     @ui @javascript
     Scenario: Specifying main taxon for configurable product
--- a/src/Sylius/Bundle/AdminBundle/Controller/NotificationController.php
+++ b/src/Sylius/Bundle/AdminBundle/Controller/NotificationController.php
@@ -23,7 +23,7 @@ use Symfony\Component\HttpFoundation\Request;
 /**
  * @author Jan Góralski <jan.goralski@lakion.com>
  */
-class NotificationController
+final class NotificationController
 {
     /**
      * @var ClientInterface
@@ -81,6 +81,8 @@ class NotificationController
             return JsonResponse::create('', JsonResponse::HTTP_NO_CONTENT);
         }
 
-        return JsonResponse::fromJsonString($hubResponse->getBody()->getContents());
+        $hubResponse = json_decode($hubResponse->getBody()->getContents(), true);
+
+        return new JsonResponse($hubResponse);
     }
 }
--- a/src/Sylius/Bundle/AdminBundle/Controller/NotificationController.php
+++ b/src/Sylius/Bundle/AdminBundle/Controller/NotificationController.php
@@ -23,7 +23,7 @@ use Symfony\Component\HttpFoundation\Request;
 /**
  * @author Jan Góralski <jan.goralski@lakion.com>
  */
-final class NotificationController
+class NotificationController
 {
     /**
      * @var ClientInterface
@@ -81,8 +81,6 @@ final class NotificationController
             return JsonResponse::create('', JsonResponse::HTTP_NO_CONTENT);
         }
 
-        $hubResponse = json_decode($hubResponse->getBody()->getContents(), true);
-
-        return new JsonResponse($hubResponse);
+        return JsonResponse::fromJsonString($hubResponse->getBody()->getContents());
     }
 }
--- a/src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
+++ b/src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
@@ -11,6 +11,7 @@
 
 namespace Sylius\Bundle\TaxonomyBundle\Doctrine\ORM;
 
+use Doctrine\ORM\QueryBuilder;
 use Sylius\Bundle\ResourceBundle\Doctrine\ORM\EntityRepository;
 use Sylius\Component\Taxonomy\Repository\TaxonRepositoryInterface;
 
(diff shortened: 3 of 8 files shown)


src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
```diff
@@ -110,4 +110,24 @@ public function createListQueryBuilder()
     {
         return $this->createQueryBuilder('o')->leftJoin('o.translations', 'translation');
     }
+
+    /**
+     * @param QueryBuilder $qb
+     * @param null $locale
```
lchrusciel: `string|null`

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
```diff
@@ -110,4 +110,24 @@ public function createListQueryBuilder()
...
+     * @return QueryBuilder
```
lchrusciel: Missing blank line before return

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
```diff
@@ -110,4 +110,24 @@ public function createListQueryBuilder()
...
+     */
+    private function createTranslationQueryBuilderPart(QueryBuilder $qb, $locale = null)
+    {
+        if ($locale !== null) {
+            $qb
```
lchrusciel: I know, that `qb` is pretty obvious, but we are not allowing abbreviation as variable names. Please, change to `queryBuilder` here and above.

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
bitbager: Is the order important? IMHO null|string is more suitable as null is the default argument value.

bitbager:
@lchrusciel, thank you for notes. I committed changes you suggested. I am writing some description and thoughts of what I've done and what I was thinking on yesterday's evening below.

The issue appeared because the user preferences allow an admin to change his locale for his admin panel.
What I've done is I made the taxonomy queries used in admin product form not depend on the current app locale. Because there are other usages of those queries in SyliuShopBundle, I decided to make the $locale parameter optional and not required for queries used in admin product view.

I also added the proper background to a feature which checks the product taxonomy association process for the GB English instead of US one because from the translation point of view, they are equal (it was required for not messing up with notification checks), but from the query point of view this are two different languages - `en_GB !== en_US`. 

There is one thing that needs to be considered - what I've done maybe just a workaround for an issue that may also appear in another place in admin. It might be that the same happens for any other admin query which has injected locale via `expr:service('sylius.context.locale').getLocaleCode()` expression. 

Maybe it would be better to modify the getLocaleCode method in a way it returns default locale for admin panel used for query injection?

stefandoorn:
I experienced the same problem I think, described in #8225. @bitbager If you adjust the title of this PR, it's more clear what it's about :)

lchrusciel:
Can you rebase to the master? Failing build should be fixed.

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
```diff
@@ -83,24 +85,23 @@ public function findRootNodes()
             ->addOrderBy('o.position')
             ->getQuery()
             ->getResult()
-        ;
+            ;
```
lchrusciel: Should not be changed.

src/Sylius/Bundle/AdminBundle/Controller/NotificationController.php
```diff
@@ -23,7 +23,7 @@
 /**
  * @author Jan Góralski <jan.goralski@lakion.com>
  */
-class NotificationController
+final class NotificationController
```
lchrusciel: Can you separate this PR from #8227?

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
lchrusciel: Still not fixed

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
```diff
@@ -110,4 +111,25 @@ public function createListQueryBuilder()
...
         return $this->createQueryBuilder('o')->leftJoin('o.translations', 'translation');
     }
+
+    /**
+     * @param QueryBuilder $queryBuilder
+     * @param null|string $locale
+     *
+     * @return QueryBuilder
+     */
+    private function createTranslationQueryBuilderPart(QueryBuilder $queryBuilder, $locale = null)
```
pamil: To be correct it does not create translation query builder part, but it configures that part on top of existing query builder.

features/product/managing_products/select_taxons_for_product.feature
```diff
@@ -10,6 +10,7 @@ Feature: Select taxon for a product
         And the store has a "T-Shirt Banana" configurable product
         And the store has a product "T-Shirt Batman"
         And I am logged in as an administrator
+        And I am using "English (GB)" locale for my panel
```
pamil: Does it translate to `en_GB` correctly?

Edit: Nope, `English (United Kingdom)` should be used.

pamil: Made #8248 to avoid this kind of issues in the future.

PR #8228, opened 2017-07-01
[Documentation][Resource] Fix the form options configuration in docs

nlzet:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | / |
| License         | MIT |

I believe this is an error in the docs. The `validation_groups` (and other form options) need to be defined under a key `options` instead of directly under `form`.

(See: https://github.com/Sylius/Sylius/blob/v1.0.0-beta.2/src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php#L149)

### DIFF
--- a/docs/bundles/SyliusResourceBundle/create_resource.rst
+++ b/docs/bundles/SyliusResourceBundle/create_resource.rst
@@ -88,7 +88,8 @@ Below you can see the usage for specifying custom options, in this case, ``valid
             _sylius:
                 form:
                     type: app_book_custom
-                    validation_groups: [sylius, my_custom_group]
+                    options:
+                        validation_groups: [sylius, my_custom_group]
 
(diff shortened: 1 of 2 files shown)


lchrusciel:
Can you rebase to the master? Failing build should be fixed.

nlzet:
@lchrusciel rebased, the checks are passing now.

lchrusciel:
Thanks Niels! I'm happy to see a new contributor :)

nlzet:
@lchrusciel no problem! Currently not using Sylius as a whole, but using some of the components like resource/grid standalone. I think there were some more things I ran into while using/reading the resource bundle docs so there will probably be some more PR's and (or) issues ;)

PR #8245, opened 2017-07-04
[Admin] add current Sylius version checker for admin panel notification

### LINKED ISSUE #8108: Sylius/GUS broken comparison logic.
There is no comparison logic on Sylius side so actually notification widget will always display that new version of sylius is available.

bitbager:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8108 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Controller/NotificationController.php
+++ b/src/Sylius/Bundle/AdminBundle/Controller/NotificationController.php
@@ -23,7 +23,7 @@ use Symfony\Component\HttpFoundation\Request;
 /**
  * @author Jan Góralski <jan.goralski@lakion.com>
  */
-class NotificationController
+final class NotificationController
 {
     /**
      * @var ClientInterface
@@ -81,6 +81,8 @@ class NotificationController
             return JsonResponse::create('', JsonResponse::HTTP_NO_CONTENT);
         }
 
-        return JsonResponse::fromJsonString($hubResponse->getBody()->getContents());
+        $hubResponse = json_decode($hubResponse->getBody()->getContents(), true);
+
+        return new JsonResponse($hubResponse);
     }
 }
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -82,6 +82,8 @@
 
 * Route `sylius_admin_address_log_entry_index` was renamed to `sylius_admin_partial_log_entry_index`.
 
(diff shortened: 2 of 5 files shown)


pamil:
Btw. looks like you don't have your Git email configured at Github :)

src/Sylius/Bundle/AdminBundle/Controller/NotificationController.php
```diff
@@ -81,6 +81,8 @@ public function getVersionAction(Request $request)
             return JsonResponse::create('', JsonResponse::HTTP_NO_CONTENT);
         }
 
-        return JsonResponse::fromJsonString($hubResponse->getBody()->getContents());
+        $hubResponse = json_decode($hubResponse->getBody()->getContents(), true);
+
+        return new JsonResponse($hubResponse);
```
pamil: What's the point of decoding JSON just to encode it again?

src/Sylius/Bundle/AdminBundle/Controller/NotificationController.php
```diff
@@ -23,7 +23,7 @@
 /**
  * @author Jan Góralski <jan.goralski@lakion.com>
  */
-class NotificationController
+final class NotificationController
```
pamil: It's a minor BC break so it's OK for it to be introduced but we'd still need some info in UPGRADE.md

src/Sylius/Bundle/AdminBundle/Controller/NotificationController.php
bitbager: JsonResponse::fromJsonString does not work with ajax success function. I had not enough time to look deeper into what causes this problem, that's why I just left the new JsonResponse instance call. Looks like the fromJsonString was not used correctly.

src/Sylius/Bundle/AdminBundle/Controller/NotificationController.php
bitbager: Ok, I will put an extra line into the UPGRADE.md file tomorrow.

src/Sylius/Bundle/AdminBundle/Controller/NotificationController.php
pamil: Looks like it's related to encoding options (Symfony uses `JSON_HEX_TAG | JSON_HEX_APOS | JSON_HEX_AMP | JSON_HEX_QUOT`).

PR #8260, opened 2017-07-06
[UI] Show filter label when form type is defined

Takeno:
Fix empty label when `form_options: { type }` is 'string'.

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

How to reproduce:
in `sylius_grid_mymodel.yml` specify form_options to avoid matching type dropdown:

```
year:
  type: string
  form_options:
    type: contains
```

And field label disappear:
schermata 2017-07-06 alle 19 16 16


lchrusciel:
Thanks Matteo!

PR #8267, opened 2017-07-11
Send environment during GUS version check

### LINKED ISSUE #8108: Sylius/GUS broken comparison logic.

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | fixes #8108 |
| License         | MIT |


src/Sylius/Bundle/AdminBundle/Tests/Controller/NotificationControllerTest.php
```diff
@@ -13,6 +13,7 @@
 
 use GuzzleHttp\ClientInterface;
 use GuzzleHttp\Exception\ConnectException;
+use GuzzleHttp\Psr7\Uri;
```
lchrusciel: Is it needed?

pamil: Not really.

PR #8268, opened 2017-07-11
Do not rely on admin's locale for taxon autocomplete / tree select

### LINKED ISSUE #8130: Can't add a new product in a taxon

### LINKED ISSUE #8225: Locale of admin user used in taxon autocomplete choice type (but unavailable content)
Might be related to #7895.

I just noticed some strange behavior in the Admin. My admin user was set to locale 'nl', while all our content is 'nl_NL'. In that case, the autocomplete taxon choice type is not working e.g., as it's searching for locale 'nl'. Not for 'nl%' e.g. 

I think it makes sense if the default locale of an admin user can only be chosen from the list of locales activated for the shop. This will prevent this and it also makes sense not to have a locale that the system is not serving at all.

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8130, fixes #8225, closes #8221 |
| License         | MIT |

Thanks @bitbager for the initial PR I based on :)

### DIFF
--- a/features/product/managing_products/select_taxons_for_product.feature
+++ b/features/product/managing_products/select_taxons_for_product.feature
@@ -10,6 +10,7 @@ Feature: Select taxon for a product
         And the store has a "T-Shirt Banana" configurable product
         And the store has a product "T-Shirt Batman"
         And I am logged in as an administrator
+        And I am using "English (United Kingdom)" locale for my panel
 
     @ui @javascript
     Scenario: Specifying main taxon for configurable product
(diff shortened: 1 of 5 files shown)


lchrusciel:
👍

PR #8253, opened 2017-07-05
Take shop firewallname in services from container parameter

jdeveloper:
Since now we have the option to configure the shop firewall name, chage it to be able to impersonate

| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | #8210  |
| License         | MIT |

PR #8270, opened 2017-07-11
Remove CoreBundle -> ShopBundle coupling on UserImpersonator

### LINKED ISSUE #8257: User impersonator couples core with shop bundle
This service is defined in core, used by admin and coupled with shop... We need some neat solution for this. Reported in https://github.com/Sylius/Sylius/pull/8253#discussion_r125844284.

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8257, introduced by #8253 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/services.xml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/services.xml
@@ -25,6 +25,11 @@
             <tag name="sylius.context.locale" priority="128" />
         </service>
 
+        <service id="sylius.security.shop_user_impersonator" class="Sylius\Bundle\CoreBundle\Security\UserImpersonator" public="false">
+            <argument type="service" id="session" />
+            <argument type="string">%sylius_shop.firewall_context_name%</argument>
+        </service>
+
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/AdminBundle/Resources/config/services.xml
```diff
@@ -25,6 +25,11 @@
             <tag name="sylius.context.locale" priority="128" />
         </service>
 
+        <service id="sylius.security.shop_user_impersonator" class="Sylius\Bundle\CoreBundle\Security\UserImpersonator" public="false">
+            <argument type="service" id="session" />
+            <argument type="string">%sylius_shop.firewall_context_name%</argument>
```
lchrusciel: Doesn't this value couple admin with shop?

pamil: It is already coupled, as currently AdminBundle uses service from CoreBundle which uses parameter from ShopBundle, this PR just gets CoreBundle out of the equation :D

lchrusciel: What if we could just have a similar parameter in admin, and override its value if shop bundle is enabled? Otherwise, we would use some default value.

PR #8273, opened 2017-07-12
[Fix][OAuth] Attach existing non-registered customer

psihius:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets |    |
| License         | MIT |

Check if customer with provided email already exists before trying to create one - fixes a UniqueConstraint violation error and is in line with default direct registration logic.
Also tries to read First name and Last name from OAuth provided data first and then reverts back to the default behaviour of reading "Real name" for first name and nickname/email for last name

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
+++ b/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
@@ -142,7 +142,7 @@ class UserProvider extends BaseUserProvider implements AccountConnectorInterface
         /** @var SyliusUserInterface $user */
         $user = $this->userFactory->createNew();
 
-        $customer = $this->customerRepository->findOneBy(['emailCanonical' => $response->getEmail()]);
+        $customer = $this->customerRepository->findOneBy(['emailCanonical' => strtolower($response->getEmail())]);
 
         if (null === $customer) {
             /** @var CustomerInterface $customer */
@@ -158,17 +158,16 @@ class UserProvider extends BaseUserProvider implements AccountConnectorInterface
 
         if (null !== $name = $response->getFirstName()) {
             $customer->setFirstName($name);
-        } else {
-            if (null !== $realName = $response->getRealName()) {
+        } elseif (null !== $realName = $response->getRealName()) {
                 $customer->setFirstName($realName);
-            }
         }
+
         if (null !== $lastName = $response->getLastName()) {
             $customer->setLastName($lastName);
-        } else {
-            if (!$user->getUsername()) {
-                $user->setUsername($response->getEmail() ?: $response->getNickname());
-            }
+        }
+
+        if (!$user->getUsername()) {
+            $user->setUsername($response->getEmail() ?: $response->getNickname());
         }
 
         // set random password to prevent issue with not nullable field & potential security hole
--- a/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
+++ b/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
@@ -142,10 +142,12 @@ class UserProvider extends BaseUserProvider implements AccountConnectorInterface
         /** @var SyliusUserInterface $user */
         $user = $this->userFactory->createNew();
 
-        $customer = $this->customerRepository->findOneBy(['emailCanonical' => strtolower($response->getEmail())]);
+        $canonicalEmail = $this->canonicalizer->canonicalize($response->getEmail());
+
+        /** @var CustomerInterface $customer */
+        $customer = $this->customerRepository->findOneBy(['emailCanonical' => $canonicalEmail]);
 
         if (null === $customer) {
-            /** @var CustomerInterface $customer */
             $customer = $this->customerFactory->createNew();
         }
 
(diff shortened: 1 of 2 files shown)


psihius:
@lchrusciel here's the PR. Trivial pretty much

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -132,21 +141,34 @@ private function createUserByOAuthUserResponse(UserResponseInterface $response)
     {
         /** @var SyliusUserInterface $user */
         $user = $this->userFactory->createNew();
-        /** @var CustomerInterface $customer */
-        $customer = $this->customerFactory->createNew();
+
+        $customer = $this->customerRepository->findOneBy(['emailCanonical' => $response->getEmail()]);
```
stloyd: You should probably call `strtolower()` cause you can't expect that response will contain email in lowercase.

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -132,21 +141,34 @@ private function createUserByOAuthUserResponse(UserResponseInterface $response)
...
+            if (null !== $realName = $response->getRealName()) {
+                $customer->setFirstName($realName);
+            }
         }
-
-        if (!$user->getUsername()) {
-            $user->setUsername($response->getEmail() ?: $response->getNickname());
+        if (null !== $lastName = $response->getLastName()) {
+            $customer->setLastName($lastName);
+        } else {
```
stloyd: This `else` looks a bit wrong, cause you are not setting in it `lastName` but the `username`.

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -132,21 +141,34 @@ private function createUserByOAuthUserResponse(UserResponseInterface $response)
...
         // set default values taken from OAuth sign-in provider account
         if (null !== $email = $response->getEmail()) {
             $customer->setEmail($email);
         }
 
-        if (null !== $realName = $response->getRealName()) {
-            $customer->setFirstName($realName);
+        if (null !== $name = $response->getFirstName()) {
+            $customer->setFirstName($name);
+        } else {
```
stloyd: Why not `elseif`?

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
psihius: Yeah, my bad, fix incoming

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
psihius: Sounds reasonable

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
psihius: Fair enough 👍

psihius:
Feedback patch applied, please review :)

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -132,17 +141,29 @@ private function createUserByOAuthUserResponse(UserResponseInterface $response)
     {
         /** @var SyliusUserInterface $user */
         $user = $this->userFactory->createNew();
-        /** @var CustomerInterface $customer */
-        $customer = $this->customerFactory->createNew();
+
+        $customer = $this->customerRepository->findOneBy(['emailCanonical' => strtolower($response->getEmail())]);
```
lchrusciel: Instead of `strtolower` we should use canonicalizer service, which is injected already to have coherent behaviour.

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -132,17 +141,29 @@ private function createUserByOAuthUserResponse(UserResponseInterface $response)
...
+
+        if (null === $customer) {
+            /** @var CustomerInterface $customer */
```
lchrusciel: This type hint should be placed when `$customer` is first time initialized

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -132,17 +141,29 @@ private function createUserByOAuthUserResponse(UserResponseInterface $response)
...
         if (null !== $email = $response->getEmail()) {
             $customer->setEmail($email);
         }
 
-        if (null !== $realName = $response->getRealName()) {
-            $customer->setFirstName($realName);
+        if (null !== $name = $response->getFirstName()) {
+            $customer->setFirstName($name);
+        } elseif (null !== $realName = $response->getRealName()) {
+                $customer->setFirstName($realName);
```
lchrusciel: Wrong indent

pjedrzejewski:
Fixed in #8604.

PR #8282, opened 2017-07-15
fix sorter

vcraescu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

### DIFF
--- a/src/Sylius/Component/Grid/Sorting/Sorter.php
+++ b/src/Sylius/Component/Grid/Sorting/Sorter.php
@@ -12,7 +12,6 @@
 namespace Sylius\Component\Grid\Sorting;
 
 use Sylius\Component\Grid\Data\DataSourceInterface;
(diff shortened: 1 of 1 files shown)


src/Sylius/Component/Grid/Sorting/Sorter.php
```diff
@@ -32,9 +31,10 @@ public function sort(DataSourceInterface $dataSource, Grid $grid, Parameters $pa
 
         foreach ($sorting as $field => $order) {
             $gridField = $grid->getField($field);
-            $property = $gridField->getSortable();
 
-            $expressionBuilder->addOrderBy($property, $order);
+            if ($gridField->getSortable()) {
```
lchrusciel: We should add explicit check with null value

PR #8283, opened 2017-07-15
[Theme] Working with json manifest file version strategy

regnisolbap:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -
| License       | MIT

While I was trying to add the promising Webpack Encore tool to my sylius app, I stumbled across some issue. When the assets are served from the webpack dev server it’s more convenient to use the json manifest file version strategy to not change the path manually when switching from dev to production. The problem is, when an asset is served from  "http://localhost:8080/build" for exemple, the getUrl method of the PathPackage class won’t detect that it’s an absolute path and returns a relative path like "/http://localhost:8080/build". The solution I found is to apply the version strategy to $path and then check if it’s an absolute url. Maybe it’s not a good solution for every cases, so let me know what you think about it !

### DIFF
--- a/src/Sylius/Bundle/ThemeBundle/Asset/Package/PathPackage.php
+++ b/src/Sylius/Bundle/ThemeBundle/Asset/Package/PathPackage.php
@@ -57,6 +57,8 @@ class PathPackage extends BasePathPackage
      */
     public function getUrl($path)
     {
+        $path = $this->getVersionStrategy()->applyVersion($path);
+
         if ($this->isAbsoluteUrl($path)) {
             return $path;
         }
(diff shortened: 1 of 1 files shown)


pamil:
I think we should clone the current behaviour of Symfony's `PathPackage`:

```php
public function getUrl($path)
{
    if ($this->isAbsoluteUrl($path)) {
        return $path;
    }

    $versionedPath = $this->getVersionStrategy()->applyVersion($path);

    // if absolute or begins with /, we're done
    if ($this->isAbsoluteUrl($versionedPath) || ($versionedPath && '/' === $versionedPath[0])) {
        return $versionedPath;
    }

    return $this->getBasePath().ltrim($versionedPath, '/');
}
```

And add theme-related logic just after versioned path is generated.

PR #8214, opened 2017-06-28
[Taxonomy] Allow parent locale for taxonSlugGenerator

theodormanolescu:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | fixes #8213 |
| License         | MIT |

PR #8286, opened 2017-07-17
Taxon slug generator made locale-aware and decoupled from repository

### LINKED ISSUE #8213: [Taxonomy] TaxonSlugGenerator allow parent locale
When generating a slug for the taxon from the parent you cannot set desired locale.
Ex:
* code: footwear
  * translation slugs:
    * en: footwear
    * ro: incaltaminte
* code: sneakers
  * translation slugs:
    * en: footwear/sneakers
    * ro: **footwear/adidasi**

it should be: **incaltaminte/adidasi**

### LINKED ISSUE #8280: Taxon slug generator - interface change
The current definition of `TaxonSlugGeneratorInterface` is as follows:

```php
interface TaxonSlugGeneratorInterface
{
    /**
     * @param string $name
     * @param int|null $parentId
     *
     * @return string
     */
    public function generate($name, $parentId = null);
}
```

This comes with the following disadvantages:
 - it assumes that the only used part will be a current taxon name
 - it assumes that the parent taxon is already persisted in the database, it makes you flush again and again when importing a taxon tree
 - it does not allow for creating a slug for a given locale (#8213 and #8214 address it)

The change I'm proposing is going to be a BC break, but rather a small one and easy to get by:

```php
interface TaxonSlugGeneratorInterface
{
    public function generate(TaxonInterface $taxon, string $localeCode): string;
}
```

This solution will move slug generation related logic directly to the generator (eg. which properties are used), make parent taxon eventually fetched by from the subject taxon instead of fetched from repository from an parent ID passed explicitly and ensure correct localisation of slug.

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | kind of |
| BC breaks?      | yes |
| Related tickets | fixes #8213, closes #8214, resolves #8280 |
| License         | MIT |

Big thanks to @theodormanolescu for the initial work and reporting this bug. :)


pjedrzejewski:
Awesome, thank you @theodormanolescu and @pamil! 👍 👍 👍

PR #8301, opened 2017-07-20
[Core] Improve token assigner

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

The logic will be more consistent if we assign a token only when there is no other value.

### DIFF
--- a/src/Sylius/Component/Core/spec/TokenAssigner/UniqueIdBasedOrderTokenAssignerSpec.php
+++ b/src/Sylius/Component/Core/spec/TokenAssigner/UniqueIdBasedOrderTokenAssignerSpec.php
@@ -41,13 +41,14 @@ final class UniqueIdBasedOrderTokenAssignerSpec extends ObjectBehavior
         $order->setTokenValue('yahboiiiii')->shouldBeCalled();
 
         $this->assignTokenValue($order);
+        $this->assignTokenValueIfNotSet($order);
     }
 
     function it_does_nothing_if_token_is_already_assigned(RandomnessGeneratorInterface $generator, OrderInterface $order)
     {
         $order->getTokenValue()->willReturn('yahboiiiii');
-        $order->setTokenValue(Argument::type('string'))->shouldNotBeCalled();
+        $order->setTokenValue(Argument::any())->shouldNotBeCalled();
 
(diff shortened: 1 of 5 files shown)


src/Sylius/Component/Core/spec/TokenAssigner/UniqueIdBasedOrderTokenAssignerSpec.php
```diff
@@ -39,9 +40,18 @@ function it_is_an_order_token_assigner()
...
         $generator->generateUriSafeString(10)->willReturn('yahboiiiii');
         $order->setTokenValue('yahboiiiii')->shouldBeCalled();
 
         $this->assignTokenValue($order);
     }
+
+    function it_does_nothing_if_token_is_already_assigned(RandomnessGeneratorInterface $generator, OrderInterface $order)
+    {
+        $order->getTokenValue()->willReturn('yahboiiiii');
+        $order->setTokenValue(Argument::type('string'))->shouldNotBeCalled();
```
pamil: `Argument::any()` will make it safer :)

lchrusciel:
I have proposed BC-Break free change. But not sure if we want to go this way. Also, this `UniqueIdBasedOrderTokenAssigner` is not the best name for this class as no id is taken into an account generating token. If we decide for a bc break and name change we can go with bigger refactoring including changing method and class names.

pamil:
Since it hasn't gone into beta.3, I'm coming back with another proposition.

 - remove assigners (as they are REALLY simple service, not an ideal extension point and still causing some confusion)
 - as assigners apart from their main job also configured the generators, introduce generators that will cover `$this->generator->generateUriSafeString(10)` from the assigner into `OrderTokenValueGenerator::__invoke() / generate()`
 - move conditional assignment logic from the assigners up one level, to the services that are directly using assigner now and inject that generator instead of the assigner

pjedrzejewski:
@pamil IIRC main reason for assigners was: state machine callbacks. :)

pjedrzejewski:
Thank you Łukasz!

PR #8303, opened 2017-07-20
Sortable fields not working when sortable: true

vcraescu:
This properly fix the sortable fields. As it is right now it's not possible to specify that a field is sortable exactly as the docs states:
[SyliusGridBundle Docs](http://docs.sylius.org/en/latest/bundles/SyliusGridBundle/configuration.html
)

`sortable: true
`

It only allows you to specify the field name otherwise a doctrine exception is thrown because `true` is used as field name.

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

### DIFF
--- a/src/Sylius/Component/Grid/Definition/Field.php
+++ b/src/Sylius/Component/Grid/Definition/Field.php
@@ -150,11 +150,11 @@ class Field
     }
 
     /**
-     * @param string|null $sortable
+     * @param string $sortable
      */
     public function setSortable($sortable)
     {
-        $this->sortable = $sortable ?: $this->name;
+        $this->sortable = $sortable;
     }
 
     /**
--- a/src/Sylius/Component/Grid/Definition/ArrayToDefinitionConverter.php
+++ b/src/Sylius/Component/Grid/Definition/ArrayToDefinitionConverter.php
@@ -92,7 +92,12 @@ final class ArrayToDefinitionConverter implements ArrayToDefinitionConverterInte
             $field->setEnabled($configuration['enabled']);
         }
         if (array_key_exists('sortable', $configuration)) {
-            $field->setSortable($configuration['sortable']);
+            $sortable = $configuration['sortable'];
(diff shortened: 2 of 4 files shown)


lchrusciel:
Can you take a look into failing builds? It seems it is failing due to your changes.

src/Sylius/Component/Grid/Definition/Field.php
```diff
@@ -152,7 +152,11 @@ public function setEnabled($enabled)
      */
     public function setSortable($sortable)
     {
-        $this->sortable = $sortable ?: $this->name;
+        if ($sortable === true) {
+            $this->sortable = $this->name;
+        }
+
+        $this->sortable = $sortable;
```
Zales0123: Don't we need any early return here? Otherwise `$this->sortable` will always be `$sortable`, no matter is it true or false.

Lutacon: My question is , why change that line ? 

> Expression expr1 ?: expr3 returns expr1 if expr1 evaluates to TRUE, and expr3 otherwise.

That's from the PHP documentation.

Also as @Zales0123 is saying with that code `$this->soratable` will always be `$sortable`.

lchrusciel: But `true` is evaluated to TRUE, but `$this->name` should be used so it is more like `$this->sortable = ($sortable === true) ? $this->name :  $sortable`

lchrusciel:
I'm totally ok with the regular if statement :) Just provided an example

vcraescu:
Please review. Thanks!

pjedrzejewski:
Looks good to me, let's wait for @pamil's opinion.

vcraescu:
@pamil any news? :)

vcraescu:
@pamil?

pamil:
Thanks @vcraescu! 🎉

PR #8304, opened 2017-07-20
Allow to create new attributes without type

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | - |
| License         | MIT |

There's no reason to forbid it.


Zales0123:
I think it was disabled, as it's not used in routing (we must specify type of attribute before entering creation form), but for sure there is no reason to forbid such a feature programmatically 👍

lchrusciel:
👍

PR #8307, opened 2017-07-23
[Documentation] improve the contribution guide for docs

gabiudrescu:
| Q               | A
| --------------- | ---
| Bug fix?        | kinda |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | n/a |
| License         | MIT |

requirements.txt for pip is located in the docs/ folder in the root of the project. also, running all these commands I ended up with an error like the one described here: https://github.com/MobSF/Mobile-Security-Framework-MobSF/issues/293#issuecomment-266879303

therefor, I thought it would be better to include that dependency as explicit.

### DIFF
--- a/docs/contributing/documentation/overview.rst
+++ b/docs/contributing/documentation/overview.rst
@@ -60,7 +60,7 @@ To test the documentation before a commit:
 
 .. code-block:: bash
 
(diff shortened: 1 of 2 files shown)


lchrusciel:
Thanks Gabi 👍

PR #8309, opened 2017-07-23
Allow multiple Accept headers in request configuration

gruberro:
By fetching all request headers of a given key instead of the first one only (which is the default of Symfonys headers `get`), configuring a request is possible event if two `Accept` headers are sent.

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | - |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/RequestConfigurationFactory.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/RequestConfigurationFactory.php
@@ -76,12 +76,18 @@ final class RequestConfigurationFactory implements RequestConfigurationFactoryIn
     {
         $parameters = [];
 
-        if (preg_match(self::API_VERSION_REGEXP, (string) $request->headers->get(self::API_VERSION_HEADER), $matches)) {
-            $parameters['serialization_version'] = $matches['version'];
+        $apiVersionHeaders = $request->headers->get(self::API_VERSION_HEADER, null, false);
+        foreach ($apiVersionHeaders as $apiVersionHeader) {
+            if (preg_match(self::API_VERSION_REGEXP, $apiVersionHeader, $matches)) {
+                $parameters['serialization_version'] = $matches['version'];
+            }
         }
 
-        if (preg_match(self::API_GROUPS_REGEXP, (string) $request->headers->get(self::API_GROUPS_HEADER), $matches)) {
-            $parameters['serialization_groups'] = array_map('trim', explode(',', $matches['groups']));
+        $apiGroupsHeaders = $request->headers->get(self::API_GROUPS_HEADER, null, false);
+        foreach ($apiGroupsHeaders as $apiGroupsHeader) {
+            if (preg_match(self::API_GROUPS_REGEXP, $apiGroupsHeader, $matches)) {
+                $parameters['serialization_groups'] = array_map('trim', explode(',', $matches['groups']));
+            }
         }
 
         return array_merge($request->attributes->get('_sylius', []), $parameters);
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ResourceBundle/Controller/RequestConfigurationFactory.php
```diff
@@ -74,12 +74,18 @@ private function parseApiParameters(Request $request)
     {
         $parameters = [];
 
-        if (preg_match(self::API_VERSION_REGEXP, $request->headers->get(self::API_VERSION_HEADER), $matches)) {
-            $parameters['serialization_version'] = $matches['version'];
+        $apiVersionHeaders = $request->headers->get(self::API_VERSION_HEADER, null, false);
+        foreach ($apiVersionHeaders as $apiVersionHeader) {
+            if (preg_match(self::API_VERSION_REGEXP, $apiVersionHeader, $matches)) {
+                $parameters['serialization_version'] = $matches['version'];
```
lchrusciel: I've approved it too quickly. According to [this issue](https://github.com/symfony/symfony/issues/16646) and [this question on SO](https://stackoverflow.com/questions/39912920/how-to-interpret-multiple-accept-headers) we should rather combine returned values instead of honoring only the last one.

gruberro: How do you think the combined values should like like? I can imagine that `serialization_groups` are concatenated, but what should happen with the `serialization_version`? IIRC the serializer context can deal with a single version only... WDYT?

lchrusciel: You are right. I'm afraid jsm serializer won't resolve multiple versions properly. I would do the following. The last version should win, but rest should be concatenated.

gruberro: ✔️

src/Sylius/Bundle/ResourceBundle/Controller/RequestConfigurationFactory.php
```diff
@@ -74,12 +74,23 @@ private function parseApiParameters(Request $request)
...
-            $parameters['serialization_groups'] = array_map('trim', explode(',', $matches['groups']));
+        $serializationGroups = [];
+        $apiGroupsHeaders = $request->headers->get(self::API_GROUPS_HEADER, null, false);
+        foreach ($apiGroupsHeaders as $apiGroupsHeader) {
+            if (preg_match(self::API_GROUPS_REGEXP, $apiGroupsHeader, $matches)) {
+                $serializationGroups = array_merge($serializationGroups, array_map('trim', explode(',', $matches['groups'])));
+            }
+        }
+
+        if (count($serializationGroups)) {
```
lchrusciel: Is it needed? `array_merge` will return an empty array in the worst case. I think we should remove this if.

gruberro: ✔️

lchrusciel:
According to travis, it will not work also. I'm afraid that this PR can be problematic. Right now we are not resolving serialization groups properly. We can either move back to your previous solution or investigate it further if the current result is ok and we need to adjust JMS Serializer and responses.

gruberro:
@lchrusciel Went back to the previous solution. This is already an improvement in favor of the initial version. I know that this does not allow to combine `serialization_groups` over multiple `Accept`-header statements but this is IMHO not a show stopper. At least not more than randomly parsing only the first `Accept` header in the request 😉 .

lchrusciel:
Thanks @gruberro and sorry for a mess 👍

pjedrzejewski:
Nice work Robert, thanks!

PR #8314, opened 2017-07-24
remove "or" from ui because it breaks with foreign translations

### LINKED ISSUE #7580: Combined buttons look strange with translations
Some languages have longer words for "or", for example the german "oder" looks strange in the buttons in the Admin-UI.

Possibly there are other languages with even longer words.

bildschirmfoto 2017-02-22 um 21 13 49

psren:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no|
| BC breaks?      | no |
| Related tickets | fixes #7580 |
| License         | MIT |

# Preview

Before:

1992009c-f944-11e6-9a72-db8c0215ed0c

After:

bildschirmfoto 2017-07-24 um 15 13 08

# Questions
It does not seem like the translation is used in any other place.
Should i remove it from the translations (which would be a very minor BC-Break, if anyone uses this key for anything else)
If I should remove it: where do i remove it? Only in the english language files or everywhere?
(Sorry, I don't know how the auto-update from crowdin works)

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/ProductVariant/generate.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/ProductVariant/generate.html.twig
@@ -25,7 +25,6 @@
     <div class="ui basic segment">
         <div class="ui buttons">
             <button class="ui labeled icon primary button" type="submit"><i class="random icon"></i>{{- 'sylius.ui.generate'|trans -}}</button>
-            <div class="or" data-text="{{ 'sylius.ui.or'|trans }}"></div>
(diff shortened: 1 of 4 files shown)


psren:
@pamil do you have an answer to my open question, too?

pamil:
When removing a translation key it's only needed to remove it from English ones and it will be synced automatically by @SyliusBot. As for this specific case I'd say it doesn't really matter whether we remove it or not, so we can leave it as is.

psren:
Okay, so this one is ready :-) Thanks for your help!

pamil:
Thanks Patrick! 🎉

PR #8318, opened 2017-07-25
Allow filterable resource routes

alcaeus:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

This adds the `filterable` option to the ResourceLoader, allowing to specify filterable routes without having to manaually redeclare the `index` route.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Routing/Configuration.php
+++ b/src/Sylius/Bundle/ResourceBundle/Routing/Configuration.php
@@ -37,6 +37,7 @@ final class Configuration implements ConfigurationInterface
                     ->prototype('scalar')
                     ->end()
                 ->end()
+                ->booleanNode('filterable')->end()
                 ->variableNode('form')->cannotBeEmpty()->end()
                 ->scalarNode('serialization_version')->cannotBeEmpty()->end()
                 ->scalarNode('section')->cannotBeEmpty()->end()
--- a/src/Sylius/Bundle/ResourceBundle/Routing/ResourceLoader.php
+++ b/src/Sylius/Bundle/ResourceBundle/Routing/ResourceLoader.php
@@ -169,6 +169,9 @@ final class ResourceLoader implements LoaderInterface
         if (!empty($configuration['criteria'])) {
             $defaults['_sylius']['criteria'] = $configuration['criteria'];
         }
+        if (array_key_exists('filterable', $configuration)) {
+            $defaults['_sylius']['filterable'] = $configuration['filterable'];
+        }
         if (isset($configuration['templates']) && in_array($actionName, ['show', 'index', 'create', 'update'], true)) {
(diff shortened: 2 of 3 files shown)


src/Sylius/Bundle/ResourceBundle/Routing/Configuration.php
```diff
@@ -37,6 +37,7 @@ public function getConfigTreeBuilder()
                     ->prototype('scalar')
                     ->end()
                 ->end()
+                ->scalarNode('filterable')->defaultValue(false)->end()
```
pamil: Shouldn't it be a boolean node?

src/Sylius/Bundle/ResourceBundle/Routing/ResourceLoader.php
```diff
@@ -169,6 +169,9 @@ private function createRoute(MetadataInterface $metadata, array $configuration,
         if (!empty($configuration['criteria'])) {
             $defaults['_sylius']['criteria'] = $configuration['criteria'];
         }
+        if (!empty($configuration['filterable'])) {
```
pamil: If it should, then it should be changed to `array_key_exists`.

src/Sylius/Bundle/ResourceBundle/Routing/Configuration.php
alcaeus: Correct.

src/Sylius/Bundle/ResourceBundle/Routing/ResourceLoader.php
alcaeus: I used `empty` to avoid adding `filterable: false` to every route specification as that's the default value anyways. I can change it, will take a moment to fix all tests though.

alcaeus:
@pamil that was less than expected. Done!

src/Sylius/Bundle/ResourceBundle/Routing/ResourceLoader.php
pamil: Understood. If so, wouldn't removing the default value `false` from the routing configuration node make it behave like before but with the added possibility to customise it? If not set at all it won't even set that `filterable` setting in `ResourceLoader` so it will fall back to the `RequestConfiguration` defaults.

alcaeus: Correct, didn't think of that! Changed it and squashed commits.

pamil: Awesome, thanks! :)

pamil:
Thanks Andreas! 🎉

pjedrzejewski:
👍

PR #8325, opened 2017-07-26
[Addressing] Use typehints introduced by PHP 7+

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | related to #8114 |
| License         | MIT |

*Introducing visual debt to Sylius...*

### Pros
 - ensuring correctness of existing docblock (ex. `ZoneMatcherInterface::matchAll` had `Collection` return docblock while it was returning an array and service using it in CoreBundle was expecting an array too)
 - removing possibility of weird method calls like `$province->setCountry();` and replacing them with explicit `$province->setCountry(null);`
 - easier maintenance
 - preventing weird type-related bugs, like using a boolean as a string etc.

### Cons
 - BC breaks - related to method signatures, customised code which implements an interface / extends a class from Sylius will need an update (which should be just copy-paste)

### DIFF
--- a/src/Sylius/Component/Addressing/Matcher/ZoneMatcher.php
+++ b/src/Sylius/Component/Addressing/Matcher/ZoneMatcher.php
@@ -24,17 +24,17 @@ use Sylius\Component\Resource\Repository\RepositoryInterface;
  * @author Gonzalo Vilaseca <gvilaseca@reiss.co.uk>
  * @author Jan Góralski <jan.goralski@lakion.com>
  */
-class ZoneMatcher implements ZoneMatcherInterface
+final class ZoneMatcher implements ZoneMatcherInterface
 {
     /**
      * @var RepositoryInterface
      */
-    protected $zoneRepository;
+    private $zoneRepository;
 
     /**
      * @var array
      */
-    protected $priorities = [
+    private const PRIORITIES = [
         ZoneInterface::TYPE_PROVINCE,
         ZoneInterface::TYPE_COUNTRY,
         ZoneInterface::TYPE_ZONE,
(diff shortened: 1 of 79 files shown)


bendavies:
are you planning on removing redundant docblocks?

pamil:
That's the tricky part, I'm for deleting these docblocks, but...:

```php
function (string $a, int $b, array $c) { ... }
```

What if `$c` is an array of some objects (like `$addresses` are `array|AddressInterface[]`)? Should we include docblocks for the whole function, all arguments, return types etc. or just for this one argument? Same while adding `@throws` docblocks.

bendavies:
Indeed. in my project (for your example), I am further documenting the arrays. FYI you can use phpstan and then https://github.com/thecodingmachine/phpstan-strict-rules to enforce this.

pamil:
As for v1.0.0 I'm planning to introduce some kind of tools to enforce a coding standard supported by static analysis checks in Travis build, so that porting bugfixes from 1.0 up would be easier. We've tried it before, but it was too restricting then and it was more discouraging contributors than helping with maintenance :)

src/Sylius/Component/Addressing/Matcher/ZoneMatcher.php
```diff
@@ -24,17 +24,17 @@
  * @author Gonzalo Vilaseca <gvilaseca@reiss.co.uk>
  * @author Jan Góralski <jan.goralski@lakion.com>
  */
-class ZoneMatcher implements ZoneMatcherInterface
+final class ZoneMatcher implements ZoneMatcherInterface
```
lchrusciel: We should add an note about finals in UPGRADE file as well

pamil: Done.

PR #8334, opened 2017-07-27
[ResourceBundle] Return event response if exists before deleting resource

dannyvw:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

Return event response so it's possible to return a different template with a list of relations (as example) before deleting the resource.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
@@ -405,6 +405,10 @@ class ResourceController extends Controller
         if ($event->isStopped()) {
             $this->flashHelper->addFlashFromEvent($configuration, $event);
 
+            if ($event->hasResponse()) {
+                return $event->getResponse();
+            }
+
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
```diff
@@ -405,6 +405,10 @@ public function deleteAction(Request $request)
         if ($event->isStopped()) {
             $this->flashHelper->addFlashFromEvent($configuration, $event);
 
+            if ($event->hasResponse()) {
+                return $event->getResponse();
```
lchrusciel: Can you provide another spec which will cover this feature?

dannyvw: @lchrusciel Can you check the PR?

pjedrzejewski:
Thanks Danny! Nice work. :)

PR #8341, opened 2017-07-31
[User] Drop pointless max length constraints for pin / token generators

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | - |
| License         | MIT |

### DIFF
--- a/src/Sylius/Component/User/Security/Generator/UniquePinGenerator.php
+++ b/src/Sylius/Component/User/Security/Generator/UniquePinGenerator.php
@@ -55,11 +55,7 @@ final class UniquePinGenerator implements GeneratorInterface
             $pinLength,
             'The value of pin length has to be an integer.'
         );
-        Assert::range(
-            $pinLength,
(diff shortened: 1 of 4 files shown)


src/Sylius/Component/User/Security/Generator/UniquePinGenerator.php
```diff
@@ -55,11 +55,7 @@ public function __construct(
             $pinLength,
```
lchrusciel: Can't we remove this assert and just add typehint and declare strict types?

lchrusciel:
Thanks @pamil 👍

src/Sylius/Component/User/Security/Generator/UniquePinGenerator.php
pamil: I'd rather make these changes in component-wide PRs like #8325.

PR #8342, opened 2017-07-31
Use stable phpspec

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | - |
| License         | MIT |

### DIFF
--- a/bin/symfony_requirements
+++ b/bin/symfony_requirements
@@ -1,8 +1,6 @@
 #!/usr/bin/env php
 <?php
 
-declare(strict_types=1);
-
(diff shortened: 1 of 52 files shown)


bin/symfony_requirements
```diff
@@ -1,8 +1,6 @@
 #!/usr/bin/env php
 <?php
 
-declare(strict_types=1);
```
pamil: Seems like it shouldn't be added by hand in #8117.

pamil:
Thanks @pamil! :DDD

PR #8343, opened 2017-07-31
[Maintenance] Make all constants public

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

Since php7.1 we are allowed to declare visibility of constants. Since all of them are public right now I suggest making it explicit and coherent with the regular variables.


pamil:
Thanks Łukasz! 💃

PR #8344, opened 2017-07-31
Check for input minimum length on customer existence check in checkout

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no|
| BC breaks?      | no|
| License         | MIT |

Based on Customer validation an email should be between 2 - 254 characters. I noticed in my 404 logs that sometimes this request is fired without any information in it (empty). That will for sure return no valid customer and it's ineffecient.

Based on https://semantic-ui.com/behaviors/api.html#/usage ('Cancelling requests') this adjustment should prohibit this call when nothing (or something too long) is entered. It could be extended to check for email validatity, but that's always an hard issue.

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/private/js/app.js
+++ b/src/Sylius/Bundle/ShopBundle/Resources/private/js/app.js
@@ -31,8 +31,14 @@
             throttle: 1500,
 
             beforeSend: function (settings) {
+                var email = $('#sylius_checkout_address_customer_email').val();
+
+                if (email.length < 3) {
+                    return false;
+                }
+
                 settings.data = {
-                    email: $('#sylius_checkout_address_customer_email').val()
+                    email: email
                 };
 
                 return settings;


sweoggy:
I am not sure that the upper limit should be hard coded here. If it's, then it should be mentioned in the doc as it is not obvious that this particular validation is client side as well as server side. The low limit is fine IMO as no valid email can consist of one character

stefandoorn:
@sweoggy Makes sense, let's see what Sylius team thinks about it and I will adjust it.

pamil:
Yeah, I'd drop the upper bound as well and it's good to go.

And an email has to have at least three characters, so we can increase the lower bound by one :)

Zales0123:
I think it's a good idea for now to check only if email is too short - this is something that can be done much more often, even accidentally, in contrast to over 255-letters email :D But let's ask for other's opinion.

stefandoorn:
@pamil @Zales0123 Alright, makes sense to me. But, if email should be atleast 3, then we might also need to change validation on PHP side? Currently it's defined inside Sylius as atleast 2 (that's how I came up with the lower limit).

stefandoorn:
Adjusted to PR to check client-side for minimum of 3 characters.

pamil:
@stefandoorn yeah, we might change that too. I wonder if it's really needed since there should be an email constraint as well which uses `/^.+\@\S+\.\S+$/` regexp enforcing at least 5 characters (nevermind though) :D

stefandoorn:
There is an Email constraint indeed. But, then it could also make sense to remove the minimum length constraint :-) Well, I submitted an issue, let's see in there what is the best approach.

pamil:
Thanks Stefan! 🎉

PR #8350, opened 2017-08-01
Remove wrong label for grid string

lasotaartur:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | |
| License         | MIT |

Field "Value" for filter string in GridBundle is not translated.


pjedrzejewski:
I will check it again because I recall there was a reason for this. :)

Zales0123:
@pjedrzejewski did you check it again? 😄

I believe it's not relevant somehow, I've just checked it and it seems to work:

zrzut ekranu 2018-11-14 o 13 44 56
zrzut ekranu 2018-11-14 o 13 40 10
zrzut ekranu 2018-11-14 o 13 40 40

Feel free to open it if there is still some problem :)

PR #8352, opened 2017-08-01
[Shipment] Shipment amount was not updated after decreasing item quantity

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

Steps to reproduce:

1. Create shipping method with `per_unit_rate` calculator
2. Add some product to cart with quantity bigger than 1
3. Go to checkout and select shipping method created in *1.*
4. Go back to cart an decrease quantity of added order item
5. Cry a lot, as shipping fee is as big as it was before.

This is just a hotfix, to be 100% that proper units are added for the only one shipment. We should definitely think about some more sophisticated solution.

### DIFF
--- a/src/Sylius/Component/Core/OrderProcessing/OrderShipmentProcessor.php
+++ b/src/Sylius/Component/Core/OrderProcessing/OrderShipmentProcessor.php
@@ -69,6 +69,10 @@ final class OrderShipmentProcessor implements OrderProcessorInterface
             return;
         }
 
+        foreach ($shipment->getUnits() as $unit) {
+            $shipment->removeUnit($unit);
+        }
+
         foreach ($order->getItemUnits() as $itemUnit) {
             if (null === $itemUnit->getShipment()) {
                 $shipment->addUnit($itemUnit);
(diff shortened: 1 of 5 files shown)


pjedrzejewski:
Looks good to me! 👍

src/Sylius/Component/Core/OrderProcessing/OrderShipmentProcessor.php
```diff
@@ -69,6 +69,10 @@ public function process(BaseOrderInterface $order)
             return;
         }
 
+        foreach ($shipment->getUnits() as $unit) {
```
pamil: Side note: This flow of removing all the entities from the collection and adding some of them back does not work with `orphanRemoval` enabled. Fortunately, it's explicitly disabled in the `Shipment` mapping, but still, it's something we should be aware of.

It's all right by now, but we can make it more reliable by performing only the necessary operations like in https://github.com/Sylake/SyliusConsumerPlugin/blob/master/src/Projector/ProductProjector.php#L216-L239.

pamil:
Thanks Mateusz! 🚂

PR #8374, opened 2017-08-05
[Locale] get translation for country name

regnisolbap:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

I added a second argument $local in LocalHelper, just like in the LocalConverter, in order to get the country name translated. So the twig filter would now look like:

`{{ locale|sylius_locale_name(app.request.locale) }}`

### DIFF
--- a/src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelper.php
+++ b/src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelper.php
@@ -27,19 +27,30 @@ final class LocaleHelper extends Helper implements LocaleHelperInterface
     private $localeConverter;
 
     /**
-     * @param LocaleConverterInterface $localeConverter
+     * @var string
      */
-    public function __construct(LocaleConverterInterface $localeConverter)
+    private $defaultLocale;
+
+    /**
+     * @param LocaleConverterInterface $localeConverter
+     * @param string $defaultLocale
+     */
+    public function __construct(LocaleConverterInterface $localeConverter, string $defaultLocale)
     {
         $this->localeConverter = $localeConverter;
+        $this->defaultLocale = $defaultLocale;
     }
 
     /**
      * {@inheritdoc}
      */
-    public function convertCodeToName(string $localeCode): ?string
+    public function convertCodeToName(string $localeCode, ?string $locale = null): ?string
     {
-        return $this->localeConverter->convertCodeToName($localeCode);
+        if (null === $locale) {
+            $locale = $this->defaultLocale;
+        }
+
+        return $this->localeConverter->convertCodeToName($localeCode, $locale);
     }
 
     /**
--- a/src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelperInterface.php
+++ b/src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelperInterface.php
@@ -22,8 +22,9 @@ interface LocaleHelperInterface extends HelperInterface
 {
     /**
      * @param string $localeCode
+     * @param string $locale
      *
      * @return string|null
      */
-    public function convertCodeToName(string $localeCode): ?string;
+    public function convertCodeToName(string $localeCode, ?string $locale): ?string;
 }
--- a/src/Sylius/Bundle/LocaleBundle/Resources/config/services.xml
+++ b/src/Sylius/Bundle/LocaleBundle/Resources/config/services.xml
@@ -51,6 +51,7 @@
 
         <service id="sylius.templating.helper.locale" class="Sylius\Bundle\LocaleBundle\Templating\Helper\LocaleHelper">
             <argument type="service" id="sylius.locale_converter" />
+            <argument>%sylius_locale.locale%</argument>
             <tag name="templating.helper" alias="sylius_locale" />
         </service>
 
--- a/src/Sylius/Bundle/LocaleBundle/spec/Templating/Helper/LocaleHelperSpec.php
+++ b/src/Sylius/Bundle/LocaleBundle/spec/Templating/Helper/LocaleHelperSpec.php
@@ -25,7 +25,7 @@ final class LocaleHelperSpec extends ObjectBehavior
 {
     function let(LocaleConverterInterface $localeConverter): void
     {
-        $this->beConstructedWith($localeConverter);
+        $this->beConstructedWith($localeConverter, 'en');
     }
 
     function it_is_a_helper(): void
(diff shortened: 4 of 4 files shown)


pjedrzejewski:
@regnisolbap Hey Pablo, thanks for your PR. What I would suggest is that the example in specs is a new one. So you should add it instead of replacing the old one, which has default values. :) WDYT?

regnisolbap:
@pjedrzejewski thanks for your feedback ! I add a new example in specs and kept the original one as you suggested, but the original one appear to be broken with this errror:

>   48  - it converts locales code to name
      method call:
        - convertCodeToName("fr_FR", "")
      on Double\LocaleConverterInterface\LocaleConverterInterface\P2169 was not expected, expected calls were:
        - convertCodeToName(exact("fr_FR"))

I'm new to PhpSpec and testing in general, what does that mean and how can I fix it ?

Thanks !

pjedrzejewski:
You should update the expected method call in the original spec to include the second argument. :) Based on the exception message.

src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelper.php
```diff
@@ -37,9 +37,9 @@ public function __construct(LocaleConverterInterface $localeConverter)
     /**
      * {@inheritdoc}
      */
-    public function convertCodeToName($localeCode)
+    public function convertCodeToName($localeCode, $locale = 'en')
     {
-        return $this->localeConverter->convertCodeToName($localeCode);
+        return $this->localeConverter->convertCodeToName($localeCode, $locale);
```
pamil: What do you think about making `$locale` null by default, injecting `%sylius_locale.locale%` (which should be the default locale of the application) here and using it is a fallback when locale is not passed?

regnisolbap: Sorry about the composer.lock... I totally agree with your proposal ! I just need some advice on the specs, should I keep the original method "it_converts_locales_code_to_name" ? I was thinking about removing it and renaming the one added here in "it_converts_locales_code_to_name_with_provided_locale_or_default_locale". WDYT ?

lchrusciel: Let's make both tests separate?  `it_converts_locales_code_to_name_with_provided_locale` and `it_converts_locales_code_to_name_with_default_locale`

regnisolbap: Yes, it makes sense. I commited the new changes.

regnisolbap:
@lchrusciel @pamil @pjedrzejewski do you still think this feature is relevant ? There's no hurry in merging of course, you have plenty of work to do !

lchrusciel:
@regnisolbap it seems good to me, but just let's wait for another approval.

src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelper.php
```diff
@@ -27,19 +27,29 @@
...
     /**
+     * @var string
+     */
+    private $defaultLocale;
+
+    /**
      * @param LocaleConverterInterface $localeConverter
      */
-    public function __construct(LocaleConverterInterface $localeConverter)
+    public function __construct(LocaleConverterInterface $localeConverter, $defaultLocale)
```
GSadee: `string $defaultLocale`

src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelper.php
```diff
@@ -27,19 +27,29 @@
     private $localeConverter;
 
     /**
+     * @var string
+     */
+    private $defaultLocale;
+
+    /**
      * @param LocaleConverterInterface $localeConverter
```
GSadee: Missing `@param string $defaultLocale`

src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelper.php
```diff
@@ -27,19 +27,29 @@
...
     {
         $this->localeConverter = $localeConverter;
+        $this->defaultLocale = $defaultLocale;
     }
 
     /**
      * {@inheritdoc}
      */
-    public function convertCodeToName(string $localeCode): ?string
+    public function convertCodeToName(string $localeCode, string $locale = null): ?string
```
GSadee: `?string $locale = null`

src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelperInterface.php
```diff
@@ -23,7 +23,9 @@
     /**
      * @param string $localeCode
      *
```
GSadee: Unnecessary line

src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelperInterface.php
```diff
@@ -23,7 +23,9 @@
...
+     * @param string $locale
+     *
      * @return string|null
      */
-    public function convertCodeToName(string $localeCode): ?string;
+    public function convertCodeToName(string $localeCode, string $locale): ?string;
```
GSadee: `?string $locale = null`

regnisolbap:
Thanks for your reviews @GSadee, I just pushed the updated code, let me know if it's ok for you.

src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelper.php
```diff
@@ -27,19 +27,30 @@
...
     {
         $this->localeConverter = $localeConverter;
+        $this->defaultLocale = $defaultLocale;
     }
 
     /**
      * {@inheritdoc}
      */
-    public function convertCodeToName(string $localeCode): ?string
+    public function convertCodeToName(string $localeCode, ?string $locale = null): ?string
```
lchrusciel: `locale` -> `resultLocale`/`displayLocale`. It is quite ambiques to have `localeCode` and `locale`

src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelper.php
```diff
@@ -27,19 +27,30 @@
...
     {
-        return $this->localeConverter->convertCodeToName($localeCode);
+        if (null === $locale) {
```
lchrusciel: Maybe we could go with $locale = $locale ?? $this->defaultLocale;

src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelper.php
```diff
@@ -27,19 +27,30 @@
...
+     * @var string
+     */
+    private $defaultLocale;
+
+    /**
      * @param LocaleConverterInterface $localeConverter
+     * @param string $defaultLocale
      */
-    public function __construct(LocaleConverterInterface $localeConverter)
+    public function __construct(LocaleConverterInterface $localeConverter, string $defaultLocale)
```
pamil: BC promise is still a work in progress, but it looks like a BC break (adding required parameters to the public method).

src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelperInterface.php
```diff
@@ -22,8 +22,9 @@
 {
     /**
      * @param string $localeCode
+     * @param string $locale
```
pamil: `string|null`

src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelper.php
regnisolbap: I agree and would go with displayLocale !

src/Sylius/Bundle/LocaleBundle/Templating/Helper/LocaleHelper.php
regnisolbap: So, should I make the defaultLocale parameter null by default (doing ?string $defaultLocale = null) ?

PR #8384, opened 2017-08-08
Moved the shipping required check to the order model itself

### LINKED ISSUE #8380: Introduce service to check if order needs shipping
See my comment here: https://github.com/Sylius/Sylius/pull/8062#discussion_r131912352. I think it is a quite common check and could be done by a service.

Persata:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | resolves #8380  |
| License         | MIT |

Hi,

As mentioned in #8380, I've moved the check for whether shipping is required onto the order model itself.

I've added it to the core `OrderInterface` since that's where the other shipping related method definitions exist, and also updated the PHPSpec tests to suit.

Let me know if there's any issues or anything I've missed!

### DIFF
--- a/src/Sylius/Component/Core/Model/Order.php
+++ b/src/Sylius/Component/Core/Model/Order.php
@@ -318,6 +318,20 @@ class Order extends BaseOrder implements OrderInterface
     }
 
     /**
+     * @return bool
+     */
+    public function requiresShipping()
+    {
+        foreach ($this->getItems() as $orderItem) {
+            if ($orderItem->getVariant()->isShippingRequired()) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function getShipments()
--- a/src/Sylius/Component/Core/spec/Checker/OrderShippingMethodSelectionRequirementCheckerSpec.php
+++ b/src/Sylius/Component/Core/spec/Checker/OrderShippingMethodSelectionRequirementCheckerSpec.php
@@ -45,21 +45,9 @@ final class OrderShippingMethodSelectionRequirementCheckerSpec extends ObjectBeh
     }
 
     function it_says_that_shipping_method_do_not_have_to_be_selected_if_none_of_variants_from_order_requires_shipping(
-        OrderInterface $order,
-        OrderItemInterface $firstItem,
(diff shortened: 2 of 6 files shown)


pjedrzejewski:
@Persata Nice! Could you add a note to the UPGRADE file? Adding this method is a BC break on the interface.

src/Sylius/Component/Core/spec/Checker/OrderShippingMethodSelectionRequirementCheckerSpec.php
```diff
@@ -45,70 +45,39 @@ function it_implements_order_shipping_necessity_checker_interface()
...
+        $order->requiresShipping()->willReturn(true);
+
         $order->getChannel()->willReturn($channel);
         $channel->isSkippingShippingStepAllowed()->willReturn(true);
 
         $order->getShipments()->willReturn([$shipment]);
 
         $shippingMethodsResolver->getSupportedMethods($shipment)->willReturn([$shippingMethod]);
 
+
```
Zales0123: Redundant ;)

src/Sylius/Component/Core/Model/Order.php
```diff
@@ -318,6 +318,20 @@ public function getLastPayment($state = null)
     }
 
     /**
+     * @return bool
+     */
+    public function requiresShipping()
```
pamil: That was an unfortunate name, cause there is already `ProductVariant::isShippingRequired()`. Let's be concise and make it the same, wdyt?

Persata: Can do, provided everyone is happy - just took the name suggested in #8380, easy enough to change!

pjedrzejewski: @Persata let's go with ``isShippingRequired`` :)

pamil: @Persata yeah, I've suggested the first name that came to my mind, that usually isn't the best one :D

pjedrzejewski:
Oups, I was too quick. :/ @Persata would you mind opening another PR with name change? Sorry for the trouble...

Issue #8313, opened 2017-07-24
[ShopBundle][Checkout] Checkout-Templates, small refactoring


psren:
Hey there,

in the checkout there is the support-box in the right column.

![bildschirmfoto 2017-07-24 um 14 38 43](https://user-images.githubusercontent.com/1016798/28523797-35b9b3d0-707e-11e7-81cc-c9ad55977e8a.png)

The Phone-Number is hard-coded.
Maybe we can remove the code completely (I know the default template just acts as an example application).

Also the code (and template events) for the right column are duplicated:
[here](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ShopBundle/Resources/views/Checkout/selectPayment.html.twig#L28-L38), [here](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ShopBundle/Resources/views/Checkout/selectShipping.html.twig#L29-L38), and [here](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ShopBundle/Resources/views/Checkout/address.html.twig#L29-L39)

## Proposal

I propose the following changes:

- [x] Remove right column from ```selectPayment.html.twig```, ```selectShipping.html.twig``` and ```address.html.twig``` and add it to ```layout.html.twig``` instead (```content```-Block)
Then override the ```content``` block in ```complete.html.twig```
- [x] Adjust the names of the Template-Events (the ```address``` namespace is wrong, because it's in ```Checkout/layout.html.twig``` wich is used by ```selectPayment|selectShipping|complete```, too.
- [x] Remove the ```_support.html.twig``` Events [here](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ShopBundle/Resources/views/Checkout/selectPayment.html.twig#L33-L37)
- [x] Add missing ```after_summary``` Event (or remove the summary-events at all. (We could introduce a ```sylius.shop.checkout.additional```  or ```sylius.shop.checkout.aside```  Event and then register a Plugin/Bundle which adds elements like support/summary to the sidebar (we know about the request context there, too, so using it for only one of the checkout-steps should be fairly easy)
- [x] remove the ```_support.html.twig``` completely (can be added with template-events by a plugin). The number is also hardcoded at the moment.

What do you think to the proposals? Are there some u like?

mheki:
How about adding a phone number to the channel? We have already a `contact email` field there.

psren:
[...]

PR #8390, opened 2017-08-10
Clean up admin templates

psren:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes (some template events changing) |
| Related tickets | partially #8313 |
| License         | MIT |


src/Sylius/Bundle/ShopBundle/Resources/views/Checkout/address.html.twig
```diff
@@ -2,40 +2,21 @@
 
 {% form_theme form '@SyliusUi/Form/theme.html.twig' %}
 
-{% block content %}
```
lchrusciel: Changing blocks name should also be mentioned in UPGRADE.md

psren: yeah, will have time at the end of the week. It is planned not to rename the block at all, i'll revert that change

psren:
This one is finished imo.
I will add the info-box to the demo-shop as soon this one is merged, or should i do the pr on the demo, now?

src/Sylius/Bundle/ShopBundle/Resources/views/Checkout/address.html.twig
psren: i documented the change now

lchrusciel:
The PR needs rebase. Also, can you squash some of the commits? 14 commits for 7 files is a little bit too much. 

Regarding your question, I would do it in separate PR but first, let's discuss if we want to put this change in the core. At the end changing blocks can slightly break existing websites.

pamil:
Thanks for the effort put into this PR! I'm sorry we've missed it on the way to stable release, we aren't able to merge it now because it contains BC breaks.

PR #8393, opened 2017-08-11
Add event before / after product box on product listing

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | yes |
| BC breaks?      | no|
| License         | MIT |

Add the ability to hook into a product listing before and after each product box.


pjedrzejewski:
Thank you Stefan, makes sense!

PR #8394, opened 2017-08-11
Fix events on product listings, resources.data should be used

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

In the product list, the events get supplied a `products` var, which is empty. The products are actually in `resources.data`, which therefore should be supplied to the events.


pjedrzejewski:
Thanks Stefan, I think ``products`` should contain the grid. But I think it makes sense to pass the products here.

PR #8401, opened 2017-08-14
[Behat] Add a less strict alternative to SymfonyPage::verifyUrl

mkilmanas:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | N/A |
| License         | MIT |

The problem at hand was that when verifying if the current url belongs to the page object it expects the URL to match perfectly, but in case of `CRUD/UpdatePage` this is not always desired, as it requires ID of the resource to be passed in order to generate the full expected URL. But in some cases the ID is not relevant and I would be happy for it to match any URL that is for an UpdatePage of my resource.

For this reason, I have added a new method `verifyRoute()` which will validate if the matched route has the same name as expected and validates any parameters that you explicitly provide (any other parameters that exist in the matched route but not provided by you will be ignored and accepted)

This means that if you wish to match an exact URL with exact parameters, you can still achieve that. But in case when you don't care for parameters, you can pass none and only route will be checked.

Feedback very welcome

### DIFF
--- a/src/Sylius/Behat/Page/SymfonyPage.php
+++ b/src/Sylius/Behat/Page/SymfonyPage.php
@@ -87,6 +87,63 @@ abstract class SymfonyPage extends Page implements SymfonyPageInterface
     }
 
     /**
+     * @param array $requiredUrlParameters
+     * @throws UnexpectedPageException
+     */
+    protected function verifyRoute(array $requiredUrlParameters = [])
+    {
+        $url = $this->getDriver()->getCurrentUrl();
+        $path = parse_url($url)['path'];
+
+        $path = preg_replace('#^/app(_dev|_test|_test_cached)?\.php/#', '/', $path);
+        $matchedRoute = $this->router->match($path);
+
+        $this->verifyRouteName($matchedRoute, $url);
+        $this->verifyRouteParameters($requiredUrlParameters, $matchedRoute);
+    }
+
+
+    /**
+     * @param array $matchedRoute
+     * @param string $url
+     * @throws UnexpectedPageException
+     */
+    private function verifyRouteName(array $matchedRoute, string $url)
+    {
+        if ($matchedRoute['_route'] !== $this->getRouteName()) {
+            throw new UnexpectedPageException(
+                sprintf(
+                    "Matched route '%s' does not match the expected route '%s' for URL '%s'",
+                    $matchedRoute['_route'],
+                    $this->getRouteName(),
+                    $url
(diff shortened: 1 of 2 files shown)


mkilmanas:
Right, I see that people were using this with different intentions/understanding and this change may cause too much trouble. I will convert this into a pure new feature (`verifyRoute()`) and update this PR accordingly

src/Sylius/Behat/Page/SymfonyPage.php
```diff
@@ -87,6 +87,63 @@ protected function verifyUrl(array $urlParameters = [])
...
+
+    /**
+     * @param array $requiredUrlParameters
+     * @param array $matchedRoute
+     * @throws UnexpectedPageException
+     */
+    private function verifyRouteParameters(array $requiredUrlParameters, array $matchedRoute)
+    {
+        foreach ($requiredUrlParameters as $key => $value) {
+            if (!isset($matchedRoute[$key]) || $matchedRoute[$key] != $value) {
```
lchrusciel: Comparison should be strict

src/Sylius/Behat/Page/SymfonyPage.php
```diff
@@ -87,6 +87,63 @@ protected function verifyUrl(array $urlParameters = [])
...
+                    $url
+                )
+            );
+        }
+    }
+
+    /**
+     * @param array $requiredUrlParameters
+     * @param array $matchedRoute
+     * @throws UnexpectedPageException
```
lchrusciel: Missing blank line

src/Sylius/Behat/Page/SymfonyPage.php
```diff
@@ -87,6 +87,63 @@ protected function verifyUrl(array $urlParameters = [])
...
+
+        $this->verifyRouteName($matchedRoute, $url);
+        $this->verifyRouteParameters($requiredUrlParameters, $matchedRoute);
+    }
+
+
+    /**
+     * @param array $matchedRoute
+     * @param string $url
+     * @throws UnexpectedPageException
```
lchrusciel: Missing blank line

mkilmanas:
@lchrusciel - Does it look better now?
I have also made the method public and added it to the interface - otherwise this method could only be used by extending the SymfonyPage, but it should not be necessary

pjedrzejewski:
Looks good to me, thanks Marijus!

PR #8409, opened 2017-08-16
[ResourceBundle] Add check before add translation mapping

### LINKED ISSUE #8328: [ResourceBundle] Doctrine inheritance type joined not supported?
If you register both entities (`PetTranslation` and `DogTranslation`) as sylius resource you will get the following error: `Property "translatable" in "AppBundle\Entity\DogTranslation" was already declared, but it must be declared only once`. The errors are triggered from `ORMTranslatableListener.php`.

Is someone using a configuration like this and has the same problem?

Example code:

```
/**
 * @ORM\Entity
 * @ORM\InheritanceType("JOINED")
 * @ORM\DiscriminatorMap({"dog"="Dog"})
 */
abstract class Pet implements TranslatableInterface
{
    use TranslatableTrait;

    public function getName()
    {
        return $this->getTranslation()->getName();
    }

    // class has multiple relations with other entities
}

class PetTranslation extends AbstractTranslation
{
    private $name;

    public function getName()
    {
        return $this->name;
    }
}

class Dog extends Pet
{
}

class DogTranslation extends PetTranslation
{
}
```

dannyvw:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | #8328 |
| License         | MIT |

With this fix we can define our own mapping to resolve #8328

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/EventListener/ORMTranslatableListener.php
+++ b/src/Sylius/Bundle/ResourceBundle/EventListener/ORMTranslatableListener.php
@@ -126,15 +126,17 @@ final class ORMTranslatableListener implements EventSubscriber
         /** @var MetadataInterface $translationResourceMetadata */
         $translationResourceMetadata = $this->resourceMetadataRegistry->get($resourceMetadata->getAlias().'_translation');
 
-        $metadata->mapOneToMany([
-            'fieldName' => 'translations',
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
Awesome, thanks Danny! 👍

Issue #7262, opened 2017-01-16
Duplicate database execute when using grid.


liverbool:
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/GridBundle/Doctrine/ORM/DataSource.php#L79
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php#L60

PS. Seem no need `$paginator->getCurrentPageResults()` in grid-bundle `$paginator->setMaxPerPage()` on resource-bundle will reset cached result.

michalmarcinkowski:
Should be fixed by https://github.com/Sylius/Sylius/pull/7319.

PR #8417, opened 2017-08-18
[Admin] Fix pagination in case of custom method with several join

### LINKED ISSUE #4211: Administration Pagination Broken in every Page

Pitoune:
…l join

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #4211 |
| License         | MIT |


lchrusciel:
And this one https://github.com/Sylius/Sylius/issues/7262 is also relevant

pamil:
Yeah, I think it's not a good solution considering #7262, but I couldn't reproduce it on any grid in the admin panel, so either it is solved or we'd need some reproducer for that.

PR #8419, opened 2017-08-20
Remove duplicate width property

juukie:
The width property was declared twice.


Zales0123:
Thank you Jürgen!

juukie:
:+1:

PR #8421, opened 2017-08-21
[Channel] Use typehints introduced by PHP7+

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | part of #8425 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Component/Channel/Model/Channel.php
+++ b/src/Sylius/Component/Channel/Model/Channel.php
@@ -24,32 +24,32 @@ class Channel implements ChannelInterface
     use TimestampableTrait, ToggleableTrait;
 
     /**
-     * @var mixed
+     * @var mixed|null
      */
     protected $id;
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $code;
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $name;
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $description;
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $hostname;
 
     /**
-     * @var string
+     * @var string|null
      */
     protected $color;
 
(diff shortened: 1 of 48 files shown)


Zales0123:
Generally 👍  But what about the docblocks? Do we want to leave them for now, or should we remove it while introducing strict types? For me, there are of course some situations where docblock is needed:
- precise the `Collection` or `array` elements
- point out what exception is thrown by a method
- describe functionality if it's really needed

I'm quite ok with leaving them for now, I just want to highlight what path should we follow 📡

pamil:
@Zales0123 we should make a main issue for these kinds of PRs for discussions :) It was discussed in the previous PR as well - https://github.com/Sylius/Sylius/pull/8325#issuecomment-318036138. I'm OK with both possibilities, removing them if they're redundant or leaving them, but I'd rather prolong it until we got a static analysis tool that can tell us when a docblock is really needed. Not removing them here also means shorter diffs :)

src/Sylius/Component/Channel/Model/Channel.php
```diff
@@ -24,32 +24,32 @@ class Channel implements ChannelInterface
...
 
     /**
-     * @var mixed
+     * @var mixed|null
      */
     protected $id;
 
     /**
-     * @var string
+     * @var string|null
```
pamil: I'm OK with both, the only reason for using `|null` is that there are more occurrences of this one in our code and that means... shorter diffs :) Anyway, we can discuss to replace them project-wide (again, with a static analysis tool configured to enforce that).

Zales0123:
@pamil makes sense, totally agreed 🚀

src/Sylius/Component/Channel/Model/Channel.php
lchrusciel: Fine for me, I have approved it already. Just a topic for discussion.

psihius:
Yeah, keep them I think.
They have use cases that are not covered by the typing system

Zales0123:
Thank you Kamil!

Issue #8425, opened 2017-08-22
Use typehints introduced by PHP7+


pamil:
A meta-issue to keep track of what needs to be done.

 - [ ] Global UPGRADE.md note about scalar and return typehints being introduced - @pamil
 - [x] Addressing / AddressingBundle - @pamil - #8325 
 - [x] AdminApiBundle - @Zales0123 - #8442
 - [x] AdminBundle - @Zales0123 - https://github.com/Sylius/Sylius/pull/8454
 - [x] Attribute / AttributeBundle - @Zales0123 - #8427
 - [x] Channel / ChannelBundle - @pamil - #8421
 - [x] Core / CoreBundle - @Zales0123 - #8478 
 - [x] Currency / CurrencyBundle - @pamil - #8428
 - [x] Customer / CustomerBundle - @pamil - #8432
 - [x] FixturesBundle - @pamil - #8459
 - [x] Grid / GridBundle - @Zales0123 - https://github.com/Sylius/Sylius/pull/8457
 - [x] Inventory / InventoryBundle - @Zales0123 - #8445 
 - [x] Locale / LocaleBundle - @pamil - #8430
 - [x] Mailer / MailerBundle - @Zales0123 - #8429
 - [x] MoneyBundle - @pamil - #8434
 - [x] Order / OrderBundle - @pamil - #8473
 - [x] Payment / PaymentBundle - @Zales0123 - #8425
 - [x] PayumBundle - @Zales0123 - #8444 
 - [x] Product / ProductBundle - @Zales0123 - https://github.com/Sylius/Sylius/pull/8466
 - [x] Promotion / PromotionBundle - @Zales0123 - https://github.com/Sylius/Sylius/pull/8436
 - [x] Registry - @pamil - #8433
 - [x] Resource / ResourceBundle - @pamil - #8439 
 - [x] Review / ReviewBundle - @Zales0123 - #8446 
 - [x] Shipping / ShippingBundle - @Zales0123 - https://github.com/Sylius/Sylius/pull/8440
 - [x] ShopBundle - @Zales0123 - https://github.com/Sylius/Sylius/pull/8453
 - [x] Taxation / TaxationBundle - @Zales0123 - https://github.com/Sylius/Sylius/pull/8451
 - [x] Taxonomy / TaxonomyBundle - @pamil - #8447 
 - [x] ThemeBundle - @pamil - #8450
 - [x] User / UserBundle - @pamil - #8458
[...]

PR #8435, opened 2017-08-22
[Payment] Scalar types in Payments

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | https://github.com/Sylius/Sylius/issues/8425 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Component/Payment/Model/PaymentInterface.php
+++ b/src/Sylius/Component/Payment/Model/PaymentInterface.php
@@ -33,50 +33,50 @@ interface PaymentInterface extends TimestampableInterface, ResourceInterface
     /**
      * @return PaymentMethodInterface
      */
-    public function getMethod();
+    public function getMethod(): ?PaymentMethodInterface;
 
     /**
-     * @param null|PaymentMethodInterface $method
+     * @param PaymentMethodInterface|null $method
      */
-    public function setMethod(PaymentMethodInterface $method = null);
+    public function setMethod(?PaymentMethodInterface $method): void;
 
     /**
-     * @return string
+     * @return string|null
      */
-    public function getState();
+    public function getState(): ?string;
 
     /**
      * @param string $state
      */
-    public function setState($state);
+    public function setState(string $state): void;
 
     /**
-     * @return string
+     * @return string|null
      */
-    public function getCurrencyCode();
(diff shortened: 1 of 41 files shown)


src/Sylius/Component/Payment/Model/PaymentInterface.php
```diff
@@ -33,50 +33,50 @@
...
      */
-    public function setAmount($amount);
+    public function setAmount(int $amount): void;
 
     /**
-     * @return array
+     * @return array|null
      */
-    public function getDetails();
+    public function getDetails(): ?array;
```
pamil: What about making it return `array` only and define the default value as an empty array?

src/Sylius/Component/Payment/Model/PaymentMethod.php
```diff
@@ -57,7 +57,7 @@ public function __construct()
     /**
      * @return string
      */
-    public function __toString()
+    public function __toString(): string
     {
         return $this->getName();
```
pamil: `return (string) $this->getName();`, `getName` return `?string`

src/Sylius/Component/Payment/Model/PaymentMethodTranslation.php
```diff
@@ -40,7 +40,7 @@ class PaymentMethodTranslation extends AbstractTranslation implements PaymentMet
     /**
      * @return string
      */
-    public function __toString()
+    public function __toString(): string
     {
         return $this->getName();
```
pamil: Same here

src/Sylius/Component/Payment/Model/PaymentMethod.php
```diff
@@ -89,87 +89,87 @@ public function setCode(?string $code): void
...
+    public function getEnvironment(): ?string
     {
         return $this->environment;
     }
 
     /**
      * {@inheritdoc}
      */
-    public function setEnvironment($environment)
+    public function setEnvironment(string $environment): void
```
pamil: I guess we need to make it nullable (Symfony forms).

pamil:
Thanks Matthias! 🎉

Issue #8371, opened 2017-08-04
[Contact] Add reply-to email when sending contact request mail


Niiko:
In fact,

We have the customer email in the mail template, but not in the reply-to field of the received email.

We have to copy paste each time we receive a mail to answer.

kochen:
@Niiko this was discuss in Slack Support channel.
have you tried this (`config.yml`)?
```sylius_mailer:
    sender:
        name: Store Name
        address: noreply@store.com```

**Comment by Niiko** [comment]
The problem is, that address depend on user which send a contact request, not a support one.

In the config you submit, this is always that address that will be used.

I have plenty of requests, and each time i had to copy/paste the one of the user inside the email template

**Comment by kochen** [comment]
How about simply creating your own: `ContactEmailManager`?
```
        <service id="sylius.email_manager.contact" class="Sylius\Bundle\ShopBundle\EmailManager\ContactEmailManager">
            <argument type="service" id="sylius.email_sender" />
        </service>
```

**Comment by Niiko** [comment]
I think it has to be in core, it's standard behaviour for requests.

In fact, you could see it in requestAction from ContactController 
```php
$this->contactEmailManager->sendContactRequest($data, [$contactEmail]);
```

Of course, i can override it, that's not the problem here ;-)

I certainly will PR that one when i will fix it for my project.

pjedrzejewski:
Implemented, thanks @loic425 & @Niiko for reporting. :)

PR #8441, opened 2017-08-23
[Mailer] add reply-to on contact request email

loic425:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | mentioned in #8371  |
| License         | MIT |

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -40,6 +40,7 @@
 ### Mailer / MailerBundle
 
 * `Email` has been made final, use decoration instead of extending it.
+$ `SenderInterface::send` method has changed to add `reply-to` argument
 
 ### Order / OrderBundle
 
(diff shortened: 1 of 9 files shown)


stefandoorn:
I think this is a BC break, as it's changing the interface. I'm using it also in a custom implementation, which won't like that.

But, +1 for this. What about BCC / CC? This is a great win already for SPF / email spam etc., but BCC / CC could also be helpful now and then.

loic425:
@stefandoorn Yes, i've changed bc break to true.

pjedrzejewski:
@loic425 Looks good, could you just add a note to UPGRADE file about the interface change? :)

loic425:
@pjedrzejewski done :)

loic425:
@pjedrzejewski @lchrusciel 
The linked issue is tagged as 1.0.0, so this one could me imho :)

pjedrzejewski:
Thank you @loic425! Nice work. :)

PR #8446, opened 2017-08-24
[Review] Scalar types in Reviews

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | https://github.com/Sylius/Sylius/issues/8425 |
| License         | MIT |


src/Sylius/Component/Core/Model/Product.php
```diff
@@ -268,31 +268,31 @@ public function getAcceptedReviews()
...
+    public function removeReview(ReviewInterface $review): void
     {
         $this->reviews->remove($review);
     }
 
     /**
      * {@inheritdoc}
      */
-    public function getAverageRating()
+    public function getAverageRating(): float
```
pamil: `?float`, isn't it? :)

src/Sylius/Component/Review/Factory/ReviewFactory.php
```diff
@@ -57,8 +59,9 @@ public function createForSubject(ReviewableInterface $subject)
     /**
      * {@inheritdoc}
      */
-    public function createForSubjectWithReviewer(ReviewableInterface $subject, ReviewerInterface $reviewer = null)
+    public function createForSubjectWithReviewer(ReviewableInterface $subject, ?ReviewerInterface $reviewer = null): ReviewInterface
```
pamil: We can remove the default `null` value here, when called without the second parameter then `createForSubject` should be used instead.

src/Sylius/Component/Review/Factory/ReviewFactoryInterface.php
```diff
@@ -28,13 +28,13 @@
...
+    public function createForSubject(ReviewableInterface $subject): ReviewInterface;
 
     /**
      * @param ReviewableInterface $subject
      * @param ReviewerInterface|null $reviewer
      *
      * @return ReviewInterface
      */
-    public function createForSubjectWithReviewer(ReviewableInterface $subject, ReviewerInterface $reviewer = null);
+    public function createForSubjectWithReviewer(ReviewableInterface $subject, ?ReviewerInterface $reviewer): ReviewInterface;
```
Zales0123: Unfortunately it's required to have it here for guest reviews. I've just made it required to pass the `$reviewer` explicitly as the default value was ridiculous.

pamil:
Thanks Matt! ☂️

Issue #7853, opened 2017-03-29
Feedback doc ProductBundle


tvequaud:
Just installed a fresh brand new Symfony 3.2 to test the ProductBundle installation guide : http://docs.sylius.org/en/latest/bundles/SyliusProductBundle/installation.html
Here is my feedback.

1. You should update the documentation to advice developers to use the dev-master branch with composer by this way
`composer require sylius/produc-bundle dev-master`
or in composer.json, add or update the 2 fields in bold :
```
{
	"name" : "thomas/ecommerce",
	"license" : "proprietary",
	"type" : "project",
	**"minimum-stability" : "dev",
	"prefer-stable" : true,**
```
Otherwise the stable version 0.19 is installed instead, unusable and not supported anymore.

2. Then, exception during `doctrine:schema:update` :
```
thomas@debian /var/www/ecommerce $ php bin/console doctrine:schema:update --force

                                                                                                                                      
  [Symfony\Component\Debug\Exception\ContextErrorException]                                                                           
  Warning: class_implements(): Class Sylius\Component\Product\Model\ProductVariantTranslation does not exist and could not be loaded
```

TheMadeleine:
Hi @tvequaud :) Thanks for your feedback, although... it is not an issue. 

Since the docs are prepared and written (kind of upfront) for the oncoming stable release, then they assume that you are using the latest version available. Which is `dev-master` indeed.

tvequaud:
@TheMadeleine : I knew you'll answer that :p  it's not obvious for a beginner. And symfony is delivered in stable so by default newcomers will have a bad first impression (like I had). Even if it's only for few days/weeks.

Anyway I'm using the dev-master branch and got the exception in the second point.

TheMadeleine:
We'll discuss the version mentioning problem at Sylius today, I promise :)

Regarding the exception you've got - did you check if the class exists in the project? Cleared the cache? If your project is clear, maybe recreating the database will help.

tvequaud:
Actually the error in console says that some astract methods are not implemented...

I opened the `/vendor/sylius/product/Model/Product.php` class and, indeed, 5 methods implemented in `Component\Product\Model\ProductInterface` are missing :

```
    /**
     * {@inheritDoc}
     * @see \Sylius\Component\Product\Model\ProductInterface::isAvailable()
     */
    public function isAvailable() {
    	// TODO: Auto-generated method stub

    }

    /**
     * {@inheritDoc}
[...]

PR #8462, opened 2017-08-26
Adds locale bundle to list of bundles

nmeirik:
It seems that the locale bundle is a dependency of the product bundle, and hence also needs to be registered with the kernel.

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | N/A |
| License         | MIT |


nmeirik:
On closer inspection, it does not seem that this change is sufficient in order for things to work properly. Registering the locale bundle with the kernel _does_ get rid of a dependency issue for the `sylius.form.type.product_attribute_value` service. However, after registering the bundle, the error is replaced by a `LocaleNotFoundException` thrown during execution of the RequestLocaleSetter. I do not know how the locale should be specified in order to remedy this issue.

loic425:
For thé locale notre round error, solution Thérèse #7853

pamil:
Thanks @nmeirik, we'll solve that other issue in #7853. 🎉

PR #8463, opened 2017-08-26
Adds slug requirement for product creation

nmeirik:
When attempting to create product without setting the slug, I get an MySQL error saying that column slug can not be null. Setting the slug persists the product. Please ensure that my proposed change is valid.

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | N/A |
| License         | MIT |


CoderMaggie:
Thank you @nmeirik ! 🍾

PR #8464, opened 2017-08-26
Documents required Variant attributes

nmeirik:
I was not able to create a variant without the code and position attributes set (MySQL complains that they cannot be null).

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | N/A |
| License         | MIT |


CoderMaggie:
@lchrusciel it is required. :)

CoderMaggie:
Thanks a lot @nmeirik! 🎉

lchrusciel:
@CoderMaggie I meant, it is required but can be assigned by gedmo. Anyway, this is probably better way

PR #8465, opened 2017-08-26
Handle exceptions thrown while deleting a resource

### LINKED ISSUE #7948: [ResourceBundle] Exceptions handling during deletion of resources
PR #7425 added a `UpdateHandlingException` exception class to handle all errors during the update of a resource and this prevents error 500 to be shown if something goes wrong. The same should be done for deletion of resources: for example, if there is a constraint on a foreign key and you try to delete something an error page is displayed. Are you willing to accept a PR to fix the issue?

ste93cry:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | fixes #7948
| License         | MIT

The aim of this PR is to handle all exceptions that can be thrown during the deletion of a resource (think for example if a `ON DELETE` constraint fails). It would be great if we could introduce a `ResourceDeleteHandler` class to let users customize how they want to delete a resource, but it would introduce a BC break because it should be a new service injected in the constructor of the controller. Also, I think that the repository should not contain a `remove` method and instead we should follow the Doctrine way that uses the `EntityManager::delete` method.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
@@ -16,6 +16,7 @@ namespace Sylius\Bundle\ResourceBundle\Controller;
 use Doctrine\Common\Persistence\ObjectManager;
 use FOS\RestBundle\View\View;
 use Sylius\Bundle\ResourceBundle\Event\ResourceControllerEvent;
+use Sylius\Component\Resource\Exception\DeleteHandlingException;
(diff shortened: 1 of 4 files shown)


ste93cry:
Tests are broken because the exceptions are now caught, but Sylius was using the `ResourceDeleteSubscriber` event subscriber to convert exceptions to a response with a flash message. Any though on how I could workaround this issue?

pjedrzejewski:
Merged in #8546, thank you Stefano for your work!

PR #8468, opened 2017-08-28
Remove 'bulk actions' from GridBundle feature list

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | no|
| BC breaks?      | no|
| License         | MIT |

The GridBundle states it supports bulk actions on the feature overview. I'm quite sure it doesn't at the moment (although it would be a great feature to have). Probably better to remove it from the docs, as this shows up on the first Google result when searching for 'Sylius bulk action'.


lchrusciel:
Thanks @stefandoorn 👍

Issue #8025, opened 2017-05-05
Promotion coupon API returns coupons that belong to other promotions


alcaeus:
Some issues are best explained using a pull request, so here goes: #8024.

There are multiple reasons this is failing: the `criteria` defined in the route for the promotion coupon API is ignored, because the route is not marked as `filterable`. Even if it were, it would not work because the `promotion` reference uses the `promotion_id`, but the API uses a promotion code. To fix this would require using a repository method, but since that requires adding a new method (`findByPromotionCode`) in `PromotionCouponRepositoryInterface`, the bug fix would result in a BC break.

Regarding the `filterable` issue, I'm not sure what other APIs might be affected by this. This would also be a good time to ensure there is enough data for the filter to actually do something, as this bug would've been caught if there were something to filter. There's no point having test if the filtered result equals the entire dataset.

PR #8471, opened 2017-08-28
[API][Promotion] Index coupons only for requested promotion

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/8025, closes https://github.com/Sylius/Sylius/pull/8024 |
| License         | MIT |

Thanks @alcaeus for pointing this out.

### DIFF
--- a/src/Sylius/Component/Promotion/Repository/PromotionCouponRepositoryInterface.php
+++ b/src/Sylius/Component/Promotion/Repository/PromotionCouponRepositoryInterface.php
@@ -43,4 +43,11 @@ interface PromotionCouponRepositoryInterface extends RepositoryInterface
      * @return PromotionCouponInterface|null
      */
     public function findOneByCodeAndPromotionCode(string $code, string $promotionCode): ?PromotionCouponInterface;
+
+    /**
+     * @param string $promotionCode
+     *
+     * @return iterable
+     */
+    public function findByPromotionCode(string $promotionCode): iterable;
 }
--- a/src/Sylius/Bundle/AdminApiBundle/Resources/config/routing/promotion_coupon.yml
+++ b/src/Sylius/Bundle/AdminApiBundle/Resources/config/routing/promotion_coupon.yml
@@ -10,8 +10,9 @@ sylius_admin_api_promotion_coupon_index:
             serialization_version: $version
             serialization_groups: [Default]
             section: admin_api
-            criteria:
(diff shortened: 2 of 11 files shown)


lchrusciel:
Thanks @alcaeus and @Zales0123! Nice job 👍

src/Sylius/Component/Promotion/Repository/PromotionCouponRepositoryInterface.php
```diff
@@ -43,4 +43,11 @@ public function countByCodeLength(int $codeLength): int;
      * @return PromotionCouponInterface|null
      */
     public function findOneByCodeAndPromotionCode(string $code, string $promotionCode): ?PromotionCouponInterface;
+
+    /**
+     * @param string $promotionCode
+     *
+     * @return iterable
+     */
+    public function findByPromotionCode(string $promotionCode): iterable;
```
pamil: All find* repository methods returns array of all results, this one returns a paginator, we should mention it somewhere or change the method name at all (`createPaginatorForPromotion`?).

pamil: Anyway, it's a BC break and should be mentioned in` UPGRADE.md`.

alcaeus: Does it need to return a paginator? IIRC the paginator is applied automatically if a route is paginated?

pamil: It works only when there's no repository method defined (here is) and for the criteria specified in routing. I'm not sure whether we can query by a nested field `promotion.code` using it.

alcaeus: Ah, your right. In that case, I'd suggest putting paginator in the name 👍

pamil: Btw. supporting relations in `criteria` looks like a useful feature.

lchrusciel: good catch, thanks @pamil

PR #8439, opened 2017-08-23
[Resource] Use typehints introduced by PHP7+

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | part of #8425 |
| License         | MIT |

PR #8480, opened 2017-08-30
fix limit type

bendavies:
fix limit being parsed to resolveMaxPerPage - must be ?int

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | #8439 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
@@ -63,7 +63,7 @@ final class ResourcesCollectionProvider implements ResourcesCollectionProviderIn
             $request = $requestConfiguration->getRequest();
 
             $paginator->setMaxPerPage($this->resolveMaxPerPage(
-                $request->query->get('limit'),
+                $request->query->has('limit') ? $request->query->getInt('limit') : null,
                 $requestConfiguration->getPaginationMaxPerPage(),
                 $paginationLimits
             ));
--- a/src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourcesCollectionProviderSpec.php
+++ b/src/Sylius/Bundle/ResourceBundle/spec/Controller/ResourcesCollectionProviderSpec.php
@@ -74,7 +74,8 @@ final class ResourcesCollectionProviderSpec extends ObjectBehavior
 
         $requestConfiguration->getRequest()->willReturn($request);
         $request->query = $queryParameters;
-        $queryParameters->get('limit')->willReturn(5);
+        $queryParameters->has('limit')->willReturn(true);
+        $queryParameters->getInt('limit')->willReturn(5);
         $queryParameters->get('page', 1)->willReturn(6);
 
         $paginator->setMaxPerPage(5)->shouldBeCalled();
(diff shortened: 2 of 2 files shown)


src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
```diff
@@ -61,9 +61,12 @@ public function get(RequestConfiguration $requestConfiguration, RepositoryInterf
 
         if ($paginator instanceof Pagerfanta) {
             $request = $requestConfiguration->getRequest();
+            if (null !== $limit = $request->query->get('limit')) {
+                $limit = (int)$limit;
+            }
 
             $paginator->setMaxPerPage($this->resolveMaxPerPage(
-                $request->query->get('limit'),
+                $limit,
```
lchrusciel: Won't the `$request->query->getInt('limit', null)` work?

bendavies: hah, i didn't see that method. will update

bendavies: @lchrusciel done

lchrusciel: Didn't know that it will break so many things... sorry :(

bendavies: I'll take a look

bendavies:
sorry @lchrusciel i've not had time to look into this yet. I've not got behat tests running locally.
Are there docs explaining how to?

lchrusciel:
And we could stay with your previous proposal, sorry about that. Maybe we will manage to add to 1.0

bendavies:
why? yours seems better. I've not investigated why the tests are failing yet

pamil:
@bendavies `getInt` is implemented as `return (int) $this->get($key, $default)` while we need something like `return $this->has($key) ? (int) $this->get($key) : null;` because `(int) null === 0`.

bendavies:
@pamil ah, are you sending `limit=` in a url somewhere?

bendavies:
I've switched to my original implementation.

bendavies:
@lchrusciel @pamil all good

src/Sylius/Bundle/ResourceBundle/Controller/ResourcesCollectionProvider.php
```diff
@@ -61,9 +61,12 @@ public function get(RequestConfiguration $requestConfiguration, RepositoryInterf
 
         if ($paginator instanceof Pagerfanta) {
             $request = $requestConfiguration->getRequest();
+            if (null !== $limit = $request->query->get('limit')) {
```
pamil: How about making it a one-liner `$limit = $request->query->has('limit') ? $request->query->getInt('limit') : null`? A little bit more readable for me (no assignment in a condition).

bendavies: done

lchrusciel:
Thanks Ben!

PR #8487, opened 2017-08-31
[Core] Removed WebServerBundle from prod env

sweoggy:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | N/A |
| License         | MIT |

[No one](http://symfony.com/doc/current/setup/web_server_configuration.html) should be using the internal web server in a production environment anyway. Makes the Kernel a tiny bit slimmer and promotes good practises (it always possible to re-add the bundle if needed).


pjedrzejewski:
Thanks Oscar!

PR #8491, opened 2017-09-03
[Admin] Mass deletion on admin grid (fixes #93)

### LINKED ISSUE #93: [Feature]: Multiple-delete from lists with checkoxes.
What are guys you think about possibility to multiple delete things from, users, lists etc ?

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | probably not |
| Related tickets | fixes #93 |
| License         | MIT |

* When enabled, on the grid between the filters and the result table, an additional element is rendered with mass action buttons. In this case only delete is available.
* In the result table checkboxes are rendered which have the ID of the resource as value.
* As soon as user clicks the mass delete button, the ID's are fetched from the checkboxes, confirmation popup is shown and if clicked through the new URL is called. ID's are sent with a DELETE method call.

Done:

* Docs
* Behat tests
* Specs
* Implement on all resources that have an 'item delete' in admin grid
* Inside ResourceController both the multiple and the single events are dispatched

Example screenshot:
zrzut ekranu 2017-10-27 o 10 03 04

To do / discuss:

- [ ] Currently let's say you can delete 10 products. If product 3 delete runs into an error, number 1 and 2 are deleted already. I couldn't find a way to wrap the deletion into a transaction without adjusting the dependencies of `ResourceController`.
- [ ] Sylius core team review.


stefandoorn:
Could someone of the Sylius team help me with a review? I'm not sure if this is the good approach, but I also think it's a valuable feature for 1.0.

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
```diff
@@ -238,6 +238,48 @@ public function indexAction(Request $request): Response
...
+
+        $this->isGrantedOr403($configuration, ResourceActions::MASS_DELETE);
+        /** @var array $resources */
+        $resources = $this->resourcesCollectionProvider->get($configuration, $this->repository);
+
+        foreach($resources as $resource) {
+            $event = $this->eventDispatcher->dispatchPreEvent(ResourceActions::DELETE, $configuration, $resource);
+
+            if ($event->isStopped() && !$configuration->isHtmlRequest()) {
+                throw new HttpException($event->getErrorCode(), $event->getMessage());
```
psihius: Should it actually do an exception here? In this case only part of the data will be removed, if there is some condition on the resource that might prevent deletion on some of them, but you need to continue further anyway? Or maybe it needs to be transnational - everything or nothing.

stefandoorn:
I'm fine with bulk. @CoderMaggie had that also as feedback.

CoderMaggie:
@gabiudrescu @stefandoorn I am strongly for "bulk action" :)

gabiudrescu:
one more question for @stefandoorn: if I select the first entry in the list and then while keeping shift on my keyboard I select the last entry, is the browser going to select all the entries in the list?

if yes, then great -> you won a beer.
if no, but you can / want / will implement it -> you can win a beer
else -> we'll negotiate this until you you pick the 2nd option

WDYT?

stefandoorn:
Haven't tested ;-) It's just select boxes - didn't do anything else fancy to it. So I guess it doesn't work.. There is also no way yet to select *ALL* - not only just what you see on the screen. I'll keep it in mind ;-)

stefandoorn:
@GSadee is helping on the implementation. Renaming to 'bulk' has been done already. More updates expected over the next days.

stefandoorn:
@GSadee WDYT about transforming `bulkDeleteAction` into `bulkAction`, requiring an `action` paramater telling which handler to load (via something like a `bulkHandlerResolver`? 

It will make things more complicated in this phase, but might make it more easy to add additional bulk actions later on (e.g. mass transform, invoice creation, sending them to external system, etc.).

stefandoorn:
@GSadee Might not be needed actually, as the configs allow to render own templates in which own routes can be called of course. First idea would mean introducing an extra dependency to ResourceController, which is not preferable. The current dependencies already suit us for delete, update & state machine transitions. Anything else can be achieved by registering own bulk action templates I think.

gabiudrescu:
@stefandoorn regarding multiple select, I did a PoC with https://github.com/rmariuzzo/checkboxes.js. you can check it out here: https://github.com/gabiudrescu/Sylius/commit/bbfcc4b53fc4a146fd69196a22e70d54999e1a30

it's a dirty PoC, I added the dependency through yarn, but didn't know how to include it in the compiled app.js. 

let me know what you think.

stefandoorn:
@gabiudrescu This allows to use shift button to select multiple? 

We also need to think about the option to select ALL (unregardless of the pagination)..

GSadee:
I guess that in this PR we should focus on basic bulk delete action functionality with proper tests, and maybe in future add to that e.g. select all button or other js magic 😃

stefandoorn:
Sure @GSadee :) Will PR it afterwards if the base functionality works :))

gabiudrescu:
@stefandoorn yes, this is what my PoC is all about. doing it with a button that selects all items on the page is a matter of a small JS:
```

jQuery(function($) {
    $('#btn-check-all').on('click', function(e) {
        $('#table1').checkboxes('check');
        e.preventDefault();
    });
});
```

you do the PHP stuff, I'll help with the UI / JS.

stefandoorn:
True, but what I meant earlier: how to do a bulk action on ALL items (e.g. on all 100 products in the database) while pagination only shows 10 on the current page? @GSadee Ideas on that one? Or should we PR that separate afterwards?

GSadee:
IMO this will be a nice feature but definitely in a seperate PR, after adding the basic functionality of bulk delete action.

gabiudrescu:
@stefandoorn i find it as a bad idea to have a "delete all" button from the database. consider this:
 - you have 400k products and you press that button
 - it will kill your webserver to handle this amount of data to be processed

I think being able to multiple select the products currently displayed in the view it's enough. 

as @GSadee says, let's ensure the PHP part reaches the codebase and we'll add more JS flavour later :grin: 

also, if you know how to add a new js dependency in sylius, I would be grateful if you could share this knowledge.

stefandoorn:
Alright, I agree.

@gabiudrescu Load it via yarn/npm, make sure it got added in `package.json` & `yarn.lock`, then adjust `Gulpfile.js` in `AdminBundle` to include it in the build process :-) 

Then run `yarn run gulp` from root and it should be in your generated JS file that is loaded in admin. Not 100% tested, from memory.

stefandoorn:
Thanks to the tremendous help of @GSadee the current PR should reflect a reviewable version :-)

stefandoorn:
Updated PR description.

pjedrzejewski:
Thank you Stefan & Grzesiu! :) Amazing work. I'd only discuss if adding the public method to the grid render interface was a BC Break. I think yes, but let me hear your thoughts. /cc @pamil

stefandoorn:
Makes sense that that is BC. Not directly sure how to get around it though.

src/Sylius/Component/Grid/Renderer/GridRendererInterface.php
```diff
@@ -48,6 +48,15 @@ public function renderAction(GridViewInterface $gridView, Action $action, $data
 
     /**
      * @param GridViewInterface $gridView
+     * @param Action $bulkAction
+     * @param mixed|null $data
+     *
+     * @return mixed
+     */
+    public function renderBulkAction(GridViewInterface $gridView, Action $bulkAction, $data = null);
```
pamil: It's a BC break for sure, we need to revert these changes and think of a better, backwards compatibility solution.

src/Sylius/Bundle/ResourceBundle/spec/Routing/ResourceLoaderSpec.php
```diff
@@ -491,7 +664,10 @@ function it_generates_routing_for_a_section(
                 'permission' => false,
             ],
         ];
-        $routeFactory->createRoute('/products/', $indexDefaults, [], [], '', [], ['GET'])->willReturn($indexRoute);
+        $routeFactory
```
pamil: Let's not do this kind of coding standard changes in a feature PR, it makes the real changes less clear.

PR #8494, opened 2017-09-04
set gender `u` as default value - resolves #8493

### LINKED ISSUE #8493: Customer::Gender is required but might be empty if removed from form
Customer::Gender is a required field:
```        <field name="gender" column="gender" type="string" length="1">
            <options>
                <option name="default">u</option>
            </options>
        </field>
```
which has a default value `u`.

but when being removed (i.e not needed for the shop) from `.../vendor/sylius/sylius/src/Sylius/Bundle/ShopBundle/Resources/views/Account/profileUpdate.html.twig`:
```<div class="field">{{ form_row(form.gender) }}</div>```
you get an error saying:
image

This issue seem to occur as well from the API - @psihius could you confirm/elaborate?

kochen:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| Related tickets | fixes #8493|
| License         | MIT |

@psihius could you confirm is fixes the API issue as well?

### DIFF
--- a/src/Sylius/Bundle/CustomerBundle/Form/Type/GenderType.php
+++ b/src/Sylius/Bundle/CustomerBundle/Form/Type/GenderType.php
@@ -31,6 +31,7 @@ final class GenderType extends AbstractType
                 'sylius.gender.male' => CustomerInterface::MALE_GENDER,
                 'sylius.gender.female' => CustomerInterface::FEMALE_GENDER,
             ],
+            'empty_data' => CustomerInterface::UNKNOWN_GENDER,
         ]);
     }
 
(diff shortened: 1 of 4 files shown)


kochen:
```Sylius\Behat\Page\UnexpectedPageException: Expected to be on "http://localhost:8080/admin/customers/666/edit" but found "http://localhost:8080/admin/login" instead in src/Sylius/Behat/Page/Page.php:132```
doesn't seem very related to my fix, is it?

lchrusciel:
@kochen it is just a random fail, that's why we have behat rerun, these scenarios pass. What you need to fix are the validation responses in API tests. There [were 3 failures](https://travis-ci.org/Sylius/Sylius/jobs/271599878#L904-L965) in CutomerApiTest

pamil:
Thank you, @kochen! 🎉

PR #8500, opened 2017-09-04
Add filter option to country choice type

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | yes |
| BC breaks?      | no|
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
+++ b/src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
@@ -45,11 +45,16 @@ final class CountryChoiceType extends AbstractType
     {
         $resolver
             ->setDefaults([
+                'choice_filter' => null,
                 'choices' => function (Options $options): iterable {
                     if (null === $options['enabled']) {
                         $countries = $this->countryRepository->findAll();
                     } else {
                         $countries = $this->countryRepository->findBy(['enabled' => $options['enabled']]);
+                    }
+
+                    if ($options['choice_filter']) {
+                        $countries = array_filter($countries, $options['choice_filter']);
                     }
 
                     usort($countries, function (CountryInterface $a, CountryInterface $b): int {
--- a/src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
+++ b/src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
@@ -70,6 +70,7 @@ final class CountryChoiceType extends AbstractType
                 'label' => 'sylius.form.address.country',
                 'placeholder' => 'sylius.form.country.select',
             ])
+            ->setAllowedTypes('choice_filter', ['null', 'callable'])
         ;
     }
 
--- a/src/Sylius/Bundle/AddressingBundle/phpunit.xml.dist
+++ b/src/Sylius/Bundle/AddressingBundle/phpunit.xml.dist
@@ -13,6 +13,7 @@
     <testsuites>
         <testsuite name="SyliusAddressingBundle Test Suite">
             <directory>./test/</directory>
+            <directory>./Tests/</directory>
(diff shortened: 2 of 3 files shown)


stefandoorn:
Hm, the logic existed previously. It might be deleted in the meanwhile - as I don't see it anymore now, but I had still some in my own project so that's what I referred to. In that case it doesn't make a lot of sense to merge this one for this sole location.

src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
```diff
@@ -64,6 +68,7 @@ public function configureOptions(OptionsResolver $resolver): void
                 'enabled' => true,
                 'label' => 'sylius.form.address.country',
                 'placeholder' => 'sylius.form.country.select',
+                'filter' => null,
```
pamil: Let's name it `choice_filter` to make it more self-explaining and consistent with `ZoneCodeChoiceType`.

stefandoorn:
Alright. Will push a change tonight.

stefandoorn:
P.S. @pamil Is that in the form of a Behat test wished or unit test?

stefandoorn:
Adjusted code, not sure what is best approach to test, cannot find an example to follow for this kind of filter :(

pamil:
@stefandoorn a unit test, you can take a look at `FixedCollectionTypeTest` or any `*TypeTest` in Sylius

stefandoorn:
I tried to get it working @pamil using an example, with registering the form in 'getExtensions' etc., but I'm probably missing one thing to make it work. Error stays that the form gets 0 arguments injected in the form constructor. If you have a moment to insert an easy test, I can finish it afterwards.

src/Sylius/Bundle/AddressingBundle/phpunit.xml.dist
```diff
@@ -13,6 +13,7 @@
     <testsuites>
         <testsuite name="SyliusAddressingBundle Test Suite">
             <directory>./test/</directory>
+            <directory>./Tests/</directory>
```
lchrusciel: We should put these tests in a one folder I suppose

src/Sylius/Bundle/AddressingBundle/Tests/Form/Type/CountryChoiceTypeTest.php
```diff
@@ -0,0 +1,128 @@
...
+        $this->assertChoicesLabels(['France', 'Poland'], ['enabled' => null]);
+    }
+
+    /**
+     * @test
+     */
+    public function it_returns_countries_in_an_alphabetical_order()
+    {
+        $this->countryRepository->findBy(['enabled' => true])->willReturn([
+            $this->france->reveal(),
```
lchrusciel: We should probably change an order of countries to have a proper assert here

stefandoorn:
Pushed adjustments based on feedback @lchrusciel.

src/Sylius/Bundle/AddressingBundle/phpunit.xml.dist
pamil: Not really, we use the following convention in other bundles (as Symfony does) and in fact `test` should be somehow merged to `Tests` (and merging it back to `test` breaks autoloading).

stefandoorn:
Now it should be good again ;-)

src/Sylius/Bundle/AddressingBundle/phpunit.xml.dist
lchrusciel: It is one of the solutions. I don't say, that we have to merge the new one with the existing one, but on the other hand we have a bundle test there. Just a thing, that we should discuss before we introduce too much test conventions in our test suite.

pjedrzejewski:
Thank you Stefan!

PR #8478, opened 2017-08-30
[Core] Scalar types in Core

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | https://github.com/Sylius/Sylius/issues/8425 |
| License         | MIT |

The last one! 🔑  🔥

PR #8507, opened 2017-09-05
[Resource] Allow parsing variable in yaml to int/float/double

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | discovered in in https://github.com/Sylius/Sylius/pull/8478 |
| License         | MIT |

While working on introducing strict scalar types in *Core*, I've realized we can have some problems with them while using parameters passed with [yml routing configuration](https://github.com/Sylius/Sylius/blob/7c13a247355d42d07a4e8cb360bbac709d706528/src/Sylius/Bundle/ShopBundle/Resources/config/routing/partial/product.yml#L16). Yaml itself supports explicit typing (https://symfony.com/doc/current/components/yaml/yaml_format.html#explicit-typing), however it does not work for variables which are always passed as a string.

I think we can use the same syntax in our resource routing to make it possible (for now only for `int`, `float` and `double`).

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ParametersParser.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ParametersParser.php
@@ -16,6 +16,7 @@ namespace Sylius\Bundle\ResourceBundle\Controller;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 use Symfony\Component\HttpFoundation\Request;
+use Webmozart\Assert\Assert;
 
 /**
  * @author Paweł Jędrzejewski <pawel@sylius.org>
@@ -75,6 +76,16 @@ final class ParametersParser implements ParametersParserInterface
 
         if (0 === strpos($parameter, 'expr:')) {
             return $this->parseRequestValueExpression(substr($parameter, 5), $request);
+        }
+
+        if (0 === strpos($parameter, '!!')) {
+            $parameter = explode(' ', $parameter);
+
+            $parser = trim($parameter[0], '!').'val';
+
+            Assert::oneOf($parser, ['intval', 'floatval', 'doubleval'], 'Variable can be parsed only to int, float or double.');
+
+            return $parser($request->get(substr($parameter[1], 1)));
         }
 
         return $parameter;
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ResourceBundle/Controller/ParametersParser.php
```diff
@@ -77,6 +78,16 @@ private function parseRequestValue($parameter, Request $request)
             return $this->parseRequestValueExpression(substr($parameter, 5), $request);
         }
 
+        if (0 === strpos($parameter, '!!')) {
+            $parameter = explode(' ', $parameter);
```
pamil: Using `2` as the third parameter will make it faster.

src/Sylius/Bundle/ResourceBundle/Controller/ParametersParser.php
```diff
@@ -77,6 +78,16 @@ private function parseRequestValue($parameter, Request $request)
...
+
+            $parser = trim($parameter[0], '!').'val';
```
pamil: `trim` -> `ltrim` in fact

src/Sylius/Bundle/ResourceBundle/Controller/ParametersParser.php
```diff
@@ -77,6 +78,16 @@ private function parseRequestValue($parameter, Request $request)
...
+
+            Assert::oneOf($parser, ['intval', 'floatval', 'doubleval'], 'Variable can be parsed only to int, float or double.');
```
pamil: We don't need `doubleval`, it's just an alias for `floatval`.

But we do need `boolval` here.

pjedrzejewski:
Thanks Mateusz, looks good to me!

src/Sylius/Bundle/ResourceBundle/Controller/ParametersParser.php
```diff
@@ -77,6 +78,16 @@ private function parseRequestValue($parameter, Request $request)
...
+
+            return $parser($request->get(substr($parameter[1], 1)));
```
pamil: This would fail unless the third parameter of `explode` is set to `2` (`explode(' ', 'a b c') === ['a', 'b', 'c']` -> `explode(' ', 'a b c', 2) === ['a', 'b c']`). We need some tests for data with spaces in it.

pamil: Anyway, it will fail if you try to typecast an expression like `!!float expr:abc`, can we try something like the following one instead:

```
return $parser($this->parseRequestValue($parameter[1], $request));
```

PR #8510, opened 2017-09-06
[AddressBook] Fix problem with cloning default address during checkout

### LINKED ISSUE #8168: Cannot delete address used in an order
When a shop user attempts to delete one of their addresses that has been previously used in an order, they cannot due to a foreign key constraint. Possibly this exception could be caught and archive the address when a user attempts to delete it instead (just hide it from the user). I'm not sure the best way to go about implementing this as it'll probably involve adding functionality to the resources bundle to allow for archiving.

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8168 |
| License         | MIT |


lchrusciel:
Thanks Grzesiu 🐋

PR #8511, opened 2017-09-06
[Checkout] Add missing form errors block

### LINKED ISSUE #8355: [Bug]: Form errors are not shown on shipping form.
I am creating shipping method and I had extension to the shipping type because I need some additional info from customer. In case I add `FormError` to the form on submit, I am returned to `Shipping` step of the order and I see no info (from symfony debug bar, I can see that there are validation errors in the form, but they're not shown in the ui).

https://github.com/Sylius/Sylius/blob/5ffc71b496e0c34b7a9c578557ae5746e4514a54/src/Sylius/Bundle/ShopBundle/Resources/views/Checkout/selectShipping.html.twig#L15 (missing `form_errors(form)`)

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8355 |
| License         | MIT |


stefandoorn:
Don't we get the errors now double on the address input form? Both inside the form (at each field) and at the top?

GSadee:
@stefandoorn If I am correct, these errors won't be duplicated. I've checked that to be sure and it seems to be good.

stefandoorn:
Cool!

pamil:
Thanks @GSadee! 🎉

PR #8519, opened 2017-09-06
Allow parent class type in ServiceRegistry

bpolaszek:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | none |
| License         | MIT |

Hello there,

[Sylius Registry](http://docs.sylius.org/en/latest/components/Registry/basic_usage.html) is a very useful standalone component that allow to locate services that share the same type.

This is a common need in many PHP projects, however the current implementation limits the type checking to services that share the same `interface`.

Some PHP projects or libraries rely on abstract classes instead of interfaces and type comparing with `instanceof` work as well. 

The aim of this PR is to use the `is_a()` function instead of `class_implements()` - thus we could register services sharing the same `class` as well as services sharing the same `interface`.

A simple example:
```php
$serviceRegistry = new ServiceRegistry('PDO', 'connection');
$serviceRegistry->register('master', new PDO('mysql:host=10.0.1.1', 'foo', 'bar'));
$serviceRegistry->register('slave', new PDO('mysql:host=10.0.1.2', 'foo', 'bar'));
```

Thank you,
Ben

### DIFF
--- a/src/Sylius/Component/Registry/ServiceRegistry.php
+++ b/src/Sylius/Component/Registry/ServiceRegistry.php
@@ -26,11 +26,11 @@ class ServiceRegistry implements ServiceRegistryInterface
     private $services = [];
 
     /**
-     * Interface which is required by all services.
+     * Interface or parent class which is required by all services.
      *
      * @var string
      */
-    private $interface;
+    private $className;
 
     /**
      * Human readable context for these services, e.g. "grid field"
@@ -40,12 +40,12 @@ class ServiceRegistry implements ServiceRegistryInterface
     private $context;
 
     /**
-     * @param string $interface
+     * @param string $className
      * @param string $context
      */
-    public function __construct(string $interface, string $context = 'service')
+    public function __construct(string $className, string $context = 'service')
     {
-        $this->interface = $interface;
+        $this->className = $className;
         $this->context = $context;
     }
 
@@ -70,9 +70,9 @@ class ServiceRegistry implements ServiceRegistryInterface
             throw new \InvalidArgumentException(sprintf('%s needs to be an object, %s given.', ucfirst($this->context), gettype($service)));
         }
 
-        if (!in_array($this->interface, class_implements($service), true)) {
+        if (!$service instanceof $this->className) {
             throw new \InvalidArgumentException(
-                sprintf('%s needs to implement "%s", "%s" given.', ucfirst($this->context), $this->interface, get_class($service))
+                sprintf('%s needs to be of type "%s", "%s" given.', ucfirst($this->context), $this->className, get_class($service))
             );
         }
 
--- a/src/Sylius/Component/Registry/spec/ServiceRegistrySpec.php
+++ b/src/Sylius/Component/Registry/spec/ServiceRegistrySpec.php
@@ -16,6 +16,7 @@ namespace spec\Sylius\Component\Registry;
 require_once __DIR__.'/Fixture/SampleServiceInterface.php';
 
 use PhpSpec\ObjectBehavior;
+use spec\Sylius\Component\Registry\Fixture\AbstractSampleService;
 use spec\Sylius\Component\Registry\Fixture\SampleServiceInterface;
 use Sylius\Component\Registry\ExistingServiceException;
 use Sylius\Component\Registry\NonExistingServiceException;
@@ -42,7 +43,7 @@ final class ServiceRegistrySpec extends ObjectBehavior
         $this->all()->shouldReturn([]);
     }
 
-    function it_registers_service_with_given_type(SampleServiceInterface $service): void
+    function it_registers_service_with_given_interface(SampleServiceInterface $service): void
     {
         $this->has('test')->shouldReturn(false);
         $this->register('test', $service);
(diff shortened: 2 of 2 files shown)


src/Sylius/Component/Registry/ServiceRegistry.php
```diff
@@ -70,9 +70,9 @@ public function register(string $identifier, $service): void
             throw new \InvalidArgumentException(sprintf('%s needs to be an object, %s given.', ucfirst($this->context), gettype($service)));
         }
 
-        if (!in_array($this->interface, class_implements($service), true)) {
+        if (!is_a($service, $this->className, true)) {
```
pamil: It should be possible to use `!$service instanceof $this->className` here then :)

src/Sylius/Component/Registry/spec/Fixture/AbstractSampleService.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace spec\Sylius\Component\Registry\Fixture;
+
+/**
+ * @author Beno!t POLASZEK
+ */
+abstract class AbstractSampleService
```
pamil: We can remove this fixture class and use `\stdClass` instead :)

src/Sylius/Component/Registry/ServiceRegistry.php
bpolaszek: Indeed, since `$service` is necessarily instanciated.

src/Sylius/Component/Registry/spec/Fixture/AbstractSampleService.php
bpolaszek: Good catch, it doesn't even need to be abstract.

pamil:
Side note: I'm not sure what the behaviour of `$serviceRegistry = new ServiceRegistry('PDO', 'connection');` would be if there was `Sylius\Component\Registry\PDO` class, it would probably use it instead of `\PDO`. Then, it's safer to use `\PDO::class` or `'\PDO'` 🎉

bpolaszek:
> Side note: I'm not sure what the behaviour of $serviceRegistry = new ServiceRegistry('PDO', 'connection'); would be if there was Sylius\Component\Registry\PDO class, it would probably use it instead of \PDO. Then, it's safer to use \PDO::class or '\PDO' :tada:

Of course, it was just for the example of this PR :)
Thanks for your review, I just made the changes.

pamil:
@bpolaszek can you rebase with master?

bpolaszek:
@pamil Don't make fun of me but I feel like a virgin: I've never done that :flushed: 
There's a first time for everything, for everyone, right? :no_mouth:

My git knowledge actually limits to checkout, pull, push, and merge branch in Phpstorm.
Can you please tell me (or give me a link) how to do this properly? 

Before doing this commit I pulled master.

Thank you,
Ben

> Give a man a fish and he will eat for a day. Teach him how to fish, and he will sit in a boat and drink beer all day.

lchrusciel:
Ben, you made my day. 

According to your rebase problem, normally it looks like this:
 * git checkout master
 * git pull upstream master
 * git checkout branch-to-rebase 
 * git rebase master

And now you need to resolve all your conflicts. There are a few tips along the way, but the commands you will need are:

 * git rebase --abort - if something screw over
 * git rebase --continue - if all conflicts will be resolved
 * git rebase --skip - if it appears that the commit is empty

But you have committed and open a PR from your master branch. This can be a little bit more complicated. The fastest solution would be probably to call `git rebase upstream/master`. But I'm not 100% sure about the result as I have never done it this way.

The cleanest solution would be to create a new branch with your latest changes.
 * git checkout -b new-branch-to-work-on # it will duplicate current state of your master to new branch
 * git checkout master # we will move back to the master branch
 * git reset --hard HEAD~1 # This command should not be executed unless you are sure, that you have a backup of your work. It will erase last commits, in this case only 1 (HEAD~1). We need to move only one commit back because you have only one commit in this PR.
 * git pull upstream master # this should update your local branch with the current state of upstream (main Sylius repository)
 * git checkout  new-branch-to-work-on
 * git rebase master

I have assumed, that you have to remotes for your local git repository origin(https://github.com/bpolaszek/Sylius) and upstream (https://github.com/Sylius/Sylius)
To check it, you execute this command:
 * `git remote -v`
To add a new remote
 * `git remote add upstream https://github.com/Sylius/Sylius`

Next time, when you would like to introduce some changes to any living repository, I would recommend to create a new branch before the first commit and then open a new PR from it. It will make rebase much easier
 * git checkout master
 * git pull upstream master
 * git push origin master # not mandatory
Here you can start making your changes
 * git checkout -b new-feature
Or here
 * git add -A
 * git commit -m "com"
 * git push origin new-feature

I hope you will enjoy your beer :)

lchrusciel:
Also, this [link](http://learngitbranching.js.org/) could be useful for the first minutes with git

pamil:
@lchrusciel wow!

@bpolaszek my first hundred rebases were based on looking into Evernote for a saved conversation with @pjedrzejewski and copy-pasting Git commands :D

bpolaszek:
Hi @lchrusciel @pamil,

Thanks a lot for your explanations!
I never really undertood the true difference between `git merge` and `git rebase`.

Anyway, here's what I've done so far:
- With PHPStorm, I reverted my changes the _soft_ way, that means `HEAD` goes back to the previous commit, and allows me to stash my differences. So it's basically a `git reset --hard HEAD~1` with an automatic backup.
- Then, I pulled upstream master to fetch the latest commits.
- Then, I created a branch `service-registry-improvements` and checked it out.
- Then, I unstashed my changes and committed to that branch.
- Then I clicked on `master` => `rebase onto`, which in the background, executes `git -c core.quotepath=false -c log.showSignature=false -c core.commentChar= rebase upstream/master` and got the following message: `Rebase Successful: service-registry-improvements is up-to-date with upstream/master`
- Then I pushed to `origin/service-registry-improvements` but [that commit](https://github.com/bpolaszek/Sylius/commit/9fc0d7c0dbae8e5b5d17e549da35741a7b403f3f) did not update the current PR.

What should I do now? Github suggests me to open a new PR...

Thank you,
Ben

lchrusciel:
Just open a new PR. This is how it is supposed to be. 

Also, de difference is that merge is adding commits at the top of the current branch. Rebase is coming back to the point of the split, then moving the cursor to the top of the rebased branch(usually the master) and then applying the commits from the branch you are actually on. 

I really recommend the link I have posted above, it should put a little bit of light on this topic. :)

bpolaszek:
Thank you @lchrusciel, I'll do this! :)

pjedrzejewski:
@lchrusciel 🍺

bpolaszek:
@lchrusciel @pjedrzejewski :beers:

PR #8523, opened 2017-09-06
Store datetimes in UTC

### LINKED ISSUE #8305: Model's datetime properties have no support for timezones handling
We use `datetime` Doctrine type everywhere so if we set a date `2017-07-21T12:30:00+02:30`, flush and get that object again we will get a different time `2017-07-21T12:30:00+00:00` (resetted timezone, not adjusted time). Is there anything stopping us from using `datetimetz` field type?

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8305 |
| License         | MIT |

Based on code provided by @mbabker in #8305 (thanks!) - added support for `\DateTimeImmutable`, stricter typehints and some tests.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml
@@ -29,6 +29,9 @@ doctrine:
         entity_managers:
             default:
                 auto_mapping: true
+    dbal:
+        types:
+            datetime: Sylius\Bundle\ResourceBundle\Doctrine\DBAL\Type\UTCDateTimeType
 
 doctrine_cache:
     providers:
--- a/src/Sylius/Bundle/ResourceBundle/Doctrine/DBAL/Type/UTCDateTimeType.php
+++ b/src/Sylius/Bundle/ResourceBundle/Doctrine/DBAL/Type/UTCDateTimeType.php
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ResourceBundle\Doctrine\DBAL\Type;
+
+use Doctrine\DBAL\Platforms\AbstractPlatform;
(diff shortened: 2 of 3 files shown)


pjedrzejewski:
Thanks Kamil & @mbabker!

stefandoorn:
I might oversee something, but doesn't this potentially mess up existing installations? I have my timezone settings set to `Europe/Amsterdam` and I guess therefore DateTime stuff is saved in that timezone in the database.

pjedrzejewski:
@pamil I think @stefandoorn is right. We could introduce a migration that will convert it to UTC, right? And add "timezone" setting to channel to display it properly in admin. WDYT @stefandoorn?

stefandoorn:
A migration is a must anyway I think - I don't think my customer will be happy seeing all his information being offset by 2 hours ;-) 

But, until this is being fixed, nobody can upgrade to dev-master without running into this issues. Makes the upgrade path to 1.0 harder. Might reverting until full safe fix be an idea?

Regarding the 'timezone' at channel level, not sure that is needed. I have PHP server level timezone set. A channel can easily handle different timezones if you are using channels as ways to handle multiple domains which can serve the whole world. Not sure this introduces use - might even make things complicated. The saving of dates always in UTC makes sense of course to me - as it makes a default and @mbabker's example clearly shows the current issues.

I also use 'datetime' at several custom resources, which I also should upgrade with a migration. If Sylius could handle that for all resources in the whole system, that would be great. Hard migration though I'm afraid. A class being adjusted is one, but wrong data at client installations is a catastrophe. 

Small side note: for 'createdAt' we could maybe use 'datetimetz_immutable'?

teohhanhui:
This is one of the most obvious mistakes that stand out when I installed Sylius. Is there any plan to fix this?

PR #8527, opened 2017-09-06
[Docs] Tune default Platform.sh docs based on project in production

### LINKED ISSUE #8124: Docs about running Sylius in production
There are several guides online regarding how to run Symfony & Doctrine best in production with what kind of settings. I think it would be good to gather some insights from several people (I'm not running Sylius in production yet) about what are the best settings they figured out in running Sylius specific. E.g. Doctrine metadata & query cache. If it's just general Symfony / Doctrine guidelines, then maybe we should point users to these guides in order to use.

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8124 |
| License         | MIT |

As requested by @CoderMaggie, some additional information on the Platform.sh docs based on findings in a project running Platform.sh in production.

### DIFF
--- a/docs/cookbook/platform-sh.rst
+++ b/docs/cookbook/platform-sh.rst
@@ -12,7 +12,7 @@ In this guide you will find sufficient instructions to have your application up 
 1. Prepare a Platform.sh project
 --------------------------------
 
-If you do not have it yet, go to the `Platform.sh store <https://accounts.platform.sh/platform/buy-now>`_ choose development plan
+If you do not have it yet, go to the `Platform.sh store <https://accounts.platform.sh/platform/buy-now>`_, choose development plan
 and go through checkout. Then, when you will have a project ready, give it a name and proceed to ``Import an existing site``.
 
 .. tip::
@@ -24,16 +24,16 @@ and go through checkout. Then, when you will have a project ready, give it a nam
 
 * Add the ``.platform.app.yaml`` file at the root of your project repository
 
-This is how this file should look like for Sylius:
+This is how this file should look like for Sylius (tuned version of the default Platform.sh example):
 
 .. code-block:: yaml
 
     # .platform.app.yaml
     name: sylius
 
-    type: "php:7.0"
+    type: "php:7.1"
     build:
-        flavor: symfony
+        flavor: composer
 
     relationships:
         database: "mysql:mysql"
@@ -45,6 +45,11 @@ This is how this file should look like for Sylius:
             - igbinary
             - memcached
             - redis
+
+    dependencies:
+        nodejs:
+            yarn: "*"
+            gulp-cli: "*"
 
     web:
         document_root: "/web"
@@ -103,7 +108,7 @@ This is how this file should look like for Sylius:
           - \.html$
           - \.pdf$
 
-    disk: 2048
+    disk: 4096
 
     mounts:
         "/var/cache": "shared:files/cache"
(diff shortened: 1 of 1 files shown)


stefandoorn:
P.S. Some parts of the example code have been removed, e.g. the cronjobs that inject the fixtures every X hours (is from the demo store).

CoderMaggie:
To me it looks great, but I will wait for @pjedrzejewski to approve it :)

docs/cookbook/platform-sh.rst
```diff
@@ -151,11 +153,11 @@ This file will load ``mysql`` and ``redis`` on your Platform.sh server.
 
     # .platform/services.yaml
     mysql:
-        type: mysql
-        disk: 2048
+        type: mysql:10.0
```
lchrusciel: Doesn't this value go along with `sever_version` of doctrine dbal? And we have [5.5](https://github.com/Sylius/Sylius/blob/master/app/config/config.yml#L35) right now in a config file.

docs/cookbook/platform-sh.rst
```diff
@@ -117,18 +122,15 @@ This is how this file should look like for Sylius:
             rm web/app_test.php
             rm web/app_test_cached.php
             rm -rf var/cache/*
-            php bin/console --env=prod --no-debug --ansi cache:clear
+            php bin/console --env=prod --no-debug --ansi cache:clear --no-warmup
+            php bin/console --env=prod --no-debug --ansi cache:warmup
             php bin/console --env=prod --no-debug --ansi assets:install
-            npm install
-            npm run gulp
+            php bin/console --env=prod --no-debug --ansi theme:assets:install
```
lchrusciel: Shouldn't it be optional only for users of theme bundle? We can just mention about it

docs/cookbook/platform-sh.rst
```diff
@@ -24,16 +24,16 @@ and go through checkout. Then, when you will have a project ready, give it a nam
...
 .. code-block:: yaml
 
     # .platform.app.yaml
     name: sylius
 
-    type: "php:7.0"
+    type: "php:7.1"
     build:
-        flavor: symfony
+        flavor: composer
```
lchrusciel: Did you notice anything when made this change?

stefandoorn: No, but Platform.sh defines 'symfony' as deprecated and advices 'composer', therefore :)

docs/cookbook/platform-sh.rst
stefandoorn: Yes, can make a comment on it indeed, it's kind of optional. Actually I don't know what happens if you don't use a theme if you run this command..

docs/cookbook/platform-sh.rst
stefandoorn: Hm, good catch. I will check that, as I have `server_version` at 5.6. I'll check that one tomorrow.

docs/cookbook/platform-sh.rst
stefandoorn: In here they use 'composer' also: https://docs.platform.sh/frameworks/symfony.html. The Symfony docs itself are behind it seems.

stefandoorn: Ah here is actually the commit I was searching for: https://github.com/platformsh/platformsh-example-symfony/commit/3de359c88fb5316ff653f0142428e658880f1b65#diff-fe14053ff51cb358f8b709415b2a2d17. Looking at the commit message, it seems 'symfony' is an alias of 'composer'. It's probably more safe to go with 'composer'.

stefandoorn:
Feedback processed!

lchrusciel:
Thanks Stefan! It will help us a lot :)

stefandoorn:
You're welcome, happy to help!

PR #8528, opened 2017-09-06
[Docs] Add advanced Platform.sh documentation topics

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | addition to #8527 |
| License         | MIT |

As an addition to #8527, some advanced topics in a separate file when using Sylius on Platform.sh in production. The default configuration is good to go, but these things have been implemented in my project and they do help / improve performance.


stefandoorn:
@CoderMaggie I have no clue how to fix that error on the build, could you have a look?

docs/cookbook/platform-sh-advanced.rst
```diff
@@ -0,0 +1,194 @@
+Advanced topics
+---------------
```
CoderMaggie: I think that the build is failing because this document does not have any title.
Let's do:
```
Advanced Platform.sh configurations
===========================
```
:)

stefandoorn:
The errors are different @CoderMaggie, but I don't understand what's wrong with the lines. Anyway, if you could fix them, that would be great.

docs/cookbook/platform-sh-advanced.rst
```diff
@@ -0,0 +1,194 @@
+Advanced Platform.sh configurations
+===================================
+
+The basic set-up let's you easily set-up a Platform.sh project running your Sylius project. It should give you an
```
lchrusciel: Shouldn't we try to avoid repetition of the same words in one sentence? `Platform.sh platform` / `Platform.sh deployment server` / `Platform.sh PaaS`?

docs/cookbook/platform-sh-advanced.rst
```diff
@@ -0,0 +1,194 @@
...
+
+Want to use the metacache, query cache or result cache Symfony and Doctrine have to offer? It comes with a caveat.
+Platform.sh doesn't allow you to connect to all your services yet from inside the `build` hook. The following
+tutorial will guide you through this and make use of Redis. In the default example Redis is already activated.
+
+* On top of your ``app/config/parameters_platform.php`` add:
+
+.. code-block:: php
+
+    $container->setParameter('metacache_driver', 'array');
```
lchrusciel: Won't it be better to have this values as default parameters in parameters.yml.dist file? Also, it should allow building your app outside platform.sh

docs/cookbook/platform-sh-advanced.rst
stefandoorn: Sure. Makes sense to come up with one single name and use that one everywhere :-)

docs/cookbook/platform-sh-advanced.rst
stefandoorn: Haven't thought about that. I think it's safe to move them in there, isn't it? I don't see downsides atleast..

CoderMaggie:
@stefandoorn will you have a second to fix what @lchrusciel suggested and make a rebase or should I take care of that?

lchrusciel:
Such a nice coworking! Thanks Magda and Stefen! 🥇

pjedrzejewski:
@stefandoorn 🍻

PR #8533, opened 2017-09-07
show original price in listing + product sheet

### LINKED ISSUE #8258: Original price not displayed in product listing
Create a product variant with price and original price.
On product listing only the price is being displayed, original price should also be displayed crossed out.

I'm ready to contribute if you see it's importance.

Snake-Tn:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | fixes #8258 |
| License         | MIT |

### DIFF
--- a/features/product/viewing_products/viewing_products_original_price.feature
+++ b/features/product/viewing_products/viewing_products_original_price.feature
@@ -1,6 +1,6 @@
 @viewing_products
 Feature: Viewing a product original price
-    In order to see products original price
+    In order to be informed about the amount of discount
     As a Visitor
     I want to be able to view a single product original price
 
@@ -9,6 +9,7 @@ Feature: Viewing a product original price
 
     @ui
     Scenario: Viewing a detailed page with product's original price
-        Given the store has a product "T-shirt banana" originally priced at "$59.00"
+        Given the store has a product "T-shirt banana" priced at "$49.00"
+        And this product has been originally priced at "$59.00"
         When I check this product's details
         Then I should see the product original price "$59.00"
--- a/features/product/viewing_products/viewing_products_original_price.feature
+++ b/features/product/viewing_products/viewing_products_original_price.feature
@@ -1,0 +1,14 @@
+@viewing_products
+Feature: Viewing a product original price
+    In order to see products original price
+    As a Visitor
+    I want to be able to view a single product original price
+
+    Background:
+        Given the store operates on a single channel in "United States"
+
+    @ui
+    Scenario: Viewing a detailed page with product's original price
+        Given the store has a product "T-shirt banana" originally priced at "$59.00"
+        When I check this product's details
+        Then I should see the product original price "$59.00"
--- a/src/Sylius/Behat/Context/Setup/ProductContext.php
+++ b/src/Sylius/Behat/Context/Setup/ProductContext.php
@@ -197,13 +197,13 @@ final class ProductContext implements Context
     }
 
     /**
-     * @Given /^the store(?:| also) has a product "([^"]+)" originally priced at ("[^"]+")$/
+     * @Given /^(this product) has been originally priced at ("[^"]+")$/
      */
-    public function storeHasAProductOriginallyPricedAt($productName, $originalPrice = 100, ChannelInterface $channel = null)
+    public function thisProductIsAlsoOriginallyPricedAtInChannel(ProductInterface $product, $originalPrice = 100, ChannelInterface $channel = null)
     {
-        $product = $this->createProduct($productName, 0, $channel);
-        $product->getVariants()->first()->getChannelPricings()->first()->setOriginalPrice($originalPrice);
-        $this->saveProduct($product);
+        $productVariant = $this->defaultVariantResolver->getVariant($product);
+        $productVariant->getChannelPricings()->first()->setOriginalPrice($originalPrice);
+        $this->objectManager->flush();
     }
 
     /**
--- a/src/Sylius/Behat/Context/Setup/ProductContext.php
+++ b/src/Sylius/Behat/Context/Setup/ProductContext.php
@@ -197,6 +197,16 @@ final class ProductContext implements Context
     }
 
     /**
+     * @Given /^the store(?:| also) has a product "([^"]+)" originally priced at ("[^"]+")$/
+     */
+    public function storeHasAProductOriginallyPricedAt($productName, $originalPrice = 100, ChannelInterface $channel = null)
+    {
+        $product = $this->createProduct($productName, 0, $channel);
+        $product->getVariants()->first()->getChannelPricings()->first()->setOriginalPrice($originalPrice);
+        $this->saveProduct($product);
+    }
+
+    /**
      * @Given /^(this product) is also priced at ("[^"]+") in ("[^"]+" channel)$/
      */
     public function thisProductIsAlsoPricedAtInChannel(ProductInterface $product, $price, ChannelInterface $channel)
--- a/src/Sylius/Behat/Context/Ui/Shop/ProductContext.php
+++ b/src/Sylius/Behat/Context/Ui/Shop/ProductContext.php
@@ -292,6 +292,14 @@ final class ProductContext implements Context
     }
 
     /**
+     * @Then I should see the product original price :originalPrice
+     */
+    public function iShouldSeeTheProductOriginalPrice($originalPrice)
+    {
+        Assert::same($this->showPage->getOriginalPrice(), $originalPrice);
+    }
+
+    /**
      * @When I set its :optionName to :optionValue
      */
     public function iSetItsOptionTo($optionName, $optionValue)
--- a/src/Sylius/Behat/Page/Shop/Product/ShowPage.php
+++ b/src/Sylius/Behat/Page/Shop/Product/ShowPage.php
@@ -178,6 +178,14 @@ class ShowPage extends SymfonyPage implements ShowPageInterface
     /**
      * {@inheritdoc}
      */
+    public function getOriginalPrice()
+    {
+        return $this->getElement('product_original_price')->getText();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function countReviews()
     {
         return count($this->getElement('reviews')->findAll('css', '.comment'));
@@ -318,6 +326,7 @@ class ShowPage extends SymfonyPage implements ShowPageInterface
             'option_select' => '#sylius_add_to_cart_cartItem_variant_%option-name%',
             'out_of_stock' => '#sylius-product-out-of-stock',
             'product_price' => '#product-price',
+            'product_original_price' => '#product-original-price',
             'reviews' => '[data-tab="reviews"] .comments',
             'selecting_variants' => '#sylius-product-selecting-variant',
             'validation_errors' => '.sylius-validation-error',
--- a/src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
+++ b/src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
@@ -63,14 +63,13 @@ class PriceHelper extends Helper
     /**
      * {@inheritdoc}
      */
-    public function getOriginalPrice(ProductVariantInterface $productVariant, array $context): int
+    public function getOriginalPrice(ProductVariantInterface $productVariant, array $context): ?int
     {
         Assert::keyExists($context, 'channel');
 
         return $this
             ->productVariantOriginalPriceCalculator
-            ->calculate($productVariant, $context)
-            ;
+            ->calculate($productVariant, $context);
     }
 
     /**
--- a/src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
+++ b/src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
@@ -29,11 +29,20 @@ class PriceHelper extends Helper
     private $productVariantPriceCalculator;
 
     /**
-     * @param ProductVariantPriceCalculatorInterface $productVariantPriceCalculator
+     * @var ProductVariantPriceCalculatorInterface
      */
-    public function __construct(ProductVariantPriceCalculatorInterface $productVariantPriceCalculator)
-    {
+    private $productVariantOriginalPriceCalculator;
+
+    /**
+     * @param ProductVariantPriceCalculatorInterface $productVariantPriceCalculator
+     * @param ProductVariantPriceCalculatorInterface $productVariantOriginalPriceCalculator
+     */
+    public function __construct(
+        ProductVariantPriceCalculatorInterface $productVariantPriceCalculator,
+        ProductVariantPriceCalculatorInterface $productVariantOriginalPriceCalculator
+    ) {
         $this->productVariantPriceCalculator = $productVariantPriceCalculator;
+        $this->productVariantOriginalPriceCalculator = $productVariantOriginalPriceCalculator;
     }
 
     /**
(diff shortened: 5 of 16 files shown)


features/product/viewing_products/viewing_products_original_price.feature
```diff
@@ -0,0 +1,14 @@
+@viewing_products
+Feature: Viewing a product original price
+    In order to see products original price
```
lchrusciel: `In order to be informed about the amount of discount`?

src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -199,6 +199,16 @@ public function storeHasAProductPricedAt($productName, $price = 100, ChannelInte
     }
 
     /**
+     * @Given /^the store(?:| also) has a product "([^"]+)" originally priced at ("[^"]+")$/
```
lchrusciel: Do we need so generic step?

features/product/viewing_products/viewing_products_original_price.feature
```diff
@@ -0,0 +1,14 @@
...
+    As a Visitor
+    I want to be able to view a single product original price
+
+    Background:
+        Given the store operates on a single channel in "United States"
+
+    @ui
+    Scenario: Viewing a detailed page with product's original price
+        Given the store has a product "T-shirt banana" originally priced at "$59.00"
```
lchrusciel:
```gherkin
Given the store has a product "T-shirt banana" priced at "$49.00"
And this product has been originally priced at "$59.00"
```
?

src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -199,6 +199,16 @@ public function storeHasAProductPricedAt($productName, $price = 100, ChannelInte
...
+     */
+    public function storeHasAProductOriginallyPricedAt($productName, $originalPrice = 100, ChannelInterface $channel = null)
```
lchrusciel: The channel parameter is not needed. 

Also, if you will use the proposal from one comment above, we can use a shared storage transformer.
```php
     /**
       * @Given /^(this product) has been originally priced at ("[^"]+")$/
       */
      public function thisProductIsAlsoPricedAtInChannel(ProductInterface $product, $price)
     {
         /** @var ProductVariantInterface $productVariant */
         $productVariant = $this->defaultVariantResolver->getVariant($product);

         // For all channel pricings for example
 
         $this->objectManager->flush();
     }
```

src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
```diff
@@ -54,6 +63,19 @@ public function getPrice(ProductVariantInterface $productVariant, array $context
...
      * {@inheritdoc}
      */
+    public function getOriginalPrice(ProductVariantInterface $productVariant, array $context): int
+    {
+        Assert::keyExists($context, 'channel');
+
+        return $this
+            ->productVariantOriginalPriceCalculator
+            ->calculate($productVariant, $context)
+            ;
```
lchrusciel: `;` should be on level of `return`

Snake-Tn:
@lchrusciel , Thanks for your help.
I fixed issues related to behat features as you mentioned. 
Can you re-check please ?

lchrusciel:
@Snake-Tn I will try to do it in nearest days :)

src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
```diff
@@ -29,11 +29,20 @@ class PriceHelper extends Helper
...
+
+    /**
      * @param ProductVariantPriceCalculatorInterface $productVariantPriceCalculator
+     * @param ProductVariantPriceCalculatorInterface $productVariantOriginalPriceCalculator
      */
-    public function __construct(ProductVariantPriceCalculatorInterface $productVariantPriceCalculator)
-    {
+    public function __construct(
+        ProductVariantPriceCalculatorInterface $productVariantPriceCalculator,
+        ProductVariantPriceCalculatorInterface $productVariantOriginalPriceCalculator
```
lchrusciel: I would create a new interface to split the logic

src/Sylius/Bundle/ShopBundle/Resources/views/Product/Show/_price.html.twig
```diff
@@ -1,5 +1,8 @@
 {% import "@SyliusShop/Common/Macro/money.html.twig" as money %}
 
+<span class="ui small header" style="text-decoration: line-through" id="product-original-price">
```
lchrusciel: Can we determine if we even want to add this html block?

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
```diff
@@ -25,7 +25,7 @@
     /**
      * {@inheritdoc}
      */
-    public function calculate(ProductVariantInterface $productVariant, array $context): int
+    public function calculate(ProductVariantInterface $productVariant, array $context): ?int
```
lchrusciel: Can we revert this change?

src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
Snake-Tn: Yes it would be better since original price is nullble and price is not. So they should not share the same calculator interface .

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
Snake-Tn: Yes, after creating a new interface for original price .

src/Sylius/Bundle/ShopBundle/Resources/views/Product/Show/_price.html.twig
Snake-Tn: You mean in case original price is null or higher than the price, we should not print this HTML?
Currently this login is hold be a twig macro.

src/Sylius/Behat/Page/Shop/Product/ShowPage.php
```diff
@@ -318,6 +326,7 @@ protected function getDefinedElements()
             'option_select' => '#sylius_add_to_cart_cartItem_variant_%option-name%',
             'out_of_stock' => '#sylius-product-out-of-stock',
             'product_price' => '#product-price',
+            'product_original_price' => '#product-original-price',
```
GSadee: Wrong order in the array

src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
```diff
@@ -54,6 +63,18 @@ public function getPrice(ProductVariantInterface $productVariant, array $context
     /**
      * {@inheritdoc}
      */
+    public function getOriginalPrice(ProductVariantInterface $productVariant, array $context): ?int
+    {
+        Assert::keyExists($context, 'channel');
+
+        return $this
+            ->productVariantOriginalPriceCalculator
+            ->calculate($productVariant, $context);
```
GSadee: Move `;` to the next line

src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
gabiudrescu: why not set the original price equal to selling price and respect the same interface?

Snake-Tn: I'm agree that null values are no more than a door for bugs to come from. And want to avoid them as possible. In best practice a method should not has multi return type. 
It's possible to update the way how a product get persisted to assure original price has an initial value(equal price). Then there is the problem of data migration to deal with also.(in order to update current products)
This is big decisions should be validated with sylius maintainers I think .

Snake-Tn: I'm agree that null values are no more than a door for bugs to come from. And want to avoid them as possible. In best practice a method should not has multi return type. 
It's possible to update the way how a product get persisted to assure original price has an initial value(equal price). Then there is the problem of data migration to deal with also.(in order to update current products)
This is big decisions should be validated with sylius maintainers I think .

src/Sylius/Behat/Page/Shop/Product/ShowPage.php
Snake-Tn: Why order is wrong , what is the correct order ?

Snake-Tn:
@lchrusciel , regarding price/original price calculator. 
We have two options : 

1. Create a separate interface for original price.
2. Original price is per default equal to the price [@gabiudrescu idea] (then original price will never be null, and it can share same interface with price)

What option do you prefer to implement ?

src/Sylius/Behat/Page/Shop/Product/ShowPage.php
GSadee: `product_original_price` key should be before `product_price` key :)

src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -197,6 +197,16 @@ public function storeHasAProductPricedAt($productName, $price = 100, ChannelInte
     }
 
     /**
+     * @Given /^(this product) has been originally priced at ("[^"]+")$/
+     */
+    public function thisProductIsAlsoOriginallyPricedAtInChannel(ProductInterface $product, $originalPrice = 100, ChannelInterface $channel = null)
```
Zales0123: `ChannelInterface $channel = null` is not needed.

Zales0123: Also, although it's they're not used widely in **Behat** yet, we should start using strict scalar types, as we support only PHP 7.1 -> so `int $originalPrice` (I don't think the default value is needed) and `: void` at the end.

src/Sylius/Behat/Context/Ui/Shop/ProductContext.php
```diff
@@ -292,6 +292,14 @@ public function iShouldSeeTheProductPrice($price)
     }
 
     /**
+     * @Then I should see the product original price :originalPrice
+     */
+    public function iShouldSeeTheProductOriginalPrice($originalPrice)
```
Zales0123: Same here, and everywhere else ;)

src/Sylius/Behat/Page/Shop/Product/ShowPage.php
Zales0123: Alphabetical order is life 🔤

src/Sylius/Bundle/ShopBundle/Resources/views/Common/Macro/money.html.twig
```diff
@@ -13,3 +13,12 @@
 
     {{- convertAndFormat(variant|sylius_calculate_price({'channel': sylius.channel})) }}
 {%- endmacro -%}
+
+{%- macro calculateOriginalPrice(variant) -%}
+    {% from _self import convertAndFormat %}
+    {% set originalPrice = variant|sylius_calculate_original_price({'channel': sylius.channel})  %}
+    {% set price = variant|sylius_calculate_price({'channel': sylius.channel})  %}
+    {% if( originalPrice > price) %}
```
Zales0123: Missing space after `if` and additional space before `originalPrice`

src/Sylius/Bundle/ShopBundle/Resources/views/Product/Show/_price.html.twig
Zales0123: I think this condition should be put here - we don't need a redundant html with no text in it :)

src/Sylius/Component/Core/spec/Calculator/ProductVariantOriginalPriceCalculatorSpec.php
```diff
@@ -0,0 +1,64 @@
...
+namespace spec\Sylius\Component\Core\Calculator;
+
+use Sylius\Component\Core\Calculator\ProductVariantOriginalPriceCalculator;
+
+/**
+ * @author Ahmed Kooli <kooliahmd@gmail.com>
+ */
+class ProductVariantOriginalPriceCalculatorSpec
+{
+    function it_is_initializable()
```
Zales0123: We've decided not to use `it_is_initializable()` functions some time ago

src/Sylius/Component/Core/spec/Calculator/ProductVariantOriginalPriceCalculatorSpec.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+namespace spec\Sylius\Component\Core\Calculator;
+
+use Sylius\Component\Core\Calculator\ProductVariantOriginalPriceCalculator;
+
+/**
+ * @author Ahmed Kooli <kooliahmd@gmail.com>
+ */
+class ProductVariantOriginalPriceCalculatorSpec
```
Zales0123: Missing `final`

src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
Zales0123: Even though we decide to implement this feature, still `$originalPrice` and `$price` are slightly different concepts and I think it should be useful to have different interfaces for their calculators.

gabiudrescu:
kind reminder

src/Sylius/Bundle/ShopBundle/Resources/views/Common/Macro/money.html.twig
```diff
@@ -13,3 +13,12 @@
...
+        {{- convertAndFormat(variant|sylius_calculate_original_price({'channel': sylius.channel})) }}
```
pamil: `{{- convertAndFormat(originalPrice) -}}`, no need to calculate it twice

src/Sylius/Bundle/ShopBundle/Resources/views/Product/Show/_price.html.twig
pamil:
```
{% set originalPrice = money.calculateOriginalPrice(product|sylius_resolve_variant) %}
{% if originalPrice is not null %}
    <span>{{ originalPrice }} </span>
{% endif %}
```

Should do the work

src/Sylius/Bundle/ShopBundle/Resources/views/Product/_box.html.twig
```diff
@@ -16,6 +16,7 @@
     <div class="content">
         <a href="{{ path('sylius_shop_product_show', {'slug': product.slug, '_locale': product.translation.locale}) }}" class="header sylius-product-name">{{ product.name }}</a>
         {% if not product.variants.empty() %}
+            <div class="sylius-product-original-price" style="text-decoration: line-through">{{ money.calculateOriginalPrice(product|sylius_resolve_variant) }}</div>
```
pamil: Same here as above (with the excess HTML being generated)

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
pamil: Yeah, this needs to be reverted in order to keep BC.

src/Sylius/Component/Core/Calculator/ProductVariantOriginalPriceCalculator.php
```diff
@@ -0,0 +1,44 @@
...
+
+        if (null === $channelPricing) {
+            throw new MissingChannelConfigurationException(sprintf(
+                'Channel %s has no price defined for product variant %s',
+                $context['channel']->getName(),
+                $productVariant->getName()
+            ));
+        }
+
+        return $channelPricing->getOriginalPrice();
```
pamil: Another solution in order to avoid making the return typehint nullable could be to return the regular price if there's no original price.

Issue #8254, opened 2017-07-05
Impersonate user from admin authenticate previously in shop, fails assignin customer shopping


jdeveloper:
Scenario:

I login as a normal user in the shop and begin to add items to the cart.

If now I authenticate in the admin, pick a customer and press impersonate and go back to the shop area, the user authenticated is in fact the impersonated user. But the impersonated user has now the cart of the previous user so, when the user end with checkout, the associated user is the one of the beginning.

Related issue: #8127 

See comments in the related issue where it is discussed a possible solution-

pjedrzejewski:
Fixed, thanks for reporting!

PR #8543, opened 2017-09-08
[Admin] Add listener to resolve cart of an impersonated customer

### LINKED ISSUE #8127: Can't simulate order switching user
The problem is the following.

I log in as a user that has role to switch user. Then I switch user with ?_switch_user and make an order but when I finish the order is places with the customer of the original auntheticated user, not the switched user.

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8127 and #8254 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Context/SessionAndChannelBasedCartContext.php
+++ b/src/Sylius/Bundle/CoreBundle/Context/SessionAndChannelBasedCartContext.php
@@ -16,6 +16,7 @@ namespace Sylius\Bundle\CoreBundle\Context;
 use Sylius\Component\Channel\Context\ChannelContextInterface;
 use Sylius\Component\Channel\Context\ChannelNotFoundException;
 use Sylius\Component\Core\Repository\OrderRepositoryInterface;
+use Sylius\Component\Core\Storage\CartStorageInterface;
 use Sylius\Component\Order\Context\CartContextInterface;
 use Sylius\Component\Order\Context\CartNotFoundException;
 use Sylius\Component\Order\Model\OrderInterface;
@@ -27,14 +28,9 @@ use Symfony\Component\HttpFoundation\Session\SessionInterface;
 final class SessionAndChannelBasedCartContext implements CartContextInterface
 {
     /**
-     * @var SessionInterface
+     * @var CartStorageInterface
      */
-    private $session;
-
-    /**
-     * @var string
-     */
-    private $sessionKeyName;
+    private $cartSessionStorage;
 
     /**
      * @var OrderRepositoryInterface
@@ -47,19 +43,16 @@ final class SessionAndChannelBasedCartContext implements CartContextInterface
     private $channelContext;
 
     /**
-     * @param SessionInterface $session
-     * @param string $sessionKeyName
+     * @param CartStorageInterface $cartSessionStorage
      * @param ChannelContextInterface $channelContext
      * @param OrderRepositoryInterface $orderRepository
      */
     public function __construct(
-        SessionInterface $session,
-        string $sessionKeyName,
+        CartStorageInterface $cartSessionStorage,
         ChannelContextInterface $channelContext,
         OrderRepositoryInterface $orderRepository
     ) {
-        $this->session = $session;
-        $this->sessionKeyName = $sessionKeyName;
+        $this->cartSessionStorage = $cartSessionStorage;
         $this->channelContext = $channelContext;
         $this->orderRepository = $orderRepository;
     }
@@ -75,17 +68,17 @@ final class SessionAndChannelBasedCartContext implements CartContextInterface
             throw new CartNotFoundException(null, $exception);
         }
 
-        if (!$this->session->has(sprintf('%s.%s', $this->sessionKeyName, $channel->getCode()))) {
+        if (!$this->cartSessionStorage->hasCartId($channel->getCode())) {
             throw new CartNotFoundException('Sylius was not able to find the cart in session');
         }
 
         $cart = $this->orderRepository->findCartByChannel(
-            $this->session->get(sprintf('%s.%s', $this->sessionKeyName, $channel->getCode())),
+            $this->cartSessionStorage->getCartId($channel->getCode()),
             $channel
         );
 
         if (null === $cart) {
-            $this->session->remove(sprintf('%s.%s', $this->sessionKeyName, $channel->getCode()));
+            $this->cartSessionStorage->removeCartId($channel->getCode());
 
             throw new CartNotFoundException('Sylius was not able to find the cart in session');
         }
--- a/src/Sylius/Bundle/CoreBundle/Context/SessionAndChannelBasedCartContext.php
+++ b/src/Sylius/Bundle/CoreBundle/Context/SessionAndChannelBasedCartContext.php
@@ -15,12 +15,10 @@ namespace Sylius\Bundle\CoreBundle\Context;
 
 use Sylius\Component\Channel\Context\ChannelContextInterface;
 use Sylius\Component\Channel\Context\ChannelNotFoundException;
-use Sylius\Component\Core\Repository\OrderRepositoryInterface;
 use Sylius\Component\Core\Storage\CartStorageInterface;
 use Sylius\Component\Order\Context\CartContextInterface;
 use Sylius\Component\Order\Context\CartNotFoundException;
 use Sylius\Component\Order\Model\OrderInterface;
-use Symfony\Component\HttpFoundation\Session\SessionInterface;
 
 /**
  * @author Anna Walasek <anna.walasek@lakion.com>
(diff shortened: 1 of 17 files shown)


src/Sylius/Bundle/CoreBundle/EventListener/UserImpersonatedListener.php
```diff
@@ -0,0 +1,86 @@
...
+        $this->session = $session;
+        $this->sessionKeyName = $sessionKeyName;
+        $this->channelContext = $channelContext;
+        $this->orderRepository = $orderRepository;
+    }
+
+    /**
+     * @param UserEvent $event
+     */
+    public function userImpersonated(UserEvent $event): void
```
pamil: `userImpersonated` -> `onUserImpersonated` / `__invoke`?

src/Sylius/Bundle/CoreBundle/EventListener/UserImpersonatedListener.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $customer = $event->getUser()->getCustomer();
+
+        /** @var ChannelInterface $channel */
+        $channel = $this->channelContext->getChannel();
+
+        /** @var OrderInterface $cart */
+        $cart = $this->orderRepository->findLatestCartByChannelAndCustomer($channel, $customer);
+
+        $sessionCartKey = sprintf('%s.%s', $this->sessionKeyName, $channel->getCode());
```
pamil: It looks like a part of session storage part is leaking out, could we introduce a unified interface to get / set / remove a cart without knowing where it is stored?

src/Sylius/Bundle/CoreBundle/spec/EventListener/UserImpersonatedListenerSpec.php
```diff
@@ -0,0 +1,83 @@
...
+    function it_sets_cart_id_of_an_impersonated_customer_in_session(
+        SessionInterface $session,
+        ChannelContextInterface $channelContext,
+        OrderRepositoryInterface $orderRepository,
+        UserEvent $event,
+        ShopUserInterface $user,
+        CustomerInterface $customer,
+        ChannelInterface $channel,
+        OrderInterface $cart
+
```
pamil: oops

src/Sylius/Bundle/CoreBundle/spec/EventListener/UserImpersonatedListenerSpec.php
```diff
@@ -0,0 +1,83 @@
...
+    function it_removes_the_current_cart_id_if_an_impersonated_customer_has_no_cart(
+        SessionInterface $session,
+        ChannelContextInterface $channelContext,
+        OrderRepositoryInterface $orderRepository,
+        UserEvent $event,
+        ShopUserInterface $user,
+        CustomerInterface $customer,
+        ChannelInterface $channel
+
```
pamil: oops

src/Sylius/Bundle/CoreBundle/Security/UserImpersonator.php
```diff
@@ -51,5 +61,7 @@ public function impersonate(UserInterface $user): void
 
         $this->session->set($this->sessionTokenParameter, serialize($token));
```
lchrusciel: If we apply @pamil 's [comment](https://github.com/Sylius/Sylius/pull/8543/files#r137765882) we should also adjust this class

GSadee: I guess that we could think about that later, because it is a slightly different case.

src/Sylius/Bundle/CoreBundle/Context/SessionAndChannelBasedCartContext.php
```diff
@@ -27,14 +28,9 @@
...
-     * @var SessionInterface
+     * @var CartStorageInterface
      */
-    private $session;
-
-    /**
-     * @var string
-     */
-    private $sessionKeyName;
+    private $cartSessionStorage;
```
lchrusciel: `cartSessionStorage` -> `cartStorage`

src/Sylius/Bundle/CoreBundle/Context/SessionAndChannelBasedCartContext.php
```diff
@@ -75,17 +68,17 @@ public function getCart(): OrderInterface
             throw new CartNotFoundException(null, $exception);
         }
 
-        if (!$this->session->has(sprintf('%s.%s', $this->sessionKeyName, $channel->getCode()))) {
+        if (!$this->cartSessionStorage->hasCartId($channel->getCode())) {
```
lchrusciel: What would you say about passing whole object and naming of method `hasCartIdForChannel`?

src/Sylius/Bundle/CoreBundle/EventListener/UserImpersonatedListener.php
```diff
@@ -0,0 +1,77 @@
...
+use Sylius\Component\Core\Repository\OrderRepositoryInterface;
+use Sylius\Component\Core\Storage\CartStorageInterface;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+
+final class UserImpersonatedListener
+{
+    /**
+     * @var CartStorageInterface
+     */
+    private $cartSessionStorage;
```
lchrusciel: `cartSessionStorage` -> `cartStorage`

src/Sylius/Bundle/CoreBundle/Security/UserImpersonator.php
lchrusciel: Maybe we should open an issue with this after merging? Just to not lost the track of the problem

pamil: That's nothing related to cart storage :)

src/Sylius/Bundle/CoreBundle/spec/Storage/CartSessionStorageSpec.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $this->setCartIdForChannel($channel, 14);
+    }
+
+    function it_returns_a_cart_id_from_a_session(SessionInterface $session, ChannelInterface $channel): void
+    {
+        $channel->getCode()->willReturn('channel_code');
+        $session->get('session_key_name.channel_code')->shouldBeCalled();
+
+        $this->getCartIdForChannel($channel);
```
pamil: We should assert the return value of this method.

pamil: Anyway, what if there is no cart for that channel?

src/Sylius/Component/Core/Storage/CartStorageInterface.php
```diff
@@ -0,0 +1,44 @@
...
+     * @param ChannelInterface $channel
+     *
+     * @return bool
+     */
+    public function hasCartIdForChannel(ChannelInterface $channel): bool;
+
+    /**
+     * @param ChannelInterface $channel
+     *
+     * @return mixed
```
pamil: We need to make it more stricter, either keep it as `getCartIdForChannel` and define an exception to throw when cart is not found or rename it to `findCartIdForChannel` and return null when there's no cart for a given channel.

src/Sylius/Component/Core/Storage/CartStorageInterface.php
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * @param ChannelInterface $channel
+     * @param mixed $cartId
+     */
+    public function setCartIdForChannel(ChannelInterface $channel, $cartId): void;
+
+    /**
+     * @param ChannelInterface $channel
+     */
+    public function removeCartIdFromChannel(ChannelInterface $channel): void;
```
pamil: The "from channel" part looks kinda wrong to me, I understand it like it implies that channel has cart it assigned. Why not "for channel" instead - `removeCartIdForChannel`?

src/Sylius/Component/Core/Storage/CartStorageInterface.php
```diff
@@ -0,0 +1,44 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\Storage;
+
+use Sylius\Component\Core\Model\ChannelInterface;
+
+interface CartStorageInterface
```
pamil: Just a thought: we can make it accept / return `OrderInterface` instances instead of cart ids so we do not have to worry about garbage being put inside this storage & we put `OrderInterface` instead of `mixed`. Then we can skip the `CartId` part from method names completely and end up with them like `getForChannel(ChannelInterface $channel): OrderInterface`.

pjedrzejewski:
Thank you Grzesiu, nice work!

Issue #7305, opened 2017-01-20
Event Consistency


mbabker:
So today I'm adding a feature for my app somewhat unrelated to the Sylius inner workings, but part of what I'm doing has me hooking into the various controller events to catch certain activities.  Looking at the base ResourceController, I'm noticing some potential inconsistencies.  Any particular reason for this?  Ideally I'd suggest things be consistent across the board and an event is always dispatched before handling the view, but this might require a B/C breaking change in `Sylius\Bundle\ResourceBundle\Controller\EventDispatcherInterface` to allow a null resource for the index action or dispatching the event with the result of the `Sylius\Bundle\ResourceBundle\Controller\ResourcesCollectionProviderInterface::get()` call which isn't a singular resource.

- `showAction`
    - Dispatches a show event before creating the view
- `indexAction`
    - No event dispatched
- `createAction`
    - pre/post events dispatched when POSTing data
    - No event dispatched for rendering a view (GET request)
- `updateAction`
    - pre/post events dispatched when submitting data (POST, PUT, PATCH requests)
    - No event dispatched for rendering a view (GET request)
- `deleteAction`
    - pre/post events dispatched

ste93cry:
This is somehow related to #7804

mvar:
I see that `updateAction()` now supports custom response by setting it to the event ([code here](https://github.com/Sylius/Sylius/commit/ca4412060c375832ce63d527ad3a24585d7ccd4c#diff-2ea8b7aa90421224fb0d96dd92b4f960R326)). 

IMHO, `createAction()` should follow the same pattern.

dannyvw:
`deleteAction()` should also follow the same pattern.

stefandoorn:
@pjedrzejewski Wouldn't it make sense to fix this before 1.0? I was looking into the 'index' event, but it's not being fired. Would be very helpful to have it I think, especially for plugins.

Issue #8539, opened 2017-09-07
[Resource] Initialize event does not return response 


dannyvw:
Initialize event does not return event response in 
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php#L369

PR #8546, opened 2017-09-08
[Resource] Resource events consistency

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/7305, replaces https://github.com/Sylius/Sylius/pull/8465 |
| License         | MIT |

Thank you @ste93cry for making a base for this PR 👍

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/Handler/ResourceDeleteHandler.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/Handler/ResourceDeleteHandler.php
@@ -1,0 +1,50 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\Doctrine\ORM\Handler;
+
+use Doctrine\ORM\ORMException;
+use Sylius\Bundle\ResourceBundle\Controller\ResourceDeleteHandlerInterface;
+use Sylius\Component\Resource\Exception\DeleteHandlingException;
+use Sylius\Component\Resource\Model\ResourceInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+final class ResourceDeleteHandler implements ResourceDeleteHandlerInterface
+{
+    /**
+     * @var ResourceDeleteHandlerInterface
+     */
+    private $decoratedHandler;
+
+    /**
+     * @param ResourceDeleteHandlerInterface $decoratedHandler
+     */
+    public function __construct(ResourceDeleteHandlerInterface $decoratedHandler)
+    {
+        $this->decoratedHandler = $decoratedHandler;
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @throws DeleteHandlingException
+     */
+    public function handle(ResourceInterface $resource, RepositoryInterface $repository): void
+    {
+        try {
+            $this->decoratedHandler->handle($resource, $repository);
+        } catch (ORMException $exception) {
+            throw new DeleteHandlingException();
+        }
+    }
+}
--- a/src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcher.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcher.php
@@ -45,7 +45,24 @@ final class EventDispatcher implements EventDispatcherInterface
     ):  ResourceControllerEvent {
         $eventName = $requestConfiguration->getEvent() ?: $eventName;
         $metadata = $requestConfiguration->getMetadata();
-        $event = $this->getEvent($resource);
+        $event = new ResourceControllerEvent($resource);
+
+        $this->eventDispatcher->dispatch(sprintf('%s.%s.%s', $metadata->getApplicationName(), $metadata->getName(), $eventName), $event);
+
+        return $event;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function dispatchMultiple(
+        string $eventName,
+        RequestConfiguration $requestConfiguration,
+        $resources
+    ):  ResourceControllerEvent {
+        $eventName = $requestConfiguration->getEvent() ?: $eventName;
+        $metadata = $requestConfiguration->getMetadata();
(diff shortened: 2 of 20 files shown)


dannyvw:
Would be nice to fix #8539 too

src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcher.php
```diff
@@ -45,7 +45,24 @@ public function dispatch(
...
+        return $event;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function dispatchMultiple(
+        string $eventName,
+        RequestConfiguration $requestConfiguration,
+        $resources
```
GSadee: `array $resources` ?

src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcherInterface.php
```diff
@@ -37,6 +37,19 @@ public function dispatch(
     /**
      * @param string $eventName
      * @param RequestConfiguration $requestConfiguration
+     * @param mixed $resources
```
GSadee: Why is it `mixed` instead of `array`?

src/Sylius/Component/Resource/spec/Exception/DeleteHandlingExceptionSpec.php
```diff
@@ -0,0 +1,43 @@
...
+
+declare(strict_types=1);
+
+namespace spec\Sylius\Component\Resource\Exception;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Resource\Exception\DeleteHandlingException;
+
+/**
+ * @author Grzegorz Sadowski <grzegorz.sadowski@lakion.com>
```
GSadee: 💥

src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcherInterface.php
Zales0123: Unfortunatelly it's not always `array` nor `Collection` :/ Take a look at `ResourcesCollectionProviderInterface`

src/Sylius/Component/Resource/spec/Exception/DeleteHandlingExceptionSpec.php
Zales0123: XD

src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcherInterface.php
stefandoorn: Isn't it always 'iterable'?

src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcher.php
```diff
@@ -45,7 +45,24 @@ public function dispatch(
...
+    ):  ResourceControllerEvent {
+        $eventName = $requestConfiguration->getEvent() ?: $eventName;
+        $metadata = $requestConfiguration->getMetadata();
+        $event = new ResourceControllerEvent($resources);
```
pamil: Shouldn't it dispatch an event for each resource?

src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcherInterface.php
pamil: `iterable|ResourceInterface[]`?

pamil:
Anyway, I'm concerned about dispatching that event for index actions, do we have any reasons to do so? What will it allow for? If we're not sure, I'd rather abstain from adding it.

stefandoorn:
It would allow to hook in on the frontend for e.g. tracking codes (I'm doing it now using the shop layout events). Other option is to add dynamic calculations (personal discounts, etc.). Not sure if these things are so valuable they need events in core though.

pamil:
I'm not sure how the current implementation might help in implementing those, it just dispatches a list of resources.

stefandoorn:
As discussed on Slack, also here for reference:

> Hm the best example is in my project inside
> 
> 
> [3:00] 
> https://github.com/stefandoorn/google-tag-manager-enhanced-ecommerce-plugin/blob/master/src/Resources/config/features/product_impressions.yml#L16 
> 
> [3:00] 
> But that gives an impression
> 
> [3:00] 
> I’m in there actually injecting JS to the view - but in my project I have one similar that register stuff to a property in a class somewhere
> 
> 
> pamil
> [3:02 PM] 
> Like you set a property on a model and it isn’t persisted then but only shown?
> 
> 
> stefandoorn
> [3:02 PM] 
> Yep
> 
> 
> [3:02] 
> I register them to inject them later in the HEAD in a certain JS var for tracking purposes
> 
> 
> [3:02] 
> Before the tracking JS gets loaded
> 
> 
> [3:02] 
> So I’m mainly willing to have that event for reading purposes
> 
> pamil
> [3:03 PM] 
> Ah, that makes sense indeed

src/Sylius/Bundle/ResourceBundle/Controller/EventDispatcherInterface.php
Zales0123: But `$resources` can be also `ResourceGridView`... I don't think it's iterable :/

stefandoorn: :-( Indeed.

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/Handler/ResourceDeleteHandler.php
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Sylius\Bundle\CoreBundle\Doctrine\ORM\Handler;
+
+use Doctrine\ORM\ORMException;
+use Sylius\Bundle\ResourceBundle\Controller\ResourceDeleteHandlerInterface;
+use Sylius\Component\Resource\Exception\DeleteHandlingException;
+use Sylius\Component\Resource\Model\ResourceInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+
+final class ResourceDeleteHandler implements ResourceDeleteHandlerInterface
```
pjedrzejewski: Why in core?

pjedrzejewski:
Thank you Mateusz! 👍

Issue #3565, opened 2015-11-12
Composer install fails on memory limit of 1Gb


stefandoorn:
I'm trying to install Sylius on my vagrant box. I set PHP to use 1024M as memory limit, but that ain't enough apparently. Ofcourse I can raise it, but I think that ain't suitable for everyone.

Not sure how it can be fixed though. Output of composer run can be found in this gist: https://gist.github.com/stefandoorn/55b1611e6c4fa9111ef5. Yes, normally I won't run it as root but it was easier on the VM to test quickly :)

PHP version = 5.6.15
Composer version = 2a8a9a4592674acec8e57de13f36357fb89ec746

stefandoorn:
When I got the memory limit to 1.5Gb I get this:

````
[1143.7MB/41.97s] Your requirements could not be resolved to an installable set of packages.
[1143.7MB/41.97s]
  Problem 1
    - Installation request for sylius/sylius 0.16.*@dev -> satisfiable by sylius/sylius[0.16.x-dev].
    - sylius/sylius 0.16.x-dev requires ext-gd * -> the requested PHP extension gd is missing from your system.
  Problem 2
    - behat/mink-extension v2.0.0 requires behat/behat ~3.0,>=3.0.5 -> satisfiable by behat/behat[v3.0.10, v3.0.11, v3.0.12, v3.0.13, v3.0.14, v3.0.15, v3.0.5, v3.0.6, v3.0.7, v3.0.8, v3.0.9].
    - behat/mink-extension v2.0.1 requires behat/behat ~3.0,>=3.0.5 -> satisfiable by behat/behat[v3.0.10, v3.0.11, v3.0.12, v3.0.13, v3.0.14, v3.0.15, v3.0.5, v3.0.6, v3.0.7, v3.0.8, v3.0.9].
    - behat/mink-extension v2.1.0 requires behat/behat ~3.0,>=3.0.5 -> satisfiable by behat/behat[v3.0.10, v3.0.11, v3.0.12, v3.0.13, v3.0.14, v3.0.15, v3.0.5, v3.0.6, v3.0.7, v3.0.8, v3.0.9].
    - behat/behat v3.0.9 requires ext-mbstring * -> the requested PHP extension mbstring is missing from your system.
    - behat/behat v3.0.8 requires ext-mbstring * -> the requested PHP extension mbstring is missing from your system.
    - behat/behat v3.0.7 requires ext-mbstring * -> the requested PHP extension mbstring is missing from your system.
[...]

PR #8559, opened 2017-09-09
Bump swiftmailer-bundle to 3.x & swiftmailer to 6.x for PHP 7 

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | yes |
| BC breaks?      | no|
| Related tickets | partially #3565  |
| License         | MIT |

Due to Sylius requiring PHP 7.1 as minimum, I think it makes sense that the libraries we are using are also up to date as possible. Swiftmailer & swiftmailer bundle have new majors that are PHP 7 only. I think in a previous commit (https://github.com/Sylius/Sylius/commit/fcad892590a11726f504e0f5d5d74437c5913042) Swiftmailer 6.x support has already been introduced, therefore I'm expecting it to work.

As discussed in #3565 the Composer update command takes use memory usage. The more strict the requirements of the platform are, the less evaluations Composer needs to do. Besides this, it promotes to be up-to-date with dependencies.


pjedrzejewski:
Thank you Stefan!

PR #8561, opened 2017-09-09
Fix DocumentRepository not compatible with Interface

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        |yes |
| New feature?    | no|
| BC breaks?      | no|
| License         | MIT |

Fixes the following error I ran in after upgrading to `dev-master` today:

```
Compile Error: Declaration of Sylius\Bundle\ResourceBundle\Doctrine\ODM\PHPCR\DocumentRepository::createPaginator(array $criteria = Array, array $sorting = Array) must be compatible with Sylius\Component\Resource\Repository\RepositoryInterface::createPaginator(array $criteria = Array, array $sorting = Array): iterable
```

Probably because I'm using the Lakion CmsPlugin, which uses PHPCR.


lchrusciel:
But we should provide something that will protect us from regression

lchrusciel:
Thanks Stefan!

PR #8565, opened 2017-09-10
Show autoload changes in readme for upgrade beta3 -> 1.0 (fixes #8564)

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no|
| BC breaks?      | no|
| Related tickets | fixes #8564 |
| License         | MIT |

PR #8570, opened 2017-09-10
Require ^3.3.8 for Symfony - when in 3.3.x

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | no|
| BC breaks?      | no |
| Related tickets | mentioned in #8565 |
| License         | MIT |

We are already at 3.3.6 if > 3.2, so think that 3.3.8 might also make perfect sense.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,7 @@
         "symfony/polyfill-intl-icu": "^1.3",
         "symfony/polyfill-mbstring": "^1.3",
         "symfony/swiftmailer-bundle": "^3.0",
-        "symfony/symfony": "~3.2.0|^3.3.6",
+        "symfony/symfony": "~3.2.0|^3.3.8",
         "twig/extensions": "^1.4",
         "twig/twig": "^2.0",
         "webmozart/assert": "^1.1",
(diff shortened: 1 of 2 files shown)


lchrusciel:
As far as I know, the current version allows to all patches from 3.2 or all versions above 3.3.6. So if we want to force using 3.3.8 we need to remove the part before the pipeline.

stefandoorn:
Your comment (https://github.com/Sylius/Sylius/pull/8565#issuecomment-328375063) is exactly the reason why I submitted this PR, to at least have it correct within the 3.3.x tree.

I don't know why 3.2 is allowed. I have the impression - if I recall correctly - it was something temporary due to compatibility issues. But not sure about that. As most tests are ran on recent versions, the question is also whether all thing still work on 3.2..

lchrusciel:
I'm voting for removing 3.2 from allowed versions. We are aiming to cover 3.4 and 4.0 as soon as possible anyway /cc @Sylius/core-team

lchrusciel:
Closing in favour of #8576

lchrusciel:
Now I hit a proper one 🏹  🎯

PR #8572, opened 2017-09-10
Remove final from controllers extending ResourceController (fixes #8563)

### LINKED ISSUE #8563: Upgrade Feedback: OrderController being final makes upgrade hard
Have already talked to several people about it on Slack, but might need to have a central place for the discussion. 

In recent commits, some Controllers that extend ResourceController have been made final. Although I like the idea of final classes, around the extensibility of the ResourceController it makes developers lives harder. This guide doesn't apply anymore for e.g. the OrderController: http://docs.sylius.org/en/latest/customization/controller.html#how-to-customize-a-resource-controller.

The use case I'm running in is that I have a PDF download connected to the OrderController, in method 'downloadAction'. It simply accepts an Order ID, fetches the file from file system and returns it as a download to the user (from inside the admin). 

Decorating the default OrderController is do-able, but there are some big downsides:

- The original controller can be injected as dependency, which is great. Unfortunately very useful helper methods as 'findOr404' and 'isGrantedOr403' are protected so therefore not callable from the decorator. This results in copy / pasting, which in the case of 'isGrantedOr403' might lead to security issues (it's harder to stay updated with Sylius updates).
- The original dependencies are hard to trace. There is no 'sylius.controller.order' service definition in a XML file, it looks to me that is done rather dynamic based on configuration inside the ResourceBundle. To mimic the correct dependencies, it's going to be a hard time I'm afraid. If copying a method like 'isGrantedOr403' I also need to make sure the correct dependencies are inside my decorator.

Some possibilities I see:

- Revert Controllers extending ResourceController being final
- Have the helper methods inside the ResourceController being public (e.g. 'findOr404') & allow to get the dependencies from the original Controller (so you don't have to declare them again)
- Allow the dependencies from the original Controller to be fetched (as above) and have the methods being shared as traits, so they can easily be implemented but still being safe during upgrades of core Sylius.

Example code of my method:

```
/**
     * @param Request $request
     *
     * @return Response
     */
    public function downloadAction(Request $request)
    {
        $configuration = $this->requestConfigurationFactory->create($this->metadata, $request);

        $this->isGrantedOr403($configuration, ResourceActions::SHOW);
        $order = $this->findOr404($configuration);

        // Find file & return
        .... // Project specific

        $disposition = $response->headers->makeDisposition(
            ResponseHeaderBag::DISPOSITION_ATTACHMENT,
            $filename
        );

        $response->headers->set('Content-Disposition', $disposition);

        return $response;
    }
```

Thinking about how to implement this in the final controller structure, is according to me quite tedious and limits extending the framework - which Sylius is great at normally. Upgrading from beta-3 to dev-master (trying it by request of @pjedrzejewski) became tremendously harder than expected :(

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | no|
| New feature?    | no|
| BC breaks?      | no|
| Related tickets | fixes #8563 |
| License         | MIT |

In case it's decided that controllers extending `ResourceController` should not be `final`, then here is the PR to fix that.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -70,10 +70,6 @@
     * `HandleException`
     * `IntegerDistributor`
     * `MissingChannelConfigurationException`
-    * `OrderController`
-    * `PaymentMethodController`
-    * `ProductTaxonController`
-    * `ProductVariantController`
     * `PromotionActionFactory`
     * `PromotionRuleFactory`
     * `ReviewerReviewsRemover`
@@ -137,7 +133,6 @@
 
 ### Product / ProductBundle
 
-* `ProductAttributeController` has been made final, use decoration instead of extending it.
 * `ProductVariantCombination` has been made final, use decoration instead of extending it.
 * `ProductVariantCombinationValidator` has been made final, use decoration instead of extending it.
 * The following methods does not longer have a default null argument and requires one to be explicitly passed:
--- a/src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
+++ b/src/Sylius/Bundle/CoreBundle/Controller/OrderController.php
@@ -19,7 +19,7 @@ use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Webmozart\Assert\Assert;
 
-final class OrderController extends BaseOrderController
+class OrderController extends BaseOrderController
 {
     /**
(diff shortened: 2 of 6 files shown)


lchrusciel:
Thanks Stefan! You saved a lot of developers from committing a suicide during upgrade to 1.0 👼

PR #8576, opened 2017-09-11
Require ^3.3.8 for Symfony - when in 3.3.x

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | mentioned in #8565, closing #8570  |
| License         | MIT |

Based on @stefandoorn work, just to merge it faster.

### DIFF
--- a/bin/symfony_requirements
+++ b/bin/symfony_requirements
@@ -1,7 +1,7 @@
 #!/usr/bin/env php
 <?php
 
-require_once dirname(__FILE__).'/../var/SymfonyRequirements.php';
+require_once dirname(__FILE__).'/./SymfonyRequirements.php';
 
 $lineSize = 70;
 $symfonyRequirements = new SymfonyRequirements();
--- a/src/Sylius/Bundle/CoreBundle/Application/Kernel.php
+++ b/src/Sylius/Bundle/CoreBundle/Application/Kernel.php
@@ -97,17 +97,14 @@ class Kernel extends HttpKernel
             new \Sylius\Bundle\FixturesBundle\SyliusFixturesBundle(),
             new \Sylius\Bundle\PayumBundle\SyliusPayumBundle(), // must be added after PayumBundle.
             new \Sylius\Bundle\ThemeBundle\SyliusThemeBundle(), // must be added after FrameworkBundle
+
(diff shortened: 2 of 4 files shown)


src/Sylius/Bundle/CoreBundle/Application/Kernel.php
```diff
@@ -97,17 +97,14 @@ public function registerBundles(): array
             new \Sylius\Bundle\FixturesBundle\SyliusFixturesBundle(),
             new \Sylius\Bundle\PayumBundle\SyliusPayumBundle(), // must be added after PayumBundle.
             new \Sylius\Bundle\ThemeBundle\SyliusThemeBundle(), // must be added after FrameworkBundle
+
+            new \Symfony\Bundle\WebServerBundle\WebServerBundle(), // allows to run build in web server. Not recommended for prod environment
```
loic425: why moving this here ?

pjedrzejewski: We decided that it does not bring much optimization but is useful when you have an issue appearing only on prod env. We still run apps in prod locally from time to time to discover a problem.

pjedrzejewski:
Thank you Łukasz!

src/Sylius/Bundle/CoreBundle/Application/Kernel.php
pamil: To make it available in another environments like staging (and nobody's forced to use it on prod ofc, it's just an easy way to see if things are working correctly without setting up a webserver).

loic425: ok I understand the reason. I was just surprised cause I've seen the previous pull-request moving it on dev/test environment :)

bin/symfony_requirements
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env php
 <?php
 
-require_once dirname(__FILE__).'/../var/SymfonyRequirements.php';
```
bendavies: @lchrusciel what's going on here? this path is now wrong.

bendavies: https://github.com/sensiolabs/SensioDistributionBundle/issues/324

lchrusciel: My bad, on it

PR #8577, opened 2017-09-11
Allow parent class type in ServiceRegistry

bpolaszek:
Following #8519

| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | none |
| License         | MIT |

Hello there,

[Sylius Registry](http://docs.sylius.org/en/latest/components/Registry/basic_usage.html) is a very useful standalone component that allow to locate services that share the same type.

This is a common need in many PHP projects, however the current implementation limits the type checking to services that share the same `interface`.

Some PHP projects or libraries rely on abstract classes instead of interfaces and type comparing with `instanceof` work as well. 

The aim of this PR is to use the `instanceof` operator instead of `class_implements()` - thus we could register services sharing the same `class` as well as services sharing the same `interface`.

A simple example:
```php
use PDO;
$serviceRegistry = new ServiceRegistry(PDO::class, 'connection');
$serviceRegistry->register('master', new PDO('mysql:host=10.0.1.1', 'foo', 'bar'));
$serviceRegistry->register('slave', new PDO('mysql:host=10.0.1.2', 'foo', 'bar'));
```

Thank you,
Ben

### DIFF
--- a/src/Sylius/Component/Registry/ServiceRegistry.php
+++ b/src/Sylius/Component/Registry/ServiceRegistry.php
@@ -26,11 +26,11 @@ class ServiceRegistry implements ServiceRegistryInterface
     private $services = [];
 
     /**
-     * Interface which is required by all services.
+     * Interface or parent class which is required by all services.
      *
      * @var string
      */
-    private $interface;
+    private $className;
 
     /**
      * Human readable context for these services, e.g. "grid field"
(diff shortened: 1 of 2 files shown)


pamil:
I think we can still merge it to 1.0, it's neither big nor complicated.

pjedrzejewski:
@pamil Agree, thank you Benoit! 👍

PR #8581, opened 2017-09-11
[Maintenance] Add staging environment

### LINKED ISSUE #8455: Introduce "staging" environment
We add it almost to every project so I think it makes sense to have it in both Sylius & Sylius-Standard.

Staging is just like production but has e-mails redirected to a test address, etc.

* [x] Sylius
* [x] Sylius-Standard

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8455 |
| License         | MIT |

Not sure if should duplicate the prod config or import it

### DIFF
--- a/app/config/config_staging.yml
+++ b/app/config/config_staging.yml
@@ -1,0 +1,23 @@
+# This file is part of the Sylius package.
+# (c) Paweł Jędrzejewski
+
+imports:
+    - { resource: "config.yml" }
+
+monolog:
+    handlers:
+        main:
+            type: fingers_crossed
+            action_level: error
+            handler: nested
+        nested:
+            type: stream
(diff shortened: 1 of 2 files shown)


pamil:
If we define staging environment as production one but without sending emails, then we should rather import the production configuration.

pjedrzejewski:
@pamil We discussed it and it seems safer to configure them separately and have full control. If we import prod in staging, one can be unaware that changing the former updates the latter as well.

pamil:
@pjedrzejewski that makes sense too, no solution seems ideal here.

pjedrzejewski:
@lchrusciel We should add it to docs too. :)

pjedrzejewski:
Thank you Łukasz!

PR #8582, opened 2017-09-11
Extract session key

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets |  |
| License         | MIT |

Based on #8000. I have added one more change and add a possibility to change a session key with a parameter.


src/Sylius/Bundle/ShopBundle/DependencyInjection/SyliusShopExtension.php
```diff
@@ -41,6 +41,7 @@ public function load(array $config, ContainerBuilder $container): void
         $loader->load(sprintf('services/integrations/locale/%s.xml', $config['locale_switcher']));
 
         $container->setParameter('sylius_shop.firewall_context_name', $config['firewall_context_name']);
+        $container->setParameter('sylius_shop.cart_session_key', $config['cart_session_key']);
```
pamil: IMO instead of introducing this parameter we should use `CartStorage` instead (introduced in #8543) - the parameter itself does not separate concerns fully as there are still channel codes appended.

Zales0123:
Closing in favor of https://github.com/Sylius/Sylius/pull/8605

Issue #8583, opened 2017-09-11
Invalid cookie name in Behat scenarios


bitbager:
Running background steps like `Given the store operates on a channel named "United States" in "USD" currency` causes below issue in test.log. It's because the channel identifier `United States` is passed as a code to the factory method.

`[2017-09-11 13:04:59] request.CRITICAL: Uncaught PHP Exception InvalidArgumentException: "The cookie name "_currency_United States" contains invalid characters." at vendor/symfony/symfony/src/Symfony/Component/HttpFoundation/Cookie.php line 53 {"exception":"[object] (InvalidArgumentException(code: 0): The cookie name \"_currency_United States\" contains invalid characters. at vendor/symfony/symfony/src/Symfony/Component/HttpFoundation/Cookie.php:53)"} []`

lchrusciel:
I guess its fixed

PR #8584, opened 2017-09-11
Lowercase and replace white characters of the channel code in Channel…

bitbager:
… setup context

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | #8583 |
| License         | MIT |


src/Sylius/Behat/Context/Setup/ChannelContext.php
```diff
@@ -114,7 +114,8 @@ public function storeOperatesOnASingleChannel($currencyCode = null)
      */
     public function theStoreOperatesOnAChannelNamed($channelIdentifier, $currencyCode = null)
     {
-        $defaultData = $this->defaultChannelFactory->create($channelIdentifier, $channelIdentifier, $currencyCode);
+        $channelCode = $this->resolveChannelCode($channelIdentifier);
```
Zales0123: We can use `StringInflector::nameToLowercaseCode(string $value)` method instead.

src/Sylius/Behat/Context/Setup/ChannelContext.php
```diff
@@ -114,7 +115,8 @@ public function storeOperatesOnASingleChannel($currencyCode = null)
      */
     public function theStoreOperatesOnAChannelNamed($channelIdentifier, $currencyCode = null)
     {
-        $defaultData = $this->defaultChannelFactory->create($channelIdentifier, $channelIdentifier, $currencyCode);
+        $channelCode = StringInflector::nameToLowercaseCode($channelIdentifier);
```
lchrusciel: It is not used anywhere, so I would put it in one line. But that's just a suggestion

bitbager: I'm doing it like this because IMHO it's more readable. Shall I change it?

bitbager:
Next time I will 🎅

src/Sylius/Behat/Context/Setup/ChannelContext.php
lchrusciel: Not needed. Just a suggestion. It is not something I would kill for 🔪

lchrusciel:
@bitbager it is good to go for me, but one scenario failed. Can you fix it?

src/Sylius/Behat/Context/Setup/ChannelContext.php
pamil: Since it has introduced `$channelCode`, I would rename `$channelIdentifier` to `$channelName`

bitbager:
@lchrusciel, sure 🙂

bitbager:
Ok, it's still failing. Looks like the bug appears in more scenarios. I will try my best to fix it before midnight.

src/Sylius/Behat/Context/Ui/Admin/DashboardContext.php
```diff
@@ -44,11 +45,11 @@ public function iOpenAdministrationDashboard()
...
 
     /**
-     * @When I open administration dashboard for :code channel
+     * @When I open administration dashboard for :name channel
      */
-    public function iOpenAdministrationDashboardForChannel($code)
+    public function iOpenAdministrationDashboardForChannel($name)
     {
-        $this->dashboardPage->open(['channel' => $code]);
+        $this->dashboardPage->open(['channel' => StringInflector::nameToCode($name)]);
```
pamil: Shouldn't it be `nameToLowercaseCode()`?

bitbager:
@pamil, looks like a pass 🎉

pjedrzejewski:
Thank you Mikołaj! 👍

PR #8586, opened 2017-09-11
Prevent guest orders hijacking

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

The problem was, that if we change email address in addressing checkout step, it was changed in customer entity. The quick fix for that is to create a new `Customer` if email is changed. It would sadly result in unneeded `Customer`s with wrong email in database, but it's still much better than current bug 😄

### DIFF
--- a/features/account/customer_account/viewing_orders_history/seeing_customer_orders_placed_as_guest.feature
+++ b/features/account/customer_account/viewing_orders_history/seeing_customer_orders_placed_as_guest.feature
@@ -24,6 +24,6 @@ Feature: Seeing customer's orders placed as guest
         And I select "Offline" payment method
         And I complete the payment step
         And I confirm my order
-        And I register with email "ned@stark.com" and password "lannistersAreDumb"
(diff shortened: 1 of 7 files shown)


pjedrzejewski:
Thanks Mateusz, nice work!

features/account/customer_account/viewing_orders_history/seeing_customer_orders_placed_as_guest.feature
```diff
@@ -0,0 +1,29 @@
...
+        When I complete addressing step with email "john@snow.com" and "United States" based shipping address
+        And I decide to change my address
+        And I specify the email as "ned@stark.com"
+        And I complete the addressing step
+        And I select "Free" shipping method
+        And I complete the shipping step
+        And I select "Offline" payment method
+        And I complete the payment step
+        And I confirm my order
+        And I register with previously used "ned@stark.com" email and "lannistersAreDumb" password
```
lchrusciel: image

PR #8588, opened 2017-09-11
[Maintenance] Upgrade JMS Serializer Bundle to 2.0

### LINKED ISSUE #8193: Upgrade JMS Serializer Bundle to 2.0
#8186 disabled aliasing `@serializer` service to `@jms_serializer.serializer` to increase compatibility with 3rd party addons for Sylius.

JMS Serializer Bundle in v2 disabled aliasing altogether, so probably the best course of action would be to upgrade the bundle to 2.0 instead of disabling the alias. However, it appears that Serializer Bundle 2.0 currently breaks Sylius, so this issue is opened as per @lchrusciel's instructions to track the progress.

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8193  |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/CustomerBundle/Resources/config/app/config.yml
+++ b/src/Sylius/Bundle/CustomerBundle/Resources/config/app/config.yml
@@ -4,6 +4,6 @@
 jms_serializer:
     metadata:
         directories:
-            sylius-user:
+            sylius-customer:
                 namespace_prefix: "Sylius\\Component\\Customer"
                 path: "@SyliusCustomerBundle/Resources/config/serializer"
--- a/app/config/config.yml
+++ b/app/config/config.yml
@@ -47,11 +47,6 @@ swiftmailer:
     username: "%mailer_user%"
     password: "%mailer_password%"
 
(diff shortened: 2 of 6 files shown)


src/Sylius/Bundle/CustomerBundle/Resources/config/app/config.yml
```diff
@@ -4,6 +4,6 @@
 jms_serializer:
     metadata:
         directories:
-            sylius-user:
+            sylius-customer:
```
pamil: 🥇

lchrusciel:
Good catch @pamil

src/Sylius/Bundle/CustomerBundle/Resources/config/app/config.yml
Zales0123: 🎖

pjedrzejewski:
Thanks Łukasz! 👍

PR #8589, opened 2017-09-11
[Shop] Strip _locale parameter from URL if using storage locale switcher

### LINKED ISSUE #7970: Disabling localised URLs passes _locale as query string
I've followed the steps from http://docs.sylius.org/en/latest/cookbook/disabling-localised-urls.html but now the URLs are generated with the _locale passed in the query string (?locale=en_US). Is there a way to disable that?

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | kinda |
| New feature?    | kinda |
| BC breaks?      | a really small one |
| Related tickets | fixes #7970 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/spec/Router/LocaleStrippingRouterSpec.php
+++ b/src/Sylius/Bundle/ShopBundle/spec/Router/LocaleStrippingRouterSpec.php
@@ -1,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+namespace spec\Sylius\Bundle\ShopBundle\Router;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Locale\Context\LocaleContextInterface;
+use Symfony\Component\HttpKernel\CacheWarmer\WarmableInterface;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
+use Symfony\Component\Routing\RouterInterface;
+
+final class LocaleStrippingRouterSpec extends ObjectBehavior
+{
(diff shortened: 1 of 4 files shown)


stefandoorn:
Looks good to me. Using a similar one in production, although mine is slightly different as I'm also using the CmsPlugin, therefore I'm decorating CmfRouter. Hopefully these two play well together.. :-)

pjedrzejewski:
@pamil Let's fix the conflict in the UPGRADE.md file and it is good to go.

src/Sylius/Bundle/ShopBundle/spec/Router/LocaleStrippingRouterSpec.php
```diff
@@ -0,0 +1,70 @@
...
+
+    function it_strips_locale_from_the_generated_url_if_locale_is_the_same_as_the_one_from_context(
+        RouterInterface $decoratedRouter,
+        LocaleContextInterface $localeContext
+    ): void {
+        $localeContext->getLocaleCode()->willReturn('pl_PL');
+
+        $decoratedRouter
+            ->generate('route_name', [], UrlGeneratorInterface::ABSOLUTE_PATH)
+            ->willReturn('http://generated.url/?_locale=pl_PL')
```
lchrusciel: Can we test for each case? It should be as easy as adding here two more examples and calling generate two more times

pamil: Done! 👌

pjedrzejewski:
Thanks Kamil! 👍

PR #8592, opened 2017-09-11
Disable the access to login page for already logged in customers/admins

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/3448 |
| License         | MIT |

### DIFF
--- a/src/Sylius/Behat/Context/Ui/Shop/AccountContext.php
+++ b/src/Sylius/Behat/Context/Ui/Shop/AccountContext.php
@@ -21,6 +21,7 @@ use Sylius\Behat\Page\Shop\Account\DashboardPageInterface;
 use Sylius\Behat\Page\Shop\Account\Order\IndexPageInterface;
 use Sylius\Behat\Page\Shop\Account\Order\ShowPageInterface;
 use Sylius\Behat\Page\Shop\Account\ProfileUpdatePageInterface;
+use Sylius\Behat\Page\Shop\Account\LoginPageInterface;
 use Sylius\Behat\Service\NotificationCheckerInterface;
 use Sylius\Component\Core\Formatter\StringInflector;
 use Sylius\Component\Core\Model\OrderInterface;
@@ -57,6 +58,11 @@ final class AccountContext implements Context
     private $orderShowPage;
 
     /**
+     * @var LoginPageInterface
+     */
+    private $loginPage;
+
+    /**
      * @var NotificationCheckerInterface
      */
     private $notificationChecker;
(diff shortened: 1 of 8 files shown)


src/Sylius/Bundle/UiBundle/Controller/SecurityController.php
```diff
@@ -62,6 +81,13 @@ public function __construct(
      */
     public function loginAction(Request $request): Response
     {
+        $alreadyLoggedInRedirectRoute = $request->attributes->get('_sylius')['logged_in_route'] ?? null;
+
+        if ($alreadyLoggedInRedirectRoute && $this->authorizationChecker->isGranted('IS_AUTHENTICATED_FULLY')) {
+
```
Zales0123: :(

pamil: :(

Arminek: 💃

src/Sylius/Behat/Context/Ui/Shop/AccountContext.php
```diff
@@ -432,6 +441,22 @@ public function iShouldBeAbleToChangePaymentMethodForThisOrder(OrderInterface $o
...
     /**
+     * @Then I should be redirected to my account dashboard
+     */
+    public function iShouldBeRedirectedToMyAccountDashboard()
+    {
+        Assert::true($this->dashboardPage->isOpen(), 'User should be on the account panel dashboard page but they are not.');
+    }
+
+    /**
+     * @When I want to log in
```
lchrusciel: I would go with `When I try to log in` and add this step to shop `LoginContext`. Fewer dependencies here and more coherent `LoginContext`.

pjedrzejewski: I did that initially but after including the LoginContext in the account suite, I got conflict of step definitions, so I decided to go this way and avoid too many changes. Are we ok with reworking that post release? :)

src/Sylius/Behat/Context/Ui/Shop/AccountContext.php
```diff
@@ -432,6 +441,22 @@ public function iShouldBeAbleToChangePaymentMethodForThisOrder(OrderInterface $o
     }
 
     /**
+     * @Then I should be redirected to my account dashboard
```
GSadee: IMO `@Then` steps should be after `@When` steps in context file

Mipme:
Great to hear! Does it also disable access to the register page?

Issue #8161, opened 2017-06-13
[Documentation][Theme] Remove the asset override feature from the doc


Colmea:
Hello Sylius team,

When creating a theme, override template works really well, but not static files (images, css, js, ...).
After a discussion on Slack, it appears this feature doesn't work/exist in Sylius.

If it's the case, we should remove it from the doc (in the Theme Book):
http://docs.sylius.org/en/latest/book/themes/themes.html#how-to-create-themes  (last note at the end of the page)

![image](https://user-images.githubusercontent.com/974918/27081617-8813b868-5041-11e7-93ae-68250fb575e0.png)

 Or even add a specific note to explain how to add custom CSS/JS (import a new CSS/JS file in the layout).

I can send a PR if wanted

pjedrzejewski:
@Colmea Would be awesome to get a PR for this, thanks Arthur!

Mipme:
Since i am using node-sass and other npm functions to build the theme assets, i simply added a copy-my-assets-script. I can contribute that to the docs if interested...

Colmea:
@pjedrzejewski What do you think about adding a custom Gulpfile into Sylius Demo project ? 
This implies to have a built-in custom theme in the repository, but this way users have a custom theme + override assets "out of the box".

I can make a PR if you find this interesting

lchrusciel:
I think simple cookbook will be easier to maintain, so I would vote tor @Mipme proposal. /cc @pjedrzejewski @pamil

pjedrzejewski:
@Colmea I would say let's have both base file & coobook? That would be huge help!

pjedrzejewski:
@Colmea Hi Arthur! Will you find a moment to push a PR for this? Would be huge help before 1.0.0. :)

PR #8603, opened 2017-09-11
Remove non-existent feature from theming docs

pjedrzejewski:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | Fixes https://github.com/Sylius/Sylius/issues/8161 |
| License         | MIT |


stefandoorn:
You can place assets inside a theme right? In e.g. the public folder, isn't it? That's where the themes assets install command is far as I see. So, shouldn't that be mentioned? Now it's nowhere written that this can be used for assets.

PR #8606, opened 2017-09-11
[Documentation] Customizing translatable fields of translatable models fix

### LINKED ISSUE #8023: Adding translatable field to entity requires entity ORM mapping
I tried adding a translation field to Taxon, based on this guide: http://docs.sylius.org/en/latest/customization/model.html#how-to-customize-translatable-fields-of-a-translatable-model.

Followed all steps, but an error occurs saying the 'Taxon.orm.yml' mapping is missing. I don't want to add fields to Taxon itself, but the mapping is required. I think either it should be described, but nicer would be not having it necessary to add this file.

CoderMaggie:
| Q               | A
| --------------- | ---
| Doc fix?        | yes! |
| New docs?    | no |
| Related tickets | fixes #8023 |
| License         | MIT |

### DIFF
--- a/docs/customization/model.rst
+++ b/docs/customization/model.rst
@@ -255,7 +255,8 @@ you'll need to update its form type. Check how to do it :doc:`here </customizati
 How to customize translatable fields of a translatable Model?
 -------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


pjedrzejewski:
Thanks Magda, great work! And @stefandoorn for reporting. :)

PR #8608, opened 2017-09-11
Make afterUrl nullable

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

Just ran into this issue using an offline payment method.

```
Type error: Return value of Sylius\Bundle\PayumBundle\Model\PaymentSecurityToken::getAfterUrl() must be of the type string, null returned
```

The used gateway = 'offline', so it's the Sylius internal one. Ran into it as in the resulting controller I'm doing some logging, which requested the afterUrl. Looking at the class code, it makes sense that it could return `null`, without the user having any other option finding out whether it is set at all (no other way to test this).

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/Model/PaymentSecurityToken.php
+++ b/src/Sylius/Bundle/PayumBundle/Model/PaymentSecurityToken.php
@@ -69,7 +69,7 @@ class PaymentSecurityToken implements ResourceInterface, TokenInterface
     /**
      * {@inheritdoc}
      *
-     * @return IdentityInterface|null
+     * @return IdentityInterface
      */
     public function getDetails()
     {
@@ -111,7 +111,7 @@ class PaymentSecurityToken implements ResourceInterface, TokenInterface
     /**
      * {@inheritdoc}
      */
-    public function getAfterUrl(): string
+    public function getAfterUrl(): ?string
     {
         return $this->afterUrl;
     }


src/Sylius/Bundle/PayumBundle/Model/PaymentSecurityToken.php
```diff
@@ -111,7 +111,7 @@ public function setTargetUrl($targetUrl): void
     /**
      * {@inheritdoc}
      */
-    public function getAfterUrl(): string
+    public function getAfterUrl(): ?string
```
lchrusciel: Isn't it inherited? If so, we should adjust interface as well

src/Sylius/Bundle/PayumBundle/Model/PaymentSecurityToken.php
```diff
@@ -69,7 +69,7 @@ public function setDetails($details): void
     /**
      * {@inheritdoc}
      *
-     * @return IdentityInterface|null
+     * @return IdentityInterface
      */
     public function getDetails()
```
lchrusciel: missing typehint

lchrusciel:
@stefandoorn can you update these comments in separate PR?

lchrusciel:
Anyway, thanks Stefan! You rock! 
image

stefandoorn:
It's inherited from Payum, not PayumBundle, if I'm correct. In their interface there are no type hints at all.

stefandoorn:
I couldn't get a typehint on getDetails. Have to see again what was the cause. PHPstorm also errored out on it as I recall (not behind computer now).

Issue #8360, opened 2017-08-02
Translations of category requirements in shipping method


lasotaartur:
How to add translations to category requirements in shipping method i.e. "At least 1 unit has to match the method category". This description exists in Sylius/Component/Shipping so there is a problem to use symfony translation.

Zales0123:
For sure there should be translations keys used for these messages, to allow the multi-language translation 👍  are you willing to open a PR with such a fix?

lasotaartur:
But I don't know how to fix it because in Sylius/Component we can't use symfony translations I think.
In Bundle yes, but in component no.

Zales0123:
Well, `getCategoryRequirementLabels()` method from `ShippingMethod` entity is used only on `ShippingMethodType`. I think we can remove these method and replace it with `const`->`translation key` array in form type itself. /cc @pjedrzejewski

lasotaartur:
@pjedrzejewski  ?

pjedrzejewski:
I agree with @Zales0123. Let's add these in the form type.

PR #8612, opened 2017-09-12
[Shipping] Make shipping method category requirements translatable

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/8360 |
| License         | MIT |

I was not sure what to do with `getCategoryRequirementLabel` method, as it uses the static method that would couple component with translations keys... but now there is no easy way to display shipping method category requirement... but we don't display it anywhere 😄  🎪

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -243,6 +243,9 @@
     * `ShipmentUnitInterface::setShipment`
     * `ShipmentMethodInterface::setCategory`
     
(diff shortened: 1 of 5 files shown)


pjedrzejewski:
Thanks Mateusz!

PR #8623, opened 2017-09-12
[PaymentMethod] Minor gateway fixes

### LINKED ISSUE #8060: Payment Method gateway determined by name
Hey,
when setting a new payment method, the gateway to be used is determined by the *typed* name:
image
which is inserted automatically into the _sylius_gateway_config_ table, but is *no* longer editable.

This should be determined by the config.

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8060 |
| License         | MIT |

Changes:
- display gateway as gateway factory name in payment methods grid
- set gateway name as a payment method code which is unique instead of payment method name

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/grids/payment_method.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/grids/payment_method.yml
@@ -26,10 +26,12 @@ sylius_grid:
                     label: sylius.ui.name
                     sortable: translation.name
                 gateway:
-                    type: string
-                    path: gatewayConfig.gatewayName
+                    type: twig
+                    path: gatewayConfig.factoryName
                     label: sylius.ui.gateway
-                    sortable: gatewayConfig.gatewayName
+                    sortable: gatewayConfig.factoryName
+                    options:
+                        template: "@SyliusUi/Grid/Field/humanized.html.twig"
                 enabled:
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
@makasim does it look good to you? Thanks for your time!

src/Sylius/Bundle/AdminBundle/Resources/config/grids/payment_method.yml
```diff
@@ -26,10 +26,12 @@ sylius_grid:
                     label: sylius.ui.name
                     sortable: translation.name
                 gateway:
-                    type: string
-                    path: gatewayConfig.gatewayName
+                    type: twig
+                    path: gatewayConfig.factoryName
```
makasim: I am not sure but it worth to menition that the gateway's factoryName property and its getter and setter are [deprecated](https://github.com/Payum/Payum/blob/bc09fd5f2c4561b6c42dc913bbdebb98134cad6e/src/Payum/Core/Model/GatewayConfigInterface.php#L21)

The factory name is inside the gateway's config. the key is `factory` (similar as in the payum.yml)

pjedrzejewski:
Thanks Grzesiu!

PR #8626, opened 2017-09-12
[Documentation] More detailed docs of Promotion Actions vs. Filters

### LINKED ISSUE #8118: Wrong 'order item percentage discount' promotions in cart
I have a weird situation currently regarding promotions. The shop I'm working on has about 10 products. 9 out of 10 products have each a single unique promotion rule assigned with the following settings (because they need to be switched on/off regularly):

* Exclusive = false
* Coupon based = 0
* Channel: only channel I have

Rules:
* Contains product: only selected the one single product it applies on

Actions:
* Item percentage discount
* Percentage: 25 (some have different percentages, but for this example I took 25)

Now I add one of these products to the cart (let's call it Product A), the normal price and the discounted price show perfect.

Then I add this single product to the cart that does not have promotions (let's call it Product B). The same 25% discount is also applied on this product (B). Which is not what I expected, as I expected 'order percentage discount' to do this, and not 'order item percentage'. I expected only discount on A.

Then I edit the promotion rule (connected to A) to exclude this product that has no promotions (so I exclude the product B). Then I update the cart, and only the product WITHOUT promotions (B) has 25% discount, but the product (A) that originally should have 25% discount does NOT have discount.

Behavior stays like this adding more products. A single product in the cart (regardless quantity) doesn't give problems. Quantities are never the problem.

I have the impression this is a bug, as I expected the action 'order item percentage' to only apply that discount to that single order line item and not to other order line items in the cart.

CoderMaggie:
| Q               | A
| --------------- | ---
| Doc fix?        | no |
| New docs?    | yes |
| Related tickets | fixes #8118 |
| License         | MIT |

### DIFF
--- a/docs/book/orders/promotions.rst
+++ b/docs/book/orders/promotions.rst
@@ -102,16 +102,21 @@ Each PromotionRule type has a very specific structure of its configuration array
 Promotion Actions
 -----------------
 
-What happens with the Order when the rules of a Promotion are fulfilled - this is an **PromotionAction**.
+Promotion Action is basically what happens when the rules of a Promotion are fulfilled, what discount is applied on the whole Order (or its Shipping cost).
 
 There are a few kinds of actions in **Sylius**:
 
-* fixed discount on the order (for example: -5$ discount)
+* fixed discount on the order (for example: -5$ off the order total)
 * percentage discount on the order (for example: -10% on the whole order)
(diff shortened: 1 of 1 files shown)


lchrusciel:
Thanks, Magda! 🐋

stefandoorn:
I think this is what @gabiudrescu meant when I had no clue how it worked. Thanks from us :)

PR #8637, opened 2017-09-12
[Resouce] Fix 0 page paginator

### LINKED ISSUE #8631: [RFC] [Resource] Cannot paginate with 0 items per page
The Api can't paginate requesting 0 items per page.
Read on for a bad side effect:

Use case: 
with a resource configured as config such as:
```
_sylius:
    paginate: $limit
    filterable: true
    sortable: true
```
I want to get the total number of resources, so I would want to do.
`GET /resource/?limit=0`
and get the value of to the `total` property.
Of course, this doesn't really make sense as you can't calculate the number of pages with per page of 0.

**However...**

what happens is that ALL resources are returned, due to https://github.com/Sylius/Sylius/blob/163c99e9585579df89329e78359a98fd4311170d/src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php#L308 casting 0 to false.
and then
https://github.com/Sylius/Sylius/blob/163c99e9585579df89329e78359a98fd4311170d/src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php#L298
casting 0 to false.

so requesting 0 items actually returns all items, unpaginated.
This will kill your server pretty quick if you have a large number of resources. not great!

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8631  |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
@@ -307,7 +307,9 @@ class RequestConfiguration
      */
     public function isPaginated()
     {
-        return (bool) $this->parameters->get('paginate', true);
+        $pagintion = $this->parameters->get('paginate', true);
+
+        return $pagintion !== false && $pagintion !== null;
     }
 
     /**
--- a/tests/Controller/CustomerApiTest.php
+++ b/tests/Controller/CustomerApiTest.php
@@ -160,6 +160,20 @@ EOT;
     /**
      * @test
      */
+    public function it_returns_page_not_found_if_limit_is_set_to_0()
+    {
+        $this->loadFixturesFromFile('authentication/api_administrator.yml');
+        $this->loadFixturesFromFile('resources/customers.yml');
+
+        $this->client->request('GET', '/api/v1/customers/?limit=0', [], [], static::$authorizedHeaderWithAccept);
+
+        $response = $this->client->getResponse();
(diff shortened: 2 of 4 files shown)


tests/Controller/CustomerApiTest.php
```diff
@@ -160,6 +160,20 @@ public function it_allows_to_get_customers_list()
     /**
      * @test
      */
+    public function it_returns_page_not_found_if_page_numer_0_requested()
```
pamil: numer -> number, also not a page :)

pjedrzejewski:
Thank you Łukasz!

src/Sylius/Bundle/ResourceBundle/Controller/RequestConfiguration.php
```diff
@@ -307,7 +307,9 @@ public function getLimit()
      */
     public function isPaginated()
     {
-        return (bool) $this->parameters->get('paginate', true);
+        $pagintion = $this->parameters->get('paginate', true);
```
plozmun: $pagination

lchrusciel: Are you willing to open a PR?

plozmun: Ready!

PR #8640, opened 2017-09-12
Remove return type in PaymentSecurityToken, might not always be correct

stefandoorn:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

After discussion with @lchrusciel in #8608 and @pamil on Slack.


lchrusciel:
Thanks Stefan!

PR #8648, opened 2017-09-13
Remove pre-stable migrations before tagging stable

rvanlaak:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | N/A |
| License         | MIT |

Migrations should be removed when `1.0.0` is tagged, as they are only relevant for pre-1.0 users, and these migrations shouldn't be "supported".


Zales0123:
Do we really want to remove these migrations? I think it's good they were removed from `Sylius-Standard` but for people already using **Sylius** it would be easier to upgrade if migration would still be here. /cc @pjedrzejewski

rvanlaak:
Releasing them with `1.0.0` would also mean Sylius should support them. Supporting pre-stable migrations is a thing I think you shouldn't do. The "state" of the `1.0.0` tag should be as clean as possible. So, yes this is a breaking change.

That doesn't mean the migrations shouldn't be available anymore of course, I think there should be some sort of "upgrade from x to 1.0.0" dist with the tag release ;)

Imagine users doing a clean install of `1.0.0`. They will create the database, and then the migrations will run. I'm not sure, but I think that will result in migration failures.

pamil:
These migrations start from the scratch and the installer actually uses them, so it shouldn't cause any failures. Anyway, the die is cast.

rvanlaak:
`sylius-standard` replaced all migrations with one migration that adds all default data (like fixtures would do), so all migrations over here could also get replaced with that single migration. That would "cleanup" the `pre-1.0.0` state.

https://github.com/Sylius/Sylius-Standard/tree/master/app/migrations

Issue #8628, opened 2017-09-12
Product names are not stored in Orders/OrderItem so can change in the future


johnrisby:
At the moment when an order is created it saves the products as orderItems linked to the Order. 

The orderItem class - correctly - saves the current price of the item, as this could change in the future and that can't affect orders/invoices in the past. The order should be immutable. 

But the product itself is linked to only by product id. 

So if the product name is changed in the future, older orders would inherit that change.

Some may argue never to change product names, but that's not realistic and shop owners will do that. There are times you would want to (ie a small name change or a vintage year in a wine) and you don't want to create a new product for SEO or other reasons.

It would be confusing to the customer and the administrators if older orders suddenly change their contents, even if subtly. 

My suggestion is the current product name is also copied in to the OrderItem. This would make it immutable and would mean a site owner could rename or even delete the product in the future without any impact on previous orders.

czende:
It will be great to have this option. So we can delete old products, but keep orders for statistics. For now, I have added product and product variant trash (archived_at column), so we can archive some older products/variants and not affect orders.

gabiudrescu:
I don't like the idea of deleting products from the database, no matter if the title is copied to the order. 

but otherwise, I'm totally for this functionality - keeping some sensible (legally binding) data copied on order / order items.

stefandoorn:
Think this is one of the major things missing right now to be compliant with laws. Not everyone integrated invoice PDF probably yet ;-)

psihius:
With time there are a lot of products to delete. Like right now the migrated database has 6034 products with 23667 variants. I clean out the old stuff and the new Sylius database ends up with 3016 products with 14344 variants. That's a lot of old stuff, and they have been clearing the old database over the years. The stuff may look the same, but it gets new SKU's, so they need to add new products anyway.

stefandoorn:
@johnrisby is working on a PR ;-)

gabiudrescu:
@psihius personally, I'm against deleting products from the database. archiving them is one thing, but deleting is a slippery slope to me. 

imagine what would happen, from the reporting / accounting point of view if you would have 1 SKU for 2 different products - one from 10 years ago, one from today.

PR #8667, opened 2017-09-13
#8628 Added immutable product/variant names and codes on OrderItem

johnrisby:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    |yes |
| BC breaks?      | no |
| Related tickets | adds functionality in issue #8628 |
| License         | MIT |

This PR is to add immutable product and variant names and codes to orders.

At the moment, only a reference to the variant is stored in the order.

This leads to the problem that if a product description or code is changed in the future, all previous orders will now reflect the new details.

This could cause problems both from a shop owners point of view and a customers point of view. As an example, we sell wine and sometimes the same product changes only by the year it was produced - ie, Wine Name 2015 can become Wine Name 2016. 

In this case, customers previous orders would now say 2016 when they bought 2015, and if they were to refer to their online account it would cause confusion. 

This maybe isn't the best example but it's one that - for us - would cause problems.

More importantly, from a legal point of view, the orders should be immutable and at the moment they aren't. This patch fixes that.

The migration file creates four new fields in the OrderItem table - immutable_product_name, immutable_product_code, immutable_variant_name and immutable_variant_code.

It also populates these fields with the existing data from the products in the orders (ie, it goes through all OrderItems, gets the variant, and sets the immutable fields from the current data). Once done, it makes the fields not null.

The controller itself returns the immutable fields if they exist, or the product-stored data if not, so even without the migration it shouldn't break BC.

This is my first ever contribution to an open source project (thanks to @stefandoorn for the help with git ;) ) - so please be gentle if it's terrible or I've done something wrong. Thanks ;)


johnrisby:
The question of translations has been raised on Slack. 

At the moment, this stores the names in the language of the shopper as it's their order that is being created. But there is some feeling that could cause problems for administrators for some multi-language sites who may want to view the products in their native language rather than the customers language.

After a good discussion I'm still not totally convinced by the need for further translations to be stored by default.

As a developer using the Sylius framework, it's something you could override yourself if you wanted to save more languages. And of course it's always possible to get the current (but not historic) translation yourself from the product/variant ID.  In all likelihood the product would be displayed with the image and the product/variant code - and a link to the product - anyway so I'm not sure how big an issue this really is in real-life for most uses.

One compromise would be to store the names in both the language of the shopper and the base language of the shop/company. Although that feels a bit like a half-measure.

A fuller solution, but sticking with this basic principle, would be to serialise the translation collections of the product and the variant and store that in a json field, but it feels like a bit of hack to save that much data for each item in an order (even if disk-space is 'free' these days). Although in most cases, there's only ever going to be a small number of languages per product. Not many sites would have 150 languages in use. But even so - it feels like throwing resources at the solution rather being smart.

For my purposes right now, this does what I need but I'd love to know if anyone has any views on the translation issue. Thanks.

johnrisby:
Sorry, not quite sure what I've done here. I've managed to mess up the PR somehow by updating it to Sylius/Master (which is now 1.1) which added thousands of changes. I've put it back to 1.0 but even now this has added other commits apart from mine. I'm not used to github so not sure how best to fix this. Any advice would be great, thanks ;)

regnisolbap:
Hi John, don't worry, it seems to me that you just have to wait for a core team member to modify the base branch of your PR.

pamil:
@regnisolbap PR owner can also change the base branch by clicking on `Edit` button next to the title :)

pamil:
But anyway, as it's a new feature, it should be based on master (1.1).

johnrisby:
Ok, I'll just leave it as it is for now - no guarantee it will be accepted anyway. Sorry again for the mess up.

GSadee:
@johnrisby if you have problem with rebasing this PR, you can try to fix that by:
```
git rebase -i master
```
and comment out the unnecessary commits

GSadee:
Closing in favour of #8874

Issue #8047, opened 2017-05-14
Taxation model doesn't seem to work for EU B2B/B2C sales


ioweb-gr:
Hi,

Sorry for the newbie question, but we only recently decided to try out Sylius for our company and noticed something that I don't see a way to fix other than customizing the taxation logic but most ecommerce alternatives support out of the box e.g. Magento, Prestashop, Woocommerce. 

The problem comes with calculating VAT Tax in European Union countries. 

Let's say we have 4 types of customers, 

Local Retail Customers (LRC) 
Foreign Retail Customers (FRC)
Local Businesses (LBC)
Foreign Business (FBC)

For retail customers there are two cases I've encountered so far based on individual country's law
1. add the same VAT Tax % for every sale to any LRC  or FRC
2. add the VAT tax % of the recipient's country for FRCs (for example 20% if the person purchasing is residing in Germany, 24% if the person is in Greece).  

For business customers I see two different cases as well
1. put VAT Tax to every sale to any LBC
2. don't put any VAT Tax (0% VAT) for FBC

I thought of setting up Zones first with scope for taxes but I can't see how to set it up the way the system  currently works. I can't see any association between "Customer Group and Tax" or  between "Zones and Customer Group" that would enable this sort of association I'm describing. Also there's not a way to automatically check during checkout if the guest is a company or an individual. 

One common implementation I see for this in checkout when purchasing from abroad is the following:
1a. when the guest fills in a company name, he is considered a BC 
1b. the guest selects if he's buying as individual or company
2. If the country he selected is within the EU, a VAT Identification Number field shows up to be filled. 
3. The tax is calculated based on the type of customer and country selected. 

Am I missing something simple here maybe because it looks to me like a serious flaw. I appreciate any advice. Thank you in advance

ioweb-gr:
After checking around  abit more I still can't find a solution to this within sylius. There's no relation between tax rate and customer tax class. 

Here's an example in Magento where we have tax rates, product tax class and customer tax class to define taxation for products. 

![image](https://cloud.githubusercontent.com/assets/20220341/26094242/d3872b4e-3a21-11e7-9dba-2c49bada819d.png)

Here's an example from virtuemart 

![image](https://cloud.githubusercontent.com/assets/20220341/26094454/bfeeba56-3a22-11e7-8f14-78c9f4856709.png)

pjedrzejewski:
[...]

PR #8676, opened 2017-09-15
[RFC][POC][Taxation] Customer tax category concept

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | associated with #8047 |
| License         | MIT |

This PR is the first approach to solving the problem mentioned in #8047. We would like to introduce new entity `CustomerTaxCategory` which can be associated with a `CustomerGroup`. This `CustomerTaxCategory` could be selected for a specific `TaxRate`, so it gives the possibility to create different tax rates for different customers also depending on product tax categories and zones.

One main question is how to resolve the customer tax category during checkout if there is a guest user. Possible solutions for that have been mentioned in #8047. For now, there is a default customer tax category for the channel as it does for the default tax zone. So, when the customer has no group, or the customer group has no tax category, then the customer tax category is taken from the current channel.

This is only a basic concept, probably not a final solution, so please, let us know what you think about this approach. Any comments and ideas would be really appreciate.


ioweb-gr:
From what I can see this works great for logged in customers where the administrator can change the customer group. For most businesses that's OK too because for business customers they want to sign them up manually after checking out their proof of registration so they know they're dealing with real businesses and not people trying to exploit and buy things without taxes. The relationship between `Customer` and `CustomerTaxCategory` would be `ManyToOne` to prevent accidents right?

It's pretty standard to assume any person is a retail customer until declared otherwise so the default tax and default group for guests would be great. 

Later on we have options to chose category for guest users. For example detection based on external services like VIES check (http://ec.europa.eu/taxation_customs/vies/) for tax identification number could allow us to automatically assign a generic `EU B2B` `CustomerTaxCategory` when a guest user fills up his tax identification number. At least for EU this will work correctly because if a business wants to conduct trade with another country in the EU then they have to be VIES Registered. 

For other countries I don't know if there is a system like that where we can check validity of whether a business really exists but there could be something interactive that will dynamically set guest tax category based on user input. It's very common on e-shops selling worldwide great volumes to just make you check a checkbox that says you're indeed a real business and you assume responsibility if it's not true.

GSadee:
@ioweb-gr thank you very much for your answer. I know that this solution isn't the best and it probably doesn't resolve every B2C/B2B cases. Hovewer, I would like to provide the basic concept, which of course works properly and then we could think about some more complex resolving `CustomerTaxCategory` based on customer data.

GSadee:
@ioweb-gr thank you very much for your answer. I know that this solution isn't the best and it probably doesn't resolve every B2C/B2B cases. Hovewer, I would like to provide the basic concept, which of course works properly and then we could think about some more complex resolving `CustomerTaxCategory` based on customer data.

pamil:
As for guest checkout and other cases, shouldn't the tax category be resolved based on billing address?

ioweb-gr:
It depends. If the product is a service it can differ based on where the service is provided so it might be based on the location of the business that provides the service or the billing address or the shipping address. Taxation is a chaos

psihius:
Well, at least in EU, you charge your VAT in the delivery country usually - at least the VAT rate for me for services and goods usually is my 21% that is my country's tax rate.

For guest checkouts I resolve your tax rate via the delivery address - we don't use the billing address at all, but since delivery address is just copied, then it can work. The only caveat here is when user changes the delivery address and billing address is not used - the billing address should change too and that's not the case as far as I know at the moment.

As for the what tax rate to use for services - I think this is probably just zone and default rate configuration issue - if the zone is not configured, then the default tax rate is used and that should cover the case when company uses the tax rate based on the service provider location, not client location.

stefandoorn:
Inside EU it's actually a bit more complicated. By default you charge your own country VAT (as where the shop is located). But, as soon as you get over certain threshold to a certain country, you have to charge the VAT rate of that country for that customer (shipping address). This applies the rest of that year and the year after. If you meet the threshold next year again, obviously it stays in place year after. If you don't, the year after it's back to normal. It goes into effect directly after passing over the threshold 

So, this could mean that only to U.K. customers you charge the U.K. Rate, and to all other EU countries you charge your own default rate. 

Reference: https://ec.europa.eu/growth/tools-databases/dem/watify/selling-online/how-can-i-comply-vat-obligations?language=en (2nd bullet).

Downside of this is that you get tax rates and including vat prices that are different to each client. It's therefore hard to show properly on the frontend catalogue views.

src/Sylius/Component/Core/OrderProcessing/OrderTaxesProcessor.php
```diff
@@ -111,9 +126,29 @@ private function getTaxZone(OrderInterface $order)
     }
 
     /**
+     * @param OrderInterface $order
+     *
+     * @return CustomerTaxCategoryInterface|null
+     */
+    private function getCustomerTaxCategory(OrderInterface $order): ?CustomerTaxCategoryInterface
```
pamil: Shouldn't it be extracted to `CustomerTaxCategoryProviderInterface`?

src/Sylius/Component/Core/OrderProcessing/OrderTaxesProcessor.php
```diff
@@ -77,15 +88,19 @@ public function process(BaseOrderInterface $order): void
...
 
         $zone = $this->getTaxZone($order);
-
         if (null === $zone) {
             return;
         }
 
+        $customerTaxCategory = $this->getCustomerTaxCategory($order);
+        if (null === $customerTaxCategory) {
+            return;
```
pamil: We should probably throw an exception here.

src/Sylius/Component/Core/Provider/CustomerTaxCategoryProviderInterface.php
```diff
@@ -0,0 +1,27 @@
...
+use Sylius\Component\Core\Model\OrderInterface;
+
+interface CustomerTaxCategoryProviderInterface
+{
+    /**
+     * @param OrderInterface $order
+     *
+     * @return CustomerTaxCategoryInterface|null
+     */
+    public function getCustomerTaxCategory(OrderInterface $order): ?CustomerTaxCategoryInterface;
```
pamil: How about making it return a customer tax category or throw an exception if not found?

tests/Controller/CheckoutAddressingApiTest.php
```diff
@@ -32,7 +32,7 @@ public function it_denies_order_addressing_for_non_authenticated_user()
     }
 
     /**
-     * @test
+
```
pamil: ?

src/Sylius/Component/Core/spec/OrderProcessing/OrderTaxesProcessorSpec.php
```diff
@@ -137,4 +161,35 @@ function it_does_not_process_taxes_if_there_is_no_tax_zone(
 
         $this->process($order);
     }
+
+    function it_does_not_process_taxes_if_there_is_no_customer_tax_category(
```
pamil: Is it a valid assumption? It would be safer to throw an exception than to make someone pay no taxes in our shop because of misconfigured taxes.

src/Sylius/Behat/Page/Admin/Channel/UpdatePage.php
```diff
@@ -85,6 +85,17 @@ public function chooseDefaultTaxZone($taxZone)
     /**
      * {@inheritdoc}
      */
+    public function chooseDefaultCustomerTaxCategory(?string $customerTaxCategoryName = null): void
+    {
+        $this->getDocument()->selectFieldOption(
+            'Default customer tax category',
+            (null === $customerTaxCategoryName) ? '' : $customerTaxCategoryName
```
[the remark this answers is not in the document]
GSadee: Why? If `$customerTaxCategoryName` is not null, it is a string, so IMO casting it to the string is not necessary.

src/Sylius/Component/Core/spec/OrderProcessing/OrderTaxesProcessorSpec.php
GSadee: IMO the same case applies to tax zones, I wanted to implement this concept similar to tax zone. What do you say to discuss this later and apply proper behaviour in a separate PR for both concepts?

src/Sylius/Behat/Page/Admin/Channel/UpdatePage.php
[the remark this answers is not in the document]
pamil: `(string) null === ''`, it's a shortcut for this ternary operator

psren:
this is not stale

gorkalaucirica:
This is not stale

afederau:
Last week I downloaded Sylius and tested it, to check if it is usable as replacement for an old ecommerce software. And first thing I missed was the possibility set up VAT/taxation functionality/logic dependent to customer group. Checked the roadmap found this issue, but I didn't understood this:

> This CustomerTaxCategory could be selected for a specific TaxRate, so it gives the possibility to create different tax rates for different customers also depending on product tax categories and zones.

IMO the Tax/VAT for all customers depends from Country or Province (Zone), but in some cases the B2B customers or charity organizations don't have to pay taxes. Do You know the cases where the customer have other/special tax rate?
I think the tax related information for customer group should define:

- how to display the prices for the customer group, VAT included or excluded
- how to pay the order, VAT included or excluded
- the order-documents (order confirmation, invoice, etc.) for specific customer group may contain specific information and may have special layout
- specific requirements to customer data. For example the retail customers in EU should have valid VAT Number. Not sure but the charity organizations also should have a valid identification. Or it could be a country-specific identification for locale (country resided) companies/organisations.
 
(some can complete the above)

pamil:
It's an important feature, but I think we need to get to it one more time (cc @CoderMaggie @GSadee).

michaelKaefer:
@pamil @GSadee Maybe also one of you can answer my question here: https://github.com/Sylius/Sylius/issues/8047. Would be nice :)

PR #8677, opened 2017-09-15
[Docs] Corrections on OrderItemQuantityModifier

juukie:
`$orderItem->getTotal();` should be 2 * 1000 = 2000
Changed the `"0"` and `"2"` because getQuantity() returns an integer.


pamil:
Thanks @juukie! 🎉

PR #8678, opened 2017-09-15
[ReviewBundle] Recalculate subject rating on review post create event

loic425:
| Q               | A
| --------------- | ---
| Bug fix?        | kind of |
| New feature?    | kind of  |
| BC breaks?      | no |
| Related tickets |  |
| License         | MIT |

In case you handle your reviews without moderation (directly with accepted status), you need your subject would be recalculate on post create event. You also need this feature if you handle ratings without title and comment.

### DIFF
--- a/src/Sylius/Bundle/ReviewBundle/EventListener/ReviewChangeListener.php
+++ b/src/Sylius/Bundle/ReviewBundle/EventListener/ReviewChangeListener.php
@@ -71,8 +71,6 @@ final class ReviewChangeListener
             return;
         }
 
-        if (ReviewInterface::STATUS_ACCEPTED === $subject->getStatus()) {
-            $this->averageRatingUpdater->update($subject->getReviewSubject());
-        }
+        $this->averageRatingUpdater->update($subject->getReviewSubject());
     }
 }
--- a/src/Sylius/Bundle/ReviewBundle/EventListener/ReviewChangeListener.php
+++ b/src/Sylius/Bundle/ReviewBundle/EventListener/ReviewChangeListener.php
@@ -13,10 +13,10 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\ReviewBundle\EventListener;
 
+use Doctrine\ORM\Event\LifecycleEventArgs;
 use Sylius\Bundle\ReviewBundle\Updater\ReviewableRatingUpdaterInterface;
 use Sylius\Component\Resource\Exception\UnexpectedTypeException;
 use Sylius\Component\Review\Model\ReviewInterface;
-use Symfony\Component\EventDispatcher\GenericEvent;
 
 /**
  * @author Mateusz Zalewski <mateusz.zalewski@lakion.com>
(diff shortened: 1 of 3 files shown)


Zales0123:
Shouldn't this be implemented inside a custom application? I think this is a custom feature and should not be provided together with default configuration, it's a different behavior :)
I think that instead of implementing this, you can write a cookbook how to customize product reviews recalculation? :) /cc @GSadee @CoderMaggie

loic425:
@Zales0123 In sylius, this recalculates for nothing but I think it adds useless complexity on this bundle for a common use case. Indeed, I can write a cookbook to remove moderation on a review if you prefer.

pamil:
I'd say the current behaviour is a bug, eg. the current average rating won't be even be calculated when using fixtures or when programatically adding already accepted reviews to the product.

pamil:
Anyway, the other thing is that we rely on ResourceBundle's events, so the cases mentioned above wouldn't be handled with this PR, but then making review already accepted would be a no-brainer, currently you might spend quite some time figuring out why average rating isn't updated.

loic425:
@pamil 
We can use doctrine lifecycle event instead. What do you think ?

loic425:
@pamil "but then making review already accepted would be a no-brainer" If you implements simple rating without comments, it make sense.
Do you want me to make another pull-request with lifecycle events instead of resource events and you can choose which one could be merged ? I think lifecycle events would be better to test average rating with behat.

pamil:
@loic425 sorry for the delay, must've marked this PR as read by an accident. The current version is all right for me (although while being a bugfix, it should target `1.0` branch).

If you're willing to make it use Doctrine lifecycle events, that's even better - it would be easier to test as well! :)

loic425:
@pamil I'm on it, build is failed but i'm working on fixing phpspec and behat scenarios.

src/Sylius/Bundle/ReviewBundle/EventListener/ReviewChangeListener.php
```diff
@@ -37,13 +37,38 @@ public function __construct(ReviewableRatingUpdaterInterface $averageRatingUpdat
...
+    public function recalculateSubjectRating(LifecycleEventArgs $args): void
+    {
+        $subject = $args->getObject();
+
         if (!$subject instanceof ReviewInterface) {
-            throw new UnexpectedTypeException($subject, ReviewInterface::class);
+            return;
         }
 
         if (ReviewInterface::STATUS_ACCEPTED === $subject->getStatus()) {
```
pamil: What do you think about removing this condition? It won't change the rating if an already accepted review gets declined.

loic425: @pamil I think it's necessary. Using doctrine lifecycle events, you always have to check you handle wanted entity: http://symfony.com/doc/current/doctrine/event_listeners_subscribers.html

pamil: What I mean was to leave that `!$subject instanceof ReviewInterface` if, but remove `ReviewInterface::STATUS_ACCEPTED === $subject->getStatus()` one, so when a review is already accepted and you decide to change its status, then the average rating is updated and all logic stays in `AverageRatingUpdater`.

loic425: yes, you're right. I'm gonna remove that.

loic425:
@pamil I think it's ok:) All changes have been done.

src/Sylius/Bundle/ReviewBundle/spec/EventListener/ReviewChangeListenerSpec.php
```diff
@@ -33,11 +34,11 @@ function let(ReviewableRatingUpdaterInterface $averageRatingUpdater): void
...
     function it_recalculates_subject_rating_on_accepted_review_deletion(
         $averageRatingUpdater,
-        GenericEvent $event,
+        LifecycleEventArgs $event,
         ReviewInterface $review,
         ReviewableInterface $reviewSubject
     ): void {
-        $event->getSubject()->willReturn($review);
+        $event->getObject()->willReturn($review);
         $review->getStatus()->willReturn(ReviewInterface::STATUS_ACCEPTED);
```
pamil: This one is not needed now as well since review status does not matter in this listener :)

loic425:
@pamil Done! 🥇

pjedrzejewski:
@loic425 Thanks for your work, could you rebase against 1.0 branch? This is a bug fix and should be merged to 1.0 but after changing the base we have some conflicts. :)

loic425:
@pjedrzejewski done :)

pjedrzejewski:
Thanks Loïc! 👍

PR #8680, opened 2017-09-17
add sylius version to the footer in admin

### LINKED ISSUE #8679: Adding the Sylius version in admin
I was curious to see which Sylius version is running on demo.sylius.org and there is no mention in admin or web.

What do you think about adding it to the footer of the admin? Like wordpress does.

gabiudrescu:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | fixes #8679  |
| License         | MIT |

It seems the version was missing from admin due to a bug in a twig file - the footer block in admin got renamed post_content at some point in history. 

On top of fixing this bug, I added a new constant in the kernel class, APP_NAME with the default value being "Sylius". In case at a later point someone decides to craft a forked version of Sylius called "Supercharged Sylius" and it needs to brag about the name. Before hand, Sylius was hardcoded in the twig template. 

I have declared both Sylius\Bundle\CoreBundle\Application\Kernel::VERSION and Sylius\Bundle\CoreBundle\Application\Kernel::APP_NAME as twig global variables and used them to be in the twig layout template for admin. 

I consider this useful when debugging a Sylius application over the phone with a non-technical person. First question I would ask is: what version of Sylius do you run? 

Also, I find it a good reason to brag about: look, I'm running Sylius 1.0.0!

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/layout.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/layout.html.twig
@@ -39,8 +39,9 @@
     {{ sonata_block_render_event('sylius.admin.layout.sidebar_down') }}
 {% endblock %}
 
-{% block footer %}
-    {{ 'sylius.ui.powered_by'|trans }} <a href="http://sylius.org" target="_blank">Sylius</a>.
+{% block post_content %}
+    <div class="ui divider"></div>
+    {{ 'sylius.ui.powered_by'|trans }} <a href="http://sylius.org" target="_blank">Sylius v{{ sylius_meta.version }}</a>
 {% endblock %}
 
 {% block javascripts %}
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml
@@ -93,6 +93,8 @@ twig:
     exception_controller: "FOS\\RestBundle\\Controller\\ExceptionController::showAction"
     globals:
         sylius: "@sylius.context.shopper"
+        sylius_meta:
+            version: !php/const:Sylius\Bundle\CoreBundle\Application\Kernel::VERSION
 
 sonata_block:
     blocks:


gabiudrescu:
let's take them one by one:
1. APP_NAME I'm ok with that, we can remove it if you consider it too much. 
2. I don't think my PR introduces a BC Break, because the `footer` block is not defined in Sylius/src/Sylius/Bundle/UiBundle/Resources/views/Layout/sidebar.html.twig nor Sylius/src/Sylius/Bundle/AdminBundle/Resources/views/layout.html.twig; all I did is change Sylius/src/Sylius/Bundle/AdminBundle/Resources/views/layout.html.twig to match the block name in Sylius/src/Sylius/Bundle/UiBundle/Resources/views/Layout/sidebar.html.twig; doing it the other way around would probably mean a BC Break.

lchrusciel:
I totally agree with the second point. I just don't think that we need this `APP_NAME` constant. If you would like to remove it, I'm ready to merge it right away. But if you think, that it would be better to introduce it, then just wait for others and if any of them would merge this PR, I'm also ok with that.

gabiudrescu:
OK, I have reverted that change.

pamil:
Couldn't spot any BC break here, is it still there? /cc @Zales0123

lchrusciel:
For me it is good to go

app/config/config.yml
```diff
@@ -61,3 +61,8 @@ fos_rest:
         rules:
             - { path: '^/api/.*', priorities: ['json', 'xml'], fallback_format: json, prefer_extension: true }
             - { path: '^/', stop: true }
+
+twig:
```
pamil: Can we add it to `@CoreBundle/Resources/config/app/config.yml` instead? Btw. we can name it `sylius_meta`.

pjedrzejewski: 👍

gabiudrescu: done.

lchrusciel:
Thanks Gabi!

Issue #8683, opened 2017-09-18
Add disable_purger option to SyliusFixtures


bitbager:
It would be nice to decide which fixture requires purging the database. For example, my CMS plugin needs some block or pages fixtures to be configurable under its suite, but they should not affect the whole db state.

pamil:
#8685 for the rescue :)

PR #8684, opened 2017-09-18
[Fixtures] Add disable_purger option

bitbager:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | no |
| Related tickets | mentioned in #8683 |
| License         | MIT |


src/Sylius/Bundle/FixturesBundle/Listener/ORMPurgerListener.php
```diff
@@ -76,20 +80,25 @@ protected function configureOptionsNode(ArrayNodeDefinition $optionsNode): void
 
         $optionsNodeBuilder
             ->enumNode('mode')
-                ->values(['delete', 'truncate'])
```
lchrusciel: I'm pretty sure we used to add another level of indention for node specification.

PR #8690, opened 2017-09-19

SyliusBot:


lchrusciel:
@pamil probably we need @SyliusBot to apply changes to 1.0 branch instead of master

PR #8692, opened 2017-09-19
Remove `|` character

ad3n:
Remove `|` character

| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |


lchrusciel:
This is a proper value. It is a little bit hackish, but we need a pipeline to treat following lines as a string instead of an array. It is related to Symfony internal typehints

PR #8709, opened 2017-09-21

SyliusBot:


Zales0123:
@pamil is there any possibility to force @SyliusBot to open PRs to branch `1.0`?

pamil:
We will need to set up versioning strategy on CrowdIn and customise SyliusBot, it'll take a while. We can make PRs and base on 1.0 but then `master` would be out of sync (but that'll do just fine for the very nearest future).

pamil:
I'm merging it to master and then I'm going to sync 1.0.

PR #8737, opened 2017-09-26
[ProductAttribute] Fix select product attribute choice value rename

### LINKED ISSUE #8308: Select attribute values get lost on attribute value rename
### Steps to reproduce:
1. create a select attribute with the following configuration: name "Mug material" - values: "Banana skin" and "Prcelan"
2. create a product and assign the attribute "Mug material" with the value "Prcelan"
3. realize you have a typo in the product attribute value: it is "Prcelan" instead of "Porcelain" 
image
4. modify the attribute value from "Prcelan" to "Porcelain"
image
5. observe the product attributes
image

### What happens:
The product attribute value gets lost

### What should happen:
"Prcelan" should be linked to the product through a reference, otherwise, whenever the attribute value gets changed, the products get obsolete references.

Based on the feedback from @pamil on #8220

GSadee:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | partially fixes #8308 |
| License         | MIT |

This PR fixes losing reference to an attribute value on a product when select attribute choice value renames.


pamil:
Can we split these two fixes to two standalone PRs? Would be easier to review & merge then :)

pamil:
Btw. as these are bugfixes, it should be targeted against 1.0

src/Sylius/Behat/Context/Setup/ProductAttributeContext.php
```diff
@@ -117,6 +140,24 @@ public function thisAttributeHasSetMinValueAsAndMaxValueAs(ProductAttributeInter
...
 
     /**
+     * @Given /^(this product) has (.+?) attribute "([^"]+)" with value "([^"]+)", "([^"]+)"$/
+     */
+    public function thisProductHasSelectAttributeWithValue(
+        ProductInterface $product,
+        string $productAttributeType,
+        string $productAttributeName,
+        string $attributeValue,
+        string $language = 'en_US'
```
pamil: Isn't the language set to the last `"([^"]+)"`?

src/Sylius/Behat/Context/Setup/ProductAttributeContext.php
```diff
@@ -117,6 +140,24 @@ public function thisAttributeHasSetMinValueAsAndMaxValueAs(ProductAttributeInter
...
+    ) {
+        $attribute = $this->provideProductAttribute($productAttributeType, $productAttributeName);
+        $attribute->setConfiguration(['multiple' => false, 'choices' => [$attributeValue], 'min' => null, 'max' => null]);
+        $attributeValue = $this->createProductAttributeValue(array_keys($values), $attribute, $language);
```
pamil: Where's `$values` defined?

src/Sylius/Behat/Context/Setup/ProductAttributeContext.php
```diff
@@ -117,6 +140,24 @@ public function thisAttributeHasSetMinValueAsAndMaxValueAs(ProductAttributeInter
     }
 
     /**
+     * @Given /^(this product) has (.+?) attribute "([^"]+)" with value "([^"]+)", "([^"]+)"$/
+     */
+    public function thisProductHasSelectAttributeWithValue(
```
pamil: Looks like an abandoned step.

src/Sylius/Behat/Context/Ui/Admin/ManagingProductAttributesContext.php
```diff
@@ -233,6 +254,25 @@ public function iWantToSeeAllProductAttributesInStore()
     }
 
     /**
+     * @When /^(the administrator) changes the value "([^"]*)" to "([^"]*)" of (this product attribute)$/
```
pamil: It doesn't feel right reading this step, what about the something like: the administrator changes this product attribute value "foo" to "bar"?

src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/Configuration/SelectAttributeChoicesCollectionType.php
```diff
@@ -67,16 +74,10 @@ public function getBlockPrefix(): string
     }
 
     /**
-     * @param string $value
-     *
      * @return string
      */
-    private function getValidFormKey(string $value): string
+    private function getValidFormKey(): string
```
pamil: getValidFormKey -> getUniqueKey?

tests/Responses/Expected/product_attribute/create_select_response.json
```diff
@@ -3,11 +3,7 @@
     "code": "mug_color",
     "type": "select",
     "configuration": {
-        "choices": {
-            "yellow": "yellow",
-            "green": "green",
-            "black": "black"
-        },
+        "choices": @array@,
```
pamil: IMO we should change it to something like:
```
{
 "@string@": "yellow",
 "@string@": "green",
 // etc..
}
```

features/product/managing_product_attributes/adding_select_product_attribute.feature
```diff
@@ -10,21 +10,21 @@ Feature: Adding a new select product attribute
 
     @ui @javascript
     Scenario: Adding a new select product attribute
-        Given I want to create a new select product attribute
-        When I specify its code as "mug_material"
+        When I want to create a new select product attribute
+        And I specify its code as "mug_material"
         And I name it "Mug material" in "English (United States)"
-        And I add material "-100% Banana Skin"
+        And I add value "-100% Banana Skin"
```
pamil: Value is kinda misleading, my first - AttributeValue, the second - the same. What about naming it an "option" or a "select option"?

tests/Responses/Expected/product_attribute/create_select_response.json
GSadee: I know and I agree that current example doesn't check behaviour properly, but this feature is not implemented in php-matcher (there is even an issue: https://github.com/coduo/php-matcher/issues/92)

features/product/managing_product_attributes/adding_select_product_attribute.feature
GSadee: But we have values in configuration of a product attribute, so I wanted to be consistent with the layout:
zrzut ekranu 2017-09-28 o 15 54 29

pamil: I see the reasoning behind it, but it's still confusing, not sure whether we should solve it already or leave it as is :)

tests/Responses/Expected/product_attribute/create_select_response.json
pamil: What about adding an additional assertion in the test file, eg. asserting that the response matches this file contents +  getting `$response['configuration']['choices']` and asserting they are what we expect (like https://github.com/Sylius/SyliusElasticSearchPlugin/blob/master/tests/Controller/SearchControllerApiTest.php#L436)?

GSadee: Done 👍

pamil:
Thank you Rzegś!

PR #8742, opened 2017-09-26
Do flush after enabling a user in user registration listener

### LINKED ISSUE #8720: No flush in UserRegistrationListener if channel does not require verification

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes #8720 |
| License         | MIT |


Zales0123:
Just wondering, maybe we can provide a scenario that would reproduce this bug to be 100% sure it's correct?

pamil:
@Zales0123 it's currently working right now cause there's something else in the code making the flush, so e2e scenario won't work correctly in this case.

Zales0123:
Great 🌮

PR #8746, opened 2017-09-27
Use sylius-labs/coding-standard

pamil:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT


pamil:
What do you mean by that? This PR only replaces explicit configuration of filter set by the implicit one provided by 3rd party.

pjedrzejewski:
Thank you Kamil! What guys meant is to document that we are using this tool at all. So that other contributors know about it and can maybe contribute something to the SyliusLabs/Coding-Standard repo.

pamil:
All right, that's what I had in mind, to make this coding standard easily available (Sylius/PluginSkeleton#53) and a common one for Sylius-based projects. I think I'll include it in plugins section in the docs.

composer.json
```diff
@@ -91,7 +91,8 @@
         "phpunit/phpunit": "^5.6",
         "se/selenium-server-standalone": "^2.52",
         "stripe/stripe-php": "^4.1",
-        "symplify/easy-coding-standard": "^2.3"
+        "symplify/easy-coding-standard": "^2.4",
```
TomasVotruba: I think you can drop this dependency, since it's already included in your custom coding standard: https://github.com/SyliusLabs/CodingStandard/blob/e50212b2be8f1e5984c69b9bdf205b32daa06688/composer.json#L15

PR #8751, opened 2017-09-29
[ProductAttribute] Fix select product attribute choice value delete

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially fixes #8308
| License         | MIT

PR #8206, opened 2017-06-26
[WIP][Attribute][Translation][Select] Add translation to the select attribute

Lowlo:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | yes |
| BC breaks?      | yes (with migration script) |
| License         | MIT |

A test in order to implement translation on a select attribute

PR #8766, opened 2017-10-04
[ProductAttribute] Make select attribute translatable

### LINKED ISSUE #8220: RFC - Translatable atribute values behavior
Let's say we have a Sylius instance that operates in two languages:
 - en_US
 - fr_FR

And this shop is selling tShirts, that have 3 mandatory attributes:
 - fabric
 - max. degrees for washing
 - collar type

Fabric is a select attribute, that allows the following options:
 - cotton
 - linen
 - polyester

Max. degrees for washing is an integer attribute, while Collar type is a string attribute.

----

The current implementation for selectable attributes in Sylius allows the following configuration for selectable attributes:

image

As a business owner, 
I am not able to define in this screen the french translation of cotton or linen. 

The only way I can do that with the current format is put all values, no matter what's the locale and select the translation on product level. Example:

image

English attribute value for product:

image

French attribute value for product:
image

This solution is cumbersome and does not scale very well when you have more than 2 locales or attributes with more than 10 options for each locale. 

----

Another issue discovered with the current implementation is with integer or percentage attributes. Those attributes do not require a translation, as do string attributes.

Based on my initial example:
image

While it makes sense to add different translatable values on product for string attributes, for integer and percentage, it doesn't. See above that, as a content editor, you enter the value 100 twice.  This is not a thing you want to do when you have like... 6 locales.

or 4-5 integer attributes on a 6 locales shop.

----

My proposal is to define attribute values as translatable or not translatable. The followings must be translatable:
 - select
 - text
 - textarea
 
While the followings do not make any sense to be translatable:
 - checkbox
 - integer
 - percentage
 - date
 - datetime

Also, in my opinion, the select attribute values must be translatable, not the assignment on the product. Our custom implementation is like this:
Selectable attribute on product (en_US):
image
Same selectable attribute on product (ro_RO):
image

And for en_US.
image

The attribute options are configured in a screen like this:
image

While for string attributes it looks like this:

image

And for integer attributes like this:

image

----

My proposal with this RFC is to rework some of the current behavior in attributes until Sylius gets stable. I understood from @pjedrzejewski this implies BC Break and we should consult first, before any decision.

Looking forward to hearing your opinions.

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #8220
| License         | MIT

Based on the PR #8206 created by @Lowlo and partially based on the PR #8751

### DIFF
--- a/app/migrations/Version20171003103916.php
+++ b/app/migrations/Version20171003103916.php
@@ -22,7 +22,7 @@ class Version20171003103916 extends AbstractMigration implements ContainerAwareI
     /**
      * {@inheritdoc}
      */
-    public function setContainer(ContainerInterface $container = null)
+    public function setContainer(?ContainerInterface $container = null): void
     {
         $this->container = $container;
     }
@@ -30,7 +30,7 @@ class Version20171003103916 extends AbstractMigration implements ContainerAwareI
     /**
      * @param Schema $schema
      */
-    public function up(Schema $schema)
+    public function up(Schema $schema): void
     {
         // this up() migration is auto-generated, please modify it to your needs
         $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
@@ -66,7 +66,7 @@ class Version20171003103916 extends AbstractMigration implements ContainerAwareI
     /**
      * @param Schema $schema
      */
-    public function down(Schema $schema)
+    public function down(Schema $schema): void
     {
         // this down() migration is auto-generated, please modify it to your needs
         $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
--- a/app/migrations/Version20171003103916.php
+++ b/app/migrations/Version20171003103916.php
@@ -4,13 +4,15 @@ namespace Sylius\Migrations;
 
 use Doctrine\DBAL\Migrations\AbstractMigration;
 use Doctrine\DBAL\Schema\Schema;
+use Sylius\Component\Attribute\AttributeType\SelectAttributeType;
+use Sylius\Component\Product\Model\ProductAttributeInterface;
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Auto-generated Migration: Please modify to your needs!
  */
-class Version20170626083012 extends AbstractMigration implements ContainerAwareInterface
+class Version20171003103916 extends AbstractMigration implements ContainerAwareInterface
 {
     /**
      * @var ContainerInterface
@@ -32,27 +34,31 @@ class Version20170626083012 extends AbstractMigration implements ContainerAwareI
     {
         // this up() migration is auto-generated, please modify it to your needs
         $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
-        $defaultLocale = $this->container->getParameter('locale');
-        $conn = $this->container->get('doctrine.dbal.default_connection');
 
-        $attributes = $conn->executeQuery('SELECT id, configuration FROM sylius_product_attribute WHERE `type` = "select"')->fetchAll();
-        foreach ($attributes as $attribute) {
-            $updatedConfig = ['choices' => []];
-            $configuration = unserialize($attribute['configuration']);
+        $defaultLocale = $this->container->getParameter('locale');
+        $productAttributeRepository = $this->container->get('sylius.repository.product_attribute');
+
+        $productAttributes = $productAttributeRepository->findBy(['type' => SelectAttributeType::TYPE]);
+        /** @var ProductAttributeInterface $productAttribute */
+        foreach ($productAttributes as $productAttribute) {
+            $configuration = $productAttribute->getConfiguration();
+            $upgradedConfiguration = [];
 
             foreach ($configuration as $configurationKey => $value) {
-                if ($configurationKey === 'choices') {
+                if ('choices' === $configurationKey) {
                     foreach ($value as $key => $choice) {
-                        $updatedConfig[$configurationKey][$key][$defaultLocale] = $choice;
+                        $upgradedConfiguration[$configurationKey][$key][$defaultLocale] = $choice;
                     }
-                } else {
-                    $updatedConfig[$configurationKey] = $value;
+
+                    continue;
                 }
+
+                $upgradedConfiguration[$configurationKey] = $value;
             }
 
             $this->addSql('UPDATE sylius_product_attribute SET configuration = :configuration WHERE id = :id', [
-                'id' => $attribute['id'],
-                'configuration' => serialize($updatedConfig),
+                'id' => $productAttribute->getId(),
+                'configuration' => serialize($upgradedConfiguration),
             ]);
         }
     }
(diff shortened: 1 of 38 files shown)


app/migrations/Version20171003103916.php
```diff
@@ -0,0 +1,103 @@
...
+{
+    /**
+     * @var ContainerInterface
+     */
+    private $container;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setContainer(ContainerInterface $container = null)
```
pamil: `?ContainerInterface`?

features/product/viewing_products/viewing_select_product_attributes_in_different_locales.feature
```diff
@@ -0,0 +1,30 @@
...
+        And this product has also a select attribute "T-shirt colour" with value "Yellow" in "English (United States)" locale
+
+    @ui
+    Scenario: Viewing a detailed page with product's select attribute in default locale
+        When I view product "T-shirt banana"
+        Then I should see the product attribute "T-shirt material" with value "Banana skin"
+        And I should also see the product attribute "T-shirt colour" with value "Yellow"
+
+    @ui
+    Scenario: Viewing a detailed page with product's select attribute after locale change
```
pamil: `after locale change` -> `in different locale`?

src/Sylius/Behat/Page/Admin/ProductAttribute/UpdatePage.php
```diff
@@ -68,18 +69,25 @@ public function hasAttributeValue(string $value): bool
...
 
     /**
      * {@inheritdoc}
      */
     public function deleteAttributeValue(string $value): void
     {
-        $attributeChoiceElement = $this->getElement('attribute_choice_list_element', ['%value%' => $value])->getParent();
+        $attributeChoiceElement = $this
+            ->getElement('attribute_choice_list_element', ['%value%' => $value])
+            ->getParent()->getParent()->getParent()->getParent()
```
pamil: Hmm, is there any way we can get rid of those parent calls?

src/Sylius/Bundle/AdminBundle/Resources/views/Form/theme.html.twig
```diff
@@ -1,5 +1,24 @@
 {% extends '@SyliusUi/Form/theme.html.twig' %}
 
+{% block _sylius_product_attribute_configuration_choices_entry_widget %}
+    {% import '@SyliusUi/Macro/flags.html.twig' as flags %}
+
+    <div class="ui styled fluid accordion">
+        {% for translation in form %}
+            <div data-locale="en_US">
```
pamil: Hardcoded `en_US`?

src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/Configuration/SelectAttributeChoicesCollectionType.php
```diff
@@ -80,4 +98,21 @@ private function getUniqueKey(): string
...
+    /**
+     * @param array $values
+     *
+     * @return array
+     */
+    private function resolveValues(array $values): array
+    {
+        $fixedValues = [];
+        foreach ($values as $locale => $value) {
+            if (!empty($value)) {
```
pamil: This will filter out `0`.

src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/Configuration/SelectAttributeValueTranslationsType.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setDefaults([
+            'entries' => $this->definedLocalesCodes,
+            'entry_name' => function ($localeCode) {
```
pamil: Strict types can be added here (and to this class entirely)

src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/SelectAttributeType.php
```diff
@@ -26,6 +26,19 @@
...
 {
     /**
+     * @var string $defaultLocale
+     */
+    protected $defaultLocale;
+
+    /**
+     * @param string $defaultLocale
+     */
+    public function __construct($defaultLocale)
```
pamil: Missing typehint

src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/SelectAttributeType.php
```diff
@@ -26,6 +26,19 @@
 final class SelectAttributeType extends AbstractType
 {
     /**
+     * @var string $defaultLocale
+     */
+    protected $defaultLocale;
```
pamil: Can be private

src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/SelectAttributeType.php
```diff
@@ -67,12 +80,28 @@ public function configureOptions(OptionsResolver $resolver): void
...
             ->setNormalizer('choices', function (Options $options) {
                 if (is_array($options['configuration'])
                     && isset($options['configuration']['choices'])
                     && is_array($options['configuration']['choices'])) {
-                    $choices = array_flip($options['configuration']['choices']);
+                    $choices = [];
+                    $localeCode = $options['locale_code'];
+
+                    foreach ($options['configuration']['choices'] as $key => $choice) {
+                        if (!empty($choice[$localeCode])) {
```
pamil: Filtering out zeros.

src/Sylius/Bundle/AttributeBundle/Resources/views/Types/select.html.twig
```diff
@@ -2,9 +2,17 @@
...
+            {% else %}
+                {{ values[value][fallbackLocale] }}{% if loop.last == false %}, {% endif %}
+            {% endif %}
         {% endfor %}
     {% else %}
-        {{ values[attribute.value] }}
+        {% if values[attribute.value][locale] is not empty %}
+            {{ values[attribute.value][locale] }}
+        {% else %}
+            {{ values[attribute.value][fallbackLocale] }}
```
pamil: What happens if `values[attribute.value][fallbackLocale]` is not defined?

src/Sylius/Behat/Page/Admin/ProductAttribute/UpdatePage.php
GSadee: I've been trying to come up with better solution, but without success. If you have any suggestion, please share with me 😃

src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/Configuration/SelectAttributeChoicesCollectionType.php
```diff
@@ -35,24 +49,28 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+                    if (!array_key_exists($this->defaultLocaleCode, $values)) {
                         continue;
                     }
 
                     $newKey = $this->getUniqueKey();
-                    $fixedArray[$newKey] = $value;
+                    $fixedData[$newKey] = $this->resolveValues($values);
 
-                    if ($form->offsetExists($key)) {
+                    if (!is_null($newKey) && $form->offsetExists($key)) {
```
pamil: `$newKey` cannot be null here as `getUniqueKey` returns `string`.

src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/Configuration/SelectAttributeValueTranslationsType.php
```diff
@@ -0,0 +1,76 @@
...
+                    'required' => $localeCode === $this->defaultLocaleCode,
+                ];
+            }
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getParent()
```
pamil: Missing typehints here (+ other method definitions).

src/Sylius/Bundle/AttributeBundle/Form/Type/AttributeType/SelectAttributeType.php
```diff
@@ -26,6 +26,19 @@
...
+     * @var string $defaultLocale
+     */
+    private $defaultLocale;
+
+    /**
+     * @param string $defaultLocale
+     */
+    public function __construct(string $defaultLocale)
+    {
+        $this->defaultLocale = $defaultLocale;
```
pamil: `SelectAttributeChoicesCollectionType` uses `LocaleProviderInterface::getDefaultLocaleCode()` to get its default locale, this one has base locale injected as `$defaultLocale`, is that intentional?

pamil: Anyway, if it's intentional, I guess `$baseLocale` would be more appropriate name then.

src/Sylius/Bundle/AttributeBundle/Resources/config/services/attribute_types.xml
```diff
@@ -56,5 +56,20 @@
...
+
+        <service id="sylius.form.type.attribute_type.select.choices_collection" class="Sylius\Bundle\AttributeBundle\Form\Type\AttributeType\Configuration\SelectAttributeChoicesCollectionType">
+            <argument id="sylius.translation_locale_provider" type="service"/>
+            <tag name="form.type" />
+        </service>
+
+        <service id="sylius.attribute_type.select.value.translations" class="Sylius\Bundle\AttributeBundle\Form\Type\AttributeType\Configuration\SelectAttributeValueTranslationsType">
+            <argument id="sylius.translation_locale_provider" type="service"/>
+            <tag name="form.type" />
+        </service>
```
pamil: I guess there's still something wrong with those definitions, `%locale%` points to a base locale code, whereas `sylius.translation_locale_provider` service has `getDefaultLocaleCode` method which returns current channel's default locale. Is there any reason for using these two simultaneously instead of only one of them?

GSadee: I didn't notice that 👍 
I've changed to use `sylius.translation_locale_provider` instead of `%locale%`.

pamil:
Thanks @GSadee, especially for your patience! :)

PR #8771, opened 2017-10-04
Make Sylius compatible with PHP 7.2

stefandoorn:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Run test suite already against PHP 7.2 to pro-actively make adjustments before PHP 7.2 is released (if possible). Not sure if this should be against 1.0 or master though ;-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,14 @@ matrix:
                 - memcached
         -
             sudo: false
+            php: 7.2
+            env: SYLIUS_SUITE="packages"
+            addons:
+                apt:
+                    packages:
+                        - parallel
+        -
+            sudo: false
             php: 7.1
             env: SYLIUS_SUITE="docs packages"
             addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,14 @@ matrix:
             services:
                 - memcached
         -
+            sudo: required
+            php: 7.2
+            env:
+                - SYLIUS_SUITE="application"
+                - TRAVIS_NODE_VERSION="7.5"
+            services:
+                - memcached
+        -
             sudo: false
             php: 7.1
             env: SYLIUS_SUITE="docs packages"
--- a/composer.json
+++ b/composer.json
@@ -37,15 +37,18 @@
         "fzaninotto/faker": "^1.6",
         "gedmo/doctrine-extensions": "^2.4.12",
         "incenteev/composer-parameter-handler": "^2.1",
+        "jeremykendall/php-domain-parser": "^4.0.3@alpha",
         "jms/serializer-bundle": "^2.0",
         "knplabs/knp-gaufrette-bundle": "^0.3",
         "knplabs/knp-menu-bundle": "^2.1",
+        "league/uri": "^4.2.3",
         "liip/imagine-bundle": "^1.9.1",
         "ocramius/proxy-manager": "^2.1",
         "payum/payum": "^1.4",
         "payum/payum-bundle": "^2.2",
         "php-http/guzzle6-adapter": "^1.1",
         "polishsymfonycommunity/symfony-mocker-container": "^1.0",
+        "ramsey/uuid": "^3.7",
         "sensio/distribution-bundle": "^5.0.21",
         "sonata-project/block-bundle": "^3.3",
         "sonata-project/intl-bundle": "^2.2",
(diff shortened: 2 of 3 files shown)


stefandoorn:
First issue spotted: `friendsofphp/php-cs-fixer` needs to be at 2.6.1 at least to support PHP 7.2.

stefandoorn:
I'm not sure how to figure out what the Behat errors are. Output seems scarce.

gabiudrescu:
@stefandoorn if you check this etc/travis/run-suite after_failure "${SYLIUS_SUITE}" the build log, you can reach this:

http://termbin.com/uw75

which is pretty verbose.

stefandoorn:
Ok..

```
idn_to_ascii(): INTL_IDNA_VARIANT_2003 is deprecated in /home/travis/build/Sylius/Sylius/vendor/league/uri/src/Components/HostnameTrait.php line 68
```

`league/uri` is required by Payum at `~4.0`. Not sure about it, but issue might be fixed in `league/uri` 5.x. Doesn't seem Payum already supports this, so this might be a hard path to get fixed.

stefandoorn:
Anyone also knows the URL on which the imgur uploaded screenshots can be found?

pamil:
@stefandoorn looks like our imgur API key is not valid so uploading screenshots has stopped working.

.travis.yml
```diff
@@ -18,13 +19,31 @@ matrix:
...
+            sudo: required
+            php: 7.2
+            env:
+                - SYLIUS_SUITE="application"
+                - TRAVIS_NODE_VERSION="7.5"
+            services:
+                - memcached
+        -
+            sudo: required
+            php: nightly
```
pamil: I'd skip the nightly PHP version for now as there's still some time until PHP 7.3 gets released & builds will take more time to finish.

pamil: Instead, running packages build on PHP 7.2 is a must :)

composer.json
```diff
@@ -88,7 +88,7 @@
         "matthiasnoback/symfony-dependency-injection-test": "^1.0",
         "mikey179/vfsStream": "^1.6",
         "pamil/prophecy-common": "^0.1",
-        "phpspec/phpspec": "^4.0",
+        "phpspec/phpspec": "^4.0@dev",
```
pamil: Same change applies for every Sylius package using phpspec (packages build on PHP 7.2 would fail without this one).

pamil (changes requested):

pamil:
> Not sure if this should be against 1.0 or master though ;-)

IMO it should go for 1.0 branch as 1.0 requires PHP ^7.1, so PHP 7.2 should also be supported there.

pamil:
As for that Payum-related errors:
 - Payum itself requires PHP 5.5+
 - Payum requires `thephpleague/uri`:
   - `thephpleague/uri` v4 requires PHP 5.5+
   - `thephpleague/uri` v4 uses the deprecated function call
   - `thephpleague/uri` v4 requires `jeremykendall/php-domain-parser` v3:
     - `jeremykendall/php-domain-parser` v3 uses the deprecated function call
   - `thephpleague/uri` v5 requires PHP 7.0+
   - `thephpleague/uri` v5 requires `jeremykendall/php-domain-parser` v4.0.3-alpha:
     -`jeremykendall/php-domain-parser` v4.0.3-alpha does not use the deprecated function call

Conclusion:
 - Getting Payum to use the newer `thephpleague/uri` would require to drop PHP 5.5 / 5.6 support for them (and probably a lot of work)
 - We can try to send bugfixes to `thephpleague/uri` v4 and `jeremykendall/php-domain-parser` v3 to use that not deprecated variant in `idn_to_*` functions (like done here - https://github.com/jeremykendall/php-domain-parser/blob/4.0.3-alpha/src/Pdp/Parser.php#L318-L322)

pamil:
jeremykendall/php-domain-parser#204 and thephpleague/uri-src#41 have been opened, let's wait for the replies.

.travis.yml
stefandoorn: Mainly added nightly for now as it might discover some issues now already that we can tackle directly. Will take it out before finalising PR.

composer.json
stefandoorn: Alright. Need to see if this change is required at all before finalising PR.

stefandoorn:
@pamil Ready to review. Had to explicitly require the updated dependencies to get the proper versions installed. 

Build on PHP 7.3 fails due to PHP-CS-Fixer not allowing 7.3. I'm not sure we should disallow installation on PHP 7.3 due to this.

pamil (commented):
Awesome! Can you add `packages` build for PHP 7.2? It's like the one for PHP 7.1 but with `docs` removed from `$SYLIUS_SUITE` 🎉

pamil (approved):

pamil:
Thanks Stefan, great job! 🥇

PR #8775, opened 2017-10-05
[ProductAttribute] Fix displaying attributes depending on locales and channels

### LINKED ISSUE #8220: RFC - Translatable atribute values behavior

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially fixes #8220
| License         | MIT

### DIFF
--- a/features/product/viewing_products/viewing_product_attributes_in_different_locales_and_channels.feature
+++ b/features/product/viewing_products/viewing_product_attributes_in_different_locales_and_channels.feature
@@ -1,0 +1,45 @@
+@viewing_products
+Feature: Viewing product's attributes in different locales and channels
+    In order to see product's specification in different locales and channels
+    As a Visitor
+    I want to be able to see product's attributes in a chosen locale and a channel
+
+    Background:
+        Given the store operates on a channel named "US Channel"
+        And that channel allows to shop using "English (United States)" and "Polish (Poland)" locales
+        And it uses the "English (United States)" locale by default
+        And the store operates on another channel named "PL Channel"
+        And that channel allows to shop using "English (United States)" and "Polish (Poland)" locales
+        And it uses the "Polish (Poland)" locale by default
+        And the store has a product "T-shirt banana" available in "US Channel" channel
+        And this product is also available in "PL Channel" channel
+        And this product has text attribute "T-shirt material" with value "Banana skin" in "English (United States)" locale
+        And this product has text attribute "T-shirt material" with value "Skórka banana" in "Polish (Poland)" locale
+        And this product has textarea attribute "T-shirt details" with value "Banana is a very good material." in "English (United States)" locale
+        And I am browsing the channel "US Channel"
+
+    @ui @todo
+    Scenario: Viewing a detailed page with product's attribute for current channel with its default locale
+        When I view product "T-shirt banana" in the "Polish (Poland)" locale
+        Then I should see the product attribute "T-shirt material" with value "Skórka banana"
+        And I should also see the product attribute "T-shirt details" with value "Banana is a very good material."
+
+    @ui @todo
+    Scenario: Viewing a detailed page with product's attribute for current channel in different locale
+        When I view product "T-shirt banana" in the "English (United States)" locale
+        Then I should see the product attribute "T-shirt material" with value "Banana skin"
+        And I should also see the product attribute "T-shirt details" with value "Banana is a very good material."
+
+    @ui @todo
+    Scenario: Viewing a detailed page with product's attribute for different channel with its default locale
+        When I change my current channel to "PL Channel"
+        And I view product "T-shirt banana" in the "Polish (Poland)" locale
+        Then I should see the product attribute "T-shirt material" with value "Skórka banana"
+        And I should also see the product attribute "T-shirt details" with value "Banana is a very good material."
+
+    @ui @todo
+    Scenario: Viewing a detailed page with product's attribute for different channel in different locale
+        When I change my current channel to "PL Channel"
+        And I view product "T-shirt banana" in the "English (United States)" locale
+        Then I should see the product attribute "T-shirt material" with value "Banana skin"
+        And I should also see the product attribute "T-shirt details" with value "Banana is a very good material."
--- a/features/product/viewing_products/viewing_product_attributes_in_different_locales_and_channels.feature
+++ b/features/product/viewing_products/viewing_product_attributes_in_different_locales_and_channels.feature
@@ -18,26 +18,26 @@ Feature: Viewing product's attributes in different locales and channels
         And this product has textarea attribute "T-shirt details" with value "Banana is a very good material." in "English (United States)" locale
         And I am browsing the channel "US Channel"
 
-    @ui @todo
+    @ui
     Scenario: Viewing a detailed page with product's attribute for current channel with its default locale
         When I view product "T-shirt banana" in the "Polish (Poland)" locale
         Then I should see the product attribute "T-shirt material" with value "Skórka banana"
         And I should also see the product attribute "T-shirt details" with value "Banana is a very good material."
 
-    @ui @todo
+    @ui
     Scenario: Viewing a detailed page with product's attribute for current channel in different locale
         When I view product "T-shirt banana" in the "English (United States)" locale
         Then I should see the product attribute "T-shirt material" with value "Banana skin"
         And I should also see the product attribute "T-shirt details" with value "Banana is a very good material."
 
-    @ui @todo
+    @ui
     Scenario: Viewing a detailed page with product's attribute for different channel with its default locale
         When I change my current channel to "PL Channel"
         And I view product "T-shirt banana" in the "Polish (Poland)" locale
         Then I should see the product attribute "T-shirt material" with value "Skórka banana"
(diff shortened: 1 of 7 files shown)


GSadee:
@pamil I've added some specs for test this behaviour

src/Sylius/Component/Product/Model/Product.php
```diff
@@ -440,22 +448,49 @@ protected function createTranslation(): ProductTranslationInterface
...
+     */
+    private function hasNotEmptyAttributeByCodeAndLocale(string $attributeCode, string $localeCode): bool
+    {
+        $attributeValue = $this->getAttributeByCodeAndLocale($attributeCode, $localeCode);
+
+        if (null === $attributeValue) {
+            return false;
+        }
+
+        if (empty($attributeValue->getValue())) {
```
pamil: I guess we do not want it to be an `empty` call.

> The following things are considered to be empty:
>  - "" (an empty string)
>  - 0 (0 as an integer)
>  - 0.0 (0 as a float)
>  - "0" (0 as a string)
>  - NULL
>  - FALSE
>  - array() (an empty array)
>  - $var; (a variable declared, but without a value)

src/Sylius/Component/Product/spec/Model/ProductSpec.php
```diff
@@ -120,6 +122,66 @@ function it_refuses_to_remove_non_product_attribute(AttributeValueInterface $att
...
+        $attributeValuePL->getCode()->willReturn('colour');
+        $attributeValuePL->getValue()->willReturn('Niebieski');
+
+        $this->addAttribute($attributeValueEN);
+        $this->addAttribute($attributeValuePL);
+
+        $this
+            ->getAttributesByLocale('pl_PL', 'en_US')
+            ->toArray()
+            ->shouldReturn([$attributeValuePL->getWrappedObject()])
```
pamil: Let's add another one which is testing the same call but returning the attribute from the fallback locale.

src/Sylius/Component/Product/spec/Model/ProductSpec.php
```diff
@@ -120,6 +122,66 @@ function it_refuses_to_remove_non_product_attribute(AttributeValueInterface $att
...
+        $attributeValueFR->getCode()->willReturn('colour');
+        $attributeValueFR->getValue()->willReturn('Bleu');
+
+        $this->addAttribute($attributeValueEN);
+        $this->addAttribute($attributeValueFR);
+
+        $this
+            ->getAttributesByLocale('pl_PL', 'en_US', 'fr_FR')
+            ->toArray()
+            ->shouldReturn([$attributeValueEN->getWrappedObject()])
```
pamil: Let's add another ones which are testing the same call but:
 - returning the attribute from `$localeCode` locale
 - returning the attribute from `$baseLocaleCode` locale

src/Sylius/Component/Product/spec/Model/ProductSpec.php
pamil: Btw. instead of `$this->getAttributesByLocale(...)->toArray()->shouldReturn([$mock->getWrappedObject()]);` you can use `$this->getAttributesByLocale(...)->shouldIterateAs([$mock]);`

src/Sylius/Component/Product/Model/Product.php
pamil: I'm afraid we're missing the tests for attributes that have values existing in base locale, but these values are considered empty (especially for booleans or integers).

Like with base locale `fr_FR` and current locale `pl_PL` and boolean attribute, what happens if:
 - boolean attribute has `false` value in `fr_FR` but `true` in `pl_PL`
 - integer attribute has `0` value in `fr_FR` but `1` in `pl_PL`

pamil:
I've been thinking for some time whether to classify this as a bug fix or a new feature, but it looks like it is the latter. `getAttributeByCodeAndLocale` behaviour stays the same if the third parameter is not passed.

What do you think about targeting `master` branch with this PR? Some parts can be still merged to 1.0, like the first two Behat scenarios and specs that does not call that method with the thrid parameter, but it's not necessary.

GSadee:
@pamil IMO this is a bug fix, because even if the behaviour of `getAttributeByCodeAndLocale` has not been changed, adding the third parameter in twig template on shop fixes the behaviour of displaying attributes there.

And these behat scenarios cannot be merge to `1.0` without changing the `getAttributeByCodeAndLocale`, because they will fail 💥

pamil:
From my POV the current attributes behaviour is like for any other translatable resources: try to fetch a translation for current locale, if not found - try fallback locale. This PR adds a new feature, which is including also attributes in base locale.

As for Behat, the first two scenarios should pass, the other two are describing new features.

Anyway, it might be helpful to hear some other opinion than ours :)

features/product/viewing_products/viewing_product_attributes_in_different_locales_and_channels.feature
```diff
@@ -0,0 +1,45 @@
...
+        And I am browsing the channel "US Channel"
+
+    @ui
+    Scenario: Viewing a detailed page with product's attribute for current channel with its default locale
+        When I view product "T-shirt banana" in the "Polish (Poland)" locale
+        Then I should see the product attribute "T-shirt material" with value "Skórka banana"
+        And I should also see the product attribute "T-shirt details" with value "Banana is a very good material."
+
+    @ui
+    Scenario: Viewing a detailed page with product's attribute for current channel in different locale
```
lchrusciel: The name of this scenario should have it is name changed with the scenario above. This one is using a default locale

src/Sylius/Component/Attribute/Model/AttributeSubjectInterface.php
```diff
@@ -28,10 +28,15 @@ public function getAttributes(): Collection;
...
      * @param string $fallbackLocaleCode
+     * @param string|null $baseLocaleCode
      *
      * @return Collection|AttributeValueInterface[]
      */
-    public function getAttributesByLocale(string $localeCode, string $fallbackLocaleCode): Collection;
+    public function getAttributesByLocale(
+        string $localeCode,
+        string $fallbackLocaleCode,
+        ?string $baseLocaleCode = null
```
lchrusciel: Maybe we could start adding TODO messages with the information, that from 2.0 the third parameter would be mandatory?

GSadee:
I couldn't 100% agree because this PR fixes the scenario: 
 - the base locale is `en_US`
 - the product has attribute with values in `en_US` and `pl_PL` locales 
 - this product is in channel with default locale `fr_FR`
 - in the shop, product is displaying in locale `pl_PL`

**Before:** the attribute woldn't be shown at all and it isn't a standard behaviour for translatable entities for me :)
**After:** the attribute would be shown in `pl_PL`

src/Sylius/Component/Attribute/Model/AttributeSubjectInterface.php
pamil: If we intend to make it required then `E_USER_DEPRECATED` error should be triggered if value was not passed. To be honest, I'd rather make fallback locale code optional as well (there's no method to get product attributes values in given locale only now).

pamil:
Thanks @GSadee for your patience! ☮️

PR #8788, opened 2017-10-06
[Shop] Use first variant image in cart if one exists

### LINKED ISSUE #8787: Dedicated Media for Product Variant is not shown in cart
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| BC Break report? | unknown
| RFC?             | maybe - if it turns out the current implementation is intended
| Sylius version   | 1.1

From Slack:

> jan
> do you know if there is any special reason why I can only add (and associate) media to a variant in the product form but not in the product variant form?
> 
> I also recognized, that if you add two variants of a product into the cart, where each variant has its own associated image, the cart only shows the first image of the product but not the associated image to the variant. Would you say this is a desired behavior or a bug?
> 
> Posted in #supportToday at 3:58 PM

----

Steps to reproduce:
 - go to admin to a product with multiple variants
 - upload two different pictures for two variants
 - add both variants to cart
 - go to cart

What happens:
 - observe the images in cart for the product are the same

What should happen:
 - you should see the image assigned to the variant:

image

image

disclaimer: I have used a picture from bestvalue.eu just because it was handy :grin: it was not intended. some other time I used one with Donald Trump.

castler:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #8787
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/Product/_info.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/Product/_info.html.twig
@@ -1,7 +1,11 @@
 {% set product = variant.product %}
 
 <div class="ui header">
-    {% include '@SyliusShop/Product/_mainImage.html.twig' with {'product': product, 'filter': 'sylius_shop_product_tiny_thumbnail'} %}
+    {% if variant.hasImages %}
+        {% include '@SyliusShop/Product/_mainImage.html.twig' with {'product': variant, 'filter': 'sylius_shop_product_tiny_thumbnail'} %}
+    {% else %}
+        {% include '@SyliusShop/Product/_mainImage.html.twig' with {'product': product, 'filter': 'sylius_shop_product_tiny_thumbnail'} %}
+    {% endif %}
     <div class="content">
         <a href="{{ path('sylius_shop_product_show', {'slug': product.slug, '_locale': product.translation.locale}) }}" class="sylius-product-name">{{ product.name }}</a>
         <span class="sub header sylius-product-variant-code">


gabiudrescu:
I would add a scenario to ensure this won't happen again. 
also, what about API behavior?

Zales0123:
Behat scenario would be indeed useful 👍  And about API behaviour, according to [this serializer configuration](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/AdminApiBundle/Resources/config/serializer/Model.ProductVariant.yml) images are not shown for `ProductVariant`. We could adjust it, but I think it's a different feature and should be done in separate PR.

**MOREOVER**
We don't have for now managing product variants images on Admin, so it would be a dead feature 💀  as it's only possible to set images on product variant programatically. So probably the discuss should be - are we willing to manage images on variants? And if yes, do we still want images on products? /cc @Sylius/core-team

castler:
I'm fighting with behat and docker to get all behat tests working. Then I will work an a scenario for this one here.

But you can manage images for variants (while editing the product, not the variant) in admin, so its not a dead feature.

stefandoorn:
I would like to be able to manage my images at product level. Some of the products in my current project have >800 variants. In a later stage (as they depend on a color) we want to customise this, but for now it's simply not an option. I think more people will run into this option. So finetuning +1 per variant, but keep the option to revert back to the product images itself at high level.

Besides this, at product level they could be used for more marketing type of images, while on variant more on the details of it.

lchrusciel:
I would say that it is a feature request rather than bug fix and we should add it to the `1.1` milestone. Somebody has to rethink how it should behave at the first place

gabiudrescu:
I think this PR should be merged with the current master branch and get released as soon as possible (1.0.2 maybe?), because it confuses current users in terms of the item they buy.

and separately, open feature requests for the other things mentioned here (admin, api behavior and co.)

pamil:
@gabiudrescu @Zales0123 how would you implement that Behat scenario?

Zales0123:
I think we can setup variant with 2 images in the scenario background and check buy URL do we have a proper image displayed on cart page. Yes, it couples the test strongly with the HTML implementation, but only on Page object level - so it can be easily changed if needed in the future :)

gabiudrescu:
pretty much what @Zales0123 said above:
 - create 2 products for each if branch: one with variant image and one without
 - add each product in cart
 - assert each variant has the correct image displayed next to it

Zales0123:
Hi, @castler, it's been a while, but do you think you would be able to add Behat scenarios covering this feature, as proposed? It would be nice to have it finally merged :)

Zales0123:
Closing in favor of #10209

PR #8798, opened 2017-10-09
[Installation] Add setting a username during installation

### LINKED ISSUE #8613: Cannot reset password admin without setting up a username
Steps to reproduce:
- install a fresh new copy of sylius (dev-master)
- go to admin
- try to set a new password of the admin created during install

what happens:
- you are required to set up a username for that user

what should happen:
- the username should be set during install process - imho - and it should be the email address
- as an alternative, the user should be asked during install if there is any preference for the username, otherwise default username to email

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #8613
| License         | MIT

### DIFF
--- a/features/installer/install_command.feature
+++ b/features/installer/install_command.feature
@@ -12,9 +12,9 @@ Feature: Sylius Install Feature
     Scenario: Trying to register administrator account without email
         Given I do not provide an email
         When I run Sylius CLI installer
-        Then I should see output "E-mail: This value should not be blank"
+        Then I should see output "This value should not be blank"
 
     Scenario: Trying to register administrator account with an incorrect email
         Given I do not provide a correct email
         When I run Sylius CLI installer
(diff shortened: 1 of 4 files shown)


pamil:
It looks like a new feature to me (and possibly can break some automated installation procedures).
What about setting the username to be the same as email for 1.0 branch and then adding this additional question in 1.1?

stefandoorn:
Can't it be an optional parameter?

GSadee:
Username is an optional parameter, when it is left blank, email is set as a username. However, I understand that there is a possibility, that this change could break some automated installation procedures as @pamil has written. I will create new PR for `1.0` branch and change the base branch of this PR.

pamil:
Thanks! 🍰

PR #8808, opened 2017-10-10
Apply coding standard fixes

pamil:


pamil:
We should automate it somehow 🗡

PR #8812, opened 2017-10-10
[Installation] Add setting an email as username during installation

### LINKED ISSUE #8613: Cannot reset password admin without setting up a username

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #8613, related with #8798 
| License         | MIT


pamil:
Thanks @GSadee! 🌮

PR #8778, opened 2017-10-05
Changed CustomerChoiceType for a ResourceToIdentifierTransformer in OrderType of SyliusAdminApiBundle

jdeveloper:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #8776 
| License         | MIT

PR #8815, opened 2017-10-11
Revert "Changed CustomerChoiceType for a ResourceToIdentifierTransformer in OrderType of SyliusAdminApiBundle"

pamil:
Reverts #8778.

Replacing `CustomerChoiceType` with `TextType` looks like a BC break - behaviour would change if anyone has relied on form type being `CustomerChoiceType` and therefore extended it by a type extension. Moreover, it feels like it does not feel right semantically.

In order to improve performance we should use a choice loader inside our choice types. Since patch release will be out today (or at last tomorrow), reverting it and coming up with more backward compatible yet performant solution later seems like the best decision.

### DIFF
--- a/src/Sylius/Bundle/AdminApiBundle/Form/Type/OrderType.php
+++ b/src/Sylius/Bundle/AdminApiBundle/Form/Type/OrderType.php
@@ -14,13 +14,13 @@ declare(strict_types=1);
 namespace Sylius\Bundle\AdminApiBundle\Form\Type;
 
 use Sylius\Bundle\ChannelBundle\Form\Type\ChannelChoiceType;
+use Sylius\Bundle\CustomerBundle\Form\Type\CustomerChoiceType;
 use Sylius\Bundle\LocaleBundle\Form\Type\LocaleChoiceType;
 use Sylius\Bundle\ResourceBundle\Form\DataTransformer\ResourceToIdentifierTransformer;
 use Sylius\Bundle\ResourceBundle\Form\Type\AbstractResourceType;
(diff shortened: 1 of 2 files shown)


pjedrzejewski:
You are right, thanks Kamil!

jdeveloper:
@pamil @pjedrzejewski I'm sorry that wasn't aware of the BC.

I'll try to submit a pull request with the performance optimization you're proposing. But I now don't know if this has to be put in version 1.0 or 1.0.1

pamil:
No worries, thanks for contributing! If that's a performance thing only, targeting 1.0 branch is a good idea (it means the next patch release, `1.0.x`, and is merged up to `master` by us).

Chrysweel:
Is there a solution for this error? ping @pamil @pjedrzejewski

I prove Sylius with a database with 30 customers, 15.000 and 30.000.

symfony profiler

symfony profiler2

symfony profiler-0

I need fix this problem... what solution can do I ?
Somebody can say me how to overwrite in my project the class src/Sylius/Bundle/AdminApiBundle/Form/Type/OrderType.php ?

Thanks in advance

Chrysweel:
ping @pamil @pjedrzejewski 

some feedback?

pamil:
I don't have too much time right now to look into this issue and try implementing it myself, but imitating `Symfony\Bridge\Doctrine\Form\ChoiceList\DoctrineChoiceLoader` and using it would be a good start.

jdeveloper:
@pamil @pjedrzejewski I have submited a new pull request. Look for the comment I put there. Maybe the approach I took doesn't satisfy you.

PR #8839, opened 2017-10-13
Add a requirements for the minimal memory_limit setting when installing

nstapelbroek:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | None
| License         | MIT

Hi! :raising_hand_man: 

When installing Sylius for the first time I got presented with an
```
[Symfony\Component\Debug\Exception\OutOfMemoryException]                                 
  Error: Allowed memory size of 134217728 bytes exhausted (tried to allocate 32768 bytes)
```
This PR should warn the user upfront when installing Sylius about any minimal memory requirements.
I'd love to hear your thoughts about this PR. I've just started my Sylius journey so I'm not that familair with the conventions yet. For example: Should we enforce a new memory_limit if `ini_set` is available?

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Installer/Requirement/SettingsRequirements.php
+++ b/src/Sylius/Bundle/CoreBundle/Installer/Requirement/SettingsRequirements.php
@@ -18,6 +18,7 @@ use Symfony\Component\Translation\TranslatorInterface;
 final class SettingsRequirements extends RequirementCollection
 {
     public const RECOMMENDED_PHP_VERSION = '7.0';
+    public const MINIMUM_MEMORY_REQUIREMENT = '256';
 
     /**
      * @param TranslatorInterface $translator
@@ -47,7 +48,12 @@ final class SettingsRequirements extends RequirementCollection
                 !$this->isOn('session.auto_start'),
                 false
             ))
-        ;
+            ->add(new Requirement(
+                $translator->trans('sylius.installer.settings.memory_limit.header', []),
+                $this->hasSufficientMemory(),
+                false,
+                $translator->trans('sylius.installer.settings.memory_limit.help', [])
+            ));
     }
 
     /**
@@ -60,5 +66,48 @@ final class SettingsRequirements extends RequirementCollection
         $value = ini_get($key);
 
         return !empty($value) && $value !== 'off';
+    }
+
+    /**
+     * @return bool
+     */
+    private function hasSufficientMemory(): bool
+    {
+        $currentLimit = ini_get('memory_limit');
+        if ($currentLimit == -1) {
+            return true;
+        }
+
+        $recommendedBytes = 1024 * 1024 * self::MINIMUM_MEMORY_REQUIREMENT;
+        if ($this->getMemoryInBytes($currentLimit) >= $recommendedBytes) {
+            return true;
+        }
+
+        return false;
+    }
+
+    /**
+     * Converts a memory_limit value to amount of bytes
+     *
+     * @see https://github.com/composer/composer/blob/master/bin/composer#L26
+     *
+     * @param $value
+     * @return int
+     */
+    private function getMemoryInBytes($value)
+    {
+        $unit = strtolower(substr($value, -1, 1));
+        $value = (int)$value;
+        switch ($unit) {
+            case 'g':
+                $value *= 1024;
+            // no break (cumulative multiplier)
+            case 'm':
+                $value *= 1024;
+            // no break (cumulative multiplier)
+            case 'k':
+                $value *= 1024;
+        }
+        return $value;
     }
 }
--- a/src/Sylius/Bundle/CoreBundle/Installer/Requirement/SettingsRequirements.php
+++ b/src/Sylius/Bundle/CoreBundle/Installer/Requirement/SettingsRequirements.php
@@ -53,7 +53,8 @@ final class SettingsRequirements extends RequirementCollection
                 $this->hasSufficientMemory(),
                 false,
                 $translator->trans('sylius.installer.settings.memory_limit.help', [])
-            ));
+            ))
+        ;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/CoreBundle/Resources/translations/messages.en.yml
```diff
@@ -202,6 +202,7 @@ sylius:
                 help: Timezone %timezone% is deprecated. Fix your php.ini file (list of deprecated timezones http://php.net/manual/en/timezones.others.php).
             version: PHP version
             version_recommended: Recommended PHP version
+            memory_limit: memory_limit
```
GSadee: You should add translation key only in this file with the proper message and revert adding this line in other `messages.*.yml` files

pamil:
I don't think we can estimate how much memory is really needed, further PHP versions optimises its memory usage so we might end up with installation failing because of this requirement only.

nstapelbroek:
@pamil good point, would you like me to change the requirement into a "warning"? I haven't tried this one out on multiple systems. But form here it looks like the default of 128m is just not enough for running the installation.

src/Sylius/Bundle/CoreBundle/Resources/translations/messages.en.yml
nstapelbroek: Alright! I'll fix this tonight

nstapelbroek:
Implemented PR feedback and squashed the commits. Changes:
- Instead of a requirement, the assertion is now a check and will not block the installation
- Removed duplicate translations
- Added a help message

pjedrzejewski:
@nstapelbroek Nice, looks like a good change to me! Could you have a look at the failing build, please? Thanks!

src/Sylius/Bundle/CoreBundle/Installer/Requirement/SettingsRequirements.php
```diff
@@ -47,7 +48,12 @@ public function __construct(TranslatorInterface $translator)
                 !$this->isOn('session.auto_start'),
                 false
             ))
-        ;
+            ->add(new Requirement(
+                $translator->trans('sylius.installer.settings.memory_limit.header', []),
+                $this->hasSufficientMemory(),
+                false,
+                $translator->trans('sylius.installer.settings.memory_limit.help', [])
+            ));
```
GSadee: `;` should be in the next line

src/Sylius/Bundle/CoreBundle/Installer/Requirement/SettingsRequirements.php
```diff
@@ -61,4 +67,47 @@ private function isOn(string $key): bool
...
+            case 'g':
+                $value *= 1024;
+            // no break (cumulative multiplier)
+            case 'm':
+                $value *= 1024;
+            // no break (cumulative multiplier)
+            case 'k':
+                $value *= 1024;
+        }
+        return $value;
```
GSadee: Missing blank line before `return`

src/Sylius/Bundle/CoreBundle/Installer/Requirement/SettingsRequirements.php
```diff
@@ -61,4 +67,47 @@ private function isOn(string $key): bool
...
+     *
+     * @see https://github.com/composer/composer/blob/master/bin/composer#L26
+     *
+     * @param $value
+     * @return int
+     */
+    private function getMemoryInBytes($value)
+    {
+        $unit = strtolower(substr($value, -1, 1));
+        $value = (int)$value;
```
GSadee: Missing space after `(int)`

src/Sylius/Bundle/CoreBundle/Installer/Requirement/SettingsRequirements.php
```diff
@@ -61,4 +67,47 @@ private function isOn(string $key): bool
...
+        return false;
+    }
+
+    /**
+     * Converts a memory_limit value to amount of bytes
+     *
+     * @see https://github.com/composer/composer/blob/master/bin/composer#L26
+     *
+     * @param $value
+     * @return int
```
GSadee: Missing blank line before `@return`

src/Sylius/Bundle/CoreBundle/Installer/Requirement/SettingsRequirements.php
```diff
@@ -61,4 +67,47 @@ private function isOn(string $key): bool
...
+     */
+    private function getMemoryInBytes($value)
```
GSadee: `private function getMemoryInBytes(string $value): int`?

pamil:
I agree it's a nice change, but still, we can't estimate how much memory the installation would take. I've just made a few installs and these are the results:

PHP version | Environment | Options | Peak memory usage
-------------|--------------|---------|----------------
PHP 7.1 | dev | sample data | 174 MiB
PHP 7.1 | dev | no sample data | 146 MiB

Even when using `dev` environment, every installation was below those required 256 MiB (and it can go even lower with OPCache enabled for CLI or next PHP versions).

stefandoorn:
@pamil How does this memory usage go when doing a full `composer update` on an installation?

pamil:
@stefandoorn this one relates to `sylius:install` command only, which happens after Composer installation / update. Last time I've checked, `composer update` took ~1.2 GiB on PHP 7.1.

pamil:
I've just done a `composer update` on `Sylius/Sylius` with PHP 7.1 and peak memory usage was 700 MiB.

nstapelbroek:
Thanks, all. If I understand correctly I only need to apply the requested changes of @GSadee ?
[Composer sets their own memory_limit](https://github.com/composer/composer/blob/eb5495b8e40811fc637e64bf0971a2e2349400ad/bin/composer#L46) if `ini_set` is available. Might this be a solution we could use here?

nstapelbroek:
Hiya :raising_hand_man: , any updates on this? Do I need to do something to get this merged?

pamil:
Hey! I'm sorry to keep it stale. As I've mentioned in a comment above, Sylius can be installed using only 146MiB memory, so 256MiB requirement would prevent it from a successful installation and that's not what we want.

I guess it's perfectly fine to try to change the memory limit to 256MiB if it's below that level, just as [Composer does](https://github.com/composer/composer/blob/master/bin/composer#L45-L47), but I'd rather not prevent end users from trying to install Sylius just because of a lower memory limit.

PR #8863, opened 2017-10-17
Add flash from event in ResourceController.php

atarko:
Add flash from event if it stopped instead of success message generation in createAction, updateAction, deleteAction and applyStateMachineTransitionAction of ResourceController.php

| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
@@ -286,7 +286,11 @@ class ResourceController extends Controller
                 return $this->viewHandler->handle($configuration, View::create($newResource, Response::HTTP_CREATED));
             }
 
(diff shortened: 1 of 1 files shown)


GSadee:
There is no need for that, because this behaviour is earlier in the code of these methods:
https://github.com/Sylius/Sylius/blob/701b0b9abe891c3d2cff1d46fc02e313a03074a3/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php#L268:L276

atarko:
@GSadee It wouldn't work because event becomes stopped only after dispatching it here:
```php
             $postEvent = $this->eventDispatcher->dispatchPostEvent(ResourceActions::UPDATE, $configuration, $resource);
```

GSadee:
@atarko my bad.

PR #8874, opened 2017-10-19
[Order][OrderItem] Immutable product/variant names

### LINKED ISSUE #8628: Product names are not stored in Orders/OrderItem so can change in the future

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #8628
| License         | MIT

It replaces a PR (#8667) created by @johnrisby using his squashed commits.

### DIFF
--- a/app/migrations/Version20170913125128.php
+++ b/app/migrations/Version20170913125128.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Sylius\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20170913125128 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_order_item ADD immutable_product_name VARCHAR(255), ADD immutable_product_code VARCHAR(255), ADD immutable_variant_name VARCHAR(255), ADD immutable_variant_code VARCHAR(255)');
+
+        $this->addSql('update sylius_order_item INNER JOIN sylius_order as o ON sylius_order_item.order_id = o.id INNER JOIN sylius_product_variant as pv1 ON sylius_order_item.variant_id = pv1.id INNER JOIN sylius_product as p1 ON pv1.product_id = p1.id  INNER JOIN sylius_product_translation as pt1 ON pt1.translatable_id = p1.id INNER JOIN sylius_order as o2 on pt1.locale = o2.locale_code set sylius_order_item.immutable_product_name =  pt1.name,  sylius_order_item.immutable_product_code = p1.code');
+
+        $this->addSql('update sylius_order_item INNER JOIN sylius_order as o ON sylius_order_item.order_id = o.id INNER JOIN sylius_product_variant as pv1 ON sylius_order_item.variant_id = pv1.id INNER JOIN sylius_product_variant_translation as pt1 ON pt1.translatable_id = pv1.id INNER JOIN sylius_order as o2 on pt1.locale = o2.locale_code set sylius_order_item.immutable_variant_name =  pt1.name,  sylius_order_item.immutable_variant_code = pv1.code');
+
+        $this->addSql('ALTER TABLE sylius_order_item CHANGE immutable_product_name immutable_product_name VARCHAR(255) DEFAULT NULL, CHANGE immutable_product_code immutable_product_code VARCHAR(255) DEFAULT NULL, CHANGE immutable_variant_name immutable_variant_name VARCHAR(255) DEFAULT NULL, CHANGE immutable_variant_code immutable_variant_code VARCHAR(255) DEFAULT NULL');
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function down(Schema $schema)
+    {
+        // this down() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_order_item DROP immutable_product_name, DROP immutable_product_code, DROP immutable_variant_name, DROP immutable_variant_code');
+    }
+}
--- a/app/migrations/Version20170913125128.php
+++ b/app/migrations/Version20170913125128.php
@@ -20,21 +20,23 @@ class Version20170913125128 extends AbstractMigration
 
         $this->addSql('ALTER TABLE sylius_order_item ADD product_name VARCHAR(255), ADD variant_name VARCHAR(255)');
 
-        $this->addSql('UPDATE sylius_order_item 
-                        INNER JOIN sylius_order AS o ON sylius_order_item.order_id = o.id 
-                        INNER JOIN sylius_product_variant AS pv ON sylius_order_item.variant_id = pv.id 
-                        INNER JOIN sylius_product AS p ON pv.product_id = p.id  
-                        INNER JOIN sylius_product_translation AS pt ON pt.translatable_id = p.id 
-                        INNER JOIN sylius_order AS ol ON pt.locale = ol.locale_code 
-                        SET sylius_order_item.product_name =  pt.name
+        $this->addSql('
+                        UPDATE sylius_order_item 
+                        INNER JOIN sylius_order ON sylius_order_item.order_id = sylius_order.id 
+                        INNER JOIN sylius_product_variant ON sylius_order_item.variant_id = sylius_product_variant.id 
+                        INNER JOIN sylius_product ON sylius_product_variant.product_id = sylius_product.id  
+                        INNER JOIN sylius_product_translation ON sylius_product_translation.translatable_id = sylius_product.id
+                        SET sylius_order_item.product_name = sylius_product_translation.name 
+                        WHERE sylius_product_translation.locale = sylius_order.locale_code
        
         ');
-        $this->addSql('UPDATE sylius_order_item 
-                        INNER JOIN sylius_order AS o ON sylius_order_item.order_id = o.id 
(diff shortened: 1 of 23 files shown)


stefandoorn:
It currently covers product + variant information. It would make sense to me also to write the underlying product option values with the variant. For my project they say way more than the variant name / code, as we have 1000s of variants per products and they just have something generated. The underlying naming says way more to the client and is also printed on invoices. Would be great to have that also added. 

The basics now already implemented are also a must, so it's great to see it moving on :)

johnrisby:
Thanks @GSadee!

GSadee:
@johnrisby no problem and of course I hope you don't mind that I continue your work 😃

johnrisby:
Not at all @GSadee, it's great. That's the whole point!

app/migrations/Version20170913125128.php
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_order_item ADD immutable_product_name VARCHAR(255), ADD immutable_product_code VARCHAR(255), ADD immutable_variant_name VARCHAR(255), ADD immutable_variant_code VARCHAR(255)');
+
+        $this->addSql('UPDATE sylius_order_item 
+                        INNER JOIN sylius_order AS o ON sylius_order_item.order_id = o.id
```
lchrusciel: It would much easier to follow the SQL logic if we would use some more appropriate names for aliases rather than abbreviate the table name

app/migrations/Version20170913125128.php
```diff
@@ -0,0 +1,53 @@
...
+                        INNER JOIN sylius_order AS ol ON pt.locale = ol.locale_code 
+                        SET sylius_order_item.immutable_product_name =  pt.name,  sylius_order_item.immutable_product_code = p.code
+       
+        ');
+        $this->addSql('UPDATE sylius_order_item 
+                        INNER JOIN sylius_order AS o ON sylius_order_item.order_id = o.id 
+                        INNER JOIN sylius_product_variant AS pv ON sylius_order_item.variant_id = pv.id 
+                        INNER JOIN sylius_product_variant_translation AS pt ON pt.translatable_id = pv.id 
+                        INNER JOIN sylius_order AS ol ON pt.locale = ol.locale_code 
+                        SET sylius_order_item.immutable_variant_name =  pt.name,  sylius_order_item.immutable_variant_code = pv.code
```
lchrusciel: Is it really worthy to add another SQL statement?

src/Sylius/Behat/Page/Admin/Order/ShowPage.php
```diff
@@ -149,15 +149,24 @@ public function countItems()
...
             $rows = $this->tableAccessor->getRowsWithFields(
-                $this->getElement('table'),
+                $table,
                 ['item' => $productName]
             );
 
-            return 1 === count($rows);
+            foreach ($rows as $row) {
+                $field = $this->tableAccessor->getFieldFromRow($table, $row, 'item');
+                $name = $field->find('css', '.sylius-product-name');
```
lchrusciel: Shouldn't we assert that the field would exist?

src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/Summary/_item.html.twig
```diff
@@ -6,7 +6,7 @@
 {% set taxAdjustment = constant('Sylius\\Component\\Core\\Model\\AdjustmentInterface::TAX_ADJUSTMENT') %}
 
 {% set variant = item.variant %}
-{% set product = variant.product %}
+{% set product = item.variant.product %}
```
lchrusciel: What's the difference?

src/Sylius/Bundle/AdminBundle/Resources/views/Product/_info.html.twig
```diff
@@ -1,11 +1,9 @@
-{% set product = variant.product %}
-
 <div class="ui header">
     {% include '@SyliusAdmin/Product/_mainImage.html.twig' with {'product': product, 'filter': 'sylius_admin_product_tiny_thumbnail'} %}
     <div class="content">
-        <a href="{{ path('sylius_admin_product_update', {'id': product.id}) }}" class="sylius-product-name" title="{{ product.name }}">{{ product.name }}</a>
-        <span class="sub header sylius-product-variant-code" title="{{ variant.code }}">
-            {{ variant.code }}
+        <div class="sylius-product-name" title="{{ item.immutableProductName }}">{{ item.immutableProductName }}</div>
```
lchrusciel: Do we need to add `immutable` prefix? It is set to order item, so it shouldn't change anyway. 

Just thinking

src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/OrderItem.orm.xml
```diff
@@ -17,6 +17,11 @@
                                       http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
 
     <mapped-superclass name="Sylius\Component\Core\Model\OrderItem" table="sylius_order_item">
+        <field name="immutableProductName" column="immutable_product_name" type="string" nullable="true"/>
```
lchrusciel: I'm not sure if the product name on order item should be nullable. It is pretty important data

src/Sylius/Component/Core/Model/OrderItem.php
```diff
@@ -37,17 +87,65 @@ public function getProduct(): ?ProductInterface
...
+     */
+    public function getImmutableProductCode(): ?string
+    {
+        return $this->immutableProductCode ?: $this->variant->getProduct()->getCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setImmutableProductCode(?string $immutableProductCode): void
```
lchrusciel: Maybe we could skip this new setter?

src/Sylius/Behat/Page/Admin/Order/ShowPage.php
GSadee: Not in this case, it would change the behaviour of this method.

src/Sylius/Component/Core/Model/OrderItem.php
```diff
@@ -27,8 +27,46 @@ class OrderItem extends BaseOrderItem implements OrderItemInterface
...
      */
+    public function getVariant(): ?ProductVariantInterface
+    {
+        return $this->variant;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setVariant(?ProductVariantInterface $variant): void
```
pamil: Hmm, what if a product name changes after the product was added to the cart but before the checkout was completed?

app/migrations/Version20170913125128.php
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_order_item ADD product_name VARCHAR(255), ADD variant_name VARCHAR(255)');
+
+        $this->addSql('UPDATE sylius_order_item 
+                        INNER JOIN sylius_order AS o ON sylius_order_item.order_id = o.id
```
lchrusciel: Can we not abbreviate rest of table aliases. It is not easy to understand what happened here

src/Sylius/Component/Core/Model/OrderItem.php
GSadee: Good point 👍  What is your suggestion? Should I set immutable names on checkout `complete` transition?

GSadee: I've created `OrderImmutableNamesSetter` which is called on checkout `complete` transition.

src/Sylius/Component/Core/Order/OrderImmutableNamesSetterInterface.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Sylius\Component\Core\Order;
+
+use Sylius\Component\Core\Model\OrderInterface;
+
+interface OrderImmutableNamesSetterInterface
+{
+    /**
+     * @param OrderInterface $order
+     */
+    public function setNames(OrderInterface $order): void;
```
pamil: This can be changed later (up to v1.1.0 release), so it's not a big deal, only this interface name looks weird.

`OrderImmutableNamesSetterInterface::setNames` -> `OrderItemProductNamesSetterInterface::__invoke` / `OrderFinalizer::__invoke` / something else?

src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine/sylius_order_checkout.yml
```diff
@@ -63,3 +63,7 @@ winzou_state_machine:
                     on: ["complete"]
                     do: ["@sylius.state_resolver.order_shipping", "resolve"]
                     args: ["object"]
+                sylius_set_order_immutable_names:
```
pjedrzejewski: We should move it to the main order state machine. This is not really specific to checkout, it should happen whenver we create a new order (cart -> new transition).

pjedrzejewski:
Thank you John & Grzesiu, nice work guys! /cc @johnrisby

johnrisby:
Great stuff. Sorry I've not contributed more to this recently but I've got a lot on. Thanks for picking up the baton Grzesiu!

Issue #8224, opened 2017-06-30
Make core model interfaces return docblocks more specific


lasotaartur:
Hi,  I have problem everywhere with sylius interface i.e.

```
$product = $productVariant->getProduct();
```
will return object with interface `Sylius\Component\Product\Model\ProductInterface`

then when I write this
```
$taxon = $product->getMainTaxon();
```
PHPstorm will show that the not exists in interface `Sylius\Component\Product\Model\ProductInterface` but exists in `\Sylius\Component\Core\Model\ProductInterface` (edited)

And problem exists for others components also

lchrusciel:
Yes, we are aware of this. That way, there is a lot of variable type hints around Sylius, But this probably won't be resolved, as it is a trade-off, that we are probably willing to pay. Unless somebody has an idea how to make it better. Not sure, if there is any point in keeping it open? /cc @pamil @pjedrzejewski

pamil:
Actually, we might end up overriding return type docblocks in Core entities so that it won't happen if traversing from another Core entity.

pjedrzejewski:
This is one of the most annoying DX issues we have IMO, let's try to fix it for 1.0.0.

pamil:
Tried to implement it before beta.3, but return types will make it much more easier.

PR #8876, opened 2017-10-19
Narrow typehints for models - part 1

pamil:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partly #8224
| License         | MIT

Thanks @ondrejmirtes for initiating this work :)

### DIFF
--- a/src/Sylius/Component/Core/Model/Product.php
+++ b/src/Sylius/Component/Core/Model/Product.php
@@ -18,6 +18,7 @@ use Doctrine\Common\Collections\Collection;
 use Sylius\Component\Channel\Model\ChannelInterface as BaseChannelInterface;
 use Sylius\Component\Product\Model\Product as BaseProduct;
 use Sylius\Component\Product\Model\ProductTranslationInterface as BaseProductTranslationInterface;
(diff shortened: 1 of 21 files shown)


src/Sylius/Component/Payment/Model/PaymentMethod.php
```diff
@@ -25,6 +26,7 @@ class PaymentMethod implements PaymentMethodInterface
     use TimestampableTrait, ToggleableTrait;
     use TranslatableTrait {
         __construct as initializeTranslationsCollection;
```
pamil: Side note: should be "__construct as **private** ...".

src/Sylius/Component/Core/Model/Product.php
```diff
@@ -353,6 +354,16 @@ public static function getVariantSelectionMethodLabels(): array
     }
 
     /**
+     * @param null|string $locale
```
GSadee: `string|null` ?

Issue #8423, opened 2017-08-22
[RFC] Remove all author blocks


Zales0123:
Hello folks!

According to discussion from https://github.com/Sylius/Sylius/pull/8310 we should decide, whether we want to remove all `@author` blocks from the code or we want to keep it as it is.

There are some good reasons to do that:
- it would prevent such a ridiculous PR as mentioned
- we can easily know the author of the code from git
- there are probably some more accurate places to keep the contributors list (on the website, on the composer.json etc.).

On the other hand, some people (especially those not active in **Sylius** development right now) can feel bad when their author blocks would be removed without asking them for permission to do that.

Personally, I'm ok with removing those blocks, but I move the discussion here, to make it more visible for anyone interested in this case 🐃

PR #8310, opened 2017-07-24
[Critical] Change Zales0123 mail

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | |
| License         | MIT |

To allow world to communicate with me easier 🌎  📫

PR #8882, opened 2017-10-20
Remove all author docblocks

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | https://github.com/Sylius/Sylius/pull/8310, https://github.com/Sylius/Sylius/issues/8423
| License         | MIT

The topic was discussed more than once. These author docblocks are not only redundant (such a data can be easily fetched from git history) but sometimes even misleading (some of these emails are outdated, for example mine 😄 ). I know that it can be sad for somebody, not being mentioned explicitly in the code, but I strongly believe there is not point in keeping this comments in our code anymore 🐃


pjedrzejewski:
👍 👍 👍

PR #8888, opened 2017-10-23
[Core][Payum] Set default gateway name in gateway config

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Without this fix, when creating payment method programatically, it was required to fetch gateway config from payment method, and set it's gateway name manually to prevent doctrine error (gateway name cannot be null).

```
$paymentMethod = $this->paymentMethodFactory->createWithGateway('offline');
$gatewayConfig = $paymentMethod->getGatewayConfig();
$gatewayConfig->setGatewayName('Offline');
$paymentMethod->setGatewayConfig($gatewayConfig);
```

The other idea is to modify `createWithGateway` method to get `?string $gatewayName` as second argument, and `ucfirst($gatewayFactory)` as default value, wdyt?

### DIFF
--- a/src/Sylius/Component/Core/Factory/PaymentMethodFactory.php
+++ b/src/Sylius/Component/Core/Factory/PaymentMethodFactory.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sylius\Component\Core\Factory;
 
+use Payum\Core\Model\GatewayConfigInterface;
 use Sylius\Component\Core\Model\PaymentMethodInterface;
 use Sylius\Component\Resource\Factory\FactoryInterface;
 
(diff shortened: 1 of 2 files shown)


src/Sylius/Component/Core/Factory/PaymentMethodFactory.php
```diff
@@ -54,8 +55,10 @@ public function createNew(): PaymentMethodInterface
      */
     public function createWithGateway(string $gatewayFactory): PaymentMethodInterface
     {
+        /** @var GatewayConfigInterface $gatewayConfig */
         $gatewayConfig = $this->gatewayConfigFactory->createNew();
         $gatewayConfig->setFactoryName($gatewayFactory);
+        $gatewayConfig->setGatewayName(ucfirst($gatewayFactory));
```
GSadee: If I think correctly, the `gatewayName` have to be unique in Payum, so that is probably not the best solution /cc @Arminek 

Now, the gateway name is set as `code` of `PaymentMethod` and this change could change current behaviour.
https://github.com/Sylius/Sylius/blob/49a1ac51277e5ba98b45db7e2d4f2bfad1155e71/src/Sylius/Bundle/CoreBundle/Form/Extension/PaymentMethodTypeExtension.php#L55-L58

pjedrzejewski:
@Zales0123 Please use the correct branch. :)

PR #8903, opened 2017-10-25
Plugin naming convention RFC implementation (closes #8848)

### LINKED ISSUE #8848: RFC: Sylius plugin naming convention - allow Sylius in plugin name
| Q                | A
| ---------------- | -----
| RFC?             | yes
| Sylius version   | ALL

Currently, the Sylius plugin manual dictates NOT using `Sylius` in the name: http://docs.sylius.org/en/latest/plugins/creating-plugin.html (at the bottom). It doesn't clearly state how to name the repository of a plugin, but most plugins until now also exclude `Sylius` from the repository name. The composer.json states type `sylius-plugin`, but that's only discoverable through Packagist search: https://packagist.org/search/?type=sylius-plugin.

On one of my plugins I got an issue filed to rename the plugin to something more recognisable, basically adding Sylius to the naming - which I can only agree with: https://github.com/stefandoorn/google-tag-manager-enhanced-ecommerce-plugin/issues/13.

I see multiple benefits from renaming plugins to include the `Sylius` name:

- Direct distinction for which platform this plugin is suitable. 
- Easier to spot plugins for potential new users (Google search for plugins will improve e.g.). Plugin adoption can rise, ecosystem improves.
- More marketing for name `Sylius`. Seeing the name more around make people wonder what it is and does. 
- Via @pamil (see comment below): The additional benefit would be to have `sylius_` prefix in the configuration files. SitemapPlugin only would result in sitemap key, whereas `SyliusSitemapPlugin` would result in `sylius_sitemap` which looks less confusing for me (and reduces potential conflicts).

Proposed changes:

- Repository name should start with `sylius-` and end with `plugin`. E.g.: `sylius-sitemap-plugin`.
- Namespace can stay as is for me, e.g.: `SitemapPlugin`, because as soon as a Sylius user is using it it's rather clear. 

Alternative for namespacing change:

- Alternative for namespacing might be: `SyliusSitemapPlugin` to make it even more clear inside the code that you are using a Sylius specific thing - instead of something Symfony wide of even PHP wide. 
- Another alternative is to require to add the author also to the namespace, to make even more clear it's not maintained by Sylius core devs when browsing through code, e.g.: `StefanDoorn\SyliusSitemapPlugin`.

stefandoorn:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes (sort of.. but no in docs)
| Deprecations?   | no
| Related tickets | fixes #8848
| License         | MIT

After discussion in #8848, hereby the PR. Difference from the discussion is that I've chosen to name the repository after the class bundle name, to have it all consistent and in the same style. 

Before merging this, there should also come a PR to the PluginSkeleton to rename things I think.

### DIFF
--- a/docs/plugins/creating-plugin.rst
+++ b/docs/plugins/creating-plugin.rst
@@ -11,11 +11,11 @@ which has built-in infrastructure for designing and testing using `Behat`_.
 
 .. code-block:: bash
 
-    $ composer create-project sylius/plugin-skeleton MyPlugin
+    $ composer create-project sylius/plugin-skeleton SyliusMyFirstPlugin
 
 .. note::
 
-    The plugin can be created anywhere, not only inside Sylius application, because it already has the test environment inside.
+    The plugin can be created anywhere, not only inside a Sylius application, because it already has the test environment inside.
 
 2. Get familiar with basic plugin design.
 -----------------------------------------
@@ -31,9 +31,13 @@ and the testing infrastructure in ``tests``.
 3. Remove boilerplate files and rename your bundle.
 ---------------------------------------------------
 
-In most cases you don't want your Sylius plugin to greet customer like it is now, so feel free to remove unnecessary
-controllers, assets and features. You will also want to change the plugin's namespace from ``Acme\ExamplePlugin`` to a
+In most cases you don't want your Sylius plugin to greet the customer like it is now, so feel free to remove unnecessary
+controllers, assets and features. You will also want to change the plugin's namespace from ``Acme\SyliusExamplePlugin`` to a
 more meaningful one. Keep in mind that these changes also need to be done in ``tests/Application`` and ``composer.json``.
+
+.. tip::
+
+    Refer to chapter 5 for the naming conventions to be used.
 
 4. Implement your awesome features.
 -----------------------------------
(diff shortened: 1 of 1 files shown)


docs/plugins/creating-plugin.rst
```diff
@@ -64,11 +64,14 @@ to ensure your plugin's extraordinary quality.
 
 Besides the way you are creating plugins (based on our skeleton or on your own), there are a few naming conventions that should be followed:
 
- * Bundle class must have a ``Plugin`` suffix instead of ``Bundle`` in its name (e.g. InvoicePlugin).
+ * Bundle class must have a ``Sylius`` prefix, and a ``Plugin`` suffix (instead of ``Bundle``) in its name (e.g. ``SyliusInvoicePlugin``).
```
pamil: I'd make it *vendor name* + `Sylius` + *actual name* + `Plugin`. The reasoning behind it is to distinct official Sylius plugins from the community made ones. `SyliusInvoicePlugin` uses `sylius_invoice` as config root name, whereas `AcmeSyliusInvoicePlugin` uses `acme_sylius_invoice.`

docs/plugins/creating-plugin.rst
```diff
@@ -64,11 +64,14 @@ to ensure your plugin's extraordinary quality.
...
  * Bundle class must use the ``Sylius\Bundle\CoreBundle\Application\SyliusPluginTrait``.
- * The name of the extension in DependencyInjection folder must follow the regular Symfony rules (e.g. InvoiceExtension).
- * The plugin shouldn't have Sylius prefix in its name. ``Plugin`` suffix in terms of bundles is unique for Sylius at the moment.
+ * The name of the extension in the DependencyInjection folder must follow the regular Symfony rules (e.g. ``SyliusInvoiceExtension``).
```
pamil: Then the extension name changes accordingly.

docs/plugins/creating-plugin.rst
stefandoorn: True.. Downside is that e.g. in my case using my own name we get to something like this: `StefanDoornSyliusInvoicePlugin`.. sounds silly :-) The vendor is already in the namespace and the repository is in another vendor, isn't that enough to make people (developers only) understand what the source of it is?

pamil: These are only two files, bundle one and extension one, it's most likely to be an one time thing. The issue here is to have distinct config keys.

For Composer package name and Github repos it's already solved by prefixing with `vendor-name/`.

stefandoorn: Ah that way. Ok, I'll check it. Alternative could be to adjust the trait to adjust the config key I think based on PSR4 namespace.

CoderMaggie: Is there really a need to distinct official Sylius plugins from the community made plugins? There should be one rule for them all IMO. 
And `Sylius` + `name` + `Plugin` seems simple and enough.

pamil: Using vendor names in package names / configuration keys is a common practice, I can't think of Symfony bundle which does not follow this convention.

Let's just take our most common plugin, CMS one, there's one from Lakion and one from BitBag. If we stay with `SyliusCmsPlugin`, the configuration key will always be `sylius_cms`, which might cause confusion. And then FriendsOfSylius or Sylius itself wants to unify them... creating another one with the same configuration key.

stefandoorn: I see the point. Is it possible to do this by adjusting the trait? What I'm trying is to get the name of the repo and the class name aligned, but having diversified configs keys :)

pamil: As for the currently existing trait, changing its behaviour will be a BC break and having two of them might be misleading.

stefandoorn: Makes sense, not an option. Could you (@pamil, @pjedrzejewski, @CoderMaggie) let me know what the naming convention regarding this will be? Then I will implement accordingly.

pamil: I guess we can end up with:

Composer package name: `vendor-name/sylius-something-plugin`
Bundle / extension classname: `VendorNameSyliusSomethingPlugin` / `VendorNameSyliusSomethingExtension` - namespace does not matter, probably in `VendorName\SyliusSomethingPlugin`

stefandoorn:
Changed a lot and added examples. Let me know what you think.

docs/plugins/creating-plugin.rst
```diff
@@ -64,11 +68,82 @@ to ensure your plugin's extraordinary quality.
...
 
- * Bundle class must have a ``Plugin`` suffix instead of ``Bundle`` in its name (e.g. InvoicePlugin).
- * Bundle class must use the ``Sylius\Bundle\CoreBundle\Application\SyliusPluginTrait``.
- * The name of the extension in DependencyInjection folder must follow the regular Symfony rules (e.g. InvoiceExtension).
- * The plugin shouldn't have Sylius prefix in its name. ``Plugin`` suffix in terms of bundles is unique for Sylius at the moment.
+ * Repository name should use dashes as separator, must have a ``sylius`` prefix and a ``plugin`` suffix, e.g.: ``sylius-invoice-plugin``.
+ * Bundle class name should start with vendor name, followed by ``Sylius`` and suffixed by ``Plugin`` (instead of ``Bundle``), e.g.: ``VendorNameSyliusInvoicePlugin``.
+ * Bundle extension should be named similar, but suffixed by the Symfony standard ``Extension``, e.g.: ``VendorNameSyliusInvoiceExtension``.
+ * Bundle class must use the ``Sylius\Bundle\CoreBundle\Application\SyliusPluginTrait`` trait.
+ * Namespace should follow _`PSR-4 <http://www.php-fig.org/psr/psr-4/>`. It should start with vendor name, followed by ``Sylius`` and suffixed by ``Plugin`` (e.g. ``VendorName\SyliusInvoicePlugin``)
```
pamil: It should start with vendor name, followed by ``Sylius`` and suffixed by ``Plugin`` (e.g. ``VendorName\SyliusInvoicePlugin``) -> Top-level namespace should be the vendor name and the second-level one should be prefixed with...?

docs/plugins/creating-plugin.rst
```diff
@@ -64,11 +68,82 @@ to ensure your plugin's extraordinary quality.
...
+    }
+
+**4.** In ``composer.json``, define the correct namespacing for the PSR-4 autoloader:
+
+.. code-block:: json
 
-The following rules are applied to all bundles which will provide an integration with the whole Sylius platform
-(``sylius/sylius`` or ``sylius/core-bundle`` in vendors). Reusable components for the whole Symfony community, which will be based
-just on some Sylius bundles should follow regular Symfony conventions.
+    "autoload": {
```
pamil: The build is failing because the given JSON snippet is invalid, let's change it to:

```json
{
  "autoload": {
    "psr-4": {
      "VendorName\\SyliusInvoicePlugin\\": "src/",
      "Tests\\VendorName\\SyliusInvoicePlugin\\": "tests/"
    }
  }
}

**Comment by stefandoorn** [comment]
@pamil I think we are good to go now.

**Comment by pamil** [comment]
This one should be merged together with Sylius/PluginSkeleton#56.

**Comment by pjedrzejewski** [comment]
Thank you very much Stefan! 👍  Great work.

File: docs/plugins/creating-plugin.rst [hunk 5]
```diff
@@ -11,11 +11,11 @@ which has built-in infrastructure for designing and testing using `Behat`_.
 
 .. code-block:: bash
 
-    $ composer create-project sylius/plugin-skeleton MyPlugin
+    $ composer create-project sylius/plugin-skeleton SyliusMyFirstPlugin
```
lsmith77: did you intentional leave out the vendor name here?

lsmith77:
I noticed that the conventions do not see to cover DI service as well as command names. Should this maybe be covered as well?

PR #8915, opened 2017-10-30
[Travis] Update Yarn to stable 1.2.1, use Node v6

venyii:
| Q               | A
| --------------- | -----
| Branch?         | 1.0 
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

This also uses LTS Node v6 since v7 is not officially supported:
```
yarn install v1.2.1
warning You are using Node "7.5.0" which is not supported and may encounter bugs or unexpected behavior. Yarn supports the following semver range: "^4.8.0 || ^5.7.0 || ^6.2.2 || ^8.0.0"
```

To use Node v8, we'd need either Trusty (which causes some problems as you know) or upgrade gcc ourselves, but I'd just wait and see if Trusty works in the near future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
             php: 7.1
             env:
                 - SYLIUS_SUITE="application"
-                - TRAVIS_NODE_VERSION="7.5"
+                - TRAVIS_NODE_VERSION="6.11"
             services:
                 - memcached
         -
@@ -22,7 +22,7 @@ matrix:
             php: 7.2
             env:
                 - SYLIUS_SUITE="application"
-                - TRAVIS_NODE_VERSION="7.5"
+                - TRAVIS_NODE_VERSION="6.11"
             services:
                 - memcached
         -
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -14,15 +14,15 @@ matrix:
             php: 7.1
             env:
                 - SYLIUS_SUITE="application"
-                - TRAVIS_NODE_VERSION="7.5"
+                - TRAVIS_NODE_VERSION="6.11"
```
pamil: Let's use `8.9` version instead, it's the latest supported one.

etc/travis/suites/application/before_install.sh
```diff
@@ -15,4 +15,4 @@ run_command "rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/
 run_command "sudo apt-key adv --fetch-keys http://dl.yarnpkg.com/debian/pubkey.gpg"
 run_command "echo \"deb http://dl.yarnpkg.com/debian/ stable main\" | sudo tee /etc/apt/sources.list.d/yarn.list"
 run_command "sudo apt-get update -qq"
-run_command "sudo apt-get install -y -qq yarn=0.21.3-1"
+run_command "sudo apt-get install -y -qq yarn=1.2.1-1"
```
pamil: Is `1.3.2` available?

.travis.yml
pamil: Aaah okay, missed the description.

pamil:
Thanks @venyii! 🎉

etc/travis/suites/application/before_install.sh
venyii: Hah, of course they release 1.3 two days after I open this PR 😄  I'll save it for another time if you don't mind, at least we're stable now 🤞

pamil: Yeah, we can wait for Trusty setup to make both Node 8.x and Yarn 1.3.x

PR #8925, opened 2017-11-01
Do not instantiate abstract class

rvanlaak:
... as Twig_Filter is deprecated in `1.x` and `Twig_Filter` class was removed in Twig `2.0`.

We get this error on Twig 1.35.0 , so as Sylius is on Twig 2.0 and that deprecation was removed, I'm not sure how it's possible that this didn't crash tests before. See https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Filter.php

This means that all Sylius' users that use `Twig <2.x` can't use Sylius, so it's safest to use `SimpleFilter` for now. Also see https://github.com/twigphp/Twig/blob/2.x/lib/Twig/SimpleFilter.php

| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/MoneyBundle/Twig/FormatMoneyExtension.php
+++ b/src/Sylius/Bundle/MoneyBundle/Twig/FormatMoneyExtension.php
@@ -36,7 +36,7 @@ final class FormatMoneyExtension extends \Twig_Extension
     public function getFilters(): array
     {
         return [
-            new \Twig_Filter('sylius_format_money', [$this->helper, 'formatAmount']),
+            new \Twig_SimpleFilter('sylius_format_money', [$this->helper, 'formatAmount']),
         ];
     }
 }


lchrusciel:
But the minimal twig version [is 2.0](https://github.com/Sylius/Sylius/blob/1.0/composer.json#L65), so the code is correct as `Twig_Filter` exists in Twig 2.0 - https://github.com/twigphp/Twig/blob/2.x/lib/Twig/Filter.php

rvanlaak:
Yes, you're absolutely right! I now see that in 2.0 Twig just added back `Twig_Filter`, which has the same name as the pre-2.0 version, so that's quite unfortunate of them as for `1.x` users the abstract `Twig_Filter` class now is conflicting.

By using `Twig_SimpleFilter` everywhere you still are BC compatible with `1.x` for the ones that use only some of the Sylius bundles in combination with Twig `1.x`.

Until Twig `3.0` is released - and supported by Sylius - it would be great to use the BC classes 👍 I can understand that Sylius is going Twig `2.x` only though.

lchrusciel:
I guess the main problem was with testing it. There was a pretty long topic about it #7629 a few months ago. Anyway, it is worth considering /cc @pamil @pjedrzejewski

rvanlaak:
From Sylius standard perspective I absolutely understand choosing to only support `2.x`, but as we don't use standard but just a couple of bundles that's a bit different.

As example, the `sylius/promotion-bundle` does not require Twig `2.x` itself, as that's required via core or standard. As result every Symfony application can require that bundle - which is great - but this did fail for us as we were still at Twig `1.x`.

Thanks for linking the discussion on #7629. That mainly covers details on dropping `1.x` support and doing necessary upgrades for core, which I fully agree with. The main food for thought I want to mention here is choosing one of the following directions:

1. If bundles drop support for Twig `1.x`, that should be defined via their own `composer.json` files as well
2. If not, then BC classes provided by Twig - like `Twig_SimpleFilter` and `Twig_SimpleFunction` - should be used as much as possible. This is different from providing actual support for it in `core`, but just a matter of being BC friendly

For now, feel free to close this PR when you won't merge it, as we upgraded to Twig 2 in the meanwhile.

pamil:
If we officially support Twig 1.x, we should have some tests for both Twig 1.x and 2.x, which is a tough one.

I'd say we should stay with supporting Twig 2.x only by making it conflict with Twig 1.x (so that Twig isn't really required by bundles, but when it is used, it can't be 1.x version).

rvanlaak:
@pamil I'm absolutely convinced you should *not* support Twig `1.x`.

What I'm trying to mention here is that if Sylius would use the Twig BC classes - the ones that are provided by 2.x - that would mean that Sylius bundle users (the ones without `core`) will not be forced to upgrade to Twig `2.x`.

> so that Twig isn't really required by bundles

Actually the bundles are the ones that come with the Twig extensions, and that's also the way it should be I guess. As example, `sylius/money-bundle/Resources/config/services.xml` does define a Twig extension.

pamil:
Seems like the activity from the other people in this issue says that there is no need to support Twig 1.x or the upgrade to Twig 2.x is simpler, therefore I'm closing this issue.

rvanlaak:
👍 Eventually we would have been absolutely fine with a "no we stick to the plan and don't BC support Twig 1.x" at that same day or week, as we now were sort of forced to upgrade to Twig 2.x as forking Sylius isn't an option due to it's subtree branching model. This made our release back then quite interesting ;-)

Please also keep in mind, all other bundles don't have Twig 2 as dependency, so for all these bundles this problem could still happen again to someone that wants to use as single bundle. Would be best to add Twig 2 as dependency for all bundles that introduce Filters / Functions so it's clear that Twig 2 is the version that's supported.

pamil:
I've opened #9136 with PoC.

rvanlaak:
Yes, cool! Using `conflict` indeed is even better 👍

PR #8939, opened 2017-11-07
Revert "[Admin] Mass deletion on admin grid"

pamil:
Reverts #8491 which introduces a BC break by adding `GridRendererInterface::renderBulkAction` method, we need to figure an other way for that.


pamil:
Closing in favour of #8940.

Issue #8852, opened 2017-10-16
[ProductAttribute] Creation "select" type attribute from the Admin UI


rudestan:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.0.1

1) In the Admin UI during the creation of "Select" type Attribute there is no possibility to specify the name of the items in the choice array. As a result automatically generated choice key names will be given in format "9c75a866-b01d-11e7-a44b-791088f6a089". It might be good to provide the ability to give custom names per choice, for example if we would like to use the choice names to create a search request.

The auto-inserted demo fixtures have proper human-readable choice item keys.

From fixtures:

`a:4:{s:8:"multiple";b:0;s:7:"choices";a:4:{s:19:"invisible_porcelain";s:21:"Invisible porcelain 1";s:11:"banana_skin";s:11:"Banana skin";s:9:"porcelain";s:9:"Porcelain";s:9:"centipede";s:9:"Centipede";}s:3:"min";N;s:3:"max";N;}`

Created from the Admin UI:

`a:4:{s:7:"choices";a:1:{s:36:"6c84a1f6-b01e-11e7-a64a-5db6175fa47a";s:13:"Intel Core i3";}s:8:"multiple";b:0;s:3:"min";N;s:3:"max";N;}`

2) Small suggestion:

It is a bit confusing that the `storage_type` of "select" attribute is "json" however, in fact, it is a PHP serialized array. I know this is how the Doctrine handles the array type fields. However, would not it make more sense either to rename the `storage_type` from json to something like `serialized_collection` (or something similar that will corresponds to the real type) or change the type of `configuration` field from `array` to `json` so the Doctrine will store the real json?

pjedrzejewski:
We switched to UUID to avoid bugs with renaming the values, etc. Also, you are right about the column type... it should be JSON. We need a PR with a migration for that, would you find some time to contribute this? Thanks for reporting anyway!

rudestan:
@pjedrzejewski Regarding the JSON type, sure i would love to help and contribute and will do it in the nearest future. 
[...]

PR #8946, opened 2017-11-08
Fixed attribute configuration field type to json_array

rudestan:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | none
| License         | MIT

Small fix of the `sylius_product_attribute` table: changed `configuration` Doctrine column type to store `json_array` instead of php serialized array. Issue: https://github.com/Sylius/Sylius/issues/8852


rudestan:
Ok, changed configuration field type of the ShippingMethod as well.

pamil:
Thanks for your effort, but I guess we misinformed you in https://github.com/Sylius/Sylius/issues/8852#issuecomment-338885435 because that column type definition is correct (as it's related to attribute instead of attribute value). The first part of that issue still remains relevant.

PR #8987, opened 2017-11-26
Add PHP 7.1 typehints

TomasVotruba:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | depends
| Deprecations?   | no
| License         | MIT

Since this project already uses PHP 7.1 as minimum, why not use its features?

This would lead to more consistent and tested code, as annotation are sometimes flawed.

I know this might be BC break for external API classes. I'd send this PR earlier, before 1 became stable, if I'd known I haven't send it yet. Sorry for being late. **Those can be excluded though.**

This PR can be resend at any time in matter of minutes, so feel free to consider it flexible.

----

I can look on failed tests after initial feedback on this.


TomasVotruba:
> The only way to make it is to apply a single fixer at a time to given directories (eg. we can't add or modify typehints anywhere in the src/Bundle or src/Component)

Thanks for feedback, I understand this. How could I start for example?

TomasVotruba:
Not relevant anymore. I'm closing to clean my opened PRs.

Feel free to cherry pick anything you need.

PR #8996, opened 2017-11-30
Setup "Stale" probot

pamil:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

This one will help to focus on the most important issues and pull requests.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,13 @@
+daysUntilStale: 90
+daysUntilClose: 7
+exemptLabels:
+  - Bug
+  - Critical
+  - RFC
+staleLabel: Stale
+markComment: >
+  This issue has been automatically marked as stale because it has not had any
+  recent activity. It will be closed in a week if no further activity occurs.
+  Thank you for your contributions.
+unmarkComment: false
+closeComment: false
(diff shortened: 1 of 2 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,12 @@
+daysUntilStale: 90
+daysUntilClose: 7
+exemptLabels:
+  - Critical
+  - RFC
```
lchrusciel: Shouldn't we also leave the `Feature` label?

pamil: This probot works in two phases:
 - when an issue / PR is not updated for 90 days, it posts a comment and marks it as stale
 - when a stale issue / PR is not updated for another 7 days, it is closed
 - when a stale issue / PR is updated within 7 days, it removes stale label and the whole process starts again

So the reasoning behind it was to close those issues / PRs that people are not longer interested in. If there's still an interest, only four actions in a year would prevent it from closing :)

CoderMaggie: :+1:

pamil:
For anyone disturbed by tons of new notifications, make use of this JS script which removes the notifications when last reply is from Stalebot:

```
[...document.querySelectorAll('.list-group-item.js-notification.js-navigation-item.unread')].filter(_ => _.innerHTML.includes('Sylius/Sylius')).filter(_ => [..._.querySelectorAll('.avatar')].shift().href.match(/\/stale\[bot\]$/)).forEach(_ => _.querySelector('.delete-note').click())
```

PR #9008, opened 2017-12-05
[Documentation][ThemeBundle] Add warning note about theme performance issue

RadnoK:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/docs/components_and_bundles/bundles/SyliusThemeBundle/your_first_theme.rst
+++ b/docs/components_and_bundles/bundles/SyliusThemeBundle/your_first_theme.rst
@@ -50,6 +50,11 @@ Same comes with the bundle resources, eg. for ``FOSUserBundle`` the paths should
     └── views
         └── template.html.twig
 
+.. danger::
+
+   Due to performance issues, we recommend to have as little files in theme directories as it is possible. If you have i.e. ``node_modules`` in your theme, move it to your application bundle's resources or to the project root directory.
+
+
 Enabling themes
 ---------------
 


stefandoorn:
What kind of performance issues pop up?

RadnoK:
@stefandoorn As I found out, ThemeBundle iterates over every file in the active theme. If you have such a large directories like `node_modules`, it can be a performance killer for the PHP. 

Profiler results:
themebundle

stefandoorn:
Hm, that's quite interesting. Maybe better to get that fixed if it's not needed :-D

tadcka:
We use theme structure:
```
assets
templates
package.json
```
config.yml

```
sylius_theme:
    sources:
        filesystem:
            enabled: true
            filename: composer.json
            directories:
                - "%kernel.root_dir%/../themes/blabla/templates"
```

docs/components_and_bundles/bundles/SyliusThemeBundle/your_first_theme.rst
```diff
@@ -50,6 +50,11 @@ Same comes with the bundle resources, eg. for ``FOSUserBundle`` the paths should
     └── views
         └── template.html.twig
 
+.. danger::
+
+   Due to performance issues, we recommend to have as little files in theme directories as it is possible. If you have i.e. ``node_modules`` in your theme, move it to your application bundle's resources or to the project root directory.
```
CoderMaggie: can we break this line in two?

docs/components_and_bundles/bundles/SyliusThemeBundle/your_first_theme.rst
```diff
@@ -50,6 +50,11 @@ Same comes with the bundle resources, eg. for ``FOSUserBundle`` the paths should
...
+
```
CoderMaggie: redundant line

psren:
Is there a ticket for that. Whats the reason for iterating over every file?
Can this be fixed?
Or maybe something similar to .gitignore

pamil:
I think we can solve this one by actually removing this bottleneck completely. Looks like the current behaviour is to recursively look for `composer.json` in `app/themes`.

Some of the proposed solutions:
 - using faster files locator (glob?)
 - creating a loader like `filesystem` in which you provide exact path to look for the configuration files (like `app/themes/my-theme/composer.json`)
 - cache located files during container compilation (it won't work well in dev though)

stefandoorn:
Or simply only allow composer.json to be in the root of a theme?

pamil:
Yeah, restricting it to scan only one directory deep might also be a solution.

stefandoorn:
Added PR to scan max 1 level deep (`composer.json` should be in root). That's also how docs state it.

stefandoorn:
Updated PR with option to specify depth yourself. Also added information in the docs regarding how to use it. Doesn't mean this PR is obsolete btw.

pamil:
We can change the recommendation to customise `scan_depth` option in case of performance issues.

psren:
Hate stalebot :-)

pamil:
Isn't `scan_depth` already mentioned there and should make this notice unneeded (http://docs.sylius.com/en/latest/components_and_bundles/bundles/SyliusThemeBundle/configuration_sources/filesystem.html)?

RadnoK:
@pamil yup, this was fixed and documented by @stefandoorn so this PR is no longer valid 🎉

PR #9012, opened 2017-12-07
[Documentation][MailerBundle] Clean installation part

odolbeau:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

The mailer bundle don't need doctrine.

There is also a lot of unused dependencies in the require-dev part of the composer.json file of the bundle. Tests also register useless bundles (such as doctrine). Should I create another PR to clean both composer.json & tests?


lchrusciel:
Thanks Olivier! And welcome to our community :)

Also, if you only have some free time, please help us with composer and tests part :)

PR #9013, opened 2017-12-07
Restrict scanning for composer.json in themes to certain directory depth (optional)

stefandoorn:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | mentioned in #9008
| License         | MIT

### DIFF
--- a/UPGRADE-1.0.md
+++ b/UPGRADE-1.0.md
@@ -1,3 +1,10 @@
+# UPGRADE FROM 1.0.7 to 1.0.8
+
+* Scanning for `composer.json` file inside themes directories is recursive by default, which can result in slow performance
+  when e.g. a `node_modules` folder is present inside a theme folder. Supply the optional `scan_depth` (integer) setting
+  to the `sylius_theme` configuration to restrict scanning for the theme configuration file to a specific depth inside
+  the specified theme directories.
+
 # UPGRADE FROM 1.0.1 to 1.0.2
 
 * `Sylius\Bundle\AdminApiBundle\Model\ClientManager`'s `findClientByPublicId($publicId): ClientInterface` signature
--- a/UPGRADE-1.0.md
+++ b/UPGRADE-1.0.md
@@ -1,10 +1,3 @@
-# UPGRADE FROM 1.0.7 to 1.0.8
-
-* Scanning for `composer.json` file inside themes directories is recursive by default, which can result in slow performance
-  when e.g. a `node_modules` folder is present inside a theme folder. Supply the optional `scan_depth` (integer) setting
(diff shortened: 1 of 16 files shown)


UPGRADE-1.0.md
```diff
@@ -1,3 +1,7 @@
+# UPGRADE FROM 1.0.5 to 1.0.6
+
+* Scanning for `composer.json` file inside themes used to be recursive. This is now limited to the root of each theme. This has always been the intended location for this file.
```
pamil: > This is now limited to the root of each theme. This has always been the intended location for this file.

That's true. What has not been said is that themes root have to be not more than one level below `app/themes`. Some people might have set up themes like `app/themes/vendor/package-name/composer.json` and some like `app/themes/FooBarTheme/composer.json` so for the first ones it would result in a BC break.

We might add an option to `FilesystemConfigurationSourceFactory` named `depth`, which defaults to `null` (being the current behaviour with no limit) and then setting it to `1` would restrict the depth in the recursive file locator.

liverbool: Just only one allow `app/themes/vendor/package-name/composer.json`.

stefandoorn:
@pamil Can you have a check again?

I've restricted the configuration now to a single integer (optional, default = null). The depth method call inside the recursive file locator can also be called multiple times, allowing users to define e.g. >1 & < 4, but I'm not sure we need something like that. In that rare case user can maybe create their own solution. I think with this change we cover the biggest performance issue (e.g. scanning a node_modules folder).

stefandoorn:
P.S. Do you also want more tests (functional / phpunit) etc? ThemeBundle has quite some I see.

lchrusciel:
@stefandoorn but we need to make current configuration tests green again

stefandoorn:
Hm damn I thought they would pass yesterday. Will look again into it then.

pamil:
@stefandoorn looks all right! Some functional tests would be great (and an another one for configuration actually). For example, you might add `scan_depth` option in `Tests/Functional/app/config/config.yml`, create another theme in `Tests/Fixtures/themes` with `YetAnother/TestTheme/composer.json` path and check that it is not found.

Btw. can you rebase it to 1.1? It feels more like a new feature (restricting scanning depth) and the release is less than a month away anyway :)

stefandoorn:
@pamil @lchrusciel Ready to be checked again. Rebased against 1.1 and added functional test / configuration checks.

pamil:
Thanks Stefan, themes will be lightning fast right now 🥇

PR #9014, opened 2017-12-08
fix a jasvacript error: double #id in province validation on checkout…

victordit:
…/address
------
I found the error in checkout/address when I was trying to do an event.preventDefault () to validate a field before submit
------
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no/yes
| Deprecations?   | no/yes 
| License         | MIT

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /.php_cs.cache
+/.phpintel
 
 /app/config/parameters.yml
 /app/config/*.local.yml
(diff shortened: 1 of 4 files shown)


pamil:
Thanks for the effort, as @Zales0123 has said, this PR touches too many parts of Sylius (OrderEmailManager), break BC and should be rather handled in the templates itself instead of the Controller code. I'll open an issue for that, feel free to contact us if you're still willing to solve this one.

pamil:
https://github.com/Sylius/Sylius/issues/9134

PR #9020, opened 2017-12-11
Restrict packages to use Symfony's DI in ^3.2 version for tests

pamil:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

We're fetching all services to smoke test whether they are properly wired and Symfony in 4.0 marks all services as private by default, so it's breaking the build. This PR is rather a fast workaround than the proper solution, but it'll do fine for now (and we'll be able to merge other PRs).

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/composer.json
+++ b/src/Sylius/Bundle/AddressingBundle/composer.json
@@ -30,6 +30,7 @@
     "require-dev": {
         "doctrine/orm": "~2.5.0",
         "phpspec/phpspec": "^4.0",
+        "symfony/dependency-injection": "^3.2",
         "symfony/form": "^3.2",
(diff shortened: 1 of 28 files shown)


lchrusciel:
Can't we define all services as public just in kernel env?

pamil:
`_defaults` was introduced in Symfony 3.3 and 1.0 requires Symfony 3.2

lchrusciel:
Aren't we able to do some magic in TestAppKernel?

pamil:
None that I know of - it's only a dev dependency though and there's no official Symfony 4 support as for now.

PR #9037, opened 2017-12-18
Fix mixed flashes

### LINKED ISSUE #8051: Admin & Shop flashes are mixed
Just noticed something not a big bug, but a bit 'weird'. I was editing a product in the Admin, and while saving it opened the front page on the shop. On the shop I got the flash message 'Product is succesfully edited' in my locale, and didn't get to see it in the Admin. It looks to me like the flashes are shared, which could result in this :)

cv65kr:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #8051
| License         | MIT


pamil:
It looks more like a bugfix, so it should target 1.0, but some tests (probably Behat ones) to prevent regression would be needed.

cv65kr:
@pamil rebased, but i don't have a lot experience with Behat. Would be great if anyone do it instead me.

src/Sylius/Bundle/CoreBundle/EventListener/SessionNameSubscriber.php
```diff
@@ -0,0 +1,78 @@
...
+        $firewallConfig = $this->firewallMap->getFirewallConfig($event->getRequest());
+        $cookies = $event->getRequest()->cookies;
+
+        if ($cookies->has($this->session->getName())) {
+            if($this->session->getId() !== $cookies->get($this->session->getName()) ) {
+                $this->session->setName(
+                    sprintf('sylius_%s', $firewallConfig->getName())
+                );
+            }
+        }
```
Zales0123: I would use some early returns to avoid nested `if` statements:
```php
if (!$cookies->has($this->session->getName()) {
    return;
}

if ($this->session->getId() !== $cookies->get($this->session->getName()) ) {
    $this->session->setName(sprintf('sylius_%s', $firewallConfig->getName()));
}
```

src/Sylius/Bundle/CoreBundle/Resources/config/services/listeners.xml
```diff
@@ -83,5 +83,11 @@
             <argument type="service" id="sylius.product_variant_resolver.default" />
             <tag name="kernel.event_listener" event="sylius.product.initialize_update" method="lock" />
         </service>
+
+        <service id="sylius.set_session_name" class="Sylius\Bundle\CoreBundle\EventListener\SessionNameSubscriber">
```
Zales0123: What about something like `sylius.session_name_subscriber`?

cv65kr:
@Zales0123 done.

lchrusciel (approved):

Zales0123:
Tests are green now, but some new scenarios would be indeed needed to prevent from possible future problems - we can't merge it without them. @cv65kr maybe you're willing to try writing them, even if you're not experienced with Behat? :) I'm sure community would help you with any problems you can have and it's a great opportunity to start your adventure with such a wonderful testing tool 🚀

cv65kr:
@Zales0123 I will try do this :) before, i tested this "unprofessionally" because i add ```die()``` between flash message setter and redirect, next i manipulating browser cart (one with admin page, one with shop page) and refresh button and it's worked properly.

I wrote some scenario for this bugfix, but i think it's not good enough.

```
Feature: Show flash message in the appropriate views

  Background:
      Given: I am refresh page with a success flash message
      And: I am logged in as an administrator

  Scenario: Show success flash message on admin page
    Given An "success" flash message not exists
    When I reload the page for "success" flash message
    Then I should see an "success"  flash message
```

Zales0123:
Hello @cv65kr, it's been a while :) It's still an issue, and for sure it should be fixed. Your solution seems correct, but indeed creating a scenario is a little bit tricky :/ Does anybody has some idea how it could be implemented? cc @pamil @lchrusciel

cv65kr:
@Zales0123 rebased 😄

lchrusciel (approved):
We should finally solve this issue in Sylius. However, I'm wondering how it will behave in `api` contexts, where we should avoid starting a new session. Also, what @Zales0123 mention, maybe we could test it just with unit tests. 

Should we blacklist all firewalls with `api` word in them? Should we merge it and iterate over it in the next PR? /cc @Zales0123 @pamil @GSadee

jakubtobiasz:
Closing from the same reason as in #8051.

PR #9048, opened 2017-12-21
Exclude the web/media directory from firewall restrict list.

Hailong:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | no
| License         | MIT

This change fixes the "... locale is unavailable in this channel" exception when loading product images in some condition.

Suppose client's browser locale is en_US, while the store's locale is zh_CN, this bug would prevent any product images from being loaded because the "NonChannelLocaleListener" class wrongly breaks the requests.

So the fix is to exclude the "/media" path from shop firewall match pattern.

### DIFF
--- a/app/config/security.yml
+++ b/app/config/security.yml
@@ -4,7 +4,7 @@
 parameters:
     sylius.security.admin_regex: "^/admin"
     sylius.security.api_regex: "^/api"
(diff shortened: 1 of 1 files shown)


pamil:
Thank you @Hailong! 🥇

PR #9101, opened 2018-01-12
Fix wrong target-entity in exchange-rate entity

alcaeus:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

This fixes an error that surfaces when trying to extend the `Currency` entity: due to the wrong mapping, it's not actually possible to extend the `Currency` entity.

I'm not sure how to reliably test this, apart from looking at all 97 `target-entity` attributes and ensuring they point to interfaces (except a Payum `GatewayConfig` that doesn't have an interface).


pamil:
Thank you Andreas! 🥇

PR #9127, opened 2018-01-29
Use random_*() instead of rand(), mt_rand() and microtime()

pamil:


src/Sylius/Component/Core/Uploader/ImageUploader.php
```diff
@@ -45,7 +45,7 @@ public function upload(ImageInterface $image): void
         }
 
         do {
-            $hash = md5(uniqid((string) mt_rand(), true));
+            $hash = bin2hex(random_bytes(16));
```
pamil: MD5 returns 128 bit (16 bytes) long hash

src/Sylius/Component/Promotion/Generator/PromotionCouponGenerator.php
```diff
@@ -99,7 +99,7 @@ private function generateUniqueCode(int $codeLength, array $generatedCoupons): s
         Assert::nullOrRange($codeLength, 1, 40, 'Invalid %d code length should be between %d and %d');
 
         do {
-            $hash = sha1((string) microtime(true));
+            $hash = bin2hex(random_bytes(20));
```
pamil: MD5 returns 160 bit (20 bytes) long hash

lchrusciel:
Thanks Kamil!

PR #9129, opened 2018-01-30
Add symfony/thanks Composer plugin

pamil:


stefandoorn:
Does this actually work in the meanwhile? Previous time I tried it on my Sylius project the GitHub API failed due to the size of the dependencies graph it seems.. :-D

pamil:
@stefandoorn it might be connected to token's permissions (see symfony/thanks#41).

stefandoorn:
Hm I had another error but it's been a while so can't recall correctly. Tested again locally, seems better now.

nicolas-grekas:
@stefandoorn please open an issue on symfony/thanks if you reproduce that error again

PR #9135, opened 2018-01-30
Fixed Customer API docs with invalid "groups" parameter

gorkalaucirica:
| Q               | A
| --------------- | -----
| Branch?         | 1.0 (affects all)
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | no
| License         | MIT

According to the Customer API docs, "groups" fields is expected to be an array while it has a 1 to N relationship in DB, and it does not work if an array is passed.

### DIFF
--- a/docs/api/customers.rst
+++ b/docs/api/customers.rst
@@ -53,7 +53,7 @@ If you request for a more detailed data, you will receive an object with followi
 +-------------------------+-------------------------------------------+
 | birthday                | Customers birthday                        |
 +-------------------------+-------------------------------------------+
-| groups                  | Array of groups customer belongs to       |
+| group                   | Customer group code                       |
 +-------------------------+-------------------------------------------+
 
 .. note::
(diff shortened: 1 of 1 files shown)


docs/api/customers.rst
```diff
@@ -139,9 +139,7 @@ Exemplary Response
         "firstName":"John",
         "lastName":"Diggle",
         "gender":"m",
-        "group":[
-
-        ]
+        "group":{}
```
pamil: Shouldn't it be `"group":null` or `"group":"GROUP_CODE"` then?

gorkalaucirica: That's the response given by the API right now

lchrusciel: Perhaps we are missing serialization definition of customer groups.

pamil: That's the current behaviour that needs to be documented anyway :)

pamil:
Rebased & build restarted.

pamil:
Thanks Gorka! 🥇

Issue #9105, opened 2018-01-15
[Urgent] An error appear in the Payum capture action when checking out an order


songecko:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.0

This issue appear after a composer update. It updates the Payum bundle to the 1.5.0 tag and subsequently it updates the League Uri library to 5.x. 

Steps to reproduce:

1- Do a composer update.
2- Do a checkout and the error appear when click in the "Place Order" button on the checkout last step.

After a research i did, i found that the League URI need to write a cache file in **/vendor** directory instead **/var/cache**. And of course the /vendor directory is not writable.

Here an screenshot showing the stack trace. URL: /payment/capture/{token}

![2018-01-13](https://user-images.githubusercontent.com/429512/34950833-f781d146-fa1c-11e7-9217-d270d964b994.png)

stefandoorn:
This would also be problematic on hosting environments like Platform.sh, where by default folders are read-only unless specifically marked as a persistent volume between deployments.

Tetragramat:
related issue in payum https://github.com/Payum/Payum/issues/734

pamil:
Tracked it downstream to thephpleague/uri-hostname-parser#2.

Takeno:
It should be a permission error for cache folder: `vendor/league/uri-hostname-parser/data/`.

Easy fix: chmod on that directory.

Tetragramat:
Wasn't that was already fixed in league/uri-components:^1.7.1 ?

Zales0123:
It should be :)

Takeno:
Yeah, it should be fixed with new version of `league/uri-components` but if you can't upgrade for some reason, chmod is a way.

Zales0123:
Nice, thank you @Takeno for useful information, thank you @Tetragramat for clarification, I think it could be closed then :)

PR #9145, opened 2018-02-01
Run Travis with readonly project directories

pamil:
Trying to reproduce #9105.


stefandoorn:
Could you run this one again @pamil? This https://github.com/Payum/Payum/issues/734#issuecomment-366235402 might be interesting. Probably still good to have caching, but this seems to fix the biggest issue for now.

pamil:
@stefandoorn done, let's see what we will get :)

Issue #9107, opened 2018-01-15
Updating product position does not work


bitbager:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.0.5

```
Given I have a product with position set to 1
When I go to the product list page that contains this product
And I set the position to 2
And I press the "Save positions" button
Then I am not notified that the position has been updated
And the position is still 1 😢 
```

Also happens on Sylius demo, take book taxonomy as an example.

pamil:
Could you provide a PR with failing Behat scenario for that?

bitbager:
How should I tag it?

pamil:
No, let it fail (just like #9145).

bitbager:
Done 🎉

pamil:
It's not reproducible on the last releases of `1.1.x` / `1.2.x` and `1.3.x`.

PR #9160, opened 2018-02-07
Add failing Behat scenario for updating product position

bitbager:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |  mentioned in #9107 
| License         | MIT

### DIFF
--- a/features/product/managing_products/changing_products_positions.feature
+++ b/features/product/managing_products/changing_products_positions.feature
@@ -1,0 +1,22 @@
+@managing_products
+Feature: Changing product positions
+    In order to sort products by position
+    As an Administrator
+    I want to be able to change positions of existing products
+
+    Background:
+        Given I am logged in as an administrator
+
+    @ui @javascript
+    Scenario: Changing position of existing products
+        Given the store has a product "Lamborghini Gallardo Model"
+        And this product has a position set to 1
+        And the store has a product "Maserati Quattroporte"
+        And this product has a position set to 2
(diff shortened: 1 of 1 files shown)


pamil:
Looks like the scenario fails because of undefined steps:
```
--- Use --snippets-for CLI option to generate snippets for following ui_managing_products suite steps:
    And this product has a position set to 1
    And this product has a position set to 2
    When I go to the list page for this products
    And I change the position for "Maserati Quattroporte" to 1
    And I change the position for "Lamborghini Gallardo Model" to 2
    And I click save positions
    Then I should be notified that the positions has been updated
    And "Maserati Quattroporte" should now have position set to 1
    And "Lamborghini Gallardo Model" should now have position set to 2
```

bitbager:
Thought you want me to provide the Behat case only. I might not have time to provide a full PR and actually IMHO it is pointless to do the job partially just by providing a Behat tests.

pamil:
That's fine as well, just the more complete the failing scenario is the more probable is that somebody would be willing to make it work :) I wouldn't say defining these steps is pointless, it's the casual process when fixing bugs - reproduce in tests, make tests green, refactor.

pamil:
There's some point in providing that scenario - I could rerun it now and check if it still does not work - couldn't reproduce it manually on Sylius demo though.

PR #9161, opened 2018-02-07
Make images removal tests great again

pamil:
There were lots of failures because the saved image URL was a data URI.


src/Sylius/Behat/Page/Admin/Taxon/UpdatePage.php
```diff
@@ -94,12 +94,16 @@ public function isImageWithTypeDisplayed($type)
     {
         $imageElement = $this->getImageElementByType($type);
 
+        var_dump($imageElement ? $imageElement->find('css', 'img')->getAttribute('src') : 'null');
```
pamil: Debugging JS ain't easy 🌮

Issue #8900, opened 2017-10-25
product select attribute translation error


benyang2015:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.1.0

<img width="951" alt="attribute-error" src="https://user-images.githubusercontent.com/15188607/31993214-1c39038c-b9af-11e7-9070-e3d4b2f0d3b3.png">

GSadee:
Did you do `yarn run gulp`?

benyang2015:
yarn run gulp and clear cache，also get the error

benyang2015:
@GSadee some js error?

GSadee:
I have no idea for now, as I said before, I was trying to reproduce that, but everything works fine for me.

benyang2015:
I update the code from your mege Again， also can't click the second language

pamil:
Can you provide all the changes made before? Your channels configurations, your locales, your base locale etc.

benyang2015:
![image](https://user-images.githubusercontent.com/15188607/32499496-b1b9efda-c40d-11e7-85b5-c5b539245a9d.png)
![image](https://user-images.githubusercontent.com/15188607/32499543-ca493268-c40d-11e7-9c0c-6d8fde8e0831.png)

Zales0123:
I think I've managed to reproduce it 😄 there is a fix PR already opened

PR #9165, opened 2018-02-08
Make accordion in select attribute values great again

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/8900
| License         | MIT

For sure it's not the best solution from the js point of view, but it works :)
The problem was that newly added select attribute values has translation accordion inside which needs to be triggered to make them work.

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
+++ b/src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
@@ -73,5 +73,11 @@
 
         $(document).previewUploadedImage('#sylius_product_images');
         $(document).previewUploadedImage('#sylius_taxon_images');
+
+        $('body').on('DOMNodeInserted', '[data-form-collection="item"]', function() {
+            if ($(this).find('.accordion').length > 0) {
+                $(this).find('.accordion').accordion();
+            }
+        });
     });
 })(jQuery);


src/Sylius/Bundle/AdminBundle/Resources/private/js/sylius-select-attribute.js
```diff
@@ -0,0 +1,24 @@
...
+
+(function ($) {
+    'use strict';
+
+    $.fn.extend({
+        handleSelectAttributeTranslatableValues: function () {
+            var element = $(this);
+
+            element.find('[data-form-collection="add"]').on('click', function() {
+                setTimeout(function() {
```
stefandoorn: What is the benefit of waiting 100ms? Or why doesn't it work without I mean..

Zales0123: If we listen on "click" of the **Add** button, it takes some miliseconds to add this value, triggering `accordion()` immediately don't catch the latest element. As I said, I'm not a JS-guy at all, I just know that it helps 😄 I'll appreciate any better solution.

stefandoorn: Probably listening to the adding of the element is 'more clean' (https://stackoverflow.com/questions/18746825/adding-an-event-to-catch-when-a-list-item-is-added-to-a-select-jquery), but this also does the job probably :-D

Zales0123: @stefandoorn this `DOMNodeInserted` trick did the job! Thanks 😄

PR #9169, opened 2018-02-08
Allow for using Doctrine ORM >=2.6

pamil:


src/Sylius/Bundle/ProductBundle/spec/EventListener/SelectProductAttributeChoiceRemoveListenerSpec.php
```diff
@@ -43,8 +47,9 @@ function it_removes_select_product_attribute_choices(
 
         $productAttribute->getType()->willReturn(SelectAttributeType::TYPE);
 
-        $entityManager->getUnitOfWork()->willReturn($unitOfWork);
-        $unitOfWork->getEntityChangeSet($productAttribute)->willReturn([
```
pamil: Need to use Mockery for UnitOfWork mock as Doctrine ORM 2.6 changes `getEntityChangeSet` definition to return a reference instead of a value and it's no supported by Prophecy.

pamil: To be honest this spec should be rewritten to follow "don't mock what you don't own" rule and behaviour tested by Behat scenarios, but this one is the quickest solution, allows us to install Doctrine ORM 2.6 and fixes the error seen in #9164.

src/Sylius/Bundle/ProductBundle/spec/EventListener/SelectProductAttributeChoiceRemoveListenerSpec.php
```diff
@@ -100,10 +107,8 @@ function it_does_not_remove_select_product_attribute_choices_if_there_is_only_ad
             ],
         ]);
 
-        $entityManager
-            ->getRepository('Sylius\Component\Product\Model\ProductAttributeValue')
-            ->shouldNotBeCalled()
-        ;
+        $entityManager->getUnitOfWork()->willReturn($unitOfWork);
+        $entityManager->getRepository('Sylius\Component\Product\Model\ProductAttributeValue')->shouldNotBeCalled();
```
Zales0123: Is there any reason for not using `ProductAttributeValue::class`?

src/Sylius/Bundle/ProductBundle/spec/EventListener/SelectProductAttributeChoiceRemoveListenerSpec.php
Zales0123: Sad but true

src/Sylius/Bundle/ProductBundle/spec/EventListener/SelectProductAttributeChoiceRemoveListenerSpec.php
pamil: Not really, just wanted not to change too much as I would rather rewrite the whole test instead :)

Issue #9062, opened 2017-12-27
Symfony 4.0 support


pamil:
Symfony 4.0 has been released together with PHP 7.2 last month (see #9009 for PHP 7.2 support status), so here's the plan:

Sylius 1.0 won't officially support Symfony 4.0 - it will stay with Symfony 3.3-3.4. The reasoning behind this is that we can only rely on Symfony 3.3 features there, so that if something has got deprecated in Symfony 3.4 and there was a forward compatibility layer provided for that, then we still can't use that without breaking backwards compatibility with Symfony 3.3. And some of our components / bundles requires ^3.2 Symfony components.

Sylius 1.1 will require Symfony 3.4 - support for Symfony 4.0 will come as soon as we solve all the deprecations in our and vendors' code.

The only missing step to start fighting deprecations is to set up Travis to include Symfony 4.0 jobs, which I'll have taken care of in this year :)

sstok:
https://github.com/SymfonyTest/symfony-bundle-test maybe this can help with testing the bundles (that don't have have any additional tests) for multiple Symfony versions.

pamil:
@sstok yeah, we have similar setup for the application tests (requiring `symfony/symfony` in a specific version) but for the packages tests we would need to require Symfony components in a specific version.

Eg. User package:
https://github.com/Sylius/Sylius/blob/cf9493076f3b59557be80efc967e240bb99b991d/src/Sylius/Component/User/composer.json#L30-L37

Then we should require `symfony/security` in a given version, but abstain from requiring `symfony/polyfill-mbstring` in that version too, as it has different release cycle.

sstok:
Symfony introduced the concept of https://github.com/symfony/lts to prevent getting "old" versions, 
but @dunglas (sorry for pinging) created https://github.com/dunglas/symfony-lock as an alternative to lock to one specific major version (ensuring all components/bundles/bridges are of the same version and no 4.0 is installed when locking to 3.x), especially for CI.

pamil:
Major locks would work here, I'm still thinking how to test the changes between minor versions (https://github.com/dunglas/symfony-lock/issues/3), as a project this big usually hits a regression or a few even when upgrading a minor. This way we could test Symfony 3.4 and 4.0 in every build and include another Symfony versions (like 3.3 for Sylius 1.0 or 4.1 for Sylius 1.1+) in nightly build.

sstok:
[...]

PR #9183, opened 2018-02-13
[POC] Going through dependency hell to Symfony 4.0

pamil:
| Q               | A
| --------------- | -----
| Branch?         | 1.1 / master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | related to #9062
| License         | MIT

This PR aims to show how many libraries needs our help to support 4.0 before we can even think of supporting 4.0 in Sylius itself 🎉  I'll probably make a Github project and assign related issues for each dependency there.


composer.json
```diff
@@ -32,39 +32,39 @@
...
         "symfony/swiftmailer-bundle": "^3.0",
-        "symfony/symfony": "^3.4",
+        "symfony/symfony": "^4.0",
         "symfony/thanks": "^1.0",
         "twig/extensions": "^1.4",
         "twig/twig": "^2.0",
         "webmozart/assert": "^1.1",
         "white-october/pagerfanta-bundle": "^1.0.8",
-        "willdurand/hateoas": "~2.10.0",
+        "willdurand/hateoas": "^2.11@dev",
```
pamil: 2.11.0 release introduces a regression from 2.10.0 and it should be fixed and merged to 1.0 first.

composer.json
```diff
@@ -32,39 +32,39 @@
...
         "polishsymfonycommunity/symfony-mocker-container": "^1.0",
         "ramsey/uuid": "^3.7",
-        "sensio/distribution-bundle": "^5.0.21",
         "sonata-project/block-bundle": "^3.3",
         "sonata-project/intl-bundle": "^2.2",
         "stof/doctrine-extensions-bundle": "^1.2",
         "swiftmailer/swiftmailer": "^6.0",
-        "sylius-labs/association-hydrator": "^1.0",
+        "sylius-labs/association-hydrator": "^1.1",
+        "sylius-labs/sensio-distribution-bundle": "^6.0",
```
pamil: I've forked DistributionBundle because issue about Symfony 4 support has been immediately closed, so we have to either use Flex (whereas we need only script executor from there now) or our DistributionBundle fork (but we take the maintenance burden, while we're lacking human resources for that).

composer.json
```diff
@@ -73,28 +73,26 @@
...
+        "friends-of-behat/cross-container-extension": "^1.1",
         "friends-of-behat/service-container-extension": "^1.0",
-        "friends-of-behat/symfony-extension": "^1.0",
-        "friends-of-behat/variadic-extension": "^1.0",
-        "hwi/oauth-bundle": "^0.5",
-        "lakion/api-test-case": "^1.0",
+        "friends-of-behat/symfony-extension": "^1.1",
+        "friends-of-behat/variadic-extension": "^1.1",
+        "hwi/oauth-bundle": "^0.6",
+        "lakion/api-test-case": "dev-test-matrix-coduo",
```
pamil: 3.0 release should be ready soon**ish** and it probably would include Lakion/ApiTestCase#111.

composer.json
pamil: See sensiolabs/SensioDistributionBundle#334.

composer.json
```diff
@@ -73,28 +73,26 @@
...
         "lakion/mink-debug-extension": "^1.2.3",
-        "malukenho/kawaii-gherkin": "^0.1",
         "matthiasnoback/symfony-config-test": "^2.0",
         "matthiasnoback/symfony-dependency-injection-test": "^1.0",
         "mikey179/vfsStream": "^1.6",
         "pamil/prophecy-common": "^0.1",
         "phpspec/phpspec": "^4.0",
-        "phpunit/phpunit": "^5.6",
+        "phpunit/phpunit": "^6.0",
```
pamil: We can even think of ^7.0 tbh.

composer.json
umpirsky: @pamil Where is the fork? :)

PR #9212, opened 2018-02-26
[Reviews] nullable title for reviews

### LINKED ISSUE #8172: [Review] Why title is required ?
On many websites, you can rate something without comment it.
On Review model, comment is nullable but title is not, it's weird. On few projects, I use review bundle in this use case and I needed to work around.

What do you think ?

loic425:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #8172 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ReviewBundle/Resources/config/validation/Review.xml
+++ b/src/Sylius/Bundle/ReviewBundle/Resources/config/validation/Review.xml
@@ -17,7 +17,7 @@
             <constraint name="NotBlank">
                 <option name="message">sylius.review.title.not_blank</option>
                 <option name="groups">
-                    <value>sylius</value>
+                    <value>sylius_review</value>
                 </option>
             </constraint>
             <constraint name="Length">
--- a/app/migrations/Version20180226142349.php
+++ b/app/migrations/Version20180226142349.php
@@ -1,0 +1,28 @@
+<?php declare(strict_types = 1);
+
+namespace Sylius\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20180226142349 extends AbstractMigration
+{
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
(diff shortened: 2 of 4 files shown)


src/Sylius/Bundle/ReviewBundle/Resources/config/validation/Review.xml
```diff
@@ -17,7 +17,7 @@
             <constraint name="NotBlank">
```
pamil: Shouldn't this constraint be removed now?

loic425: I don't think so. IMHO, when you write a commented review, you need a title and a comment, when you have a "rating system" by only clicking on stars for example, title is not required.

pamil: Does it work like that now (title is required if comment is filled)? I guess we'd need some Behat scenario to cover that.

loic425: No, I just kept existing behaviour on sylius core :)
We can go further but it's more complex. For exemple, you need to separate ratings from commented reviews on frontend. Moreover, the status workflow is different. A single rate doesn't need to be accepted.

loic425: @pamil I think it's just a first step for sylius. It's just an improvment for sylius review bundle and an easier way to improve sylius review behaviour in a sylius shop.

loic425: @Zales0123 @pamil @pjedrzejewski @lchrusciel 
Could you think about that, discuss this and tell me what do you think about the best behaviour I could implement.

Currently, with existing Pull request, I added a new validation group (sylius_review) on title to allow using only "sylius" validation group. Another solution could be the idea suggested by Pamil: title is required only if comment is filled. I guess it needs a form listener to do that.

loic425: Any idea?

Zales0123: Sorry @loic425 for the long response. I thing that requiring comment only if title is filled is the best approach, just seems as the most natural solution. /cc @pamil @pjedrzejewski @lchrusciel

loic425: @Zales0123 Ok, I'm gonna work on that approach.

pamil:
Thanks Loic!

pamil:
I've created #9239 to track that conditional validation feature.

PR #9219, opened 2018-02-28
Improve flags support

### LINKED ISSUE #9184: [Locale] Flags does not work properly
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.1.0

When you create a new locale, for some countries the form return two letters as code (example : France => "fr"), and five for the others one (example : United States => en_US).
The problem is that in the template, when the application try to display flag that doesn't work for the locale with only two letters : 

capture d ecran 2018-02-14 a 04 55 01

shadypierre:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9184 
| License         | MIT

This PR improve flags support, it's not perfect en doesn't work for every country but at least that support more countries than previously.


pamil:
Thanks @shadypierre! 🥇

Issue #8744, opened 2017-09-26
[Admin] Change position of options causes bug in product variant form


czende:
Steps to reproduce on [demo shop](http://demo.sylius.org/) with some T-Shirt:
1. Change position of `t_shirt_size` in Product options to `0`
2. Check for example T-Shirt "autem" variants
3. Grid has proper values but if you want to edit some variant, there are incorrect values in select boxes (first value without `selected` attribute).

Grid:
![screen shot 2017-09-26 at 16 38 25](https://user-images.githubusercontent.com/5628485/30866341-3e60d148-a2d9-11e7-8523-a7a2a8aa5021.png)

Form:
![screen shot 2017-09-26 at 16 39 34](https://user-images.githubusercontent.com/5628485/30866369-5072b216-a2d9-11e7-9a3a-984612e19bca.png)

I tested it on different configurable products and options and values, dumped different variables in code, but I'm always getting empty string in `FormView.vars.value` so twig macro hasn't default value to be selected.

czende:
@Zales0123 Please any update on this? Or suggestion what is causing it? I'm ready to PR, but I'm not able to find it 😞  Thank you!

czende:
Fixed in this PR #9223

PR #9223, opened 2018-03-02
[ProductBundle] Fix data value ProductOptionValue in ProductOptionValueCollectionType

### LINKED ISSUE #8999: Product variant option value not selected in admin
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | ???
| RFC?             | no
| Sylius version   | 1.0.3

## Problem

In the Sylius admin, when editting a product variant with two options: Color and Size a situation can occur when the correct option value (eg: red or XL) is not loaded in the `<select>` and the default value is selected instead.  

It turns out to be caused by the order in which the option values are saved. This is visible in \Symfony\Component\Form\ChoiceList\ArrayChoiceList::getValuesForChoices where $this->options is a list of colors, but the $choices array contains a size.

## Expected situation
The correct option value is selected regardless of the order the options are saved in the database.

## Solution
As for a solution I can think of three possibilities:
 - Correctly sort the values in the query (but doctrine fecthes the options, so i think this is not the best solution/possible)
 - Add a sort somewhere in the code, before values/options are passed to the Symfony formbuilder.
 - Always make sure the options/values are in the correct order in the database (bad practise imo.)
 
 
I have not managed to find a solution. Maybe I can create a patch if someone can confirm the situation/point in a direction.

### Working situation:
```
SELECT
  t0.code AS code_1,
  t0.id AS id_2,
  t0.option_id AS option_id_3
FROM
  sylius_product_option_value t0
  INNER JOIN sylius_product_variant_option_value
    ON t0.id = sylius_product_variant_option_value.option_value_id
WHERE sylius_product_variant_option_value.variant_id = 5206;
```

```
code_1   id_2  option_id_3  
sz-s     25    3            
cl-navy  29    4  
```   

form_working       
code_working

### Bug situation:
```
SELECT
  t0.code AS code_1,
  t0.id AS id_2,
  t0.option_id AS option_id_3
FROM
  sylius_product_option_value t0
  INNER JOIN sylius_product_variant_option_value
    ON t0.id = sylius_product_variant_option_value.option_value_id
WHERE sylius_product_variant_option_value.variant_id = 5196;
```

```
code_1   id_2  option_id_3  
cl-aqua  24    4            
sz-s     25    3 
```           

form_broken
code_broken

Maybe related to:
https://github.com/Sylius/Sylius/issues/8144

monro93:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #8999
| License         | MIT

We found a bug in ProductOptionValueCollectionType where in our ProductVariant edition form were not taken in consideration the actual value of the ProductOption to show it as a default value. We are using the ProductBundle as standalone.


stefandoorn:
Is this the solution for https://github.com/Sylius/Sylius/issues/8999?

monro93:
Yes, this should fix this issue

czende:
Approved that this should fix #8744 and #8999, thank you @monro93 !

pamil:
Could you provide a Behat scenario to prevent regressions?

monro93:
Hi, I've pulled from the last version of 1.1 and the test are passing :D. 
I've never did before a Behat scenario, this is my frist time, so please tell me if something is wrong, let's go:
```
Given a store with a product option with the name "T-Shirt color" and values "Red", "Black", and "White"
Given a store with a product with the name "T-Shirt Hawaian" and the options "T-Shirt color"
Given a store with a product variant with the name "T-Shirt Hawaian White" and the option "T-Shirt color" setted to "White"
When I edit the product variant "T-Shirt Hawaian White"
Then the product variant option selected value should be "White" 
```

Thank you!

Zales0123:
@monro93 it's good you want to try Behat, it's useful when working with Sylius :) Regarding your proposal, you can take a look at the [existing Sylius features about product variants](https://github.com/Sylius/Sylius/tree/master/features/product/managing_product_variants), to know how they should look like. Two advices at the beginning:
1. Behat scenarios should be written as naturally as they can (e.g. `And the product "PHP Mug" has "Medium PHP Mug" variant priced at "$40.00"`)
2. Keywords at the beginning of sentences should always be used in the same order (`Given` -> `When` -> `Then`) with conjunctions between (`And` or `But`). Good luck ;)

pamil:
We need two more things to get this merged:
 - rebase to `1.0` (as the bug it fixes, #8999, was reported for 1.0.3)
 - include Behat scenario in the PR to prevent regressions and ensure things are working all right after merging it (as I can't trust my common sense when it comes to forms :))

nbjohan:
I have very little experience with Behat, but I've looked into adding a Behat scenario. The regression only occurs when some data in the database is saved in 'reversed' order. So I think adding a behat scenario will not be able to test for this error (#8999) (but validating the default values regardless of #8999 still might be useful in general).

Could this fix be tested by adding a unittest to \Sylius\Bundle\ProductBundle\Form\Type\ProductOptionValueCollectionType ?

stefandoorn:
@pamil @Zales0123 Could we have a look into this issue? I can confirm similar behavior, only issue is that I have no clue yet how it can be reproduced, so it's hard to define a feature for it right now.

Zales0123:
I'll take a look at it as soon as possible 👍 I'm still thinking about a Behat scenario that could be written to simulate a bug... I feel unsafe with merging something without tests 😄 🐰

Zales0123:
Merged in https://github.com/Sylius/Sylius/pull/10364, thank you! 🖖

PR #9235, opened 2018-03-07
[ResourceBundle] make controller public by default

bendavies:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

unsure if this is a bug fix. shall i raise target 1.0?

Symfony's `ContainerControllerResolver` is is triggering a deprecation warning on when retrieving controllers from the container because they are private by default. They should be public.
```
The "..." service is private, getting it from the container is deprecated since Symfony 3.2 and will fail in 4.0. You should either make the service public, or stop using the container directly and use dependency injection instead.
```


pamil:
Thanks Ben! 🎉

PR #9237, opened 2018-03-08
Fix select province in checkout address

### LINKED ISSUE #8330: [Checkout][Address] Province cannot be selected using address book
Issue found on *Sylius standard 1.0.0-beta.3*.

**How to reproduce:**

1. Define provinces for a country in administration
2. Log in as an customer
3. Create an address with a defined province selected
4. Go to checkout
5. Select the address from the address book

**Current state:**

When the address is selected, province is selected only if the same country was already selected.

Even if the province was selected, the province selection is reloaded and the province is unselected.

The #7944 makes it even worse as no validation error is shown.

**Expected state:**

When the address is selected, province should be selected as well

serieznyi:
| Q                     | A
| ------------------- | -----
| Branch?           | 1.0
| Bug fix?           | yes
| New feature?   | no
| BC breaks?      | no/yes
| Deprecations? | no
| Related tickets | fixes #8330
| License            | MIT

Before changes
https://drive.google.com/open?id=1ddyaMFtmW0M3oXX1JrfV8HuTvlk1jFY4

After changes
https://drive.google.com/open?id=1quA3mWzYAUoikCnTQLvgW0PBrLuHWaJH

Maybe not a best solution, but it`s work

Also it is work for billing address

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/private/js/sylius-address-book.js
+++ b/src/Sylius/Bundle/ShopBundle/Resources/private/js/sylius-address-book.js
@@ -19,14 +19,12 @@
                 forceSelection: false,
 
                 onChange: function (name, text, choice) {
-                    var provinceCode = choice.data()['provinceCode'];
-                    var provinceName = choice.data()['provinceName'];
+                    var provinceCode = choice.data()['provinceCode'],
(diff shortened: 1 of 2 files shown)


pamil:
Thank you Maxim! 🥇

PR #9249, opened 2018-03-12
[Fixtures] Creating images from array for Taxons

RadnoK:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

From Fixtures level we are not able to add images to Taxons as we can in Products. This PR adds missing configuration 😉

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/BookProductFixture.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/BookProductFixture.php
@@ -93,6 +93,10 @@ class BookProductFixture extends AbstractFixture
                     'name' => 'Books',
                 ],
             ],
+            'images' => [
+                [sprintf('%s/../Resources/fixtures/%s', __DIR__, 'books.jpg'), 'main'],
+                [sprintf('%s/../Resources/fixtures/%s', __DIR__, 'books.jpg'), 'thumbnail'],
+            ]
         ]]]);
 
         $bookGenres = [
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Factory/TaxonExampleFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Factory/TaxonExampleFactory.php
@@ -14,12 +14,15 @@ declare(strict_types=1);
 namespace Sylius\Bundle\CoreBundle\Fixture\Factory;
 
 use Sylius\Component\Core\Formatter\StringInflector;
+use Sylius\Component\Core\Model\TaxonImageInterface;
 use Sylius\Component\Core\Model\TaxonInterface;
+use Sylius\Component\Core\Uploader\ImageUploaderInterface;
 use Sylius\Component\Locale\Model\LocaleInterface;
 use Sylius\Component\Resource\Factory\FactoryInterface;
 use Sylius\Component\Resource\Repository\RepositoryInterface;
 use Sylius\Component\Taxonomy\Generator\TaxonSlugGeneratorInterface;
 use Sylius\Component\Taxonomy\Repository\TaxonRepositoryInterface;
+use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Component\OptionsResolver\Options;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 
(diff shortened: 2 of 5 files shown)


src/Sylius/Bundle/CoreBundle/Fixture/Factory/TaxonExampleFactory.php
```diff
@@ -128,9 +151,30 @@ protected function configureOptions(OptionsResolver $resolver): void
...
             ->setDefault('children', [])
             ->setAllowedTypes('children', ['array'])
+            ->setDefault('images', [])
+            ->setAllowedTypes('images', ['array'])
         ;
     }
 
+    /**
+     * @param array $image
+     * @return TaxonImageInterface
```
Zales0123: Missing blank line above

lchrusciel: Missing blank line above

lchrusciel:
Thanks Koni! 👍

src/Sylius/Bundle/CoreBundle/Fixture/Factory/TaxonExampleFactory.php
```diff
@@ -55,22 +63,33 @@ class TaxonExampleFactory extends AbstractExampleFactory implements ExampleFacto
...
      * @param FactoryInterface $taxonFactory
+     * @param FactoryInterface $taxonImageFactory
      * @param TaxonRepositoryInterface $taxonRepository
      * @param RepositoryInterface $localeRepository
      * @param TaxonSlugGeneratorInterface $taxonSlugGenerator
+     * @param ImageUploaderInterface $imageUploader
      */
     public function __construct(
         FactoryInterface $taxonFactory,
+        FactoryInterface $taxonImageFactory,
```
pamil: It's a BC break

src/Sylius/Bundle/CoreBundle/Fixture/Factory/TaxonExampleFactory.php
```diff
@@ -55,22 +63,33 @@ class TaxonExampleFactory extends AbstractExampleFactory implements ExampleFacto
...
         TaxonRepositoryInterface $taxonRepository,
         RepositoryInterface $localeRepository,
-        TaxonSlugGeneratorInterface $taxonSlugGenerator
+        TaxonSlugGeneratorInterface $taxonSlugGenerator,
+        ImageUploaderInterface $imageUploader
```
pamil: Same here

src/Sylius/Bundle/CoreBundle/Fixture/Factory/TaxonExampleFactory.php
```diff
@@ -128,9 +151,30 @@ protected function configureOptions(OptionsResolver $resolver): void
...
+            ->setDefault('images', [])
+            ->setAllowedTypes('images', ['array'])
         ;
     }
 
+    /**
+     * @param array $image
+     * @return TaxonImageInterface
+     */
+    private function createImages(array $image): TaxonImageInterface
```
pamil: `['foo.jpg', 'this', 'does', 'not', 'matter', 'thumbnail']` has the same effect as `['foo.jpg', 'thumbnail']`

src/Sylius/Bundle/CoreBundle/Fixture/BookProductFixture.php
```diff
@@ -93,6 +93,10 @@ public function load(array $options): void
                     'name' => 'Books',
                 ],
             ],
+            'images' => [
+                [sprintf('%s/../Resources/fixtures/%s', __DIR__, 'books.jpg'), 'main'],
```
pamil: This how it would look like in YAML:

```yml
images:
  - ["path.jpg", "thumbnail"]
```

I guess instead of indexed array it's better to use associative array:

```yml
images:
  - { path: "foo.jpg", type: "thumbnail" }
```

pamil:
🚔 Reverting because of these BC breaks (and it should be sent to `master` branch then)

pamil:
The way to add this functionality without a BC break would be to add nullable arguments at the end of the constructor and skipping loading these images if these dependencies are not injected.

PR #9267, opened 2018-03-15
Add PaymentMethod::instructions option to fixtures

stefandoorn:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Not a real bug, but just a missing option. Default is still `null`, so it doesn't confuse as it's an optional text.


pamil:
Could you add a configuration test for that (`PaymentMethodFixtureTest`)?

stefandoorn:
Done @pamil.

Zales0123:
Thank you Stefan!

PR #9268, opened 2018-03-15
Documentation - Use nullable return type

Holicz:
In doctrine config the field flag is defined as nullable, so we should use nullable type in getter.

| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/docs/customization/model.rst
+++ b/docs/customization/model.rst
@@ -59,9 +59,9 @@ Assuming that you would want to add another field on the model - for instance a 
         private $flag;
 
         /**
-         * @return bool
+         * @return bool|null
          */
-        public function getFlag(): bool
+        public function getFlag(): ?bool
         {
             return $this->flag;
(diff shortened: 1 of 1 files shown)


docs/customization/model.rst
```diff
@@ -61,7 +61,7 @@ Assuming that you would want to add another field on the model - for instance a
         /**
          * @return bool
```
pamil: bool|null

lchrusciel:
To which branch should it be merged? 1.1 or 1.0? /cc @Zales0123

Zales0123:
You're right, probably it should be `1.0` 👍

pamil:
Thank you, Lukas! 🎉

PR #9277, opened 2018-03-23
[ThemeBundle] Remove deprecated translator selector usage

adrienlucas:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | NA
| License         | MIT

>  Passing a "Symfony\Component\Translation\MessageSelector" instance into the "Symfony\Component\Translation\Translator::__construct" as a second argument is deprecated since Symfony 3.4 and will be removed in 4.0. Inject a "Symfony\Component\Translation\Formatter\MessageFormatterInterface" implementation instead.

See https://github.com/symfony/symfony/blob/b9a2e21326ad93aab1350e566afced701942ac6d/src/Symfony/Component/Translation/Translator.php#L90

### DIFF
--- a/src/Sylius/Bundle/ThemeBundle/Tests/Translation/TranslatorTest.php
+++ b/src/Sylius/Bundle/ThemeBundle/Tests/Translation/TranslatorTest.php
@@ -17,9 +17,9 @@ use PHPUnit\Framework\TestCase;
 use Sylius\Bundle\ThemeBundle\Translation\Provider\Loader\TranslatorLoaderProvider;
 use Sylius\Bundle\ThemeBundle\Translation\Provider\Resource\TranslatorResourceProvider;
 use Sylius\Bundle\ThemeBundle\Translation\Translator;
+use Symfony\Component\Translation\Formatter\MessageFormatterInterface;
 use Symfony\Component\Translation\Loader\ArrayLoader;
 use Symfony\Component\Translation\MessageCatalogue;
-use Symfony\Component\Translation\MessageSelector;
 
 /**
  * @see \Symfony\Component\Translation\Tests\TranslatorTest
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/ThemeBundle/Tests/Translation/TranslatorTest.php
```diff
@@ -363,8 +363,14 @@ private function createTranslator($locale = 'en', $options = []): Translator
     {
         $loaderProvider = new TranslatorLoaderProvider();
         $resourceProvider = new TranslatorResourceProvider();
-        $messageSelector = $this->getMockBuilder(MessageSelector::class)->getMock();
 
-        return new Translator($loaderProvider, $resourceProvider, $messageSelector, $locale, $options);
+        $messageFormatter = new class implements MessageFormatterInterface {
+            public function format($message, $locale, array $parameters = array())
+            {
+                return strtr($message, $parameters);
```
adrienlucas: This can even be simplified to `return $message;`

lchrusciel:
I believe it should be open to 1.0.

adrienlucas:
@lchrusciel it seems that the 1.0 version require `"symfony/symfony": "^3.3.13"` while the deprecation has been added to Symfony 3.4.
The 3.4 version is required as of Sylius 1.1, so maybe I should request the change against this last one ?

lchrusciel:
You are right. It should be opened to 1.1

adrienlucas:
@lchrusciel done !

src/Sylius/Bundle/ThemeBundle/Translation/Translator.php
```diff
@@ -47,14 +47,14 @@
...
      * @param TranslatorResourceProviderInterface $resourceProvider
-     * @param MessageSelector $messageSelector
+     * @param MessageFormatterInterface $messageFormatter
      * @param string $locale
      * @param array $options
      */
     public function __construct(
         TranslatorLoaderProviderInterface $loaderProvider,
         TranslatorResourceProviderInterface $resourceProvider,
-        MessageSelector $messageSelector,
```
pamil: 🚨 BC break alert 🚨 

The way we can solve it here is to remove the typehint and do instanceof checks inside to accept `MessageSelector|MessageFormatterInterface`. We'll also need tests which use both of them.

pamil:
Thanks for the effort to make this PR! In my opinion, it should go to 1.2 (master) as it isn't a bugfix but a deprecation removal (and I'd like to keep bugfix branches changes as small as possible).

PR #9284, opened 2018-03-28
[Grid] Filtering orders bug

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | Fixes https://twitter.com/nickchambers/status/977964619787587585
| License         | MIT

Bug was reporter on Twitter 🐦 by Nick Chambers. I'm not 100% sure about this approach, as it requires providing default time in both form type and filter itself. Nevertheless, it works nice (as shown in scenario). I assume that proposed approach (`00:00` as default time for *from* and `23:59` as default time for *to*) should be appropriate for most cases :)

### DIFF
--- a/features/order/managing_orders/filtering_orders_by_date.feature
+++ b/features/order/managing_orders/filtering_orders_by_date.feature
@@ -7,15 +7,15 @@ Feature: Filtering orders
     Background:
         Given the store operates on a single channel in "United States"
         And the store has customer "Mike Ross" with email "ross@teammike.com"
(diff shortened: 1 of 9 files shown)


pamil:
Can we add a test for filtering something made on 23:59:59?

Zales0123:
@pamil done

pamil:
Thank you Matt! 🎋

Issue #8544, opened 2017-09-08
UUID for entities, routing to be changed


PapsOu:
Hello,

according to these issues (#3379, #6511), there still an implicit usage of IDs as integer.

For example :
```yml
sylius_shop_partial_product_review_latest:
    path: /latest/{count}
    methods: [GET]
    defaults:
        _controller: sylius.controller.product_review:indexAction
        _sylius:
            template: $template
            repository:
                method: findLatestByProductId
                arguments:
                    - !!int $productId
                    - !!int $count
        count: 3
```

$productId is still handled only as integer (double and float allowed too in vendor/sylius/sylius/src/Sylius/Bundle/ResourceBundle/Controller/ParametersParser.php

This issue happen when overriding the Sylius product with custom entity that uses Uuid instead of autogenerated int ID. It cause this exception :
```log
An exception has been thrown during the rendering of a template ("An exception occurred while executing 'SELECT l0_.title AS title_0, l0_.rating AS rating_1, l0_.comment AS comment_2, l0_.status AS status_3, l0_.created_at AS created_at_4, l0_.updated_at AS updated_at_5, l0_.id AS id_6, l0_.product_id AS product_id_7, l0_.author_id AS author_id_8 FROM librinfo_ecommerce_productreview l0_ WHERE l0_.product_id = ? AND l0_.status = ? ORDER BY l0_.created_at DESC LIMIT 3' with params [9223372036854775807, "accepted"]:

SQLSTATE[22P02]: Invalid text representation: 7 ERROR: invalid input syntax for uuid: "9223372036854775807"").
```

PapsOu:
By searching `- !!int ` in whole sylius project, it seems that the only entity id forced to int is of this route (sylius_shop_partial_product_review_latest,  defined in vendor/sylius/sylius/src/Sylius/Bundle/ShopBundle/Resources/config/routing/partial/product_review.yml)

Do you need a PR to remove this route constrain ?

pamil:
@PapsOu yes, IDs shouldn't be typehinted cause we make them `mixed` everywhere, a PR will be more than welcome! :)

pjedrzejewski:
@pamil So we should change it everywhere to mixed?

PapsOu:
@pjedrzejewski @pamil Here the PR : https://github.com/Sylius/Sylius/pull/8549

PapsOu:
[...]

PR #9286, opened 2018-03-28
Do not assume that model id is int

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes (scalar type hint)
| Deprecations?   | no
| Related tickets | follow up for #8544
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/Model/GatewayConfig.php
+++ b/src/Sylius/Bundle/PayumBundle/Model/GatewayConfig.php
@@ -19,14 +19,14 @@ use Sylius\Component\Resource\Model\ResourceInterface;
 class GatewayConfig extends BaseGatewayConfig implements ResourceInterface
 {
     /**
-     * @var int
+     * @var mixed
      */
     protected $id;
 
     /**
-     * @return int
+     * {@inheritdoc}
      */
-    public function getId(): int
+    public function getId()
     {
         return $this->id;
     }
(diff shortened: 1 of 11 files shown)


teohhanhui:
There's 1 place where there's BC break because of the return type hint being removed. It means we promised before that it should return an int, but now we revoke that promise. The client code might not have handled it.

stefandoorn:
Isn't `null|int` and `getId(): ?int` better in this case? Sylius models don't use UUID's or something as far as I recall. Or is the idea to make it possible to do something like that? I currently rely on the fact that it's an ID, which won't change of course with this change but it might become more unclear what's being returned (less strict contract).

teohhanhui:
The idea is that Sylius should not care whether `id` is an `int` or not. It should not matter. Changing the contract means we expect that Sylius and extensions should not assume that `id` is `int`.

teohhanhui:
About the BC promise, maybe we're fine because of http://docs.sylius.com/en/latest/contributing/code/bc.html#models-model-interfaces

src/Sylius/Bundle/PayumBundle/Model/GatewayConfig.php
```diff
@@ -19,14 +19,14 @@
...
-     * @var int
+     * @var mixed
      */
     protected $id;
 
     /**
-     * @return int
+     * {@inheritdoc}
      */
-    public function getId(): int
```
pamil: Need to check whether it is a BC break or not.

teohhanhui:
How do we proceed with this one?

lchrusciel:
Hello Han, 

I know it is been a while since we moved anything in this PR. IMHO removing `int` typehint is not a BC break and we could have it finally merged. Could you rebase to the newest master?

PR #9306, opened 2018-04-06
[Resource] Make sure Sylius resources services are public

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Since Symfony 3.4 services are made private by default. Getting them from the container explicitly is now deprecated, but not possible in Symfony 4 (which we're heading to).

### DIFF
--- a/tests/Resource/ResourceServicesTest.php
+++ b/tests/Resource/ResourceServicesTest.php
@@ -1,0 +1,43 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Tests\Resource;
+
+use Doctrine\ORM\EntityManager;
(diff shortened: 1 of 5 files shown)


lchrusciel:
Shouldn't it be opened to 1.1, as we required Symfony 3.4 as a minimal version? https://github.com/Sylius/Sylius/blob/1.1/composer.json#L61

tests/Resource/ResourceServicesTest.php
```diff
@@ -0,0 +1,43 @@
...
+final class ResourceServicesTest extends WebTestCase
+{
+    /**
+     * @test
+     */
+    public function it_allows_to_access_resource_services_from_container(): void
+    {
+        $client = parent::createClient();
+
+        $productRepository = $client->getContainer()->get('sylius.repository.product');
```
pamil: Can we make this tests ResourceBundle specific? We don't need the whole Sylius app to test public services.

pamil:
Thanks Mati! 🎉

PR #9308, opened 2018-04-06
[Adjustment] Inject adjustment types that shall be cleared

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Due to the fact, that cleared adjustments types has been set as static array, after adding new adjustment type (that is also processed during checkout), it was necessary to implement whole new class (with very similar/the same implementation, but different static array) and override the service definition. With this change, only definition overriding is needed.

### DIFF
--- a/src/Sylius/Component/Core/OrderProcessing/OrderAdjustmentsClearer.php
+++ b/src/Sylius/Component/Core/OrderProcessing/OrderAdjustmentsClearer.php
@@ -22,20 +22,19 @@ final class OrderAdjustmentsClearer implements OrderProcessorInterface
     /**
      * @var array
      */
-    private static $adjustmentsToRemove = [
-        AdjustmentInterface::ORDER_ITEM_PROMOTION_ADJUSTMENT,
-        AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT,
-        AdjustmentInterface::ORDER_SHIPPING_PROMOTION_ADJUSTMENT,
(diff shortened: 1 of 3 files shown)


Zales0123:
My bad, they've just vanished magically 😄 ⭐️

pamil:
IIRC we can get this info by calling `func_num_args()`, it will be `0` if there are no arguments passed.

```
function foo(array $arg = []) {
  var_dump(func_num_args());
}

foo(); // 0
foo([]); // 1
```

We also need to trigger user deprecation error there, just as Symfony does it:

```
@trigger_error('Pass adjustment types explicitly, ok?', E_USER_DEPRECATED);
```

src/Sylius/Component/Core/OrderProcessing/OrderAdjustmentsClearer.php
```diff
@@ -26,6 +26,18 @@
 
     public function __construct(array $adjustmentsToRemove = [])
     {
+        if (0 === func_num_args()) {
+            @trigger_error('Adjustment types should be passed explicitly', E_USER_DEPRECATED);
```
pamil: We also might want to mention that not passing them is deprecated since 1.2 and will be prohibited by 2.0

pamil:
Yay, first deprecation in Sylius, thank you Mateusz! 🥇

Zales0123:
I'm very happy I could provide that, I hope we will be able to have as many deprecations as possible xD

PR #9312, opened 2018-04-06
Fix JS error when autocomplete field is empty

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

We should be good to go for browser compatibility because `Array.prototype.filter()` has been supported since IE 9: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter#Browser_compatibility and jQuery 3.x (which we're using) only supports IE 9+: https://jquery.com/browser-support/

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-auto-complete.js
+++ b/src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-auto-complete.js
@@ -52,7 +52,7 @@
                         }
                     });
 
-                    if (0 < autocompleteValue.split(',').length) {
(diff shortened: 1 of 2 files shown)


pamil:
Looks like Symfony 3.4.x update broke the build, can you rebase with current `1.0` branch (it was fixed by #9322).

teohhanhui:
Rebased.

pamil:
Thank you, Teoh! 🥇

PR #9321, opened 2018-04-09
removing an unnecessary hidden div

axzx:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT


Zales0123:
I'm not an UI guy but... does it indeed looks better "after"? 😄

pamil:
I feel like "after" screenshot looks worse than "before".

axzx:
:)
now is better

Zales0123:
What is your resolution and browser? Because this is how it looks for me:

**Before**
before

**After**
after

Issue #9294, opened 2018-04-02
[RFC] Update ChannelNotFoundException to LocaleNotFoundException format


diimpp:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | yes
| Sylius version   | 1.x.y

[`ChannelNotFoundException`](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Channel/Context/ChannelNotFoundException.php) accepts as constructor only another exception, while [`LocaleNotFoundException`](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Locale/Context/LocaleNotFoundException.php), [`CurrencyNotFoundException`](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Currency/Context/CurrencyNotFoundException.php) allows for custom message. 

Would be cool to upgrade `ChannelNotFoundException` to same format. For example, I've a custom shop-api, which have channel defined in URL and there're multiple cases why this exception could be thrown.

pamil:
Could you provide a PR for 1.2 (master)?

PR #9324, opened 2018-04-11
ChannelNotFoundException updated to other exceptions' format without BC break

bartoszpietrzak1994:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | yes 
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/9294
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Channel/Context/ChannelNotFoundException.php
+++ b/src/Sylius/Component/Channel/Context/ChannelNotFoundException.php
@@ -15,23 +15,22 @@ namespace Sylius\Component\Channel\Context;
 
 class ChannelNotFoundException extends \RuntimeException
 {
-    private const EXCEPTION_MESSAGE = 'Channel could not be found! Tip: You can use the Web Debug Toolbar to switch between channels in development.';
-
     /**
      * {@inheritdoc}
      */
     public function __construct($messageOrPreviousException = null, ?\Throwable $previousException = null)
     {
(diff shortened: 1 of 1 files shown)


src/Sylius/Component/Channel/Context/ChannelNotFoundException.php
```diff
@@ -15,11 +15,23 @@
 
 class ChannelNotFoundException extends \RuntimeException
 {
+    private static const EXCEPTION_MESSAGE = 'Channel could not be found! Tip: You can use the Web Debug Toolbar to switch between channels in development.';
+
     /**
      * {@inheritdoc}
      */
-    public function __construct(?\Exception $previousException = null)
+    public function __construct($messageOrPreviousException = null, ?\Throwable $previousException = null)
```
pamil: WDYT about the following implementation:

```php
$message = 'Default message';

if ($messageOrPreviousException instanceof \Throwable) {
  @trigger_error('Passing previous exception as the first argument is deprecated since 1.2 and will be prohibited since 2.0.', E_USER_DEPRECATED);

  $previousException = $messageOrPreviousException;
}

if (is_string($messageOrPreviousException)) {
  $message = $messageOrPreviousException;
}

parent::__construct($message, 0, $previousException);
```

It removes the need for triple parent constructor call and preserves only one execution path (and also trigger the deprecation error).

pamil:
Thank you, Bartosz Pietrzak! 🎉

PR #9328, opened 2018-04-12
Fix invalid YAML tag syntax

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | maybe? (to me invalid YAML syntax could be considered a bug)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

1.1 because it requires Symfony 3.4

Symfony introduced invalid YAML syntax but they've since fixed it. This removes a deprecation message. See https://github.com/symfony/symfony/pull/22913

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml
@@ -95,7 +95,7 @@ twig:
         sylius: "@sylius.context.shopper"
         sylius_base_locale: "%locale%"
         sylius_meta:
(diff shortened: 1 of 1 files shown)


pamil:
Do we need to quote this parameter (like in #9334)?

teohhanhui:
No, there's no reason to quote it. It's a YAML tag. There's something not quite right with #9334 (I'll look into it further.)

teohhanhui:
@pamil Anything else you want me to change here? Should it go into a different branch?

pamil:
It just got out of view, thanks Teoh! :)

Issue #8929, opened 2017-11-02
Overriding bundle templates via Twig namespaces is not possible


emodric:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.0.2

In Symfony, it is possible to override any bundle template by using Twig namespaces. Basically, you register an additional path for a namespace equal to the namespace of the bundle you wish to override.

However, it looks like that `Sylius\Bundle\ThemeBundle\Templating\TemplateNameParser` breaks this functionality as it assumes that for example `@SyliusShop` Twig namespace always corresponds to `SyliusShopBundle` bundle. In reality, `@SyliusShop` namespace can match any file system path as defined by `twig.paths` config.

emodric:
@pamil Any news on this? Can this be fixed for 1.1?

pamil:
@emodric sorry to keep you waiting! At the time I've written ThemeBundle I think the paths were not handled like this, so it looks like a bug caused by Twig changes in the meantime. I'd love to spend some time on it, but I'm quite limited on this matter, so I'd really appreciate if you or someone else could help with this issue. I would be happy to answer all the questions, you can also catch me on Slack :)

As for releases, we can fix it for both 1.0.x and 1.1.x branch.

emodric:
From my very limited investigation, I think that just removing `Sylius\Bundle\ThemeBundle\Templating\TemplateNameParser` altogether might be a way to go, since its functionality is already included in Symfony, by default. However, I don't know enough about Sylius as a whole to say with certainty that removing it would be a valid fix.

If community has anything to add, I'll gladly listen :)

teohhanhui:
For a possible workaround (basically just disabling `SyliusThemeBundle` if you're not using it), see https://github.com/Sylius/Sylius/issues/7491#issuecomment-378336035

pamil:
#9804 should fix this one.

PR #9329, opened 2018-04-12
Remove sylius.theme.twig.loader decorator

### LINKED ISSUE #7491: please make possible to inherit from templates instead of overwriting them
Hello, 

Sylius is a great tool, it is so easy to customize it!

One way to make this customisation process more pleasant would be to let us inherit your template and alter only required blocks instead of having to overwrite them.

it would be possible that way (for example)
- fooBarAbstract.html.twig
- fooBar.html.twig

with `fooBarAbstract.html.twig`containig all your current code
and `fooBar.html.twig` extending the abstract without defining anything, it is the one beeing calld to be rendered.

This way, if whe overwrite `fooBar.html.twig` whe can extend  `fooBarAbstract.html.twig` and change only needed blocks.

I used a similar architecture on several projects, it is quite easy to maintain/use.

teohhanhui:
Just use `twig.loader` instead. The class itself is not removed but should be deprecated in v1.2.

| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no (can only be done in 1.2)
| Related tickets | fixes #7491, #8929
| License         | MIT


teohhanhui:
Let's see if it'd break anything. :stuck_out_tongue:

teohhanhui:
Okay, this is not the right way to do it.

PR #9332, opened 2018-04-12
Remove sylius.theme.templating.name_parser decorator

### LINKED ISSUE #7491: please make possible to inherit from templates instead of overwriting them

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no (must be done in 1.2)
| Related tickets | fixes #7491, #8929
| License         | MIT

Let's try @emodric's solution in https://github.com/Sylius/Sylius/issues/8929#issuecomment-363372536

### DIFF
--- a/src/Sylius/Bundle/ThemeBundle/Resources/config/services/integrations/templating.xml
+++ b/src/Sylius/Bundle/ThemeBundle/Resources/config/services/integrations/templating.xml
@@ -27,11 +27,6 @@
             <argument type="service" id="kernel" />
         </service>
 
-        <service id="sylius.theme.templating.name_parser" class="Sylius\Bundle\ThemeBundle\Templating\TemplateNameParser" decorates="templating.name_parser" decoration-priority="256" public="false">
-            <argument type="service" id="sylius.theme.templating.name_parser.inner" />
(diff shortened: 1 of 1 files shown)


teohhanhui:
Nope.

emodric:
Damn :(

teohhanhui:
But I think it's clear that both of these decorators should be removed. What is left is for someone to fix the template loading. Lol... I don't use themes at this point (yet), so I can't justify investing time in that.

PR #9334, opened 2018-04-13
Quote usage of !!int  to remove deprecation warning

stefandoorn:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Removes some Symfony deprecation warnings.

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/routing/partial/customer.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/routing/partial/customer.yml
@@ -10,7 +10,7 @@ sylius_admin_partial_customer_latest:
             repository:
                 method: findLatest
                 arguments:
-                    - !!int $count
+                    - '!!int $count'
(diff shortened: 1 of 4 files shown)


teohhanhui:
May I know what is the Symfony issue / PR? This seems wrong from a YAML perspective?

stefandoorn:
Not sure @teohhanhui what is the related Symfony change. If you run a project locally, I found it in the deprecations in the profiler.

teohhanhui:
The correct syntax should be `!!php/int`, but I'm not sure if that's implemented in Symfony.

stefandoorn:
`vendor/symfony/symfony/src/Symfony/Component/Yaml/Inline.php` at line 727 is where the exception is. Symfony 3.4.2.

pamil:
Thank you, Stefan! 🥇

PR #9335, opened 2018-04-13
Make sure controller services are public

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.0 (it's harmless to explicitly declare controller services as public)
| Bug fix?        | no (but it removes deprecation messages since Symfony 3.4)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

Services are private by default since Symfony 3.4 (see https://symfony.com/blog/new-in-symfony-3-4-services-are-private-by-default)


pamil:
Thank you, Teoh! 🎉

PR #9355, opened 2018-04-19
State machine transitioned transition fixes

### LINKED ISSUE #9326: Already transitioned resource transition
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | yes
| Sylius version   | 1.1.*

Steps to reproduce this scenario:
1. create custom order
2. as admin1 open detail of the order above
3. open anonymous tab in browser, login as admin2 and open detail of the order above
4. as admin2 in anonymous tab click on cancel order
5. as admin1 in normal tab click on cancel order - **BadRequestHttpException - Sylius/Bundle/ResourceBundle/Controller/ResourceController.php (line 556)** 

I think this is not common use case. But we have many admins and it already happened multiple times, that admin1 and admin2 have displayed order detail at the same time without refreshing the page.

czende:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | maybe
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | fixes #9326 
| License         | MIT

As described in related ticket - it happens in order, order shipment and order payment state machine. 

As this is my first bigger PR in Sylius and I'm not PHP guru as you guys, help, feedback and RFC would be welcomed :-)

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
@@ -485,7 +485,9 @@ class ResourceController extends Controller
         }
 
         if (!$this->stateMachine->can($configuration, $resource)) {
-            throw new BadRequestHttpException();
+            $this->flashHelper->addErrorFlash($configuration, 'apply_state_machine_transition_error', $resource);
+
+            return $this->redirectHandler->redirectToReferer($configuration);
         }
 
         try {
--- a/src/Sylius/Bundle/ResourceBundle/Controller/ResourceUpdateHandler.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/ResourceUpdateHandler.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sylius\Bundle\ResourceBundle\Controller;
 
 use Doctrine\Common\Persistence\ObjectManager;
+use Sylius\Component\Resource\Exception\UpdateHandlingException;
 use Sylius\Component\Resource\Model\ResourceInterface;
 
 final class ResourceUpdateHandler implements ResourceUpdateHandlerInterface
@@ -39,8 +40,8 @@ final class ResourceUpdateHandler implements ResourceUpdateHandlerInterface
         RequestConfiguration $configuration,
         ObjectManager $manager
     ): void {
-        if ($configuration->hasStateMachine()) {
-            $this->stateMachine->apply($configuration, $resource);
+        if ($configuration->hasStateMachine() && !$this->stateMachine->apply($configuration, $resource)) {
+            throw new UpdateHandlingException();
         }
 
         $manager->flush();
--- a/src/Sylius/Bundle/ResourceBundle/Controller/StateMachine.php
+++ b/src/Sylius/Bundle/ResourceBundle/Controller/StateMachine.php
@@ -14,6 +14,8 @@ declare(strict_types=1);
 namespace Sylius\Bundle\ResourceBundle\Controller;
 
 use SM\Factory\FactoryInterface;
+use SM\SMException;
+use Sylius\Component\Resource\Exception\UpdateHandlingException;
 use Sylius\Component\Resource\Model\ResourceInterface;
 
 final class StateMachine implements StateMachineInterface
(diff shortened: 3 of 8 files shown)


src/Sylius/Bundle/ResourceBundle/Controller/StateMachineInterface.php
```diff
@@ -28,6 +29,10 @@ public function can(RequestConfiguration $configuration, ResourceInterface $reso
     /**
      * @param RequestConfiguration $configuration
      * @param ResourceInterface $resource
+     *
+     * @return bool
+     *
+     * @throws UpdateHandlingException
      */
-    public function apply(RequestConfiguration $configuration, ResourceInterface $resource): void;
+    public function apply(RequestConfiguration $configuration, ResourceInterface $resource): bool;
```
pamil: It's a BC break, we can't change return typehints once we set them 🎉

czende: Thanks for your response pamil. If we have own SMinterface, I think it has to be the same as Winzou SMinterface - apply method return typehint is bool. Or is there any reason, why our apply method has void typehint?

pamil: I guess it has never returned anything, so we added that `void` typehint to ensure that. Changing it right now would cause BC breaks for anyone reimplementing this interface in the userland code. I think instead of relying on this method returning `bool`, we can catch the exceptions outside of it and process them as a `false` :)

czende: Ok, I'll try to look at it. Thank you.

Zales0123: @czende did you find the way to avoid this BC break?

czende: @Zales0123 Not quite yet :/ I have it dirty hardcoded in my project. To be sure, I have no clue how to avoid BC break. When you look at `SM\StateMachine\StateMachineInterface` both methods `can()` and `apply()` return `bool`, so I don't know why we have `bool` and `void` in `Sylius\Bundle\ResourceBundle\Controller\StateMachineInterface`.

src/Sylius/Bundle/ResourceBundle/Controller/StateMachine.php
```diff
@@ -46,12 +48,20 @@ public function can(RequestConfiguration $configuration, ResourceInterface $reso
...
-    public function apply(RequestConfiguration $configuration, ResourceInterface $resource): void
+    public function apply(RequestConfiguration $configuration, ResourceInterface $resource): bool
     {
         if (!$configuration->hasStateMachine()) {
             throw new \InvalidArgumentException('State machine must be configured to apply transition, check your routing.');
         }
 
-        $this->stateMachineFactory->get($resource, $configuration->getStateMachineGraph())->apply($configuration->getStateMachineTransition());
+        try {
+            if (!$this->stateMachineFactory->get($resource, $configuration->getStateMachineGraph())->apply($configuration->getStateMachineTransition())) {
```
pamil: Could it be solved by using `can` method on state machine factory before we call `apply`?

src/Sylius/Bundle/ResourceBundle/Controller/ResourceUpdateHandler.php
```diff
@@ -39,8 +40,8 @@ public function handle(
         RequestConfiguration $configuration,
         ObjectManager $manager
     ): void {
-        if ($configuration->hasStateMachine()) {
-            $this->stateMachine->apply($configuration, $resource);
+        if ($configuration->hasStateMachine() && !$this->stateMachine->apply($configuration, $resource)) {
```
pamil: Why don't we call `can` here instead of `apply`?

Zales0123:
Hello @czende, what is the status of this PR? do you think you would be able to fix mentioned issues? It would good to have it merged :)

src/Sylius/Bundle/ResourceBundle/Controller/ResourceController.php
```diff
@@ -485,7 +485,9 @@ public function applyStateMachineTransitionAction(Request $request): Response
         }
 
         if (!$this->stateMachine->can($configuration, $resource)) {
-            throw new BadRequestHttpException();
```
venyii: IMO this will break the Admin API. The exception should still be thrown if `!$configuration->isHtmlRequest()` like some lines below to handle a `UpdateHandlingException`: https://github.com/Sylius/Sylius/pull/9355/files#diff-2ea8b7aa90421224fb0d96dd92b4f960L494

WDYT?

PR #9366, opened 2018-04-25
[BC BREAK] Symfony 4.0 compatibility, part #1

### LINKED ISSUE #9186: Test packages on both Symfony 3 and Symfony 4
Creating a sub-issue from #9066, to properly support Symfony 4 we also need to test packages on both Symfony 3.4 and 4.0.

pamil:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | yes
| Related tickets | related to #9062, fixes #9186, closes #9277
| License         | MIT


pamil:
If the build passes for Symfony 3.4.*, it's good to merge to start more work based on it.

src/Sylius/Behat/Resources/config/services/pages/admin/order.xml
```diff
@@ -20,6 +20,8 @@
     </parameters>
 
     <services>
+        <defaults public="true" />
```
lchrusciel: Is it needed, as all (as I suppose) services here are explicitly declared as not public?

pamil: I just added it automatically to every services file, it can be removed in the future without much effort and no BC break.

PR #9369, opened 2018-04-26
Missing 'required' class added

piermont:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Province field was not displayed as mandatory during the checkout, even thought it's required.

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/Common/Form/_address.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/Common/Form/_address.html.twig
@@ -7,7 +7,7 @@
 
 {% include '@SyliusShop/Common/Form/_countryCode.html.twig' with {'form': form.countryCode} %}
 
-<div class="province-container field" data-url="{{ path('sylius_shop_ajax_render_province_form') }}">
+<div class="province-container required field" data-url="{{ path('sylius_shop_ajax_render_province_form') }}">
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/ShopBundle/Resources/views/Common/Form/_address.html.twig
```diff
@@ -7,7 +7,7 @@
 
 {% include '@SyliusShop/Common/Form/_countryCode.html.twig' with {'form': form.countryCode} %}
 
-<div class="province-container field" data-url="{{ path('sylius_shop_ajax_render_province_form') }}">
+<div class="province-container required field" data-url="{{ path('sylius_shop_ajax_render_province_form') }}">
```
pamil: Can we make it dependent on the `form.provinceCode`'s / `form.provinceName`'s `required` option?

piermont: I agree that's a better solution, my solution can probably be applied as a template override to fix it temporarily, thanks @pamil .

pamil: I've made a bit of research and it's required only when `form.provinceCode` is defined.

PR #9370, opened 2018-04-26
[Behat] Describe 100 percent discount scenario

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Description of 100% promotion rule, which appears to not work on Sylius demo, while it works when done in Behat...

I tried it twice on a demo and received a bug notice from one of Sylius users.


lchrusciel:
Seems to be duplication of #9512

PR #9371, opened 2018-04-26
Fix sourcemaps

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

We should be fine for Node.js compatibility, since yarn doesn't support node < v4 anyway. And `for...of` with arrays has been supported since node v4: https://node.green/#ES2015-syntax-for--of-loops-with-arrays


teohhanhui:
Note that this does not completely fix the problem with sourcemaps, because all of the source files have their directory structure flattened. It could probably only be solved by passing the `base` option to `gulp.src`, which means we'll have to split up the calls based on the base directory.

teohhanhui:
And the problem with the directory structure is fixed! :smile:

pamil:
Niice, thank you a lot, Teoh! 🎉

PR #9384, opened 2018-05-02
Set "logout_on_user_change" firewall option to "true"

### LINKED ISSUE #9327: Deprecation: logout_on_user_change must be set to true on firewall config
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.x

We need to make `User` implement [`EquatableInterface`](https://github.com/symfony/symfony/blob/v3.4.8/src/Symfony/Component/Security/Core/User/EquatableInterface.php) to fix a Symfony deprecation message safely. See https://github.com/symfony/symfony/pull/23882 and https://github.com/symfony/symfony-docs/issues/8428

pamil:
Fixes #9327.

---
This change is [Reviewable](https://reviewable.io/reviews/sylius/sylius/9384)

### DIFF
--- a/src/Sylius/Component/User/Model/User.php
+++ b/src/Sylius/Component/User/Model/User.php
@@ -548,6 +548,24 @@ class User implements UserInterface
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function isEqualTo(\Symfony\Component\Security\Core\User\UserInterface $user): bool
+    {
+        if (!$user instanceof UserInterface) {
+            return false;
+        }
+
+        return $this->getPassword() === $user->getPassword()
+            && $this->getUsername() === $user->getUsername()
+            && $this->getEmail() === $user->getEmail()
+            && $this->isEnabled() === $user->isEnabled()
+            && $this->getId() === $user->getId()
+        ;
+
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


src/Sylius/Component/User/Model/User.php
```diff
@@ -547,6 +547,24 @@ public function unserialize($serialized): void
...
+     */
+    public function isEqualTo(\Symfony\Component\Security\Core\User\UserInterface $user): bool
+    {
+        if (!$user instanceof UserInterface) {
+            return false;
+        }
+
+        return $this->getPassword() === $user->getPassword()
+            && $this->getUsername() === $user->getUsername()
+            && $this->getEmail() === $user->getEmail()
```
Zales0123: User email is never set in `ShopUserExampleFactory` (used in e.g. `I am a logged in customer` step) and because of that scenarios fail. We should either don't check it here or modify factory to set email on newly created shop user. The problem is, in `ShopUser` entity setting user email is [in fact setting it's customer email](https://github.com/Sylius/Sylius/blob/26ee97a41fb2add65966f16775e9dfb0349ebf6b/src/Sylius/Component/Core/Model/ShopUser.php#L68) 😄

Issue #9374, opened 2018-04-28
First shipping method is default applied even if shipping categories does not match


stefandoorn:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.0.x

We have this set-up:

* 2 shipping categories.
* 2 shipping methods (first 6.95 euro, second 40 euro). First method is tied to the first category, second method is tied to the second category.
* A single variant in the cart which is set to have category 2.

Until a shipping method is selected, the first shipping method is applied, even though that one cannot be available for this variant. So until the user get's past the shipping method page, it's always 6.95 shipping, and then on the last page it gets adjusted into 40. 

It can be reproduced on the demo:

* Add 2 shipping categories.
* Adjust the first shipping method to be tied to the first shipping category.
* Adjust the second shipping method to be tied to the second shipping category.
* Adjust a variant to be tied to the second shipping category.
* Add variant to cart

Then you see in cart the shipping costs of the first shipping method (in current setup 11.15). Continue checkout, fill in address info. Then on the shipping overview page, the shipping method which costs 11.15 is not available, but only the two others. Select one of them and proceed, then after selecting the payment method the checkout complete page shows the correct shipping price.

stefandoorn:
I think the scenario in #7178 describes the behavior as mentioned above. Let's see if it fails and does what I expect.

pamil:
Should be already solved by #9398 and #8629.

stefandoorn:
Yep, confirmed.

PR #8629, opened 2017-09-12
[Shipping][OrderProcessing] Default shipping method fixes

Zales0123:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | no |
| BC breaks?      | no |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/7965 |
| License         | MIT |
 
Two bugs where involved in this issue:
1. `DefaultShippingMethodResolver` where returning only enabled shipping methods for channel, without considering their eligibility for proper zone.
2. When the shipment was already set on `Order` it was taken confidently, without checking does it's method is eligible for current state of the order.

It was especially frustrating when shipping step was skipped.

Current solution is not perfect (even more complexity in `OrderShipmentProcessor`) but better than not working at all 😄  🐃

PR #9388, opened 2018-05-02
Add behat scenario because of wrong default shipping resolving in cart

stefandoorn:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | not yet
| New feature?    | no
| BC breaks?      | maybe
| Deprecations?   | no/yes 
| Related tickets | mentioned in #9374, related to #8629
| License         | MIT


pamil:
Should be already solved by #9398 and #8629.

PR #9389, opened 2018-05-02
[AdminBundle] Specify default _format in individual AJAX routes

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | kind of? (configuration)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

Specify default `_format` in individual AJAX routes. If it is set at a higher level, it can never be overridden per route. See https://github.com/symfony/symfony/issues/10779

This fixes `sylius_admin_ajax_render_province_form`.


pamil:
Thanks, Teoh! 🥇

PR #9419, opened 2018-05-22
Moved IE css polyfills

czende:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | no
| License         | MIT

If there are more styles included through `@SyliusUi/_stylesheets.html.twig` (for example bootstrap from cdn) it causes duplicate includes of IE 9 polyfills.

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/layout.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/layout.html.twig
@@ -13,6 +13,11 @@
     {% endblock %}
 
     {% block stylesheets %}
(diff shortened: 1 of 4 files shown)


pamil:
Thank you, Jan! 🎉

teohhanhui:
Do we still support IE 8? The jQuery version we use only supports IE 9+ though: https://jquery.com/browser-support/

teohhanhui:
Isn't it worse to have a partially working site? Lol...

PR #9405, opened 2018-05-14
Upgrade to gulp 4

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no?
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

Upgrade to [gulp 4](https://github.com/gulpjs/gulp/releases/tag/v4.0.0)

Use `gulpfile.babel.js` so that we can write in ES2015+ :tada:

/cc @pamil

PR #9441, opened 2018-06-01
Remove vendorPath command-line argument from root gulpfile

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

Since it was unnecessarily added in #9405, and not properly implemented anyway.

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -1,18 +1,11 @@
 import chug from 'gulp-chug';
 import gulp from 'gulp';
-import upath from 'path';
 import yargs from 'yargs';
 
 const { argv } = yargs
   .options({
     rootPath: {
       description: '<path> path to web assets directory',
-      type: 'string',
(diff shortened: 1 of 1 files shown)


gulpfile.babel.js
```diff
@@ -1,6 +1,5 @@
 import chug from 'gulp-chug';
 import gulp from 'gulp';
-import upath from 'path';
```
pamil: Is it the only place `upath` is used? If so, we can delete it from `package.json` as well :)

teohhanhui: No, it's used in the other `gulpfile`s :smile:

pamil:
Thank you, Teoh! 🎉

PR #9454, opened 2018-06-04
Symfony 4.1 support (together with dropping Symfony 4.0 support)

pamil:
It would make Sylius easier to maintain, so that:
 - if you prefer stable releases, stay on Symfony 3.4
 - if you prefer cutting edge solutions, upgrade directly from Symfony 3.4 to Symfony 4.1


Zales0123:
🎉

Issue #6580, opened 2016-10-28
[Admin] Bring back the Product show view


inspiran:
In the new alpha the product show view page is gone.  This is regretable because it did offer a convenient overview of  the product with all it's variant.   Some reasons why I hope it should come back.
- Authorisation levels: previously you could make a dinstinction between the Show (sales clerk) and Edit (product manager) view.  This is gone.
- The show page is also the location where you display non-editable product information.  For instance, product related statistics (eg. number of views, graphics, competition price trends).  
- Having many variants (+100) for a given product made the Show page very handy .
- I also liked having pictures next to the variants

michalmarcinkowski:
Thanks Daniel for your feedback! You presented good arguments for this feature, we will consider bringing it back.

PS. Renamed `Backend` to `Admin` in order to have a consistent vocabulary 😉

gabiudrescu:
I also vote for the arguments of @inspiran. As I have briefly tested the latest version, I found that same need.

Also, I'd add one more button: show product on website.

So far, we're struggling with upgrading from Sylius 0.18 to Sylius 0.19 and we will still rely on the old admin interface, as we did a lot of customization to it so it can better fit our needs.

But we plan that in early 2017 we migrate our instance to Sylius 1.0 and to the new admin. Then we will be able to provide a better insight from a team of 30+ people using the system - from content management people to sales and customer support.

psihius:
I'm also starting a project on Sylius, and this view was one of the things that my customer was excited beyond belief because they deal with products that have like 30-40 variants. We where looking at 0.19 at the time, and because it was 2 months till the 1.0.0 alpha and they had the hot season, we decided to wait. But we really, really, really need that view. I'm prepared to do it myself beyond the functionality it had originally (tabular input en mass - stocks, prices, discounts), but I need some guidance from the team probably to do it right. I'm going to implement it one way or the other anyway, better to contribute to the project directly.

gabiudrescu:
my opinion haven't changed since then. Admin Product Show view is still a feature we want :) I know it's easy to customize, but why not include it in core?

gabiudrescu:
up. this should not be forgotten and closed!

gabiudrescu:
Yey! :D

Issue #5846, opened 2016-08-19
Multi-channel on same host with different path


DanPhyxius:
Hi guys!

As always, thank you for your amazing work.  Today, i have a strange request for a project. :)

For a customer, our company has to set multi-channel on same host but with different path.

For example, with an e-commerce platform host ecommerce.com, our customer needs:

Store EN -> ecommerce.com/en/....
Store IT   -> ecommerce.com/it/....
Store PL  -> ecommerce.com/pl/

My question is: 
It is possible to discriminate more channel on the same host, but with different initial path?

Thank you,
Dan

DanPhyxius:
Hi all (Hi @pjedrzejewski :) ),

our company can't get this feature done, could you give us some advices on that?

We found this old issue: https://github.com/Sylius/Sylius/issues/3929
it seems that someone has used to change channel by url prefix (instead of hostname).

I'm looking for a setChannel method somewhere but i can't find it.

Some idea? :)

Thank you,
Dan

michalmarcinkowski:
Hi @DanPhyxius,

it is simple 😉  there is [request based channel context](https://github.com/michalmarcinkowski/Sylius/blob/60847f564c6f3ab63b1df94d61a3ed6cb5d5bbdc/src/Sylius/Component/Channel/Context/RequestBased/ChannelContext.php#L48-L48) where [HostnameBasedRequestResolver](https://github.com/michalmarcinkowski/Sylius/blob/4a804418869250f825dfd034f2033c2680482e5a/src/Sylius/Component/Channel/Context/RequestBased/HostnameBasedRequestResolver.php#L40-L40) is used to find channel based on hostname. All you have to do is implement your own RequestBasedResolver that will resolve the channel based on url prefix and use it instead of the default one 🎉

DanPhyxius:
Hi @michalmarcinkowski,

great answer, thank you! I'll try tomorrow morning this way! :)

DanPhyxius:
Hi @michalmarcinkowski,

it seems that override HostnameBasedRequestResolver (create our own) is not the right way.

Actually, i override sylius.context.channel.request_based.resolver.hostname_based.class parameter in my app/config/config.yml with a custom HostnameBasedRequestResolver and i put right hostname value in sylius_channel table. 

When i go to my host/subdirectory, for example:

myhost.com/fr

this resolver class is not automatically called and Sylius doesn't get right channel.

Any idea? What am i wrong? :)

Thank you for your help,
Dan

[...]

PR #9475, opened 2018-06-11
Make Stalebot less annoying

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

Stalebot is a great tool, that helps a lot with some old, deprecated issues and pull requests. However, sometimes it's quite frustrating when we have the issue that needs to be kept open, but currently, there is no time or possibility to process it (2 examples from the recent past: https://github.com/Sylius/Sylius/issues/6580, https://github.com/Sylius/Sylius/issues/5846). 
I propose to introduce a new label, that would be used in such situations. I don't know is it the best option (even the label name is a neologism, I think 😄), but could be good enough to try. The only thing we must keep in mind is not to abuse this label, so we don't wake up with dozens of "unstaleable"  issues at some point of the time.

I would love to read community's opinion about this proposal (and label name as well) :)

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -3,6 +3,7 @@ daysUntilClose: 7
 exemptLabels:
   - Bug
   - Critical
+  - Do not stale
 staleLabel: Stale
 markComment: >
   This issue has been automatically marked as stale because it has not had any


kochen:
@Zales0123 this is a nice idea, but I suggest using one of the existing tags, otherwise there might be a (i.e) BUG, but it'll still be stalled by the bot because it's not tagged with the "right" tag.

This btw, won't solve the problem with issues that don't even get the team's attention, even when they are repeatedly re-opened.

Zales0123:
@kochen if the issue would be tagged as "Bug" or "Critical" it won't be staled anyway (see the rest of the configuration file). Of course, there is always a place for oversight, but we're all just humans. It's of course very subjective which issue should be labeled that way, but we can base on a community feedback for sure.
Regarding issues that are not getting enough attention from the @Sylius/core-team. I realize it can be a problem, we're trying hard to minimalize such situations, but that's also the place where we need community really much, we can't handle it all by our own :)

stefandoorn:
+1 for having a tag (or smth) that prevents closing.

.github/stale.yml
```diff
@@ -3,6 +3,7 @@ daysUntilClose: 7
 exemptLabels:
   - Bug
   - Critical
+  - Unstaleable
```
pamil: Is this a real word? Is this just fantasy?

What do you think about `Do not close`?

Zales0123: It's totally my fantasy 😄 `Do not close` is pretty good for me, says exactly what should (not) be done 👌

kochen:
Actually, `Do not stale` is better, because that's what the bot does in the first place (only after that, it closes the issue).

Zales0123:
OK, then the winner is `Do not stale` 🏅

pamil:
Thank you, Matt! 🎉

PR #9512, opened 2018-06-20
[Behat] Describe percentage promotion issue

### LINKED ISSUE #7836: [Promotions] Not longer eligible for promotion at checkout-complete with "Order total" rule

* Create a promotion coupon, configure it as "Order total" X sum and "Order discount" 50%
* Add a products, so the total cart sum is just over "X" (or at least when discount is applied, the resulting sum is bellow "X").
* Apply the coupon in the cart
* Proceed with the checkout
* When pressing "Complete order" and "complete" step - get an error "You are no longer eligible for this promotion Y"

After debugging the code, I found out that "ItemTotalRuleChecker" is checking against Order::getItemsTotal that is returning an already adjusted sum, not the item original cost, that results in the order being no longer eligible for the promotion.

System should check against original sum for items, not an adjusted one when validating eligibility.

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | covers #7836
| License         | MIT

With collaboration with @CoderMaggie and @gabiudrescu

### DIFF
--- a/features/checkout/order_promotion/order_promotion_integrity_validation.feature
+++ b/features/checkout/order_promotion/order_promotion_integrity_validation.feature
@@ -10,13 +10,13 @@ Feature: Order promotions integrity
         And the store ships everywhere for free
         And the store has a product "PHP T-Shirt" priced at "$100.00"
         And there is a promotion "Christmas sale"
-        And this promotion gives "$10.00" discount to every order
-        And this promotion expires tomorrow
         And I am a logged in customer
 
     @ui
     Scenario: Preventing customer from completing checkout with already expired promotion
-        Given I added product "PHP T-Shirt" to the cart
+        Given this promotion gives "$10.00" discount to every order
+        And this promotion expires tomorrow
+        And I added product "PHP T-Shirt" to the cart
         And I have proceeded selecting "Offline" payment method
         And this promotion has already expired
         When I confirm my order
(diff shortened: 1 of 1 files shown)


pamil:
After digging into Sylius' internals, it's caused by integrity checker. It uses promotion rule checkers, which were designed and works correctly only before applying the promotion they check. I'll create an issue about deprecating the current integrity checkers and my idea the brand new ones in the coming days.

pamil:
The idea for a new integrity checker is the following:
 1. Make a service which returns a DTO containing all the current promotions codes and names, prices per items / shipping / taxes / total etc. based on `OrderInterface`
 2. Implement an integrity checker like the current ones, which does:
     - transform the current order state into that DTO using the service mentioned before
     - process the order once again using order processor (this removes all the adjustments and apply promotions once again etc.)
     - transform the current order state into an another DTO
     - compare these DTOs, if they are the same - it's safe for the order to be placed, if not - throw an error and don't let that order to be placed

pamil:
Fixed by #9596.

pamil:
@lchrusciel could you rebase this PR?

pamil:
Modified the scenario, because the older structure has been always failing because steps weren't implemented correctly.

pamil:
Thank you, Łukasz, Magda and Gabi! 🎉

PR #9541, opened 2018-07-01
Require Symfony 4.1.1 and remove hotfixes for 4.1.0

pamil:


composer.json
```diff
@@ -58,7 +58,7 @@
         "symfony/polyfill-intl-icu": "^1.3",
         "symfony/polyfill-mbstring": "^1.3",
         "symfony/swiftmailer-bundle": "^3.0",
-        "symfony/symfony": "^3.4|^4.1",
+        "symfony/symfony": "^3.4|^4.1.1",
```
lchrusciel: Shouldn't we just conflict `4.1.0`?

pamil: This is the same IIRC.

stefandoorn: Does the same I think, but the difference is the 'communication'. With the conflict it's clear to everyone that some incompatibility occured.

pamil: @stefandoorn agreed! This would require a bit of extra work, I wouldn't mind just accepting a PR that does it :)

stefandoorn: I wouldn't mind helping on that, but that would be earliest next week somewhere. You can add an issue on my name if you want.

PR #9553, opened 2018-07-04
Fix wrong type for "images" option in ProductFixture

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

It wasn't actually working with `ProductExampleFactory`, but the tests don't catch it.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
@@ -359,13 +359,21 @@ class ProductExampleFactory extends AbstractExampleFactory implements ExampleFac
     private function createImages(ProductInterface $product, array $options): void
     {
         foreach ($options['images'] as $image) {
-            $imagePath = array_shift($image);
+            // BC, to be deprecated in 1.3 and removed in 2.0
+            if (!array_key_exists('path', $image)) {
+                $imagePath = array_shift($image);
+                $imageType = array_pop($image);
+            } else {
+                $imagePath = $image['path'];
+                $imageType = $image['type'] ?? null;
+            }
+
             $uploadedImage = new UploadedFile($imagePath, basename($imagePath));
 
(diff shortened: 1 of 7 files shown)


teohhanhui:
The best to me would be:

```yaml
images:
    main: 'path1.jpg'
    thumbnail: 'path2.jpg'
```

teohhanhui:
But yeah, it'd be a problem when we want to leave the type empty. So I'll go with your second option.

pamil:
Yeah, that was my first thought too, but the type is nullable and is not unique.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
```diff
@@ -359,13 +359,13 @@ private function createChannelPricings(ProductVariantInterface $productVariant,
     private function createImages(ProductInterface $product, array $options): void
     {
         foreach ($options['images'] as $image) {
-            $imagePath = array_shift($image);
+            $imagePath = array_key_exists('path', $image) ? $image['path'] : array_shift($image);
```
teohhanhui: @pamil Should we actually `@trigger_error` here?

teohhanhui: We could add the deprecation in another PR targeting `master`, WDYT?

pamil: Ah, that's right indeed!

teohhanhui: Done.

pamil:
Thanks, Teoh! 🥇

pamil:
Created #9581 for that deprecation addition.

PR #9562, opened 2018-07-06
Make percentage discount promotions great again

### LINKED ISSUE #7836: [Promotions] Not longer eligible for promotion at checkout-complete with "Order total" rule

pamil:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | closes #9512, fixes #7836
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Setup/PromotionContext.php
+++ b/src/Sylius/Behat/Context/Setup/PromotionContext.php
@@ -348,6 +348,19 @@ final class PromotionContext implements Context
     }
 
     /**
+     * @Given /^([^"]+) gives ("[^"]+%") discount to every order with items total at least ("[^"]+")$/
+     */
+    public function itGivesPercentageDiscountToEveryOrderWithItemsTotalAtLeast(
+        PromotionInterface $promotion,
+        $discount,
+        $targetAmount
+    ) {
+        $channelCode = $this->sharedStorage->get('channel')->getCode();
+        $rule = $this->ruleFactory->createItemTotal($channelCode, $targetAmount);
+        $this->createPercentagePromotion($promotion, $discount, [], $rule);
+    }
+
+    /**
      * @Given /^([^"]+) gives ("[^"]+%") off on every product when the item total is at least ("(?:€|£|\$)[^"]+")$/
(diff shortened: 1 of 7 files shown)


src/Sylius/Behat/Context/Setup/PromotionContext.php
```diff
@@ -347,6 +347,19 @@ public function itGivesFixedDiscountToEveryOrderWithItemsTotalAtLeast(
...
     }
 
+    /**
+     * @Given /^([^"]+) gives ("[^"]+%") discount to every order with items total at least ("[^"]+")$/
+     */
+    public function itGivesPercentageDiscountToEveryOrderWithItemsTotalAtLeast(
+        PromotionInterface $promotion,
+        $discount,
+        $targetAmount
+    ) {
```
Zales0123: What about strict types?

src/Sylius/Behat/Resources/config/suites/ui/checkout/checkout_order_promotion.yml
```diff
@@ -0,0 +1,63 @@
+# This file is part of the Sylius package.
+# (c) Paweł Jędrzejewski
+
+default:
+    suites:
+        ui_checkout:
```
Zales0123: `ui_checkout_order_promotion`?

pamil:
It looks like it would require much more extra work than this, closing for now.

PR #9563, opened 2018-07-06
[AdminApi] Fix command

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminApiBundle/Command/CreateClientCommand.php
+++ b/src/Sylius/Bundle/AdminApiBundle/Command/CreateClientCommand.php
@@ -20,8 +20,23 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
+/**
+ * @deprecated Fetching dependencies directly from container is not recommended from Symfony 3.4. Extending `ContainerAwareCommand` will be removed in 2.0
+ */
 final class CreateClientCommand extends ContainerAwareCommand
 {
+    /**
+     * @var ClientManagerInterface
+     */
+    private $clientManager;
+
+    public function __construct(?string $name = null, ClientManagerInterface $clientManager = null)
+    {
+        parent::__construct($name);
+
+        $this->clientManager = $clientManager;
+    }
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/AdminApiBundle/Command/CreateClientCommand.php
```diff
@@ -15,13 +15,25 @@
...
 
-final class CreateClientCommand extends ContainerAwareCommand
+final class CreateClientCommand extends Command
 {
+    /**
+     * @var ClientManagerInterface
+     */
+    private $clientManager;
+
+    public function __construct(string $name = null, ClientManagerInterface $clientManager)
```
pamil: Unfortunately, adding a mandatory argument to the constructor method is considered a BC break.

lchrusciel: I can add a separate class and override service definition in the configuration if current Symfony version is 4.1. 

This class will be deprecated then.

lchrusciel: What about this solution?

src/Sylius/Bundle/AdminApiBundle/Resources/config/services.xml
```diff
@@ -23,6 +23,8 @@
         <defaults public="true" />
 
         <service id="Sylius\Bundle\AdminApiBundle\Command\CreateClientCommand">
+            <argument />
```
pamil: This passes an empty string now.

pamil: `<argument>null</argument>` is the solution for passing null.

lchrusciel: Won't it just pass `null` as a string?

pamil: Nope, I've checked it myself, writing an integration test for that would be 11/10 though :)

src/Sylius/Bundle/AdminApiBundle/Resources/config/services.xml
```diff
@@ -23,6 +23,8 @@
         <defaults public="true" />
 
         <service id="Sylius\Bundle\AdminApiBundle\Command\CreateClientCommand">
+            <argument on-invalid="null"/>
```
pamil: It still passes an empty string as command name which causes `InvalidArgumentException` to be thrown. What was wrong with `<argument>null</argument>`?

pamil:
Thank you, Łukasz! 🎉

PR #9578, opened 2018-07-10
[Core] Nullable customer on order

### LINKED ISSUE #9577: [Core] [BC-Break] Not nullable customer on order
**Sylius version affected**: 1.1.8

**Description**  
BC-Break introduced in https://github.com/Sylius/Sylius/pull/9552/commits/261c614f0e38d0660693cca4c30751376194434c#diff-fb2a1dbd12848847780522681726f61dR129. We should still allow for null value here.

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   |
| Related tickets | fixes #9577 
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Core/Model/Order.php
+++ b/src/Sylius/Component/Core/Model/Order.php
@@ -126,7 +126,7 @@ class Order extends BaseOrder implements OrderInterface
      */
     public function setCustomer(?BaseCustomerInterface $customer): void
     {
-        Assert::isInstanceOf($customer, CustomerInterface::class);
+        Assert::nullOrisInstanceOf($customer, CustomerInterface::class);
 
         $this->customer = $customer;
     }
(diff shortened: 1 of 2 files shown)


pamil:
Could you target `1.1` with this fix?

lchrusciel:
It was meant for 1.1. Sorry 🙈

pamil:
Thank you, Łukasz! 🥇

teohhanhui:
The intention is to allow attaching orders to non-customer resources, right?

lchrusciel:
It was a BC breaking change against 1.0, so we had to revert it. By default, it is not possible to assign a null as a customer during checkout in the default checkout process. However, you can customise it and then it is possible.

teohhanhui:
Yes, I understand that. But what I'm curious about is whether attaching orders to something other than a customer or a user (say, a company) is something you've thought about and whether that's designed for.

PR #9605, opened 2018-07-20
Allowing fields comparison in grid filters

Roshyo:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Adding a support to fields comparison. So we can filter for example an entity related to an other one.
There is still the PHPCRODM to verify (marked as TODO: ) since I don't know if that works

@pamil @CoderMaggie  : I believe it's the right branch this time (master for features :) )

### DIFF
--- a/src/Sylius/Component/Grid/Data/ExpressionBuilderInterface.php
+++ b/src/Sylius/Component/Grid/Data/ExpressionBuilderInterface.php
@@ -56,6 +56,22 @@ interface ExpressionBuilderInterface
 
     /**
      * @param string $field
+     * @param string $secondField
+     *
+     * @return mixed
+     */
+    public function fieldsEquals(string $field, string $secondField);
+
+    /**
+     * @param string $field
+     * @param string $secondField
+     *
+     * @return mixed
+     */
+    public function fieldsNotEquals(string $field, string $secondField);
+
+    /**
+     * @param string $field
      * @param mixed $value
      *
      * @return mixed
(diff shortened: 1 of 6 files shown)


src/Sylius/Component/Grid/Data/ExpressionBuilderInterface.php
```diff
@@ -54,6 +54,22 @@ public function equals(string $field, $value);
      */
     public function notEquals(string $field, $value);
 
+    /**
+     * @param string $field
+     * @param string $secondField
+     *
+     * @return mixed
+     */
+    public function fieldsEquals(string $field, string $secondField);
```
pamil: Adding a method to an interface is a BC break unfortunately :/

Roshyo: Oh, indeed... What would be the best to do it then ?

Roshyo:
Do not stale ?

How to proceed ? do we have to wait until 2.0 for this feature ?

vvasiloi:
Isn't it possible to implement this use-case with a custom filter and `Sylius\Bundle\GridBundle\Doctrine\ORM\ExpressionBuilder::comparison`?

Zales0123:
Hello Stephane, it's been a while. In the meantime, **SyliusGridBundle** was extracted to the [separate repository](https://github.com/Sylius/SyliusGridBundle/), so I suppose it should be implemented there. Thank you! 🖖

PR #9607, opened 2018-07-21
Add alias to enable autowire of SenderInterface

lfjeff:
Add alias so dependency injection of SenderInterface will work with autowire

| Q               | A
| --------------- | -----
| Branch?         | 1.0, 1.1 or master 
| Bug fix?        | no/yes
| New feature?    | no/yes
| BC breaks?      | no/yes
| Deprecations?   | no/yes 
| License         | MIT


src/Sylius/Bundle/MailerBundle/Resources/config/services.xml
```diff
@@ -32,6 +32,9 @@
             <argument>%sylius.mailer.sender_name%</argument>
             <argument>%sylius.mailer.sender_address%</argument>
         </service>
+        
+        <service id="sylius.email_sender.alias" alias="sylius.email_sender" public="true" />
```
pamil: I guess this line is not needed.

src/Sylius/Bundle/MailerBundle/Resources/config/services.xml
```diff
@@ -32,6 +32,9 @@
...
+        <service id="Sylius\Component\Mailer\Sender\SenderInterface" alias="sylius.email_sender.alias" />
```
pamil: Changing `sylius.email_sender.alias` to `sylius.email_sender` should be enough.

PR #9608, opened 2018-07-21
Remove instances of loop.index0

dnna:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

When using Sylius with [PHP-PM](https://github.com/php-pm/php-pm), it cannot be guaranteed that the elements of an ArrayCollection will have continuous indexes (for example if you add two items to the cart, then remove one then add a third one the final items collection will have indexes 0 and 2). This sometimes breaks Sylius in Twig templates where loop.index0 is used because it may not match the current item index.

This PR fixes the issue by using the actual index within the loop instead of loop.index0.

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/Cart/Summary/_items.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/Cart/Summary/_items.html.twig
@@ -16,8 +16,8 @@
             </tr>
             </thead>
             <tbody>
-                {% for index,item in cart.items %}
-                    {% include '@SyliusShop/Cart/Summary/_item.html.twig' with {'item': item, 'form': form.items[index]} %}
+                {% for key, item in cart.items %}
+                    {% include '@SyliusShop/Cart/Summary/_item.html.twig' with {'item': item, 'form': form.items[key]} %}
                 {% endfor %}
             </tbody>
         </table>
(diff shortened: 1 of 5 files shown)


src/Sylius/Bundle/ShopBundle/Resources/views/Cart/Summary/_items.html.twig
```diff
@@ -16,8 +16,8 @@
             </tr>
             </thead>
             <tbody>
-                {% for item in cart.items %}
-                    {% include '@SyliusShop/Cart/Summary/_item.html.twig' with {'item': item, 'form': form.items[loop.index0]} %}
+                {% for index,item in cart.items %}
```
lchrusciel: I'm not sure about `index` keyword here, as it can be confused with `loop.index` in long-term. Perhaps `key` would be a better word? Ref: https://twig.symfony.com/doc/2.x/tags/for.html#iterating-over-keys-and-values

pamil: 👍 for using `key` and adding a space after comma would be great 🎉

dnna:
@lchrusciel @pamil comments have been addressed

pamil:
Thank you, Dimosthenis! 🥇

PR #9610, opened 2018-07-23
[Resource] Fix deprecation referencing controllers with a single colon

dannyvw:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | 
| License         | MIT

Fix "User Deprecated: Referencing controllers with a single colon is deprecated since Symfony 4.1. Use app.controller.user::indexAction instead."

### DIFF
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -22,7 +22,7 @@ sylius_shop_default_locale:
     path: /
     methods: [GET]
     defaults:
-        _controller: sylius.controller.shop.locale_switch:switchAction
+        _controller: sylius.controller.shop.locale_switch::switchAction
 
 _liip_imagine:
     resource: "@LiipImagineBundle/Resources/config/routing.yaml"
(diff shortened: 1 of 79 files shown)


dannyvw:
@pamil This change fails for symfony 3.4. Any idea how we can solve this?

pamil:
We need to support Symfony 3.4 at least until Symfony 4.4 is released, I've got no idea how to fix it now other than waiting for dropping Symfony 3.4 support in Sylius.

teohhanhui:
Symfony 4.4 is scheduled for November 2019 (!)

I think we have a problem if issues like this keep piling up.

pamil:
Since this PR couldn't be merged in a foreseeable future (and the work itself will be needed to be done again then), I'm closing it and creating #9784 to keep track of it. Thanks for mentioning it, Danny!

emodric:
@pamil One alternative is to use single action controllers with `__invoke` magic method. That way you don't need to provide the method name, only the FQCN or the service ID. And it works on Symfony 3.4 and even Symfony 2.8.

It is still a breaking change, but maybe others will benefit from the info :)

PR #9613, opened 2018-07-24
Fix ShopBasedCartContext resetting

dnna:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

ShopBasedCartContext keeps an instance of the cart internally. When using [PHP-PM](https://github.com/php-pm/php-pm), this attribute needs to be reset at the end of each request otherwise unexpected behaviors happen. This solution uses the kernel.reset tag introduced specifically for this purpose in https://github.com/symfony/symfony/issues/23984

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services/context.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services/context.xml
@@ -33,6 +33,7 @@
         <service id="sylius.context.cart.new_shop_based" class="Sylius\Component\Core\Cart\Context\ShopBasedCartContext" decorates="sylius.context.cart.new" decoration-priority="256" public="false">
             <argument type="service" id="sylius.context.cart.new_shop_based.inner" />
             <argument type="service" id="sylius.context.shopper" />
+            <tag name="kernel.reset" method="reset" />
         </service>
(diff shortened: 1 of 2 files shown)


pamil:
Thank you, Dimosthenis! 🎉 Don't hesitate to share your findings about running Sylius in PHP-PM :)

pamil:
Although it could you some quick unit test, I'll try to prepare one.

dnna:
Thank you! I wasn't sure how to unit test this part. So far it seems to be working great and fast with PHP-PM after this and https://github.com/Sylius/Sylius/pull/9608 :) It will go for user testing in the next few weeks, so I'll be happy to make PR for any other issues we find running it this way.

pamil:
Awesome to hear that! Added some tests for that new method in #9617.

czende:
@dnna Are you using PHP-PM with or without Docker? I've done some testing with nginx setup and it seems to be mind-blowing!

dnna:
@czende I'm using it with Docker. The performance gain is pretty insane, but there are unfortunately some hard-to-fix leaks (see https://github.com/Sylius/Sylius/issues/9629) especially in the admin that cause inconsistent behavior. The shop front-end seems quite stable and I haven't noticed any issues there since merging this PR. All in all I'd say its not production ready yet, but its closer than I expected!

PR #9615, opened 2018-07-25
Simplify code of `sylius-product-images-preview` module

### LINKED ISSUE #9595: Complicated error prone previewUploadedImage plugin
Hi. I was adding images to one of my entities and found this: https://github.com/Sylius/Sylius/blob/281d33c1c7b3b83e9ac420e7be917e4632cb2593/src/Sylius/Bundle/AdminBundle/Resources/private/js/sylius-product-images-preview.js#L32

```js
$.fn.extend({
  previewUploadedImage(root) {
    $(`${root} input[type="file"]`).each((idx, el) => {
      $(el).change((event) => {
        displayUploadedImage(event.currentTarget);
      });
    });

    $(`${root} [data-form-collection="add"]`).on('click', (evt) => {
      const self = $(evt.currentTarget);

      setTimeout(() => {
        self.parent().find('.column:last-child input[type="file"]').on('change', (event) => {
          displayUploadedImage(event.currentTarget);
        });
      }, 500);
    });
  },
});
```

So as I understand it, if somebody adds new form part via add button, function is fired after 500ms that adds change handler on the file input.

What if there will be 2 inputs in 500ms? (last input will have 2 handlers and input before that none).
What if user removes that input in 500ms? (last input will have 2 handlers).

Why not write it just like:
```js
$.fn.extend({
  previewUploadedImage(root) {
    $(root).on('change', 'input[type="file"]', function() {
      displayUploadedImage(this);
    });
  },
});
```

? It seems to be simpler and less error prone.

nenadalm:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | in rare cases
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9595
| License         | MIT

`$el.on(event, selector, handler)` attaches event handler to the `$el` element. If event is triggered on the `$el`, it is checked that element that triggered the event (child of `$el`) matches given `selector`. In case the element matches the `selector`, `handler` is called - otherwise nothing happens.

So it's not necessary to attach event handler for each element - it's enough to attach it to some parent and specify `selector` of childs for which the `handler` should be called. See https://api.jquery.com/on/#on-events-selector-data-handler for details.

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/private/js/sylius-product-images-preview.js
+++ b/src/Sylius/Bundle/AdminBundle/Resources/private/js/sylius-product-images-preview.js
@@ -31,20 +31,8 @@ const displayUploadedImage = function displayUploadedImage(input) {
 
 $.fn.extend({
   previewUploadedImage(root) {
-    $(`${root} input[type="file"]`).each((idx, el) => {
-      $(el).change((event) => {
-        displayUploadedImage(event.currentTarget);
-      });
-    });
-
-    $(`${root} [data-form-collection="add"]`).on('click', (evt) => {
-      const self = $(evt.currentTarget);
-
-      setTimeout(() => {
-        self.parent().find('.column:last-child input[type="file"]').on('change', (event) => {
-          displayUploadedImage(event.currentTarget);
-        });
-      }, 500);
+    $(root).on('change', 'input[type="file"]', function() {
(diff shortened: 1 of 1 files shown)


pamil:
Thank you, Miloslav! 🥇

PR #9618, opened 2018-07-26
Reproduce CSRF token validation failure when deleting an user in admin panel

### LINKED ISSUE #8936: adminbundle : invalid csrf token , when trying to delete a customer
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.0.1

I am using the default templates and no custom code here , yet i still get this csrf token error.

screenshot from 2017-11-06 14-28-49

pamil:
Fixes #8936.


features/user/managing_users/deleting_account.feature
```diff
@@ -5,7 +5,8 @@ Feature: Deleting the customer account
     I want to be able to remove customer account details from the system
 
     Background:
-        Given there is a user "theodore@example.com" identified by "pswd"
+        Given the store has customer "customer@example.com"
```
pamil: We generate CSRF token based on customer ID instead of user ID. Before this change, user ID and customer ID were the same, so it was passing.

Issue #9594, opened 2018-07-17
Migrations incompatible with MySQL 8 


pamil:
**Sylius version affected**: 1.0.0 - forward

**Description**  
Our migrations fails on MySQL 8.

**Steps to reproduce**  
Run migrations (or just `bin/console sylius:install`) with MySQL 8.

**Possible Solution**  
1) Add MySQL 8 to the testing pipeline (more jobs, yay)
2) Fix the CI build :)

pamil:
Fixed by #9622.

PR #9622, opened 2018-07-30
Quote row_number identifier for MySQL queries

alcaeus:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 0
| License         | MIT

When running Migrations on MySQL 8, one query causes an SQL error:
> SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'row_number FROM sylius_product INNER JOIN sylius_product' at line 2

The name has to be properly quoted for MySQL 8 where `ROW_NUMBER` is a reserved identifier.

### DIFF
--- a/app/migrations/Version20161223091334.php
+++ b/app/migrations/Version20161223091334.php
@@ -17,7 +17,7 @@ class Version20161223091334 extends AbstractMigration
         $this->addSql('ALTER TABLE sylius_product_variant ADD position INT NOT NULL');
         $this->addSql('SET @row_number = -1');
         $this->addSql('CREATE TEMPORARY TABLE IF NOT EXISTS variants_count
-                        SELECT sylius_product.id AS product_id, COUNT(sylius_product_variant.id) AS row_number FROM sylius_product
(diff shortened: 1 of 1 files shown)


pamil:
Fixes #9594, thanks! 🎉

alcaeus:
Ah, didn't find the issue in question. I'll see if I can easily add MySQL 8 to the build pipeline.

pamil:
I'm afraid adding MySQL 8 to the build pipeline would make the whole feedback cycle way too long. I've tested it manually and migrations works correctly on that database. Anyway, I'll merge this PR, as it solves the problem, and we can think how to stop regressions later on. Thanks, Andreas! 🥇

stefandoorn:
@pamil What about combining https://docs.travis-ci.com/user/customizing-the-build/#matching-jobs-with-allow_failures & https://docs.travis-ci.com/user/customizing-the-build/#fast-finishing? It's still extra, but it also guarantees higher project quality to test against multiple dependencies.

PR #9625, opened 2018-07-30
Remove automatic database environment suffix

pamil:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

People new to Sylius don't usually know that we add a suffix to the database name (`your_configured_name` + `dev/test/staging` [or none for production]). Also, while working on a Docker environment for Sylius-Standard, it also caused a small issue that needs some workaround.

Therefore I propose to remove these suffixes for Sylius 1.3 to give people full control of their database name by default. It is a backwards compatible change, as it does not change the behaviour for existing projects.

### DIFF
--- a/app/config/config_dev.yml
+++ b/app/config/config_dev.yml
@@ -8,10 +8,6 @@ framework:
     router: { resource: "%kernel.project_dir%/app/config/routing_dev.yml" }
     profiler: { only_exceptions: false }
 
-doctrine:
-    dbal:
-        dbname: "%database_name%_dev"
-
 monolog:
     handlers:
         main:
--- a/app/config/config_staging.yml
+++ b/app/config/config_staging.yml
@@ -15,9 +15,5 @@ monolog:
             path: "%kernel.logs_dir%/%kernel.environment%.log"
             level: debug
 
-doctrine:
-    dbal:
-        dbname: "%database_name%_staging"
-
 swiftmailer:
     disable_delivery: true
--- a/app/config/config_test.yml
+++ b/app/config/config_test.yml
@@ -15,10 +15,6 @@ framework:
     session:
         storage_id: session.storage.mock_file
 
-doctrine:
-    dbal:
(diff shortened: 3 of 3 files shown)


lchrusciel:
I'm against this change. 

I'm used to this, and I feel it is super useful. Once we change it, I need to either move quickly to docker or adjust it for each future project. Wasn't it a default Symfony behaviour?

pamil:
@lchrusciel that was never the default Symfony behaviour. Basically, it's possible to bring back the previous behaviour by changing `%database_name%` to `database_%kernel.environment%` (the production environment would also include a suffix then).

gabiudrescu:
I'm also against this, as this brings running doctrine:database:drop in production one step closer. 

what are the issues that appear in docker?

pamil:
@lchrusciel @loic425 @gabiudrescu I understand your concerns, what do you think about database name always being `sylius_ENVIRONMENT`? The only change for projects starting from 1.3 would be the production environment database named `sylius_prod` instead of `sylius` and `test_cached` environment database named `sylius_test_cached` instead of `sylius_test` (which should not really affect 3rd party apps).

This would allow us to connect to the database using [an URL](https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url), meaning one parameter, instead of 6 like it is now.

loic425:
@pamil If combining database url and automatic database_name depends on environment is possible, I totally agree!

gabiudrescu:
@pamil right now it is based on the environment.

for example, in config.yml the name of the database is `%database_name%`, while in config_dev.yml is `'%database_name%dev`

but i get what you are proposing and that's a good idea :)

pamil:
Cool, closing this one then and it will be changed together with Symfony 4 directory structure and environmental variables.

PR #9627, opened 2018-08-03
Narrow down selectors to prevent unexpected bugs

### LINKED ISSUE #9438: Problem with UiBundle javascript, creates white overlay when it shouldn't
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.1.6

There is a bug with the way your admin javascript works. Whenever you hit the "Create" (or "Save changes" if you're editing) button, a semitransparent white overlay with a 'loading' icon (spinner) appears while the request is sent and the new response comes (ie, showing the next page, be it the index page, the update page... whatever you configure when you add the admin routing with the ResourceBundle).

My problem comes when I try to use https://github.com/danielantelo/AddressableBundle within Sylius. This bundle provides a new form type called AddressMapType which consists in a map (you can move the marker to get an address) and a text field with a "Search" button to geocode the search string into an actual address as well as longitude+latitude if you wish instead of using the map itself. Then you can fine-tune the fields for street name, street number, city, country and zip code.

When I click that search button in the AddressMapType, the white overlay + spinner are shown. I can see the AddressMapType is geocoding correctly, finishes its work, but that overlay never goes away.

Tracing the error I've seen that the `form.loadable` selector gets a .loading css class whenever I use any `<button>`. And doing a global search of `.loadable` within the Sylius project, I got where the error is: UiBundle/Resources/private/js/app.js
So, I think we should narrow the selector in UiBundle javascript to only affect the submit buttons, leaving alone plain buttons (non submit). This works for this specific bundle but not sure for others.
The offending line is this:
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/UiBundle/Resources/private/js/app.js#L30
Changing it with this works for me.
```js
$('form.loadable button[type=submit]').on('click', function() {
      return $(this).closest('form').addClass('loading');
    });
```
But again, not sure if that's enough or if we should try to see other use cases.

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9438
| License         | MIT

/cc @migmolrod


pamil:
Rebased the PR to see if it's passing with the most recent version of the `1.1` branch.

PR #9631, opened 2018-08-07
Fix controller notation (single colon is deprecated)

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes?
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

Using the single colon (`service:method`) notation for controllers has been deprecated since Symfony 4.1. See https://github.com/symfony/symfony/pull/26085

Currently we have hundreds of deprecation errors on the first page load because of this.

### DIFF
--- a/app/config/routing.yml
+++ b/app/config/routing.yml
@@ -22,7 +22,7 @@ sylius_shop_default_locale:
     path: /
     methods: [GET]
     defaults:
-        _controller: sylius.controller.shop.locale_switch:switchAction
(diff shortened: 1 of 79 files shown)


pamil:
Related to #9610, the issue is that we still need to support Symfony 3.4.

teohhanhui:
Sorry. Didn't notice that.

PR #9643, opened 2018-08-09
Symfony 4 directory structure

pamil:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | yes 
| Related tickets | -
| License         | MIT

This PR switches Sylius to Symfony 4 directory structure. The PR introducing Symfony Flex in Sylius will follow shortly 🎉


.env.dist
```diff
@@ -0,0 +1,23 @@
...
+APP_ENV=dev
+APP_DEBUG=1
+APP_SECRET=EDITME
+###< symfony/framework-bundle ###
+
+###> doctrine/doctrine-bundle ###
+# Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url
+# For a sqlite database, use: "sqlite:///%kernel.project_dir%/var/data.db"
+# Set "serverVersion" to your server version to avoid edge-case exceptions and extra database calls
+DATABASE_URL=mysql://root@127.0.0.1/sylius_${APP_ENV}?serverVersion=5.5
```
teohhanhui: `serverVersion` should not be set here. See https://github.com/symfony/recipes/blob/cb6ea10b7bb60c4bd9c39ca3e92fb53e7ccce7e5/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml#L12

teohhanhui: https://github.com/symfony/recipes/pull/258

composer.json
```diff
@@ -191,7 +185,7 @@
         "psr-4": {
             "Sylius\\Tests\\": "tests/"
         },
-        "classmap": ["app/AppKernel.php", "app/AppCache.php", "app/TestAppKernel.php"]
+        "classmap": ["src/Kernel.php"]
     },
     "config": {
         "bin-dir": "bin"
```
teohhanhui: This should be removed.

composer.json
```diff
@@ -203,9 +197,6 @@
         "symfony-web-dir": "web",
```
teohhanhui: This is wrong.

pamil: It'll be removed after adding Symfony Flex and removing SensioDistributionBundle in the next PR, I'll change it to `public` before that.

teohhanhui: But why in a separate PR?

pamil: This one is already +500 -800, the diff is getting unreadable and I want to make tests passing before proceeding further. It's `master` so there's still over one and a half month until the next release.

pamil: Nevertheless, Symfony 4 config structure and Symfony Flex are two different changes, that could go separately.

src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml
```diff
@@ -23,6 +23,9 @@ imports:
     - { resource: "@SyliusCoreBundle/Resources/config/app/state_machine.yml" }
     - { resource: "@SyliusCoreBundle/Resources/config/app/fixtures.yml" }
 
+parameters:
+    sylius_core.public_dir: "%kernel.project_dir%/web"
```
teohhanhui: This is of course wrong. But also, is it really necessary for this to be configurable?

pamil: This is introduced only to make the changes backwards compatible - so that it won't fail if you upgrade from 1.2 to 1.3. It is overwritten in `config/_sylius.yaml` to `%kernel.project_dir%/public`.

pamil: What do you mean by introducing more config files? `config/packages/_sylius.yaml` follows the Symfony 4 standard of having config files for each package. We can't remove the config files in bundles in `1.x` branch. As for configs in Sylius-Standard and PluginSkeleton and their synchronisation, this is a whole different issue to be handled and I believe doing it all at once (Symfony 4 directory structure, Symfony Flex, config files synchronisation) would result in not having anything that works (and if it does, it would probably result in some nasty BC breaks).

pamil:
Merging it for now, if there's any change needed, we can still revert it or customise before v1.3.0 release in late September.

Issue #9365, opened 2018-04-25
Product removed from taxon if taxon not shown


blankcheck:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | from 1.0.0-beta2 up

Steps to reproduce (I did it on my older install and on current demo):

1. Go to edit product
2. Have at least one taxon with children, i.e.:

 - taxon 1
 - taxon 2
 - taxon 3
    -- taxon 3.1
    -- taxon 3.2

Add product to taxon 1 and taxon 3.1.
Without expanding taxon 3, add product to taxon 2
As a consequence, product is removed from taxon 3.1

I guess reason being that symfony form doesn't now anything about taxon 3.1 until it was fetched (in current sylius via ajax).

blankcheck:
Just verified that this is bug indeed, here is root cause and initial sub-optimal solution:

In this file https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/AdminBundle/Resources/private/js/sylius-lazy-choice-tree.js , function assigned to variable `bindCheckboxAction` basically goes through checkboxes and builds array of **selected** taxons (later that is saved as comma separated string, but not important for bug itself). Of course, taxons which haven't been fetched (because they are 2,3 levels deep) will not be part of selected taxons and will be detached from product. My initial solution to plug the hole:

Override choiceTree method and replace bindCheckboxAction

```
            var bindCheckboxAction = function (checkboxElement) {
                checkboxElement.checkbox({
                    onChange: function () {
                        var $checkboxes = tree.find('.checkbox');
                        var checkedValues = $input.val().split(',');

                        $checkboxes.each(function () {
                            var checkboxValue = $(this).data('value');
                            if ($(this).checkbox('is checked')) {
                                checkedValues.push(checkboxValue);
                            } else {
                                checkedValues = $.grep(checkedValues, function(value) {
                                    return value != checkboxValue;
                                });
                            }
                        });

                        $input.val(checkedValues.join());
                    }
                });
            };


````
This repeats taxons and likely js code should filter against that, but also that fact shouldn't affect taxon assignments as such.

pamil:
Fixed by #9646. 🎉

PR #9646, opened 2018-08-13
[Admin][Product edit] Change the value of the taxons individually when checked/unchecked.

sbarbat:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | [#9365](https://github.com/Sylius/Sylius/issues/9365)
| License         | MIT

Before this fix was checking which checkbox were checked but is loading the tree in lazy mode, so if you didn't load some part of the tree where you had taxons checked, those taxons were lost when saving.

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/private/js/sylius-lazy-choice-tree.js
+++ b/src/Sylius/Bundle/AdminBundle/Resources/private/js/sylius-lazy-choice-tree.js
@@ -155,13 +155,13 @@
                 checkboxElement.checkbox({
                     onChecked: function() {
                         var value = checkboxElement.data('value');
-                        var checkedValues = $input.val().split(",");
+                        var checkedValues = $input.val().split(",").filter(x => x);
                         checkedValues.push(value);
                         $input.val(checkedValues.join());
                     },
                     onUnchecked: function() {
                         var value = checkboxElement.data('value');
-                        var checkedValues = $input.val().split(",");
+                        var checkedValues = $input.val().split(",").filter(x => x);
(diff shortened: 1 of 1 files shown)


pamil:
Thank you, Santiago! 🎉

tchapi:
Hi,

I'm running into a very peculiar issue with Sylius 1.1 and I _think_ this commit is related.

Sylius 1.1 uses **Uglify** (In this Gulpfile here for instance : https://github.com/Sylius/Sylius/blob/1.1/src/Sylius/Bundle/AdminBundle/Gulpfile.js).

The problem is that Uglify does not support ES6, and arrow functions are an ES6 notation.
This commit introduces 2 arrow functions ([here](https://github.com/Sylius/Sylius/pull/9646/files#diff-f5e80a9df49d2db624723f3713c4cd16R158) and [here](https://github.com/Sylius/Sylius/pull/9646/files#diff-f5e80a9df49d2db624723f3713c4cd16R164)) and thus, breaks `npm run gulp` for the 1.1 branch. This is the only file in the 1.1 branch where the arrow notation is used AFAIK

Maybe I've missed something ?

If I'm right, is there a fix for 1.1, or a quick workaround you can think about ? 
I'd gladly update to Sylius 1.2 but I can't for now due to other requirements.

Thanks a lot!

sbarbat:
You are probably right. Maybe you can change the x => x to function(x) { return x;} and see if it works 😊

tchapi:
Yes it does work with a "regular" function, hence my comment. I'll try to make a PR

PR #9660, opened 2018-08-20
[Docs] Fixes services file path

mmenozzi:
Because probably the customization goes into the end application and the proper location for the service config file is `app/config/services.yml`. Moreover if the extension is registered in `AppBundle/Resources/config/services.yml` it doesn't work if that file is not loaded through the AppBundle DI extension (https://symfony.com/doc/current/bundles/extension.html).


pamil:
Thank you, Manuele! 🥇

mmenozzi:
You're welcome!

mmenozzi:
@pamil why did you merged only into `1.1` branch? Isn't this change related also to the `1.2` version?

pamil:
`1.1` branch is then merged into `1.2` and then into `master` :)

PR #9662, opened 2018-08-21
Fix Grids archive ability to work on multi-param urls

diimpp:
| Q               | A
| --------------- | -----
| Branch?         | 1.0, 1.1, 1.2, master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | no
| License         | MIT

Hi,

I was adding `ArchivableInterface` to my `ProductVariant`, when I've noticed, that `archive.html.twig` can't work on urls with multiple parameters, like `productId`, `id`, e.g. `/admin/products/206/variants/195`

Solution is copy-pasted from `delete.html.twig`, which doesn't have such problem. https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/UiBundle/Resources/views/Grid/Action/delete.html.twig

Works for me on `shipping-methods` and custom `product/product-variants` views.

Let me know, if something is unclear.

Cheers.


diimpp:
Does anybody know what the heck is going with all this mass-commits from regular online PR? Is it because that I already had clone of the project? How do I avoid it?

pamil:
Thank you, Dmitri! 🥇

PR #9664, opened 2018-08-22
Replace private by protected to allow override.

SteadEXE:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | None
| License         | MIT

I submit this PR because if you want to override the 'add' or 'remove' action you need to copy all the private methods.


pamil:
Could you let us know what is your use-case that requires overriding those add/remove actions?

PR #9665, opened 2018-08-22
[WIP] Introduce Symfony Flex

pamil:
TODO:
 - [ ] Tests passing
 - [ ] Testing both with Symfony 3.4 and 4.1 on Travis CI

PR #9672, opened 2018-08-24
Provide a BC layer for files in "app/config/" referenced by PluginSkeleton

pamil:
Files `app/config/routing.yml`, `app/config/routing_dev.yml` and `app/config/security.yml` are referenced in PluginSkeleton. This PR provides a backwards compatibility layer for them.

### DIFF
--- a/app/config/_container_deprecation.php
+++ b/app/config/_container_deprecation.php
@@ -15,6 +15,4 @@
 
 declare(strict_types=1);
 
-@trigger_error('The "AppKernel" class located at "app/AppKernel.php" is deprecated since Sylius 1.3. Use "Kernel" class located at "src/Kernel.php" instead.', E_USER_DEPRECATED);
-
-class_alias(Kernel::class, AppKernel::class);
+@trigger_error('Importing files from Sylius/Sylius\'s "app/config" directory is deprecated since Sylius 1.3.', \E_USER_DEPRECATED);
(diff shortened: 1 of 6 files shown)


app/config/parameters.yml.dist
```diff
@@ -1,24 +0,0 @@
-# This file is auto-generated during the composer install
```
pamil: This file is not referenced by neither Sylius-Standard nor PluginSkeleton.

app/config/_container_deprecation.php
```diff
@@ -0,0 +1,18 @@
...
+ * file that was distributed with this source code.
+ */
+
+/**
+ * @internal
+ */
+
+declare(strict_types=1);
+
+@trigger_error('Importing files from Sylius/Sylius\'s "app/config" directory is deprecated since Sylius 1.3.', \E_USER_DEPRECATED);
```
pamil: Deprecating YAML files is kinda weird 🎉

app/config/parameters.yml.dist
pamil: Ah, we still need it here, it is being removed in #9665.

lchrusciel:
Thanks Kamil!

pamil:
@teohhanhui I'm trying to solve that by making Sylius-Standard a part of Sylius/Sylius (which is subtree split back into Sylius-Standard). This way we can run tests on this repo while having one less place with configuration.

teohhanhui:
Yes, make them part of the application, not part of Sylius. Sylius-Standard should be treated purely as a distribution, aka recommended scaffolding / project template. No BC guarantees.

PR #9683, opened 2018-08-29
Do not run tests if not needed

pamil:
PRs containing only docs changes shouldn't trigger test run, same for the ones only modifying a readme.


pamil:
It looks like it's indeed working, but let's manually check the next PRs' builds for a day or two to see whether it does not skip too much :)

PR #9684, opened 2018-08-29
Deprecate `setState` methods

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets |
| License         | MIT

State machine should be the only possible way to manage the entity state. Because of the existence of `setState` methods, we could do such abominations as [this step in Behat context](https://github.com/Sylius/Sylius/blob/b509d978834e1d356e7041b584da3bc85b671a1c/src/Sylius/Behat/Context/Setup/OrderContext.php#L423).

### DIFF
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -35,3 +35,11 @@
     
         * `sylius.mongodb_odm.repository.class`
         * `sylius.phpcr_odm.repository.class`
+
+* All methods for setting entity states (associated with state machine) has been removed.
+Only state machine should be used to manage entity state:
+
+    * `Sylius\Component\Order\Model\OrderInterface::setState`
+    * `Sylius\Component\Payment\Model\PaymentInterface::setState`
+    * `Sylius\Component\Shipping\Model\ShipmentInterface::setState`
+    * `Sylius\Component\Review\Model\ReviewInterface::setStatus`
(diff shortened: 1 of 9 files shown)


src/Sylius/Component/Order/Model/OrderInterface.php
```diff
@@ -114,6 +114,7 @@ public function getState(): string;
 
     /**
      * @param string $state
+     * @deprecated this method is deprecated since Sylius 1.3 and will be removed in Sylius 2.0, use state machine to manage order state
```
pamil: We need to trigger deprecation errors in our implementations of these methods.

Zales0123: Done 🎉

src/Sylius/Component/Order/Model/Order.php
```diff
@@ -271,6 +271,11 @@ public function getState(): string
      */
     public function setState(string $state): void
     {
+        @trigger_error(
```
pamil: `setState` is used by state machine itself so I'm not sure we can deprecate it now (unless we provide another way for the state machine to change it).

UPGRADE-2.0.md
```diff
@@ -35,3 +35,11 @@
     
         * `sylius.mongodb_odm.repository.class`
         * `sylius.phpcr_odm.repository.class`
+
+* All methods for setting entity states (associated with state machine) has been removed.
```
pjedrzejewski: has -> have

PR #9142, opened 2018-02-01
[Channel] [Checkout] Added configurable shipping countries (#9122)

JakobTolkemit:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | mentioned in #9122 
| License         | MIT

PR #9697, opened 2018-09-07
[Channel] [Checkout] Configurable shipping countries

### LINKED ISSUE #9122: Countries available in the checkout made configurable per channel

| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.2

Currently, if we have 2 channels "US Webstore" and "EU Webstore", during the checkout, all countries are available. An improvement would be to add a relation between every Channel and many Countries. And if Channel has like 2 out of 5 countries selected, these are available during checkout. If no countries are selected - all are available. So we retain the current behavior (no BC break) but add new, useful feature.

To achieve this, I would suggest to modify the AddressType form type. In the CheckoutAddressType form type we could identify the channel of the order and use appropriate countries as options to choose from.

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9122, replaces https://github.com/Sylius/Sylius/pull/9142
| License         | MIT

This is rebased and improved version of an awesome PR by @JakobTolkemit. The main change is a decoration of `ShippingMethodEligibilityChecker` in _Core_, to respect shippable countries (e.g. while providing a default shipping method).


features/checkout/addressing_order/choosing_country_on_a_channel_with_shipping_countries.feature
```diff
@@ -0,0 +1,26 @@
+@checkout
+Feature: Choosing a country from a set of shipping countries
+    In order to avoid selecting an invalid country during the addressing checkout step step
+    As a Customer
+    I want to only be able to choose from countries that are shippable
+
+    Background:
+        Given the store operates on a single channel in the "United States" named "Webstore"
+        And the store operates in "Germany" and "France"
```
pamil: What do you think about: the store **also** operates in "Germany" and "France"?

src/Sylius/Component/Core/Resolver/ShippableCountriesResolver.php
```diff
@@ -0,0 +1,54 @@
...
+     * @return CountryInterface[]
+     */
+    public function __invoke(ChannelInterface $channel): array
+    {
+        $countries = $channel->getShippableCountries()->filter(function (CountryInterface $country): bool {
+            return $country->isEnabled();
+        });
+
+        if ($countries->isEmpty()) {
+            $countries = new ArrayCollection($this->countryRepository->findBy(['enabled' => true]));
```
pamil: If a channel has assigned only disabled countries, then I wouldn't expect it automatically means all enabled countries.

src/Sylius/Component/Core/Checker/ShippingMethodEligibilityChecker.php
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        $channel = $subject->getOrder()->getChannel();
+
+        $shippableCountries = ($this->shippableCountriesResolver)($channel);
+        $zoneCountries = $this->zoneCountriesProvider->getCountriesInWhichZoneOperates($method->getZone());
+
+        foreach ($zoneCountries as $zoneCountry) {
+            if (!in_array($zoneCountry, $shippableCountries)) {
+                return false;
```
pamil: Given the following setup:
  - Channel A with shippable countries Poland and Germany
  - Shipping method B for zone "All around the world" which contains all countries

Shipping method B would always be not eligible.

Shouldn't it only consider whether the currently selected shipping country is shippable inside the channel and inside the zone of the shipping method?

pamil:
@Zales0123 could you remind all of us what our discussion was about this feature? I remember we have backed off this one for some reason, but I couldn't remember what exactly it was :)

Zales0123:
As far as I remember, the main reason why we've postponed it was possible BC breaks and uncertainty how to resolve default shipping method properly, taking care about configurable shipping countries. I'll go back to this PR in the nearest days and try to figure it out :) 🐎

CoderMaggie:
@Zales0123 any news here? :)

pamil:
Yeah, there was another conflict with shipping method resolution, as they rely on zones, so after nailing it down to them, you also had to filter out the countries.

In my opinion, we should leverage zones from shipping methods to limit the country and province choices for shipping address. If that's not an option, we still should assign zones to the channel instead of countries as it allows for the greater flexibility.

mamazu:
Still looking forward to it. Later we should also expose this via the API.

PR #9698, opened 2018-09-08
Fix grid events to fire on every request in PHP-PM

dnna:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

When using Sylius with [PHP-PM](https://github.com/php-pm/php-pm), the `ArrayGridProvider` service generates a `$grids` array internally. With PHP-PM, this occurs only for the first request, then the array is cached. This means events like `sylius.grid.admin_*` only fire once and cannot be used to modify the grid dynamically.

This PR fixes this by refactoring `ArrayGridProvider` a bit and resetting the grids state at the end of every request.

### DIFF
--- a/src/Sylius/Component/Grid/Provider/ArrayGridProvider.php
+++ b/src/Sylius/Component/Grid/Provider/ArrayGridProvider.php
@@ -20,6 +20,16 @@ use Sylius\Component\Grid\Exception\UndefinedGridException;
 final class ArrayGridProvider implements GridProviderInterface
 {
     /**
+     * @var ArrayToDefinitionConverterInterface
+     */
+    private $converter;
+
+    /**
+     * @var array
+     */
+    private $gridConfigurations;
+
+    /**
      * @var Grid[]
      */
     private $grids = [];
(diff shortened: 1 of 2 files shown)


src/Sylius/Component/Grid/Provider/ArrayGridProvider.php
```diff
@@ -30,20 +40,35 @@
...
     public function __construct(ArrayToDefinitionConverterInterface $converter, array $gridConfigurations)
     {
-        foreach ($gridConfigurations as $code => $gridConfiguration) {
-            if (isset($gridConfiguration['extends'], $gridConfigurations[$gridConfiguration['extends']])) {
-                $gridConfiguration = $this->extend($gridConfiguration, $gridConfigurations[$gridConfiguration['extends']]);
+        $this->converter = $converter;
+        $this->gridConfigurations = $gridConfigurations;
+    }
+
+    private function convertGrids()
```
lchrusciel: Missing return types

pamil: Btw. could you move it below public and protected methods?

src/Sylius/Component/Grid/Provider/ArrayGridProvider.php
```diff
@@ -30,20 +40,35 @@
...
+            if (isset($gridConfiguration['extends'], $this->gridConfigurations[$gridConfiguration['extends']])) {
+                $gridConfiguration = $this->extend($gridConfiguration, $this->gridConfigurations[$gridConfiguration['extends']]);
             }
 
-            $this->grids[$code] = $converter->convert($code, $gridConfiguration);
+            $this->grids[$code] = $this->converter->convert($code, $gridConfiguration);
         }
     }
 
+    public function reset()
```
pamil: We can also add a `: void` typehint here :)

dnna: Done :)

dnna:
@pamil @lchrusciel I have made the requested changes.

@lchrusciel actually I have been thinking about how to best test PHP-PM, for this project and others as well. So far the pattern I have noticed is that almost all bugs with PHP-PM are arising from the fact that some services save state internally, assuming it will be destroyed at the end of the request. In fact Sylius is one of the projects not doing it very frequently, which is why it works pretty well.

I think a good way to catch this is with static analysis - maybe phpstan could detect when services are modifying their internal state. I will look into making a contribution there, but I cannot promise about the timeframe. In the meantime I will make PRs here with the issues we are noticing in production :)

dnna:
FYI I have created https://github.com/phpstan/phpstan-symfony/pull/11, which should catch most of the cases where PHP-PM could break in Sylius. It does not catch this specific issue (we would need a separate rule for this, for example detecting when events are getting dispatched from DI constructors), but it did catch issues like https://github.com/Sylius/Sylius/pull/9613 in my tests.

All in all there are about 15 services in Sylius that got flagged by the new rule, but this includes services like CookieStorage which may be OK to modify themselves. I will go through the list at some point and see which are relevant.

I'm hoping this rule will get accepted in the main phpstan-symfony, otherwise I'll look into creating a third party ruleset for PHP-PM specific checks.

pamil:
@dnna I've created #9757 with an alternative solution, let me know what you think! :)

dnna:
@pamil nice solution! I did a quick test and it seems to work, so I think this is the way to go :)

pamil:
Fixed by #9757, thanks Dimosthenis for bringing it up and initial implementation! 🎉

PR #9705, opened 2018-09-12
Update winzou/state-machine-bundle

### LINKED ISSUE #9510: Lacking debug:winzou
**Sylius version affected**: v1.2.0

**Description**  

I docs

> http://docs.sylius.com/en/latest/cookbook/shop/checkout.html

There is presented command:

    php bin/console debug:winzou:state-machine

but applying this give error

```                     
  There are no commands defined in the "debug:winzou" namespace.  
                                                                  
  Did you mean one of these?                                      
      debug                                                       
      debug:payum 
```

**Steps to reproduce**  

Install Sylius in Ubuntu using composer

> https://sylius.com/download/

Use command

    php bin/console debug:winzou:state-machine

> https://sylius.com/download/

**Possible Solution**  

Update docs or dependencies.

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no?
| Deprecations?   | no
| Related tickets | fixes #9510
| License         | MIT

### DIFF
--- a/docs/book/architecture/state_machine.rst
+++ b/docs/book/architecture/state_machine.rst
@@ -45,7 +45,7 @@ Having states configured we can have a transition between the ``cart`` state to 
        sylius_order_checkout:
            transitions:
                address:
-                   from: [cart, addressed, shipping_selected, payment_selected]  # here you specify which state is the initial 
+                   from: [cart, addressed, shipping_selected, payment_selected]  # here you specify which state is the initial
                    to: addressed                                                 # there you specify which state is final for that transition
 
 Callbacks
@@ -121,7 +121,7 @@ would become ``shipping_selected``.
                        args: ["object"]
                    sylius_create_order:
                        on: ["complete"]
-                       do: ["@sm.callback.cascade_transition", "apply"]
+                       do: ["@SM\\Callback\\CascadeTransitionCallback", "apply"]
                        args: ["object", "event", "'create'", "'sylius_order'"]
                    sylius_hold_inventory:
                        on: ["complete"]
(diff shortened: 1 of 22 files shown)


lchrusciel:
But we need to make the `SM\Factory\FactoryInterface` service public in order to make it work :(

teohhanhui:
It's made public in the compiler pass. In order to avoid that, we'd have to stop injecting the container directly into the test classes.

teohhanhui:
```
  [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]
  The "SM\Factory\FactoryInterface" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.
```

Uhh... I wonder why it's not working...

docs/book/architecture/state_machine.rst
```diff
@@ -121,7 +121,7 @@ would become ``shipping_selected``.
                        args: ["object"]
                    sylius_create_order:
                        on: ["complete"]
-                       do: ["@sm.callback.cascade_transition", "apply"]
+                       do: ["@SM\\Callback\\CascadeTransitionCallback", "apply"]
```
pamil: If I understand correctly, `sm.callback.cascade_transition` is removed in the `0.4` release of that bundle, which might break the existing Sylius-based application in case one have customised the state machine configuration / written their own configuration for it.

src/Sylius/Component/Core/Updater/UnpaidOrdersStateUpdater.php
```diff
@@ -13,7 +13,7 @@
 
 namespace Sylius\Component\Core\Updater;
 
-use SM\Factory\Factory;
+use SM\Factory\FactoryInterface;
```
pamil: This change could make it into Sylius as a bugfix to `1.1` branch.

docs/book/architecture/state_machine.rst
pamil: Indeed!

artggd:
Any news on this @teohhanhui? It would be really nice to have that feature back and not wasting the effort you already put on this PR 🙂

PR #9716, opened 2018-09-19
Move symfony/flex dependency to require-dev

pamil:
The ultimate goal is to get rid of the application from Sylius/Sylius without sacrificing the DX while contributing. However, we should allow one not to use Symfony Flex while using Sylius.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,6 @@
         "symfony/expression-language": "^3.4|^4.1",
         "symfony/filesystem": "^3.4|^4.1",
         "symfony/finder": "^3.4|^4.1",
-        "symfony/flex": "^1.1",
         "symfony/form": "^3.4|^4.1",
         "symfony/framework-bundle": "^3.4|^4.1",
         "symfony/http-foundation": "^3.4|^4.1",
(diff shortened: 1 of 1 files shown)


lchrusciel:
Thanks Kamil!

markahesketh:
Quick question: When I do a `composer install --no-dev` during a deployment to Dokku, I am running into the following error when `@auto-scripts` runs post-install as `symfony/flex` is not installed:

```
remote:        > symfony-cmd
remote:        sh: 1: symfony-cmd: not found
remote:        Script symfony-cmd handling the auto-scripts event returned with error code 127
remote:        Script @auto-scripts was called via post-install-cmd
```

Is this expected?

lchrusciel:
Your deployment should not be related to Sylius/Sylius itself. If you are creating a new app, then you should use Sylius/Sylius-Standard where symfony/flex should be added I suppose (and it's not atm).

As a result, you should require flex in your end app and can you open an issue for it in Sylius-Standard repository?

markahesketh:
Thanks, I'll take a look - but this was a project created using `composer create-project sylius/sylius-standard acme` as I was trying out 1.3

lchrusciel:
Sure, it is our bug. We should introduce flex to sylius-standard probably, or change installation instruction to reflect flex usage :D /cc @Sylius/core-team

PR #9720, opened 2018-09-21
Adding a reference to all entities in the override doc

Roshyo:
It's not practical to have to search for where the entity has been defined in the first place. And especially the first time you override sylius_user.resources.shop.user, it's not an obvious path. The same is for ChannelPricings.
So this could be really helpful. I'll try to keep it up to date with my further overridings.

| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | doc
| License         | MIT

### DIFF
--- a/docs/customization/model.rst
+++ b/docs/customization/model.rst
@@ -389,6 +389,36 @@ Which we strongly recommend over updating the schema.
     $ php bin/console doctrine:migrations:diff
     $ php bin/console doctrine:migrations:migrate
 
+List of all core models and how to override them :
+--------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


pamil:
Thank you for the PR! I understand the motivation behind it and we are working on providing Sylius-Standard with all core models already overwritten (see https://github.com/Sylius/Sylius-Standard/pull/265). This way the documentation will not be outdated, will also contain the examples for translated models and dynamic models (like ShopUser) and the implementation itself. I hope that solves the tension.

PR #9712, opened 2018-09-14
Fix wrong template path syntax

teohhanhui:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

Stop using legacy template paths, as it doesn't work well with template overriding through the `templates` directory.

PR #9726, opened 2018-09-24
Use native Twig references for templates

### LINKED ISSUE #9719: Symfony flex structure : templates override
**Sylius version affected**: 1.3-DEV

**Description**  

When you use new symfony flex with sylius, you can't override templates using this method
https://symfony.com/doc/current/bundles/override.html#templates

For example if you want to rewrite SyliusAdminBundle/Channel/_form.html.twig, you should be able to put your file in templates/bundles/SyliusAdminBundle/Channel/_form.html.twig

**Possible Solution**  
This issue and proposed fix is relevant #8929 but not totally.
See https://github.com/symfony/symfony/issues/25661
```yaml
templates:
  form: SyliusAdminBundle:Channel:_form.html.twig
```
should be
```yaml
templates:
  form: @SyliusAdmin/Channel/_form.html.twig
```

wadjeroudi:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | yes
| Deprecations?   | no 
| Related tickets | fixes #9719
| License         | MIT


src/Sylius/Bundle/GridBundle/Tests/DependencyInjection/SyliusGridExtensionTest.php
```diff
@@ -83,15 +83,15 @@ public function it_sets_filter_templates_as_parameters(): void
         $this->load([
             'templates' => [
                 'filter' => [
-                    'string' => 'AppBundle:Grid/Filter:string.html.twig',
-                    'date' => 'AppBundle:Grid/Filter:date.html.twig',
+                    'string' => 'Grid/Filter/string.html.twig',
```
pamil: It's a test, but I guess there's `@` missing in the beginning.

wadjeroudi: It was on purpose, you can't use @App. Not totally sure if it's the correct value.

pamil: Ah yeah, you're right! :)

pamil:
LGTM apart from the answer for the "BC breaks?" question, what exactly is a BC break here?

wadjeroudi:
@pamil do you know behat suite didn't pass anymore, can't spot the mistaske ?
For the BC break, it's because of the new structure with this syntax, not sure if it does have side effects to be handled between old structure projects and new flex structure.

src/Sylius/Bundle/ShopBundle/Resources/views/Checkout/complete.html.twig
```diff
@@ -20,7 +20,7 @@
 
             {{ sonata_block_render_event('sylius.shop.checkout.complete.before_summary', {'order': order, 'form': form}) }}
 
-            {% include 'SyliusShopBundle:Common/Order:_summary.html.twig' %}
+            {% include '@SyliusShop/Common/Order/' %}
```
pamil: Something went wrong here :)

wadjeroudi:
- Thank you @pamil. Green test.

- I didn't see this duplicate PR #9712.
@teohhanhui chose to replace the docs too, not sure about that.
Sylius 1.3 should still supports symfony 3.4 and old structure ?

teohhanhui:
I don't see any problem with using the "new" syntax in the docs. In fact, it's not new and already supported since Twig 2.0?

src/Sylius/Bundle/AdminBundle/Resources/config/routing/admin_user.yml
```diff
@@ -3,7 +3,7 @@ sylius_admin_admin_user:
         alias: sylius.admin_user
         section: admin
         path: users
-        templates: SyliusAdminBundle:Crud
+        templates: "@SyliusAdmin\\Crud"
```
teohhanhui: I thought it should be a forward slash, not a backslash?

wadjeroudi:
@teohhanhui I'm not opposed, just maybe if we replace, we should add a documentation that explains clearly the two structures and how to override (even if it's a common symfony practice) ?

pamil:
Thank you, @wadjeroudi and @teohhanhui! 🎉

PR #9735, opened 2018-09-27
Make Flex recipes to be based on Sylius possible

pamil:


Zales0123:
I think that after removing `sensio-distribution-bundle` we have also removed `sensiolabs/security-checker` which provides `security:check` command used in Travis build :(

pamil:
@Zales0123 could you add that package to the `require-dev`? I'll be able to do that at least at 5pm, the faster the better now :)

Zales0123:
You're reading my mind @pamil! 😄 https://github.com/Sylius/Sylius/pull/9736

PR #9747, opened 2018-10-01
Add failing test for showing correct shipping method fees after shipping discount applied

stefandoorn:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | not yet
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/features/shipping/applying_shipping_fee/seeing_correct_shipping_fee_after_discount.feature
+++ b/features/shipping/applying_shipping_fee/seeing_correct_shipping_fee_after_discount.feature
@@ -1,0 +1,30 @@
+@receiving_discount
+Feature: Receiving percentage discount on shipping
+  In order to pay decreased amount for shipping
+  As a Customer
+  I want to see the correct fees for the shipping methods available
+
+  Background:
+    Given the store operates on a single channel in "United States"
+    And the store has "DHL" shipping method with "$10.00" fee
+    And the store has a product "PHP T-Shirt" priced at "$100.00"
+    And there is a promotion "Holiday promotion"
+    And I am a logged in customer
+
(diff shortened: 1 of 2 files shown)


pamil:
The scenario is now failing because of missing steps, are you willing to implement them or reuse the existing ones to make this PR include a really failing scenario?

stefandoorn:
Done, @pamil.

stefandoorn:
@pamil @Zales0123 I'm not sure what's the best fix here, could you have a look?

Zales0123:
Hmmm, it's a good question 😄 Write now displayed shipping cost is calculated [in this file](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ShippingBundle/Form/Type/ShippingMethodChoiceType.php#L109). I think the best option would be to either to extend this form type in _Core_ and calculate costs one more time including discount or provide a new shipping calculator (also in _Core_) that would include these discounts as well... The easiest option I see here is just to display a message "These are full prices of shipping, go to the end of the checkout to see how much your shipping would actually cost" but it seems like a not so good customer experience ⚖️ 🏎

lchrusciel:
Related to #10813

lchrusciel:
Hello Stefan 👋 

I've checked this description, and the solution for this problem is a custom shipping calculator, which will allow for free delivery above some threshold. We are not able to recalculate the price of shipping with the promotion applied. This PR will go along with the #10813 request. Would you (or @loevgaard or @igormukhingmailcom) like to contribute to this new calculator? Behat tests are almost ready in this PR. We just need to change one step here.

PR #9746, opened 2018-10-01
[Documentation] v1.3 Update

CoderMaggie:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Doc fix?        | yes
| Related tickets | 
| License         | MIT

First partial update of docs for the v1.3

PR #6901, opened 2016-11-28
[Currency] [Critical] Channel based currencies support

lchrusciel:
| Q               | A
| --------------- | ---
| Bug fix?        | yes |
| New feature?    | yes |
| BC breaks?      | yes |
| Related tickets | fixes #6641  |
| License         | MIT |

Rebased and fixed version of #6868

Contains also #6896

PR #9752, opened 2018-10-02
Update installation.rst for Flex

dunglas:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | mentioned in #9746
| License         | MIT

Complementary to #9746, both should be merged.

### DIFF
--- a/docs/book/installation/installation.rst
+++ b/docs/book/installation/installation.rst
@@ -34,10 +34,8 @@ To create a new project using Sylius Standard Edition, run this command:
 
     Make sure to use PHP ^7.1. Using an older PHP version will result in installing an older version of Sylius.
 
-This will create a new Symfony project in ``acme`` directory. When all the
-dependencies are installed, you'll be asked to fill the ``parameters.yml``
-file via interactive script. Please follow the steps. After everything is in
-place, run the following commands:
+This will create a new Symfony project in ``acme`` directory.
+After everything is in place, run the following commands:
 
 .. code-block:: bash
 
@@ -49,7 +47,8 @@ This package has the whole ``sylius/sylius`` package in vendors, so you can easi
 .. warning::
 
     During the ``sylius:install`` command you will be asked to provide important information, but also its execution ensures
-    that the default **currency** (USD) and the default **locale** (English - US) are set they can be later on changed in the ``parameters.yml`` file.
+    that the default **currency** (USD) and the default **locale** (English - US) are set.
+    They can be changed later, respespectively in the "Configuration > Channels" section of the admin and in the ``app/config/services.yaml`` file.
     From now on all the prices will be stored in the database in USD as integers, and all the products will have to be added with a base american english name translation.
 
 Installing assets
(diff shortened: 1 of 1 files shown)


docs/book/installation/installation.rst
```diff
@@ -49,7 +47,7 @@ This package has the whole ``sylius/sylius`` package in vendors, so you can easi
 .. warning::
 
     During the ``sylius:install`` command you will be asked to provide important information, but also its execution ensures
-    that the default **currency** (USD) and the default **locale** (English - US) are set they can be later on changed in the ``parameters.yml`` file.
+    that the default **currency** (USD) and the default **locale** (English - US) are set they can be later on changed.
```
CoderMaggie: they're now stored in the `config/services.yaml` file, I think. Am I right @pamil ?

dunglas: On my local fresh install there is only the `locale` (as in any Symfony standard), but I cannot find the `currency` parameter.

Btw, the install seems to currently be broken (assets are not installed correctly).

CoderMaggie: what happens if you add the `currency: EUR` below the locale?

CoderMaggie: Edit: It works :)
Check via `bin/console debug:container --parameters | grep currency`

dunglas: It may work, but the currency is set somewhere (I chosen `EUR` during the install, and this currency is indeed used).

Zales0123: afaik we don't use a `currency` parameters for last two years or something :D we have now base currencies on channels

dunglas: It looks like this config is now stored in the DB (in the `sylius_currency` table). Am I right?

dunglas: @CoderMaggie there is no such parameter in the container of my app.

Zales0123: @dunglas there is no such parameter [since Nov 2016](https://github.com/Sylius/Sylius/pull/6901/commits/29bf4cfb1c220f6f192b1fb438d86ef2b863980e), it's a documentation residue, that should be removed for sure

dunglas: Got it, but the installer still ask for the locale. It may worth it to explain how to change it afterward. In the admin panel? (Sorry I’m a total newcomer to Sylius).

Zales0123: At least one currency is required for Sylius to work, it's then set as a base currency of a channel, that is also created during the installation process. No problem, asking questions is good :)

Zales0123: And I totally agree it should be explained in installation docs, instead of this legacy information about parameters 👍

CoderMaggie: :+1:

CoderMaggie:
And this PR should target the `1.3` branch :)

docs/book/installation/installation.rst
dunglas: Updated

pamil:
Thanks, Kevin! 🎉

PR #9758, opened 2018-10-02
[ThemeBundle] Fix risky tests

pamil:


src/Sylius/Bundle/ThemeBundle/phpunit.xml.dist
```diff
@@ -6,6 +6,8 @@
 >
     <php>
         <server name="KERNEL_CLASS" value="AppKernel" />
+
+        <server name="SHELL_VERBOSITY" value="-1" />
```
Zales0123: why changing the verbosity?

pamil: Without this change, it is writing a lot of logs to console output when executing tests. It's also the standard Symfony setting: https://github.com/symfony/recipes/blob/master/phpunit/phpunit/4.7/phpunit.xml.dist#L16

Zales0123: Great! 🎆

Zales0123:
Nice job, Kamil! 🍉

PR #9765, opened 2018-10-02
Keep the existing pagination when changing sorting on product list page

### LINKED ISSUE #7787: [Shop] Can not set both sorting and limit
Can not set both sorting and limit for displaying products on a taxon page.

pamil:
Fixes #7787.


Zales0123:
Agreed, thank you, Kamil! 🖖

PR #9775, opened 2018-10-03
Limit products shown in associated products autocomplete field

### LINKED ISSUE #9364: Product association : ajax product search fails
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.1

With more than 100 products in database and long text contents, espacially in description field (including html), when you modify a product in back-office and want to make a new product association, if you search a product in the dropdown selector, the request is really slow or fails.

The first thing is that the response returns all fields of the product in the JSON Response while we only need the name and the code of the product. So, it takes a lot of time to serialize the whole object and after, it takes a lot of time to display the list in the javascript function.

I didn't find if it's possible to request a part of the product field and not the whole object.

That's problematic because in a lot of case the request take too much time and stops, so we can't select any product for the association...

pamil:
Fixes #9364.

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Product/Tab/_associations.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Product/Tab/_associations.html.twig
@@ -4,8 +4,8 @@
     {% for associationForm in form.associations %}
         <div class="field">
             {{- form_label(associationForm) -}}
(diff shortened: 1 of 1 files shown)


stefandoorn:
Is there an issue to solve it better in the future?

Zales0123:
Good idea 👍

Issue #9778, opened 2018-10-04
Overriden error templates do not work


emodric:
**Sylius version affected**: 1.3

**Description**  

It looks like that override for Sylius error templates is not applied. This is probably due to Symfony using `templates/bundles` for bundle template overrides instead of `template/bundle` as it is here.

**Steps to reproduce**  

Install Sylius 1.3
Access `/_error/404` in dev mode, or a non existent page in prod mode
Error page will be a default "Oops! An Error Occurred" Symfony error page.

**Possible Solution**  

Renaming `templates/bundle` to `templates/bundles` makes the Sylius error template for 404 and 500 errors show up. Curiously enough, this does not work for other errors (`/_error/401` still shows the Symfony error template for some reason) or even for `SyliusShopBundle` (for example `templates/bundles/SyliusShopBundle/layout.html.twig`).

emodric:
@pamil I believe this can be closed :)

PR #9779, opened 2018-10-04
Fix error templates path

pamil:
Reported in #9778.


Zales0123:
As far as I understand, it only resolves a problem with error template override? Overriding another Sylius templates is still bugged?

pamil:
@Zales0123 it fixes some of that error templates (404, 500), but not others (401) - but overall the path was wrong - `templates/bundle/` instead of `templates/bundles/`.

Zales0123:
Thank you, Kamil! ⭐️

emodric:
@Zales0123 You're right, yes, overriding Sylius templates is still bugged. For example, I tried overriding the main layout template with `templates/bundles/SyliusShopBundle/layout.html.twig` path, and it doesn't work.

pamil:
@emodric I'm on it! :)

nechocandia:
I just tried to do the same path from the tutorial: https://docs.sylius.com/en/1.3/customization/template.html  

`\templates\bundle\SyliusShopBundle\login.html.twig` and it's not working for me, I already did the `cache:clear` and I can't see anything.

I'm using the latest version.

Zales0123:
There is a bug in documentation, should be `\templates\bundles` not `\templates\bundle` :)

pamil:
@nechocandia docs should be fixed already (#9841).

Gerardlc:
> @Zales0123 You're right, yes, overriding Sylius templates is still bugged. For example, I tried overriding the main layout template with `templates/bundles/SyliusShopBundle/layout.html.twig` path, and it doesn't work.

Is there any solution? Im stuck on this

PR #9789, opened 2018-10-06
[FixtureBundle] Add autoconfiguration for instance of FixtureInterface

adrienlucas:
| Q               | A
| --------------- | ---
| Bug fix?        | no
| New feature?    | kindda
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Improve DX by bringing autoconfiguration to services implementing the `Sylius\Bundle\FixturesBundle\Fixture\FixtureInterface`. This allow developper to omit the `sylius_fixtures.fixture` tag on their service definitions (and even omit the whole definition if autowiring is on).

I would suggest to put the tag value in a constant of the `FixtureRegistryPass`, seems like a good idea to you ?

### DIFF
--- a/src/Sylius/Bundle/FixturesBundle/Tests/DependencyInjection/SyliusFixturesExtensionTest.php
+++ b/src/Sylius/Bundle/FixturesBundle/Tests/DependencyInjection/SyliusFixturesExtensionTest.php
@@ -14,7 +14,10 @@ declare(strict_types=1);
 namespace Sylius\Bundle\FixturesBundle\Tests\DependencyInjection;
 
 use Matthias\SymfonyDependencyInjectionTest\PhpUnit\AbstractExtensionTestCase;
+use Sylius\Bundle\FixturesBundle\DependencyInjection\Compiler\FixtureRegistryPass;
+use Sylius\Bundle\FixturesBundle\DependencyInjection\Compiler\ListenerRegistryPass;
 use Sylius\Bundle\FixturesBundle\DependencyInjection\SyliusFixturesExtension;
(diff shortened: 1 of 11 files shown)


src/Sylius/Bundle/FixturesBundle/Tests/DependencyInjection/SyliusFixturesExtensionTest.php
```diff
@@ -50,3 +68,13 @@ protected function getContainerExtensions(): array
         return [new SyliusFixturesExtension()];
     }
 }
+
+
```
lchrusciel: redundant line

lchrusciel:
Good idea with the const, but the same should be done with the listener tag I suppose. Maybe we should create a `FixturesTag` class, which would be final with a private constructor, and put both constants there?

src/Sylius/Bundle/FixturesBundle/Tests/DependencyInjection/SyliusFixturesExtensionTest.php
```diff
@@ -50,3 +68,13 @@ protected function getContainerExtensions(): array
         return [new SyliusFixturesExtension()];
     }
 }
+
+class DummyFixture implements FixtureInterface
```
pamil: Could we extract this class to anther file?

adrienlucas:
@lchrusciel done of rthe fixture and the listener

as it makes the PR really bigger, let me know if you want the const issue to be addressed in an other one.

PR #9792, opened 2018-10-06
[ResourceBundle] Deprecate passing container to ORMTranslatableListener

### LINKED ISSUE #9677: Deprecate passing container to ORMTranslatableListener
**Sylius version affected**: Since 1.0.0, but fix should be applied to the master branch

**Description**:

`ORMTranslatableListener` requires `ContainerInterface` as the second argument only to get one service from it:

https://github.com/Sylius/Sylius/blob/1494e9696ca9b7699abfea304509c03d9f5f3164/src/Sylius/Bundle/ResourceBundle/EventListener/ORMTranslatableListener.php#L45-L51

The solution would be to:
1) Drop typehint on that argument
2) Rename it to `$translatableEntityLocaleAssigner`
3) Make a series of conditions:
    - if an instance of `ContainerInterface` is passed, trigger a deprecation and get that service from container
    - if an instance of `TranslatableEntityLocaleAssignerInterface` is passed, do nothing
    - if something else is passed, throw an exception

kayneth:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | fixes #9677
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/EventListener/ORMTranslatableListener.php
+++ b/src/Sylius/Bundle/ResourceBundle/EventListener/ORMTranslatableListener.php
@@ -40,14 +40,14 @@ final class ORMTranslatableListener implements EventSubscriber
 
     /**
      * @param RegistryInterface $resourceMetadataRegistry
-     * @param ContainerInterface $container
+     * @param object $translatableEntityLocaleAssigner
      */
     public function __construct(
         RegistryInterface $resourceMetadataRegistry,
-        ContainerInterface $container
+        object $translatableEntityLocaleAssigner
     ) {
         $this->resourceMetadataRegistry = $resourceMetadataRegistry;
-        $this->translatableEntityLocaleAssigner = $container->get('sylius.translatable_entity_locale_assigner');
+        $this->translatableEntityLocaleAssigner = $this->processTranslatableEntityLocaleAssigner($translatableEntityLocaleAssigner);
     }
 
     /**
@@ -213,5 +213,31 @@ final class ORMTranslatableListener implements EventSubscriber
         }
 
         return false;
+    }
+
+    /**
+     * @param object $translatableEntityLocaleAssigner
+     * @return TranslatableEntityLocaleAssignerInterface
+     */
+    private function processTranslatableEntityLocaleAssigner(object $translatableEntityLocaleAssigner): TranslatableEntityLocaleAssignerInterface
+    {
+        if ($translatableEntityLocaleAssigner instanceof ContainerInterface) {
+            @trigger_error(
+                sprintf('Passing an instance of "%s" is deprecated since 1.4. Use "%s" instead.',
+                    ContainerInterface::class, TranslatableEntityLocaleAssignerInterface::class), E_USER_DEPRECATED
+            );
+            $translatableEntityLocaleAssigner = $translatableEntityLocaleAssigner->get('sylius.translatable_entity_locale_assigner');
+        }
+
+        if(!$translatableEntityLocaleAssigner instanceof TranslatableEntityLocaleAssignerInterface)
+        {
+            throw new \InvalidArgumentException(sprintf(
+                '`$translatableEntityLocaleAssigner` was expected to return an instance of "%s" , "%s" found',
+                TranslatableEntityLocaleAssignerInterface::class,
+                get_class($translatableEntityLocaleAssigner)
+            ));
+        }
+
+        return $translatableEntityLocaleAssigner;
     }
 }
--- a/CHANGELOG-1.4.md
+++ b/CHANGELOG-1.4.md
@@ -1,1 +1,3 @@
 # CHANGELOG FOR `1.4.X`
+
+- [#9677](https://github.com/Sylius/Sylius/pull/9790) Deprecate passing container to ORMTranslatableListener  (@kayneth)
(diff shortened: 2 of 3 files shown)


src/Sylius/Bundle/ResourceBundle/EventListener/ORMTranslatableListener.php
```diff
@@ -40,14 +40,32 @@
...
         RegistryInterface $resourceMetadataRegistry,
-        ContainerInterface $container
+        $translatableEntityLocaleAssigner
     ) {
         $this->resourceMetadataRegistry = $resourceMetadataRegistry;
-        $this->translatableEntityLocaleAssigner = $container->get('sylius.translatable_entity_locale_assigner');
+
+        if ($translatableEntityLocaleAssigner instanceof ContainerInterface) {
+            @trigger_error(
+                sprintf('Passing an instance of "%s" is deprecated since 1.0. Use %s instead.',
```
pamil: `1.0` -> `Sylius 1.4` probably :)

pamil: And we might use double quotes around the second class name as well.

src/Sylius/Bundle/ResourceBundle/EventListener/ORMTranslatableListener.php
```diff
@@ -40,14 +40,32 @@
 
     /**
      * @param RegistryInterface $resourceMetadataRegistry
-     * @param ContainerInterface $container
+     * @param $translatableEntityLocaleAssigner
      */
     public function __construct(
         RegistryInterface $resourceMetadataRegistry,
-        ContainerInterface $container
+        $translatableEntityLocaleAssigner
```
pamil: Since this PR targets `master`, we could put `object` as a typehint.

src/Sylius/Bundle/ResourceBundle/EventListener/ORMTranslatableListener.php
```diff
@@ -40,14 +40,32 @@
...
+            );
+            $translatableEntityLocaleAssigner = $translatableEntityLocaleAssigner->get('sylius.translatable_entity_locale_assigner');
+        }
+
+        if(!$translatableEntityLocaleAssigner instanceof TranslatableEntityLocaleAssignerInterface)
+        {
+            throw new \InvalidArgumentException(sprintf(
+                '`$translatableEntityLocaleAssigner` was expected to return an instance of "%s" , "%s" found',
+                TranslatableEntityLocaleAssignerInterface::class,
+                is_object($translatableEntityLocaleAssigner) ? get_class($translatableEntityLocaleAssigner) : gettype($translatableEntityLocaleAssigner)
```
pamil: Following up the comment about constructor `object` typehint, we might remove this ternary operator and use `get_class` only.

src/Sylius/Bundle/ResourceBundle/EventListener/ORMTranslatableListener.php
```diff
@@ -40,14 +40,32 @@
...
+            ));
+        }
+
+        $this->translatableEntityLocaleAssigner = $translatableEntityLocaleAssigner;
```
pamil: In the end, we might extract this constructor logic to a private method like `processTranslatableEntityLocaleAssigner` (that is probably not the best name :)), returning an instance of `TranslatableEntityLocaleAssignerInterface` or throwing an exception if couldn't get it.

kayneth: I canno't come with a better name for that method :/

kayneth:
Hi ! Sorry for the delay but I managed to update this PR :+1:

pamil:
No problem. Thank you, Dylan! 🥇

src/Sylius/Bundle/ResourceBundle/EventListener/ORMTranslatableListener.php
```diff
@@ -214,4 +214,30 @@ private function hasUniqueConstraint(ClassMetadata $metadata, array $columns): b
...
+
+    /**
+     * @param object $translatableEntityLocaleAssigner
+     * @return TranslatableEntityLocaleAssignerInterface
+     */
+    private function processTranslatableEntityLocaleAssigner(object $translatableEntityLocaleAssigner): TranslatableEntityLocaleAssignerInterface
+    {
+        if ($translatableEntityLocaleAssigner instanceof ContainerInterface) {
+            @trigger_error(
+                sprintf('Passing an instance of "%s" is deprecated since 1.4. Use "%s" instead.',
```
teohhanhui: The error message is rather confusing...

src/Sylius/Bundle/ResourceBundle/EventListener/ORMTranslatableListener.php
```diff
@@ -214,4 +214,30 @@ private function hasUniqueConstraint(ClassMetadata $metadata, array $columns): b
...
+                    ContainerInterface::class, TranslatableEntityLocaleAssignerInterface::class), E_USER_DEPRECATED
+            );
+            $translatableEntityLocaleAssigner = $translatableEntityLocaleAssigner->get('sylius.translatable_entity_locale_assigner');
+        }
+
+        if(!$translatableEntityLocaleAssigner instanceof TranslatableEntityLocaleAssignerInterface)
+        {
+            throw new \InvalidArgumentException(sprintf(
+                '`$translatableEntityLocaleAssigner` was expected to return an instance of "%s" , "%s" found',
```
teohhanhui: The error message is wrong...

src/Sylius/Bundle/ResourceBundle/EventListener/ORMTranslatableListener.php
kayneth: Because we don't explicitly tell the user why ?

teohhanhui: I think the main problem is that we ask the user to "use `TranslatableEntityLocaleAssignerInterface` instead of `ContainerInterface`" (???)

PR #9795, opened 2018-10-07
[OrderBundle] Add autoconfiguration for cart contexts and order processor services.

adrienlucas:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | 
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

This is made to improve DX by allowing auto-configuration (and thus, autowiring) of services implementing the `CartContextInterface` and the `OrderProcessorInterface` (as for the FixtureBundle - see #9789).

The PR also contains a "fix" about creating dummy classes for the FixtureBundle : phpUnit does that very well and it costs much less lines of codes :)

### DIFF
--- a/src/Sylius/Bundle/OrderBundle/DependencyInjection/Compiler/RegisterCartContextsPass.php
+++ b/src/Sylius/Bundle/OrderBundle/DependencyInjection/Compiler/RegisterCartContextsPass.php
@@ -17,12 +17,14 @@ use Sylius\Bundle\ResourceBundle\DependencyInjection\Compiler\PrioritizedComposi
 
 final class RegisterCartContextsPass extends PrioritizedCompositeServicePass
 {
+    public const CART_CONTEXT_SERVICE_TAG = 'sylius.context.cart';
+
     public function __construct()
     {
         parent::__construct(
             'sylius.context.cart',
             'sylius.context.cart.composite',
-            'sylius.context.cart',
+            self::CART_CONTEXT_SERVICE_TAG,
             'addContext'
         );
     }
--- a/src/Sylius/Bundle/OrderBundle/DependencyInjection/SyliusOrderExtension.php
+++ b/src/Sylius/Bundle/OrderBundle/DependencyInjection/SyliusOrderExtension.php
@@ -13,7 +13,11 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\OrderBundle\DependencyInjection;
 
+use Sylius\Bundle\OrderBundle\DependencyInjection\Compiler\RegisterCartContextsPass;
+use Sylius\Bundle\OrderBundle\DependencyInjection\Compiler\RegisterProcessorsPass;
 use Sylius\Bundle\ResourceBundle\DependencyInjection\Extension\AbstractResourceExtension;
+use Sylius\Component\Order\Context\CartContextInterface;
+use Sylius\Component\Order\Processor\OrderProcessorInterface;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
@@ -36,5 +40,14 @@ final class SyliusOrderExtension extends AbstractResourceExtension
 
         $container->setParameter('sylius_order.cart_expiration_period', $config['expiration']['cart']);
         $container->setParameter('sylius_order.order_expiration_period', $config['expiration']['order']);
+
+        $container
+            ->registerForAutoconfiguration(CartContextInterface::class)
+            ->addTag(RegisterCartContextsPass::CART_CONTEXT_SERVICE_TAG)
+        ;
+        $container
+            ->registerForAutoconfiguration(OrderProcessorInterface::class)
+            ->addTag(RegisterProcessorsPass::PROCESSOR_SERVICE_TAG)
+        ;
     }
 }
--- a/docs/components_and_bundles/bundles/SyliusOrderBundle/processors.rst
+++ b/docs/components_and_bundles/bundles/SyliusOrderBundle/processors.rst
@@ -6,7 +6,7 @@ Order processors are responsible of manipulating the orders to apply different p
 Registering custom processors
 -----------------------------
 
-Once you have your own :ref:`component_order_processors_order-processor-interface` implementation you need to register it as a service.
+Once you have your own :ref:`component_order_processors_order-processor-interface` implementation, if services autowiring and auto-configuration are not enabled, you need to register it as a service.
 
 .. code-block:: xml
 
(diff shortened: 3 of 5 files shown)


src/Sylius/Bundle/FixturesBundle/Tests/DependencyInjection/SyliusFixturesExtensionTest.php
```diff
@@ -53,14 +55,14 @@ public function it_autoconfigures_fixtures_and_listeners(): void
...
-                ->setClass(DummyFixture::class)
+                ->setClass(self::getMockClass(FixtureInterface::class))
                 ->setAutoconfigured(true)
         );
 
         $this->container->setDefinition(
             'acme.listener_autoconfigured',
             (new Definition())
-                ->setClass(DummyListener::class)
+                ->setClass(self::getMockClass(ListenerInterface::class))
```
lchrusciel: nice 👍

src/Sylius/Bundle/OrderBundle/DependencyInjection/SyliusOrderExtension.php
```diff
@@ -36,5 +40,10 @@ public function load(array $config, ContainerBuilder $container): void
 
         $container->setParameter('sylius_order.cart_expiration_period', $config['expiration']['cart']);
         $container->setParameter('sylius_order.order_expiration_period', $config['expiration']['order']);
+
+        $container->registerForAutoconfiguration(CartContextInterface::class)
```
lchrusciel: Just my personal view, but maybe it would be better to order chained functions like this:
```php
$container
    ->registerForAutoconfiguration(CartContextInterface::class)
    ->addTag(RegisterProcessorsPass::CART_CONTEXT_SERVICE_TAG)
;
```

WDYT?

adrienlucas:
@lchrusciel 
> 1. While the fixtures and their listeners could be autoloaded without a doubt, with contexts and order processors, it is important to remember about their priorities, as it may affect the end result. Your proposed solution will work, and hand configured tags may be just an available configuration option for more advanced users, but I just wanted to raise this issue. Do we have a plan for the default behavior of this /cc @Sylius/core-team? Maybe we should add an interface with getPriority function, which may (but doesn't have to) be used in compiler pass?

The interface idea may be a good idea indeed; and it would solve some other cases (I'm thinking about `type`, `alias`, `label` and other required tag properties we have in Sylius) if applied to other bundles.
Also, i'm not an e-commerce expert and as the documentation did not mentioned the `priority` as a significant property (it is just set to `0` in the code sample) I thought it was ok to autoload processors with no priority (the `` set it to `0` by default). What are the use-cases of adding you own Order Processor ? (it would be nice to have it in the documentation too)

As a side effect, what are the use-cases of creating your own Cart Context ? I would be happy to include a new page of documentation about it, the subject is not covered yet.

> 2. If we agree on some solution for 1st point, there are also three more contexts available in code base: LocaleContext, ChannelContext, and CurrencyContext. Are you willing to configure autoconfiguration also for them?

Yes. Should I do it in this PR ? I will wait for `1.` to be resolved of course.

src/Sylius/Bundle/OrderBundle/DependencyInjection/SyliusOrderExtension.php
adrienlucas: done (same for `SyliusFixtureBundle` for consistency)

pamil:
I think the current solution with autoconfiguration using the default priority (0) is simple enough, but some docs on how to change a priority of autoconfigured service might be useful for newcomers.

Basically what we're using with the most of the tagged services (cart contexts, order processors, locale contexts, etc.) is a composite pattern. When called to do something, it goes through the list of tagged services and returns the first significant answer (usually we either return something or throw an exception that we couldn't resolve it). Therefore priorities are quite important if you want to modify the existing behaviour, let's take the channel context for example:
 - we've got test channel context with priority 128, which is used to forcibly use a given channel based on some cookie or get parameter
 - we've got the regular channel context with priority 0, which is used to resolve channel based on the hostname
 - we've got the default channel context with priority -128, which returns a channel if there's only one

So that if you want to always forcibly set one of the channels, you want to register your own channel context with priority >128 returning a channel. 
If you want to resolve a channel based on path (see #5846), you want to register a channel context with default priority 0 (so if hostname rule does not apply, you try the path rule).
If you want to add more logic to the fallback channel resolving, you might want to use priority <0 and so on.

I'm against the interface as it brings significant overhead (you need to get the priority from an object of that class so that you need to initialise all the dependencies and the transient dependencies and so on, so you might end up with issues like some repository being a transient dependency calling the database while clearing the cache when you don't have any database set up yet etc.).

Could you make issues for that use cases related questions? Then we can handle it asynchronously and independently from this PR :)

src/Sylius/Bundle/FixturesBundle/DependencyInjection/SyliusFixturesExtension.php
```diff
@@ -37,10 +37,14 @@ public function load(array $config, ContainerBuilder $container): void
 
         $this->registerSuites($config, $container);
 
-        $container->registerForAutoconfiguration(FixtureInterface::class)
-            ->addTag(FixtureRegistryPass::FIXTURE_SERVICE_TAG);
-        $container->registerForAutoconfiguration(ListenerInterface::class)
-            ->addTag(ListenerRegistryPass::LISTENER_SERVICE_TAG);
+        $container
```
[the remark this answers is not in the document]
adrienlucas: done : https://github.com/Sylius/Sylius/pull/9810/files

src/Sylius/Bundle/OrderBundle/DependencyInjection/Compiler/RegisterCartContextsPass.php
```diff
@@ -17,12 +17,14 @@
 
 final class RegisterCartContextsPass extends PrioritizedCompositeServicePass
 {
+    PUBLIC const CART_CONTEXT_SERVICE_TAG = 'sylius.context.cart';
```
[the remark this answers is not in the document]
adrienlucas: oups :/

fixed.

adrienlucas:
@pamil  here are the documentation issues : https://github.com/Sylius/Sylius/issues/9811 https://github.com/Sylius/Sylius/issues/9812

pamil:
Thank you, Adrien! 🎉

belmeopmenieuwesim:
So how does one change the priority now of an autoconfigured context? Because right now, my custom cart context will always execute first, while I want it to be executed last.. And there is no way for me to do it, since manually defining my context with the proper priority set will just duplicate it and I end up with two instances of my cart context added to the composite one.

belmeopmenieuwesim:
It seems that if you set `autoconfigure: false` on your service definition you can properly set the priority again without the autoconfiguration creating a duplicate.

PR #9801, opened 2018-10-08
Update template.rst

bitbager:
According to the official Symfony4 docs, it should be `bundles`. Source: https://symfony.com/doc/current/bundles/override.html 🙂 .

| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/docs/customization/template.rst
+++ b/docs/customization/template.rst
@@ -20,7 +20,7 @@ Methods of templates customizing
 
     There are three ways of customizing templates of Sylius:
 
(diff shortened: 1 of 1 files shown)


bitbager:
There's one more problem with this documentation. Although it should be valid, in reality it does not work  🙁

Related: #9802

PR #9804, opened 2018-10-09
[ThemeBundle] Add support for Twig namespaced paths and "templates/" top-level directory

### LINKED ISSUE #8929: Overriding bundle templates via Twig namespaces is not possible

### LINKED ISSUE #9719: Symfony flex structure : templates override

### LINKED ISSUE #9802: Overriding templates from Sylius bundles does not work
**Sylius version affected**: 1.3.0

**Description**  
Overriding templates from Sylius bundles does not work. 

**Steps to reproduce**  
After I create a file under `templates/bundles/SyliusShopBundle/Homepage/index.html.twig` path and I change the content to blank, I still get the Sylius vendor's homepage index template rendered. The weird thing is that overriding templates from Twig bundle (exceptions) works fine.

**Possible Solution**  
I think this might have something to do with the old templates structure, custom namespaces or theme loader for Sylius bundles.

**Tip**
If I remove the core template from vendor, then the custom template from `templates/bundles/Homepage/index.html.twig` is loaded. It seems like the vendor has a higher priority than the app. 
Might seem obvious, but the bug was detected on Sylius-Standard and replicated in the PluginSkeleton.

pamil:
| Q               | A
| --------------- | -----
| Branch?         | 1.3?
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9719, fixes #8929, fixes #9802
| License         | MIT

It's both a new feature (added support) and a bugfix (Symfony 4 directory structure & Twig namespaced paths aren't handled correctly now, which prevents people from using Sylius 1.3). That's why I target 1.3 with this PR.

First commit is containing all the new tests, second one provides the implementation.

### DIFF
--- a/src/Sylius/Bundle/ThemeBundle/Locator/BundleResourceLocator.php
+++ b/src/Sylius/Bundle/ThemeBundle/Locator/BundleResourceLocator.php
@@ -19,20 +19,12 @@ use Symfony\Component\HttpKernel\KernelInterface;
 
 final class BundleResourceLocator implements ResourceLocatorInterface
 {
-    /**
-     * @var Filesystem
-     */
+    /** @var Filesystem */
     private $filesystem;
 
-    /**
-     * @var KernelInterface
-     */
+    /** @var KernelInterface */
     private $kernel;
 
-    /**
-     * @param Filesystem $filesystem
-     * @param KernelInterface $kernel
-     */
     public function __construct(Filesystem $filesystem, KernelInterface $kernel)
     {
         $this->filesystem = $filesystem;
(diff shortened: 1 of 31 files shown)


pamil:
If someone could test it manually with Sylius-Standard, I'd be very grateful. :)

Zales0123:
zrzut ekranu 2018-10-10 o 09 18 10
zrzut ekranu 2018-10-10 o 09 18 21

I've just checked it on a simple case and it seems to work :)

Zales0123:
Thank you, Kamil! 🥇

emodric:
@pamil Any timeline on when this will be released?

pamil:
Not later than tomorrow's evening.

PR #9827, opened 2018-10-15
Custom homepage controller as public service

davidroberto:
It seems that the Custom homepage controller service needs to be marked as public as well, otherwise it cannot be fetched from the container. 
It looks like there is a default service configuration for the controllers in config.yml which is supposed to make public all controllers in AppBundle but it doesn't resolve this issue

| Q               | A
| --------------- | -----
| Branch?         | 1.2, 1.3 or master 
| Bug fix?        | no/yes
| New feature?    | no/yes
| BC breaks?      | no/yes
| Deprecations?   | no/yes 
| License         | MIT


pamil:
Thank you, David! 🥇

PR #9828, opened 2018-10-15
access the container inside the customized Standard Controller

davidroberto:
| Q               | A
| --------------- | -----
| Branch?         | 1.2, 1.3 or master 
| Bug fix?        | no/yes
| New feature?    | no/yes
| BC breaks?      | no/yes
| Deprecations?   | no/yes 
| License         | MIT

### DIFF
--- a/docs/customization/controller.rst
+++ b/docs/customization/controller.rst
@@ -188,4 +188,8 @@ Remember to import the ``app/config/services.yml`` into the ``app/config/config.
 
 From now on your ``customAction`` of the ``HomepageController`` will be available alongside the ``indexAction`` from the base class.
 
+.. use services::
+
+If you need to access the container inside your action, your customized Standard Controller needs to extends ``Symfony\Bundle\FrameworkBundle\Controller\Controller``
+
 .. include:: /customization/plugins.rst.inc


docs/customization/controller.rst
```diff
@@ -188,4 +188,8 @@ Remember to import the ``app/config/services.yml`` into the ``app/config/config.
 
 From now on your ``customAction`` of the ``HomepageController`` will be available alongside the ``indexAction`` from the base class.
 
+.. use services::
+
+If you need to access the container inside your action, your customized Standard Controller needs to extends ``Symfony\Bundle\FrameworkBundle\Controller\Controller``
```
lchrusciel: This is a bad practice. We should recommend using DI instead.

davidroberto: I agree about the use of the container, but Sylius makes heavy use of it in it's doc for all the services:
https://docs.sylius.com/en/1.2/book/architecture/resource_layer.html

Zales0123: I think that these docs are written this way just to make things shorter and easier to read. But they can be changed too, to not spread the wrong practices :)

pamil: Anyway, you don't need to extend Symfony's Controller to get the container, you could just inject it instead.

PR #9834, opened 2018-10-17
Use SyliusShop simple menu template

maximehuran:
## Adding a menu template to ShopBundle

I created the file `Menu/simple.html.twig` in `SyliusShopBundle` in order to be able to override this template in theme and only impact the front. With that, we can customize our shop without break the admin render.

## Change UiBundle calls in ShopBundle

I changed the menu called in `ShopBundle/Resources/views/Account/layout.html.twig`.
The `Menu/simple.html.twig` is only called in this template. In the future, if I want to use it on admin and/or in front I will be able to customize it.

Thank you for your amazing work on Sylius 🚀
If you want more information, ask me 👀

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/Menu/simple.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/Menu/simple.html.twig
@@ -1,0 +1,1 @@
+{% extends '@SyliusUi/Menu/simple.html.twig' %}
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ShopBundle/Resources/views/Menu/simple.html.twig
```diff
@@ -0,0 +1 @@
+{% extends '@SyliusUi/Menu/simple.html.twig' %}
```
lchrusciel: Missing blank line at the end of file

maximehuran:
Missing blank line is now here

pamil:
Thank you! 🥇

PR #9837, opened 2018-10-17
Repaired shipping method fixture

JakobTolkemit:
| Q               | A
| --------------- | -----
| Branch?         | 1.2 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| License         | MIT

Using fixtures it was impossible to set categories for shipping methods, because because the option in the fixture and the corresponding factory had different names.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/ShippingMethodFixture.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/ShippingMethodFixture.php
@@ -37,7 +37,7 @@ class ShippingMethodFixture extends AbstractResourceFixture
                 ->scalarNode('description')->cannotBeEmpty()->end()
                 ->scalarNode('zone')->cannotBeEmpty()->end()
                 ->booleanNode('enabled')->end()
-                ->scalarNode('shipping_category')->end()
+                ->scalarNode('category')->end()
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/CoreBundle/Fixture/ShippingMethodFixture.php
```diff
@@ -37,7 +37,7 @@ protected function configureResourceNode(ArrayNodeDefinition $resourceNode): voi
                 ->scalarNode('description')->cannotBeEmpty()->end()
                 ->scalarNode('zone')->cannotBeEmpty()->end()
                 ->booleanNode('enabled')->end()
-                ->scalarNode('category')->end()
+                ->scalarNode('shipping_category')->end()
```
pamil: I think we should fix the example factory to allow for `category` instead (as shipping method has `taxCategory` and `category` relations, there's no `shippingCategory`).

pamil: This change would make any fixture with `category` defined failing.

lchrusciel:
May we add usage of this method to default fixture suite? I will protects us from regression for the future.

lchrusciel:
Thank you, Jakob! :1st_place_medal:

PR #9843, opened 2018-10-18
Image / File display in edit form

davidroberto:
If there is a collection of file fields inside the form (for a custom resource with a collection of images for example) and there one or more image already persisted, display a thumbnail of that image for each (I'm not sure of the best way to do it, i think my snippet  could be improved a lot).
Currently there is nothing displayed, the file field in the edit page looks empty even if there is already a file saved, so it's a bit confusing for the admin.

| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no/yes
| Deprecations?   | no/yes 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig
+++ b/src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig
@@ -114,6 +114,11 @@
 {% macro collection_item(form, allow_delete, button_delete_label, index) %}
     {% spaceless %}
         <div data-form-collection="item" data-form-collection-index="{{ index }}">
+            {% if form.vars.value.path|default(null) is not null %}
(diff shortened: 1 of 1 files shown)


Zales0123:
It's been a while since this change has been proposed, sorry David for a long response time ⌛️ It would be nice to have it merged, I miss 2 things here:

- a screenshot, showing how does it look now 🖼 
- a scenario, that would prevent us from an unneeded regression (take a look at [this Behat step](https://github.com/Sylius/Sylius/blob/master/features/product/viewing_products/viewing_product_image.feature#L16)) ☠️ 

Do you think you would be possible to add them to your PR? Thanks for the contribution!

Zales0123:
Thank you @davidroberto, maybe then it's a good occasion to become familiar with it :) good luck! 🖖

Zales0123:
Merged in #10372 🎉

PR #9853, opened 2018-10-23
Add sylius.customer.post_register when the user is created

### LINKED ISSUE #9750: Event not triggered when creating user from OAuth authentication
**Sylius version affected**: 1.2.7

**Description**  
Using HWIOAuthBundle I successfully set two providers (google and a custom one). When a new account is created, no email validation is triggered and also no `kernel.event_listener` with  `sylius.*.post_create` tags are triggered.

**Steps to reproduce**  
Configure OAuth with the provider Google for example and log-in with an not existant user email

**Possible Solution** 
Dispatch generic events in a DoctrineEventSubscriber to trigger those events before and after persisting `Entity`

**Precisions**
I'm in a dev environment with default data in the database (eg: Channel "Default" with `accountVerificationRequired` to `true`)

`app/config/config_dev.yml`as :
```
swiftmailer:
    disable_delivery: false #to allow mailhog to catch dev mails
```

Prometee:
When a new user is created by `Sylius\Bundle\CoreBundle\OAuth\UserProvider` no mail was sent to the user to check the email or to just confirm that the account is well created.

| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9750 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
+++ b/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
@@ -151,7 +151,7 @@ class UserProvider extends BaseUserProvider implements AccountConnectorInterface
      *
      * @return SyliusUserInterface
      */
-    private function createUserByOAuthUserResponse(UserResponseInterface $response): SyliusUserInterface
+    protected function createUserByOAuthUserResponse(UserResponseInterface $response): SyliusUserInterface
     {
         /** @var SyliusUserInterface $user */
         $user = $this->userFactory->createNew();
@@ -203,7 +203,7 @@ class UserProvider extends BaseUserProvider implements AccountConnectorInterface
      *
      * @return SyliusUserInterface
      */
-    private function updateUserByOAuthUserResponse(UserInterface $user, UserResponseInterface $response): SyliusUserInterface
+    protected function updateUserByOAuthUserResponse(UserInterface $user, UserResponseInterface $response): SyliusUserInterface
     {
         /** @var SyliusUserInterface $user */
         Assert::isInstanceOf($user, SyliusUserInterface::class);
--- a/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
+++ b/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
@@ -26,6 +26,8 @@ use Sylius\Component\Resource\Repository\RepositoryInterface;
 use Sylius\Component\User\Canonicalizer\CanonicalizerInterface;
 use Sylius\Component\User\Model\UserOAuthInterface;
 use Sylius\Component\User\Repository\UserRepositoryInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\EventDispatcher\GenericEvent;
 use Symfony\Component\Security\Core\Exception\UsernameNotFoundException;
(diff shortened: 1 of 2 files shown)


Prometee:
In my own project I clone the class and declare it as service with this `UserProvider` as parent service, adding `@event_listener` as a `call` method. The only thing which can be helpful and allow us to trigger this event more easily : is to make all `private` methods and attributes `protected`.

lchrusciel:
I would say that service decoration is a much cleaner solution (https://symfony.com/doc/3.4/service_container/service_decoration.html). Therefore, I would vote in favor of finalizing this class, but we are not allowed to do it, because of BC promise.

Anyway, the current implementation is no-go, as we cannot introduce BC-breaks. We need to create a separate service (preferably final, with event dispatcher and maybe some additional refactoring) or decorate an existing one.

Zales0123:
Hello @Prometee, do you think you would be able to continue work on this PR and get rid of BC breaks mentioned by @lchrusciel? Thanks :)

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -85,7 +93,8 @@ public function __construct(
         RepositoryInterface $oauthRepository,
         ObjectManager $userManager,
         CanonicalizerInterface $canonicalizer,
-        CustomerRepositoryInterface $customerRepository
+        CustomerRepositoryInterface $customerRepository,
+        EventDispatcherInterface $eventDispatcher
```
lchrusciel, line 97: This is the main source of BC Break, if we will allow for null value here and check for it later, we can proceed and merge this PR to 1.7.

```suggestion
        ?EventDispatcherInterface $eventDispatcher = null
```

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -95,6 +104,7 @@ public function __construct(
         $this->userFactory = $userFactory;
         $this->userManager = $userManager;
         $this->customerRepository = $customerRepository;
+        $this->eventDispatcher = $eventDispatcher;
```
lchrusciel, line 107:
```suggestion
        if ($eventDispatcher === null) {
            @trigger_error(sprintf('Not passing a $eventDispatcher to %s constructor is deprecated since Sylius 1.7 and will be removed in Sylius 2.0.', self::class), \E_USER_DEPRECATED);
        }
        $this->eventDispatcher = $eventDispatcher;
```

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -116,8 +126,11 @@ public function loadUserByOAuthUserResponse(UserResponseInterface $response): Us
             if (null !== $user) {
                 return $this->updateUserByOAuthUserResponse($user, $response);
             }
+            $user = $this->createUserByOAuthUserResponse($response);
+
+            $this->eventDispatcher->dispatch('sylius.customer.post_register', new GenericEvent($user->getCustomer()));
```
lchrusciel, line 131:
```suggestion
        if ($eventDispatcher === null) {
            $this->eventDispatcher->dispatch('sylius.customer.post_register', new GenericEvent($user->getCustomer()));
        }
```

PR #9864, opened 2018-10-26
[Behat] Mitigate race conditions in Behat tests

anthid:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes (well, not exactly...)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Sometimes, in a semi-random way (some pages are affected more frequently than others) Javascript Behat tests experience race conditions, leading to failed tests. 

This solution -while probably not the most orthodox- works in the 100% of tests i've done. And given that we are running tests in a controled environment (localhost calls) I think it's just fine.

(There is more advanced solutions involving spinning until success or timeout, but in our case this solutions must be implemented multiple times at context level and requires more code, and deeper changes)

### DIFF
--- a/src/Sylius/Behat/Page/Admin/Crud/UpdatePage.php
+++ b/src/Sylius/Behat/Page/Admin/Crud/UpdatePage.php
@@ -44,6 +44,7 @@ class UpdatePage extends SymfonyPage implements UpdatePageInterface
      */
     public function saveChanges()
     {
+        usleep(500000);
         $this->getDocument()->pressButton('sylius_save_changes_button');
     }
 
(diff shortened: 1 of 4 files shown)


Zales0123:
Hello @anthid, it's been a long time, sorry for that ☮️ I would personally vote to focus on lowering an amount of js Behat tests (in general they're a terrible idea). Fixing some random fails with some random `sleep` methods do not assure us more that they will be working correctly. But thanks for your contribution, reopen this PR (you, or anyone form @Sylius/core-team) if you think the discussion should go on 🖖

PR #9661, opened 2018-08-20
typo maybe!

xeptore:
isn't it `sylius`

| Q               | A
| --------------- | -----
| Branch?         | 1.1, 1.2 or master 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no/yes 
| Related tickets | fixes #X, partially #Y, mentioned in #Z
| License         | MIT

PR #9873, opened 2018-10-29
[CoreBundle] Typo in sylius_order.yml callback name

sergejvoronov:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine/sylius_order.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine/sylius_order.yml
@@ -78,7 +78,7 @@ winzou_state_machine:
                     on: ["cancel"]
                     do: ["@sm.callback.cascade_transition", "apply"]
                     args: ["object", "event", "'cancel'", "'sylius_order_shipping'"]
-                sylis_cancel_order:
+                sylius_cancel_order:
                     on: ["cancel"]
                     do: ["@sylius.inventory.order_inventory_operator", "cancel"]
                     args: ["object"]


Zales0123:
Please @sergejvoronov, if you could rebase to `1.2` as it should be a base branch :)

pamil:
may become messed-up.**

execute `git fetch origin && git reset "master"` to update your local branch.
 
src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine/sylius_order.yml
```diff
@@ -78,7 +78,7 @@ winzou_state_machine:
                     on: ["cancel"]
                     do: ["@sm.callback.cascade_transition", "apply"]
                     args: ["object", "event", "'cancel'", "'sylius_order_shipping'"]
-                sylis_cancel_order:
+                sylius_cancel_order:
```
pamil: While it's great not to have any typo in callback name, I'd hold it until we figure out what should we do in those cases (there's also #9661).

This fix would break external applications in which this callback was previously customised or disabled. It's a rare edge case, but a possible one.

stefandoorn: Potential BC indeed. Isn't there an option to have them both in there and deprecate the 'wrong' named one (e.g. with a deprecation service call added to it via a listener or whatever)? Then remove them in an upcoming version.

PR #3703, opened 2015-12-10
[Core] Extract path generation logic from ImageUploader

pamil:
Brand new `ImageUploader` allows us to change the way paths are generated without modifying or copying it. You just need to implement `PathGeneratorInterface` and inject it to `sylius.image_uploader` service as second argument.

Anyway, the specifications of new classes uses some not-merged-yet features of Prophecy: phpspec/prophecy#237 and phpspec/prophecy#238.

TODO:
 - [x] Add a decorator for `PathGeneratorInterface` that put uploaded files into nested directories (eg. `deadbeef.png` becomes `de/ad/beef.png`
 - [x] Create an easy way to switch between path generators

PR #9885, opened 2018-10-30
fixed ad blocking issue

### LINKED ISSUE #9884: Adblockers will block image paths containing 'ad'
**Sylius version affected**: 1.*

**Description**  
This issue is made from this forum thread: https://forum.sylius.com/t/adblockers-blocking-certain-images/672 so creds to [William](https://forum.sylius.com/u/William) for figuring it out. I just wanted to add this issue before it was forgotten since it's quite important in my perspective.

Images are not displayed when the path contains `ad`.

**Steps to reproduce**  
I haven't tried this myself, but a quick google search tells me the problem is evident: If you have certain ad blockers, images with paths containing `ad` will not be shown because the ad blocker believes it's an ad.

**Possible Solution**  
The [class](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Uploader/ImageUploader.php#L54) handling file paths should check for the phrase `ad`.

loevgaard:
| Q               | A
| --------------- | -----
| Branch?         | 1.2, 1.3, master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9884
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Core/Uploader/ImageUploader.php
+++ b/src/Sylius/Component/Core/Uploader/ImageUploader.php
@@ -54,7 +54,7 @@ class ImageUploader implements ImageUploaderInterface
         do {
             $hash = bin2hex(random_bytes(16));
             $path = $this->expandPath($hash . '.' . $file->guessExtension());
-        } while ($this->filesystem->has($path));
+        } while ($this->filesystem->has($path) || $this->isAdBlockingProne($path));
 
         $image->setPath($path);
 
@@ -99,5 +99,17 @@ class ImageUploader implements ImageUploaderInterface
     private function has(string $path): bool
     {
         return $this->filesystem->has($path);
+    }
+
+    /**
+     * Will return true if the path is prone to be blocked by ad blockers
+     *
+     * @param string $path
+     *
+     * @return bool
+     */
+    private function isAdBlockingProne(string $path): bool
+    {
+        return stripos($path, 'ad') !== false;
     }
 }


loevgaard:
Well, basically the `random_bytes` function should be preseeded with a value, just like you can do with i.e. the [Carbon library](https://carbon.nesbot.com/docs/#api-testing). But this means that you either need to create a lib to do that or find one that does.

Or you could do what @Zales0123 says ;)

Do you want to do any of this in this PR? If so I probably need some pointers as to where you want the `PathGenerator` class or if you have a lib in mind, which one.

Zales0123:
We've just had a little discussion about this with @lchrusciel in the office 😄 We both think that service extraction would be the best choice, but on the other hand, it would be a BC break and lead to many extra things to do (deprecations, fallback to old function etc.)... I don't know is the value of such a test worth so many actions 😄 cc @Sylius/core-team

loevgaard:
Well, personally I probably wouldn't, but on the other hand it lowers the quality of this exact class ;) I don't know if you any system in place where you put issues like this for future implementation? :)

pamil:
The idea of extracting the path generation logic reminds me of a PR I made some time ago, #3703.

src/Sylius/Component/Core/Uploader/ImageUploader.php
```diff
@@ -100,4 +100,16 @@ private function has(string $path): bool
...
+    /**
+     * Will return true if the path is prone to be blocked by ad blockers
+     *
+     * @param string $path
+     *
+     * @return bool
+     */
+    private function isAdBlockingProne(string $path): bool
+    {
+        return stripos($path, 'ad') !== false;
```
pamil: It filters out 11-12% of generated paths. Do adblockers block the images if `ad` is the directory name or also if `ad` is in the middle of a name?

pamil: No need for case-insensitive check here though.

src/Sylius/Component/Core/Uploader/ImageUploader.php
```diff
@@ -54,7 +54,7 @@ public function upload(ImageInterface $image): void
         do {
             $hash = bin2hex(random_bytes(16));
             $path = $this->expandPath($hash . '.' . $file->guessExtension());
-        } while ($this->filesystem->has($path));
+        } while ($this->filesystem->has($path) || $this->isAdBlockingProne($path));
```
pamil: We should probably call the new method first, as filesystem calls usually take longer than string lookup.

src/Sylius/Component/Core/Uploader/ImageUploader.php
loevgaard: As I mentioned in the intro I have not any ad blocker installed and I couldn't find any valid source  when trying to google it, so basically I just made the assumption that it would filter both paths and filenames

PR #9896, opened 2018-11-05
[Taxon] Make createTranslationBasedQueryBuilder protected

pierre-H:
For customizing repository

| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT


pamil:
Could you tell us what is your use case for protected method?

pierre-H:
I need to create a custom function in the repository and when I extend it, I'd like to use this function so it would be more easy for me.

pamil:
Making this method protected would make it a part of our public API so that we wouldn't be able to change it in the future as much as we'd like to and it'd be a difficult part to keep backwards compatibility. We strongly recommend just copying this method contents in your project.

Zales0123:
Hello Pierre, I hope Kamil explanation about why we would like to keep this method private is sufficient :) Feel free to open this PR if any more discussion is required. Thanks for the contribution!

PR #9901, opened 2018-11-06
Handle the base locale code in fixtures

pjedrzejewski:
| Q               | A
| --------------- | -----
| Branch?         | 1.2, 1.3, master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

During the training, we ran into an issue where we added the default locale code in the custom fixture suite and it has been added twice. This should avoid the duplicate error constraint. :)

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/LocaleFixture.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/LocaleFixture.php
@@ -53,9 +53,7 @@ class LocaleFixture extends AbstractFixture
      */
     public function load(array $options): void
     {
-        $localesCodes = array_merge([$this->baseLocaleCode], $options['locales']);
-
-        foreach ($localesCodes as $localeCode) {
+        foreach ($options['locales'] as $localeCode) {
             /** @var LocaleInterface $locale */
             $locale = $this->localeFactory->createNew();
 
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/CoreBundle/Fixture/LocaleFixture.php
```diff
@@ -53,9 +53,7 @@ public function __construct(FactoryInterface $localeFactory, ObjectManager $loca
      */
     public function load(array $options): void
     {
-        $localesCodes = array_merge([$this->baseLocaleCode], $options['locales']);
```
pamil: Why don't we use `array_unique` after the merge? It should solve the issue.

src/Sylius/Bundle/CoreBundle/Tests/Fixture/LocaleFixtureTest.php
```diff
@@ -39,6 +39,28 @@ public function locales_can_be_set(): void
         $this->assertConfigurationIsValid([['locales' => ['en_US' => true, 'pl_PL' => false, 'es_ES' => true]]], 'locales');
     }
 
+    /**
+     * @test
+     */
+    public function default_locale_is_always_added() : void
+    {
+        $this->assertProcessedConfigurationEquals(
+            [['locales' => ['en_US' => true, 'pl_PL' => false, 'es_ES' => true]]],
```
pamil: Looks like this configuration test has been broken since #7346 because the fixture itself expect locales to be values, not keys.

Zales0123:
Ping, @pjedrzejewski, are you willing to continue work on this PR? 😄

pamil:
Already fixed by https://github.com/Sylius/Sylius/pull/10342.

PR #9906, opened 2018-11-07
Product images deletion fix

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### Before

The problem was, whenever product deletion failed (for example because the variant is already used in a placed order), images files were still removed. The reason for that was, that the `ImagesRemoveListener` were listening on `postRemove` doctrine event, which is called **even if** flush ends with an error.

### Proposed solution

I've refactored `ImagesRemoveListener` to listen on both `onFlush` and `postFlush` events. It saves paths from scheduled to remove entities with `ImageInterface` type and removes them with `ImageUploader` and `CacheManager` when the whole transaction passes (is there any BC Break? cc @pamil).

### Test

During the implementation, I've encountered some problems with images assertions in Behat tests. The problem was, the images files in the cache were not generated while visiting them with mink driver in Behat page. I propose a new service `ImageExistenceChecker`, which is based on checking the existence of a file, rather than visiting checked file URL (which I find a little bit more appropriate). I'm not sure about the implementation - I suppose I could have made too many assumptions about the folders structure (parameter `%__symfony__.kernel.project_dir%/web/` should be for sure changed in Sylius 1.3+) and I would be grateful for any ideas how this service can be improved/simplified.

Nonetheless, it works and fixes a bug, which is definitely a good thing 😄 🐃

### DIFF
--- a/src/Sylius/Behat/Resources/config/services.xml
+++ b/src/Sylius/Behat/Resources/config/services.xml
@@ -51,6 +51,11 @@
 
         <service id="sylius.behat.table_accessor" class="Sylius\Behat\Service\Accessor\TableAccessor" public="false" />
 
+        <service id="sylius.behat.checker.image_existence" class="Sylius\Behat\Service\Checker\ImageExistenceChecker">
+            <argument type="service" id="__symfony__.sylius.liip.filter_service" />
+            <argument type="string">%__symfony__.kernel.project_dir%/web/</argument>
+        </service>
+
         <service id="sylius.behat.paypal_api_mocker" class="Sylius\Behat\Service\Mocker\PaypalApiMocker" public="false">
             <argument type="service" id="sylius.behat.mocker" />
             <argument type="service" id="sylius.behat.response_loader" />
--- a/src/Sylius/Behat/Resources/config/services.xml
+++ b/src/Sylius/Behat/Resources/config/services.xml
@@ -53,7 +53,7 @@
 
         <service id="sylius.behat.checker.image_existence" class="Sylius\Behat\Service\Checker\ImageExistenceChecker">
             <argument type="service" id="__symfony__.sylius.liip.filter_service" />
-            <argument type="string">%__symfony__.kernel.project_dir%/web/</argument>
+            <argument type="string">%__symfony__.sylius_core.public_dir%</argument>
         </service>
 
         <service id="sylius.behat.paypal_api_mocker" class="Sylius\Behat\Service\Mocker\PaypalApiMocker" public="false">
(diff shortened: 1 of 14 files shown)


src/Sylius/Bundle/CoreBundle/EventListener/ImagesRemoveListener.php
```diff
@@ -13,12 +13,16 @@
...
-use Doctrine\ORM\Event\LifecycleEventArgs;
+use Doctrine\ORM\Event\OnFlushEventArgs;
+use Doctrine\ORM\Event\PostFlushEventArgs;
 use Liip\ImagineBundle\Imagine\Cache\CacheManager;
 use Liip\ImagineBundle\Imagine\Filter\FilterManager;
 use Sylius\Component\Core\Model\ImageInterface;
 use Sylius\Component\Core\Uploader\ImageUploaderInterface;
 
+/**
+ * @internal
```
pamil: 👍

src/Sylius/Bundle/CoreBundle/EventListener/ImagesRemoveListener.php
```diff
@@ -30,20 +34,34 @@
...
+        }
+    }
 
-        if ($image instanceof ImageInterface) {
-            $this->imageUploader->remove($image->getPath());
-            $this->cacheManager->remove($image->getPath(), array_keys($this->filterManager->getFilterConfiguration()->all()));
+    public function postFlush(PostFlushEventArgs $event): void
+    {
+        foreach ($this->imagesToDelete as $imagePath) {
+            $this->imageUploader->remove($imagePath);
```
pamil: We should also remove deleted images from `$this->imagesToDelete` array.

pamil: An additional test for that should be nice also (PHP-PM community would be thankful) :)

src/Sylius/Behat/Resources/config/services.xml
```diff
@@ -51,6 +51,11 @@
 
         <service id="sylius.behat.table_accessor" class="Sylius\Behat\Service\Accessor\TableAccessor" public="false" />
 
+        <service id="sylius.behat.checker.image_existence" class="Sylius\Behat\Service\Checker\ImageExistenceChecker">
+            <argument type="service" id="__symfony__.sylius.liip.filter_service" />
+            <argument type="string">%__symfony__.kernel.project_dir%/web/</argument>
```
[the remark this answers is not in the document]
Zales0123: I was looking for this parameter 😄 🎉

src/Sylius/Bundle/CoreBundle/EventListener/ImagesRemoveListener.php
Zales0123: 👍

PR #9912, opened 2018-11-09
[WIP] sylius infection

loic425:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/bin/test-package
+++ b/bin/test-package
@@ -14,6 +14,7 @@ test_package() {
     if [[ "$(composer run-script --list 2> /dev/null | awk '{ print $1 }' | grep -c ^test$)" = "0" ]]; then
         run_phpspec "$1" || exit_code=$?
         run_phpunit "$1" || exit_code=$?
+        run_infection "$1" || exit_code=$?
     else
         run_command "composer test" || exit_code=$?
     fi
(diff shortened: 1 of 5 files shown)


mamazu:
I don't think that infection testing is necessary as those only make it harder to contribute to the project and there are other ways to measure a code coverage.

Zales0123:
I think that it's definitely not something that should be run on our Travis (possible red build until the end of the world 😄) but on the other hand it can be extremely useful for specific components or generally for improving our tests. Maybe we can keep this PR open for testing purposes and think about some kind of cookbook, like "How to infect Sylius with `infection`"? Mutation tests are probably the most radical kind of tests we can deal with and it would be worth to keep in mind that they exist and can be used to improve your code. cc @CoderMaggie

loic425:
@Zales0123 Yes.
Moreover infection running can take a lot of time. But I think we can keep coverage phpspec toots cause it's very useful. We can use an alternative phpspec configuration with code coverage (like Lukaz did on its lchrushiel/beery project) in order to not constraint contributors to install xdebug or phpdbg.

PR #9923, opened 2018-11-13
Use oneline phpdocs for property type info

pamil:


Zales0123:
Thank you, Kamil! +2,099 −6,301 seems great :)

PR #9931, opened 2018-11-15
[Payum] infinite loop on state machine exception fixed

### LINKED ISSUE #9928: Infinite loop in Payum afterCaptureAction
**Sylius version affected**: 1.x

**Description**  
We had server down due to memory leak when requesting after-pay website. After debugging I found that it was caused by this particular extension getting into infinite loop:

https://github.com/tautelis/Sylius/blob/8f4542a2744aada6d66d93e259a09370d2b67829/src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php#L93

The problem is that if `updatePaymentState` method throws an exception `onPostExecute` method is called once again.

**Steps to reproduce**  
1. Screw u you state machine.
2. Go to `sylius_shop_order_after_pay` page

**Possible Solution**  
Catch if context has an exception.

tautelis:
| Q               | A
| --------------- | -----
| Branch?         | 1.0
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9928 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
+++ b/src/Sylius/Bundle/PayumBundle/Extension/UpdatePaymentStateExtension.php
@@ -84,6 +84,10 @@ final class UpdatePaymentStateExtension implements ExtensionInterface
             return;
         }
 
+        if (null !== $context->getException()) {
+            return;
+        }
+
         $context->getGateway()->execute($status = new GetStatus($payment));
         $value = $status->getValue();
         if ($payment->getState() !== $value && PaymentInterface::STATE_UNKNOWN !== $value) {


Zales0123:
It appeared in my notifications that the issue is stale now, so I've reminded about this PR :) of course the PHPSpec is not a point (stupid me from the December 😠), but I still wonder how can we test it to be 100% sure we're fixing a bug? The change does not seem to harm anything, but it should not be a sufficient reason to say it's definitely ok 😄 cc @tautelis @pamil @lchrusciel

Zales0123:
Ok, it's a long time since it was opened... I agree that we should merge it, probably to `1.5` now, even without any tests, but I would like to see one more opinion from @pamil about that. And thank you @tautelis for your patience 🖖

tautelis:
@Zales0123 speaking about tests, there is no structure to perform such test since its an edge case of system working incorrectly and throwing exceptions. It's not possible to test these exceptions with Behat or Spec. The only way would be PHPUnit Functional test which we have none.

alexrowesoap:
This really needs merging in, coverage or not. This was absolutely crippling to locate when the loop it causes means you don't even get a log in Sentry when in production mode.

alexrowesoap:
Just to add, adding the above patch indeed fixed a loop in our UAT environment and successfully the underlying cause: A doctrine error in regards to a key error in a custom table

pamil:
may become messed-up.**

execute `git fetch origin && git reset "issue_9928"` to update your local branch.
 
pamil:
Thanks, Vytautas! :tada:

alexrowesoap:
Thank you both @tautelis and @pamil , with props to @CoderMaggie for the hustle

PR #9967, opened 2018-11-26
Fix for #8759 (ability to specify custom fixture suite during install)

### LINKED ISSUE #8759: Ability to specify fixture suite during sylius:install
When we have some custom fixture suite - steps to install Sylius become not user-friendly. 
We should:

1. Run `php bin/console sylius:install`
2. Answer 'Y' for 'Would you like to reset it?
3. Answer 'N' for 'Loading sample data for environment dev. Continue?
4. Type dummy E-mail and password (they will be rewrited during loading fixtures - next step)
5. Load our custom fixture suite (`php bin/console sylius:fixtures:load CUSTOM_SUITE_NAME --no-interaction`)

Note, that we can't use `--no-interaction` at first step as far as in that case `default` fixture suite will be executed and (in case we have some entities customized) it will fail (like `Some custom fields can't be null`) during creation. Thats why we should perform so many steps and its not good.

May I perform pull request?

igormukhingmailcom:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #8759
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Command/InstallDatabaseCommand.php
+++ b/src/Sylius/Bundle/CoreBundle/Command/InstallDatabaseCommand.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sylius\Bundle\CoreBundle\Command;
 
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Style\SymfonyStyle;
 
@@ -31,6 +32,7 @@ final class InstallDatabaseCommand extends AbstractInstallCommand
 The <info>%command.name%</info> command creates Sylius database.
 EOT
             )
+            ->addOption('fixture-suite', 's', InputOption::VALUE_OPTIONAL, 'Load specified fixture suite during install', null)
         ;
     }
 
@@ -39,6 +41,8 @@ EOT
      */
     protected function execute(InputInterface $input, OutputInterface $output): void
     {
+        $suite = $input->getOption('fixture-suite');
+
         $outputStyle = new SymfonyStyle($input, $output);
         $outputStyle->writeln(sprintf(
             'Creating Sylius database for environment <info>%s</info>.',
@@ -54,6 +58,10 @@ EOT
         $this->runCommands($commands, $output);
         $outputStyle->newLine();
 
-        $this->commandExecutor->runCommand('sylius:install:sample-data', [], $output);
+        $parameters = [];
+        if ($suite) {
+            $parameters['--fixture-suite'] = $suite;
+        }
+        $this->commandExecutor->runCommand('sylius:install:sample-data', $parameters, $output);
     }
 }
--- a/src/Sylius/Bundle/CoreBundle/Command/InstallDatabaseCommand.php
+++ b/src/Sylius/Bundle/CoreBundle/Command/InstallDatabaseCommand.php
@@ -59,7 +59,7 @@ EOT
         $outputStyle->newLine();
 
         $parameters = [];
-        if ($suite) {
+        if (null !== $suite) {
             $parameters['--fixture-suite'] = $suite;
         }
         $this->commandExecutor->runCommand('sylius:install:sample-data', $parameters, $output);
--- a/src/Sylius/Bundle/CoreBundle/Command/InstallSampleDataCommand.php
+++ b/src/Sylius/Bundle/CoreBundle/Command/InstallSampleDataCommand.php
@@ -15,6 +15,7 @@ namespace Sylius\Bundle\CoreBundle\Command;
 
 use Symfony\Component\Console\Helper\QuestionHelper;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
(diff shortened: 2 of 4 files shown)


igormukhingmailcom:
Guys, I also suggest to not ask confirmation question if suite was specified (at https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Command/InstallSampleDataCommand.php#L55) like this:

```php
if (!$questionHelper->ask(
    $input, 
    $output, 
    new ConfirmationQuestion(
        'Continue? (y/N) ', 
        null !== $suite // <------------- Question will be answered YES by default in no interaction mode if suite was provided
    )
)) {
```

This not breaks BC as far as new option/`$suite` is `null` by default (so `null !== $suite ` is `false` by default as it was before), but allows us to run install commands with `--fixture-suite=custom --no-interaction` options and load fixtures automatically (now, if you run install with `--no-interaction` option - sample data loading step will be skipped).

igormukhingmailcom:
Hi, @lchrusciel
I've fixed build fail

src/Sylius/Bundle/CoreBundle/Command/InstallSampleDataCommand.php
```diff
@@ -68,8 +72,13 @@ protected function execute(InputInterface $input, OutputInterface $output): ?int
             return 1;
         }
 
+        $parameters = ['--no-interaction' => true];
+        if ($suite) {
+            $parameters['suite'] = $suite;
```
lchrusciel: Can you make explict comparition here? What do we accept as a proper value? Not empty string?

src/Sylius/Bundle/CoreBundle/Command/InstallDatabaseCommand.php
```diff
@@ -54,6 +58,10 @@ protected function execute(InputInterface $input, OutputInterface $output): void
         $this->runCommands($commands, $output);
         $outputStyle->newLine();
 
-        $this->commandExecutor->runCommand('sylius:install:sample-data', [], $output);
+        $parameters = [];
+        if ($suite) {
```
lchrusciel: Can you make explict comparition here? What do we accept as a proper value? Not empty string?

src/Sylius/Bundle/CoreBundle/Command/InstallSampleDataCommand.php
```diff
@@ -43,16 +45,18 @@ protected function execute(InputInterface $input, OutputInterface $output): ?int
...
+        $suite = $input->getOption('fixture-suite');
 
         $outputStyle = new SymfonyStyle($input, $output);
         $outputStyle->newLine();
         $outputStyle->writeln(sprintf(
-            'Loading sample data for environment <info>%s</info>.',
-            $this->getEnvironment()
+            'Loading sample data for environment <info>%s</info> from suite <info>%s</info>.',
+            $this->getEnvironment(),
+            $suite
```
Zales0123: We should probably have some default value here? `$suite ?? 'default'`?

src/Sylius/Bundle/CoreBundle/Command/InstallDatabaseCommand.php
```diff
@@ -54,6 +58,10 @@ protected function execute(InputInterface $input, OutputInterface $output): void
         $this->runCommands($commands, $output);
         $outputStyle->newLine();
 
-        $this->commandExecutor->runCommand('sylius:install:sample-data', [], $output);
+        $parameters = [];
+        if (null !== $suite) {
```
Zales0123: Do we need to check this? If we pass just `['--fixture-suite' => $suite]` even if suite is null it should still work?

src/Sylius/Bundle/CoreBundle/Command/InstallSampleDataCommand.php
igormukhingmailcom: You're right

src/Sylius/Bundle/CoreBundle/Command/InstallDatabaseCommand.php
igormukhingmailcom: Yes, we need it

igormukhingmailcom:
Hi, @Zales0123 . 
Any change to merge this?

mamazu:
That would indeed be a nice feature to have.

igormukhingmailcom:
@Zales0123 @lchrusciel Could any of you review please?

src/Sylius/Bundle/CoreBundle/Command/InstallSampleDataCommand.php
```diff
@@ -33,6 +34,7 @@ protected function configure(): void
 The <info>%command.name%</info> command loads the sample data for Sylius.
 EOT
             )
+            ->addOption('fixture-suite', 's', InputOption::VALUE_OPTIONAL, 'Load specified fixture suite during install', null)
```
lchrusciel: Won't it be better to have `default` as a default value instead of null? We will reduce one if statement `$suite ?? 'default'`, while `null !== $suite` could be changed to `'default' !== $suite`. 

This will mean, that we cannot load not interactively default fixtures. However, I'm not convinced about this change anyway. The feature is useful, but consequences may be huge, so I would not load it.

lchrusciel:
I'm ok with the changes, thanks a lot for the contribution 👍

Zales0123:
Thank you, Igor! 🏅

Issue #9925, opened 2018-11-14
Parameters.locale conflicts with channels without the specified locale


Peteck:
**Sylius version affected**: < 1.3

**Description**
So I was creating some channels that won't use the locale specified in `parameters.locale` and stuff started breaking.

So far I've found that the web debug toolbar breaks.

**Steps to reproduce**  
- Set `parameters.locale` to some locale
- Use a channel without the specified `parameters.locale`

**Possible Solution**  
Make a `parameters.locale` follow the locale context. This was the behavior of 1.2.

pamil:
Could you send us some screenshots or the error messages that you get? Also, the specific setup you are using would be helpful.

Peteck:
![v1_3_locale_problem](https://user-images.githubusercontent.com/1520592/48557183-55898c80-e8e6-11e8-8e62-82f37f3a54e2.jpg)
This is the response of the web debug toolbar when `parameters.locale` is set to `en_US` and I'm on a channel without `en_US` as a locale.

Peteck:
Updated description because I seem to not having problems with `path()` anymore. I don't know what might have went wrong there. I'll report back if I'll experience the issue again.

Zales0123:
<img width="1440" alt="zrzut ekranu 2018-11-15 o 15 49 37" src="https://user-images.githubusercontent.com/6212718/48560439-16f7d000-e8ee-11e8-9cfe-890be3e6cd8c.png">

Just checked it on `1.2` and it indeed seems to be an issue.

Peteck:
This issue was fixed with pull request #9975

PR #9975, opened 2018-11-28
Ignore locale request restriction for profiler and it's toolbar

Peteck:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #9925 
| License         | MIT

Fixed issue with toolbar and profiler breaking whenever you are on a channel without the locale specified in `parameters.locale`.

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/EventListener/NonChannelLocaleListener.php
+++ b/src/Sylius/Bundle/ShopBundle/EventListener/NonChannelLocaleListener.php
@@ -57,6 +57,10 @@ final class NonChannelLocaleListener
         }
 
         $request = $event->getRequest();
+        if($request->attributes && in_array($request->attributes->get('_route'), ['_wdt', '_profiler'])) {
+            return;
+        }
+
         $currentFirewall = $this->firewallMap->getFirewallConfig($request);
         if (!$this->isFirewallSupported($currentFirewall)) {
             return;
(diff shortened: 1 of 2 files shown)


Zales0123:
may become messed-up.**

execute `git fetch origin && git reset "fix-profiler-toolbar"` to update your local branch.
 
src/Sylius/Bundle/ShopBundle/EventListener/NonChannelLocaleListener.php
```diff
@@ -57,6 +57,10 @@ public function restrictRequestLocale(GetResponseEvent $event): void
         }
 
         $request = $event->getRequest();
+        if($request->attributes && in_array($request->attributes->get('_route'), ['_wdt', '_profiler'])) {
```
Zales0123: Is there any possibility the `$request->attributes` would be false? can it be `null`? I believe it can be an empty `ParameterBag` but then it would be covered with the second condition

Zales0123: And missing space after `if` :)

Peteck: Yes, when running the phpspec tests there will not always be any attributes set. Actually it's only in the one test I made :) I'll add the space.

Zales0123:
Thanks, Peter! :tada:

PR #9988, opened 2018-12-02
#9969 Fix when trying to delete shop user having same ID than logged …

### LINKED ISSUE #9969: Deleting shop user in admin might result in "Cannot remove currently logged in user."
**Sylius version affected**: 1.3.4

**Description**  
If your admin user has the same ID as the shop user and you want to delete the shop user, you get the error "Cannot remove currently logged in user."

**Steps to reproduce**  
1. In the backend login with the default admin (ID: 1) and create one customer (ID: 1) and enable that customer for shop login (setting password, ...)
1. On the customer use "Show" and then "Delete"

This results in: "Cannot remove currently logged in user."

**Possible Solution**  
The issue seems to be that the `\Sylius\Bundle\UserBundle\EventListener\UserDeleteListener` only checks for the same ID but not if they are of the same type, so e.g. the listener is called and compared with the admin object.

laurent35240:
…in admin user

| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9969 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/UserBundle/EventListener/UserDeleteListener.php
+++ b/src/Sylius/Bundle/UserBundle/EventListener/UserDeleteListener.php
@@ -57,7 +57,7 @@ final class UserDeleteListener
 
     private function isTryingToDeleteLoggedInAdminUser(UserInterface $user): bool
     {
-        if (!$user->hasRole('ROLE_ADMINISTRATION_ACCESS')){
+        if (!$user->hasRole('ROLE_ADMINISTRATION_ACCESS') && !$user->hasRole('ROLE_API_ACCESS')){
             return false;
         }
 
--- a/src/Sylius/Bundle/UserBundle/EventListener/UserDeleteListener.php
+++ b/src/Sylius/Bundle/UserBundle/EventListener/UserDeleteListener.php
@@ -14,7 +14,6 @@ declare(strict_types=1);
 namespace Sylius\Bundle\UserBundle\EventListener;
 
 use Sylius\Bundle\ResourceBundle\Event\ResourceControllerEvent;
-use Sylius\Component\Core\Model\AdminUserInterface;
 use Sylius\Component\User\Model\UserInterface;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface;
@@ -58,15 +57,20 @@ final class UserDeleteListener
 
     private function isTryingToDeleteLoggedInAdminUser(UserInterface $user): bool
     {
-        $isTryingToDeleteLoggedInUser = false;
-        if ($user instanceof AdminUserInterface) {
-            $token = $this->tokenStorage->getToken();
-            if ($token) {
-                $loggedUser = $token->getUser();
-                $isTryingToDeleteLoggedInUser = ($loggedUser instanceof AdminUserInterface) && ($loggedUser->getId() === $user->getId());
-            }
+        if (!$user->hasRole('ROLE_ADMINISTRATION_ACCESS')){
+            return false;
         }
 
-        return $isTryingToDeleteLoggedInUser;
+        $token = $this->tokenStorage->getToken();
+        if (!$token) {
+            return false;
+        }
+
+        $loggedUser = $token->getUser();
+        if (!$loggedUser) {
+            return false;
+        }
+
+        return $loggedUser->getId() === $user->getId();
     }
 }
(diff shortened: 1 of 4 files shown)


src/Sylius/Bundle/UserBundle/composer.json
```diff
@@ -35,6 +35,7 @@
         "php": "^7.1",
 
         "doctrine/orm": "^2.5",
+        "sylius/core": "^1.2",
```
lchrusciel: We cannot require core component in others bundle, sorry :(

src/Sylius/Bundle/UserBundle/EventListener/UserDeleteListener.php
```diff
@@ -56,4 +55,18 @@ public function deleteUser(ResourceControllerEvent $event): void
...
+        $isTryingToDeleteLoggedInUser = false;
+        if ($user instanceof AdminUserInterface) {
+            $token = $this->tokenStorage->getToken();
+            if ($token) {
+                $loggedUser = $token->getUser();
+                $isTryingToDeleteLoggedInUser = ($loggedUser instanceof AdminUserInterface) && ($loggedUser->getId() === $user->getId());
+            }
+        }
+
+        return $isTryingToDeleteLoggedInUser;
```
Zales0123: I would say we need a little bit of refactoring with early returns here:

```php
$token = $this->tokenStorage->getToken();
if (!$user instanceof AdminUserInterface || $token === null) {
    return false;
}

if ($loggedUser instanceof AdminUserInterface && $loggedUser->getId() === $user->getId()) {
    return true;
}

return false;
```

wdyt?

Zales0123:
It would be also to have a scenario for that, would you be able to provide it, @laurent35240?

laurent35240:
@lchrusciel @Zales0123 I feel that class `UserDeleteListener` should then be moved to AdminBundle since this delete protection concerns only admin users.

What do you think?

src/Sylius/Bundle/UserBundle/EventListener/UserDeleteListener.php
```diff
@@ -56,4 +55,18 @@ public function deleteUser(ResourceControllerEvent $event): void
...
     }
+
+    private function isTryingToDeleteLoggedInAdminUser(UserInterface $user): bool
+    {
+        $isTryingToDeleteLoggedInUser = false;
+        if ($user instanceof AdminUserInterface) {
+            $token = $this->tokenStorage->getToken();
+            if ($token) {
+                $loggedUser = $token->getUser();
+                $isTryingToDeleteLoggedInUser = ($loggedUser instanceof AdminUserInterface) && ($loggedUser->getId() === $user->getId());
```
Zales0123: Hmm... right now I'm wondering, would it even work? `$loggedUser` would always be `AdminUserInterface`, as we perform delete only from admin panel. We should probably instead check if `$user` is an admin because if it is **and** its ID is the same as logged user id, then deletion should be broken.

laurent35240: I removed the test checking that logged user is `AdminUserInterface`
I check now that `$user` is an admin by checking that he has the role `ROLE_ADMINISTRATION_ACCESS`

For writing a scenario in this case, I am not sure that he can easily be handled with behat because we need to force IDs of Admin User and Customer in database.
That's why I tested this change with phpspec

laurent35240:
@lchrusciel @Zales0123 I removed dependency to `sylius/core` and refacto the part checking if Admin User is trying to delete himself.

Let me know if it is good for you

Zales0123:
Nice 👌 Do you think you would be able to test this feature with some Behat scenario?

laurent35240:
@Zales0123 I do not think I can easily write Behat scenario for this bugfix because I would have to find a way of manipulating IDs for entries in `sylius_admin_user` and `sylius_customer` tables.
That is why I wrote phpspec test for it.

Rest of normal behaviour for deleting admin user or customer through back office was already covered by other behat tests.

There was just this case of deleting a customer with same ID than logged in admin that should be fixed.

BTW, another idea for fixing this issue would be to use UUID instead of ID (by using this doctrine library for example https://github.com/ramsey/uuid-doctrine) but I prefer leave that for another PR or issue ;-)

lchrusciel:
Thanks, Laurent! :1st_place_medal:

loic425:
@lchrusciel @Zales0123 This changes are based on ROLES... This is not a good idea to use UserBundle out of the box (https://travis-ci.org/Monofony/SymfonyStarter/jobs/514593093)

I think we can handle that by comparing user classes with something like that:
```
if (
    (null !== $token)
    && ($loggedUser = $token->getUser())
    && ($loggedUser->getId() === $user->getId())
    && (get_class($loggedUser) === get_class($user))
) {
```

laurent35240:
@loic425 It's true that your solution is more elegant

PR #10002, opened 2018-12-05
Avoid deprecated notice when using symfony/config > 4.2

odolbeau:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

See: https://github.com/symfony/symfony/blob/master/UPGRADE-4.2.md#config

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -42,3 +42,4 @@ parameters:
         # Symfony 4.0 BC layer
         - '/Symfony\\Component\\ExpressionLanguage\\ParserCache\\ParserCacheInterface/'
         - '/Symfony\\Component\\ExpressionLanguage\\ParserCache\\ParserCacheAdapter/'
+        - '/Class Symfony\\Component\\Config\\Definition\\Builder\\TreeBuilder does not have a constructor and must be instantiated without any parameters/'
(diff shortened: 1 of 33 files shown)


odolbeau:
PHPStan is obviously not happy by this change due to non declared arguments in constructor with symfony/config < 4.2.

Should I update this PR to do something like this?
```php
if (method_exists($treeBuilder, 'getRootNode')) {
    $treeBuilder = new TreeBuilder('swarrot');
    $rootNode = $treeBuilder->getRootNode();
} else {
    // BC layer for symfony/config 4.1 and older
    $treeBuilder = new TreeBuilder();
    $rootNode = $treeBuilder->root('swarrot');
}
```

Same result with more lines but at least PHPStan will be happy. :)

pamil:
@odolbeau I think your proposed solution to make PHPStan happy is a good one and will make deprecation removal easier in the future. Could you update the PR with it? :)

odolbeau:
There's still errors with phpstan... :/

Zales0123:
:( The only possible solution I see right now is extracting the creation of `TreeBuilder` and getting its root node to a separate class - and then disabling a PHPStan analysis on it 😄 Maybe it even makes sense, as it's the same logic in each class which is changed in this PR... @odolbeau @pamil wdyt?

pamil:
We can use PHPStan config (`phpstan.neon`) to ignore that error:

```
parameters:
    ignoreErrors:
        - '/Class Symfony\\Component\\Config\\Definition\\Builder\\TreeBuilder does not have a constructor and must be instantiated without any parameters/'
```

odolbeau:
> All checks have passed
\o/
I added an ignore rule in `phpstan.neon` as suggested by @pamil

pamil:
Thanks, Olivier! :tada:

PR #10004, opened 2018-12-05
[Console] Add command for showing available Sylius plugins

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | no
| New feature?    | no?
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

zrzut ekranu 2018-12-5 o 12 04 49

### DIFF
--- a/src/Sylius/Behat/Context/Cli/ShowingAvailablePluginsContext.php
+++ b/src/Sylius/Behat/Context/Cli/ShowingAvailablePluginsContext.php
@@ -50,10 +50,7 @@ final class ShowingAvailablePluginsContext implements Context
         $this->command = $this->application->find('sylius:show-available-plugins');
         $this->tester = new CommandTester($this->command);
 
-        try {
-            $this->tester->execute(['command' => 'sylius:show-available-plugins']);
-        } catch (\Exception $e) {
-        }
+        $this->tester->execute(['command' => 'sylius:show-available-plugins']);
     }
 
     /**
--- a/src/Sylius/Behat/Context/Cli/ShowingAvailablePluginsContext.php
+++ b/src/Sylius/Behat/Context/Cli/ShowingAvailablePluginsContext.php
@@ -1,0 +1,72 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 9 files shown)


src/Sylius/Bundle/CoreBundle/Command/ShowAvailablePluginsCommand.php
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): void
+    {
+        $output->writeln('<comment>Available official Sylius Plugins:</comment>');
+
+        $pluginTable = new TableRenderer($output);
+
+        $pluginTable->setHeaders(['Plugin', 'Description', 'URL']);
+        $pluginTable->addRow(['<info>Admin Order Creation</info>', 'Creating (and copying) orders in the administration panel.', 'https://github.com/Sylius/AdminOrderCreationPlugin']);
```
lchrusciel: IMHO a list of plugins with whole data as an array and for loop with this function would be nicer solution. Even better if we add VO :)

Zales0123: VO `PluginInfo(string $name, string $description, string $url)` would be great 👌 but an array of data would also work well :)

src/Sylius/Behat/Resources/config/suites.yml
```diff
@@ -1,11 +1,13 @@
 # This file is part of the Sylius package.
+# This file is part of the Sylius package.
```
Zales0123: Duplicated :)

src/Sylius/Behat/Context/Cli/ShowingAvailablePluginsContext.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        $this->application = new Application($this->kernel);
+        $this->application->add(new SetupCommand());
+
+        $this->command = $this->application->find('sylius:show-available-plugins');
+        $this->tester = new CommandTester($this->command);
+
+        try {
+            $this->tester->execute(['command' => 'sylius:show-available-plugins']);
+        } catch (\Exception $e) {
```
Zales0123: Should we silence the exception? And `$e` -> `$exception` :)

bartoszpietrzak1994:
@stefandoorn We will consider it in further iterations :)

bartoszpietrzak1994:

Issue #9560, opened 2018-07-05
More flexible password hashing


pamil:
⚠️ 🛣 **This issue is a part of the** [Sylius Public Roadmap](https://github.com/orgs/Sylius/projects/2) 🛣 ⚠️

Our currently used, default password hashing algorithm is `sha512`, which shouldn't be used anymore and it has been already mentioned in issues like #3045.

My proposed solution would be to provide a general framework that would make changing password hashing algorithm much easier. The solution can be as simple as:

 - adding the `hashingAlgorithm` field to user models
 - resolving `Symfony\Component\Security\Core\Encoder\PasswordEncoderInterface` instance used to hash passwords based on that field
 - to provide a seamless upgrade path, rehashing the password using the current hashing algorithm when a user logs in

Since it's a new feature targeted to `master` branch being the next `v1.3.0` release, which also requires at least PHP 7.2, we can jump straight into the world of modern cryptography and use [`libsodium`](https://paragonie.com/book/pecl-libsodium), which is distributed with PHP since 7.2 release.

Further improvements might include adding a grace period after which users are obliged to reset their passwords (so that you wouldn't have to store `sha512`-hashed passwords years after you changed the hashing algorithm), but it is not a part of this RFC.

Checklist:
 - [x] Support for multiple password encoders at once (#10008, #10080)
 - [x] Argon2i used for new users while sha512 kept for old users (#10080)
 - [x] Seamless encoders changes - listening for login event, rehashing the password then (#10084)
 - [x] Integration tests in Sylius as an application (#10084)

teohhanhui:
For the default, we should use `argon2i` as it ships with PHP 7.2.

https://symfony.com/blog/new-in-symfony-3-4-argon2i-password-hasher

psren:
> Further improvements might include adding a grace period after which users are obliged to reset their passwords (so that you wouldn't have to store sha512-hashed passwords years after you changed the hashing algorithm), but it is not a part of this RFC.

Maybe this can be replaced by an "upgrade" command which encodes the already encoded password?

e.g. the pass is stores as hashed sha512 it could "rehash" the password.
and store the hashing algos as csv (e.g. ```hashingAlgorithm``` => ```sha512,argon2i```)
Then a ```ChainEncoder``` could run both ```PasswordEncoderInterface``` in serial.

Benefit is that a one-time upgrade of all passwords would be possible without annoying the users.
Down-Side: Hashing and comparing would be slightly slower.

[...]

PR #10008, opened 2018-12-06
ShopUser class is now EncoderAware to provide more flexibility for ch…

kortwotze:
…oosing how to encode passwords

| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | mentioned in #9560 
| License         | MIT

### DIFF
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -10,7 +10,8 @@ security:
         sylius_shop_user_provider:
             id: sylius.shop_user_provider.email_or_name_based
     encoders:
-        Sylius\Component\User\Model\UserInterface: sha512
+        sylius_default_password_encoder:
+            id: 'sylius.security.user_password_encoder'
     firewalls:
         admin:
             switch_user: true
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -10,8 +10,8 @@ security:
         sylius_shop_user_provider:
             id: sylius.shop_user_provider.email_or_name_based
     encoders:
-        sylius_default_password_encoder:
-            id: 'sylius.security.user_password_encoder'
+        sylius_password_encoder_sha512:
+            id: 'sylius.security.user_password_encoder.sha512'
     firewalls:
         admin:
             switch_user: true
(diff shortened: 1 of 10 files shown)


kortwotze:
Can someone help me finding out which test-configuration to change to add the new encoder from security.yml to the test-environments?

kortwotze:
CC @pamil, because this references his issue.

pamil:
Rebased the PR so that the build should be passing now.

config/packages/security.yaml
```diff
@@ -10,7 +10,8 @@ security:
         sylius_shop_user_provider:
             id: sylius.shop_user_provider.email_or_name_based
     encoders:
-        Sylius\Component\User\Model\UserInterface: sha512
+        sylius_default_password_encoder:
```
pamil: I would name it `sha512` (`default` looks tricky and might break the application if someone suddenly changes the algorithm).

config/packages/security.yaml
```diff
@@ -10,7 +10,8 @@ security:
...
+            id: 'sylius.security.user_password_encoder'
```
pamil: Same here - `sylius.security.user_password_encoder.sha512`?

config/packages/security.yaml
kortwotze: Actually, I would leave the encoder name as is and change the service ID of the "default" encoder. So the default encoder is configured to be the sha512 one

src/Sylius/Bundle/UserBundle/Tests/Functional/app/config/config.yml
```diff
@@ -61,8 +61,6 @@ security:
     encoders:
         sylius_password_encoder_sha512:
             id: 'sylius.security.user_password_encoder.sha512'
-        sylius_password_encoder_argon2i:
```
kortwotze: This is intended, after adding and committing I tried to check if argon2i is actually enabled in my sylius container. Seeing that it was not, I chose not to take any risk and remove it again.

It most likely will come back sooner or later. :)

pamil:
Thanks, Peter! :1st_place_medal:

PR #4583, opened 2016-03-21
[Sylius][ResourceBundle] Fix mapping with interface

gperdomor:
| Q               | A
| --------------- | ---
| Bug fix?        |yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #4447
| License         | MIT
| Doc PR          | -

Remove <code>ResolveDoctrineTargetEntitiesPass</code> and handle interfaces in <code>RegisterResourcesPass</code>

PR #10013, opened 2018-12-07
[Sylius][ResourceBundle] Fix missing service doctrine.orm.listeners.resolve_target_entity

gseidel:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | mentioned in #4583
| License         | MIT

After i add some mappings to ``doctrine.orm.resolve_target_entities`` i got an exception:

**ServiceNotFoundException** The service "doctrine.dbal.default_connection" has a dependency on a non-existent service "doctrine.orm.listeners.resolve_target_entity" in CheckExceptionOnInvalidReferenceBehaviorPass.php line 86

I search for this error i found a comment on the pull request #4583 facing the same error:

> @pjedrzejewski honestly i don't know why fails. I don't know why doctrine.orm.listeners.resolve_target_entity is missing in the container, but all other test are green

After some research i found out, that this error occurs because the service ``doctrine.orm.resolve_target_entities`` is marked as private. The ``CheckExceptionOnInvalidReferenceBehaviorPass`` will check for invalid reference and therefor it check if the container has the mentioned service. 

```
 if (ContainerInterface::EXCEPTION_ON_INVALID_REFERENCE < $value->getInvalidBehavior() || $this->container->has($id = (string) $value)) {
```

I don't know why but the ``DoctrineTargetEntitiesResolverPass`` in combination with  ``doctrine.orm.resolve_target_entities`` leads to check the service and fail. 

With this fix, the ``ServiceNotFoundException`` will not be thrown.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/DependencyInjection/Compiler/DoctrineTargetEntitiesResolverPass.php
+++ b/src/Sylius/Bundle/ResourceBundle/DependencyInjection/Compiler/DoctrineTargetEntitiesResolverPass.php
@@ -31,6 +31,7 @@ final class DoctrineTargetEntitiesResolverPass implements CompilerPassInterface
         try {
             $resources = $container->getParameter('sylius.resources');
             $resolveTargetEntityListener = $container->findDefinition('doctrine.orm.listeners.resolve_target_entity');
+            $resolveTargetEntityListener->setPublic(true);
         } catch (InvalidArgumentException $exception) {
             return;
         }


gseidel:
@pamil I just add a simple interface mapping. 
```
doctrine:
  orm:
    resolve_target_entities:
      App\Entity\Jellybean: App\Entity\GrumpyJellybean
```



I reproduce the bug in a clean symfony/sylius-resource project.

https://github.com/gseidel/sylius-pull-10013

The bug only occures in symfony packages with version `4.2.0` but `4.2.1` works fine. I try to find which change fix the bug for better understanding, but i can't find out which of these packages  
is responsible for that. It seems that this bug is not related to a doctrine package.

https://github.com/gseidel/sylius-pull-10013/blob/master/composer.json#L29-L91

Maybe you have any ideas.

PR #10014, opened 2018-12-07
[BuildFix] Symfony version improvements

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Because of recent changes in `symfony/routing` component, there is some bug with `UrlMatcher` that is losing its mind in out API routing - especially for `index` and `create` which has the same path (but different methods). We should wait for a fix on the Symfony side and use a previous version of `symfony/routing` still.

Thank you @lchrusciel and @nicolas-grekas for help in investigating what's wrong!

I've also changed supported symfony version to `4.1.*`, as `^4.1.1` was allowing even Symfony 4.2, which we have no tests for yet and we cannot guarantee it works properly (I personally had some problems with the **Stopwatch Component** during the update). For sure we need to think about the support for Symfony 4.2 (sooner than later, as the maintenance of Symfony 4.1 is ending soon).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,8 @@
         "symfony/polyfill-intl-icu": "^1.3",
         "symfony/polyfill-mbstring": "^1.3",
         "symfony/swiftmailer-bundle": "^3.0",
-        "symfony/symfony": "^3.4|^4.1.1",
+        "symfony/symfony": "^3.4|4.1.*",
+        "symfony/routing": "^3.4|4.1.8",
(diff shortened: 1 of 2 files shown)


Zales0123:
Closing in favor of https://github.com/Sylius/Sylius/pull/10020

PR #10016, opened 2018-12-08
Column 'position' cannot be null

zspine:
It seems gedmo 'sortable' extension required by this bundle. Otherwise it throws an exception

"Integrity constraint violation: 1048 Column 'position' cannot be null"

| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT


lchrusciel:
Make sense. At least we have it configured that way

lchrusciel:
Thanks, M#3! :tada:

Issue #10015, opened 2018-12-07
Symfony 4.2 support


Zales0123:
During fixing our build (https://github.com/Sylius/Sylius/pull/10014), I've realized that we should think about updating supported Symfony version to 4.2. Looking at these versions official roadmap (https://symfony.com/roadmap/4.1 and https://symfony.com/roadmap/4.2), the maintenance of **Symfony 4.1** end in January '19. On the other hand, **Symfony 4.2** has just been released this month, and will be maintained until July '19.

I really believe we should discuss such changes with our community, to avoid frustration and disappointment as with [PHP version update](https://github.com/Sylius/Sylius/issues/9838) (not being malicious, just empathic :)).

Roshyo:
The disappointment with PHP Version was not because of PHP 7.2 support, but because of PHP 7.1 drop without any obvious reason. As long as Symfony 4.1 is still supported, I believe noone will be disappointed with Symfony 4.2 support.
I also think Symfony 4.1 should be supported actively at least untill end January, and left in supported versions as long as there is no obvious incompatibility.

stefandoorn:
Agree with @Roshyo :-) Obviously project needs to move forward and can't support things forever (also takes time), but e.g. not restricting it would already be a great help. Community can also help on keeping things compatible in case we (for whatever reason) cannot update yet to the latest version. Therefore I also hope S3.4 & PHP7.1 will be allowed for a while, even though if the mainline product moves forward.

liverbool:
@pamil did you see this https://github.com/symfony/symfony/issues/29637

pamil:
Solved by #10067.

liverbool:
@pamil i think this is not solve https://github.com/symfony/symfony/issues/29637 because the Sylius kernel was not used in my app.

pamil:
@liverbool could you open a separate issue for that (as it also happens with symfony/dependency-injection 4.1.8) and it's not connected with Symfony 4.2 support directly?

PR #10020, opened 2018-12-09
[HotFix][BuildFix] Use old PhpMatcherDumper to avoid trailing slash problems

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | replaces https://github.com/Sylius/Sylius/pull/10014
| License         | MIT

The previous proposition did not work, although I tried hard :( Happily, I've investigated wich specific class caused these problems - `PhpMatcherDumper` (it was also the reason of problems in Symfony 4.1.8, but different ones).

As a quick fix, I propose overriding this class in Sylius, until these bugs will be resolved in Symfony itself. It can be also a good excuse to bump Symfony version to 4.2, which is (as far as I know) free of these bugs - see https://github.com/Sylius/Sylius/issues/10015.

**Pros:**
- No Symfony version has to be conflicted, as we override a problematic class (only specific component in one bundle)
- It should work nicely for both Sylius 1.2 and 1.3

**Cons:**
- We have a big, ugly service definition and even bigger and uglier service implementation in our code base 😢

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -22,6 +22,9 @@ parameters:
         - '**MongoDB**'
         - '**ODM**'
 
+        # Temporary PhpMatcherDumper class
+        - 'src/Sylius/Bundle/CoreBundle/Routing/Matcher/Dumper/PhpMatcherDumper.php'
+
     ignoreErrors:
         # Magic calls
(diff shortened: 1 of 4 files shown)


pamil:

PR #10032, opened 2018-12-14
Avoid useless search with minCharacter

pierre-H:
I noticed with the Admin Order Creation Plugin that event when there is no character, a search is done.

| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-auto-complete.js
+++ b/src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-auto-complete.js
@@ -25,6 +25,7 @@
                             search: 250
                         },
                         forceSelection: false,
+                        minCharacters: 1,
                         apiSettings: {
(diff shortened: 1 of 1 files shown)


pamil:
may become messed-up.**

execute `git fetch origin && git reset "patch-2"` to update your local branch.
 
pierre-H:
@Zales0123 I don't understand why the build is failing … That's strange ...

Zales0123:
Do these scenarios work on your machine?

pamil:
The build is failing because the test client does not fill in the field so there are no characters.
Anyway, I think we fixed it with #10404, limiting default results to 25 makes the search much more performant and also allows for selecting a choice without filling the input (which might be helpful eg. if you have not that much taxons in your app).

PR #10041, opened 2018-12-18
FEATURE: Expose order token-value in admin-api

daniellienert:
In order to create a cart using the admin-api, which can later
be picked up with the shop-api, I need to be able to specify
the token value.

| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT


daniellienert:
Hey guys, 

I am pretty new to Sylius and Symfony as well. So I guess I need some guidance on adding this feature. It works / locally tested, but I don't know If i missed something.
Especially I am not sure if a validator is needed for that value.

mamazu:
If you ware just exposing the value (sending it) then no validator is needed. However if you want to get a cart by token value, then a validator would be good to have a nicer error handling.

kortwotze:
I am wondering why you would create a cart in admin-api and then use shop-api to handle it. As far as I know the current shop-api will give you a cart-token (which is an order-token) upon using the GET cart/{tokenValue} request without submitting a tokenValue (since you dont know at the beginning of a new buy-process)

kortwotze:
See this merged PR in Shop-API: https://github.com/Sylius/ShopApiPlugin/pull/338

You can POST to apiurl/carts to get a fresh cart with a token, without having a token beforehand.

daniellienert:
Hey @kortwotze,

Hey thanks for asking!

My scenario is the following: 
On a separate CMS under a separate URL, I add a product to the cart. The CMS uses the admin API to add the items to the Sylius cart. 

When I am logged in that CMS: When I klick on "checkout", I am automatically logged in to Sylius via OAuth single-signon and get my cart there. That works fine.

When I am not logged in, the API adds an anonymous cart. Now I need to assign that cart to my Sylius session when I change to the shop. I wrote a plugin, that does that.

Now, the thing which is left, is the possibility to specify the token-value using the admin API.

Well, I assume that approach sounds a bit weird at first. So If you've got any best practice to work in such a two stack scenario I am pretty open to hear about it.

Cheers,
Daniel

kortwotze:
I feel you - we use Sylius exactly like you - nothing exposed from sylius itself, we have Sulu in front of it and only use shop-api to handle eCommerce things.

So if we want to add something to a cart, we just POST the data to shop-api. Depending on the CMS session, we either know a cart/order token and submit it, or we dont. If we dont, shop-api creates a new cart, adds the item and returns the new cart.

Guest-Mode:
If we are not logged in and add something to a cart, we get a "guest-cart-token" and store it in our cms-session. If we then log in, we submit that  token and we built something in shop-api that puts all items from the submitted cart (and all items from a cart that is already assigned to the logged in customer) and put them into a third cart. then we delete the two old ones and return the new cart token to the CMS which then uses that from there on to interact with the cart.

From where I am standing, there is no need for the admin-api at all. Everything can be done with Shop-API.

daniellienert:
Hey @kortwotze,
thanks fo explaining. Good to hear, that it is not too odd what we are doing.
I also could use the shop API. But is it somehow possible to transfer an order / cart, created with the shop API to a user session in order to use the default Sylius interface / forms?

kortwotze:
If I understand correctly: Yes. If you are logged in (e.g. have a JWT for UserA) and you create a cart with that token and put items in there and so on with API and THEN log into that customer on a sylius page, it actually should just load the last cart for that customer and display it.
No need to merge anything there.

If you have the credentials of your customer, you could "prepare" a cart for them by using those credentials to create a JWT (route: login_check, blabla) and submit that when interacting with the POST /cart endpoint.

Login with that customer on sylius should just do it then.

kortwotze:
If you check the database after you have interacted with a customer cart or created one, it should be assigned to a customer ID. Every cart has a customer ID - and some customers even have a shop_user (which is the entity you log in with).

daniellienert:
There is no problem when I can identify a registered customer. My problem is an "anonymous" session.

daniellienert:
Apart from that, I could need some help with the tests. Seems, the introduced field is somehow marked as required. (While my projects integration tests state, that it is not). How can I configure that field as optional?

daniellienert:
> The CMS puts something in a not-logged-in-cart and then someone hits a Sylius page, logs in and 
 expects that same cart to be there?

Not quite, for anonymous user it works like this:

1. CMS puts data in an anonymous cart using the shop-api
2. User is sent to `https://shop.project.com/pick-cart/<token-value-uuid>`
3. The cart identified by the token value is assigned to the current user session
4. Checkout using the default Sylius interface

daniellienert:
@kortwotze - well everything I wrote above is implemented and works as expected. The missing piece is this PR. So lets come back to that and see how we get it in :)

kortwotze:
Allright - I was just trying to show you a way without using the admin-api. ;)

For the tests (after looking into the failed travis build):
I think you need to check the failed tests, discover which response-file they load and add something like "tokenValue": @@string@@ to it, do designate the fields existence and that it can contain any string.

src/Sylius/Bundle/AdminApiBundle/Form/Type/OrderType.php
```diff
@@ -63,6 +64,10 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                     new NotBlank(['groups' => ['sylius']]),
                 ],
             ])
+            ->add('tokenValue', TextType::class, [
```
pamil: Is it needed for returning the token value?

Zales0123:
Hello @daniellienert, it's been a while, but I think it's still worth to have it merge :) Do you think you would be able to add some tests as @pamil proposed? Thanks! 🚀

PR #10043, opened 2018-12-18
Behat JS scenarios war vol.1

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

This is a first PR from (planned, hope it will work :D) series of PR's refactoring our javascript Behat scenarios.

I think we're all aware of the value that Behat brings to us, but also we're all struggling with the unbelievably slow scenarios containing `@javascript` tag, run by Selenium. Just looking a the execution times can be scary:

- no-js scenarios: 11525 steps, ~6m18.29s
- js scenarios: 2413 steps, ~16m45.64s

I think that eventually, we should think about some faster and more reliable tool for testing our js features. On the other way, I have a deep premonition that we can still do something already, without even changing a testing tool. I've started with two simple things:

- removing `@javascript` tag whenever it's not needed and it requires no or minimum work
- refactoring "rating product" scenarios to use radio buttons rather than semantic-ui stars (why? because clicking on **stars** is _de facto_ testing a UI framework, which we should assume works properly and test a real value. Maybe we should leave one step tagged with `@javascript`, just to be sure that everything works?)

I know, it's just a few tens of seconds right now, but I strongly believe, that even without changing our testing tools we can make build closer to 20min than 30-35min. I investigated most of the places when javascript is required for scenarios and will try to figure out is it relevant or can it be removed with no big changes and no BC breaks :)

### DIFF
--- a/features/account/customer_account/viewing_orders_history/viewing_order_items.feature
+++ b/features/account/customer_account/viewing_orders_history/viewing_order_items.feature
@@ -16,7 +16,7 @@ Feature: Viewing details of an order
         And I addressed it to "Lucifer Morningstar", "Seaside Fwy", "90802" "Los Angeles" in the "United States" with identical billing address
         And I chose "Free" shipping method with "Cash on Delivery" payment
 
-    @ui @javascript
+    @ui
     Scenario: Viewing basic information about an order
         When I view the summary of the order "#00000666"
         And it should has number "#00000666"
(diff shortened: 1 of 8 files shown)


pamil:
Thanks, Mateusz! :tada:

PR #10045, opened 2018-12-19
Add scalar types in Behat/Page/Admin directory

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Million years ago, we introduced scalar types to Sylius code (https://github.com/Sylius/Sylius/issues/8425) and it is a blessing :) However, our (big and strong) Behat-related code is still missing them. I know, they're not so valuable for the test code, but still allows us to reduce amount of unnecessary inaccuracies (and code lines :)). We should do something about it, or we will stay without these PHP7 features until PHP8 😄 

This PR aims to add scalar types only in **admin pages**, which means `src/Sylius/Behat/Page/Admin` directory. Rest of the changes are just required to avoid failing tests. I believe it's doable to introduce scalar types in a reasonable amount of time and energy for the whole Behat suite, but it should be done in separated PR's per section.

To be honest, it was not so bad as I've expected 🐃

### DIFF
--- a/src/Sylius/Behat/Behaviour/ChecksCodeImmutability.php
+++ b/src/Sylius/Behat/Behaviour/ChecksCodeImmutability.php
@@ -17,15 +17,9 @@ use Behat\Mink\Element\NodeElement;
 
 trait ChecksCodeImmutability
 {
-    /**
-     * @return NodeElement
-     */
-    abstract protected function getCodeElement();
+    abstract protected function getCodeElement(): NodeElement;
 
-    /**
-     * @return bool
-     */
-    public function isCodeDisabled()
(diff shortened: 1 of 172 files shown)


pamil:
Thank you, Mateusz! :tada:

Zales0123:
I'm totally aware it's not 100% of coverage 😄 But yes, I also believe that with all these scalar types introduced, people would be more eager to add them when they're missing, during a regular refactoring ;) Thanks!

PR #10059, opened 2018-12-31
Cover specs with PHPStan

pamil:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

Tried to add `proget-hq/phpstan-phpspec` to our PHPStan but ended up reporting https://github.com/proget-hq/phpstan-phpspec/issues/8 and covering our specs with PHPStan and some ignored errors.

This led me to find over ten specs which were not run due to class name/namespace/filename mismatches and some wrong `Argument::` calls in specs.


composer.json
```diff
@@ -190,7 +189,57 @@
...
+            "spec\\Sylius\\Bundle\\PromotionBundle\\": "src/Sylius/Bundle/PromotionBundle/spec/",
+            "spec\\Sylius\\Bundle\\ResourceBundle\\": "src/Sylius/Bundle/ResourceBundle/spec/",
+            "spec\\Sylius\\Bundle\\ReviewBundle\\": "src/Sylius/Bundle/ReviewBundle/spec/",
+            "spec\\Sylius\\Bundle\\ShippingBundle\\": "src/Sylius/Bundle/ShippingBundle",
+            "spec\\Sylius\\Bundle\\ShopBundle\\": "src/Sylius/Bundle/ShopBundle/spec/",
+            "spec\\Sylius\\Bundle\\TaxationBundle\\": "src/Sylius/Bundle/TaxationBundle/spec/",
+            "spec\\Sylius\\Bundle\\TaxonomyBundle\\": "src/Sylius/Bundle/TaxonomyBundle/spec/",
+            "spec\\Sylius\\Bundle\\ThemeBundle\\": "src/Sylius/Bundle/ThemeBundle/spec/",
+            "spec\\Sylius\\Bundle\\UiBundle\\": "src/Sylius/Bundle/UiBundle/spec/",
+            "spec\\Sylius\\Bundle\\UserBundle\\": "src/Sylius/Bundle/UserBundle/spec/"
```
TomasVotruba: These repeated autoloads can be very hard to maintain and keep completed.

For monorepo and `autoload-dev`, the post autoload script might be sustainable solution: https://github.com/Symplify/Symplify/commit/4d9ed30520d05e1feaf180a0bfcd89b8e80a1559

pamil: It looks like a pretty nice idea, but I think I prefer explicit configuration within `composer.json` to implicit one done by some PHP code which almost no one expects. Our list of components and bundles is quite stable, so maintaining won't be an issue here. Thanks for the heads up! :)

TomasVotruba: Sure :) 
Since it's monorepo and this removes the maintenance completely, I see the values in clear `composer.json` overrulling the lost explicitnes.

I got some feedback that monorepo autoloads are pretty WTF, compared to single psr-4 lines.

PR #10070, opened 2019-01-05
#9699 Fix for viewing products when they belong to a taxon and to one…

### LINKED ISSUE #9699: Product pagination bug
**Sylius version affected**: every

**Description**  
https://demo.sylius.com/en_US/taxons/t-shirts So we request 9 products per page, but only 6 are visible, the others are on the second page. My guess is this should not happen.

**Steps to reproduce**  
Visit https://demo.sylius.com/en_US/taxons/t-shirts

**Possible Solution**  
Dunno.

laurent35240:
… of children taxon

| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9699 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
@@ -70,6 +70,7 @@ class ProductRepository extends BaseProductRepository implements ProductReposito
         array $sorting = []
     ): QueryBuilder {
         $queryBuilder = $this->createQueryBuilder('o')
+            ->distinct()
             ->addSelect('translation')
             ->innerJoin('o.translations', 'translation', 'WITH', 'translation.locale = :locale')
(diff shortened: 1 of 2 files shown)


lchrusciel:
Yup. I prefer to ensure, that we are not adding any regression. However, there should be some tests for that anyway.

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
```diff
@@ -70,6 +70,7 @@ public function createShopListQueryBuilder(
         array $sorting = []
     ): QueryBuilder {
         $queryBuilder = $this->createQueryBuilder('o')
+            ->distinct()
```
pamil: It's so simple when you see it, but couldn't figure it out when I was trying to solve this issue, good job 🎉

pamil:
Thank you, Laurent! :tada:

Issue #9999, opened 2018-12-04
Documentation about Shipping Component are out of date


Konafets:
The docs about Shipping Component is outdated and misleading. Additional it lacks on proper explanation of the concepts.

Issues are:
* non existent classes are mentioned as described in #9855.
* default Shipment state is wrong https://docs.sylius.com/en/latest/components_and_bundles/components/Shipping/state_machine.html
* The available states are wrong. See states at `vendor/sylius/sylius/src/Sylius/Component/Shipping/Model/ShipmentInterface.php`
* Class `ShipmentItem` does not exist

Furthermore the concept of all the classes are not explained. 

* What is a shipment? Is it a package? 
* What is a ShippingCategory? Why I need one?
* What is a ShippingMethod? Do I need one?
* What is a ShipmentUnit?

cansozeri:
@Zales0123 The main problem in this project is the documentation, there is a great effort to programming but the documentation is weak. It could be hard to take a time to a learning curve because of a weak documentation. I hope you gonna improve the docs in the near future :)

Zales0123:
Thank you for your opinion, @cansozeri :) I believe that the worst part of our documentation is **Components & Bundles** section, which is the most outdated and incomplete... On the other way, e.g. "The Book" or "The Cookbook" are kept in a relatively good condition. 
The thing is, we, as a Core Team, have plenty of things to do with this project and documentation is not always our top priority (I know it's sometimes should). And that's why contributions from the community, especially to the documentation, are a huge help for us :) Moreover, I believe sometimes people outside of the Core Team have a better insight on where are the documentation weaknesses, as you use it more than us 😄

We will for sure try to fix documentation step by step, but we also strongly encourage you to contribute to it. Thanks! :)

cansozeri:
@Zales0123 Thank you for your efforts, I have to understand what is going on in this shipment component and bundle first before to contribute to the docs :) I want to use it in my project but because of documentation I have to dig in the codes :) but I need time for it, if I can, I would be very happy to contribute to it ..

Konafets:
I can second what @cansozeri wrote. To describe a behavior of f.e. the shipping component you have to understand it first. But if the documentation is wrong, the only thing someone can do is dive into the code and do trial and error. The problem here is: Writing a documentation from guessing does not help anybody.

How about the idea to have a documentation hackathon or similar, where interested ppls from the community are in a room with the core team to discuss and document?

lchrusciel:
[...]

PR #10073, opened 2019-01-07
[Docs] Bundles turbo refactoring

CoderMaggie:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Doc fix?        | yes
| Related tickets | fixes all issues regarding outdated bundle docs
| License         | MIT

Due to our bundle docs being highly outdated we have decided to remove them and approach leaving only technical, non-ecommerce bundle docs here. The useful articles from the removed bundle docs will come back in cookbooks as soon as possible.

We were wondering who and why is nowadays using Sylius ecommerce bundles standalone, and we thought that a much better place for bundle documentation is _each bundle's readme file._


stefandoorn:
This thing is rather double. I can understand it's outdated and takes significant time to update. On the contrary, I think there's very useful information in there. Adoption of the framework is highly based on the quality of documentation in my opinion. 

Not sure what's the solution though, as I understand the core team cannot support such piece of documentation.

CoderMaggie:
@stefandoorn That's true and we are sorry for this move. We are getting very few contributions to the docs, we are currently incapable of updating them, and it's better not to have them at all than to have them outdated. We are doing the same thing with components. 
As I wrote in the description, we will be recommending to put Bundle docs into bundles' readmes.

4c0n:
I also highly value the existing documentation. Although it's outdated, I still manage to find useful information in it all the time.

Maybe it is a good idea to require documentation updates as a condition to get pull requests merged. If that is not the case already. That way it should stay up to date, but does bring in a bit more overhead.

stefandoorn:
Makes sense @CoderMaggie. Alternative: is there a method to mark doc pages as 'outdated' asking community to support in updates? Then anyone can see it might be outdated, but still have access to the information. Not sure it's possible in the documentation tooling though.

Konafets:
I proposed the idea to do an documentation hackathon (#9999) but so far now answer. The problem with contributing to the doc are in my case that I have no idea what the intended is. This makes it hard to describe anything. At a hackathon the ways are short and ppls can discuss things. 

I consider removing them now completely is not the right direction to go. Instead you should go through the doc and remove only those parts which are obviously outdated.

CoderMaggie:
Closing in favor of #10100

PR #10076, opened 2019-01-08
[Docs] Remove vagrant references

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

The vagrant box is outdated and not maintained. I will not find time in the nearest future to pick it up. I want to remove Vagrant references, so it will not be misleading for newcomers. Also, if there is some good spirit, who would like to maintain our Vagrant box, we would be more than happy to transfer the reposiotry itself. Otherwise it will be abandon/read-only.

Also, I owe they a :beer: ;)


pamil:
Thanks, Łukasz! :tada:

prigal:
Was reviewing 1.3.7 changelog and found this. If needed, you can send users to Homestead like Symfony do in its documentation. I have tested it yesterday for a quick plugin install on a clean sylius standard and it just work.

PR #10079, opened 2019-01-09
[Channel] Shop billing data

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | associated with https://github.com/Sylius/RefundPlugin/issues/106, https://github.com/Sylius/InvoicingPlugin/issues/52
| License         | MIT

During the development of `InvoicingPlugin`, we introduced a concept of shop billing data (set per channel) that is required to be displayed on an invoice. On the other hand, we also should have it in a `RefundPlugin` to generate a credit memo properly.

We considered a few options, but in the end, it's quite a common use case to have such a bunch of data manageable in Admin panel.

zrzut ekranu 2019-01-9 o 11 04 31

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/ShopBillingData.orm.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/ShopBillingData.orm.xml
@@ -21,11 +21,11 @@
             <generator/>
         </id>
 
-        <field name="company" />
-        <field name="taxId" />
-        <field name="countryCode" />
-        <field name="street" />
-        <field name="city" />
-        <field name="postcode" />
+        <field name="company" nullable="true" />
+        <field name="taxId" nullable="true" />
+        <field name="countryCode" nullable="true" />
+        <field name="street" nullable="true" />
+        <field name="city" nullable="true" />
+        <field name="postcode" nullable="true" />
     </entity>
 </doctrine-mapping>
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/ShopBillingData.orm.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/ShopBillingData.orm.xml
@@ -22,8 +22,8 @@
         </id>
 
         <field name="company" nullable="true" />
-        <field name="taxId" nullable="true" />
-        <field name="countryCode" nullable="true" />
+        <field name="taxId" column="tax_id" nullable="true" />
+        <field name="countryCode" column="country_code" nullable="true" />
         <field name="street" nullable="true" />
         <field name="city" nullable="true" />
         <field name="postcode" nullable="true" />
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/ShopBillingData.orm.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/ShopBillingData.orm.xml
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+
+ This file is part of the Sylius package.
+
+ (c) Paweł Jędrzejewski
+
+ For the full copyright and license information, please view the LICENSE
+ file that was distributed with this source code.
+
+-->
+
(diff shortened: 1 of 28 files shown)


src/Sylius/Component/Core/Model/ShopBillingDataInterface.php
```diff
@@ -0,0 +1,43 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\Model;
+
+interface ShopBillingDataInterface
+{
+    public function getId(): ?int;
```
pamil: I think we shouldn't enforce this method, as it serves no purpose - shop billing data should be always assigned to either a channel, an invoice or a credit memo.

src/Sylius/Component/Core/Model/ChannelInterface.php
```diff
@@ -60,4 +60,8 @@ public function setSkippingPaymentStepAllowed(bool $skippingPaymentStepAllowed):
     public function isAccountVerificationRequired(): bool;
 
     public function setAccountVerificationRequired(bool $accountVerificationRequired): void;
+
+    public function getShopBillingData(): ?ShopBillingDataInterface;
```
pamil: If all fields are not required, we might enforce returning an instance of ShopBillingDataInterface so that there's no need for an additional check for nullability.

src/Sylius/Component/Core/Model/ShopBillingData.php
```diff
@@ -0,0 +1,104 @@
...
+    /** @var string|null */
+    private $street;
+
+    /** @var string|null */
+    private $city;
+
+    /** @var string|null */
+    private $postcode;
+
+    public function getId(): ?int
```
pamil: Let's remove this getter also if it's not required by the interface.

src/Sylius/Component/Core/Model/ShopBillingData.php
```diff
@@ -0,0 +1,104 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\Model;
+
+/** @final */
+class ShopBillingData implements ShopBillingDataInterface
+{
+    /** @var int|null */
+    private $id;
```
pamil: I guess we should make all properties protected as in other models.

pamil:
Thank you, Mateusz! :tada:

prigal:
@Zales0123 after some time thinking about the model and how we will do it on our side, I suggest you some modifications for the core PR.

Adding billing data to channel is not really correct, as long as a channel is a "way for selling goods". A same company can have several channels : online shop, API, mobile app, physical store, ... that all share the same billing data.

We can duplicate data, but in my opinion we should add a higher level in Sylius : a "company" entity. The company will store the billing data, and we will select per channel which is the company that "run" the channel.

When it comes to exporting data to a accounting software / ERP. You will need to export it "per company" and not really "per channel".

tautelis:
@Zales0123 several questions:
* why is the mapping is not mapped-superclass? now the entity is not extendible
* why this new entity does not use default Sylius entity management and is not defined in CoreBundle's Configuration.php ?

src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/ShopBillingData.orm.xml
```diff
@@ -0,0 +1,31 @@
...
+ file that was distributed with this source code.
+
+-->
+
+<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
+                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping
+                                      http://doctrine-project.org/schemas/orm/doctrine-mapping.xsd">
+
+    <entity name="Sylius\Component\Core\Model\ShopBillingData" table="sylius_shop_billing_data">
```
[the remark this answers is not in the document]
tautelis: @Zales0123 several questions:

- why is the mapping is not mapped-superclass? now the entity is not extendible
- why this new entity does not use default Sylius entity management and is not defined in CoreBundle's Configuration.php ?

Issue #9654, opened 2018-08-15
Plugin's template cannot be overriden by theme


jacquesbh:
**Sylius version affected**: 1.2.3

**Description**  

If you want to override a plugin's template (a *plugin*, not a *bundle*), and you want to use the theme for that, it's not working.

**Steps to reproduce**  

Let's say you install the [BitBagCms](https://github.com/BitBagCommerce/SyliusCmsPlugin) plugin.

Then you want to override the template `@BitBagSyliusCmsPlugin/Shop/Page/show.html.twig`.
And you want to do tat in your theme.

So you create this file:

`app/themes/MyTheme/BitBagSyliusCmsPlugin/Shop/Page/show.html.twig`

And according to everything it should work. But it doesn't.

**Possible Solution**  

The bug is actually in <https://github.com/Sylius/Sylius/blob/1.2/src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php#L65-L75>.

I did that to fix the problem (very succinct):

```diff
diff --git a/src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php b/src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php
index 0d7fbb1f5b..22191df875 100644
--- a/src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php
+++ b/src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php
@@ -66,8 +66,14 @@ final class TemplateNameParser implements TemplateNameParserInterface
             return $this->decoratedParser->parse($name);
         }
 
+        if (isset($matches[1]) && preg_match('/^.+Plugin$/', $matches[1])) {
+            $bundleName = $matches[1];
+        } else {
+            $bundleName = $matches[1] ? $matches[1] . 'Bundle' : '';
+        }
+
         $template = new TemplateReference(
-            $matches[1] ? $matches[1] . 'Bundle' : '',
+            $bundleName,
             $matches[2],
             $matches[3],
             $matches[4],
```

It works well since we call the templates with the `Plugin` prefix:

```
@BitBagSyliusCmsPlugin/Shop/Page/show.html.twig
                ^^^^^^
```

chaenu:
Hello,
is this still true for version 1.3.3? I tried to overwrite templates for the CMS plugin as well, but failed. I wonder if I'm doing something wrong or if it's still this bug? Couldn't make it work with above fix either.

[...]

PR #10082, opened 2019-01-10
[Theme] Allow overriding templates from plugins (1.2.*)

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/9654
| License         | MIT

It's hard to test this change with some functional test in a current tests structure, as it would require to use `SyliusPluginTrait` in test bundle/plugin, but it's in a core bundle :/ I have some idea have to test it properly with a Behat scenario, but I need to spend a few more minutes on that :)

It would be nice if someone could incorporate these changes into their application and see does it work as expected (cc @jacquesbh :)). I've of course done that, but _better safe than sorry_ 🐃 

This change only applies to `1.2` branch, for Sylius `^1.3` take a look here: https://github.com/Sylius/Sylius/pull/10083.

### DIFF
--- a/src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php
+++ b/src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php
@@ -57,7 +57,7 @@ final class TemplateNameParser implements TemplateNameParserInterface
         }
 
         $template = new TemplateReference(
-            $matches[1] ? $matches[1] . 'Bundle' : '',
+            $this->getBundleOrPluginName($matches[1]),
             $matches[2],
             $matches[3],
             $matches[4],
@@ -73,5 +73,14 @@ final class TemplateNameParser implements TemplateNameParserInterface
         }
 
         return $this->cache[$name] = $template;
+    }
+
+    private function getBundleOrPluginName(string $name): string
+    {
+        if (preg_match('/^.+Plugin$/', $name)) {
+            return $name;
+        }
+
+        return $name ? $name . 'Bundle' : '';
     }
 }
--- a/src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php
+++ b/src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php
@@ -77,7 +77,7 @@ final class TemplateNameParser implements TemplateNameParserInterface
 
     private function getBundleOrPluginName(string $name): string
     {
-        if (preg_match('/^.+Plugin$/', $name)) {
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php
```diff
@@ -74,4 +74,13 @@ public function parse($name): TemplateReferenceInterface
 
         return $this->cache[$name] = $template;
     }
+
+    private function getBundleOrPluginName(string $name): string
+    {
+        if (preg_match('/^.+Plugin$/', $name)) {
```
lchrusciel: oh 🗡

src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php
```diff
@@ -74,4 +74,13 @@ public function parse($name): TemplateReferenceInterface
...
+            return $name;
+        }
+
+        return $name ? $name . 'Bundle' : '';
```
pamil: Do we need this ternary operator here?

Zales0123: Ok, in fact we need, as it's explained in `it_generates_template_references_from_root_namespaced_paths` specification

src/Sylius/Bundle/ThemeBundle/Templating/TemplateNameParser.php
```diff
@@ -57,7 +57,7 @@ public function parse($name): TemplateReferenceInterface
         }
 
         $template = new TemplateReference(
-            $matches[1] ? $matches[1] . 'Bundle' : '',
+            $this->getBundleOrPluginName($matches[1]),
```
jacquesbh: This can generate a notice. You have to check the existence of the index `1`.

image

jacquesbh: Hum, according to the regex above, you should not process this code if `$matches[1]` doesn't exist, since it's not matching in that case.

pamil: Actually, there should always be `$matches[0-5]` if the `preg_match` returns anything `>=1` (if not, the `if` above has an early return), so we can assume that they exist. IIRC that regexp is copied from Symfony source code.

pamil:

PR #10083, opened 2019-01-10
[Theme] Allow overriding templates from plugins (^1.3)

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes (kinda) https://github.com/Sylius/Sylius/issues/9654
| License         | MIT

This is, of course, a quick fix and quite straight-forward, but it works 😄 And it is still required to use `@SyliusTestPlugin` notation (maybe it's worth to be documented somehow?). This PR fixes the functionality for Sylius `^1.3`, for working fix on `1.2` take a look here: https://github.com/Sylius/Sylius/pull/10082.

Regarding testing it better (as I've also written in fix for `1.2`):

> It's hard to test this change with some functional test in a current tests structure, as it would require to use `SyliusPluginTrait` in test bundle/plugin, but it's in a core bundle :/ I have some idea have to test it properly with a Behat scenario, but I need to spend a few more minutes on that :)

### DIFF
--- a/src/Sylius/Bundle/ThemeBundle/Locator/BundleResourceLocator.php
+++ b/src/Sylius/Bundle/ThemeBundle/Locator/BundleResourceLocator.php
@@ -85,15 +85,23 @@ final class BundleResourceLocator implements ResourceLocatorInterface
     private function locateResourceBasedOnTwigNamespace(string $resourcePath, ThemeInterface $theme): string
     {
         $twigNamespace = substr($resourcePath, 1, strpos($resourcePath, '/') - 1);
-        $bundleName = $twigNamespace . 'Bundle';
         $resourceName = substr($resourcePath, strpos($resourcePath, '/') + 1);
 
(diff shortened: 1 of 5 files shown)


src/Sylius/Bundle/ThemeBundle/Locator/BundleResourceLocator.php
```diff
@@ -85,15 +85,23 @@ private function locateResourceBasedOnBundleNotation(string $resourcePath, Theme
...
         if ($this->filesystem->exists($path)) {
             return $path;
         }
 
         throw new ResourceNotFoundException($resourcePath, $theme);
     }
+
+    private function getBundleOrPluginName(string $twigNamespace): string
+    {
+        if (preg_match('/Plugin$/', $twigNamespace)) {
```
pamil: Could we incorporate the same change as in #10082?

pamil:

PR #10094, opened 2019-01-13
Overriding plugin templates in themes tests

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | https://github.com/Sylius/Sylius/pull/10082, https://github.com/Sylius/Sylius/pull/10083
| License         | MIT

I can imagine this PR can be a little bit controversial 😄 but I hope it can be a nice step forward to make `ThemeBundle` more reliable.

As there is no easy way to test overriding plugin templates inside `ThemeBundle` itself (as it would require us to use trait from `CoreBundle`), I thought it would be nice to implement such a "test plugin" for Behat tests. It's maybe not the most elegant way, but it tests themes connected with plugins quite well and I don't see having a test plugin only for scenarios a bad thing, as we already have [test themes](https://github.com/Sylius/Sylius/tree/master/src/Sylius/Bundle/ThemeBundle/Tests/Fixtures/themes) or [test bundles](https://github.com/Sylius/Sylius/tree/master/src/Sylius/Bundle/ThemeBundle/Tests/Functional/TestBundle) in our test suite.

I was also thinking, that maybe such a test should be implemented on our https://github.com/Sylius/PluginSkeleton, but I really consider overriding plugins template as a `ThemeBundle` feature, rather than plugins... possibility 🚀


src/Sylius/Behat/Resources/config/services/pages.xml
```diff
@@ -15,6 +15,7 @@
     <imports>
         <import resource="pages/admin.xml" />
         <import resource="pages/shop.xml" />
+        <import resource="pages/test-plugin.xml" />
```
pamil: We usually use underscores instead of hyphens for config files in Sylius.

pamil:
Thank you, Mateusz! :tada:

PR #10114, opened 2019-01-23
Added productTaxon to product list query to prevent an SQL error

yohang:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fix bug introduced in #10070
| License         | MIT

#10070 has introduced a SQL Error when displaying products by taxon (`for SELECT DISTINCT, ORDER BY expressions must appear in select list`).

This is due to the ordering by ProductTaxon position, as they are not in the `SELECT` clause.

One possible solution is to simply add the ProductTaxon to the `SELECT` DQL clause, as done in this PR.


pamil:
It breaks one scenario, could you have a look at the build?

yohang:
@pamil Yes, saw it. I'll have a look and fix it.

pamil:
Thank you, we fixed the build and included your commit in #10136. 🎉

PR #10115, opened 2019-01-24
[WIP] Fix deprecation for getExtendedTypes

isometriks:
| Q               | A
| --------------- | -----
| Branch?         | all
| Bug fix?        | fix deprecation sf4
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Form/Extension/CollectionTypeExtension.php
+++ b/src/Sylius/Bundle/ResourceBundle/Form/Extension/CollectionTypeExtension.php
@@ -48,4 +48,12 @@ final class CollectionTypeExtension extends AbstractTypeExtension
     {
         return CollectionType::class;
     }
+    
+    /**
+     * @inheritdoc
+     */
+    public static function getExtendedTypes()
+    {
+        return [CollectionType::class];
+    }
 }


isometriks:
Ah there's actually a bunch more I could add this too, will update this PR tomorrow.

Zales0123:
Hello @isometriks, sorry for keeping this without an answer for such a long time 🖖 I know the `getExtendedType()` function is deprecated since Symfony **4.2**, but we still support also Symfony **3.4**, so if it's not required I don't think we should add that 😕 It should be for sure recommended in our documentation, but I believe there are already some works being done in this case :)

isometriks:
No problem, I meant to add more to it but never got around to it. It would actually stay BC compatible, and would remove deprecation notices in Symfony 4.2+. Symfony 4 actually throws the notice if this method doesn't exist, not if the old one still does, so keeping both until you removed support for 3.4 would work for both: https://github.com/symfony/symfony/blob/4.2/src/Symfony/Component/Form/DependencyInjection/FormPass.php#L96

Although it does appear that if you still have the tag for `extended-type` it won't use the static method anyway so I guess all it would do is suppress the deprecation.

PR #10132, opened 2019-01-30
Add missing Length constraint on product translation slug

venyii:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no (?)
| Deprecations?   | no
| Related tickets | -
| License         | MIT

This previously caused database errors when trying to insert data. Discovered this while working with the API, but this may also occur when trying to save a product in the admin GUI and the slug generator converts one character into > 1 characters like `ß` -> `ss`.

>SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'slug' at row 1

About the BC breaks yes/no: I guess someone could have increased the field length for the slug which this would prevent after updating. I personally don't see it as a BC break, but it may be useful to put a "warning" note in the changelog to change the constraint if you modified anything. WDYT?

### DIFF
--- a/src/Sylius/Bundle/ProductBundle/Resources/config/validation/ProductTranslation.xml
+++ b/src/Sylius/Bundle/ProductBundle/Resources/config/validation/ProductTranslation.xml
@@ -40,6 +40,11 @@
                 <option name="message">sylius.product.slug.not_blank</option>
                 <option name="groups">sylius</option>
             </constraint>
+            <constraint name="Length">
+                <option name="max">255</option>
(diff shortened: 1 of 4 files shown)


pamil:
Thank you, @venyii! :1st_place_medal:

pamil:
As for the note, the PR title would be included in the changelog and if someone modified it so far, I bet they also have an automatic test for it. It won't crash any shop with increased length limit, though it might prevent updating a product, so it's not critical.

Issue #7389, opened 2017-02-01
Problem with get products list by taxon on MySQL 5.7


4oXdev:
When I testing Sylius with mysql 5.7, there is an error in Reports:

request.INFO: Matched route "sylius_shop_product_index". {"route":"sylius_shop_product_index","route_parameters":{"_controller":"sylius.controller.product:indexAction","_sylius":{"template":"@SyliusShop/Product/index.html.twig","grid":"sylius_shop_product"},"slug":"aem-induction","_route":"sylius_shop_product_index"},"request_uri":"http://xxxx/taxons/xxxx","method":"GET"} []

[...]

PR #10136, opened 2019-01-31
[HotFix] 500 on taxons list error fix

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | introduced in https://github.com/Sylius/Sylius/pull/10070, replaces https://github.com/Sylius/Sylius/pull/10114 and https://github.com/Sylius/Sylius/pull/10134, fixes https://github.com/Sylius/Sylius/issues/7389
| License         | MIT

Based on a PR by @yohang (thank you for your work!). The test was failing, as there is another bug with setting max results with paginator (that's why I've reduced the amount of products in the scenario) 🚀 😄 But the 500 error bug is fixed (I've also checked it manually).

### DIFF
--- a/features/product/viewing_products/viewing_products_from_taxon_and_children.feature
+++ b/features/product/viewing_products/viewing_products_from_taxon_and_children.feature
@@ -23,9 +23,6 @@ Feature: Viewing products from taxon children
         And the store has a product "T-Shirt Watermelon" available in "Poland" channel
         And this product belongs to "T-Shirts"
         And this product belongs to "Men"
-        And the store has a product "T-Shirt Lemon" available in "Poland" channel
-        And this product belongs to "T-Shirts"
-        And this product belongs to "Men"
 
     @ui
     Scenario: Viewing products from taxon children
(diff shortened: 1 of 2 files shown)


Zales0123:
And thank you @yohang! 🎉

witrin:
@Zales0123 This breaks the ShopApiPlugin. See https://github.com/Sylius/ShopApiPlugin/blob/master/src/ViewRepository/Product/ProductCatalogViewRepository.php#L100, https://github.com/Sylius/ShopApiPlugin/blob/master/src/Factory/Product/ProductViewFactory.php#L79 and for example https://travis-ci.org/Sylius/ShopApiPlugin/jobs/488729007#L853.

I've patched it several days ago for one of my projects. It works but it's ugly:

```
->addSelect('productTaxon.position as HIDDEN position')
```

After updating to 1.3.8 I'm screwed may be I'll copy the previous version and keep it until this is solved. ;)

PR #10077, opened 2019-01-08
Fix select attributes according to recent Symfony form changes

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | related to https://github.com/symfony/symfony/pull/29307
| License         | MIT

Due to the latest Symfony 4.1.10/3.4.21 release and changes introduced in linked PR, a bug in select attribute type araised 🌤 Apparently, possible to select values in attribute configuration was saved with `TextType` form type, even though `SelectAttributeValueTranslationsType` should be used. It was somehow parsed before, but know it's not possible and this PR fixes this absurdity 🖖

PR #10141, opened 2019-02-01
[1.1] Fix select attributes according to recent Symfony form changes

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | related with https://github.com/Sylius/Sylius/pull/10077
| License         | MIT

Even though we don't support Sylius 1.1 yet, it would be nice to have the build passing 😄 This version should still get security fixes and it can be hard to determine if this security fix does not break anything if the build would be failing anyway 🔐 👁


pamil:

Issue #10148, opened 2019-02-04
Expired cart remover efficiency


stefandoorn:
Hi,

Last week I ran into an issue where the removing of expired carts was causing issues. Basically due to my error an SQL error occurred on some orders (wrong cascading of null values). Anyway, as the flush in such commands is done at the end, none of the expired carts get removed. Basically it starts to build up and up, never removing any carts anymore. After a while, the amount of carts to be processed is so high that the command can also run out of memory.

I'm not sure what is the best solution here. Calling `flush()` after every removal in `src/Sylius/Bundle/OrderBundle/Command/RemoveExpiredCartsCommand.php` is probably also not a good solution. Obviously I'm working on fixing my error, but memory issues can occur anyway if people only setup such commands later in their projects and loading an huge amount of orders.

wadjeroudi:
@stefandoorn I think the removal should be handled by pure SQL. And maybe notify only the ids removed not the loaded cart entities.

stefandoorn:
I'm not sure this fully solves the issue @pamil. The index makes it more efficient, but it still can run out of memory if too many carts are loaded. Let's say somebody didn't schedule this command for 2 years and then wants to use it. That becomes rather hard. It probably has to move the expiry date month by month to get to a workable solution I'm afraid.

nicogommen:
Hi,

We have the same issue.
We don't have any error, but just a lot of traffic.
We have the cronjob set up every night, to remove carts that have been idle for 2 days.
But in 2 days, we have too many carts and the command can't handle it... and always run out of memory.
We now have more than 9 000 000 carts in the database...
We're going to manually remove the items with an SQL query, but the issue will happen again very soon, so it would be nice to optimize the command like @stefandoorn suggested (https://github.com/Sylius/Sylius/issues/10148#issuecomment-460176496).

Thanks.

Zales0123:
I'll reopen it then, to trigger a discussion :) If you @stefandoorn or @nicogommen have any possible, better solution for the problem, PR would be very welcomed 😄 🖖

wadjeroudi:
@nicogommen did you make the SQL queries ?

nicogommen:
Hi @wadjeroudi, 

Yes, we did the query manually => delete the items from the carts' table for the carts older than 2 days (I can't remember the exact query).
Then we ended up increasing the frequency of execution of the sylius:remove-expired-carts command.
We put it every hours, so there's way less items to delete each time it's executed.

[...]

PR #10149, opened 2019-02-04
Add index on order.cart + order.updated_at for faster expired cart removal selection

stefandoorn:
| Q               | A
| --------------- | -----
| Branch?         | 1.2
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially #10148 
| License         | MIT

Makes selection of expired carts way more efficient. Especially helpful in growing projects.

Before:

image

After:

image

### DIFF
--- a/app/migrations/Version20190204092544.php
+++ b/app/migrations/Version20190204092544.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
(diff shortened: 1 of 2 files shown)


pamil:
We only add new migrations in minor releases, so that you could update your patch version anytime and don't care about schema changes.

pamil:
Thanks, Stefan! :tada:

PR #10156, opened 2019-02-06
Fix recent Composer deprecations

pamil:
`Deprecation warning: require-dev.mikey179/vfsStream is invalid, it should not contain uppercase characters. Please use mikey179/vfsstream instead. Make sure you fix this as Composer 2.0 will error.`


Zales0123:

PR #10169, opened 2019-02-12
Don't fail on billing or shipping address not set ver.2

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | replaces https://github.com/Sylius/Sylius/pull/7381
| License         | MIT

This is an old issue, I've rebased the previous PR and added some test. Thank you @DmitriyTrt for your work!

zrzut ekranu 2019-02-12 o 09 52 48
zrzut ekranu 2019-02-12 o 09 56 01

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/_addresses.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/_addresses.html.twig
@@ -1,14 +1,18 @@
-<h4 class="ui top attached styled header">
-    {{ 'sylius.ui.shipping_address'|trans }}
-</h4>
-<div class="ui attached segment" id="shipping-address">
-    {% include '@SyliusAdmin/Common/_address.html.twig' with {'address': order.shippingAddress} %}
-</div>
-<h4 class="ui attached styled header">
+{% if order.shippingAddress is not null %}
+    <h4 class="ui top attached styled header">
+        {{ 'sylius.ui.shipping_address'|trans }}
+    </h4>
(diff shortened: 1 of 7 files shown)


src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/_addresses.html.twig
```diff
@@ -1,17 +1,21 @@
...
+{% endif %}
+{% if order.billingAddress is not null %}
+    <h4 class="ui attached styled header{% if order.shippingAddress is null %} top{% endif %}">
+        {{ 'sylius.ui.billing_address'|trans }}
+    </h4>
+    <div class="ui attached segment" id="billing-address">
+        {% include '@SyliusAdmin/Common/_address.html.twig' with {'address': order.billingAddress} %}
     </div>
+{% endif %}
+<div class="ui attached segment" id="edit-addresses">
```
pamil: Shouldn't we hide it when no addresses are set?

pamil:

src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/_addresses.html.twig
Zales0123: I was wondering about that, but I can imagine that SOMEHOW addresses could not be set by mistake... and then we want to be able to set them in the Admin panel, I suppose 😄

Issue #10179, opened 2019-02-18
Extract technical bundles and components to standalone packages


pamil:
Sylius has a wide range of technical components and bundles, which can be used standalone and aren't specific to eCommerce domain, like:

 - [x] FixturesBundle (#10182)
 - [x] ThemeBundle (#10184)
 - [x] Registry (#10188)
 - [x] Mailer / MailerBundle (#10325)
 - [x] Grid / GridBundle (#10326)
 - [x] Resource / ResourceBundle (#10327)

There are a few problems related to their maintenance and development:

 - **Release cycle is the same as for Sylius eCommerce Framework**: While this is great for other eCommerce packages, it would be nice to be able to release a minor version with new features more often than three times a year. 
We've missed autowiring by class when adding support for it in autogenerated services by ResourceBundle, which might come handy in some cases and there's https://github.com/Sylius/Sylius/issues/10003#issuecomment-455350768 reporting it. Though it could be solved pretty quickly, it got lost in other issues and we need to wait for v1.5.0 to release a version helping with that. We lose the agility to react to your feedback fast.

 - **Long build times on CI server**: Our Travis CI build tests the whole application, builds the documentation and tests all the packages, which takes a lot of time. This is why some contributors might find it discouraging when they have to wait so long. We could try to improve the performance of our testing suite, but we'd never go down to a few minutes.

- **Tough to test with custom setup**: If a package needs some specific setup (eg. MongoDB in ResourceBundle), we need to either set it up for all the packages or skip that and "trust" that it works. Adding additional checks or customising the testing logic is not straightforward - https://github.com/Sylius/Sylius/blob/master/bin/install-package#L70-L72. 
Moreover, splitting those packages would allow us to increase the code quality easily (use a different level of static analysis) and require to write better tests (we no longer could rely on E2E Sylius tests that much).

- **Hard to introduce new people to the project**: It's much easier to introduce new people to libraries that are small and aren't made of over 50 standalone components and an application combining all of them. The small packages wouldn't be that overwhelming and it'd make onboarding easier, especially for community contributors and maintainers.

My idea is to move the development and maintenance of these packages to their own repositories. Eg. for ThemeBundle it would happen in https://github.com/Sylius/SyliusThemeBundle. With the split, there are also some new challenges:

 - Documentation needs to be hosted standalone (Github Pages should be OK)
 - A new release cycle needs to be defined (semantic versioning + feature based releases)
[...]

PR #10182, opened 2019-02-18
Extract FixturesBundle

pamil:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | related to #10179 
| License         | MIT

- Removed FixturesBundle source code from `src`
- Made it required in `composer.json`
- Documentation not removed yet, waiting for standalone one to be up and running

Targeting the `1.1` branch, because it's the last one supported (with security fixes). During upmerging, `composer.json` requirements need appropriate customisation` to always require the same minor version of the package:

 - 1.1 - `~1.1.15`
 - 1.2 - `~1.2.15`
 - 1.3 - `~1.3.8`
 - 1.4 - `~1.4.0`


stefandoorn:
Probably good to add a link to the new repository/packagist in UPGRADE e.g.?

pamil:
@stefandoorn there's no end-user action required, so I guess `CHANGELOG-1.3.md` with `TL;DR` (also included in release notes) should be enough, what do you think?

stefandoorn:
Ah, true. Was thinking about plugins etc., but they rely on Sylius also and therefore no issue.

PR #10199, opened 2019-02-26
Allowing options to be given with resource[0].id syntax

Roshyo:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | potentially ?
| Deprecations?   | no
| Related tickets | 
| License         | MIT

When customizing grids quite a lot, we might get a queryBuilder returning an array of array instead of an array of entities. this array is built like this : 
```php
$result = [
    0 => new Resource(),
    'custom_value' => 8,
    'other_custom_value' => 'aaa',
]
```

And for this reason, it might be usefull to allow this kind of option parsing.

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/Grid/Parser/OptionsParser.php
+++ b/src/Sylius/Bundle/ResourceBundle/Grid/Parser/OptionsParser.php
@@ -71,6 +71,10 @@ final class OptionsParser implements OptionsParserInterface
             return $this->parseOptionResourceField(substr($parameter, 9), $data);
         }
 
+        if (0 === strpos($parameter, 'resource[')) {
+            return $this->parseOptionResourceField(substr($parameter, 8), $data);
+        }
+
         return $parameter;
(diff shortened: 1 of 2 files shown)


Zales0123:
ping @Roshyo :)

Roshyo:
Oh yeah indeed. I totally forgot this.

What sort of tests should I add ?
I'm afraid Behat will be hard to cover this. It would require creating a whole new grid, with a whole new page and a custom queryBuilder.

Zales0123:
Some unit test in `OptionsParserSpec` at least, we can optionally think about something more sophisticated later :)

Roshyo:
Here we are.

Roshyo:
ping @pamil @Zales0123

lchrusciel:
Thanks, Stephane! :tada:

Roshyo:
Thanks to you !

PR #10209, opened 2019-03-04
[Shop] Use first variant image on a cart page

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | replaces https://github.com/Sylius/Sylius/pull/8788, fixes https://github.com/Sylius/Sylius/issues/8787
| License         | MIT

This is a resurrection of linked PR, with some Behat tests added. I'm not very happy with the scenario implementation, as it requires overwriting the whole `ImageUploader` in a test environment, to keep uploaded image name - to give availability to check it in the scenario :/
IMO the best option would be to extract some service that would provide a new name for the uploaded image. I can do it in this or separate PR... What do you think? cc @Sylius/core-team @castler @gabiudrescu

### DIFF
--- a/src/Sylius/Behat/Context/Ui/Shop/CartContext.php
+++ b/src/Sylius/Behat/Context/Ui/Shop/CartContext.php
@@ -409,6 +409,14 @@ final class CartContext implements Context
         Assert::same($this->summaryPage->getCartTotal(), $total);
     }
 
+    /**
+     * @Then /^(\d)(st|nd|rd|th) item in my cart should have "([^"]+)" image displayed$/
+     */
+    public function itemShouldHaveImageDisplayed(int $itemNumber, string $image): void
+    {
+        Assert::contains($this->summaryPage->getItemImage($itemNumber), $image);
+    }
+
     private function getPriceFromString(string $price): int
     {
         return (int) round((float) str_replace(['€', '£', '$'], '', $price) * 100, 2);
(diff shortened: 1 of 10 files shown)


gabiudrescu:
I have no say in this, as long as it fixes https://github.com/Sylius/Sylius/issues/8787.

I mentioned this, so there will be a link from this PR to my initial bug report.

src/Sylius/Behat/Context/Ui/Shop/CartContext.php
```diff
@@ -409,6 +409,14 @@ public function myCartSTotalShouldBe($total)
         Assert::same($this->summaryPage->getCartTotal(), $total);
     }
 
+    /**
+     * @Then /^(\d)st item in my cart should have "([^"]+)" image displayed$/
```
lchrusciel:
```suggestion
     * @Then /^(\d)(st|nd|rd|th) item in my cart should have "([^"]+)" image displayed$/
```

src/Sylius/Bundle/ShopBundle/Resources/views/Product/_info.html.twig
```diff
@@ -1,7 +1,11 @@
 {% set product = variant.product %}
 
 <div class="ui header">
-    {% include '@SyliusShop/Product/_mainImage.html.twig' with {'product': product, 'filter': 'sylius_shop_product_tiny_thumbnail'} %}
+    {% if variant.hasImages %}
+        {% include '@SyliusShop/Product/_mainImage.html.twig' with {'product': variant, 'filter': 'sylius_shop_product_tiny_thumbnail'} %}
```
lchrusciel: This `product` variable is not the best naming since we are passing `variants` as well :( Could we make some unification?

lchrusciel:

PR #10213, opened 2019-03-05
Fix product form submit

### LINKED ISSUE #6548: Creating Product press enter to submit form does nothing - Easy Pick
If you press enter to submit the form on a create simple product page. It gives a js error which stops the page from processing the form. 

Issue is on this line in product-attributes.js (line 58 i believe)

$(this).parent().find('select').val().forEach(function(item) {
data += 'sylius_product_attribute_choice[]=' + item + "&";
                });

If i click submit button it's fine. But only happens when i press enter on a form field.

kulczy:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | 
| Related tickets | fixes #6548
| License         | MIT

the problem was that the browser treats the button without the given type as submit, so after clicking enter it activated not this form


pamil:
Thank you, Szymon! :tada:

PR #10217, opened 2019-03-08
[ResourceBundle] Make it possible to autowire services without interfaces

loic425:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

It allows to custom your repositories and your factories without creating an interface and keep benefits of autowiring.

```
sylius_resource:
    resources:
        app.book:
            classes:
                model: App\Entity\Book
                factory: App\Factory\BookFactory
                repository: App\Repository\BookRepository
```

It aliases:

- **App\Factory\BookFactory** with **app.factory.book**
- **App\Repository\BookRepository** with **app.repository.book**

### DIFF
--- a/src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
+++ b/src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
@@ -94,6 +94,8 @@ abstract class AbstractDriver implements DriverInterface
         $factoryClass = $metadata->getClass('factory');
         $modelClass = $metadata->getClass('model');
 
+        $container->setAlias($factoryClass, $metadata->getServiceId('factory'));
+
         $definition = new Definition($factoryClass);
         $definition->setPublic(true);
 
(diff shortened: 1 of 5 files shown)


src/Sylius/Bundle/ResourceBundle/DependencyInjection/Driver/AbstractDriver.php
```diff
@@ -94,6 +94,8 @@ protected function addFactory(ContainerBuilder $container, MetadataInterface $me
         $factoryClass = $metadata->getClass('factory');
         $modelClass = $metadata->getClass('model');
 
+        $container->setAlias($factoryClass, $metadata->getServiceId('factory'));
```
pamil: What would happen if there are two resources which use the same factory class for them? Could you include that scenario in the test suite and see what happens?

loic425: @pamil The second factory alias replace the first one. I've just added a test to check that.

pamil: Just to bring this discussion on Slack here, we figured out the solution to be named autowiring like `Factory $bookFactory` and `Factory $authorFactory`, not `Factory $whateverFactory`.

pamil:
Since ResourceBundle has been extracted, I created https://github.com/Sylius/SyliusResourceBundle/pull/95 and co-authored commits with you there. I've also resolved the issue discussed in comments.

PR #10224, opened 2019-03-12
Allow filtration of Admin API ProductOption

Acrylangel:
Added the grid configuration to admin api ProductOption to allow filtering

| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminApiBundle/Resources/config/routing/product_option.yml
+++ b/src/Sylius/Bundle/AdminApiBundle/Resources/config/routing/product_option.yml
@@ -7,6 +7,7 @@ sylius_admin_api_product_option:
         alias: sylius.product_option
         section: admin_api
         serialization_version: $version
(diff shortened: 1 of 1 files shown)


Zales0123:
@Acrylangel ping, it would be still good to have it merged :)

vvasiloi:
@Zales0123 whould it still be good, considering there's the new API?

Zales0123:
Definitely, we don't want to develop the _old_ API right now, just focus on the _new_ one. The idea is to cover all the Behat scenarios that we have implemented already with the UI, to ensure features parity.

HOWEVER

I see no problem in merging upcoming contributions from the Community, even if they're for the part of the application that will be slowly left out 🖖

PR #10283, opened 2019-04-02
[UserBundle] Fix user comparaison on user delete listener

### LINKED ISSUE #9969: Deleting shop user in admin might result in "Cannot remove currently logged in user."

loic425:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | Fixed #9969  Mentioned #9988 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/UserBundle/EventListener/UserDeleteListener.php
+++ b/src/Sylius/Bundle/UserBundle/EventListener/UserDeleteListener.php
@@ -44,7 +44,7 @@ final class UserDeleteListener
 
         Assert::isInstanceOf($user, UserInterface::class);
 
-        if ($this->isTryingToDeleteLoggedInAdminUser($user)) {
+        if ($this->isTryingToDeleteLoggedInUser($user)) {
             $event->stopPropagation();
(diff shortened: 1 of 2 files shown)


lchrusciel:
Thanks, Loïc! :tada:

loic425:
@lchrusciel You're welcome! :)

PR #10295, opened 2019-04-09
[ChannelPricing] Fixed channelPricing not blank validation

hatem20:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Removing price from a simple product or variant can cause not null exception if product is disabled in channel.

### DIFF
--- a/features/product/managing_products/product_validation.feature
+++ b/features/product/managing_products/product_validation.feature
@@ -143,6 +143,15 @@ Feature: Products validation
         Then I should be notified that slug has to be unique
         And product with code "7-WONDERS-BABEL" should not be added
 
+    @ui
+    Scenario: Trying to remove price from an existing simple product while disabling channel
+        Given the store has a "Dice Brewing" product
+        And I want to modify this product
+        When I remove its price from channel "Web"
+        And I disable it in channel "Web"
+        And I try to save my changes
+        Then I should be notified that price cannot be blank
+
     @ui @javascript
     Scenario: Trying to add a new product with a text attribute without specifying its value in default locale
         When I want to create a new simple product
--- a/src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
+++ b/src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
@@ -977,4 +977,31 @@ final class ManagingProductsContext implements Context
             $this->updateConfigurableProductPage,
         ]);
     }
+
+    /**
+     * @When I remove its price from channel :channel
+     */
+    public function iRemoveItsPriceFromChannel(ChannelInterface $channel)
+    {
+        $this->updateSimpleProductPage->specifyPrice($channel->getName(), '');
+    }
+
+    /**
+     * @When I disable it in channel :channel
+     */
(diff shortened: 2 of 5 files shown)


pamil:
may become messed-up.**

execute `git fetch origin && git reset "channel_pricing_validation"` to update your local branch.
 
features/product/managing_products/product_validation.feature
```diff
@@ -143,6 +143,15 @@ Feature: Products validation
...
         And product with code "7-WONDERS-BABEL" should not be added
 
+    @ui
+    Scenario: Trying to remove price from an existing simple product while disabling channel
+        Given the store has a "Dice Brewing" product
+        And I want to modify this product
+        When I remove its price from channel "Web"
+        And I disable it in channel "Web"
+        And I try to save my changes
+        Then I should be notified that price cannot be blank
```
pamil: I think I'm missing something here - if a product is disabled in a given channel, then the price for that channel shouldn't be required.

hatemBBN: true, but when a product has a price in channel, 
then you disable product in this channel and remove its price 
a notNullException is thrown. 
are you suggesting that in this case channelPricing record should be removed instead of being updated with null @pamil ?

hatem20: true, but when a product has a price in channel,
then you disable product in this channel and remove its price
a notNullException is thrown.
are you suggesting that in this case channelPricing record should be removed instead of being updated with null @pamil ?

pamil: When is that NotNullException thrown?

hatemBBN: create a simple product and enable it for web channel and give it price in this channel 
after a product  is created, disable it from channel and remove price (set it empty)
then click save changes.
without this PR, a NotNullException should be thrown.

hatem20: create a simple product and enable it for web channel and give it price in this channel
after a product is created. update it by disable it from channel and remove price (set it empty)
then click save changes.
without this PR, a NotNullException should be thrown.

src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
```diff
@@ -977,4 +977,31 @@ private function resolveCurrentPage()
             $this->updateConfigurableProductPage,
         ]);
     }
+
+    /**
+     * @When I remove its price from channel :channel
+     */
+    public function iRemoveItsPriceFromChannel(ChannelInterface $channel)
```
lchrusciel: `public` functions should be placed before `private`. Also, do we have similar functions for `specifyPrice` and `checkChannel`? Maybe we can unify calls.

features/product/managing_products/product_validation.feature
Zales0123: I believe it's a description of the incorrect behavior we have in Sylius right now :( We should prevent this `NotNullException` for sure, but also give a possibility to remove the product price if it's possible

PR #10308, opened 2019-04-14
Fix base locale

### LINKED ISSUE #10307: Fixtures: Avoid using `$baseLocale` at `LocaleFixture`
**Sylius version affected**: any

**Description**  
`LocaleFixture` adding `$baseFixture` to list of fixtures to load.
This not allow us to load `locale:` fixtures second time:

```bash
An exception occurred while executing 'INSERT INTO sylius_locale (code, created_at, updated_at) VALUES (?, ?, ?)' with params ["en_US", "2019-04-12 18:10:00", "2019-04-12 18:10:00"]:  
                                                                                                                                                                                          
  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'en_US' for key 'UNIQ_7BA1286477153098'         
```

**Steps to reproduce**  
- Load default fixtures suite
- Load another suite with `orm_purger: false` and `locale:` fixture

**Possible Solution**  
- Avoid using `$baseLocale`. 
- Specify basic locale right at `fixtures.yml`:

```yaml
sylius_fixtures:
    suites:
        default:
            listeners:
                orm_purger: ~
                logger: ~
            fixtures:
                locale:
                    options:
                        locales:
                            - "%locale%"
```

I will make PR if you guys OK with this.
Let me know please if I can do this on `1.4` or even `1.3` rather than `master` - I'm not sure how strict BC compatibility should be...

igormukhingmailcom:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | not sure (it backward compatible on default fixtures configuration)
| Deprecations?   | not sure (deprecation messages was added, but not appear on default fixtures configuration)
| Related tickets | fixes #10307, partially #7346
| License         | MIT

So, in BC in mind, with this PR it now will work like this:

- **If you have default (empty) configuration, as was before**:

    ```yaml
    sylius_fixtures:
        # ...
        fixtures:
            locale: ~
    ```
  Then, base locale **will be added** to configuration.

- *But if you provided a list of locales*:

    ```yaml
    sylius_fixtures:
        suites:
            my_custom_suite_to_add_to_existing_database:
                listeners:
                    orm_purger: false
                fixtures:
                    locale:
                        options:
                            locales:
                                - "de_DE"
    ```
   Then, base locale **will NOT be added** to configuration.

Deprecation message was added to avoid usage base locale at all in future. To suppress that message on default configuration, `"%locale%"` was passed directly to the `locales` option at `src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures.yml`.

# TODO

- [ ] Define Sylius version where base locale will be removed OR remove deprecation message if base locale should be kept by default (when fixture defined like `locale: ~`) - I'm not 100% sure about this so want to discuss this first.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/LocaleFixture.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/LocaleFixture.php
@@ -45,9 +45,7 @@ class LocaleFixture extends AbstractFixture
      */
     public function load(array $options): void
     {
-        $localesCodes = array_merge([$this->baseLocaleCode], $options['locales']);
-
-        foreach ($localesCodes as $localeCode) {
+        foreach ($options['locales'] as $localeCode) {
             /** @var LocaleInterface $locale */
             $locale = $this->localeFactory->createNew();
 
@@ -73,6 +71,21 @@ class LocaleFixture extends AbstractFixture
     protected function configureOptionsNode(ArrayNodeDefinition $optionsNode): void
     {
         $optionsNode
+            ->beforeNormalization()
+                ->ifEmpty()
+                ->then(function(){
+                    @trigger_error(
+                        'Base locale deprecated since 1.3. Please, pass %locale% directly to locales fixture at fixtures option.',
+                        E_USER_DEPRECATED
+                    );
+
+                    return [
+                        'locales'=>[
+                            $this->baseLocaleCode
+                        ]
+                    ];
+                })
+            ->end()
             ->children()
                 ->arrayNode('locales')
                     ->scalarPrototype()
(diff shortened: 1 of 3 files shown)


GSadee:
may become messed-up.**

execute `git fetch origin && git reset "fix-base-locale"` to update your local branch.
 
lchrusciel:
Thank you, Igor! :tada:

pamil:
This is actually a BC break and should be reverted, we need to figure out a solution which allows you not to load the default locale and prevents throwing an error if the default locale is defined explicitly. 

I used to do a project, in which fixtures would break after this change because we removed the default locale from the explicit list but still wanted it to be created.

pamil:
Created #10341 which reverts it.

pamil:
Idea by @lchrusciel:

Add a boolean option `load_default_locale` (or `prepend_default_locale`) which is true by default. Disabling it will stop LocaleFixture from adding the default locale.

lchrusciel:
Ref. #10342

PR #10331, opened 2019-04-23
[Shop] Update grid action and filter keys to decouple shop from admin

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/config/app/config.yml
+++ b/src/Sylius/Bundle/ShopBundle/Resources/config/app/config.yml
@@ -9,10 +9,10 @@ imports:
 sylius_grid:
     templates:
         action:
-            show: "@SyliusUi/Grid/Action/show.html.twig"
-            pay: "@SyliusShop/Account/Order/Grid/Action/pay.html.twig"
+            shop_show: "@SyliusShop/Grid/Action/show.html.twig"
+            shop_pay: "@SyliusShop/Account/Order/Grid/Action/pay.html.twig"
         filter:
-            string: "@SyliusUi/Grid/Filter/string.html.twig"
+            shop_string: "@SyliusShop/Grid/Filter/string.html.twig"
 
 sylius_shop:
     checkout_resolver:
--- a/src/Sylius/Bundle/ShopBundle/Resources/config/app/config.yml
+++ b/src/Sylius/Bundle/ShopBundle/Resources/config/app/config.yml
@@ -2,6 +2,8 @@
 # (c) Paweł Jędrzejewski
 
 imports:
+    - { resource: "@SyliusUiBundle/Resources/config/app/config.yml" }
+
     - { resource: "@SyliusShopBundle/Resources/config/app/sylius/sylius_mailer.yml" }
     - { resource: "@SyliusShopBundle/Resources/config/grids/account/order.yml" }
     - { resource: "@SyliusShopBundle/Resources/config/grids/product.yml" }
--- a/src/Sylius/Bundle/UiBundle/Resources/config/app/config.yml
+++ b/src/Sylius/Bundle/UiBundle/Resources/config/app/config.yml
@@ -1,0 +1,20 @@
+sylius_grid:
+    templates:
+        action:
+            default: "@SyliusUi/Grid/Action/default.html.twig"
+            create: "@SyliusUi/Grid/Action/create.html.twig"
+            delete: "@SyliusUi/Grid/Action/delete.html.twig"
+            show: "@SyliusUi/Grid/Action/show.html.twig"
+            update: "@SyliusUi/Grid/Action/update.html.twig"
+            apply_transition: "@SyliusUi/Grid/Action/applyTransition.html.twig"
+            links: "@SyliusUi/Grid/Action/links.html.twig"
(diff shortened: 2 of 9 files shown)


src/Sylius/Bundle/ShopBundle/Resources/config/app/config.yml
```diff
@@ -9,10 +9,10 @@ imports:
 sylius_grid:
     templates:
         action:
-            show: "@SyliusUi/Grid/Action/show.html.twig"
-            pay: "@SyliusShop/Account/Order/Grid/Action/pay.html.twig"
+            shop_show: "@SyliusShop/Grid/Action/show.html.twig"
```
pamil: I think it's a BC break - anyone using this action in their own grid will be affected.

I'd recommend introducing new actions with those prefixes and deprecating the current Twig templates from `@SyliusUi` - we'll probably need to bump up Twig requirements, see https://twig.symfony.com/doc/2.x/tags/deprecated.html.

lchrusciel: IMHO, the main place where we can consider BC break is here: https://github.com/Sylius/Sylius/pull/10331/files#diff-09570101558a8ee92c483bda1a32d2e0R54

That's true, it may affect all users, that turned off AdminBundle. That's why I have postulated moving from `AdminBundle` all configuration coupled to `UiBundle` to `UiBundle` itself.

GSadee: I've moved basic grid actions and filters to `UiBundle`, now both `AdminBundle` and `ShopBundle` are importing config from `UiBundle`. If I think correctly, there is no BC break right now.

Zales0123: I think now the problem is, `SyliusUiBundle` does not depend (and should not) on `SyliusGridBundle`, so having a `sylius_grid` configuration in `UiBundle/Resources/config/app/config.yml` would not work if you use `UiBundle` as a standalone package... The question is, should it even be possible to use it without the rest of the Sylius, as it provides strictly Sylius-related UI :) But still, it is possible now, so the problem exists.

GSadee: I do not see much sense in using `UiBundle` as a standalone package, what is more, I've checked that before and `UiBundle`  is coupled with `GridBundle` by for example using [twig filters from `GridBundle`](https://github.com/Sylius/Sylius/blob/1.3/src/Sylius/Bundle/UiBundle/Resources/views/Grid/_default.html.twig#L22). So here nothing changes imo 😃

Zales0123: I don't see any sense as well, I just say it's possible 💃 If the incorrect coupling is there anyway, I would be 👌 with the proposed changes, but we should also keep in mind to either couple the `UiBundle` with `GridBundle` in `composer.json` or (IMO better choice) don't allow to use it as standalone. 🖖

GSadee: For now, I've coupled `UiBundle` with `GridBundle` in `composer.json`

lchrusciel:
Thank you, Grzegorz! :tada:

src/Sylius/Bundle/UiBundle/Resources/config/app/config.yml
```diff
@@ -0,0 +1,20 @@
...
+            create: "@SyliusUi/Grid/Action/create.html.twig"
+            delete: "@SyliusUi/Grid/Action/delete.html.twig"
+            show: "@SyliusUi/Grid/Action/show.html.twig"
+            update: "@SyliusUi/Grid/Action/update.html.twig"
+            apply_transition: "@SyliusUi/Grid/Action/applyTransition.html.twig"
+            links: "@SyliusUi/Grid/Action/links.html.twig"
+            archive: "@SyliusUi/Grid/Action/archive.html.twig"
+        filter:
+            string: "@SyliusUi/Grid/Filter/string.html.twig"
+            boolean: "@SyliusUi/Grid/Filter/boolean.html.twig"
```
lchrusciel: We could also sort this config alphabetically

src/Sylius/Bundle/ShopBundle/Resources/config/app/config.yml
```diff
@@ -2,17 +2,19 @@
...
+
     - { resource: "@SyliusShopBundle/Resources/config/app/sylius/sylius_mailer.yml" }
     - { resource: "@SyliusShopBundle/Resources/config/grids/account/order.yml" }
     - { resource: "@SyliusShopBundle/Resources/config/grids/product.yml" }
 
 sylius_grid:
     templates:
         action:
-            show: "@SyliusUi/Grid/Action/show.html.twig"
-            pay: "@SyliusShop/Account/Order/Grid/Action/pay.html.twig"
```
pamil: This renaming is still a BC break, we should leave `pay`, deprecate its template and add `shop_payz to follow the convention.

PR #10340, opened 2019-04-26
Fix #9646 by removing lambdas in JS file

tchapi:
| Q               | A
| --------------- | -----
| Branch?         | 1.1
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9646 again
| License         | MIT

Using lambdas on the `1.1` branch does not work since the `1.1` branch uses **uglifyJS** without Babel, hence, ES6 notation cannot be used.

Without this PR, you cannot run `GULP_ENV=prod npm run gulp` on the `1.1` branch, and thus you cannot build the assets needed for the backoffice.

This PR reverts the lambdas to "regular" functions, unbreaking the asset building for this branch.

Please do not hesitate if I missed some guidelines for this PR
Best

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/private/js/sylius-lazy-choice-tree.js
+++ b/src/Sylius/Bundle/AdminBundle/Resources/private/js/sylius-lazy-choice-tree.js
@@ -155,13 +155,13 @@
                 checkboxElement.checkbox({
                     onChecked: function() {
                         var value = checkboxElement.data('value');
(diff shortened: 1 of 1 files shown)


pamil:
We support the `1.1` branch with security fixes only, but since I'm in the process of extracting technical bundles to their standalone repositories which involves this branch, I can make an exception with this PR and merge it.

lchrusciel:
Thanks, tchap! :tada:

PR #10350, opened 2019-05-02
fix default repository for variant and association type resources

loic425:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ProductBundle/DependencyInjection/Configuration.php
+++ b/src/Sylius/Bundle/ProductBundle/DependencyInjection/Configuration.php
@@ -13,6 +13,8 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\ProductBundle\DependencyInjection;
 
+use Sylius\Bundle\ProductBundle\Doctrine\ORM\ProductAssociationTypeRepository;
+use Sylius\Bundle\ProductBundle\Doctrine\ORM\ProductVariantRepository;
 use Sylius\Bundle\ProductBundle\Form\Type\ProductAssociationType;
 use Sylius\Bundle\ProductBundle\Form\Type\ProductAssociationTypeTranslationType;
 use Sylius\Bundle\ProductBundle\Form\Type\ProductAssociationTypeType;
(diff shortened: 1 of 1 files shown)


Zales0123:
Thanks, Loïc! :1st_place_medal:

lchrusciel:
Was there any discussion behind this bug? Loic, can you explain briefly, where did you encounter a problem?

I'm not saying, that it is a bad solution, I don't know why we want to change this configuration. The problem is that the same configuration is true for 55 other repositories
image
and all repositories where configured dynamically in:
https://github.com/Sylius/SyliusResourceBundle/blob/92c8e365a7c15aa436429b07dd14e7b2f8965f2f/src/Bundle/DependencyInjection/Driver/Doctrine/DoctrineORMDriver.php#L38-L69
or if overridden in YAML:
https://github.com/Sylius/Sylius/blob/c56d62eae80638f961073eaaeb477afc50cbc7ca/src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius/sylius_order.yml#L10

If I recall correctly, the main idea was to decouple bundles from ORM by default. I don't know if it is still a case, but I would vote for a little discussion here. /cc @Sylius/core-team

loic425:
> I'm assuming, that you are working with `doctrine/orm` driver, am I correct?
> Can you check if this line works properly:https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ProductBundle/DependencyInjection/SyliusProductExtension.php#L43 ?
> 
> It should load this file: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ProductBundle/Resources/config/services/integrations/doctrine/orm.xml and extend basic configuration.

Yes, this load this file. This is indeed ok with product variant repository but not ok for association type repository:

**sylius.repository.product_variant**
Sylius\Bundle\ProductBundle\Doctrine\ORM\ProductVariantRepository

**sylius.repository.product_association_type**                                                                   Sylius\Bundle\ResourceBundle\Doctrine\ORM\EntityRepository

Normal, cause this second repository is not defined in https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ProductBundle/Resources/config/services/integrations/doctrine/orm.xml

lchrusciel:
At least we found a source of 🐛 Would you like to open a PR with a fix? :D

loic425:
Yes I will open a pull request next week :)

PR #10362, opened 2019-05-07
Update release process with dates for 1.5 - 1.7 releases

pamil:


docs/contributing/organization/release-process.rst
```diff
@@ -53,17 +53,19 @@ Past Releases
 +---------+--------------------+--------------------+--------------------+-----------------------+
 | 1.4     | Feb 4, 2019        | Oct 4, 2019        | Jun 4, 2020        | Fully supported       |
 +---------+--------------------+--------------------+--------------------+-----------------------+
+| 1.5     | May 10, 2019       | Jan 10, 2020       | Sep 10, 2020       | Fully supported       |
```
lchrusciel: Shouldn't we stop support for 1.1 and add security support only for 1.2?

pamil: The development phase of 1.5 was shortened from 4 months to 3 months, so it will happen in June (01.06 - end of maintenance for 1.3, 13.06 - end of life for 1.1). Not only Lannisters always pay their debts.

PR #10365, opened 2019-05-08
[Admin][Promotion] Fix removing taxon used in promotion rule

### LINKED ISSUE #10225: Removing taxon used in promotion breaks the promotion edit page
**Sylius version affected**: 1.2.x

**Description**  
It seems that admin user is able to delete a taxon that is used in the rules on a promotion. Deleting the taxon makes the promotion edit page unusable:

```
Uncaught PHP Exception Symfony\Component\Form\Exception\TransformationFailedException: "Unable to transform value for property path "[taxon]": Object "AppBundle\Entity\Taxon" with identifier "code"="identifier-code-here" does not exist." at /app/vendor/symfony/symfony/src/Symfony/Component/Form/Form.php line 1048
```

Probably best that the taxon cannot be used, or the promotion stays accessible. Anyhow the admin needs to know this is happening, as it might lead to unexpected results.

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10225
| License         | MIT

Zrzut ekranu 2019-05-10 o 12 50 29

### DIFF
--- a/src/Sylius/Behat/Context/Ui/Admin/ManagingPromotionsContext.php
+++ b/src/Sylius/Behat/Context/Ui/Admin/ManagingPromotionsContext.php
@@ -213,10 +213,13 @@ final class ManagingPromotionsContext implements Context
     }
 
     /**
-     * @When /^I add the "([^"]+)" action configured with a percentage value of (?:|-)([^"]+)% for ("[^"]+" channel)$/
+     * @When /^I add the "([^"]+)" action configured with a percentage value of (?:|-)([^"]+)% for ("[^"]+") channel$/
      */
-    public function iAddTheActionConfiguredWithAPercentageValueForChannel($actionType, $percentage = null, $channelName)
-    {
+    public function iAddTheActionConfiguredWithAPercentageValueForChannel(
+        string $actionType,
+        string $percentage = null,
+        string $channelName
+    ): void {
         $this->createPage->addAction($actionType);
         $this->createPage->fillActionOptionForChannel($channelName, 'Percentage', $percentage);
     }
@@ -408,7 +411,6 @@ final class ManagingPromotionsContext implements Context
     /**
      * @Given I want to modify a :promotion promotion
      * @Given /^I want to modify (this promotion)$/
-     * @Then I should be able to modify a :promotion promotion
      */
     public function iWantToModifyAPromotion(PromotionInterface $promotion)
     {
@@ -618,11 +620,21 @@ final class ManagingPromotionsContext implements Context
     }
 
     /**
+     * @Then I should be able to modify a :promotion promotion
+     */
+    public function iShouldBeAbleToModifyAPromotion(PromotionInterface $promotion): void
+    {
+        $this->iWantToModifyAPromotion($promotion);
+        $this->updatePage->saveChanges();
+    }
+
+    /**
      * @Then the :promotion promotion should have :ruleName rule configured
      */
     public function thePromotionShouldHaveRuleConfigured(PromotionInterface $promotion, string $ruleName): void
     {
         $this->iWantToModifyAPromotion($promotion);
+        $this->updatePage->saveChanges();
 
         Assert::true($this->updatePage->hasRule($ruleName));
     }
@@ -636,7 +648,6 @@ final class ManagingPromotionsContext implements Context
 
         Assert::false($this->updatePage->hasAnyRule());
     }
-
 
     /**
      * @param string $element
--- a/src/Sylius/Behat/Context/Ui/Admin/ManagingPromotionsContext.php
+++ b/src/Sylius/Behat/Context/Ui/Admin/ManagingPromotionsContext.php
@@ -618,6 +618,27 @@ final class ManagingPromotionsContext implements Context
     }
 
     /**
+     * @Then the :promotion promotion should have :ruleName rule configured
+     */
+    public function thePromotionShouldHaveRuleConfigured(PromotionInterface $promotion, string $ruleName): void
+    {
+        $this->iWantToModifyAPromotion($promotion);
+
+        Assert::true($this->updatePage->hasRule($ruleName));
+    }
+
+    /**
+     * @Then the :promotion promotion should not have any rule configured
+     */
+    public function thePromotionShouldNotHaveAnyRuleConfigured(PromotionInterface $promotion): void
+    {
+        $this->iWantToModifyAPromotion($promotion);
+
+        Assert::false($this->updatePage->hasAnyRule());
+    }
+
+
+    /**
      * @param string $element
      * @param string $expectedMessage
      */
--- a/src/Sylius/Behat/Page/Admin/Promotion/UpdatePage.php
+++ b/src/Sylius/Behat/Page/Admin/Promotion/UpdatePage.php
@@ -92,6 +92,29 @@ class UpdatePage extends BaseUpdatePage implements UpdatePageInterface
             && $this->getElement('ends_at_time')->getValue() === date('H:i', $timestamp);
     }
 
+    public function hasAnyRule(): bool
+    {
+        $items = $this->getElement('rules')->findAll('css', 'div[data-form-collection="item"]');
+
+        return 0 < count($items);
+    }
+
+    public function hasRule(string $name): bool
+    {
+        $items = $this->getElement('rules')->findAll('css', 'div[data-form-collection="item"]');
+
+        foreach ($items as $item) {
+            $selectedOption = $item->find('css', 'option[selected="selected"]');
+
+            /** @var NodeElement $selectedOption */
+            if ($selectedOption->getText() === $name) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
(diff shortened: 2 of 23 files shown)


src/Sylius/Behat/Context/Ui/Admin/ManagingPromotionsContext.php
```diff
@@ -617,6 +617,27 @@ public function thePromotionsShouldHavePriority(PromotionInterface $promotion, i
...
+     * @Then the :promotion promotion should not have any rule configured
+     */
+    public function thePromotionShouldNotHaveAnyRuleConfigured(PromotionInterface $promotion): void
+    {
+        $this->iWantToModifyAPromotion($promotion);
+
+        Assert::false($this->updatePage->hasAnyRule());
+    }
+
+
```
Tomanhez: Redundant blank line

src/Sylius/Behat/Context/Ui/Admin/ManagingPromotionsContext.php
```diff
@@ -617,6 +620,26 @@ public function thePromotionsShouldHavePriority(PromotionInterface $promotion, i
         Assert::same($this->updatePage->getPriority(), $priority);
     }
 
+    /**
+     * @Then the :promotion promotion should have :ruleName rule configured
+     */
+    public function thePromotionShouldHaveRuleConfigured(PromotionInterface $promotion, string $ruleName): void
+    {
+        $this->iWantToModifyAPromotion($promotion);
```
lchrusciel: Can we also make a `submit` on this page? Just to ensure, that clicking a button will not 💥

src/Sylius/Bundle/CoreBundle/EventListener/TaxonDeletionListener.php
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $promotionRules = $this->promotionRuleRepository->findBy(['type' => HasTaxonRuleChecker::TYPE]);
+
+        /** @var PromotionRuleInterface $promotionRule */
+        foreach ($promotionRules as $promotionRule) {
+            $configuration = $promotionRule->getConfiguration();
+            if (in_array($taxonCode, $configuration['taxons'])) {
+                $taxons = array_filter(
+                    $configuration['taxons'],
+                    function (string $code) use ($taxonCode) {
```
lchrusciel: we could add return typehint here

src/Sylius/Bundle/CoreBundle/EventListener/TaxonDeletionListener.php
```diff
@@ -0,0 +1,91 @@
...
+    /** @var EntityManagerInterface */
+    private $manager;
+
+    public function __construct(RepositoryInterface $promotionRuleRepository, EntityManagerInterface $manager)
+    {
+        $this->promotionRuleRepository = $promotionRuleRepository;
+        $this->manager = $manager;
+    }
+
+    public function removeTaxonFromPromotionRules(GenericEvent $event): void
```
lchrusciel: I have two problems with this implementation. First of all, it will work only for taxons deleted through Resource Controller. Secondly, we are hardcoded to only two existing rules that contain relation to the taxon(or taxons).

I would prefer, to decouple this logic from ResourceController, but I'm not sure if it's a good place to do it. For sure, it is a good place to start a discussion.
According to the second problem, it could be resolved with some new interface and iteration over all promotion rules in our store. Something like: TaxonsAwarePromotionRule or something similar. WDYT?

src/Sylius/Bundle/CoreBundle/EventListener/TaxonDeletionListener.php
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private function resolveHasTaxonRules(string $taxonCode): void
+    {
+        $promotionRules = $this->promotionRuleRepository->findBy(['type' => HasTaxonRuleChecker::TYPE]);
+
+        /** @var PromotionRuleInterface $promotionRule */
+        foreach ($promotionRules as $promotionRule) {
+            $configuration = $promotionRule->getConfiguration();
+            if (in_array($taxonCode, $configuration['taxons'])) {
```
lchrusciel: WDYT about simplifing this `if`?
```php
if (($key = array_search($taxonCode, $configuration['taxons'])) !== false) {
    unset($configuration['taxons'][$key]);
}
```
or
```php
$configuration['taxons'] = array_diff($configuration['taxons'], [$taxonCode])
```

src/Sylius/Behat/Context/Ui/Admin/ManagingPromotionsContext.php
GSadee: Added in other step

src/Sylius/Bundle/CoreBundle/EventListener/TaxonDeletionListener.php
GSadee: We've discussed the first problem, that this behaviour could be associated with the resource event.
About the second one, I've moved the logic of removing taxon from configuration to separate services.

src/Sylius/Behat/Page/Admin/Promotion/UpdatePage.php
```diff
@@ -92,6 +92,29 @@ public function hasEndsAt(\DateTimeInterface $dateTime): bool
             && $this->getElement('ends_at_time')->getValue() === date('H:i', $timestamp);
     }
 
+    public function hasAnyRule(): bool
+    {
+        $items = $this->getElement('rules')->findAll('css', 'div[data-form-collection="item"]');
+
+        return 0 > count($items);
```
lchrusciel: Shouldn't it be like this:
```suggestion
        return 0 < count($items);
```
?

src/Sylius/Bundle/CoreBundle/EventListener/TaxonDeletionListener.php
```diff
@@ -0,0 +1,62 @@
...
+    private $hasTaxonRuleUpdater;
+
+    /** @var TaxonAwareRuleUpdaterInterface */
+    private $totalOfItemsFromTaxonRuleUpdater;
+
+    /** @var SessionInterface */
+    private $session;
+
+    public function __construct(
+        TaxonAwareRuleUpdaterInterface $hasTaxonRuleUpdater,
```
lchrusciel: Would it work with variadic in the constructor?

```php
    public function __construct(
		SessionInterface $session,
        TaxonAwareRuleUpdaterInterface ....$ruleUpdaters
        
    ) {
        $this->session = $session;
        $this->ruleUpdaters = $ruleUpdaters;
    }
```

and iterate over them?

src/Sylius/Behat/Service/NotificationChecker.php
```diff
@@ -35,12 +35,21 @@ public function checkNotification(string $message, NotificationType $type): void
...
         throw new NotificationExpectationMismatchException($type, $message);
     }
+
+    private function resolveClass(NotificationType $type): string
+    {
+        if ($type->__toString() === 'info') {
+            return 'info';
+        }
+
+        return $type->__toString() === 'success' ? 'positive' : 'negative';
```
Zales0123: What about having some type-class-map? Sth like:

```php
private function resolveClass(NotificationType $type): string
{
    $typeClassMap = [
        'info' => 'info',
        'failure' => 'negative',
        'success' => 'positive',
    ];

    Assert::keyExist($typeClassMap, $type->__toString());

    return $typeClassMap($type->__toString());
}
```

? I know it's a little bit longer, but much cleaner IMO and more extendable. But also it can be done as an improvement in a separate PR, so no pressure 🖖

src/Sylius/Bundle/CoreBundle/EventListener/TaxonDeletionListener.php
```diff
@@ -0,0 +1,59 @@
...
+    /** @var SessionInterface */
+    private $session;
+
+    /** @var TaxonAwareRuleUpdaterInterface[] */
+    private $ruleUpdaters;
+
+    public function __construct(
+        SessionInterface $session,
+        TaxonAwareRuleUpdaterInterface ...$ruleUpdaters
+
```
Zales0123: 😱

src/Sylius/Behat/Service/NotificationChecker.php
```diff
@@ -35,12 +36,25 @@ public function checkNotification(string $message, NotificationType $type): void
...
+
+    private function resolveClass(NotificationType $type): string
+    {
+        $typeClassMap = [
+            'failure' => 'negative',
+            'info' => 'info',
+            'success' => 'positive',
+        ];
+
+        Assert::keyExists($typeClassMap, $type->__toString());
```
lchrusciel: I'm just wondering, what is better
```suggestion
        Assert::keyExists($typeClassMap, (string) $type);
```

lchrusciel:

PR #10371, opened 2019-05-09
[Docs] How to find out the resource config required when customizing models

4c0n:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Added information/instructions on how to find out what the correct resource configuration should be in `_sylius.yaml`.


docs/customization/model.rst
```diff
@@ -96,6 +96,139 @@ That in Sylius-Standard configuration is overriden already.
...
+.. code-block:: bash
+
+    ---------------------------- -------------------------------------------
+     Parameter                    Value
+    ---------------------------- -------------------------------------------
+     sylius.model.country.class   App\Entity\Country
+    ---------------------------- -------------------------------------------
+
+
+In some cases however, you will see an error:
```
lchrusciel: I would vote to add something like this to separate header.

4c0n: I was absolutely thinking the same thing, but what to call this?
It seems very related to "extended" models (https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/inheritance-mapping.html#class-table-inheritance ?), but how can we explain or easily find out if that is the case for the model that the developer wants to customize?

Another option could be to add a "Troubleshooting" section for example, would you be okay with that?

lchrusciel: `Troubleshooting` seems reasonable for me, but this is a @CoderMaggie call :)

docs/customization/model.rst
```diff
@@ -96,6 +96,139 @@ That in Sylius-Standard configuration is overriden already.
                 classes:
                     model: App\Entity\Addressing\Country
 
+You can check if the configuration in ``config/packages/_sylius.yaml`` is correct by running:
+
+.. code-block:: bash
+
+    $ php bin/console debug:container --parameter=sylius.model.country.class
+
+
```
GSadee: Unnecessary blank line

docs/customization/model.rst
```diff
@@ -96,6 +96,139 @@ That in Sylius-Standard configuration is overriden already.
...
+.. code-block:: bash
+
+    ---------------------------- -------------------------------------------
+     Parameter                    Value
+    ---------------------------- -------------------------------------------
+     sylius.model.country.class   App\Entity\Country
+    ---------------------------- -------------------------------------------
+
+
```
GSadee: Unnecessary blank line

docs/customization/model.rst
4c0n: @GSadee I removed it.

docs/customization/model.rst
4c0n: @GSadee I removed it.

4c0n:
> Hello Youri 
> 
> This tip is very valuable, although it's description is lengthy, don't you think? 
> 
> I suggest you to refactor this and:
> 
> 1. add a new doc (`Tips & Tricks`) to the Customization Guide section `docs/customization/tips_and_tricks.rst`;
> 2. In this doc add a section `How to get Sylius Resource configuration from the container?`
> 3. And there give the tips you are giving here
> 4. Here in the models doc you can then refer to that part from a `.. tip::` :)
> 
> How do you feel about it?

Hi Magdalena,

Thank you very much for your feedback. I really appreciate it. :D
It is quite lengthy indeed, I totally agree with your suggestions and will process them.

I feel at some point in time we should somehow edit the original model customization guide and introduce a step 1 that allows the reader to determine what kind of model they are trying to customize and immediately point them to the right section (so having separate instructions for each case, like we have a section for regular models and a section for translated models now).
However we currently don't have a good way to do that, so we can always revisit it when we have that figured out.

4c0n:
@CoderMaggie Can you review this again please? Thanks in advance!

pamil:
@4c0n thanks for your patience! Merging this PR as soon as the build passes 🎉

pamil:
Thanks, Youri! :tada:

4c0n:
At your service Kamil :+1:

PR #10389, opened 2019-05-15
[UI] Hide filters by default on index pages

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | 

On some pages, especially on the orders index page, filters block is enormous and take almost half of the screen:

Zrzut ekranu 2019-05-15 o 10 47 49

It seems a reasonable change to hide filters by default, as they're just an addition to the index page functionality, not the core of it:

Zrzut ekranu 2019-05-15 o 10 47 58

PR #10228, opened 2019-03-14
Improve taxon UI

kulczy:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9347
| License         | MIT

mov

Zrzut ekranu 2019-03-14 o 10 14 32
Zrzut ekranu 2019-03-14 o 10 14 52

PR #10396, opened 2019-05-20
[Admin] Avoid javascript in saving positions

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

These scenarios were firstly problematic during https://github.com/Sylius/Sylius/pull/10228, so I've fixed it (apparently nope) by adding a [focus() method](https://github.com/Sylius/Sylius/pull/10228/commits/2cec0b211e4a06e8a8144ac3d276cee3025c418b#diff-f7ba951e53d61c2c9086e0a60c8cf788R49) on the element. However, it's still a problem in https://github.com/Sylius/Sylius/pull/10389. The thing is, in the javascript scenarios some other element is covering the "Save position" button if the screen size is too small 🚀 💃 

I've realized that we can avoid using an ajax call for saving positions, by using a `form` attribute on inputs. If this change will be welcomed, we can do the same for the variants' positions.


src/Sylius/Bundle/AdminBundle/Resources/views/Product/Grid/Action/updatePositions.html.twig
```diff
@@ -1,4 +1,6 @@
-<form class="sylius-update-product-taxons" action="{{ path('sylius_admin_ajax_product_taxons_update_position') }}" data-csrf-token="{{ csrf_token('update-product-taxon-position') }}">
+<form id="sylius-update-product-taxons" class="sylius-update-product-taxons" action="{{ path('sylius_admin_ajax_product_taxons_update_position') }}" method="post">
+    <input type="hidden" name="_method" value="PUT"/>
```
Tomanhez:
```suggestion
    <input type="hidden" name="_method" value="PUT" />
```
Missing space

src/Sylius/Bundle/AdminBundle/Resources/views/Product/Grid/Action/updatePositions.html.twig
```diff
@@ -1,4 +1,6 @@
-<form class="sylius-update-product-taxons" action="{{ path('sylius_admin_ajax_product_taxons_update_position') }}" data-csrf-token="{{ csrf_token('update-product-taxon-position') }}">
+<form id="sylius-update-product-taxons" class="sylius-update-product-taxons" action="{{ path('sylius_admin_ajax_product_taxons_update_position') }}" method="post">
```
lchrusciel: Is it still an ajax call? We cannot remove this route, but maybe depreciation and adding a new one would be a good option?

src/Sylius/Bundle/CoreBundle/Controller/ProductTaxonController.php
```diff
@@ -37,23 +37,25 @@ public function updatePositionsAction(Request $request): Response
...
-
-                $this->manager->flush();
+                $productTaxonFromBase = $this->repository->findOneBy(['id' => $id]);
+                $productTaxonFromBase->setPosition((int) $position);
             }
+
+            $this->manager->flush();
         }
 
-        return new JsonResponse();
```
lchrusciel: On the one hand, we kept the contract, but if someone was using this action it may stop working as expected. Same as with routes, we should probably deprecate this class and create a new implementation

features/product/managing_products/sorting_products_within_a_taxon_by_position.feature
```diff
@@ -47,3 +47,10 @@ Feature: Sorting listed products from a taxon by position
         And I start sorting products by position
         Then I should see 3 products in the list
         And they should have order like "Small pug", "Young pug" and "Old pug"
+
+    @ui
+    Scenario: Being unable to use a non-numeric string as a product position
+        When I am browsing products from "Soft Toys" taxon
```
GSadee: `Given`

src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
```diff
@@ -828,9 +828,9 @@ public function iSaveMyNewConfiguration()
     /**
      * @When I set the position of :productName to :position
      */
-    public function iSetThePositionOfTo($productName, $position)
+    public function iSetThePositionOfTo(string $productName, string $position): void
```
GSadee: `int $position`?

Zales0123: The idea is to allow setting a `string` as it's possible from the UI (you can type any string in the position field) and therefore we can check if validation is working

pamil:
Thank you, Mateusz! :tada:

PR #10404, opened 2019-05-23
Fix huge autocomplete queries issue

bitbager:
In case you have a large amount of data (i.e. 100k products), the autocomplete query fetches all matching results which brokes down the whole web server. This simple fix limits the results to 25 as in most cases the end user does not need more than this amount of suggested hits. 
Could be improved by adding the limit parameter to the method signature.

| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductRepository.php
+++ b/src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductRepository.php
@@ -43,6 +43,7 @@ class ProductRepository extends EntityRepository implements ProductRepositoryInt
             ->andWhere('translation.name LIKE :name')
             ->setParameter('name', '%' . $phrase . '%')
             ->setParameter('locale', $locale)
+            ->setMaxResults(25)
             ->getQuery()
             ->getResult()
         ;
--- a/src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductRepository.php
+++ b/src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductRepository.php
@@ -36,14 +36,14 @@ class ProductRepository extends EntityRepository implements ProductRepositoryInt
     /**
      * {@inheritdoc}
      */
-    public function findByNamePart(string $phrase, string $locale): array
+    public function findByNamePart(string $phrase, string $locale, ?int $limit = null): array
     {
         return $this->createQueryBuilder('o')
             ->innerJoin('o.translations', 'translation', 'WITH', 'translation.locale = :locale')
             ->andWhere('translation.name LIKE :name')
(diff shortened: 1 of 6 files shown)


src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductRepository.php
```diff
@@ -43,6 +43,7 @@ public function findByNamePart(string $phrase, string $locale): array
             ->andWhere('translation.name LIKE :name')
             ->setParameter('name', '%' . $phrase . '%')
             ->setParameter('locale', $locale)
+            ->setMaxResults(25)
```
vvasiloi: Maybe make it an optional argument with 25 as default value?

vvasiloi: Nevermind, this is an interface method.

pamil: My idea would be to add an optional parameter to the method definition:

```
public function findByNamePart(string $phrase, string $locale, ?int $limit = null): array
```

This way, the behaviour wouldn't change, but we'd allow for specifying the max results amount.

The second step to actually make use of it would be to modify the routing of the action which provides autocompletion (`sylius_admin_ajax_product_by_name_phrase`) and limit the query to 25 there.

It would be best if we could change it for both `ProductRepositoryInterface` and `TaxonRepositoryInterface`, they both have the same method with the same usage.

What do you think?

vvasiloi: @pamil That's what I thought of too, but wouldn't it be a BC break?

bitbager: Yeap, as I said - this could be improved by adding the default limit to the signature. However, this would be a BC if someone customized the AJAX route in his local environment. As this would be a bugfix, I think that it makes more sense to add the limit in the default repository interface like this: 
`public function findByNamePart(string $phrase, string $locale, ?int $limit = 25): array`

pamil: Even if the route got overwritten, it should then default to not including the limit - tested this on Sylius-Standard.

pamil:
may become messed-up.**

execute `git fetch origin && git reset "patch-2"` to update your local branch.
 
pamil:
Thank you, Mikołaj! :1st_place_medal:

PR #10434, opened 2019-06-06
[Shipping] Shipping method rules

### LINKED ISSUE #10302: Shipping method rules
I would love to be able to add a shipping method checker or whatever the name might be just as you can add promotion rules using tags. Could be something like:

```xml
<service id="App\Shipping\Checker\MyChecker">
    <tag name="sylius.shipping_method_eligibility_checker" />
</service>
```

I am just creating this issue so that it isn't forgotten. I will hopefully get around to creating a PR. Would be great with some feedback on naming from the Sylius team.

loevgaard:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | No
| New feature?    | Yes
| BC breaks?      | No
| Deprecations?   | Yes
| Related tickets | Fixes #10302
| License         | MIT

This pull request adds rules to shipping methods. If any rule returns false the shipping method will not be shown. It is very much based on the same logic as the promotion engine.

- [x] Functionality
- [x] Specs
- [x] Behat tests
- [x] Docs

Here is a screenshot of the functionality:

image


loevgaard:
@CoderMaggie, @Zales0123: Is it possible to get a code review before writing the Behat tests. I am still a Behat newb, so I would like to know that the functionality is what you expect and everything else is okay before I dive into that :D

loevgaard:
Ping @lchrusciel, @pamil

loevgaard:
Do not stale

src/Sylius/Component/Core/spec/Resolver/EligibleDefaultShippingMethodResolverSpec.php
```diff
@@ -13,6 +13,7 @@
 
 namespace spec\Sylius\Component\Core\Resolver;
 
+use InvalidArgumentException;
```
Zales0123: Not neede, I suppose

src/Sylius/Component/Shipping/Checker/ShippingMethodEligibilityChecker.php
```diff
@@ -13,9 +13,16 @@
 
 namespace Sylius\Component\Shipping\Checker;
 
+use const E_USER_DEPRECATED;
+use Sylius\Component\Shipping\Checker\Eligibility\CompositeShippingMethodEligibilityChecker;
 use Sylius\Component\Shipping\Model\ShippingMethodInterface;
 use Sylius\Component\Shipping\Model\ShippingSubjectInterface;
 
+@trigger_error(sprintf('The "%s" class is deprecated since Sylius 1.7, use "%s" instead.', ShippingMethodEligibilityChecker::class, CompositeShippingMethodEligibilityChecker::class), E_USER_DEPRECATED);
```
Zales0123: `\E_USER_DEPRECATED` and we don't need a use case

src/Sylius/Component/Shipping/spec/Checker/Eligibility/CompositeShippingMethodEligibilityCheckerSpec.php
```diff
@@ -0,0 +1,99 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace spec\Sylius\Component\Shipping\Checker\Eligibility;
+
+use InvalidArgumentException;
+use PhpSpec\ObjectBehavior;
+use stdClass;
```
Zales0123: 💃

src/Sylius/Component/Shipping/Checker/ShippingMethodEligibilityChecker.php
loevgaard: What do you mean by `we don't need a use case`?

src/Sylius/Component/Core/spec/Resolver/EligibleDefaultShippingMethodResolverSpec.php
loevgaard: Usually I add use statements for all imported classes. If this line is removed I would use `\ InvalidArgumentException` in the code, which the PhpStorm inspections complain about just as it would with something like `\Sylius\Component\Order\Model\Order` directly in the code instead of imported as `use Sylius\Component\Order\Model\Order`

src/Sylius/Bundle/CoreBundle/Validator/Constraints/OrderShippingMethodEligibilityValidator.php
```diff
@@ -14,7 +14,7 @@
 namespace Sylius\Bundle\CoreBundle\Validator\Constraints;
 
 use Sylius\Component\Core\Model\OrderInterface;
-use Sylius\Component\Shipping\Checker\ShippingMethodEligibilityCheckerInterface;
+use Sylius\Component\Shipping\Checker\Eligibility\ShippingMethodEligibilityCheckerInterface;
```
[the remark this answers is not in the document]
loevgaard: `\Sylius\Component\Shipping\Checker\ShippingMethodEligibilityCheckerInterface` (the old interface)

extends

`\Sylius\Component\Shipping\Checker\Eligibility\ShippingMethodEligibilityCheckerInterface` (the new interface)

Isn't that correct? It means if you inject something that implements the old interface that would still be possible.

src/Sylius/Bundle/CoreBundle/Resources/config/services/shipping.xml
```diff
@@ -11,20 +11,37 @@
...
-<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+<container xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://symfony.com/schema/dic/services"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <services>
-        <defaults public="true" />
+        <defaults public="true"/>
 
-        <service id="sylius.shipping_calculator.flat_rate" class="Sylius\Component\Core\Shipping\Calculator\FlatRateCalculator">
+        <service id="sylius.shipping_calculator.flat_rate"
+                 class="Sylius\Component\Core\Shipping\Calculator\FlatRateCalculator">
```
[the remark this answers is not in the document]
loevgaard: I just press `cmd + option + L` and that's what it gives me. I am not much into doing formatting manually. I think the XML formatter just makes sure that no lines extend past the 'vertical line indicator' (unless the line can't be broken)

image

src/Sylius/Bundle/ShippingBundle/Form/EventSubscriber/AddTotalFormSubscriber.php
```diff
@@ -0,0 +1,44 @@
...
+namespace Sylius\Bundle\ShippingBundle\Form\EventSubscriber;
+
+use Sylius\Bundle\CoreBundle\Form\Type\ChannelCollectionType;
+use Sylius\Bundle\ShippingBundle\Form\Type\AmountType;
+use Sylius\Component\Core\Model\ChannelInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Form\FormEvent;
+use Symfony\Component\Form\FormEvents;
+
+final class AddTotalFormSubscriber implements EventSubscriberInterface
```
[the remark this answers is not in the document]
loevgaard: I guess my thought were more if other developers would like to use it for their future rules. Maybe even a more generic `AddAmountFormSubscriber` would be better. One where the constructor took two arguments like so:

`public function __construct(string $name = 'amount', string $label = 'sylius.form.amount')`.

wdyt?

src/Sylius/Bundle/ShippingBundle/Form/Type/Core/AbstractConfigurationCollectionType.php
```diff
@@ -0,0 +1,89 @@
...
+
+use Sylius\Component\Registry\ServiceRegistryInterface;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\CollectionType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\Form\FormView;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+abstract class AbstractConfigurationCollectionType extends AbstractType
```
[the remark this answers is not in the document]
loevgaard: Imo it would be great to have some abstract class that developers could use to make rule engines. Me and Igor have copied this logic to other plugins also because it wasn't made in a generic way

src/Sylius/Bundle/ShippingBundle/Form/Type/AbstractConfigurableShippingMethodElementType.php
```diff
@@ -0,0 +1,108 @@
...
+use Sylius\Bundle\ResourceBundle\Form\Registry\FormTypeRegistryInterface;
+use Sylius\Bundle\ResourceBundle\Form\Type\AbstractResourceType;
+use Sylius\Component\Shipping\Model\ConfigurableShippingMethodElementInterface;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormEvent;
+use Symfony\Component\Form\FormEvents;
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+abstract class AbstractConfigurableShippingMethodElementType extends AbstractResourceType
```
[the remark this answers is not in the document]
loevgaard: Yes. You are correct that I just wanted to be consistent. I am not sure these corrections should go into this PR, though?

src/Sylius/Bundle/ShippingBundle/Form/Type/AmountType.php
```diff
@@ -0,0 +1,67 @@
...
+
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver
+            ->setRequired('channel')
+            ->setAllowedTypes('channel', [ChannelInterface::class])
+
+            ->setDefaults([
+                'label' => static function (Options $options): string {
+                    return $options['channel']->getName();
```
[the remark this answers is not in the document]
loevgaard: But you need the channel name to distinguish the amounts for different channels

loevgaard:
@Zales0123: I have answered/fixed your comments :)

src/Sylius/Bundle/ShippingBundle/Form/EventSubscriber/AddTotalFormSubscriber.php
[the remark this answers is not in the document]
Zales0123: Hmmm... still making a subscriber just for one field is a little exaggeration for me 😄 but I would like to see others opinions, cc @pamil @lchrusciel @GSadee

src/Sylius/Bundle/ShippingBundle/Form/Type/AbstractConfigurableShippingMethodElementType.php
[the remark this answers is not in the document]
Zales0123: Yeah, maybe too much at once 🖖 The only real difference between these files is `getRegistryIdentifier` method... but when should we put such a base abstract class that these two classes would extend? In the resource bundle? But it involves the second repo... I'm 👌 with this implementation and later refactorization, but, again, would like to see if @pamil @lchrusciel @GSadee have some better idea 🖖

src/Sylius/Bundle/ShippingBundle/Form/Type/AmountType.php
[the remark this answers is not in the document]
Zales0123: Yes, so we could inject 2 variables: `label` (would pass the channel's name) and `currency` (would pass the channel's base currency code). We even **cannot** use `ChannelInterfae` there, as it couples `ShippingBundle` with `Core` component 🏎️

src/Sylius/Component/Core/Shipping/Checker/Rule/OrderTotalRuleChecker.php
```diff
@@ -0,0 +1,47 @@
...
+
+        $amount = $configuration['total'][$channel->getCode()]['amount'] ?? null;
+        if (null === $amount) {
+            return false;
+        }
+
+        return $this->compare($order->getTotal(), $amount);
+    }
+
+    abstract protected function compare(int $total, $threshold): bool;
```
Zales0123, line 46: Missing type for `$threshold`

src/Sylius/Component/Core/spec/Shipping/Checker/Rule/OrderTotalGreaterThanOrEqualRuleCheckerSpec.php
```diff
@@ -0,0 +1,97 @@
...
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Model\ShipmentInterface;
+use Sylius\Component\Core\Shipping\Checker\Rule\OrderTotalGreaterThanOrEqualRuleChecker;
+use Sylius\Component\Shipping\Checker\Rule\RuleCheckerInterface;
+use Sylius\Component\Shipping\Model\ShipmentInterface as BaseShipmentInterface;
+
+final class OrderTotalGreaterThanOrEqualRuleCheckerSpec extends ObjectBehavior
+{
+    public function it_implements_rule_checker_interface(): void
```
Zales0123, line 26: `public` are not needed in specs 🖖

src/Sylius/Component/Shipping/Checker/Eligibility/CompositeShippingMethodEligibilityChecker.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    /** @var ShippingMethodEligibilityCheckerInterface[] */
+    private $eligibilityCheckers;
+
+    /**
+     * @param ShippingMethodEligibilityCheckerInterface[] $eligibilityCheckers
+     */
+    public function __construct(array $eligibilityCheckers)
+    {
+        Assert::notEmpty($eligibilityCheckers);
```
Zales0123, line 30: I would skip this assertion. If one does not register any checkers, this service would just not check anything 🚀

src/Sylius/Component/Core/Shipping/Checker/Rule/OrderTotalRuleChecker.php
```diff
@@ -0,0 +1,47 @@
...
+        if (null === $channel) {
+            return false;
+        }
+
+        $amount = $configuration['total'][$channel->getCode()]['amount'] ?? null;
+        if (null === $amount) {
+            return false;
+        }
+
+        return $this->compare($order->getTotal(), $amount);
```
Zales0123, line 43: 🚨 

As far as I've investigated, this line is braking the scenario. The reason is when the cart is on the shipping selection step, it already has a default shipping method included which fee is a part of the order's total 🤯 So I suppose we should use `getItemsTotal()` here

Zales0123:
@loevgaard some more comment, rebase is also needed, but generally, it's close to be merged, in my opinion 🖖

src/Sylius/Component/Core/spec/Shipping/Checker/Rule/OrderTotalGreaterThanOrEqualRuleCheckerSpec.php
loevgaard: Always wondered why you didn't do that.

src/Sylius/Bundle/ShippingBundle/Form/Type/AmountType.php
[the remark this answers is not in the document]
loevgaard: I moved this type and associated classes to the core bundle since we need the channel in the form type:

image

src/Sylius/Component/Shipping/Checker/ShippingMethodEligibilityCheckerInterface.php
```diff
@@ -13,10 +13,19 @@
...
+use const E_USER_DEPRECATED;
+use Sylius\Component\Shipping\Checker\Eligibility\ShippingMethodEligibilityCheckerInterface as NewShippingMethodEligibilityCheckerInterface;
 use Sylius\Component\Shipping\Model\ShippingMethodInterface;
 use Sylius\Component\Shipping\Model\ShippingSubjectInterface;
 
-interface ShippingMethodEligibilityCheckerInterface
+@trigger_error(sprintf('The "%s" interface is deprecated since Sylius 1.7, use "%s" instead.', ShippingMethodEligibilityCheckerInterface::class, NewShippingMethodEligibilityCheckerInterface::class), E_USER_DEPRECATED);
+
+/**
+ * @deprecated since Sylius 1.7. Use Sylius\Component\Shipping\Checker\Eligibility\ShippingMethodEligibilityCheckerInterface instead
```
lchrusciel, line 24:
```suggestion
 * @deprecated since Sylius 1.8. Use Sylius\Component\Shipping\Checker\Eligibility\ShippingMethodEligibilityCheckerInterface instead
```

src/Sylius/Component/Shipping/spec/Checker/Eligibility/CategoryRequirementEligibilityCheckerSpec.php
```diff
@@ -0,0 +1,70 @@
...
+        ShippingCategoryInterface $shippingCategory,
+        ShippableInterface $shippable
+    ): void {
+        $shippingMethod->getCategory()->willReturn($shippingCategory);
+        $shippingMethod->getCategoryRequirement()->willReturn(ShippingMethodInterface::CATEGORY_REQUIREMENT_MATCH_ANY);
+
+        $shippable->getShippingCategory()->willReturn($shippingCategory);
+        $subject->getShippables()->willReturn(new ArrayCollection([$shippable->getWrappedObject()]));
+
+        $this->isEligible($subject, $shippingMethod)->shouldReturn(true);
```
lchrusciel, line 43: I think we should cover all possible approval cases. We can either write a specification for each of possible result or pack them into approde/decline groups or per type (any, none, all). If you would like to group them, here is a useful sample of repeated usage of the same config in phpspec:
```suggestion
        $shippingMethod->getCategory()->willReturn($shippingCategory);
        $shippingMethod->getCategoryRequirement()->willReturn(
            ShippingMethodInterface::CATEGORY_REQUIREMENT_MATCH_ANY,
            ShippingMethodInterface::CATEGORY_REQUIREMENT_MATCH_ALL,
            ShippingMethodInterface::CATEGORY_REQUIREMENT_MATCH_NONE
        ); # every time we will execute this test different requirement will be returned

        $shippable->getShippingCategory()->willReturn(
            $shippingCategory,
            $shippingCategory,
            $shippingCategory2
        ); # first two times we will return the same shippingCategory. But we need different for the third test to pass
        
        $subject->getShippables()->willReturn(new ArrayCollection([$shippable->getWrappedObject()]));

        $this->isEligible($subject, $shippingMethod)->shouldReturn(true);
        $this->isEligible($subject, $shippingMethod)->shouldReturn(false);
        $this->isEligible($subject, $shippingMethod)->shouldReturn(true);
```

src/Sylius/Component/Shipping/spec/Checker/Eligibility/CategoryRequirementEligibilityCheckerSpec.php
```diff
@@ -0,0 +1,70 @@
...
+
+    function it_denies_category_requirement_if_categories_do_not_match(
+        ShippingSubjectInterface $subject,
+        ShippingMethodInterface $shippingMethod,
+        ShippingCategoryInterface $shippingCategory,
+        ShippingCategoryInterface $shippingCategory2,
+        ShippableInterface $shippable
+    ): void {
+        $shippingMethod->getCategory()->willReturn($shippingCategory);
+        $shippingMethod->getCategoryRequirement()->willReturn(ShippingMethodInterface::CATEGORY_REQUIREMENT_MATCH_ANY);
```
lchrusciel, line 63: IMHO we should cover more cases

lchrusciel:
I've added a few mostly minor comments (with update of 1.7 to 1.8 version in deprecation). We are getting close to merging it ;)

src/Sylius/Bundle/CoreBundle/Form/EventSubscriber/AddTotalFormSubscriber.php
```diff
@@ -0,0 +1,44 @@
...
+namespace Sylius\Bundle\CoreBundle\Form\EventSubscriber;
+
+use Sylius\Bundle\CoreBundle\Form\Type\ChannelCollectionType;
+use Sylius\Bundle\CoreBundle\Form\Type\AmountType;
+use Sylius\Component\Core\Model\ChannelInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\Form\FormEvent;
+use Symfony\Component\Form\FormEvents;
+
+final class AddTotalFormSubscriber implements EventSubscriberInterface
```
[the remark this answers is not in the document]
loevgaard: Removed :)

src/Sylius/Bundle/ShippingBundle/Form/Type/ShippingMethodType.php
```diff
@@ -137,6 +145,13 @@ public function buildView(FormView $view, FormInterface $form, array $options):
         }
     }
 
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        parent::configureOptions($resolver);
+
+        $resolver->setDefault('rules_help', 'sylius.form.shipping_method.rules_help');
```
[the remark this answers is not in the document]
loevgaard: I have no idea. I have added it directly to the template now. The intention is the same as the form option `help`, yes. But I don't think Sylius have implemented this yet? That's why I put it here, because I like having help/documentation directly where you need it.

src/Sylius/Component/Shipping/Checker/ShippingMethodEligibilityChecker.php
```diff
@@ -13,9 +13,17 @@
...
 use Sylius\Component\Shipping\Model\ShippingSubjectInterface;
 
+@trigger_error(sprintf('The "%s" class is deprecated since Sylius 1.7, use "%s" instead.', 'Sylius\Component\Shipping\Checker\ShippingMethodEligibilityChecker', 'Sylius\Component\Shipping\Checker\Eligibility\CompositeShippingMethodEligibilityChecker'), E_USER_DEPRECATED);
+
+/**
+ * @deprecated since Sylius 1.7. Use Sylius\Component\Shipping\Checker\Eligibility\CompositeShippingMethodEligibilityChecker instead
+ *
+ * @psalm-suppress DeprecatedInterface
+ */
 final class ShippingMethodEligibilityChecker implements ShippingMethodEligibilityCheckerInterface
```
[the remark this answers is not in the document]
loevgaard: This class should be dead, yes. Not sure what your thoughts are here?

src/Sylius/Component/Shipping/spec/Checker/Eligibility/CategoryRequirementEligibilityCheckerSpec.php
loevgaard: Added :)

src/Sylius/Component/Shipping/spec/Checker/Eligibility/CategoryRequirementEligibilityCheckerSpec.php
loevgaard: Done :)

Zales0123:
Thank you, Joachim! :1st_place_medal:

src/Sylius/Component/Core/Shipping/Checker/Rule/OrderTotalRuleChecker.php
MyUtd4Life: @Zales0123  / @loevgaard Can one of you please tell me how this issue is resolved? I think it can still be a problem. 
For example:
Set two shipping rules:
Method A. Order total less than or equal to 99.99 with the shipping costs of 5.00
Method B. Order total greater than or equal to 100 with the shipping costs of 0.00

Create a cart with an items total of 97.50.
Method A can't be used because the order total is now 102.50 (97.50 items total + 5.00 shipping). 
Method B can't be used either because, with the included shipping costs of this method, the order total of 100 (97.50 items total + 0 shipping) is not reached.

loevgaard: I agree that it would probably make more sense that the shipping costs are not included in what we consider the order total for these rules.

Could you make an issue instead @MyUtd4Life?

PR #10440, opened 2019-06-10
Fix removing taxons with numeric codes from products

### LINKED ISSUE #10439: [Admin] Removing products from taxons
**Sylius version affected**: 1.4, 1.5

**Description**  
If the taxon code in integer (technically it's string of course but contains integer value e.g. "6"), The products cannot be unassigned from such taxon. It can be however added to it.

**Steps to reproduce**  

- Create taxon with code "6"
- Edit any product and assign it to that taxon
- Edit product again and try to unassign it from that taxon

**Workaround**  
Change the taxon's code so it does not contain digits only.

vvasiloi:
Use native `.dataset` instead of jQuery `.data()` to avoid type conversion which make it impossible to remove taxons with numeric codes from products

| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10439
| License         | MIT


pamil:
Thank you, Victor! :tada:

Issue #10408, opened 2019-05-24
Doctrine class metadata not found for extended interface


teohhanhui:
**Sylius version affected**: 1.x

**Description**  
For any resource, it should be possible to get the Doctrine class metadata using any of the interfaces for the resource defined at the component / core / app level.

**Steps to reproduce**  
1. Create a new project from Sylius-Standard.
2. Try to get Doctrine class metadata using `Sylius\Component\Core\Model\ProductInterface`.

**Possible Solution**  
I think there is a design problem.

If you look at https://github.com/Sylius/SyliusResourceBundle/blob/v1.5.0/src/Bundle/DependencyInjection/Compiler/DoctrineTargetEntitiesResolverPass.php

it does not go up the inheritance chain - and it's not safe to do so, as we can't tell which interfaces should be mapped, and which should not. So if we set `interface` to a child interface, things would stop working if you try using the parent interface(s).

What is actually expected is that all of the interfaces should work, e.g.:

- `Sylius\Component\Product\Model\ProductInterface`
- `Sylius\Component\Core\Model\ProductInterface`
- `App\Model\ProductInterface` (for example)

So we need a way to map multiple interfaces.

vvasiloi:
I think going up or down the inheritance chain isn't a good idea.
What if an interface is extended/implemented by multiple interfaces/classes?
Like `Sylius\Component\Core\Model\ProductInterface` is implemented by `App\Entity\Product`
and also extended by `App\Model\FooProductInterface`, which is then implemented by `App\Entity\FooProduct`.
You will end up having either `App\Model\FooProductInterface` mapped to `App\Entity\Product` or `Sylius\Component\Core\Model\ProductInterface` mapped to `App\Entity\FooProduct`, depending on the order they're processed.

Maybe it could be done by having a way to explicitely define multiple/additional interfaces for a class.
WDYT?

teohhanhui:
Yes, that's what I've mentioned. We cannot go up / down the inheritance chain as that's unsafe. So we must provide a way to map multiple interfaces. :smile:

vvasiloi:
Oh, sorry, I read that poorly. 😄 
So, maybe something like that:
```
// Sylius\Bundle\ResourceBundle\DependencyInjection\Configuration::addResourcesSection
->arrayNode('interfaces')
    ->cannotBeEmpty()
    ->scalarPrototype()
    ->beforeNormalization()->castToArray()
 ->end()
```
and 
```
[...]

PR #10450, opened 2019-06-12
Fix interfaces mapping in Doctrine for admin user and shop user

### LINKED ISSUE #9360: Sylius Shop user resource declaration does not have an interface config
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | yes
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.1.5

When I try to configure Doctrine mapping reference by interface for ShopUser, I am getting the class not found error:

```
$ bin/console do:sche:cre

 !                                                                                                                      
 ! [CAUTION] This operation should not be executed in a production environment!                                         
 !                                                                                                                      

 Creating database schema...


In MappingException.php line 96:
                                                                        
  Class 'Sylius\Component\Core\Model\ShopUserInterface' does not exis
```

This might be caused by missing interface config in `src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius/sylius_user.yml` file.

Below is my Doctrine config. I hope this will give you a better understanding of the problem.

```
<?xml version="1.0" encoding="UTF-8"?>

<doctrine-mapping xmlns="http://doctrine-project.org/schemas/orm/doctrine-mapping"
                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                  xsi:schemaLocation="http://doctrine-project.org/schemas/orm/doctrine-mapping

    <mapped-superclass name="BitBag\SyliusWishlistPlugin\Entity\Wishlist" table="bitbag_wishlist">
        <id name="id" column="id" type="integer">
            <generator strategy="AUTO" />
        </id>

        <many-to-many field="products" target-entity="Sylius\Component\Product\Model\ProductInterface">
            <join-table name="bitbag_wishlist_products">
                <join-columns>
                    <join-column name="wishlist_id" referenced-column-name="id" nullable="false" on-delete="CASCADE" />
                </join-columns>
                <inverse-join-columns>
                    <join-column name="product_id" referenced-column-name="id" nullable="false" on-delete="CASCADE" />
                </inverse-join-columns>
            </join-table>
        </many-to-many>

        <one-to-one field="shopUser" target-entity="Sylius\Component\Core\Model\ShopUserInterface">
            <join-column name="shop_user_id" referenced-column-name="id" nullable="true" />
        </one-to-one>

    </mapped-superclass>

</doctrine-mapping>
```

### LINKED ISSUE #10447: Customer mapping is using the wrong interface
**Sylius version affected**: 1.4.5 (possibly earlier versions as well)

**Description**  
The `sylius.shop_user` resource uses the `UserInterface` instead of the new `ShopUserInterface` which should be used to make doctrine mappings more concise.

This issue is related with this: #9360 (and its solution proposal). @bitbager probably discovered this during the development of the WishlistPlugin which also broke because the `ShopUserInterface` exists, but is neither used in the doctrine mapping, nor in the resource configuration.

When trying to fix this by using `resolve_target_entities` for example, one ends up breaking both `shop_user` and `admin_user` because both use the same `UserInterface`.

**Steps to reproduce**  
None, really. :( There is more information in the linked issue above.

**Possible Solution**  
1) Adding `ShopUserInterface` as `interface` to the default resource configuration of `sylius.shop_user`, `sylius.admin_user` and `sylius.oauth_user`
2) Fix the `.../CoreBundle/Resources/config/doctrine/model/Customer.orm.xml` and use the correct interface there as well. (If another entity refers to a `user` with a `UserInterface` one should look at those as well.)

This will make entity extension and overriding much easier and is semantically more precise and its using existing classes designed for that exact purpose.

pamil:
Both the first and the second commit should pass before merging this PR. This assures as the changes as backwards compatible (second change resigns from BC layer).

| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10447, fixes #9360, related to #10408
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/DependencyInjection/Compiler/BackwardsCompatibility/ResolveShopUserTargetEntityPass.php
+++ b/src/Sylius/Bundle/CoreBundle/DependencyInjection/Compiler/BackwardsCompatibility/ResolveShopUserTargetEntityPass.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\DependencyInjection\Compiler\BackwardsCompatibility;
+
+use Sylius\Component\User\Model\UserInterface;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\InvalidArgumentException;
+
+/**
+ * @internal
+ */
+final class ResolveShopUserTargetEntityPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        try {
+            $resolveTargetEntityListener = $container->findDefinition('doctrine.orm.listeners.resolve_target_entity');
+            $shopUserClass = $container->getParameter('sylius.model.shop_user.class');
+        } catch (InvalidArgumentException $exception) {
+            return;
+        }
+
+        $resolveTargetEntityListener->addMethodCall(
+            'addResolveTargetEntity',
+            [UserInterface::class, $shopUserClass, []]
+        );
+    }
+}
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius/sylius_user.yml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius/sylius_user.yml
@@ -7,12 +7,14 @@ sylius_user:
             user:
                 classes:
                     model: Sylius\Component\Core\Model\AdminUser
+                    interface: Sylius\Component\Core\Model\AdminUserInterface
                     repository: Sylius\Bundle\UserBundle\Doctrine\ORM\UserRepository
                     form: Sylius\Bundle\CoreBundle\Form\Type\User\AdminUserType
         shop:
             user:
                 classes:
                     model: Sylius\Component\Core\Model\ShopUser
+                    interface: Sylius\Component\Core\Model\ShopUserInterface
                     repository: Sylius\Bundle\CoreBundle\Doctrine\ORM\UserRepository
                     form: Sylius\Bundle\CoreBundle\Form\Type\User\ShopUserType
         oauth:
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/Customer.orm.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/Customer.orm.xml
@@ -20,7 +20,7 @@
             </cascade>
         </one-to-one>
 
-        <one-to-one field="user" target-entity="Sylius\Component\User\Model\UserInterface" mapped-by="customer">
+        <one-to-one field="user" target-entity="Sylius\Component\Core\Model\ShopUserInterface" mapped-by="customer">
             <cascade>
                 <cascade-all/>
             </cascade>
--- a/src/Sylius/Bundle/CoreBundle/SyliusCoreBundle.php
+++ b/src/Sylius/Bundle/CoreBundle/SyliusCoreBundle.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\CoreBundle;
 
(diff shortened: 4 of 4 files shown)


pamil:
`UserOAuth` mappings to `UserInterface` is another issue that needs to be solved, but rather in a different PR.

kortwotze:
Thank you so much! :rocket:

src/Sylius/Bundle/CoreBundle/DependencyInjection/Compiler/BackwardsCompatibility/ResolveShopUserTargetEntityPass.php
```diff
@@ -0,0 +1,31 @@
...
+    public function process(ContainerBuilder $container): void
+    {
+        try {
+            $resolveTargetEntityListener = $container->findDefinition('doctrine.orm.listeners.resolve_target_entity');
+            $shopUserClass = $container->getParameter('sylius.model.shop_user.class');
+        } catch (InvalidArgumentException $exception) {
+            return;
+        }
+
+        $resolveTargetEntityListener->addMethodCall(
```
kortwotze: Are you sure this does not break the AdminUser entity? As far as I see this will replace its target-entity with a ShopUser as well.

pamil: Before this PR, `Sylius\Component\User\Model\UserInterface` targets `ShopUser` entity (cause it's the last one to define `Sylius\Component\User\Model\UserInterface` as its interface). This will preserve backwards compatibility (could be removed in 2.0).

kortwotze:
Currently at lunch, but I think I remember the AdminUser mapping using the same interface for its user-relation.

src/Sylius/Bundle/CoreBundle/DependencyInjection/Compiler/BackwardsCompatibility/ResolveShopUserTargetEntityPass.php
kortwotze: Currently at lunch, but I think I remember the AdminUser mapping using the same interface for its user-relation.

pamil: That's correct, but shop user comes next and overwrites it :)

kortwotze: Isn't `resolveTargetEntity` replacing all occurrences of `OriginalTargetEntity` with `NewTargetEntity` and thus both `AdminUser` and `ShopUser` use the `NewTargetEntity` (ShopUserInterface) as its target-entity?

pamil: It works like this:

```php
$interfaces = [];
$interfaces[UserInterface::class] = AdminUser::class;
$interfaces[UserInterface::class] = ShopUser::class;

var_dump(str_replace(array_keys($interfaces), array_values($interfaces), UserInterface::class));
// string(8) "ShopUser"
```

There can be only one target entity for a given interface.

kortwotze: Wait - I think thats on me. 
I got ResourceBundle config and mappings mixed up. 
The `AdminUser` and `ShopUser` had the same interface configured for its resource, but the `ShopUser` is the only one being used with `UserInterface` in a relation (Customer in this case).

The `AdminUser::class` has no target-entity for nothing, so my worries fizzle.
The `ShopUser::class` is (only) being used by the `Customer` as an inverse relation with target-entity `UserInterface` (which gets corrected anyways) and thus re-mapping `UserInterface` to `ShopUserInterface` (or the currently configured resource-class) in the whole application is totally fine. 

My bad, thanks for your time though :heart:

PR #10456, opened 2019-06-14
Make image uploader easier to customize

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | 
| License         | MIT

It always annoyed me, that if you want to change anything in the images upload process (for example keep the file name, as it's done in our Behats) you need to override the whole, big service. With this change, a path generation logic has been extracted - I believe it provides much better DX right now (and fulfills SOLID principles as well) :)


src/Sylius/Component/Core/Uploader/ImageUploader.php
```diff
@@ -23,9 +24,21 @@ class ImageUploader implements ImageUploaderInterface
...
+    /** @var ImagePathGeneratorInterface */
+    protected $imagePathGenerator;
+
+    public function __construct(
+        Filesystem $filesystem,
+        ?ImagePathGeneratorInterface $imagePathGenerator = null
+    ) {
         $this->filesystem = $filesystem;
+
+        if ($this->imagePathGenerator === null) {
```
pamil: This will always trigger a deprecation ⚠️

src/Sylius/Component/Core/Uploader/ImageUploader.php
```diff
@@ -71,6 +83,17 @@ public function remove(string $path): bool
...
 
+    private function getImagePath(ImageInterface $image): string
+    {
+        if (null !== $this->imagePathGenerator) {
+            return $this->imagePathGenerator->generate($image);
+        }
+
+        $hash = bin2hex(random_bytes(16));
+
+        return $this->expandPath($hash . '.' . $image->getFile()->guessExtension());
```
pamil: The logic in these methods (`getImagePath` and `expandPath`) looks the same as in `UploadedImagePathGenerator`. This way we can create and assign `UploadedImagePathGenerator` instance in the constructor if no image path generator has been passed and get rid of this code in this class.

src/Sylius/Component/Core/spec/Uploader/ImageUploaderSpec.php
```diff
@@ -16,37 +16,46 @@
...
+        ImagePathGeneratorInterface $imagePathGenerator,
+        ImageInterface $image
+    ): void {
         $filesystem->has(Argument::any())->willReturn(false);
 
         $file = new File(__FILE__);
         $image->getFile()->willReturn($file);
 
-        $this->beConstructedWith($filesystem);
+        $this->beConstructedWith($filesystem, $imagePathGenerator);
```
pamil: Could we add a test for the deprecation (`$this->shouldTrigger(\E_USER_DEPRECATED)->during...()` matcher).

config/packages/test/sylius_uploader.yaml
```diff
@@ -4,8 +4,5 @@ services:
         arguments:
             - "%sylius.uploader.filesystem%"
         factory: ['@knp_gaufrette.filesystem_map', 'get']
-    sylius.image_uploader:
-        class: Sylius\Behat\Service\Uploader\ImageUploader
-        arguments:
-            - "@sylius_test.filesystem"
```
pamil: Are we sure we don't need to use `sylius_test.filesystem` anymore?

Zales0123: Good point 😄

config/packages/test/sylius_uploader.yaml
```diff
@@ -5,7 +5,9 @@ services:
             - "%sylius.uploader.filesystem%"
         factory: ['@knp_gaufrette.filesystem_map', 'get']
     sylius.image_uploader:
-        class: Sylius\Behat\Service\Uploader\ImageUploader
+        class: Sylius\Component\Core\Uploader\ImageUploader
         arguments:
             - "@sylius_test.filesystem"
-        public: true
+            - "@Sylius\\Component\\Core\\Generator\\ImagePathGeneratorInterface"
+    Sylius\Component\Core\Generator\ImagePathGeneratorInterface:
```
pamil: I guess after the recent changes we just need this service to make tests work and the rest of definitions in this file can be removed.

src/Sylius/Component/Core/Generator/ImagePathGeneratorInterface.php
```diff
@@ -0,0 +1,21 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\Generator;
+
+use Sylius\Component\Core\Model\ImageInterface;
+
+interface ImagePathGeneratorInterface
+{
+    public function generate(ImageInterface $image): string;
```
pamil: In `ImageUploader` we use this method in a way that needs to be defined further, possibly as a comment.

This method cannot be idempotent, every call to this method should return a different value. If the same value is returned, then `ImageUploader` might enter an infinite loop.

pamil:
Thank you, Mateusz! :tada:

Zales0123:
And Kamil! 🚀

PR #10461, opened 2019-06-18
Fix product show page elements

kulczy:
- Fix channels display
- Remove options row from simple products
- Add margin to the stock column

Zrzut ekranu 2019-06-18 o 15 10 06
Zrzut ekranu 2019-06-18 o 15 15 18


GSadee:
Thanks, Szymon! :tada:

PR #10489, opened 2019-07-02
Make it possible to have no shipping methods for Order fixtures

TiMESPLiNTER:
Currently it's not possible to have fixture suites with no shipping methods (a.k.a. virtual shipping only). This PR makes it possible.

| Q               | A
| --------------- | -----
| Branch?         | 1.4 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | #10490
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php
@@ -175,6 +175,8 @@ class OrderFixture extends AbstractFixture
     {
         $numberOfItems = random_int(1, 5);
         $products = $this->productRepository->findAll();
+        $shippingMethods = $this->shippingMethodRepository->findAll();
+        $shippingMethodsAvailable = (bool) count($shippingMethods);
 
         for ($i = 0; $i < $numberOfItems; ++$i) {
             /** @var OrderItemInterface $item */
@@ -182,6 +184,7 @@ class OrderFixture extends AbstractFixture
 
             $product = $this->faker->randomElement($products);
             $variant = $this->faker->randomElement($product->getVariants()->toArray());
+            $variant->setShippingRequired($shippingMethodsAvailable);
 
             $item->setVariant($variant);
             $this->orderItemQuantityModifier->modify($item, random_int(1, 5));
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php
```diff
@@ -175,13 +175,16 @@ private function generateItems(OrderInterface $order): void
     {
         $numberOfItems = random_int(1, 5);
         $products = $this->productRepository->findAll();
+        $shippingMethods = $this->shippingMethodRepository->findAll();
+        $shippingMethodsAvailable = (bool) count($shippingMethods);
```
lchrusciel:
```suggestion
        $shippingMethodsAvailable = count($shippingMethods) > 0;
```

lchrusciel: Or we can even skip this variable

src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php
```diff
@@ -175,13 +175,16 @@ private function generateItems(OrderInterface $order): void
...
 
         for ($i = 0; $i < $numberOfItems; ++$i) {
             /** @var OrderItemInterface $item */
             $item = $this->orderItemFactory->createNew();
 
             $product = $this->faker->randomElement($products);
             $variant = $this->faker->randomElement($product->getVariants()->toArray());
+            $variant->setShippingRequired($shippingMethodsAvailable);
```
lchrusciel: Later, we should improve these fixtures and allow for configuration from yaml file

lchrusciel:
And welcome to our contributors :)

src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php
TiMESPLiNTER: Should I improve the issue description and we keep it open or should I create a new issue for making it configurable?

src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php
TiMESPLiNTER: I want to keep it as I don't want to count n times the same result set in the for loop. :)

lchrusciel:
Thank you, Pascal! :tada:

PR #10506, opened 2019-07-06
[RFC] Limited: LazyOption::findAll returned records count

### LINKED ISSUE #10493: LazyOption::randomOne* functions uses findAll which loads all data
**Sylius version affected**: any

**Description**  
`LazyOption`'s `randomOne`, `randomOneOrNull` and `randomOnes` functions uses `findAll` which trying to load all data from repository. This become a problem when you have thousands of records in that repository.

**Steps to reproduce**  
Try to build multi-channel (5-10 channels) store fixtures with close-to-realworld set of taxons (100-200), with products (even few per taxon) with size options (few sizes per product). After that getting random product variant will take infinity time/resources.

**Possible Solution**  
Not sure, maybe one of:
- Select random chunks per one db request and only then pass to php's `array_rand`
- Check table records count - if there not much records - use `findAll` as before, if more than X000 - then use `rand()` in db query.

igormukhingmailcom:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no for core fixtures
| Deprecations?   | no
| Related tickets | fixes #10493 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/OptionsResolver/LazyOption.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/OptionsResolver/LazyOption.php
@@ -14,6 +14,8 @@ declare(strict_types=1);
 namespace Sylius\Bundle\CoreBundle\Fixture\OptionsResolver;
 
 use Doctrine\Common\Collections\Collection;
+use Sylius\Bundle\ResourceBundle\Doctrine\ORM\EntityRepository;
+use Sylius\Component\Resource\Model\ResourceInterface;
 use Sylius\Component\Resource\Repository\RepositoryInterface;
 use Symfony\Component\OptionsResolver\Options;
 use Webmozart\Assert\Assert;
@@ -39,7 +41,7 @@ final class LazyOption
     public static function randomOne(RepositoryInterface $repository): \Closure
     {
         return function (Options $options) use ($repository) {
-            $objects = $repository->findAll();
+            $objects = self::findLimitedAll($repository);
 
             if ($objects instanceof Collection) {
                 $objects = $objects->toArray();
@@ -58,7 +60,7 @@ final class LazyOption
                 return null;
             }
 
-            $objects = $repository->findAll();
+            $objects = self::findLimitedAll($repository);
 
             if ($objects instanceof Collection) {
                 $objects = $objects->toArray();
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/CoreBundle/Fixture/OptionsResolver/LazyOption.php
```diff
@@ -133,4 +135,8 @@ public static function findOneBy(RepositoryInterface $repository, string $field)
             return $repository->findOneBy([$field => $previousValue]);
         };
     }
+
+    private static function findLimitedAll(RepositoryInterface $repository, int $limit = 2000) {
```
Zales0123: Move bracket to the next line and add a return type

igormukhingmailcom: Fixed

src/Sylius/Bundle/CoreBundle/Fixture/OptionsResolver/LazyOption.php
```diff
@@ -133,4 +134,9 @@ public static function findOneBy(RepositoryInterface $repository, string $field)
             return $repository->findOneBy([$field => $previousValue]);
         };
     }
+
+    private static function findLimitedAll(RepositoryInterface $repository, int $limit = 2000): array
```
lchrusciel: I know that it may be annoying, that I'm just randomly jumping into this PR (and some changes are done due to my previous review), but after some time I would say that giving a limit here doesn't make any sense. It is not possible to use it anyway (class is final)

igormukhingmailcom: Its ok :)

As for this function - it used inside `LazyOption` class (all `findAll`s replaced by this function call). Please see full diff for context:
https://github.com/Sylius/Sylius/pull/10506/files

lchrusciel: I get it. What I try to say, that adding a `limit` parameter here makes no sense, as it is not used anywhere. This function is private and it is used 4 times, always without second parameter

igormukhingmailcom: Ah, you mean I should just inject that code directly. Will do that soon

lchrusciel: You can just remove this `limit` option here or propagate it's usage up to the functions that use them like
```php
public static function all(RepositoryInterface $repository, int $limit = 2000): \Closure
```

igormukhingmailcom: Removed (replaced by private constant)

PR #10509, opened 2019-07-08
[Admin] Add link to product in variant breadcrumb

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

**New link:**
Screenshot 2019-07-08 at 10 38 28

**After clicking:**
**to url:**
Screenshot 2019-07-08 at 13 59 57

**Product edit page:**
Screenshot 2019-07-08 at 14 00 05

PR #10512, opened 2019-07-09
[Admin] Improve breadcrumbs (especially for ProductVariants and PromotionCoupons)

CoderMaggie:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | no
| New feature?    | rather an improvement
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | triggered by #10509 😄 
| License         | MIT

The clue is to use entities **names** instead of **codes**, wherever it is possible. especially in entities that are _under_ other entities like coupons or variants.

Product variant edit:
Before:
image

After:
image

Without the variant name:
image

Channels edit:
Before:
image

After:
image

 And while trying to unsuccessfully remove the name:
image

After (translation of names to Admin locale)
image
image
image

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/ProductVariant/Update/_breadcrumb.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/ProductVariant/Update/_breadcrumb.html.twig
@@ -3,9 +3,9 @@
 {% set breadcrumbs = [
         { label: 'sylius.ui.administration'|trans, url: path('sylius_admin_dashboard') },
         { label: 'sylius.ui.products'|trans, url: path('sylius_admin_product_index') },
-        { label: app.request.attributes.get('productId'), url: path('sylius_admin_product_update', {'id': app.request.attributes.get('productId')}) },
-        { label: (metadata.applicationName~'.ui.'~metadata.pluralName)|trans, url: path(configuration.getRouteName('index'), configuration.vars.route.parameters|default({})) },
-        { label: resource.code|default(resource.id) },
+        { label: resource.product.name, url: path('sylius_admin_product_update', {'id': resource.product.id}) },
+        { label: 'sylius.ui.variants'|trans, url: path('sylius_admin_product_variant_index', {'productId': resource.product.id}) },
+        { label: resource.name|default(resource.code) },
         { label: 'sylius.ui.edit'|trans }
     ]
 %}
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Crud/Update/_breadcrumb.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Crud/Update/_breadcrumb.html.twig
@@ -3,7 +3,7 @@
 {% set breadcrumbs = [
         { label: 'sylius.ui.administration'|trans, url: path('sylius_admin_dashboard') },
         { label: (metadata.applicationName~'.ui.'~metadata.pluralName)|trans, url: path(configuration.getRouteName('index'), configuration.vars.route.parameters|default({})) },
-        { label: resource.code|default(resource.id)},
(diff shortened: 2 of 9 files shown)


src/Sylius/Bundle/AdminBundle/Resources/config/routing/product.yml
```diff
@@ -79,3 +79,19 @@ sylius_admin_product_create_simple:
                     form: "@SyliusAdmin/Product/_form.html.twig"
                 route:
                     name: sylius_admin_product_create_simple
+
+sylius_admin_product_update:
```
GSadee: You should add `update` to except actions in `sylius_admin_product` route (line 6) or even change this line to `only: ['delete']`

src/Sylius/Bundle/AdminBundle/Resources/config/routing/promotion.yml
```diff
@@ -18,3 +18,19 @@ sylius_admin_promotion:
                 templates:
                     form: "@SyliusAdmin/Promotion/_form.html.twig"
     type: sylius.resource
+
+sylius_admin_promotion_update:
```
GSadee: As above

src/Sylius/Bundle/AdminBundle/Resources/views/Product/Update/_breadcrumb.html.twig
```diff
@@ -0,0 +1,11 @@
+{% import '@SyliusAdmin/Macro/breadcrumb.html.twig' as breadcrumb %}
+
+{% set breadcrumbs = [
+        { label: 'sylius.ui.administration'|trans, url: path('sylius_admin_dashboard') },
+        { label: 'sylius.ui.products'|trans , url: path('sylius_admin_product_index') },
+        { label: resource.name|trans },
```
GSadee: It doesn't work this way, name that is returned from resource is already translated

pjedrzejewski:
IIRC the reason to use codes was to avoid issues when editing. Because if you try to change name to empty, the breeadcrumb will take the data from the entity, which will have empty value bound by the forms logic... :)

vvasiloi:
> IIRC the reason to use codes was to avoid issues when editing. Because if you try to change name to empty, the breeadcrumb will take the data from the entity, which will have empty value bound by the forms logic... :)

@pjedrzejewski `resource.name|default(resource.code)`? 😄

CoderMaggie:
The editing issue is very minor compared to the logic of displaying codes (that are an internal property used for processing things), while we have names (that are designed to be displayed everywhere). :)

I guess the weirdy forms' logic is not an issue that should be solved by displaying our immutable codes instead of names.

Although I will use the tip from @vvasiloi for now, thanks 💪

src/Sylius/Bundle/AdminBundle/Resources/views/ProductVariant/Update/_breadcrumb.html.twig
```diff
@@ -3,9 +3,9 @@
 {% set breadcrumbs = [
         { label: 'sylius.ui.administration'|trans, url: path('sylius_admin_dashboard') },
         { label: 'sylius.ui.products'|trans, url: path('sylius_admin_product_index') },
-        { label: app.request.attributes.get('productId'), url: path('sylius_admin_product_update', {'id': app.request.attributes.get('productId')}) },
+        { label: resource.product.name|trans, url: path('sylius_admin_product_update', {'id': app.request.attributes.get('productId')}) },
```
GSadee:
```suggestion
        { label: resource.product.name|trans, url: path('sylius_admin_product_update', {'id': resource.product.id}) },
```

Zales0123:
@CoderMaggie rebase should fix the build :)

GSadee:
Thank you, Magdalena! :tada:

PR #10516, opened 2019-07-09
Fix Psalm false-positives

pamil:
Psalm does not recognize conditional definitions like the one used by Symfony's EventDispatcherInterface: https://github.com/symfony/contracts/blob/master/EventDispatcher/EventDispatcherInterface.php#L16


Zales0123:
Thanks, Kamil! :tada:

PR #10518, opened 2019-07-10
[Admin] Unify order link in Orders, Payments, Shipments

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

**Add link in orders**
Screenshot 2019-07-12 at 07 38 21

**Add link in payments and delete Show button**
Screenshot 2019-07-12 at 07 39 12

**Add link in shipment and delete Show button**
Screenshot 2019-07-12 at 07 39 25

### DIFF
--- a/src/Sylius/Behat/Page/Admin/Payment/IndexPage.php
+++ b/src/Sylius/Behat/Page/Admin/Payment/IndexPage.php
@@ -33,9 +33,9 @@ final class IndexPage extends BaseIndexPage implements IndexPageInterface
         return $this->getField($orderNumber, 'state')->getText();
     }
 
-    public function showOrderPage(string $orderId): void
+    public function showOrderPageForNthPayment(int $position): void
     {
-        $this->getDocument()->find('css', '.table tr td:nth-child(2)')->clickLink($orderId);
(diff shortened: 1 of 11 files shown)


src/Sylius/Behat/Page/Admin/Payment/IndexPage.php
```diff
@@ -33,9 +33,9 @@ public function getPaymentStateByOrderNumber(string $orderNumber): string
         return $this->getField($orderNumber, 'state')->getText();
     }
 
-    public function showOrderPageForNthPayment(int $position): void
+    public function showOrderPage(string $orderId): void
```
CoderMaggie: I do not understand why are you changing this method so much. In my opinion in this step definition, we should be still finding a row with specific payment, finding by order ID is not a good solution as there can be many payments associated with one order :)

Tomanhez: You a right, I will fix it

GSadee:
Thanks, Tomasz! :tada:

PR #10524, opened 2019-07-12
Fix form validation in variant generate

### LINKED ISSUE #10515: Form validation error after add new channel and generate variants for products
**Sylius version affected**: 1.4.0

**Description**  
Form validation throws errors after add new channel and generate variants for products
- as an admin
- I have store with one or more channels
- add new channel is disabled
- in random product i chose menageVariants->Generate-> I do not set price in new added channel 
- click generate
- I get below screen:
Screenshot 2019-07-09 at 14 20 24

**Steps to reproduce**  

    @managing_product_variants
    Feature: Product variant validation
        In order to avoid making mistakes when managing a product variant
        As an Administrator
        I want to be prevented from adding it without specifying required fields

    Background:
        And the store has a "Wyborowa Vodka" configurable product

    @ui
    Scenario: Adding a new product variant without specifying its price in new added channel
        Given this product has "Wyborowa Exquisite" variant priced at "$90" identified by "VODKA_WYBOROWA_PREMIUM"
        And I add new channel in "Europe"
        And I want to create a new variant of this product
        And I do not set its price in "Europe" channel
        And I try to add it
        Then I should be notified that prices in all channels must be defined
**Possible Solution**  
Add constraints in a form validation

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Fixed #10515

When channel is disabled, validating working ok:
Screenshot 2019-07-12 at 14 47 56

### DIFF
--- a/features/product/managing_product_variants/generating_product_variant_after_add_new_channel_validation.feature
+++ b/features/product/managing_product_variants/generating_product_variant_after_add_new_channel_validation.feature
@@ -1,0 +1,19 @@
+@managing_product_variants
+Feature: Generating product variant after add new channel
+    In order to avoid making mistakes when generating variants
+    As an Administrator
+    I want to be prevented from generating it without specifying required fields
+
+    Background:
+        Given the store operates on a channel named "United States"
+        And the store also operates on a channel named "Europe"
+        And the store has a "Wyborowa Vodka" configurable product
+        And this product has option "Taste" with values "Orange" and "Melon"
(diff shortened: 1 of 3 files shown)


Zales0123:
For sure we need some scenario, that reproduces the issue 🚀

PR #10531, opened 2019-07-17
Improve filters UI

kulczy:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/private/sass/_filters.scss
+++ b/src/Sylius/Bundle/UiBundle/Resources/private/sass/_filters.scss
@@ -1,0 +1,64 @@
+.sylius-filters {
+    display: flex;
+    flex-wrap: wrap;
+    margin-left: -10px;
+    margin-right: -10px;
+
+    &__field {
+        flex-grow: 1;
+        min-width: 360px;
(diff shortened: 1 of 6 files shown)


Roshyo:
IMO, filters should be set in the <th> of each attribute if possible, and in a separate block if they deal with multiple/no specific attribute.

That's what any other CMS do and it feels very intuitive.

The matching can be done by name (replacing '.' by '_') or with a specific configuration line.

And for the filter block itself. The global width of number inputs is way too large.

vvasiloi:
In the first screenshot the number value field has an icon, what's that? Should it be for the customer input?

kulczy:
@vcraescu my bad, it's just a chrome addon :D

gabiudrescu:
poor @vcraescu... @kulczy I thought you meant @vvasiloi 

they are both Moldavians, but there's a river separating them.

vcraescu:
@kulczy it's ok! 😁

GSadee:
Thanks, Szymon! :tada:

PR #10538, opened 2019-07-18
[Shop] Use a menu builder for the tabs on the product page

mbabker:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | probably
| Deprecations?   | yes 
| Related tickets | N/A
| License         | MIT

Right now there isn't a way for plugins to hook into the frontend product page and add new tabs to the display, so this pull request aims to make that a possibility using the same concept from https://github.com/Sylius/Sylius/pull/7468.

Note, I list BC break as "probably" because the `@SyliusShop/Product/Show/_tabs.html.twig` and `@SyliusShop/Product/Show/Tabs/_menu.html.twig` layouts are completely rewritten, which theoretically isn't an issue if you're just using the default layouts or you've overridden both of them in your own install; the default layout still produces the same end result and the override flow isn't impacted by this.  What probably is a break is if you've overwritten the `@SyliusShop/Product/Show/_tabs.html.twig` layout and not the `@SyliusShop/Product/Show/Tabs/_menu.html.twig` layout, but I don't have any bright ideas on this one without coming up with some awkward naming conventions for the menu builder based layouts.

### DIFF
--- a/UPGRADE-1.6.md
+++ b/UPGRADE-1.6.md
@@ -5,3 +5,7 @@ Require upgraded Sylius version using Composer:
 ```bash
 composer require sylius/sylius:~1.6.0
 ```
+
+### Deprecations
+
+The `@SyliusShop/Product/Show/Tabs/_content.html.twig` template is deprecated, product tabs should be rendered with the `sylius.shop.product_tabs` menu instead
--- a/UPGRADE-2.0.md
+++ b/UPGRADE-2.0.md
@@ -39,3 +39,7 @@
     * The following configuration options were removed:
 
         * `sylius.mailer.templates`
+
+    * The following templates were removed:
+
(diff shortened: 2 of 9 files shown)


mbabker:
I think the test failures here are actually a case of exposing another bug or bad logic.  The current `@SyliusShop/Product/Show/Tabs/_menu.html.twig` layout hides the attributes tab if the product doesn't have any, but the `@SyliusShop/Product/Show/Tabs/_content.html.twig` layout doesn't do the same check so it always renders the tab content for the attributes tab.  With the changes in this PR, that tab's content no longer renders and that's where the tests are failing.

vvasiloi:
@mbabker Hey, why did you give up on this?

Issue #9479, opened 2018-06-11
Wrong flag icon of Czech republic in administration


hladpe:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.1.6

Czech language (cs_CZ) has displayed wrong Serbian flag (css class "cs") instead of correct one (css class "cz"). This bug is in administration for example specifically during creating or editing taxon, product,...

![sylius-sylius-czech-flag](https://user-images.githubusercontent.com/23253970/41240171-31d68422-6d9a-11e8-802d-fe3a97aeece0.png)

lchrusciel:
The problem probably has its roots in Symfony Intl component.

pamil:
Fixed by #10553.

PR #10549, opened 2019-07-24
Fix country flag selection by locale

### LINKED ISSUE #10416: Wrong Semantic UI flag usage
**Sylius version affected**: (probably) all

**Description**  
Sylius is showing flags next to the translation form BUT Semantic UI flags are designed to be **country** flags, Sylius is using them as **language** flags. Why does it matter?

Semantic UI is using [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) codes
Sylius (Symfony) is using [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) codes

So in some cases wrong flags are displayed.

Affected countries:

| Country | Language | Used code (ISO 639-1) | Needed code (ISO 3166-2) | Displayed flag |
| --- | --- | --- | --- | --- |
| Czech republic | Czech | cs | cz | Serbia |
| Myanmar | Burmese | my | mm | Malaysia |
| Netherlands Antilles | Dutch | nl | an | Netherlands |
| Serbia | Serbian | sr | cs / rs | Suriname |
| Montenegro | Montenegrin | - | me | - |
| Ghana | Akan | ak | gh | - |
| Ethiopia | Amharic | am | et | Armenia |
| India | Hindi | hi | in | - |
| Belarus | Belarusian | be | by | Belgium |
| Denmark | Danish | da | dk | - |
| Sweden | Swedish | sv | se | El Salvador |
| Estonia | Estonian | et | ee | Ethiopia |

_The list is not complete because there is so many of them, but you got the point._

**Steps to reproduce**  
1) Go to the route `/admin/locales/`
2) Add `Czech` locale
3) See the list of available locales
4) In column `name` you can see `cs Czech` with flag which is not Czech.

**Possible Solution**  
1) Create mapping file which will convert ISO 639-1 to ISO 3166-2 but that doesn't sounds right

or

2) Use country name which we already have. Semantic UI allows us to use `<i class="czech republic flag"></i>` instead of `<i class="cz flag"></i>`

sim2github:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10416, #9479
| License         | MIT

Sometimes happens that the [locale](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) and [country](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) codes are not equal.

That is why for some languages has [another flag](https://semantic-ui.com/elements/flag.html) displayed or flag completely absent.

At that moment, Ukraine(UA) become a colony of the United Kingdom(UK), Vietnam(VN) been invaded by the US Virgin Islands(VI), and the Czech Republic(CZ) has declared itself Serbia(CS) ...

☮️🌍

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/views/Macro/flags.html.twig
+++ b/src/Sylius/Bundle/UiBundle/Resources/views/Macro/flags.html.twig
@@ -4,5 +4,39 @@
 
 {% macro fromLocaleCode(locale_code) %}
     {% import _self as flags %}
-    {{ flags.fromCountryCode(locale_code|slice(-2)) }}
+    {% set locale_country = { 
+  		'ak':'gh',
+  		'am':'et',
+  		'am':'et',
+  		'be':'by',
+  		'bs':'ba',
+  		'cs':'cz',
+  		'cy':'wls',
+  		'da':'dk',
+  		'el':'gr',
+  		'et':'ee',
+  		'ga':'ie',
+  		'gd':'sct',
+  		'hi':'in',
+  		'hy':'am',
+  		'ja':'jp',
+  		'kk':'kz',
+  		'lo':'la',
+  		'my':'mm',
+  		'nl':'an',
+  		'sq':'al',
+  		'sr':'cs',
+  		'sv':'se',
+  		'tg':'tj',
+  		'tl':'ph',
+  		'uk':'ua',
+  		'vi':'vn'  
+      } 
+    %}
+    {% set locale_code = locale_code|slice(-2) %}
+    {% if locale_code in locale_country|keys %}
+        {{ flags.fromCountryCode(locale_country[locale_code]) }}
+    {% else %}
+        {{ flags.fromCountryCode(locale_code) }}
+    {% endif %}
 {% endmacro %}


vvasiloi:
Hey @sim2github, did you try my suggestion from https://github.com/Sylius/Sylius/issues/10416#issuecomment-496618688?

sim2github:
@vvasiloi, i doesn't use any libs and extension from your comment. The problem is only for locale languages without defined regions. I use array-map that set fallback country value for some of them.
Its easy to use and overwrite (maybe  someone will use another pack of icons / maybe I miss some values). 

PS: just leave here [list](https://wiki.openstreetmap.org/wiki/Nominatim/Country_Codes)

vvasiloi:
I strongly believe that for locales without a specific region no flag should be displayed.
A lot of resourses backing this idea can be found on internet, to name a few:
http://www.flagsarenotlanguages.com/blog/
https://unitedlanguagegroup.com/blog/inside-design-a-language-selector-is-no-place-for-flags/

vvasiloi:
@sim2github Can you take a look at the counter-PR I launched https://github.com/Sylius/Sylius/pull/10553 and give some feedback?
I believe it's a better solution.

pamil:
Thanks for the PR @sim2github, I merged @vvasiloi's PR which seems a bit more generic though 🎉

PR #10550, opened 2019-07-24
Allow easier decoration/extension of the Order fixture

TiMESPLiNTER:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

I'm not really happy with the new `createOrder` method as it accepts three arrays which are then used to fetch a random element. But it's also not acceptable to call `findAll` n times just to pick one random element out of the collection. Any ideas how to solve this better? One solution might be properties holding the the three result sets...


lchrusciel:
Closing in favor of #10550

PR #10552, opened 2019-07-25
[Order] Change OrderItemController methods to protected

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Even though I love **Composition over inheritance** principle, this controller is already poorly written and, I believe, we should make its customization as painless as possible 🔪 🤕


pamil:

PR #10562, opened 2019-07-29
Avoid js when removing product from cart

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Welcome to the continuation of my war against JS scenarios in Sylius ⚔️ :D (see https://github.com/Sylius/Sylius/pull/10396). The idea is the same as previously, to avoid javascript scenarios whenever we can.

Currently, each scenario that contains some step with removing an item from the cart needs to be tagged as `@javascript`, as this functionality is done with the ajax call. We can easily avoid it, by setting a proper `form` attribute on each item quantity and coupon fields, as well as update, remove an item and apply coupon buttons.

The good thing is - no PHP code is changed 😄 only twigs, js and a new route in yaml. We should also probably deprecate an old `sylius_shop_ajax_cart_item_remove` route (or shouldn't we?).

As a result, the build should be at least one minute faster 🚤 (it's a good step forward, anyway 🚀)


pamil:

PR #10564, opened 2019-07-31
[Fixture] Improve order fixture

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Fixes 2 possible bugs/inaccuracies during fixture orders creation:
- does not allow to create an order with products that are not available in order's channel (especially useful, when you create more than one channel in fixtures 💃 )
- does not allow to duplicate variant within order items (the quantity is now increased in the existing item)

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php
@@ -21,6 +21,7 @@ use Sylius\Component\Core\Checker\OrderShippingMethodSelectionRequirementChecker
 use Sylius\Component\Core\Model\AddressInterface;
 use Sylius\Component\Core\Model\OrderInterface;
 use Sylius\Component\Core\Model\OrderItemInterface;
+use Sylius\Component\Core\Model\ProductInterface;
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php
```diff
@@ -175,17 +176,34 @@ private function generateItems(OrderInterface $order): void
     {
         $numberOfItems = random_int(1, 5);
         $products = $this->productRepository->findAll();
```
lchrusciel: Can we just fetch products available in orders channel?

lchrusciel:
Thank you, Mateusz! :tada:

src/Sylius/Bundle/CoreBundle/Fixture/OrderFixture.php
Zales0123: We could if we write a custom repository method for that 👌 As we have a collection of channels not a single channel on the product. But it, of course, could be improved 👍

PR #10573, opened 2019-08-05
Date field type doesn't exist

mikemix:
When using the `date` I get:

```
An exception has been thrown during the rendering of a template ("Grid field service "date" does not exist, available grid field services: "currency", "datetime", "string", "twig"").
```

| Q               | A
| --------------- | -----
| Branch?         | >= 1.4
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/docs/components_and_bundles/bundles/SyliusGridBundle/configuration.rst
+++ b/docs/components_and_bundles/bundles/SyliusGridBundle/configuration.rst
@@ -54,13 +54,14 @@ Here you will find all configuration options of ``sylius_grid``.
                         form_options: { }
                         default_value: ~
                     date:
-                        type: date # Type of filter
+                        type: datetime # Type of filter
                         label: app.ui.created_at
                         enabled: true
                         template: ~
                         position: 100
                         options:
                             field: createdAt
+                            format: 'Y-m-d H:i' # optional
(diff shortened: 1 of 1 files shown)


docs/components_and_bundles/bundles/SyliusGridBundle/configuration.rst
```diff
@@ -54,13 +54,14 @@ Here you will find all configuration options of ``sylius_grid``.
                         form_options: { }
                         default_value: ~
                     date:
-                        type: date # Type of filter
+                        type: datetime # Type of filter
```
Zales0123: Are you sure it does not work? As far as I see, it's not a field type but filter type (and it exists https://github.com/Sylius/SyliusGridBundle/blob/1.4/src/Component/Filter/DateFilter.php)

mikemix: I posted the exception I get when using the `date` type.

mikemix: Oh god my mistake.

PR #10574, opened 2019-08-05
[PayumBundle] Fixed payment amount when using partial payments

### LINKED ISSUE #8958: [PayumBundle] Wrong capture total amount, when partial payment is used
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | yes
| Sylius version   | master

Issue is that CaptureAction assumes, that payment covers order fully, which isn't the case by design as partial payments are supported.

Consequently I've encountered this issue on custom partial payment flow, so I would think it's not reproducible with vanilla Sylius.

```
---             $totalAmount = $order->getTotal();
+++             $totalAmount = $payment->getAmount();
                $payumPayment = new PayumPayment();
                $payumPayment->setNumber($order->getNumber());
```

https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/PayumBundle/Action/CapturePaymentAction.php#L63

Same for description provider.
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/PayumBundle/Provider/PaymentDescriptionProvider.php

Should I do PR?

Regards.

Leobaillard:
| Q               | A
| --------------- | -----
| Branch?         | 1.5
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #8958
| License         | MIT

As mentioned in #8958, the `AuthorizePaymentAction` and `CapturePaymentAction` (as well as the `PaymentDescriptorProvider`) should use the payment amount instead of the total order amount by default. This allows the use of partial payments as they are supported by Sylius.

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/Action/AuthorizePaymentAction.php
+++ b/src/Sylius/Bundle/PayumBundle/Action/AuthorizePaymentAction.php
@@ -52,10 +52,9 @@ final class AuthorizePaymentAction extends GatewayAwareAction
                 $this->gateway->execute($convert = new Convert($payment, 'array', $request->getToken()));
                 $payment->setDetails($convert->getResult());
             } catch (RequestNotSupportedException $e) {
-                $totalAmount = $order->getTotal();
                 $payumPayment = new PayumPayment();
                 $payumPayment->setNumber($order->getNumber());
-                $payumPayment->setTotalAmount($totalAmount);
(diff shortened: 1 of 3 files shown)


Zales0123:
Closing in favor of https://github.com/Sylius/Sylius/pull/10613

PR #10595, opened 2019-08-20
Use {{ limit }} to allow min/max value update

Prometee:
| Q               | A
| --------------- | -----
| Branch?         | all
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT


Prometee:
No problem, I just push it.

I think there is other spots where this enhancements could be applied but for now I only found this because I would like to apply an other constraint on ZipCode field by adding a `max_limit`.

lchrusciel:
Thanks, Francis! :tada:

PR #10643, opened 2019-08-29
[Fixture] channel fixture fix

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         |  master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
Deprecations?   | no
| License         | MIT


src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures/shop_configuration.yaml
```diff
@@ -29,7 +29,7 @@ sylius_fixtures:
                     options:
                         custom:
                             us_web_store:
```
Zales0123: Then we should probably change this key as well

GSadee:
Thanks, Adam! :tada:

Issue #10648, opened 2019-09-02
[RFC] Add dependencies check with maglnet/composer-require-checker


mmenozzi:
Currently, in the Sylius source (the `src` folder) are being used many symbols which are not required directly but are anyway installed because them are required by a Sylius dependency. 

This may lead to unexpected BC breaks (this is well explained here: https://github.com/maglnet/ComposerRequireChecker#whats-it-about).

To prevent this situation we could use [maglnet/composer-require-checker](https://github.com/maglnet/ComposerRequireChecker).

I tried to run it on the current `master` branch and this is the result:

```
ComposerRequireChecker unknown-development
The following unknown symbols were found:
+---------------------------------------------------------------------------+--------------------+
| unknown symbol                                                            | guessed dependency |
+---------------------------------------------------------------------------+--------------------+
| FriendsOfBehat\PageObjectExtension\Page\SymfonyPageInterface              |                    |
| FriendsOfBehat\PageObjectExtension\Page\SymfonyPage                       |                    |
| Behat\Mink\Exception\ElementNotFoundException                             |                    |
| FriendsOfBehat\PageObjectExtension\Page\PageInterface                     |                    |
| Behat\Mink\Driver\Selenium2Driver                                         |                    |
| FriendsOfBehat\PageObjectExtension\Page\UnexpectedPageException           |                    |
| Behat\Mink\Session                                                        |                    |
| Behat\Mink\Exception\UnsupportedDriverActionException                     |                    |
| Behat\Mink\Element\NodeElement                                            |                    |
| FriendsOfBehat\PageObjectExtension\Page\Page                              |                    |
| Behat\Behat\Context\Context                                               |                    |
| Doctrine\Common\Persistence\ObjectManager                                 |                    |
| Behat\Gherkin\Node\TableNode                                              |                    |
| SM\Factory\FactoryInterface                                               |                    |
| Symfony\Component\BrowserKit\Client                                       |                    |
| Symfony\Component\BrowserKit\Cookie                                       |                    |
| json_decode                                                               | ext-json           |
| Doctrine\DBAL\DBALException                                               |                    |
| Doctrine\DBAL\Exception\ForeignKeyConstraintViolationException            |                    |
| FriendsOfBehat\PageObjectExtension\Element\Element                        |                    |
| PhpSpec\ObjectBehavior                                                    |                    |
| Mockery\MockInterface                                                     |                    |
| Prophecy\Argument                                                         |                    |
[...]

Issue #10530, opened 2019-07-17
Test environment broken with php 7.2.20


JakobTolkemit:
**Sylius version affected**: >=1.4.2

**Description**  

We encountered this error when we tried to clear the cache:
```sh
[Symfony\Component\Debug\Exception\FatalErrorException]
Error: During class fetch: Uncaught ReflectionException: Class HWI\Bundle\OAuthBundle\Connect\AccountConnectorInterface not found
  in /var/www/html/sylius/vendor/sylius/sylius/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php:36
Stack trace:
#0 /var/www/html/sylius/vendor/symfony/debug/DebugClassLoader.php(160): require('/var/www/html/s...')
#1 [internal function]: Symfony\Component\Debug\DebugClassLoader->loadClass('Sylius\\Bundle\\C...')
#2 [internal function]: spl_autoload_call('Sylius\\Bundle\\C...')
#3 /var/www/html/sylius/vendor/symfony/config/Resource/ClassExistenceResource.php(78): class_exists('Sylius\\Bundle\\C...')
#4 /var/www/html/sylius/vendor/symfony/dependency-injection/ContainerBuilder.php(353): Symfony\Component\Config\Resource\ClassExistenceResource->isFresh(0)
#5 /var/www/html/sylius/vendor/symfony/dependency-injection/Compiler/AutowirePass.php(336): Symfony\Component\DependencyInjection\ContainerBuilder->getReflectionClass('Sylius\\Bundle\\C...', false)
#6 /var/www/html/sylius/vendor/symfony/dep      
```

I found out that it's due to Symfony trying to load the `Sylius\Bundle\CoreBundle\OAuth\UserProvider`, when in the test environment, which implements the `HWI\Bundle\OAuthBundle\Connect\AccountConnectorInterface`. 
This interface comes with the `hwi/oauth-bundle` which doesn't get installed by Sylius-standard.

The only reason it worked before was a bug in Php that made it possible to define a class that implements a non existent interface.

This bug got fixed in Php 7.2.20. 
https://www.php.net/ChangeLog-7.php#7.2.20
https://bugs.php.net/bug.php?id=76980

**Steps to reproduce**  

Install Php 7.2.20 and try to clear the cache in the test environment.

**Possible Solution**  

Two solutions come to mind:
1. Require `hwi\oauth-bundle` in Sylius-standard
[...]

PR #10664, opened 2019-09-09
Add composer-require-checker to the CI build

mmenozzi:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #10648 
| License         | MIT


mmenozzi:
@pamil @lchrusciel @Zales0123 what do you think?

etc/travis/suites/application/script/validate-composer-dependencies
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+source "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/../../../../bash/common.lib.sh"
+
+print_header "Validating (Composer require checker)" "Sylius"
+run_command "(composer global require maglnet/composer-require-checker && $HOME/.composer/vendor/bin/composer-require-checker --config-file=composer-require-checker.json)"
```
lchrusciel: Shouldn't `composer global require `maglnet/composer-require-checker` be part of some install step with the possibility to add this library to cache?

composer.json
```diff
@@ -63,6 +77,7 @@
         "symfony/asset": "^3.4|^4.3",
         "symfony/config": "^3.4|^4.3",
         "symfony/console": "^3.4|^4.3",
+        "symfony/debug-bundle": "^3.4|^4.3",
```
pamil: This bundle is only used in dev / test environment, it shouldn't be required for consumers of `sylius/sylius` package.

composer.json
```diff
@@ -91,13 +106,16 @@
         "symfony/translation": "^3.4|^4.3",
         "symfony/twig-bundle": "^3.4|^4.3",
         "symfony/validator": "^3.4|^4.3",
+        "symfony/web-profiler-bundle": "^3.4|^4.3",
```
pamil: Same here.

etc/travis/suites/application/script/validate-composer-dependencies
pamil: I'd vote for downloading the PHAR and caching it (it's the recommended way by the package itself). We do the same with Selenium now.

composer.json
```diff
@@ -20,33 +20,47 @@
...
         "doctrine/orm": "^2.5",
+        "doctrine/persistence": "^1.1",
         "egulias/email-validator": "~2.0",
         "friendsofsymfony/oauth-server-bundle": "^1.6",
         "friendsofsymfony/rest-bundle": "^2.1",
         "fzaninotto/faker": "^1.6",
         "gedmo/doctrine-extensions": "^2.4.12",
+        "guzzlehttp/guzzle": "^6.3",
+        "guzzlehttp/psr7": "^1.6",
+        "hwi/oauth-bundle": "^0.6",
```
pamil: Wondering about this one - we only provide an optional integration with it, it shouldn't be required by all the consumers (that's why it was in require-dev).

composer.json
```diff
@@ -91,13 +106,16 @@
         "symfony/translation": "^3.4|^4.3",
         "symfony/twig-bundle": "^3.4|^4.3",
         "symfony/validator": "^3.4|^4.3",
+        "symfony/web-profiler-bundle": "^3.4|^4.3",
+        "symfony/web-server-bundle": "^3.4|^4.3",
```
oallain: I think this one should also be placed in the "_require-dev_".

composer.json
mmenozzi: Do you mean only `hwi/oauth-bundle` and not also `guzzlehttp` packages, right?
If so, yes it's true, `hwi/oauth-bundle` is an optional dependency. I have to think about how to handle it. I think that the right thing to do here would be move this oauth feature in a separate Sylius bundle/plugin where the `hwi/oauth-bundle` is a direct dependency. But it's too much work and should be done in a separate PR.

mmenozzi: Basically I fully agree with @leonboot here https://github.com/Sylius/Sylius/issues/10530#issuecomment-515063529.
So @pamil I think that we should choose:

1. Move the `Sylius\Bundle\CoreBundle\OAuth\UserProvider` to an external package
2. Require `hwi/oauth-bundle` as strict/direct depencency of Sylius

I'd choose for the option 2. I understand that is an optional integration but I think that requiring `hwi/oauth-bundle` solves more problems than it get.

Zales0123:
What is the status of this PR? It looks good for me, to give it a try 👌 cc @pamil @lchrusciel

composer.json
lchrusciel: I would suggest adding it to `suggest` and maybe be more clear about this dependency in docs. This is another library that didn't match a stable release. May also block us from Symfony 5.0 support. Less hard dependencies we have, easier it is to maintain the project.

But it is also a question to @CoderMaggie if we want to support login from 3 party apps by default.

vvasiloi:
Hey @mmenozzi, how do you fill about reviving this?

mmenozzi:
> Hey @mmenozzi, how do you fill about reviving this?

Hi @vvasiloi, I'd be happy to revive this if Sylius core team agree to merge this as soon as it will be ready.
Reviving this PR now, after 2 years, basically means that I need to start again from scratch and I'd like to be sure that it will be merged then.

vvasiloi:
Hey @mmenozzi, you have my full support and I can assure you that the core team won't let you down on this one the second time.

mmenozzi:
Ok @vvasiloi thank you! I'll do my best to get this work done asap.

.github/workflows/application.yml
```diff
@@ -113,6 +113,15 @@ jobs:
                 name: Run PHPStan
                 run: vendor/bin/phpstan analyse
                 if: always() && steps.end-of-setup.outcome == 'success'
+            
+            -
+                name: Run ComposerRequireChecker
+                run: |
+                    (cat composer.json | jq '.["autoload-dev"]["psr-4"] |= . + {"Sylius\\Behat\\": "src/Sylius/Behat/"}' | jq 'del(.autoload["psr-4"]["Sylius\\Behat\\"])') > _composer.json
```
vvasiloi, line 120: Can you explain this?

composer.json
```diff
@@ -21,24 +21,34 @@
...
         "doctrine/orm": "^2.7",
+        "doctrine/persistence": "^2.3",
         "egulias/email-validator": "^3.1",
         "fakerphp/faker": "^1.9",
         "friendsofphp/proxy-manager-lts": "^1.0",
         "friendsofsymfony/rest-bundle": "^3.0",
         "gedmo/doctrine-extensions": "^3.2",
+        "guzzlehttp/guzzle": "^6.5",
+        "guzzlehttp/psr7": "^1.8",
+        "hwi/oauth-bundle": "^1.4",
```
vvasiloi, line 51: This is a tough one. It's meant to be an optional dependency, but the moment I wrote these 2 words I remembered [this post](https://matthiasnoback.nl/2014/04/theres-no-such-thing-as-an-optional-dependency/) from Matthias Noback about optional dependencies, which I still agree with.
Keeping it in `require` might result in dependency issue on update for those who don't use the bundle.
I suggest to move it back to `require-dev`, ignore it in composer-require-checked config  and address the issue separately.
Perhaps this integration should move from Sylius Core to a plugin.

composer.json
```diff
@@ -71,6 +82,7 @@
         "symfony/asset": "^4.4 || ^5.4",
         "symfony/config": "^4.4 || ^5.4",
         "symfony/console": "^4.4 || ^5.4",
+        "symfony/debug-bundle": "^4.4 || ^5.4",
```
vvasiloi, line 85: This one must be in require-dev, otherwise it will be installed in production envs.

composer.json
```diff
@@ -107,12 +119,15 @@
         "symfony/translation": "^4.4 || ^5.4",
         "symfony/twig-bundle": "^4.4 || ^5.4",
         "symfony/validator": "^4.4 || ^5.4",
+        "symfony/web-profiler-bundle": "^4.4 || ^5.4",
+        "symfony/web-server-bundle": "^4.4",
```
vvasiloi, line 123: These 2 also must be in require-dev.

vvasiloi (changes requested):

.github/workflows/application.yml
mmenozzi: The `Sylius\Behat` namespace is in the `autoload` section. For ComposerRequireCheker this means that all symbols used in those classes must be in the `require` section of the `composer.json` and not in the `require-dev` one. Unless we really want to add all Behat, Mink and other packages to the `require` section we have to do that.

This is a workaround. IMHO the real solution is to move the whole Behat contexts, pages and services classes to a separate package that can be installed by those who needs them in their own Behat test suites. But obviously this is not in the scope of this PR.

vvasiloi: Thanks!

composer.json
mmenozzi: We already talked about this topic in this same PR [here](https://github.com/Sylius/Sylius/pull/10664#discussion_r331455461).

I'm still for having `hwi/oauth-bundle` as a strict/hard dependency. Remember that not having it in the `require` section already caused issues to others (see [here](https://github.com/Sylius/Sylius/issues/10530#issuecomment-515063529)).

Moreover in the meantime `hwi/oauth-bundle` has become stable.

vvasiloi: I missed the first conversation, but I remember the second one.
I still think that `hwi/oauth-bundle` should not be in the `require` section of `sylius/sylius` and that we need an alternative solution for it, like extracting this feature in a plugin.

mmenozzi:
Hi @vvasiloi, currently if someone uses the `Sylius\Bundle\CoreBundle\Application\Kernel` will get errors because many bundles registered there (for example the `Doctrine\Bundle\DoctrineCacheBundle\DoctrineCacheBundle`) are not required anymore.

How you'd handle that?

composer.json
mmenozzi: Done.

composer.json
mmenozzi: Done.

vvasiloi:
@mmenozzi no one should use that class in their application, it was deprecated in Sylius 1.3.

mmenozzi:
> @mmenozzi no one should use that class in their application, it was deprecated in Sylius 1.3.

Yes @vvasiloi I see that is deprecated since Sylius 1.3. But deprecated does not mean removed. Deprecated means that we should support it until Sylius 2.0. Instead that class does not work anymore since someone removed required packages from composer.json (for example the DoctrineCacheBundle).

1. If we really want to support that old Kernel class we should re-add those dependencies to the composer.json file.
2. Or we could decide that is not worth because, in the end, its support already ended several releases ago when someone removed required dependencies. If this is the case I'd completely remove that old Kernel class.
3. The last, and more messy option, is to keep that Kernel class without adding required dependencies to the composer.json and whitelist them in the ComposerRequireChecker configuration.

I'd go with the option number 2.

vvasiloi:
@mmenozzi Tough choice. I remembered now that Sylius broke that compatibility in 1.7: https://github.com/Sylius/Sylius/issues/12463.
Re-adding some of those dependencies will not be possible, because we're [dropping support for Symfony 4](https://github.com/Sylius/Sylius/pull/13489):
- [DoctrineCacheBundle ](https://github.com/doctrine/DoctrineCacheBundle) was deprecated and doesn't support Symfony 5.
- [SonataCoreBundle](https://github.com/sonata-project/SonataCoreBundle) was deprecated and doesn't support Symfony 5.
- [Symfony WebServerBundle](https://github.com/symfony/web-server-bundle) was deprecated, but supports Symfony 5.
- [SonataIntlBundle](https://github.com/sonata-project/SonataIntlBundle) is still maintained and supports Symfony 5 and 6.
- [DoctrineFixturesBundle](https://github.com/doctrine/DoctrineFixturesBundle) is still maintained and supports Symfony 5 and 6.

The `Sylius\Bundle\CoreBundle\Application\Kernel` class also uses `Symfony\Component\HttpKernel\Kernel::getRootDir`, which was deprecated in Symfony 4 and removed in Symfony 5.

I don't think it's worth supporting that upgrade path anymore, especially since it was unfortunately broken long ago.
The damage is already done, so we can probably remove it already...
In the current state, I think it's virtually impossible to upgrade past Sylius 1.7 without upgrading the directory structure and removing the use of `Sylius\Bundle\CoreBundle\Application\Kernel` or adding those dependencies to the end project's requirements.
@stefandoorn can probably attest to that.

We might be able to bring back the BC for this class in Sylius 1.10 and 1.11, because those still support Symfony 4, but again, I don't think it's worth it.

Anyway, regardless of which path we choose, it should not be part of this PR.

@Sylius/core-team I would love to see more opinions on this issue.

Roshyo:
I would also go for option 2. As you said, it has been broken for a while now, and upgrading to 5.x/6.x should be the way.

IMO it is okay to ask a few changes in a project when upgrading if there are valid reasons. In that case, as you said again, no-one should use this Kernel, so "forcing" them to make the switch is fine as long as it is documented in upgrade files.

mmenozzi:
> Anyway, regardless of which path we choose, it should not be part of this PR.

Ok @vvasiloi but this leads to option 3 for this PR. I don't like it that much because every symbol we whitelist in the ComposerRequireChecker configuration is global and not only for a certain usage. So we have to remember that there are those "cleanup PRs" to be made in the future.

I wonder if it would be better to delete that old Kernel class in this PR.

vvasiloi:
@mmenozzi True, it's going to be a bit messy. However, going from no dependency checks at all to checking most dependencies, but with quite a few exception is still good. We can open the follow-up PR(s) right after merging this.
I'm just worried that this PR will get stale again because of those tough decisions and I just want it to move forward ASAP and sort out everything else afterwards.

mmenozzi:
Totally agree on this @vvasiloi! Let's go ahead with option 3.

lchrusciel (approved):
Hey folks. Thanks a lot for comments and pushing this topic forward. I can agree, that the current solution is the best to move this topic forward and then do it case by case. At the current state, I would love to merge it, but the build is red.

mmenozzi:
Yes @lchrusciel it's red because it's still work in progress.

mmenozzi:
Hi @vvasiloi and @lchrusciel,
I think that this PR is now ready for merge.

I created separate commits for each issue that, in my opinion, must be fixed in separate PR. In those commits comment you can find more information about the issue, they are the following:

- https://github.com/Sylius/Sylius/pull/10664/commits/30c42ff5a45e593e894d13060c6593178f27888b: there is an old Kernel class that can be removed.
- https://github.com/Sylius/Sylius/pull/10664/commits/e09551c016752eff03339cc90ab678afabc3765d: the `Sylius\Bundle\CoreBundle\EventListener\CircularDependencyBreakingExceptionListener` uses `symfony/http-kernel` symbols from version 4 but the support for Symfony 4 has been removed in Sylius. I think that this issue is the most critical because currently if this class is used will throw a fatal error.
- https://github.com/Sylius/Sylius/pull/10664/commits/5d4765e4830c1b34a2dac074f7307bc5742d5655: `hwi/oauth-bundle` is an optional dependency (see [here](https://docs.sylius.com/en/1.9/cookbook/shop/facebook-login.html)). That's ok but this forced me to put its symbols in the ComposerRequireChecker whitelist. One way to handle that is to create a separate plugin that could be installed only if someone wants to use the Facebook login.

Remember that for ComposerRequireChecker every whitelisted symbol is global and not limited to a certain usage. For this reason there should be less whitelisted symbol as possible. Think for example to the `hwi/oauth-bundle` symbols. If someone uses it during the development of a new Sylius core feature without putting the dependency in the `require` section of the `composer.json` file, ComposerRequireChecker will not trigger any error because those symbols are whitelisted. This leads to bugs.

After this PR will be merged I will create 3 separate issues to track down those problems.

vvasiloi (approved):

vvasiloi:
Thank you, @mmenozzi! Great work! 🥇 
It's now up to @lchrusciel to merge.

mmenozzi:
Thank you @vvasiloi!
@lchrusciel the build is green now, you can merge it! 😉

lchrusciel (approved):
Hey Manuele! 

Thanks a lot for your contribution! I feel sorry, that we keep you waiting for over two years, but luckily Victor bring this topic again to light. After all this time, lest finally merge it!

mmenozzi:
Yeah! Thank you guys! 🎉

PR #10665, opened 2019-09-10
Remove ShippingBundle spec from autoload-dev

### LINKED ISSUE #10657: Whole ShippingBundle in autoload-dev
Is it ok that the whole ShippingBundle is in the `autoload-dev` of the `composer.json` file?

https://github.com/Sylius/Sylius/blob/master/composer.json#L239

I know that it's not a big problem because it is also in the `autoload` with the `"Sylius\\Bundle\\": "src/Sylius/Bundle/"` mapping but it sounds weird that the whole bundle it's also in the `autoload-dev`. Especially because other bundles have only the `spec` classes in their `autoload-dev` mapping.

mmenozzi:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10657 
| License         | MIT


lchrusciel:
Thank you, Manuele! :tada:

PR #10698, opened 2019-09-24
Allow custom productName and variantName

Prometee:
| Q               | A
| --------------- | -----
| Branch?         | all
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

When setting up a custom order sometime it's useful to set some additional informations about the variant or the product name. For exemple in my case it was to add a period infos on a renewal subscription order.

I set the test on `null` value but maybe it could be safer to test it for an empty value ?

### DIFF
--- a/src/Sylius/Component/Core/Order/OrderItemNamesSetter.php
+++ b/src/Sylius/Component/Core/Order/OrderItemNamesSetter.php
@@ -26,11 +26,11 @@ final class OrderItemNamesSetter implements OrderItemNamesSetterInterface
         foreach ($order->getItems() as $item) {
             $variant = $item->getVariant();
 
-            if (null !== $variant) {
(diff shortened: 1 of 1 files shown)


lchrusciel:
Once the name is set, we will never change it and I'm not sure if this is expected behavior. I would say, that your case is totally valid, but I would achieve it with this service decoration.

Prometee:
@lchrusciel That's what I'm doing (eg: decorate this service) to be able to change the product name but it could be easier to just alter the test here.

An other solution is to build two more services to be able to build our own variant name and product name. Maybe this could be better, what do you think of this enhancement ?

lchrusciel:
New service with the current implementation seems to be good choice. Then anyone can customize to their needs

PR #10699, opened 2019-09-25
Admin avatar - fixtures

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Screenshot 2019-09-25 at 09 29 01


src/Sylius/Bundle/CoreBundle/Fixture/Factory/AdminUserExampleFactory.php
```diff
@@ -32,15 +36,28 @@ class AdminUserExampleFactory extends AbstractExampleFactory implements ExampleF
...
+    /** @var FileLocatorInterface */
+    private $fileLocator;
+
+    /** @var ImageUploaderInterface */
+    private $imageUploader;
+
+    public function __construct(
+        FactoryInterface $userFactory,
+        string $localeCode,
+        FileLocatorInterface $fileLocator,
```
lchrusciel: 🚨 BC ALERT 🚨

New dependencies should be optional to keep backward compatibility. If not provided, then we should trigger error to inform, that it should be added

lchrusciel:
Thank you, Adam! :1st_place_medal:

PR #10708, opened 2019-09-26
fix channel name on product show page

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no/yes
| New feature?    | no/yes
| BC breaks?      | no/yes
| Deprecations?   | no/yes 
| License         | MIT

Screenshot 2019-09-26 at 10 09 00


Zales0123:
@AdamKasp we need a rebase and Behat scenario fix there ;)
@lchrusciel maybe it would be a better choice, but definitely it's not a case for such a simple PR 🚀

AdamKasp:
fixed in PR #10999

PR #10710, opened 2019-09-26
[Address book] Extensibility improvements

cyrosy:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

I introduced 2 changes in the AddressBook JS handler, allowing address form to be extended with more freedom :
- checkbox and radio values won't be set to empty string, and will be correctly checked
- prevent input field name collisions : the value of one field could be duplicated in multiple fields with similar names (for instance, the value of « street » would override the value of a « street2 » or « streetComp » field)

This should not break anything in core Sylius. If you think I should rebase this on another branch, just ask as this will be quite easy to adapt :)


pamil (approved):

pamil:
Thanks, Cyro! :1st_place_medal:

PR #10721, opened 2019-09-30
Fixed: use_authorize option comparison (fixes #)

### LINKED ISSUE #10720: [Critial Bug] Strict use_authorize option value check lead to Payum fail

**Description**  
You have next recommendation at `docs/cookbook/payments/authorize.rst`:
```php
class MyGatewayGatewayConfigurationType extends AbstractType
// ...
                ->add('use_authorize', HiddenType::class, [
                    'data' => 1,
                ])
```

This form will store field at `sylius_gateway_config.config` database field as `"use_authorize": 1`. 

But at [this line](https://github.com/Sylius/Sylius/commit/6c748c9aec878687c610bd440aac9635143df0c3#diff-063b340e70ed54a7454a9c76bd3ef84eR158) **check become strict** and we expecting for `true` (strictly `boolean`) rather than `1`. 

Since that point, **Payum no longer works** and Sylius redirects user to `capture` action rather than `authorize`.

**Steps to reproduce**  
- Add a payment method at admin that have `use_authorize` field at its form type
- Check at database that field stored as integer rather than boolean (`select * from sylius_gateway_config;`)

**More context**
This issue was faced at https://github.com/Setono/SyliusQuickpayPlugin. After upgrading from Sylius 1.5 to 1.6 payment stopped work and throw exception because it become redirecting to capture payment before authorize (e.g. before even user can type his payment card information). 

Currently, we can make payment works by changing `use_authorize` value to `true` at fixtures. But if we put something like `'data' => true,` at `...GatewayGatewayConfigurationType` - we still have an issue because it store value at database like this: `"use_authorize":"1"` (e.g. `true` stored as string).

**Possible Solution**  
1. **Less strict comparison** at https://github.com/Sylius/Sylius/commit/6c748c9aec878687c610bd440aac9635143df0c3#diff-063b340e70ed54a7454a9c76bd3ef84eR158
2. As far as proposed `HiddenType` example doesn't support booleans and `'data' => true,` will not help, probably we should **expect `integer` (`0`/`1`) from `use_authorize`**?
3. Or **provide valid MyGatewayGatewayConfigurationType example** how to define `use_authorize` form field that stored as `boolean` in database

igormukhingmailcom:
| Q               | A
| --------------- | -----
| Branch?         | 1.6 (as it present only since 1.6)
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10720, related 6c748c9aec878687c610bd440aac9635143df0c3
| License         | MIT

Decided to use `(bool)` here as it converts both `1` and `'1'` to true and `0` and `'0'` to false.

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/Controller/PayumController.php
+++ b/src/Sylius/Bundle/PayumBundle/Controller/PayumController.php
@@ -155,7 +155,7 @@ final class PayumController
         /** @var GatewayConfigInterface $gatewayConfig */
         $gatewayConfig = $paymentMethod->getGatewayConfig();
 
-        if (isset($gatewayConfig->getConfig()['use_authorize']) && $gatewayConfig->getConfig()['use_authorize'] === true) {
+        if (isset($gatewayConfig->getConfig()['use_authorize']) && true === (bool) $gatewayConfig->getConfig()['use_authorize']) {
             $token = $this->getTokenFactory()->createAuthorizeToken(
                 $gatewayConfig->getGatewayName(),
                 $payment,


lchrusciel:
Thank you, Igor! :1st_place_medal:

igormukhingmailcom:
Hi, @lchrusciel 
Let me know please whether this PR will be also merged to 1.6 branch (as this is bugfix) and released as `1.6.*`?
I see that currently it merged to master only.
Thanks

pamil:
@igormukhingmailcom I've cherrypicked your commit to 1.6, will be released with the today's patch tag. :)

PR #10733, opened 2019-10-02
Fix 10719 infinite order fixture loading

### LINKED ISSUE #10719: [Bug] Infinite order fixture load when additional channel(-s) added

**Description**  
When you add additional channel to `default` suite in your app, `order` fixture could get one of that new channels randomly at `channel` option via option resolver. And as far as all `default` suite products have only `FASHION_WEB` channel assigned to them, next piece of code makes loop infinite:

```php
            if (!$product->hasChannel($order->getChannel())) {
                $i--;
                continue;
            }
```

**Steps to reproduce**  
- Just add one or more (to increase possibility of random "wrong" channel) channel fixture
- And optionally define another `order` fixture like this (to get 100% possibility of failing):

```yaml
sylius_fixtures:
    suites:
        default:
            fixtures:
                order:
                    options:
                        amount: 20
                        channel: YOUR_NEW_CHANNEL_CODE_YOU_HAVENT_PRODUCTS_AT
```

**Possible Solution**  
1. Use `$productRepository->findByChannel($order->getChannel())` rather than `->findAll()` 
2. AND Throw an exception if `$products` is empty (or just skip? or skip and warn there are no products in channel X?)
3. AND Explicitly define `channel` option for `order` fixture at `default` suite like this:

```yaml
# src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures/custom_fixtures.yaml
sylius_fixtures:
    suites:
        default:
            fixtures:
                order:
                    options:
                        amount: 20
                        channel: FASHION_WEB
```

4. Also, to prevent future problems on setups that have too much products, please use some limiting when getting products (almost same issue like described at https://github.com/Sylius/Sylius/issues/10493)

I will provide PR if you guys ok with this changes.

igormukhingmailcom:
| Q               | A
| --------------- | -----
| Branch?         | 1.6 (as it present only at 1.6 and not present at 1.5 or 1.4)
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10719 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Factory/OrderExampleFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Factory/OrderExampleFactory.php
@@ -183,19 +183,20 @@ class OrderExampleFactory extends AbstractExampleFactory implements ExampleFacto
     protected function generateItems(OrderInterface $order): void
     {
         $numberOfItems = random_int(1, 5);
-        $products = $this->productRepository->findAll();
+        $channel = $order->getChannel();
+        $products = $this->productRepository->findLatestByChannel($channel, $order->getLocaleCode(), 100);
+        if (0 === count($products)) {
+            throw new \InvalidArgumentException(sprintf(
+                'You have no enabled products at the channel "%s", but they are required to create an orders for that channel',
+                $channel->getCode()
+            ));
+        }
+
         $generatedItems = [];
 
         for ($i = 0; $i < $numberOfItems; ++$i) {
             /** @var ProductInterface $product */
             $product = $this->faker->randomElement($products);
-
-            if (!$product->hasChannel($order->getChannel())) {
-                --$i;
-
-                continue;
-            }
-
             $variant = $this->faker->randomElement($product->getVariants()->toArray());
 
             if (array_key_exists($variant->getCode(), $generatedItems)) {
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures/custom_fixtures.yaml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/fixtures/custom_fixtures.yaml
@@ -16,6 +16,7 @@ sylius_fixtures:
                 order:
                     options:
                         amount: 20
+                        channel: 'FASHION_WEB'
 
                 address:
                     options:


ostrolucky:
looks like all changes except custom_fixtures can be backported to < 1.6

src/Sylius/Bundle/CoreBundle/Fixture/Factory/OrderExampleFactory.php
```diff
@@ -183,18 +183,20 @@ protected function createOrder(ChannelInterface $channel, CustomerInterface $cus
     protected function generateItems(OrderInterface $order): void
     {
         $numberOfItems = random_int(1, 5);
-        $products = $this->productRepository->findAll();
+        $channel = $order->getChannel();
+        $products = $this->productRepository->findByChannel($channel, 100);
```
pamil: Instead of adding `findByChannel` method, we can use `findBy(['channel' => $channel], [], 100)` here.

pamil: Therefore we won't need that method and we can push it as a bugfix :)

ostrolucky: There is no `channel` prop though, it's `channels`. We need `MEMBER OF`

pamil: You're right, we either need to find some other way around it using the existing methods or this PR needs to go to the master branch.

igormukhingmailcom: How about:

```
        $channel = $order->getChannel();
        $locale = $channel->getDefaultLocale();
        $products = $this->productRepository->findLatestByChannel($channel, $locale->getCode(), 100);
        if (0 === count($products)) {
            throw new \InvalidArgumentException(sprintf(
                'You have no enabled products at the channel "%s", but they are required to create an orders for that channel',
                $channel->getCode()
            ));
        }
```
(here we stick on enabled products only, but that's ok I guess)

igormukhingmailcom: Force-pushed this change.

src/Sylius/Component/Core/Repository/ProductRepositoryInterface.php
```diff
@@ -39,6 +39,11 @@ public function createShopListQueryBuilder(
      */
     public function findLatestByChannel(ChannelInterface $channel, string $locale, int $count): array;
 
+    /**
+     * @return array|ProductInterface[]
+     */
+    public function findByChannel(ChannelInterface $channel, int $count): array;
```
lchrusciel: Not needed anymore I suppose.

igormukhingmailcom: Ah, sure. Thanks.

Will rebase later today

ostrolucky:
> 1.6 (as it present only at 1.6 and not present at 1.5 or 1.4)

This is false though, it's on 1.4 too

igormukhingmailcom:
@ostrolucky 
> This is false though, it's on 1.4 too
At 1.4 there is another file to fix - `OrderFixture` rather than `OrderExampleFactory`. I guess you can make PR for it if you want.

src/Sylius/Component/Core/Repository/ProductRepositoryInterface.php
igormukhingmailcom: @lchrusciel, it rebased now

ostrolucky:
true

lchrusciel:
Thanks, Igor! :1st_place_medal:

PR #10738, opened 2019-10-04
Issue 10463 [UI] Switch from Semantic-UI to Fomantic-UI

### LINKED ISSUE #10463: Switch from Semantic-UI to Fomantic-UI?
Switch from Semantic-UI to Fomantic-UI. Or at least study our options.

Fomantic-UI **should** be a drop-in replacement with maybe some needed modifications in templates if they changed how some components are rendered. The change log can be used for that, starting from 2.4 (which is the first release after the fork, I think):
https://fomantic-ui.com/introduction/new.html#/twofour

I tried to `yarn add fomantic-ui-css` then use it in my custom gulp file. What happens? You get a huge overhead since Sylius loads Semantic-UI and I load Fomantic-UI in my custom assets. That's just pretty much doubling the size of the assets because reasons.
To avoid that we can directly override the main admin layout stylesheets and javascripts blocks and get rid of the default style.css and app.js files to use custom ones. So we should had to recreate the AdminBundle gulp file inside our custom gulp file. That means we would need to see if Sylius doesn't change any of these files, otherwise we'd need to manually put those changes in our files. A complete pain for maintainability.

**Additional context**
Semantic-UI development is simply dead. Its owner said:

> SUI is used in production for many large apps, and will continue to receive patch/release updates about every month or so at a pace that is reasonable.

That was on 03/19/2018. The sad truth is, last commit in Semantic-UI was on 10/21/2018 (the 2.4.1 release). Which means 8 months with zero pull requests being merged (or even reviewed), bugs not fixed, outdated dependencies (FontAwesome for example) and issues piling up just because having one person as the only able to keep development going in such a big project is simply unfeasible. He's looking for some kind of financing which is completely respectable. But considering how hard that is for Open Source projects, in my honest opinion he should have given some pr reviewing/merging permissions at least to a small group of trustworthy people to help him in the meantime, which he hasn't, unfortunately for us.

Anyway, my opinion about that shouldn't matter. The thing is: in Fomantic-UI we have a better and more active (last commit was three days ago) alternative to Semantic-UI which should serve as a drop-in replacement, so it should not be a big drama to switch over.

migmolrod:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10463 
| License         | MIT

This is my first attempt. Since it's the first time I contribute to such a big project, I've probably made mistakes. Let me know how can I fix them. I probably need to add some stuff, like `CHANGELOG` and `UPGRADE` files, since this patch will need users to run `yarn install` and `yarn build`. Am I right?


pierre-H:
Any news ? 😃

kulczy:
nope, we have no plans for that

PR #10746, opened 2019-10-08
Added items option to order fixture

JakobTolkemit:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no 
| License         | MIT

I thought it would be nice if you could decide what products a fixture generated order contains.

```yaml
...
  fixtures:
    order:
      options:
        items:
          - variant: 'some_variant_code'
            quantity: 5
```

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Factory/OrderExampleFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Factory/OrderExampleFactory.php
@@ -60,6 +60,9 @@ class OrderExampleFactory extends AbstractExampleFactory implements ExampleFacto
     protected $productRepository;
 
     /** @var RepositoryInterface */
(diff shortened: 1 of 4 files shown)


src/Sylius/Bundle/CoreBundle/Fixture/Factory/OrderExampleFactory.php
```diff
@@ -182,6 +214,8 @@ protected function createOrder(ChannelInterface $channel, CustomerInterface $cus
 
     protected function generateItems(OrderInterface $order): void
     {
+        $order->clearItems();
```
lchrusciel: Is it needed, if `items` are expected to be empty to access this function?

PR #10749, opened 2019-10-08
Promotion fixture coupons

JakobTolkemit:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no 
| License         | MIT

This adds the possibility to define coupons in promotion fixtures.

```yaml
...
 promotion:
   options:
     custom:
       test:
         name: 'Test'
         coupon_based: true
         coupons:
           - code: 'abc'

           - code: 'abcde'
             expires_at: 'tomorrow'
             usage_limit: 10
             per_customer_usage_limit: 1
             reusable_from_cancelled_orders: true
```

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Factory/PromotionExampleFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Factory/PromotionExampleFactory.php
@@ -67,7 +67,7 @@ class PromotionExampleFactory extends AbstractExampleFactory implements ExampleF
         $this->configureOptions($this->optionsResolver);
 
         if ($this->couponFactory === null) {
-            @trigger_error(sprintf('Not passing a $couponFactory to %s constructor is deprecated since Sylius 1.6 and will be removed in Sylius 2.0.', self::class), \E_USER_DEPRECATED);
+            @trigger_error(sprintf('Not passing a $couponFactory to %s constructor is deprecated since Sylius 1.8 and will be removed in Sylius 2.0.', self::class), \E_USER_DEPRECATED);
         }
     }
 
@@ -164,12 +164,8 @@ class PromotionExampleFactory extends AbstractExampleFactory implements ExampleF
         ;
     }
 
-    /**
-     * @param FactoryInterface|null $couponFactory
-     *
-     * @return \Closure
-     */
-    private static function getCouponNormalizer (?FactoryInterface $couponFactory): \Closure {
+    private static function getCouponNormalizer (?FactoryInterface $couponFactory): \Closure
+    {
         return function (Options $options, array $couponDefinitions) use ($couponFactory): array {
             if (null === $couponFactory) {
                 throw new \RuntimeException('You must configure a $couponFactory');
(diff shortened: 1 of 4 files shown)


src/Sylius/Bundle/CoreBundle/Fixture/Factory/PromotionExampleFactory.php
```diff
@@ -45,11 +50,13 @@ class PromotionExampleFactory extends AbstractExampleFactory implements ExampleF
 
     public function __construct(
         FactoryInterface $promotionFactory,
+        FactoryInterface $couponFactory,
```
Zales0123: Probably, we should inject it as the last nullable parameter, to prevent from BC break. Check out [this approach](https://github.com/Sylius/Sylius/blob/b157a18a06fae0a6eafba73b2ed9489a0fc3df21/src/Sylius/Bundle/CoreBundle/Fixture/Factory/AdminUserExampleFactory.php#L48).

src/Sylius/Bundle/CoreBundle/Fixture/Factory/PromotionExampleFactory.php
```diff
@@ -142,6 +153,32 @@ protected function configureOptions(OptionsResolver $resolver): void
...
+                    $coupon->setUsageLimit($couponDefinition['usage_limit']);
+
+                    if (null !== $couponDefinition['expires_at']) {
+                        $coupon->setExpiresAt(new \DateTime($couponDefinition['expires_at']));
+                    }
+
+                    $coupons[] = $coupon;
+                }
+
+                return $coupons;
```
Zales0123: We could also extract it to the private/protected method for better readability 🖖

src/Sylius/Bundle/CoreBundle/Fixture/Factory/PromotionExampleFactory.php
```diff
@@ -142,6 +153,41 @@ protected function configureOptions(OptionsResolver $resolver): void
...
+            ->setNormalizer('coupons', self::getCouponNormalizer($this->couponFactory))
         ;
     }
+
+    /**
+     * @param FactoryInterface $couponFactory
+     *
+     * @return \Closure
+     */
+    private static function getCouponNormalizer (FactoryInterface $couponFactory): \Closure {
```
lchrusciel:
```suggestion
    private static function getCouponNormalizer (?FactoryInterface $couponFactory): \Closure {
```

And we need to provide an exception if this function is used and `$this->couponFactory` is null

src/Sylius/Bundle/CoreBundle/Fixture/Factory/PromotionExampleFactory.php
```diff
@@ -142,6 +153,41 @@ protected function configureOptions(OptionsResolver $resolver): void
...
+            ->setAllowedTypes('coupons', 'array')
+            ->setNormalizer('coupons', self::getCouponNormalizer($this->couponFactory))
         ;
     }
+
+    /**
+     * @param FactoryInterface $couponFactory
+     *
+     * @return \Closure
+     */
```
Zales0123: Dockblock not needed

lchrusciel:
Hey @JakobTolkemit, feature freeze will be by the end of the next week, so it is the last moment to add it to the core. Can you apply our small recommendations?

lchrusciel:
Thanks, Jakob! :tada:

PR #10759, opened 2019-10-11
Fix user impersonated listener when impersonating another user resource

loic425:
| Q               | A
| --------------- | -----
| Branch?         | 1.5
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

If you want to impersonate another user resource (for example, in my current project, we have a Partner user), this listener is called and is failed cause it's called on any user resource impersonating action.


pamil:
Thank you, Loïc! :tada:

PR #10776, opened 2019-10-17
[Behat] Remove final on product index and product variant index pages

loic425:
| Q               | A
| --------------- | -----
| Branch?         | 1.5
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

All Behat pages are not final except these two pages. Behat pages need to be extended easily with override method.


pamil:
Thank you, Loïc! :1st_place_medal:

PR #10780, opened 2019-10-18
Allow for no default tax zone in channel fixtures

pamil:


vvasiloi:
Just wondering, should it accept zones with "shipping" scope?

pamil:
Replaced by #10781.

@vvasiloi good point, could you create an issue for that?

PR #10781, opened 2019-10-18
Allow no default tax zone in channel fixtures

pamil:


lchrusciel:

PR #10786, opened 2019-10-20
Change rename in tests/Application/Kernel.php

maximehuran:
Change the replaced string in `tests/Application/Kernel.php` because `\Acme\SyliusExamplePlugin\AcmeSyliusExamplePlugin()` does not exists


Zales0123:
@maximehuran could you please make a change proposed by @pamil? It would be great to have it merged 🖖

maximehuran:
I have also to update `tests/Context/Ui/Shop/WelcomeContext.php` to `tests/Behat/Context/Ui/Shop/WelcomeContext.php`

PR #10802, opened 2019-10-29
[Addressing] Make sure the CountryNameExtension::translateCountryIsoCode() always returns a string

### LINKED ISSUE #10799: Support case when CountryNameExtension::translateCountryIsoCode() doesn't have translation
**Sylius version affected**: 1.6.1

**Description**  
`Sylius\Bundle\AddressingBundle\Twig\CountryNameExtension::translateCountryIsoCode()` return type hint expects non-null result, while `Intl::getRegionBundle()->getCountryName()` might return null if no translation provided. 
Some of the edge case ISO countries, like reserved names, might not have translation. (UN, TA, XK, EZ, DG, EA, IC, AC)

**Steps to reproduce**  
Add one of countries (UN, TA, XK, EZ, 
DG, EA, IC, AC) to country list and there will be 500 error on country listing at admin UI.

**Possible Solution**  
If no translation found, return raw country code.

image

While I should just drop those countries (It worked at Sylius1.2 and SF3.4, btw), it will be nice to handle this case generally.

vvasiloi:
Fallback to code when there's no translation.
Fallback to empty string when there's no code.

| Q               | A
| --------------- | -----
| Branch?         | 1.5
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10799
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Twig/CountryNameExtension.php
+++ b/src/Sylius/Bundle/AddressingBundle/Twig/CountryNameExtension.php
@@ -28,12 +28,20 @@ class CountryNameExtension extends \Twig_Extension
         ];
     }
 
+    /**
+     * @param CountryInterface|string $country
+     * @param string|null $locale
+     *
+     * @return string
+     */
     public function translateCountryIsoCode($country, ?string $locale = null): string
     {
-        if ($country instanceof CountryInterface) {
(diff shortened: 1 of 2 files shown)


vvasiloi:
Will open from a dedicated branch.

PR #10803, opened 2019-10-29
Add Webpack Encore

kulczy:
This is the first step to completely remove the gulp. For now, lets you choose whether you want to use the gulp or webpack. By default, it uses the gulp as before.

#### To switch to the webpack, just change the assets paths:

```js
// src/Sylius/Bundle/AdminBundle/Resources/views/_scripts.html.twig

{{ encore_entry_script_tags('admin-entry', null, 'admin') }}
```

```js
// src/Sylius/Bundle/AdminBundle/Resources/views/_styles.html.twig

{{ encore_entry_link_tags('admin-entry', null, 'admin') }}
```

```js
// src/Sylius/Bundle/AdminBundle/Resources/views/_logo.html.twig

{{ asset('build/admin/images/admin-logo.svg', 'admin') }}
```

And for shop:

```js
// src/Sylius/Bundle/ShopBundle/Resources/views/_scripts.html.twig

{{ encore_entry_script_tags('shop-entry', null, 'shop') }}
```

```js
// src/Sylius/Bundle/ShopBundle/Resources/views/_styles.html.twig

{{ encore_entry_link_tags('shop-entry', null, 'shop') }}
```

```js
// src/Sylius/Bundle/ShopBundle/Resources/views/_header.html.twig

{{ asset('build/shop/images/logo.png', 'shop') }}
```

The paths should be changed for each asset you use.

#### To build the assets, run:
`yarn encore dev` or `yarn encore production` or `yarn encore dev-server`

When compiling assets, errors may appear (they don't break the build), due to different babel configuration for gulp and webpack. Once you decide to use the webpack you can delete the `gulpfile.babel.js` and `.babelrc` from the root directory - then the errors will stop appearing.

---

➡️➡️➡️  [Webpack Encore docs](https://symfony.com/doc/current/frontend.html#webpack-encore)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /public/css
 /public/js
 /public/media
+/public/build
 !/public/media/image/.gitkeep
 
 /app/config/parameters.yml
@@ -40,3 +41,8 @@
 ###> phpunit/phpunit ###
 /phpunit.xml
 ###< phpunit/phpunit ###
+
+###> symfony/webpack-encore-bundle ###
+/npm-debug.log
+/yarn-error.log
+###< symfony/webpack-encore-bundle ###
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/layout.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/layout.html.twig
@@ -3,7 +3,7 @@
 {% block title %} | Sylius{% endblock %}
 
 {% block stylesheets %}
-    {% include '@SyliusUi/_stylesheets.html.twig' with {'path': 'assets/admin/css/style.css'} %}
+    {% include '@SyliusAdmin/_styles.html.twig' %}
 
     {{ sonata_block_render_event('sylius.admin.layout.stylesheets') }}
 {% endblock %}
@@ -49,7 +49,7 @@
 {% endblock %}
 
 {% block javascripts %}
-    {% include '@SyliusUi/_javascripts.html.twig' with {'path': 'assets/admin/js/app.js'} %}
+    {% include '@SyliusAdmin/_scripts.html.twig' %}
 
     {{ sonata_block_render_event('sylius.admin.layout.javascripts') }}
 {% endblock %}
--- a/src/Sylius/Bundle/ShopBundle/Resources/private/entry.js
+++ b/src/Sylius/Bundle/ShopBundle/Resources/private/entry.js
@@ -1,0 +1,15 @@
+import './js/shim/shim-jquery';
+import './js/shim/shim-lightbox';
+import './js/shim/shim-semantic-ui';
+
+import 'semantic-ui-css/semantic.css';
+import 'lightbox2/dist/css/lightbox.min.css';
+
+import 'sylius/ui-resources/js/app';
+import './js/app';
+
+import 'sylius/ui-resources/sass/main.scss';
+import './scss/style.scss';
+
+import 'sylius/ui-resources/img/logo.png';
+import 'sylius/ui-resources/img/avatar.png';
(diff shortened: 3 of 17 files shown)


gabiudrescu:
1. Reproducible builds with Webpack is nightmare; further reads here:
 - https://github.com/webpack/webpack/issues/7179
 - https://medium.com/webpack/predictable-long-term-caching-with-webpack-d3eee1d3fa31

2. 

> Once you decide to use the webpack you can delete the gulpfile.babel.js and .babelrc from the root directory - then the errors will stop appearing.

Could be automated with symfony-flex, I think. but I don't know if it should

3. can we challenge now the usage of yarn instead of npm? why not use npm instead of yarn?

src/Sylius/Bundle/AdminBundle/Resources/views/layout.html.twig
```diff
@@ -3,7 +3,7 @@
 {% block title %} | Sylius{% endblock %}
 
 {% block stylesheets %}
-    {% include '@SyliusUi/_stylesheets.html.twig' with {'path': 'assets/admin/css/style.css'} %}
```
pamil: I could see someone modifying `@SyliusUi/_stylesheets.html.twig` with an intention to include some asset in all the UI. Same applies to `@SyliusUi/_javascripts.html.twig`. If we don't include that file, it would effectively be unused and it would have no effect in the next minor Sylius release.

I can see two solutions:
 - Revert this change (should still work since the path is same)
 - Use the removed code in `@SyliusAdmin/_styles.html.twig` (so that `layout.html.twig` uses `_styles.html.twig` which uses `_stylesheets.html.twig`)

No matter what we decide to go with, it should be done for styles and scripts for both Admin and Shop in the same way.

.gitignore
```diff
@@ -40,3 +41,8 @@
 ###> phpunit/phpunit ###
 /phpunit.xml
 ###< phpunit/phpunit ###
+
+###> symfony/webpack-encore-bundle ###
+npm-debug.log
```
pamil:
```suggestion
/npm-debug.log
```

.gitignore
```diff
@@ -40,3 +41,8 @@
...
+yarn-error.log
```
pamil:
```suggestion
/yarn-error.log
```

src/Sylius/Bundle/ShopBundle/Resources/private/entry.js
```diff
@@ -0,0 +1,15 @@
+import './js/shim/shim-jquery';
+import './js/shim/shim-lightbox';
+import './js/shim/shim-semantic-ui';
+
+import 'semantic-ui-css/semantic.css';
+import 'lightbox2/dist/css/lightbox.min.css';
+
+import '../../../UiBundle/Resources/private/js/app';
```
pamil: Could we pass `uiBundlePath` variable from `webpack.config.js` here? Going outside of the current bundle seems a bit like crossing some boundary.

I don't hold this opinion strongly, but I want to discuss it first :)

src/Sylius/Bundle/AdminBundle/Resources/views/layout.html.twig
kulczy: good point, the second solution seems better to me

src/Sylius/Bundle/ShopBundle/Resources/private/entry.js
kulczy: in the `webpack.config.js` file we have the alias `sylius/ui` which is necessary for the `app.js` files. We can create another alias, e.g. `sylius/ui_resources` and then we'll be able to use it instead of `../../../UiBundle/Resources/private`

pamil: Sounds good! This way we'll get the absolute path to the resources. What is the naming convention here, should it be `sylius/ui_resources` or `sylius/ui-resources`?

kulczy: imo `sylius/ui-resources`

pamil:
Thank you, Szymon! :tada:

PR #10805, opened 2019-10-29
[Addressing] Make sure the CountryNameExtension::translateCountryIsoCode() always returns a string

### LINKED ISSUE #10799: Support case when CountryNameExtension::translateCountryIsoCode() doesn't have translation

vvasiloi:
Fallback to code when there's no translation.
Fallback to empty string when there's no code.

| Q               | A
| --------------- | -----
| Branch?         | 1.5
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10799
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Twig/CountryNameExtension.php
+++ b/src/Sylius/Bundle/AddressingBundle/Twig/CountryNameExtension.php
@@ -28,12 +28,6 @@ class CountryNameExtension extends \Twig_Extension
         ];
     }
 
-    /**
-     * @param CountryInterface|string $country
-     * @param string|null $locale
-     *
-     * @return string
-     */
     public function translateCountryIsoCode($country, ?string $locale = null): string
     {
         $countryCode = $country instanceof CountryInterface ? $country->getCode() : $country;
--- a/src/Sylius/Bundle/AddressingBundle/Twig/CountryNameExtension.php
+++ b/src/Sylius/Bundle/AddressingBundle/Twig/CountryNameExtension.php
@@ -28,12 +28,20 @@ class CountryNameExtension extends \Twig_Extension
         ];
     }
 
+    /**
+     * @param CountryInterface|string $country
+     * @param string|null $locale
+     *
+     * @return string
+     */
     public function translateCountryIsoCode($country, ?string $locale = null): string
     {
-        if ($country instanceof CountryInterface) {
-            return Intl::getRegionBundle()->getCountryName($country->getCode(), $locale);
+        $countryCode = $country instanceof CountryInterface ? $country->getCode() : $country;
+
+        if ($countryName = Intl::getRegionBundle()->getCountryName($countryCode, $locale)) {
+            return $countryName;
         }
 
-        return Intl::getRegionBundle()->getCountryName($country, $locale);
+        return $countryCode ?? '';
     }
 }
--- a/src/Sylius/Bundle/AddressingBundle/spec/Twig/CountryNameExtensionSpec.php
+++ b/src/Sylius/Bundle/AddressingBundle/spec/Twig/CountryNameExtensionSpec.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace spec\Sylius\Bundle\AddressingBundle\Twig;
 
 use PhpSpec\ObjectBehavior;
+use Sylius\Component\Addressing\Model\Country;
 use Sylius\Component\Addressing\Model\CountryInterface;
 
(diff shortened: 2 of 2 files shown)


src/Sylius/Bundle/AddressingBundle/Twig/CountryNameExtension.php
```diff
@@ -28,12 +28,20 @@ public function getFilters(): array
         ];
     }
 
+    /**
+     * @param CountryInterface|string $country
+     * @param string|null $locale
+     *
+     * @return string
+     */
```
Zales0123: Docblock not needed

src/Sylius/Bundle/AddressingBundle/Twig/CountryNameExtension.php
```diff
@@ -28,12 +28,20 @@ public function getFilters(): array
...
-        if ($country instanceof CountryInterface) {
-            return Intl::getRegionBundle()->getCountryName($country->getCode(), $locale);
+        $countryCode = $country instanceof CountryInterface ? $country->getCode() : $country;
+
+        if ($countryName = Intl::getRegionBundle()->getCountryName($countryCode, $locale)) {
+            return $countryName;
         }
 
-        return Intl::getRegionBundle()->getCountryName($country, $locale);
+        return $countryCode ?? '';
```
Zales0123: I believe it can be done even simpler:

```
if ($country instanceof CountryInterface) {
    return Intl::getRegionBundle()->getCountryName($country->getCode(), $locale) ?? $country->getCode();
}

return Intl::getRegionBundle()->getCountryName($country, $locale) ?? $country;
```

src/Sylius/Bundle/AddressingBundle/spec/Twig/CountryNameExtensionSpec.php
```diff
@@ -39,4 +40,16 @@ function it_translates_country_code_to_name_according_to_locale(): void
     {
         $this->translateCountryIsoCode('IE', 'es')->shouldReturn('Irlanda');
     }
+
+    function it_fallbacks_to_country_code_when_there_is_no_translation(): void
+    {
+        $country = 'country_code_without_translation';
```
Zales0123: Variable is not needed

src/Sylius/Bundle/AddressingBundle/spec/Twig/CountryNameExtensionSpec.php
```diff
@@ -39,4 +40,16 @@ function it_translates_country_code_to_name_according_to_locale(): void
...
+        $this->translateCountryIsoCode($country)->shouldReturn($country);
+    }
+
+    function it_fallbacks_to_an_empty_string_when_there_is_no_code(): void
+    {
+        $this->translateCountryIsoCode(new Country())->shouldReturn('');
```
Zales0123: This should be a `CountryInterface` mock with `$country->getCode()->willReturn('country_code_without_translation')` result stubbed

src/Sylius/Bundle/AddressingBundle/Twig/CountryNameExtension.php
vvasiloi: This way the `if` block can return `null` (from `$country->getCode()`).
Also, if the `$country` argument is `null`, for example when chaining the filter on a `null` value in the teplate, the returned value will be `null` again.

src/Sylius/Bundle/AddressingBundle/spec/Twig/CountryNameExtensionSpec.php
vvasiloi: You mean `$country->getCode()->willReturn(null)`?

Zales0123:
Thank you, Victor! :tada:

Issue #10387, opened 2019-05-15
Incorrect amount of taxons loaded in shop list query, causes missing taxons


PalTamasWBA:
**Sylius version affected**: 1.2.16

**Description**  

`\Sylius\Bundle\CoreBundle\Doctrine\ORM\ProductRepository::createShopListQueryBuilder` does not eager load all taxons, makes $product->getTaxons() return incorrect amount of taxons (only the one that we're filtering by)

EDIT: I've checked master, and see that there has been some improvements to include descendent taxons and sibling taxons, however this still does not fix the issue to include all taxons of a product.

In 1.2.16 the following lines are present
```
->addSelect('productTaxon')
```
I'm assuming this was done in order to eager load the product taxons, but this causes an issue  because `->andWhere('productTaxon.taxon = :taxon')` filters on the taxon, and only that particular taxon is eager loaded.

I'm not sure which exact version has introduced this change, i can see that in v1.2.3 it's not present.

see https://github.com/Sylius/Sylius/blob/v1.2.3/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php vs https://github.com/Sylius/Sylius/blob/v1.2.16/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php

**Steps to reproduce**  

1. Create Product A
2. assign it 2 taxons, B and C
3. pass the `createShopListQueryBuilder` method taxon B
4. call `$queryBuilder->getResults()`, find product A, and call `->getTaxons()->count()` on it
5. observe that count is 1, but should be 2
6. remove `->distinct()->addSelect('productTaxon')` from the `createShopListQueryBuilder` method
7. observe that count is now 2

**Possible Solution**  
Remove `->distinct()->addSelect('productTaxon')` from the `createShopListQueryBuilder` method or override the ProductRepository in your project and override the `createShopListQueryBuilder`

rowanvanekeren:
I got the same problem, the data consistency is not great between product listings and product details. 
I can confirm that when a product is in a list the taxons get limited and multiple root taxons are not visible this limits the power of the taxon system a lot (imo). I had a dirty workaround by creating the same one to many relation twice but as a different property (like: allTaxons) and set  fetch: EAGER to get all the product taxons in overviews aswell.

vvasiloi:
I explored this issue and summarized the results in this [repository](https://github.com/vvasiloi/symfony-test)

## This particular issue
[...]

Issue #10343, opened 2019-04-29
Product pagination bug (again)


vvasiloi:
**Sylius version affected**: 1.4.1 and up (didn't test lower versions)
**Description**  
The same bug described here: https://github.com/Sylius/Sylius/issues/9699
Opened this issue to keep record.

**Steps to reproduce**  
It's not persistent across all categories.
Here is one where the bug occurs: https://demo.sylius.com/en_US/taxons/t-shirts?limit=9

**Possible Solution**  
https://github.com/Sylius/Sylius/issues/9699#issuecomment-487631905

mmenozzi:
Another quick solution to this problem is to change the file `config/packages/_sylius.yaml` from:
```yaml
sylius_shop:
    product_grid:
        include_all_descendants: true
```
to:
```yaml
sylius_shop:
    product_grid:
        include_all_descendants: false
```
Of course this will stop to include all descendants product but if you have this bug probably is because products are associated to both a parent and a child category. Indeed on the Sylius demo, T-Shirts are associated both to the `T-Shirt` and `T-Shirt/Geneder` categories.

So I don't know if this is really a bug in such situations.

vvasiloi:
It is a bug.
Read https://github.com/Sylius/Sylius/issues/10387#issuecomment-504197089 for details.
I started to prepare a PR, but... didn't find time to finish it.

mmenozzi:
Ok, good to know!

pcmsystemsltd:
We can see this issue still happening on versions 1.6 and 1.7. Pagination is working fine if the list of products is sorted alphabetically but not by position or price for example.

vvasiloi:
Do not stale.

PR #10815, opened 2019-11-02
Incorrect product count in the grid

dbalabka:
| Q               | A
| --------------- | -----
| Branch?         | >=1.3
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Grid size is incorrect when one other the product from the page is included in multiple taxons.
The regular usage scenario is when there are two taxons some product category and product brand. In this example, the product will be included in two taxons in the same time. 

The SQL that contains `distinct` works incorrectly when `sylius_product_taxon.taxon_id` is included into `SELECT`. So the solution is simple. Need to remove the following line:
https://github.com/Sylius/Sylius/blob/af11ba747102a86400e5537dc16bcb180e2b70e2/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php#L76

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
@@ -73,7 +73,6 @@ class ProductRepository extends BaseProductRepository implements ProductReposito
         $queryBuilder = $this->createQueryBuilder('o')
             ->distinct()
             ->addSelect('translation')
-            ->addSelect('productTaxon')
             ->innerJoin('o.translations', 'translation', 'WITH', 'translation.locale = :locale')
             ->innerJoin('o.productTaxons', 'productTaxon');
 


dbalabka:
Probably, a better fix is to use "WHERE IN" strategy:
https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/tutorials/pagination.html

vvasiloi:
It will break the sorting by position, see https://github.com/Sylius/Sylius/pull/10136 and https://github.com/Sylius/Sylius/pull/10114
Other related issues: https://github.com/Sylius/Sylius/issues/10387 and https://github.com/Sylius/Sylius/issues/10343
Possible solution: https://github.com/Sylius/Sylius/issues/10387#issuecomment-504197089

dbalabka:
@vvasiloi build is green. I guess the regression tests were missing. It would be great to cover this.

Talking about [possible solution in #10387](https://github.com/Sylius/Sylius/issues/10387#issuecomment-504197089). As I understand it might affect performance #3775

vvasiloi:
@dbalabka Yes, it is the very reason the output walkers were disabled.
In https://github.com/doctrine/orm/issues/3648 it's mentioned that using output walkers is "slow" and https://github.com/Sylius/Sylius/pull/3495 mentions a load time of 60s for a million rows table.
If it's reasonable for tables under 200k rows, then I propose to use the defaults (`$fetchJoinCollection = true, $useOutputWalkers = null`), allow it to be configurable through the grid configuration and set both to `true` at least for the shop product grid.

tom10271:
I face the same issue today on `resource-bundle/src/Bundle/Doctrine/ORM/EntityRepository.php` `getPaginator` and `$fetchJoinCollection` has to be `true` if query passed in is joinning to other one to many table.

Consider I was trying to get a Pagerfanta for product with specific variant code pattern, with `$fetchJoinCollection` set as false. In `doctrine/orm/lib/Doctrine/ORM/Tools/Pagination/Paginator.php` `getIterator` it will execute something like

```sql
SELECT 
...
FROM product p
JOIN variant pv ON ...
WHERE
...
LIMIT 100
```

Since 1 product can have more than 1 variants, number of distinct products returned is likely too be less than 100.

With `$fetchJoinCollection` set to true, doctrine will first select all distinct IDs first and then select all tuples afterwards to ensure number of products returned matches what `setMaxResult` set.

PR #10837, opened 2019-11-08
Remove unused templating engine from RemoveAvatarAction

pamil:


src/Sylius/Bundle/AdminBundle/Action/RemoveAvatarAction.php
```diff
@@ -40,12 +36,10 @@ final class RemoveAvatarAction
 
     public function __construct(
         AvatarImageRepositoryInterface $avatarRepository,
-        EngineInterface $templatingEngine,
```
lchrusciel: Isn't it a bc break? :(

pamil: Good catch 👍 Kind of, clarifying it here: https://github.com/Sylius/Sylius/pull/10844

PR #10851, opened 2019-11-12
[DX] Adding isLocaleAvailable to LocaleProviderInterface

mamazu:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | :warning: yes
| Deprecations?   | no
| Related tickets | none
| License         | MIT

Checking if the locale is allowed (or available) in the current context should be inside the locale Provider as a method.

There could be a performance boost because if we are interacting with the LocaleRepository we don't need to fetch all and iterate through them but now can directly check if it exists. But this performance increase will probably be negligible.


src/Sylius/Component/Core/Provider/ChannelBasedLocaleProvider.php
```diff
@@ -54,6 +54,24 @@ public function getAvailableLocalesCodes(): array
...
 
+    public function isLocaleCodeAvailable(string $locale): bool
+    {
+        try {
+            /** @var ChannelInterface $channel */
+            $channel = $this->channelContext->getChannel();
+
+            return $channel
+                ->getLocales()
+                ->map(function (LocaleInterface $locale) {
```
vvasiloi: You could use `->exists()` instead of `->map()`, like:
```
$channel
    ->getLocales()
    ->exists(static function (LocaleInterface $locale) use ($localeCode) {
        return $locale->getCode() === $localeCode;
    })
```

mamazu: Nice, haven't thought about that.

mamazu:
I have no idea why this is broken.

vvasiloi:
@mamazu try to rebase/merge.

mamazu:
Ah okay, thanks

mamazu:
@Zales0123 could you restart the job and review it please.

src/Sylius/Component/Locale/Provider/LocaleProviderInterface.php
```diff
@@ -20,5 +20,7 @@ interface LocaleProviderInterface
      */
     public function getAvailableLocalesCodes(): array;
 
+    public function isLocaleCodeAvailable(string $locale): bool;
```
lchrusciel: 🚨 BC break 🚨

We need to create a new interface that may be implemented by the same class that we already have.

mamazu: Yeah, thanks for spotting it. I will add it to the description.

Zales0123: ping @mamazu :)

mamazu: Well, the problem is that there is no good backwards compatibility fix. So I guess this will not be merged?

vvasiloi: Maybe add a new abstraction instead?

mamazu: Yeah, that's probably the best idea. I will do that.

PR #10858, opened 2019-11-14
[Psalm] Treat MismatchingDocblockReturnType as errors

pamil:
Getting some proper docblocks in place. Lots of `@psalm-suppress` annotations, but those can be removed after fixing them upstream (and there's a Psalm CLI flag to find those unnecessary).


GSadee:

PR #10863, opened 2019-11-18
[SyliusUserBundle] Improve output of Promote/DemoteUserCommand

### LINKED ISSUE #10862: [SyliusUserBundle] Confusing output when promoting or demoting a user causes an error using the console command
**Sylius version affected**: 1.6.x

**Description**  
If you promote a user and add a role they already have or demote them to remove a role they don't have through the console, the output text gives the impression that any other role changes were successful, even though the user entity is is only flushed and updated in the database if no error happens. 

**Steps to reproduce**  
1. Add an admin user in the admin panel.
2. Use sylius:user:promote to add the roles ROLE_ADMINISTRATION_ACCESS ROLE_API_ACCESS
3. The command will output something like this:

User "user@domain.com" did already have "ROLE_ADMINISTRATION_ACCESS" security role.
Security role ROLE_API_ACCESS has been added to user user@domain.com

4. If you check the database, or in this case try and use the user with the api you will see the ROLE_API_ACCESS role is missing.

**Possible Solution**  
I should be able to submit a PR to fix this, but not sure on the best method to take, either:

1. Always flush the entity saving any successful role changes

 or

2. Save success messages into array and only output them if no error happens, in the case of an error output an additional message indicating no changes have been made to the user.

I think option 2 is the best option.

markbeazley:
| Q               | A
| --------------- | -----
| Branch?         | 1.5
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10862
| License         | MIT

A quick fix to make it clearer to users that if an error happens the user's roles are unchanged. Previously it was showing both success and error messages which gave the impression other role changes that didn't trigger an error had been applied when they hadn't (the entity is only flushed if no errors happen).

### DIFF
--- a/src/Sylius/Bundle/UserBundle/Command/DemoteUserCommand.php
+++ b/src/Sylius/Bundle/UserBundle/Command/DemoteUserCommand.php
@@ -49,21 +49,25 @@ EOT
     protected function executeRoleCommand(InputInterface $input, OutputInterface $output, UserInterface $user, array $securityRoles): void
     {
         $error = false;
+        $successMessages = [];
 
         foreach ($securityRoles as $securityRole) {
             if (!$user->hasRole($securityRole)) {
-                $output->writeln(sprintf('<error>User "%s" didn\'t have "%s" Security role.</error>', $user->getEmail(), $securityRole));
+                $output->writeln(sprintf('<error>User "%s" doesn\'t have "%s" Security role.</error>', $user->getEmail(), $securityRole));
                 $error = true;
 
                 continue;
             }
 
             $user->removeRole($securityRole);
-            $output->writeln(sprintf('Security role <comment>%s</comment> has been removed from user <comment>%s</comment>', $securityRole, $user->getEmail()));
+            $successMessages[] = sprintf('Security role <comment>%s</comment> has been removed from user <comment>%s</comment>', $securityRole, $user->getEmail());
         }
 
         if (!$error) {
+            $output->writeln($successMessages);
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/UserBundle/Command/DemoteUserCommand.php
```diff
@@ -49,21 +49,25 @@ protected function configure(): void
     protected function executeRoleCommand(InputInterface $input, OutputInterface $output, UserInterface $user, array $securityRoles): void
     {
         $error = false;
+        $successMessages = [];
 
         foreach ($securityRoles as $securityRole) {
             if (!$user->hasRole($securityRole)) {
-                $output->writeln(sprintf('<error>User "%s" didn\'t have "%s" Security role.</error>', $user->getEmail(), $securityRole));
+                $output->writeln(sprintf('<error>User "%s" doesn\'t have "%s" Security role.</error>', $user->getEmail(), $securityRole));
```
lchrusciel: I'm wondering if we shouldn't just throw an exception and return from the function

markbeazley: Possibly, the current way has the advantage of showing every error to the user in cases where they try and remove multiple roles that aren't set, by throwing an exception and returning they will only see the first incorrect role.

lchrusciel:
Thank you, @markbeazley! :tada:

lchrusciel:
Also, welcome to our community ;) 👋

PR #10865, opened 2019-11-21
[Admin][Promotion] Fix the prevention of generating too many coupons

### LINKED ISSUE #10795: [Promotion] Coupons generation check fails
**Sylius version affected**: 1.6, master

**Description** 
Before the coupons generation is some check mechanism that does not work properly. 

**Steps to reproduce** 
1. Create coupon based promotion (or edit one to coupon-based)
1. Go to the **Generate coupons** section
1. Enter **prefix**: _BRNO-_, **code length**: _1_
1. Enter the **amount** _8_ and generate it
1. Repeat the process 2 more times and you will get the **500**

**Possible Solution** 
Check the already generated coupons and give relevant notice to the user.
The check would not be so easy. It could be required to store additional data from coupon generation form to the database.

Btw the current check doesn't work properly - it shows 8 possible coupons when having code length 1, but the right result is 16 (0-9,A-E).

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.6
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10795
| License         | MIT

### DIFF
--- a/features/promotion/managing_coupons/being_unable_to_generate_too_many_coupons_with_prefix_and_suffix.feature
+++ b/features/promotion/managing_coupons/being_unable_to_generate_too_many_coupons_with_prefix_and_suffix.feature
@@ -1,0 +1,45 @@
+@managing_promotion_coupons
+Feature: Being unable to generate too many coupons with prefix and suffix
+    In order to avoid generating duplicated coupons for my promotions
+    As an Administrator
+    I want to be prevented from generating too many coupons with the same prefix or suffix
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And there is a promotion "Christmas sale"
+        And it is coupon based promotion
+        And I am logged in as an administrator
+
+    @todo @ui
+    Scenario: Being unable to generate too many coupons with prefix
+        Given I have generated 8 coupons for this promotion with code length 1 and prefix "CHRISTMAS_"
+        When I want to generate new coupons for this promotion
+        And I choose the amount of 2 coupons to be generated
+        And I specify their prefix as "CHRISTMAS_"
+        And I specify their code length as 1
+        And I try to generate these coupons
+        Then I should be notified that generating 2 coupons with code length equal to 1 is not possible
+        And there should still be 10 coupons related to this promotion
+
+    @todo @ui
+    Scenario: Being unable to generate too many coupons with suffix
+        Given I have generated 8 coupons for this promotion with code length 1 and suffix "_CHRISTMAS"
+        When I want to generate new coupons for this promotion
+        And I choose the amount of 2 coupons to be generated
+        And I specify their suffix as "_CHRISTMAS"
+        And I specify their code length as 1
+        And I try to generate these coupons
+        Then I should be notified that generating 2 coupons with code length equal to 1 is not possible
+        And there should still be 13 coupons related to this promotion
+
+    @todo @ui
+    Scenario: Being unable to generate too many coupons with prefix and suffix
+        Given I have generated 8 coupons for this promotion with code length 1, prefix "CHRISTMAS_" and suffix "_SALE"
+        When I want to generate new coupons for this promotion
+        And I choose the amount of 2 coupons to be generated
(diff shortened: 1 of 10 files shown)


src/Sylius/Behat/Context/Ui/Admin/ManagingPromotionCouponsContext.php
```diff
@@ -99,9 +99,9 @@ public function iWantToGenerateNewCouponsForThisPromotion(PromotionInterface $pr
      * @When /^I specify their code length as (\d+)$/
      * @When I do not specify their code length
      */
-    public function iSpecifyTheirCodeLengthAs($codeLength = null)
+    public function iSpecifyTheirCodeLengthAs(int $codeLength = null): void
```
pamil:
```suggestion
    public function iSpecifyTheirCodeLengthAs(?int $codeLength = null): void
```

src/Sylius/Behat/Context/Ui/Admin/ManagingPromotionCouponsContext.php
```diff
@@ -166,9 +166,9 @@ public function iChangeItsUsageLimitTo($limit)
      * @When I do not specify its amount
      * @When I choose the amount of :amount coupons to be generated
      */
-    public function iSpecifyItsAmountAs($amount = null)
+    public function iSpecifyItsAmountAs(int $amount = null): void
```
pamil:
```suggestion
    public function iSpecifyItsAmountAs(?int $amount = null): void
```

features/promotion/managing_coupons/being_unable_to_generate_too_many_coupons_with_prefix_and_suffix.feature
```diff
@@ -0,0 +1,45 @@
+@managing_promotion_coupons
+Feature: Being unable to generate too many coupons with prefix and suffix
+    In order to avoid generating duplicated coupons for my promotions
+    As an Administrator
+    I want to be prevented from generating too many coupons with the same prefix or suffix
+
```
pamil: When I first read this scenario, I didn't understand why it's not possible to generate 2 more coupons if 8 are generated already.

What do you think about adding the following section before `Background`?

```gherkin
Rules:
    - Generated coupon code uses 16 characters (numbers from 0 to 9, letters from A to F)
    - To prevent guessing coupons codes, it is not allowed to generate more than 50% of possible coupons
```

src/Sylius/Bundle/PromotionBundle/Doctrine/ORM/PromotionCouponRepository.php
```diff
@@ -45,6 +45,30 @@ public function countByCodeLength(int $codeLength): int
...
+        }
+        if ($suffix !== null) {
+            $codeLength += strlen($suffix);
+        }
+        $codeTemplate = $prefix . '%' . $suffix;
+
+        return (int) $this->createQueryBuilder('o')
+            ->select('COUNT(o.id)')
+            ->andWhere('LENGTH(o.code) = :codeLength')
+            ->andWhere('o.code LIKE :prefixAndSuffix')
```
pamil: Personal preference, it brings more context into this query.

```suggestion
            ->andWhere('o.code LIKE :codeTemplate')
```

src/Sylius/Component/Promotion/Repository/PromotionCouponRepositoryInterface.php
```diff
@@ -23,6 +23,12 @@ public function createQueryBuilderByPromotionId($promotionId): QueryBuilder;
 
     public function countByCodeLength(int $codeLength): int;
```
pamil: I wonder whether we could merge the new method with this one (`countByCodeLength(5) === countByCodeLengthWithPrefixAndSuffix(5, null, null)`), but I'm not sure, just thinking out loud.

pamil:
Thanks, Grzegorz! :tada:

PR #10873, opened 2019-11-21
[Psalm] Fix the build after Symfony 4.4 update

pamil:


src/Sylius/Bundle/CoreBundle/Command/InstallCommand.php
```diff
@@ -112,23 +114,23 @@ private function getProperFinalMessage(bool $errored): string
 
     private function getSyliusLogo(): string
```
pamil: I checked, it renders correctly :)

GSadee:

PR #10883, opened 2019-11-27
Rework Behat testing from classes/ids to attributes for login and registration pages

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Ui/Shop/LoginContext.php
+++ b/src/Sylius/Behat/Context/Ui/Shop/LoginContext.php
@@ -75,7 +75,7 @@ final class LoginContext implements Context
     /**
      * @When I want to log in
      */
-    public function iWantToLogIn()
+    public function iWantToLogIn(): void
     {
         $this->loginPage->open();
     }
@@ -83,7 +83,7 @@ final class LoginContext implements Context
     /**
      * @When I want to reset password
      */
-    public function iWantToResetPassword()
+    public function iWantToResetPassword(): void
     {
         $this->requestPasswordResetPage->open();
     }
@@ -91,7 +91,7 @@ final class LoginContext implements Context
     /**
      * @When /^I follow link on (my) email to reset my password$/
      */
-    public function iFollowLinkOnMyEmailToResetPassword(UserInterface $user)
+    public function iFollowLinkOnMyEmailToResetPassword(UserInterface $user): void
     {
         $this->resetPasswordPage->open(['token' => $user->getPasswordResetToken()]);
     }
@@ -99,7 +99,7 @@ final class LoginContext implements Context
     /**
      * @When I specify the username as :username
      */
-    public function iSpecifyTheUsername($username = null)
+    public function iSpecifyTheUsername(string $username = null): void
     {
         $this->loginPage->specifyUsername($username);
     }
@@ -108,7 +108,7 @@ final class LoginContext implements Context
      * @When I specify the email as :email
      * @When I do not specify the email
      */
-    public function iSpecifyTheEmail($email = null)
+    public function iSpecifyTheEmail(?string $email = null): void
     {
         $this->requestPasswordResetPage->specifyEmail($email);
     }
@@ -117,7 +117,7 @@ final class LoginContext implements Context
      * @When I specify the password as :password
      * @When I do not specify the password
      */
-    public function iSpecifyThePasswordAs($password = null)
+    public function iSpecifyThePasswordAs(string $password = null): void
     {
         $this->loginPage->specifyPassword($password);
     }
@@ -126,7 +126,7 @@ final class LoginContext implements Context
      * @When I specify my new password as :password
      * @When I do not specify my new password
      */
-    public function iSpecifyMyNewPassword(string $password = null)
+    public function iSpecifyMyNewPassword(string $password = null): void
     {
         $this->resetPasswordPage->specifyNewPassword($password);
     }
@@ -135,7 +135,7 @@ final class LoginContext implements Context
      * @When I confirm my new password as :password
      * @When I do not confirm my new password
      */
-    public function iConfirmMyNewPassword(string $password = null)
+    public function iConfirmMyNewPassword(string $password = null): void
     {
         $this->resetPasswordPage->specifyConfirmPassword($password);
     }
@@ -144,7 +144,7 @@ final class LoginContext implements Context
      * @When I log in
      * @When I try to log in
      */
-    public function iLogIn()
+    public function iLogIn(): void
     {
         $this->loginPage->logIn();
     }
@@ -153,7 +153,7 @@ final class LoginContext implements Context
      * @When I reset it
      * @When I try to reset it
      */
-    public function iResetIt()
+    public function iResetIt(): void
     {
         /** @var RequestPasswordResetPageInterface|ResetPasswordPageInterface $currentPage */
         $currentPage = $this->currentPageResolver->getCurrentPageWithForm([$this->requestPasswordResetPage, $this->resetPasswordPage]);
@@ -189,7 +189,7 @@ final class LoginContext implements Context
     /**
      * @Then I should be logged in
      */
-    public function iShouldBeLoggedIn()
+    public function iShouldBeLoggedIn(): void
     {
         $this->homePage->verify();
         Assert::true($this->homePage->hasLogoutButton());
@@ -198,7 +198,7 @@ final class LoginContext implements Context
     /**
      * @Then I should not be logged in
      */
-    public function iShouldNotBeLoggedIn()
+    public function iShouldNotBeLoggedIn(): void
     {
         Assert::false($this->homePage->hasLogoutButton());
     }
@@ -206,7 +206,7 @@ final class LoginContext implements Context
     /**
      * @Then I should be notified about bad credentials
      */
-    public function iShouldBeNotifiedAboutBadCredentials()
+    public function iShouldBeNotifiedAboutBadCredentials(): void
     {
         Assert::true($this->loginPage->hasValidationErrorWith('Error Invalid credentials.'));
     }
@@ -214,7 +214,7 @@ final class LoginContext implements Context
     /**
      * @Then I should be notified about disabled account
      */
-    public function iShouldBeNotifiedAboutDisabledAccount()
+    public function iShouldBeNotifiedAboutDisabledAccount(): void
     {
         Assert::true($this->loginPage->hasValidationErrorWith('Error Account is disabled.'));
     }
@@ -222,9 +222,12 @@ final class LoginContext implements Context
     /**
      * @Then I should be notified that email with reset instruction has been sent
      */
-    public function iShouldBeNotifiedThatEmailWithResetInstructionWasSent()
+    public function iShouldBeNotifiedThatEmailWithResetInstructionWasSent(): void
     {
-        $this->notificationChecker->checkNotification('If the email you have specified exists in our system, we have sent there an instruction on how to reset your password.', NotificationType::success());
+        $this->notificationChecker->checkNotification(
+            'If the email you have specified exists in our system, we have sent there an instruction on how to reset your password.',
+            NotificationType::success()
+        );
     }
 
     /**
@@ -238,7 +241,7 @@ final class LoginContext implements Context
     /**
      * @Then I should be notified that my password has been successfully reset
      */
-    public function iShouldBeNotifiedThatMyPasswordHasBeenSuccessfullyReset()
+    public function iShouldBeNotifiedThatMyPasswordHasBeenSuccessfullyReset(): void
     {
         $this->notificationChecker->checkNotification('has been reset successfully!', NotificationType::success());
     }
@@ -247,7 +250,7 @@ final class LoginContext implements Context
      * @Then I should be able to log in as :email with :password password
      * @Then the customer should be able to log in as :email with :password password
      */
-    public function iShouldBeAbleToLogInAsWithPassword($email, $password)
+    public function iShouldBeAbleToLogInAsWithPassword(string $email, string $password): void
     {
         $this->loginPage->open();
         $this->loginPage->specifyUsername($email);
--- a/src/Sylius/Behat/Element/Shop/Account/RegisterElement.php
+++ b/src/Sylius/Behat/Element/Shop/Account/RegisterElement.php
@@ -36,12 +36,12 @@ final class RegisterElement extends Element implements RegisterElementInterface
 
     public function register(): void
     {
-        $this->getDocument()->pressButton('Create an account');
+        $this->getElement('register_button')->click();
     }
 
     public function specifyEmail(?string $email): void
     {
-        $this->getDocument()->fillField('Email', $email);
+        $this->getElement('email')->setValue($email);
     }
 
     public function getEmail(): string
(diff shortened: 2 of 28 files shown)


src/Sylius/Bundle/UiBundle/Form/Type/SecurityLoginType.php
```diff
@@ -21,28 +21,25 @@
...
 {
-    /**
-     * {@inheritdoc}
-     */
     public function buildForm(FormBuilderInterface $builder, array $options): void
     {
         $builder
             ->add('_username', TextType::class, [
                 'label' => 'sylius.form.login.username',
+                'attr' => ['data-login-username' => ''],
```
pamil: I think we should add it to the template instead.

`form_row(form._username, {'attr': {'data-login-username': ''}})`

In order to make it more extendable, let's add a new Twig function `sylius_form_test_attribute(attribute)` which would return `['attr' => [$attribute => '']]` in test environment and `[]` in any other environment.

This way you can define the additional attributes in an easy way:

`form_row(form._username, sylius_form_test_attribute('data-login-username'))`

Or in case that there's some configuration already defined:

`form_row(form._username, sylius_form_test_attribute('data-login-username')|merge({'other': 'config'}))`

Then the attributes could be removed from the PHP form types.

src/Sylius/Bundle/ShopBundle/Resources/views/Account/resetPassword.html.twig
```diff
@@ -18,7 +18,9 @@
                 {{ form_row(form.password.first, {'label': 'sylius.form.user_reset_password.new'|trans}) }}
                 {{ form_row(form.password.second, {'label': 'sylius.form.user_reset_password.confirmation'|trans}) }}
                 {{ form_row(form._token) }}
-                <button type="submit" class="ui fluid large primary submit button">{{ 'sylius.ui.reset'|trans }}</button>
+                <button type="submit" class="ui fluid large primary submit button" data-login-password-reset-button>
```
AdamKasp:
```suggestion
                <button type="submit" class="ui fluid large primary submit button" {{ sylius_test_attribute(data-login-password-reset-button) }}>
```

AdamKasp: here and everywhere

src/Sylius/Bundle/ShopBundle/Resources/views/Account/requestPasswordReset.html.twig
```diff
@@ -22,7 +22,9 @@
                 {{ sonata_block_render_event('sylius.shop.request_password_reset_token.form', {'form': form}) }}
 
                 {{ form_row(form._token) }}
-                <button type="submit" class="ui fluid large primary submit button">{{ 'sylius.ui.reset'|trans }}</button>
+                <button type="submit" class="ui fluid large primary submit button" {{ sylius_test_attribute('data-request-password-reset-button') }}>
```
GSadee: I would be for adding `data-` prefix in extension, so you could have `{{ sylius_test_attribute('request-password-reset-button') }}`

src/Sylius/Bundle/UiBundle/Form/Type/SecurityLoginType.php
Tomanhez: You a right it is great idea 👍

src/Sylius/Bundle/ShopBundle/Resources/views/Account/requestPasswordReset.html.twig
```diff
@@ -17,12 +17,14 @@
             </h1>
             <div class="ui segment">
                 {{ form_start(form, {'action': path('sylius_shop_request_password_reset_token'), 'attr': {'class': 'ui large loadable form', 'novalidate': 'novalidate'}}) }}
-                {{ form_row(form.email) }}
+                {{ form_row(form.email, sylius_data_test_attribute_array('reset-email')) }}
```
pamil:
```suggestion
                {{ form_row(form.email, sylius_test_form_attribute('reset-email')) }}
```
It's an array, but it's highly specific for forms and should only be used there - hence my proposition to change the name.

src/Sylius/Bundle/ShopBundle/Resources/views/Account/requestPasswordReset.html.twig
```diff
@@ -17,12 +17,14 @@
...
 
                 {{ sonata_block_render_event('sylius.shop.request_password_reset_token.form', {'form': form}) }}
 
                 {{ form_row(form._token) }}
-                <button type="submit" class="ui fluid large primary submit button">{{ 'sylius.ui.reset'|trans }}</button>
+                <button type="submit" class="ui fluid large primary submit button" {{ sylius_data_test_attribute_name('request-password-reset-button') }}>
```
pamil:
```suggestion
                <button type="submit" class="ui fluid large primary submit button" {{ sylius_test_html_attribute('request-password-reset-button') }}>
```

Going further with the same logic as in the comment above, this will mark it as specific for HTML and also make the function name shorter.

src/Sylius/Bundle/UiBundle/Twig/DataTestAttributeArrayExtension.php
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * @return TwigFunction[]
+     */
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction(
+                'sylius_data_test_attribute_array',
+                function (?string $data = null, ?string $attributeValue = null): array {
```
pamil:
```suggestion
                function (string $name, ?string $value = null): array {
```

`$data` should be always passed.

I'm also wondering about `$name` for that instead of `$data`.

src/Sylius/Bundle/UiBundle/Twig/DataTestAttributeArrayExtension.php
```diff
@@ -0,0 +1,48 @@
...
+                    if (strpos($this->env, 'test') !== false) {
```
pamil:
```suggestion
                    if (strpos($this->env, 'test') === 0) {
```

I would go for checking if `test` starts the environment string instead of checking whether `test` could be found anywhere in that string.

src/Sylius/Bundle/UiBundle/Twig/DataTestAttributeArrayExtension.php
```diff
@@ -0,0 +1,48 @@
...
+                        return ['attr' => ['data-test-'.$data => sprintf('"%s"', $attributeValue)]];
```
pamil:
```suggestion
                        return ['attr' => ['data-test-'.$data => (string) $value];
```

pamil: I think we don't need `"` here.

src/Sylius/Bundle/UiBundle/Twig/DataTestAttributeNameExtension.php
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * @return TwigFunction[]
+     */
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction(
+                'sylius_data_test_attribute_name',
+                function (?string $data = null, ?string $attributeValue = null): ?string {
```
pamil:
```suggestion
                function (string $name, ?string $value = null): string {
```

src/Sylius/Bundle/UiBundle/Twig/DataTestAttributeNameExtension.php
```diff
@@ -0,0 +1,48 @@
...
+                    if (strpos($this->env, 'test') !== false) {
```
pamil: Same as above.

src/Sylius/Bundle/UiBundle/Twig/DataTestAttributeNameExtension.php
```diff
@@ -0,0 +1,48 @@
...
+                        return sprintf('data-test-%s="%s"', $data, $attributeValue);
```
pamil: Let's cast `$attributeValue` to string explicitly - `(string) $attributeValue`.

src/Sylius/Bundle/UiBundle/Twig/DataTestAttributeNameExtension.php
```diff
@@ -0,0 +1,48 @@
...
+                    }
+
+                    return null;
```
pamil: It should rather return an empty string here.

src/Sylius/Behat/Element/Shop/Account/RegisterElement.php
```diff
@@ -51,43 +51,45 @@ public function getEmail(): string
...
     protected function getDefinedElements(): array
     {
         return array_merge(parent::getDefinedElements(), [
-            'email' => '#sylius_customer_registration_email',
-            'first_name' => '#sylius_customer_registration_firstName',
-            'last_name' => '#sylius_customer_registration_lastName',
-            'password' => '#sylius_customer_registration_user_plainPassword_first',
-            'password_verification' => '#sylius_customer_registration_user_plainPassword_second',
-            'phone_number' => '#sylius_customer_registration_phoneNumber',
+            'email' => '[data-test-email]',
```
lchrusciel: Maybe we can process some helper here? It could be done in the next iteration.

lchrusciel: Something like this:
```suggestion
            'email' => TestAttribute::string('email'),
```

src/Sylius/Bundle/UiBundle/Twig/TestFormAttributeExtension.php
```diff
@@ -0,0 +1,48 @@
...
+     * @return TwigFunction[]
+     */
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction(
+                'sylius_test_form_attribute',
+                function (string $name, ?string $value = null): array {
+                    if (strpos($this->env, 'test') === 0) {
+                        return ['attr' => ['data-test-'.$name => (string) $value]];
```
lchrusciel:
```suggestion
                        return ['attr' => ['data-test-' . $name => (string) $value]];
```

lchrusciel: But in the case, we would like to introduce helper class, we can do
```suggestion
                        return ['attr' => TestAttribure::array($name, $value)];
```

src/Sylius/Bundle/UiBundle/Twig/TestHtmlAttributeExtension.php
```diff
@@ -0,0 +1,48 @@
...
+     * @return TwigFunction[]
+     */
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction(
+                'sylius_test_html_attribute',
+                function (string $name, ?string $value = null): string {
+                    if (strpos($this->env, 'test') === 0) {
+                        return sprintf('data-test-%s="%s"', $name, (string) $value);
```
lchrusciel:
```suggestion
                        return TestAttribute::string($name, $value);
```

src/Sylius/Bundle/UiBundle/Twig/TestFormAttributeExtension.php
pamil: I'd say a helper class used only in one place without any possibility to replace is not a good idea, especially for maintainers.

src/Sylius/Behat/Element/Shop/Account/RegisterElement.php
pamil: I get the idea here but I'd leave it for later when we will have more experience with the new system so that we create right abstraction layer.

src/Sylius/Behat/Context/Ui/Shop/LoginContext.php
```diff
@@ -117,7 +117,7 @@ public function iSpecifyTheEmail($email = null)
      * @When I specify the password as :password
      * @When I do not specify the password
      */
-    public function iSpecifyThePasswordAs($password = null)
+    public function iSpecifyThePasswordAs(string $password = null): void
```
pamil:
```suggestion
    public function iSpecifyThePasswordAs(?string $password = null): void
```

src/Sylius/Behat/Context/Ui/Shop/LoginContext.php
```diff
@@ -126,7 +126,7 @@ public function iSpecifyThePasswordAs($password = null)
      * @When I specify my new password as :password
      * @When I do not specify my new password
      */
-    public function iSpecifyMyNewPassword(string $password = null)
+    public function iSpecifyMyNewPassword(string $password = null): void
```
pamil:
```suggestion
    public function iSpecifyMyNewPassword(?string $password = null): void
```

src/Sylius/Behat/Context/Ui/Shop/LoginContext.php
```diff
@@ -135,7 +135,7 @@ public function iSpecifyMyNewPassword(string $password = null)
      * @When I confirm my new password as :password
      * @When I do not confirm my new password
      */
-    public function iConfirmMyNewPassword(string $password = null)
+    public function iConfirmMyNewPassword(string $password = null): void
```
pamil:
```suggestion
    public function iConfirmMyNewPassword(?string $password = null): void
```

src/Sylius/Behat/Context/Ui/Shop/LoginContext.php
```diff
@@ -75,31 +75,31 @@ public function __construct(
...
+    public function iFollowLinkOnMyEmailToResetPassword(UserInterface $user): void
     {
         $this->resetPasswordPage->open(['token' => $user->getPasswordResetToken()]);
     }
 
     /**
      * @When I specify the username as :username
      */
-    public function iSpecifyTheUsername($username = null)
+    public function iSpecifyTheUsername(string $username = null): void
```
pamil:
```suggestion
    public function iSpecifyTheUsername(?string $username = null): void
```

src/Sylius/Bundle/UiBundle/Twig/TestFormAttributeExtension.php
lchrusciel: It won't be used only in one place. This class should be used in all behat scenarios. What I would like to achieve, is to have this string (and array) building logic in one class. 

Injecting it would be even better of course. The question is if it still is worth it when we will have to inject this service to every page. On the other hand, we can just extend parent class and add it as a protected field.

src/Sylius/Bundle/UiBundle/Twig/TestFormAttributeExtension.php
```diff
@@ -0,0 +1,48 @@
...
+
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFunction;
+
+final class TestFormAttributeExtension extends AbstractExtension
+{
+    /** @var string */
+    private $env;
+
+    public function __construct(string $env)
```
pamil: Btw. I would go for `$environment`, no need for abbreviations in the code.

src/Sylius/Bundle/UiBundle/Twig/TestFormAttributeExtension.php
lchrusciel: Just to be clear. The same service (or helper class) should be used here and here: https://github.com/Sylius/Sylius/pull/10883/files#r352113248

pamil: The output is different here:
 - for rendering we need:
   - HTML attribute: `data-test-key` or `data-test-key="value"`
   - form attribute: `['attr' => ['data-test-key' => '']]` or `['attr' => ['data-test-key' => 'value']]`
 - for matching we need:
   - either purely key-based: `[data-test-key]`
   - or basing on key and value: `[data-test-key="value"]`

If we go with static class, I'd only implement two methods:
 - one that creates the identifier, passing `key` as a parameter would return string `data-test-key`
 - one that creates the matcher, passing `key` as a parameter would return string  `[data-test-key]` and passing `key` and `value` would return string `[data-test-key="value"]`

I don't see where method returning array `['data-test-'.$key => (string) $value]` for `$key, $value` could be reused.

lchrusciel: I thought about putting all creation methods into one class, but anything will be better than writing strings

lchrusciel:

src/Sylius/Bundle/ShopBundle/Resources/views/Menu/_security.html.twig
```diff
@@ -2,7 +2,7 @@
     {% if is_granted('ROLE_USER') %}
         <div class="item">{{ 'sylius.ui.hello'|trans }} {{ app.user.customer.fullName }}!</div>
         <a href="{{ path('sylius_shop_account_dashboard') }}" class="item">{{ 'sylius.ui.my_account'|trans }}</a>
-        <a href="{{ path('sylius_shop_logout') }}" class="item sylius-logout-button">{{ 'sylius.ui.logout'|trans }}</a>
+        <a href="{{ path('sylius_shop_logout') }}" class="item" {{ sylius_test_html_attribute('logout-button') }}>{{ 'sylius.ui.logout'|trans }}</a>
```
lchrusciel:
```suggestion
        <a href="{{ path('sylius_shop_logout') }}" class="item  sylius-logout-button" {{ sylius_test_html_attribute('logout-button') }}>{{ 'sylius.ui.logout'|trans }}</a>
```
Please revert this change

PR #10904, opened 2019-12-05
Decrement length constraint on postcode

loevgaard:
In Oslo, Norway there is a postal code `1`. See https://worldpostalcode.com/norway/hordaland/oslo

| Q               | A
| --------------- | -----
| Branch?         | All
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Resources/config/validation/Address.xml
+++ b/src/Sylius/Bundle/AddressingBundle/Resources/config/validation/Address.xml
@@ -114,7 +114,7 @@
                 </option>
             </constraint>
             <constraint name="Length">
-                <option name="min">2</option>
+                <option name="min">1</option>
                 <option name="max">255</option>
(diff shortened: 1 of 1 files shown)


CoderMaggie:
After a quick research, although there is this postal code you referred, but it is written as `0001` not just `1`
ref: 
1. http://www.fedex.com/be/shippingguide/postalcodeformat.html 
2. http://www.dhl.com.tw/content/dam/downloads/tw/express/forms/postcode_formats.pdf

loevgaard:
Yes, that is correct, but does users adhere for standards? No :) The same in Denmark. We also have 4 digit postal codes, but there postal codes like 0983. When people order, they write 983 obviously.

lchrusciel:
Thanks, Joachim! :1st_place_medal:

loevgaard:
Great, thank you!

PR #10911, opened 2019-12-09
Add code to exception message since name can be null

loevgaard:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
+++ b/src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
@@ -30,9 +30,10 @@ final class ProductVariantPriceCalculator implements ProductVariantPriceCalculat
 
         if (null === $channelPricing) {
             throw new MissingChannelConfigurationException(sprintf(
-                'Channel %s has no price defined for product variant %s',
+                'Channel %s has no price defined for product variant %s (%s)',
                 $context['channel']->getName(),
-                $productVariant->getName()
+                $productVariant->getName(),
+                $productVariant->getCode()
             ));
         }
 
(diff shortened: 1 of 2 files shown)


src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
```diff
@@ -30,9 +30,10 @@ public function calculate(ProductVariantInterface $productVariant, array $contex
 
         if (null === $channelPricing) {
             throw new MissingChannelConfigurationException(sprintf(
-                'Channel %s has no price defined for product variant %s',
+                'Channel %s has no price defined for product variant %s (%s)',
                 $context['channel']->getName(),
-                $productVariant->getName()
+                $productVariant->getName(),
+                $productVariant->getCode()
```
Zales0123: Imo, it would still look quite confusing if there is no name - `Channel DEFAULT has no price defined for product variant  (VARIANT_CODE)` (double space 💃). What about fallback to code if there is no name? `$productVariant->getName() ?? $productVariant->getCode()`?

Then we have either

`Channel DEFAULT has no price defined for product variant Variant name`

or

`Channel DEFAULT has no price defined for product variant VARIANT_CODE`

loevgaard: It's a good idea, but the code is the most valuable. Much easier to find, both in a db viewer or in the admin interface. I will make it nicer :)

loevgaard: What do you think now?

pamil:
Thanks, Joachim! :tada:

loevgaard:
My pleasure!

PR #10924, opened 2019-12-11
New template events architecture (introduction)

pamil:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | -
| License         | MIT

This PR aims to replace our current template events system using Sonata Block bundle and improve both the developer experience using it and Sylius extendability, especially in the light of [Sylius Store](https://sylius.com/plugins/) which has been recently opened.

Apart from the usual code review, I'd like to get feedback on high-level design and your experiences with our current template events system.
<hr/>

## Current implementation

 - We use Sonata Block bundle to create extension points in Sylius, which are recommended to be used by our plugins.

 - An extension point is created by calling `sonata_block_render_event` Twig function:

    ```twig
    {{ sonata_block_render_event('sylius.admin.order.show.before_summary', {'resource': resource}) }}

    <div class="ui segment">
        {% include '@SyliusAdmin/Order/Show/_summary.html.twig' %}
    </div>

    {{ sonata_block_render_event('sylius.admin.order.show.after_summary', {'resource': resource}) }}
    ``` 

 - A block can be injected by creating a template and registering a event listener in DIC:

    ```yaml
    services:
        my.service.name.rendering.block:
            class: Sylius\Bundle\UiBundle\Block\BlockEventListener
            arguments:
                - 'template.html.twig'
            tags:
                - { name: kernel.event_listener, event: sonata.block.event.sylius.admin.order.show.before_summary, method: onBlockEvent }
    ```

    The most obvious issue here is the need to put so much boilerplate code and to concatenate the event name with `sonata.block.event.*` prefix.

 - In order to pass any variable from the template to the rendered blocks, it needs to be defined in `sonata_block` configuration:

    ```yaml
    sonata_block:
        blocks:
            sonata.block.service.template:
                settings:
                    resource: ~
    ```

    After passing a variable to the context, it has to be accessed using `block.settings.*` prefix:

    ```twig
    {{ dump(block.settings.resource) }}
    ```

 - The only possibility to remove a previously registered block is to add a compiler pass removing the listener service that adds it.

 - There's no possibility to change rendering order of blocks.

## Proposed implementation

 - Pure Twig implementation is used, no event listeners for increased performance.
 - An extension point is created by calling `sylius_template_event` Twig function:

    ```twig
    {{ sylius_template_event('sylius.admin.order.show.before_summary', {'resource': resource}) }}

    <div class="ui segment">
        {% include '@SyliusAdmin/Order/Show/_summary.html.twig' %}
    </div>

    {{ sylius_template_events('sylius.admin.order.show.after_summary', {'resource': resource}) }}
    ```

 - A block can be injected by configuring UiBundle:

    ```yaml
    sylius_ui:
        events:
            sylius.admin.order.show.before_summary:
                blocks:
                    some_block: 'template.html.twig' # short syntax
                    another_block: # long syntax
                        template: 'another.html.twig' 
    ```

 - It is possible to pass any variable from the template to the block without any configuration changes.

    It can be accessed using the same variable name:

    ```twig
    {{ dump(resource) }}
    ```

 - It is possible to remove a block by disabling it in the config (assuming the configuration mentioned above comes from some Sylius plugin):

    ```yaml
    # Your application configuration
    sylius_ui:
        events:
            sylius.admin.order.show.before_summary:
                blocks:
                    some_block: false # short syntax
                    another_block: # long syntax
                        enabled: false
    ```

 - It is possible to change the template of some block:

    ```yaml
    # Your application configuration
    sylius_ui:
        events:
            sylius.admin.order.show.before_summary:
                blocks:
                    some_block: 'my_other_template.html.twig'
    ```

 - It is possible to define priorities for blocks, allowing plugin creators and e-commerce developers to change blocks order easily or inject a block in the middle of other ones:

    ```yaml
    # Vendor configuration
    sylius_ui:
        events:
            sylius.admin.order.show.before_summary:
                blocks:
                    first_block: 'first.html.twig' # default priority is 0
                    last_block: 
                        template: 'last.html.twig'
                        priority: -5

    # Your application configuration
    sylius_ui:
        events:
            sylius.admin.order.show.before_summary:
                blocks:
                    middle_block: 
                        template: 'middle.html.twig'
                        priority: -3  # between 0 (first_block) and -5 (last_block)
    ```

 - Having an easy way to disable and reorder blocks, we can merge events like `before_header` and `after_header` into one `header` event with proper priorities. This way we can use event blocks for core Sylius templates and allow for simpler customisations like changing the order of sidebar menus in order show page in admin: 

    Screenshot 2019-12-11 at 14 40 14

    This sidebar can consist of four blocks - customer info, addresses, payment, shipment - that can be manipulated easily.

<hr/>

## Todo & ideas

 - [x] Add deprecation notices to `UPGRADE.md` file
 - [x] Consider improving the naming (template events, blocks? looking for feedback on it)
 - [ ] Replace existing `sonata_block_render_event` method calls with `sylius_template_event`
    - [ ] Register blocks that will keep backwards compatibility and call `sonata_block_render_event`
    - [ ] Merge existing `*.before_something` and `*.after_something` events into `*.something` and use them for core functionalities as well (so that one can add a block in the middle of core ones or disable it)
 - [ ] Write more extensive documentation for template events
 - [ ] Consider creating [ADR (Architecture Decision Record)](https://github.com/joelparkerhenderson/architecture_decision_record) detailing the reasoning, alternatives and pros/cons of this solution
 - [ ] Create _best practices_ guide (what priorities to use in your application / your plugin) so that Sylius ecosystem can adopt it without struggles
 - [ ] Increase discoverability of events and blocks:
    - [ ] Debugging commands for CLI (show all registered events, show all blocks for an event)
    - [ ] HTML comments with event / block names before rendered content
 - [ ] Create issues in downstream repositories making use of current events implementation
<hr/>

## Takeaways

 - The second commit contains a functional test with the whole test application running using `MicroKernelTrait` - this might be the way to test our other bundles, our current functional testing architecture dates to Symfony 2 and still has `app/console` in there.
    - I'm not 100% happy with that test, because it requires the kernel to be configured accordingly - ideally, the test would contain both kernel config and assertions for it. Nevertheless, it's a reliable test to see whether it really works.


src/Sylius/Bundle/UiBundle/Block/BlockEventListener.php
```diff
@@ -23,6 +23,11 @@ final class BlockEventListener
 
     public function __construct(string $template)
     {
+        @trigger_error(
+            sprintf('Using "%s" to add blocks to the templates is deprecated since Sylius 1.7. Use "sylius_ui" configuration instead.', self::class),
+            \E_USER_DEPRECATED
```
gabiudrescu: I would mention here explicitly when this is going to be removed from codebase. I assume 2.0, right?

src/Sylius/Bundle/UiBundle/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,54 @@
...
+                                    ->canBeDisabled()
+                                    ->beforeNormalization()
+                                        ->ifString()
+                                        ->then(static function (string $template): array {
+                                            return ['template' => $template];
+                                        })
+                                    ->end()
+                                    ->children()
+                                        ->scalarNode('template')->isRequired()->cannotBeEmpty()->end()
+                                        ->integerNode('priority')->defaultValue(0)->end()
```
Zales0123: Can we mention the usage of priorities somewhere in the documentation? I have a deep fear that we will forget about it if we leave skip it now 😄

src/Sylius/Bundle/UiBundle/DependencyInjection/SyliusUiExtension.php
```diff
@@ -17,15 +17,46 @@
...
+    private function loadEvents(array $eventsConfig, ContainerBuilder $container): void
+    {
+        $templateEventExtensionDefinition = $container->getDefinition('sylius.twig.extension.template_event');
+
+        $blocksForEvents = [];
+        foreach ($eventsConfig as $eventName => $eventConfiguration) {
+            $blocksPriorityQueue = new SplPriorityQueue();
+
+            foreach ($eventConfiguration['blocks'] as $blockName => $details) {
+                if ($details['enabled'] === false) {
```
Zales0123: Can we pass any value in the `enabled` key? Even some string and it won't be parsed to bool?

src/Sylius/Bundle/UiBundle/spec/Twig/TemplateEventExtensionSpec.php
```diff
@@ -0,0 +1,47 @@
...
+
+declare(strict_types=1);
+
+namespace spec\Sylius\Bundle\UiBundle\Twig;
+
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Twig\Environment;
+
+class TemplateEventExtensionSpec extends ObjectBehavior
```
Zales0123: Should be final

src/Sylius/Bundle/UiBundle/DependencyInjection/Configuration.php
pamil: I'll add them as soon as I'm done with changing Sonata blocks to Sylius blocks.

src/Sylius/Bundle/UiBundle/DependencyInjection/SyliusUiExtension.php
pamil: Enabled is `booleanNode`, so it shouldn't be possible.

Zales0123: So maybe `if(!$details['enabled'])` is enought? 😄

pamil:
Will be continued in further PRs.

PR #10932, opened 2019-12-12
[Admin][Order] Add resending an order confirmation email

### LINKED ISSUE #10907: Resend order confirmation email from backend
Have a button on the order show page in `admin` where you can resend the order confirmation email.

None

**Additional context**
When a customer writes their email wrong and the admin edits it manually on the customer, then this feature is mandatory :D

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10907 
| License         | MIT

Based on https://github.com/Sylius/Sylius/pull/10936

screencapture-127-0-0-1-8007-admin-orders-23-2019-12-12-14_40_39

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Action/ResendOrderConfirmationEmailAction.php
+++ b/src/Sylius/Bundle/AdminBundle/Action/ResendOrderConfirmationEmailAction.php
@@ -1,0 +1,78 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\AdminBundle\Action;
+
+use Sylius\Bundle\AdminBundle\EmailManager\OrderEmailManagerInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Repository\OrderRepositoryInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\Session\Session;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+use Symfony\Component\HttpKernel\Exception\HttpException;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\Security\Csrf\CsrfToken;
+use Symfony\Component\Security\Csrf\CsrfTokenManagerInterface;
+
+final class ResendOrderConfirmationEmailAction
+{
+    /** @var OrderRepositoryInterface */
+    private $orderRepository;
+
+    /** @var OrderEmailManagerInterface */
+    private $orderEmailManager;
+
+    /** @var CsrfTokenManagerInterface */
+    private $csrfTokenManager;
+
+    /** @var Session */
+    private $session;
+
+    public function __construct(
+        OrderRepositoryInterface $orderRepository,
+        OrderEmailManagerInterface $orderEmailManager,
+        CsrfTokenManagerInterface $csrfTokenManager,
+        SessionInterface $session
+    ) {
+        $this->orderRepository = $orderRepository;
+        $this->orderEmailManager = $orderEmailManager;
+        $this->csrfTokenManager = $csrfTokenManager;
+        $this->session = $session;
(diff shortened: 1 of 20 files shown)


src/Sylius/Bundle/AdminBundle/Resources/config/routing/order.yml
```diff
@@ -119,3 +119,10 @@ sylius_admin_order_shipment_ship:
                     parameters:
                         orderId: $orderId
                         id: $id
+
+
```
lchrusciel:
```suggestion
```

src/Sylius/Bundle/AdminBundle/Action/ResendOrderConfirmationEmailAction.php
```diff
@@ -0,0 +1,65 @@
...
+        Session $session
+    ) {
+        $this->orderRepository = $orderRepository;
+        $this->orderEmailManager = $orderEmailManager;
+        $this->session = $session;
+    }
+
+    public function __invoke(Request $request): Response
+    {
+        $orderId = $request->attributes->get('id');
```
lchrusciel: Can we add CSRF validation?

src/Sylius/Bundle/AdminBundle/Resources/views/Order/show.html.twig
```diff
@@ -47,6 +47,7 @@
 
         {% include '@SyliusAdmin/Order/Show/_payments.html.twig' %}
         {% include '@SyliusAdmin/Order/Show/_shipments.html.twig' %}
+        {% include '@SyliusAdmin/Order/Show/_resendEmail.html.twig' %}
```
lchrusciel: It is just a personal filling, but I would say, that it would be better to add this button next to cancel buttons etc at the top of the page

GSadee: It is quite a big button and I've consulted that with @CoderMaggie that this will be a proper place

src/Sylius/Bundle/AdminBundle/EmailManager/OrderEmailManager.php
```diff
@@ -0,0 +1,42 @@
...
+
+    public function __construct(SenderInterface $emailSender)
+    {
+        $this->emailSender = $emailSender;
+    }
+
+    public function sendConfirmationEmail(OrderInterface $order): void
+    {
+        $this->emailSender->send(
+            Emails::ORDER_CONFIRMATION,
```
lchrusciel: We should use `order_confirmation_resent` as code here (or constant with it) or rethink the way, how we are sending emails

pjedrzejewski:
Reaction: I think typically it is more visible in the top (above the customer or maybe the top menu? Here it will be really hidden.

CoderMaggie:
@pjedrzejewski I know, but it not a common action, but an edge-case. It will not be performed all the time, that's why I've requested placing it on the bottom :)

src/Sylius/Bundle/AdminBundle/Action/ResendOrderConfirmationEmailAction.php
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        $this->orderEmailManager->sendConfirmationEmail($order);
+
+        $this->session->getFlashBag()->add(
+            'success',
+            'sylius.email.order_confirmation_resent'
+        );
+
+        return new RedirectResponse($request->headers->get('referer'));
```
lchrusciel: In the long term, we should think about providing a way to handle these actions in Admin API as well, because right now Admin API and normal Admin panel will drift from each other with the feature list

src/Sylius/Bundle/AdminBundle/Resources/views/Email/orderConfirmation.html.twig
```diff
@@ -0,0 +1,29 @@
+{% extends '@SyliusAdmin/Email/layout.html.twig' %}
+
+{% block subject %}
+    {{ 'sylius.email.order_confirmation.subject'|trans({}, null, localeCode) }}
+{% endblock %}
+
+{% block content %}
+    {% set url = channel.hostname is not null ? 'http://' ~ channel.hostname ~ path('sylius_shop_order_show', {'tokenValue': order.tokenValue, '_locale': localeCode}) : url('sylius_shop_order_show', {'tokenValue': order.tokenValue, '_locale': localeCode}) %}
```
lchrusciel: We need to resolve this coupling ASAP

lchrusciel:

PR #10937, opened 2019-12-13
Rework ShowPage and IndexPage in Shop/Product

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Page/Shop/Product/ShowPage.php
+++ b/src/Sylius/Behat/Page/Shop/Product/ShowPage.php
@@ -24,18 +24,12 @@ use Webmozart\Assert\Assert;
 
 class ShowPage extends SymfonyPage implements ShowPageInterface
 {
-    /**
-     * {@inheritdoc}
-     */
     public function getRouteName(): string
     {
         return 'sylius_shop_product_show';
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function addToCart()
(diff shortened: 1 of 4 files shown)


GSadee:
Thanks, Tomasz! :tada:

src/Sylius/Behat/Page/Shop/Product/ShowPage.php
```diff
@@ -145,121 +95,91 @@ public function getAttributeByName(string $name): ?string
...
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function hasReviewTitled($title)
+    public function hasAddToCartButton(): bool
     {
-        return null !== $this->getElement('reviews')->find('css', sprintf('.comment:contains("%s")', $title));
+        return $this->getDocument()->hasButton('Add to cart')
```
lchrusciel: Something like this? 
```suggestion
        return 
            $this->getDocument()->hasButton('Add to cart') &&
            false === $this->getDocument()->findButton('Add to cart')->hasAttribute('disabled')
        ;
```
I'm wondering if we cannot fetch a button, check if it is null and then check if it has disabled attribute in separate checks, as it will be easier to read

PR #10938, opened 2019-12-13
[Payum][Paypal] Use full price instead of discounted one

Prometee:
| Q               | A
| --------------- | -----
| Branch?         | 1.4 to master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | unknown
| License         | MIT

I was having trouble with many PayPal errors during the black friday and the error was :
`"The totals of the cart item amounts do not match order amounts."`

The bug was caused by the fact that PayPal make a validation of the total price with the sum of each item lines we build. This bug appear because all discounted prices are regrouped into an item line named "Discount" so all other lines shouldn't be filled with something related with discounts.


Zales0123:
may become messed-up.**

execute `git fetch origin && git reset "payum-paypal"` to update your local branch.
 
lchrusciel:
Thanks, Francis! :tada:

PR #10952, opened 2019-12-17
Add option to link taxon to a parent in a fixture

Jeroen-G:
| Q               | A
| --------------- | -----
| Branch?         | 1.6
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

The taxon fixture did not allow for a parent to be set. I added the option to pass the code of an existing parent taxon to a new one with the `parent:` attribute in a fixture yaml file.


src/Sylius/Bundle/CoreBundle/Fixture/Factory/TaxonExampleFactory.php
```diff
@@ -135,6 +140,9 @@ protected function configureOptions(OptionsResolver $resolver): void
             ->setAllowedTypes('translations', ['array'])
             ->setDefault('children', [])
             ->setAllowedTypes('children', ['array'])
+            ->setDefault('parent', function (Options $options): string {
```
lchrusciel: It seems wrong. We don't want to have the same taxon as it's parent by default. The default value should be null and we should add normalizer, which will translate some `code` to taxon itself. 
```suggestion
            ->setDefault('parent', null)
            ->setAllowedTypes('parent', ['null', 'string', TaxonInterface::class])
            ->setNormalizer('parent', LazyOption::findOneBy($this->taxonRepository, 'code'))
```

Jeroen-G:
As an example I have used the syntax in the t-shirts example fixture.

src/Sylius/Bundle/CoreBundle/spec/Fixture/Factory/TaxonExampleFactorySpec.php
```diff
@@ -25,7 +25,7 @@
 
 class TaxonExampleFactorySpec extends ObjectBehavior
 {
-    function let(
+    public function let(
```
lchrusciel: All our specs don't have any `public` accessor. Can you revert this change?

Jeroen-G: Done! It was the auto formatter from my editor 👀

lchrusciel: but there is one more file to fix :P https://github.com/Sylius/Sylius/pull/10952/files#diff-f275e213fd9b5b34a08b735a342f79d9R28

Jeroen-G: 😩 

Fixed.

lchrusciel:
One more thing, this is a new feature, therefore we need to rebase this branch to master

PR #10954, opened 2019-12-17
Make products in fixtures translatable

Jeroen-G:
| Q               | A
| --------------- | -----
| Branch?         | 1.6
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Instead of using one pair of data for all locales, I looked at how the taxons are created and made a similar implementation for products.

Example of how the yaml would look:
```yaml
example_product:
          name: product
          options:
            custom:
              - name: 'Cakevorm 25cm'
                slug: 'Cakevorm-25cm'
                code: 'sku01'
                channels:
                  - 'BAKKEN_WEB'
                enabled: true
                tracked: false
                tax_category: 'high_sales_tax'
                main_taxon: 'BAKVORMEN'
                taxons:
                  - 'BAKVORMEN'
                translations:
                  nl_NL:
                    name: 'Cakevorm 25cm'
                    slug: 'cakevorm-25cm'
                  en_GB:
                    name: 'Cake mold 25cm'
                    slug: 'cake-mold-25cm'
```

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
@@ -214,6 +214,7 @@ class ProductExampleFactory extends AbstractExampleFactory implements ExampleFac
             ->setDefault('product_attributes', [])
             ->setAllowedTypes('product_attributes', 'array')
             ->setNormalizer('product_attributes', function (Options $options, array $productAttributes): array {
+                $productAttributes = $this->filterTranslationsFromAttributes($productAttributes);
                 return $this->setAttributeValues($productAttributes);
             })
 
@@ -448,5 +449,12 @@ class ProductExampleFactory extends AbstractExampleFactory implements ExampleFac
             },
             ''
         ));
+    }
+
+    private function filterTranslationsFromAttributes(array $productAttributes): array
+    {
+        return array_filter($productAttributes, function ($code) {
+            return is_string($code);
+        }, ARRAY_FILTER_USE_KEY);
     }
 }
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
@@ -228,6 +228,9 @@ class ProductExampleFactory extends AbstractExampleFactory implements ExampleFac
 
             ->setDefault('tax_category', null)
             ->setAllowedTypes('tax_category', ['string', 'null', TaxCategoryInterface::class])
+
+            ->setDefault('translations', [])
+            ->setAllowedTypes('translations', ['array'])
         ;
 
         if ($this->taxCategoryRepository !== null) {
@@ -237,15 +240,28 @@ class ProductExampleFactory extends AbstractExampleFactory implements ExampleFac
 
     private function createTranslations(ProductInterface $product, array $options): void
     {
+        // add translation for each defined locales
         foreach ($this->getLocales() as $localeCode) {
-            $product->setCurrentLocale($localeCode);
-            $product->setFallbackLocale($localeCode);
-
-            $product->setName($options['name']);
-            $product->setSlug($options['slug']);
-            $product->setShortDescription($options['short_description']);
-            $product->setDescription($options['description']);
+            $this->createTranslation($product, $localeCode, $options);
         }
+
+        // create or replace with custom translations
+        foreach ($options['translations'] as $localeCode => $translationOptions) {
+            $this->createTranslation($product, $localeCode, $translationOptions);
+        }
+    }
+
+    private function createTranslation(ProductInterface $product, string $localeCode, array $options = []): void
+    {
+        $options = $this->optionsResolver->resolve($options);
+
+        $product->setCurrentLocale($localeCode);
+        $product->setFallbackLocale($localeCode);
+
+        $product->setName($options['name']);
+        $product->setSlug($options['slug'] ?: $this->slugGenerator->generate($product, $localeCode));
+        $product->setShortDescription($options['short_description']);
+        $product->setDescription($options['description']);
     }
 
     private function createRelations(ProductInterface $product, array $options): void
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
```diff
@@ -214,6 +214,7 @@ protected function configureOptions(OptionsResolver $resolver): void
             ->setDefault('product_attributes', [])
             ->setAllowedTypes('product_attributes', 'array')
             ->setNormalizer('product_attributes', function (Options $options, array $productAttributes): array {
+                $productAttributes = $this->filterTranslationsFromAttributes($productAttributes);
```
lchrusciel: Why is it a part of this PR? This change seems to not be related to translation at all

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
```diff
@@ -237,17 +241,30 @@ protected function configureOptions(OptionsResolver $resolver): void
...
     }
 
+    private function createTranslation(ProductInterface $product, string $localeCode, array $options = []): void
+    {
+        $options = $this->optionsResolver->resolve($options);
+
+        $product->setCurrentLocale($localeCode);
+        $product->setFallbackLocale($localeCode);
+
+        $product->setName($options['name']);
```
lchrusciel: If I understood it correctly, there is a high possibility, that this code will throw an exception if `name`, `description` or  `short_description ` won't be defined for translation. We should ensure some structure here and default values. 

Can we also write some tests for this configuration? We could take this factory and try to create different objects for different options.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
Jeroen-G: I was not completely sure what happened, but the tests failed without it. I think the translations were somehow interpreted as a product attribute, but since they lacked the options of a attribute, the test failed.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
Jeroen-G: I'm not really sure how the factory works, but I'll take a look at the taxon factory for ideas on implementing some defaults.

Jeroen-G: By the way, `name`, `description` and `short_description` are not allowed to be empty, so the translations will not be the first to throw an exception because of those values missing.

Jeroen-G: I have added some tests, als for incomplete translations!

RickvdStaaij:
Hi @lchrusciel, can you take a look at this PR? @Jeroen-G has added the tests you requested. We're hoping to include this in our Sylius Plus fixtures soon.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
lchrusciel: Can you post the exact error that happened here? I'm worrying that we will hide some issues with this change.

lchrusciel: As next week we will have a feature freeze, can you provide this rather ASAP?

Jeroen-G:
```
TypeError: Argument 1 passed to Sylius\Bundle\CoreBundle\Fixture\Factory\ProductExampleFactory::configureProductAttributeValue() must be of the type string, int given, called in ProductExampleFactory.php on line 382
```

Jeroen-G:
@lchrusciel I posted the error in the conversation above.

src/Sylius/Bundle/CoreBundle/Fixture/Factory/ProductExampleFactory.php
```diff
@@ -237,17 +241,30 @@ protected function configureOptions(OptionsResolver $resolver): void
...
-            $product->setDescription($options['description']);
+        // create or replace with custom translations
+        foreach ($options['translations'] as $localeCode => $translationOptions) {
+            $this->createTranslation($product, $localeCode, $translationOptions);
         }
     }
 
+    private function createTranslation(ProductInterface $product, string $localeCode, array $options = []): void
+    {
+        $options = $this->optionsResolver->resolve($options);
```
lchrusciel: I've spotted a problem with the current implementation. We shouldn't resolve options again here(that's why you had to add this filtration in the first place).

What should be done, is to remove filtration logic from product attributes, extract translation a logic to separate example factory and delegate work into it(more or less what is happening here). The most tricky part would be to avoid bc-breaks, so we need to keep the current behavior, and use this additional logic only for defined transition.

It should be separate service, with an allowance of injecting null

Jeroen-G: Going back and forth like this is not really doing it for me. Feel free to push the desired changes.

lchrusciel: Oh, excuse me for pushing back and forth. I didn't want to push you away from these changes. This PR could be a beneficial enhancement for the platform.

For sure, the feedback loop could be faster, and I'm sorry about that. We are doing our best to be daily available on this repository. 

However, to spot this problem and provide you finally (I hope) valuable feedback about this change, I had to fetch your branch, and thanks to your tests, debug it with xdebug. It's not possible to do with every single PR, and that is why I haven't done it in the first place. Due to our rather strict bc policy, we need to be careful what lands in the codebase. The current solution is more like a workaround so that it couldn't be merged.

 As today we've started feature freeze phase, this PR will not be released before 1.8 (September 2020), we have some time to improve this solution. I will ask our PO to add it to the backlog. Maybe it will help this feature to land into the core.

Thanks again for your time and patience. I hope we will have another opportunity to work together on some changes in the future, and it will go smoother than this time.

Jeroen-G: Thank you for your kind words. Since [we became a partner](https://sylius.com/partners/enrise/) I think we could work together in the future for sure.

Issue #10928, opened 2019-12-11
The road to Symfony 5


pamil:
Symfony 5 has been released a few weeks ago, now's the time to make the effort to support it.

## Dependencies

I've compiled a list of dependencies that either need to be updated on Sylius side (to next major releases) or need to support Symfony 5 first:

<table>
<tr><th>Status</th><th>Package</th><th>Installed version</th><th>Details</th></tr>
<tr><td>:white_check_mark:</td><td>behat/behat</td><td>v3.5.0</td><td><a href="https://github.com/Behat/Behat/pull/1256">Pull request</a></td></tr>
<tr><td>:white_check_mark:</td><td>behat/mink</td><td>1.7.x-dev</td><td><a href="https://github.com/minkphp/Mink/issues/787">Issue</a></td></tr>
<tr><td>:white_check_mark:</td><td>behat/mink-browserkit-driver</td><td>1.3.3</td><td><a href="https://github.com/minkphp/MinkBrowserKitDriver/issues/139">Issue</a></td></tr>
<tr><td>:white_check_mark:</td><td>behat/mink-extension</td><td>2.3.1</td><td><a href="https://github.com/Behat/MinkExtension/pull/355">Pull request</a></td></tr>
<tr><td>:x:</td><td>doctrine/doctrine-bundle</td><td>1.12.2</td><td>Supported in <a href="https://github.com/doctrine/DoctrineBundle/blob/2.0.0/composer.json">^2.0</a></td></tr>
<tr><td>:x:</td><td>doctrine/doctrine-cache-bundle</td><td>1.4.0</td><td>Deprecated, will be solved by updating doctrine/doctrine-bundle</td></tr>
<tr><td>:white_check_mark:</td><td>doctrine/doctrine-fixtures-bundle</td><td>v2.4.1</td><td>Supported in <a href="https://github.com/doctrine/DoctrineFixturesBundle/releases/tag/3.3.0">^3.3</a></td></tr>
<tr><td>:white_check_mark:</td><td>friends-of-behat/symfony-extension</td><td>v2.0.10</td><td><a href="https://github.com/FriendsOfBehat/SymfonyExtension/pull/100">Pull request</a></td></tr>
<tr><td>:white_check_mark:</td><td>friends-of-behat/variadic-extension</td><td>v1.2.0</td><td><a href="https://github.com/FriendsOfBehat/VariadicExtension/pull/27">Pull request</a></td></tr>
[...]

PR #10960, opened 2019-12-17
Handle pluralisation in translations with Intl ICU

pamil:
Symfony 5.0 removes `transchoice` filter and recommends using Intl ICU to handle pluralisation - https://symfony.com/doc/4.4/translation/message_format.html.

Referencing mega-issue #10928.

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/Cart/_widget.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/Cart/_widget.html.twig
@@ -5,9 +5,9 @@
 
     <i class="cart icon"></i>
     <span id="sylius-cart-total">
-        {{ money.convertAndFormat(cart.itemsTotal) }}
+        {{- money.convertAndFormat(cart.itemsTotal) -}}
     </span>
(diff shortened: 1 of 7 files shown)


src/Sylius/Bundle/ShopBundle/Resources/views/Cart/_widget.html.twig
```diff
@@ -5,9 +5,9 @@
 
     <i class="cart icon"></i>
     <span id="sylius-cart-total">
-        {{ money.convertAndFormat(cart.itemsTotal) }}
+        {{- money.convertAndFormat(cart.itemsTotal) -}}
     </span>
-    {% transchoice cart.items|length %}sylius.ui.item.choice{% endtranschoice %}
+    {{- 'sylius.ui.item.choice'|trans({'%count%': cart.items|length}) }}
```
lchrusciel: Why did you decide to trim this message only from one side?

pamil: It does the job of removing unnecessary space between price and the number of items - probably should be backported to Sylius 1.5 as a bugfix though.

It looked like "$10.00 , 1 item", now it's "$10.00, 1 item".

PR #10962, opened 2019-12-18
[Hot-Fix] Fix failing specs after prophecy update

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

It seems that these specs should never work 😄 The problem is, it's hard to specify setting a cloned object... especially if in one of this case we also have some value changed 💃 So this is obviously not a final fix, but at least it should fix a build 🚀


pamil:

PR #10969, opened 2019-12-18
Use Symfony CLI command instead of symfony/web-server-bundle

pamil:
Related to #10928.


Zales0123:
Thanks, Kamil! :tada:

PR #10971, opened 2019-12-18
Use Symfony CLI command to check security

pamil:
Related to #10928, replaces #10967.


Zales0123:
Thanks, Kamil! :tada:

PR #10974, opened 2019-12-19
Remove doctrine/doctrine-fixtures-bundle from dependencies and move doctrine/data-fixtures to dev dependencies

pamil:
Related to #10928. We don't use DoctrineFixturesBundle for a long time already and we only use DataFixtures in dev for purging the database.


GSadee:

PR #10984, opened 2019-12-23
Upgrade to Sonata Block Bundle ^4.0

pamil:
Related to #10928.


Zales0123:
As far as I see, `Sonata Block Bundle ^4.0` requires PHP7.3, so our tests on 7.2 fails :/ Is it even possible for us to support both this versions and still make progress with Symfony 5?

Zales0123:
Thank you, Kamil! :tada:

PR #10988, opened 2019-12-27
Template events system - refactoring & HTML debug comments

pamil:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | continues work from #10924
| License         | MIT

For the concept explanation, take a look at #10924.

A bit of refactoring:
 - extracted template block registry and a related DTO for template block data
 - extracted template event renderer
 - extracted template block renderer
 - removed logic from the Twig extension

As mentioned by @Prometee on our Slack, there are HTML comments added for every rendered block with some information about the block itself (if the block is a `*.html.twig` template and the `%kernel.debug%` is enabled):
Screenshot 2019-12-27 at 20 33 23


src/Sylius/Bundle/UiBundle/Registry/TemplateBlock.php
```diff
@@ -0,0 +1,61 @@
...
+        int $priority,
+        bool $enabled
+    ) {
+        $this->name = $name;
+        $this->template = $template;
+        $this->priority = $priority;
+        $this->enabled = $enabled;
+    }
+
+    public function name(): string
```
loevgaard: Is there a reason you didn't choose to add `get` to these methods. You know by convention or consistency or something :)

src/Sylius/Bundle/UiBundle/Renderer/TemplateBlockRenderer.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        $this->twig = $twig;
+        $this->debug = $debug;
+    }
+
+    public function render(string $eventName, TemplateBlock $templateBlock, array $context = []): string
+    {
+        $renderedBlock = '';
+
+        if ($this->debug && strrpos($templateBlock->template(), '.html.twig') !== false) {
```
loevgaard: Very nice

src/Sylius/Bundle/UiBundle/Renderer/TemplateEventRenderer.php
```diff
@@ -0,0 +1,43 @@
...
+        $this->templateBlockRenderer = $templateBlockRenderer;
+    }
+
+    public function render(string $eventName, array $context = []): string
+    {
+        $templateBlocks = $this->templateBlockRegistry->findEnabledForEvent($eventName);
+        $renderedTemplates = [];
+
+        foreach ($templateBlocks as $templateBlock) {
+            $renderedTemplates[] = $this->templateBlockRenderer->render($eventName, $templateBlock, $context);
```
loevgaard: Why not concatenate a string instead. My guess is you'd want a trailing new line anyways in the final source code.

So like:
`$str .= $this->templateBlockRenderer->render($eventName, $templateBlock, $context) . "\n";`

Prometee: Concat in this case could be faster https://stackoverflow.com/questions/4502654/php-many-concats-or-one-implode

src/Sylius/Bundle/UiBundle/Registry/TemplateBlock.php
pamil: It's the style for DTOs I'm used to in the internal projects, but I'm gonna change it for OS for consistency :+1:

src/Sylius/Bundle/UiBundle/Renderer/TemplateEventRenderer.php
pamil: I think we don't need trailing new line:

```twig
<div>
  {{ sylius_template_event('event') }}
</div>
```

With current implementation it will return:
```
<div>
  <p>First</p>
  <p>Second</p>
</div>
```

With trailing new line:
```
<div>
  <p>First</p>
  <p>Second</p>

</div>
```

pamil: As for the performance concerns, I think using implode is clearer than concatenation and right trimming last newline character - those comments are 9 years old, based on PHP 5 and still seem to be rather a microoptimisation.

Zales0123:
Thanks, Kamil! :tada:

PR #10989, opened 2019-12-27
Template events system - CLI command for debugging

pamil:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | related to #10924
| License         | MIT

Added `sylius:debug:template-event` command:
Screenshot 2019-12-27 at 17 49 07

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Command/DebugTemplateEventCommand.php
+++ b/src/Sylius/Bundle/UiBundle/Command/DebugTemplateEventCommand.php
@@ -1,0 +1,78 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\UiBundle\Command;
+
+use Sylius\Bundle\UiBundle\Registry\TemplateBlock;
+use Sylius\Bundle\UiBundle\Registry\TemplateBlockRegistryInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+final class DebugTemplateEventCommand extends Command
+{
+    protected static $defaultName = 'sylius:debug:template-event';
+
+    /** @var TemplateBlockRegistryInterface */
+    private $templateBlockRegistry;
+
+    public function __construct(TemplateBlockRegistryInterface $templateBlockRegistry)
+    {
+        parent::__construct();
+
+        $this->templateBlockRegistry = $templateBlockRegistry;
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setDescription('Debug template events and associated blocks')
+            ->addArgument('event', InputArgument::OPTIONAL, 'Template event name', null)
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $io = new SymfonyStyle($input, $output);
+        $eventName = $input->getArgument('event');
+
+        if ($eventName === null) {
+            $io->title('List of template events');
+
+            $io->listing(array_keys($this->templateBlockRegistry->all()));
+
+            return 0;
+        }
+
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/UiBundle/Resources/config/services/template_event.xml
```diff
@@ -26,5 +26,10 @@
             <argument type="service" id="Sylius\Bundle\UiBundle\Registry\TemplateBlockRegistryInterface" />
             <argument type="service" id="Sylius\Bundle\UiBundle\Renderer\TemplateBlockRendererInterface" />
         </service>
+
+        <service id="Sylius\Bundle\UiBundle\Command\DebugTemplateEventCommand">
+            <argument type="service" id="Sylius\Bundle\UiBundle\Registry\TemplateBlockRegistryInterface" />
+            <tag name="console.command" command="sylius:debug:template-event" />
```
loevgaard: Why put the command name here and not in the protected static property?

pamil: No hard feelings on this one - results are the same: https://symfony.com/doc/current/console/commands_as_services.html#lazy-loading.

loevgaard: Yes. I am probably more keen to put as much as I can in the code, so it's easier for people to understand it without having to find **both** the command and the service definition to figure out the command to run

src/Sylius/Bundle/UiBundle/Command/DebugTemplateEventCommand.php
```diff
@@ -0,0 +1,76 @@
...
+            );
+
+            return 0;
+        }
+
+        $io->title('List of template events');
+
+        $io->listing(array_keys($this->templateBlockRegistry->all()));
+
+        return 0;
```
Zales0123: I would this block in the if statement (with the returned condition, of course) -> less code indented

src/Sylius/Bundle/UiBundle/Command/DebugTemplateEventCommand.php
```diff
@@ -0,0 +1,76 @@
...
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+final class DebugTemplateEventCommand extends Command
+{
+    /** @var TemplateBlockRegistryInterface */
+    private $templateBlockRegistry;
+
+    public function __construct(TemplateBlockRegistryInterface $templateBlockRegistry)
+    {
+        parent::__construct();
```
Zales0123: Is it needed?

src/Sylius/Bundle/UiBundle/Command/DebugTemplateEventCommand.php
```diff
@@ -0,0 +1,76 @@
...
+                array_map(
+                    static function (TemplateBlock $templateBlock): array {
+                        return [
+                            $templateBlock->getName(),
+                            $templateBlock->getTemplate(),
+                            $templateBlock->getPriority(),
+                            $templateBlock->isEnabled() ? 'TRUE' : 'FALSE',
+                        ];
+                    },
+                    $this->templateBlockRegistry->all()[$eventName] ?? []
```
Zales0123: We already have a `findEnabledForEvent` method in a template block registry, how about putting some `findForEvent` as well there?

src/Sylius/Bundle/UiBundle/Resources/config/services/template_event.xml
pamil: Makes sense, I'll add it.

src/Sylius/Bundle/UiBundle/Command/DebugTemplateEventCommand.php
pamil: Yup :D

src/Sylius/Bundle/UiBundle/Command/DebugTemplateEventCommand.php
pamil: I did it before, but then I realised 2 of 3 methods in the registry (`findForEvent`, `findRegisteredEvents`) are used only for debugging so that's why I went with `all` method which allows for reusability but isn't a burden at the same time.

src/Sylius/Bundle/UiBundle/Command/DebugTemplateEventCommand.php
Zales0123: :(

PR #10991, opened 2019-12-27
Template events system - replace topbar with newly introduced events

pamil:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | related to #10924
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/config/app/config.yml
+++ b/src/Sylius/Bundle/ShopBundle/Resources/config/app/config.yml
@@ -54,3 +54,20 @@ liip_imagine:
 
 sonata_block:
     default_contexts: ~
+
+sylius_ui:
+    events:
+        sylius.shop.layout.topbar:
+            blocks:
+                before_switchers_legacy:
+                    template: "@SyliusShop/Block/Layout/_beforeSwitchersLegacy.html.twig"
+                currency_switcher:
+                    template: "@SyliusShop/Block/Layout/_currencySwitcher.html.twig"
+                locale_switcher:
+                    template: "@SyliusShop/Block/Layout/_localeSwitcher.html.twig"
+                before_security_widget_legacy:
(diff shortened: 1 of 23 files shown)


src/Sylius/Bundle/ShopBundle/Resources/views/Block/Layout/_afterSecurityWidgetLegacy.html.twig
```diff
@@ -0,0 +1 @@
+{{ sonata_block_render_event('sylius.shop.layout.after_security_widget') }}
```
pamil: I'm thinking about the way to generalise these calls, creating a template for every legacy function call seems like an overkill.

src/Sylius/Bundle/ShopBundle/Resources/config/app/config.yml
```diff
@@ -54,3 +54,20 @@ liip_imagine:
...
+sylius_ui:
+    events:
+        sylius.shop.layout.topbar:
+            blocks:
+                before_switchers_legacy:
+                    template: "@SyliusShop/Block/Layout/_beforeSwitchersLegacy.html.twig"
+                currency_switcher:
+                    template: "@SyliusShop/Block/Layout/_currencySwitcher.html.twig"
+                locale_switcher:
+                    template: "@SyliusShop/Block/Layout/_localeSwitcher.html.twig"
```
pamil: Need to figure out priorities for those blocks. Any ideas?

pamil: I used the following conventions:
 - last meaningful block starts at priority: 10
 - other meaningful blocks are having a bigger priority than the last one by +10
 - legacy Sonata events are +5 / -5 (before / after) from the related meaningful block

PR #10998, opened 2020-01-03
Expose payment and shipping state

loevgaard:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Adds two fields to the API response: `paymentState` and `shippingState`

### DIFF
--- a/src/Sylius/Bundle/AdminApiBundle/Resources/config/serializer/Model.Order.yml
+++ b/src/Sylius/Bundle/AdminApiBundle/Resources/config/serializer/Model.Order.yml
@@ -6,6 +6,14 @@ Sylius\Component\Core\Model\Order:
             expose: true
             type: string
             groups: [Default, Detailed, DetailedCart]
+        paymentState:
+            expose: true
+            type: string
+            groups: [Default, Detailed, DetailedCart]
+        shippingState:
+            expose: true
+            type: string
+            groups: [Default, Detailed, DetailedCart]
         currencyCode:
             expose: true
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/AdminApiBundle/Resources/config/serializer/Model.Order.yml
```diff
@@ -6,6 +6,14 @@ Sylius\Component\Core\Model\Order:
...
             type: string
             groups: [Default, Detailed, DetailedCart]
+        paymentState:
+            expose: true
+            type: string
+            groups: [Default, Detailed, DetailedCart]
+        shippingState:
+            expose: true
+            type: string
+            groups: [Default, Detailed, DetailedCart]
```
Zales0123: I wonder, should they be included in the `DetailedCart` group as well? When the order is not placed, they're always `'cart'`

loevgaard: I haven't followed up on the usage of the `DetailedCart` group, I just copied the `checkoutState` because I thought they would be relevant the same places as the `checkoutState`

loevgaard:
Yes, I will do that of course :)

loevgaard:
Do not stale

loevgaard:
Irrelevant when the new API comes

PR #10724, opened 2019-10-01
[Product] Fix bug related with pricing in product show page

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | 1.6
| Bug fix?        |yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Screenshot 2019-10-01 at 09 40 18

replaced by https://github.com/Sylius/Sylius/pull/10999

PR #10999, opened 2020-01-07
[Product] Add handling product pricing by channel

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | 1.6
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes 
| Related tickets | replacing #10724 
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Element/Product/ShowPage/VariantsElement.php
+++ b/src/Sylius/Behat/Element/Product/ShowPage/VariantsElement.php
@@ -30,6 +30,7 @@ final class VariantsElement extends Element implements VariantsElementInterface
         string $name,
         string $code,
         string $price,
+        string $channelName,
         string $currentStock
     ): bool {
         /** @var NodeElement $variantRow */
@@ -38,7 +39,7 @@ final class VariantsElement extends Element implements VariantsElementInterface
         /** @var NodeElement $variant */
         foreach ($variantRows as $variant) {
             if (
-                $this->hasProductWithGivenNameCodePriceAndCurrentStock($variant, $name, $code, $price, $currentStock)
+                $this->hasProductWithGivenNameCodePriceAndCurrentStock($variant, $name, $code, $price, $channelName, $currentStock)
             ) {
                 return true;
             }
@@ -52,6 +53,7 @@ final class VariantsElement extends Element implements VariantsElementInterface
         string $name,
         string $code,
         string $price,
+        string $channelName,
         string $currentStock
     ): bool {
         $variantContent = $variant->getParent()->find('css', sprintf(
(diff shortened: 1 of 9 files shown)


src/Sylius/Behat/Element/Product/ShowPage/VariantsElement.php
```diff
@@ -38,7 +39,7 @@ public function hasProductVariantWithCodePriceAndCurrentStock(
         /** @var NodeElement $variant */
         foreach ($variantRows as $variant) {
             if (
-                $this->hasProductWithGivenNameCodePriceAndCurrentStock($variant, $name, $code, $price, $currentStock)
+            $this->hasProductWithGivenNameCodePriceAndCurrentStock($variant, $name, $code, $price, $channelName, $currentStock)
```
Tomanhez:
```suggestion
                $this->hasProductWithGivenNameCodePriceAndCurrentStock($variant, $name, $code, $price, $channelName, $currentStock)
```
i think that this indentation should stay

composer.json
```diff
@@ -181,6 +181,7 @@
         "sylius/user-bundle": "self.version"
     },
     "conflict": {
+        "sylius/grid-bundle": "1.7.4",
```
Zales0123: Already there https://github.com/Sylius/Sylius/blob/1.6/composer.json#L185

src/Sylius/Bundle/AdminBundle/Twig/ChannelNameExtension.php
```diff
@@ -36,6 +39,8 @@ public function getFilters(): array
 
     public function getChannelNameByCode(string $code): string
     {
+        @trigger_error('Function getChannelNameByCode is deprecated since Sylius 1.7 and will be removed in Sylius 2.0.', \E_USER_DEPRECATED);
```
lchrusciel: If it will be merged to 1.6 then: 
```suggestion
        @trigger_error('Function getChannelNameByCode is deprecated since Sylius 1.6 and will be removed in Sylius 2.0.', \E_USER_DEPRECATED);
```

src/Sylius/Bundle/AdminBundle/Resources/views/Product/Show/_pricing.html.twig
```diff
@@ -11,19 +11,21 @@
...
-                    {% if channelPricing.originalPrice != null %}
-                        <td>{{ money.format(channelPricing.originalPrice, product.channels.first.baseCurrency) }}</td>
-                    {% else %}
-                        <td>-</td>
-                    {% endif %}
-                </tr>
-            {% endfor %}
+                {% set currencies = sylius_channels_currencies() %}
+                {% for channelPricing in product.variants.first.channelPricings %}
+                    {% set channel = channelPricing.channelCode|sylius_channel %}
```
pamil: This filter will always call the database for every channel, there is configured for a given variant.

I'd like to have a custom controller action which renders this template given a variant id and optimises these calls - this will also remove the need for the custom filter, which feels a bit hacky. We can achieve it by using `render(controller())` or `render(url())` instead of `include` in Twig (for this template).

src/Sylius/Bundle/AdminBundle/Resources/views/Product/Show/_variants.html.twig
```diff
@@ -123,17 +123,17 @@
                                                     </tr>
                                                     {% set currencies = sylius_channels_currencies() %}
                                                     {% for channelPricing in variant.channelPricings %}
+                                                        {% set channel = channelPricing.channelCode|sylius_channel %}
```
pamil: Same here - we can also think about merging both this template and the pricing one above.

PR #11018, opened 2020-01-14
Fix: Check PropertyPath value for add error to form

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.6
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | replaces #10689
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/OrderBundle/Controller/OrderItemController.php
+++ b/src/Sylius/Bundle/OrderBundle/Controller/OrderItemController.php
@@ -227,7 +227,11 @@ class OrderItemController extends ResourceController
     private function getAddToCartFormWithErrors(ConstraintViolationListInterface $errors, FormInterface $form): FormInterface
     {
         foreach ($errors as $error) {
-            $form->get('cartItem')->get($error->getPropertyPath())->addError(new FormError($error->getMessage()));
+            $formSelected = empty($error->getPropertyPath())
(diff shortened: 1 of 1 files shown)


Coosos:
@lchrusciel 
Last commit who remove is_null check is not included : https://github.com/Sylius/Sylius/pull/10689/commits/b4edb101c5298823fca9198dcbd114182027bade

lchrusciel:
When I tried to code one, I've found one scenario added already: https://github.com/Sylius/Sylius/blob/master/features/inventory/cart_inventory/prevent_buying_more_products_than_available_in_a_stock.feature#L26-L30

Also, tested manually and it works at the current master:
image

@Coosos can you check it as well?

Coosos:
@lchrusciel 

I've checked with demo version (1.5 and master) and I reproduced.
Example to reproduce (fresh installation) : 

In ``\Sylius\Bundle\InventoryBundle\Validator\Constraints\InStockValidator``, line 45, add (just after assert) :

```php
$this->context->addViolation(
   $constraint->message,
   ['%itemName%' => 'Example item']
);
```

And with the browser, test to add a product, no error is displayed

image

With the fix, the error is displayed :+1: 

image

In the case of the stock we have a front validation which protects, but in the case of a custom validation, that pose problem

PR #9498, opened 2018-06-18
Require PHP ^7.2 in Sylius ^1.3

pamil:
| Q               | A
| --------------- | -----
| Branch?         | 1.3
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

Sylius v1.3.0 will be released in late September, PHP 7.1 active support ends on 1st December and PHP 7.3 is about to come near the end of this year too.

PR #11021, opened 2020-01-14
Require PHP ^7.3

pamil:
Similar story to #9498. PHP 7.2 is not supported by PHP itself and PHP 7.4 was published a few months ago. Therefore, Sylius 1.7 released in February 2020 will support PHP ^7.3, dropping the maintenance burden.


lchrusciel:
Thank you, Kamil! :tada:

PR #11023, opened 2020-01-14
Fix doubled package in Composer's conflict section

pamil:


lchrusciel:
Thanks, Kamil! :tada:

PR #11028, opened 2020-01-15
Fix typo

vvasiloi:
| Q               | A
| --------------- | -----
| Branch?         | 1.4
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/Resources/config/services/action.xml
+++ b/src/Sylius/Bundle/PayumBundle/Resources/config/services/action.xml
@@ -29,7 +29,7 @@
         <service id="sylius.payum_action.offline.convert_payment" class="Sylius\Bundle\PayumBundle\Action\Offline\ConvertPaymentAction">
             <tag name="payum.action" factory="offline" alias="sylius.offline.convert_payment" />
         </service>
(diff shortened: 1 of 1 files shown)


Zales0123:
This is a duplicate of #9661... and unfortunately, as mentioned [there](https://github.com/Sylius/Sylius/pull/9661#discussion_r212303537) it's not possible to merge it because of BC policy :/

lchrusciel:
But what about adding an alias + depreciation about the usage of the old one?

vvasiloi:
My bad, didn't see the other one.
I was thinking it may be a BC break.
I was hopping that adding the `payum.action` tag with alias `sylius.offline.status` will replace the action, but unfortunately the service has to be overriden or aliased.
So, alias + deprecation will work.
I can do that if you reopen this PR and close the other one.

vvasiloi:
The question is, do I
1. deprecate the old service and alias the new one to it;
2. add the new service and alias the old one to it?

lchrusciel:
I would go for 2 with alias deprecation: https://symfony.com/blog/new-in-symfony-4-3-deprecating-service-aliases

PR #11033, opened 2020-01-16
[Checkout] Make billing address the default one

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/10575
| License         | MIT

It's the change we need and we deserve 😄 There are a lot of files changed, but it's a change that touches a lot of places in the code.

What needs to be considered - how to keep a BC, as we change the `differentBillingAddress` field to `differentShippingAddress`, which can affect customizations in some Sylius apps.

PR #11039, opened 2020-01-17
[Checkout] Add missing unit test for using deprecated different billing address flag

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | #11033
| License         | MIT

### DIFF
--- a/tests/Controller/CheckoutAddressingApiTest.php
+++ b/tests/Controller/CheckoutAddressingApiTest.php
@@ -100,6 +100,41 @@ EOT;
 
     /**
      * @test
+     *
+     * It is a test for deprecated behaviour since Sylius 1.7 and will be removed in Sylius 2.0
+     */
+    public function it_allows_to_address_order_with_the_same_shipping_and_billing_address_using_different_billing_address_flag(): void
+    {
+        $this->loadFixturesFromFile('authentication/api_administrator.yml');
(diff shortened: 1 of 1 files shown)


tests/Controller/CheckoutAddressingApiTest.php
```diff
@@ -98,6 +98,41 @@ public function it_allows_to_address_order_with_the_same_shipping_and_billing_ad
...
+                "postcode": "99-999"
+            },
+            "differentBillingAddress": false
+        }
+EOT;
+
+        $this->client->request('PUT', $this->getAddressingUrl($cartId), [], [], static::$authorizedHeaderWithContentType, $data);
+
+        $response = $this->client->getResponse();
+        $this->assertResponseCode($response, Response::HTTP_NO_CONTENT);
```
lchrusciel: Can we fetch a cart and ensure that both addresses are set? Same for billing address?

lchrusciel:
Thanks, Grzegorz! :tada:

tests/Controller/CheckoutAddressingApiTest.php
pamil: Yeah, it should be asserted as well, right now there's no clue what happens with addresses after they're processed.

PR #11048, opened 2020-01-20
Bug #10951 [Account] fix sync username with customer email

### LINKED ISSUE #10951: "Customer can login to the store" checkbox bug in the customer admin
**Sylius version affected**: 1.6.2

**Description**  
Even though customer has an account, the "user can login" checkbox is not checked.

**Steps to reproduce**  

1. Go to the [customer admin](https://demo.sylius.com/admin/customers/new)
2. Type: `first name`, `last name` and the `email`, DON'T check the "user can login" checkbox
3. Press save and wait for the page to reload
4. Check `user can login` checkbox, type a password and press save
5. Notice the checkbox is NOT checked

When you check it, the form disappears LOL ! This bug makes my clients go crazy.

Peek 2019-12-16 19-38

hatem20:
If applied, this commit will

-make sync behavior works on both customer and shop user,
resulting in username will never by setting username value whenever
shop user or customer is inserted or updated.

| Q               | A
| --------------- | -----
| Branch?         | 1.6 or master 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10951
| License         | MIT

I could have used the solution proposed here 
https://github.com/Sylius/Sylius/issues/10951#issuecomment-566190560
but it might be better if the username is always in sync with customer email

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/EventListener/DefaultUsernameORMListener.php
+++ b/src/Sylius/Bundle/CoreBundle/EventListener/DefaultUsernameORMListener.php
@@ -18,6 +18,7 @@ use Doctrine\ORM\Event\OnFlushEventArgs;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\UnitOfWork;
 use Sylius\Component\Core\Model\CustomerInterface;
+use Sylius\Component\Core\Model\ShopUserInterface;
 
 /**
  * Keeps user's username synchronized with email.
@@ -35,13 +36,20 @@ final class DefaultUsernameORMListener
 
     private function processEntities(array $entities, EntityManagerInterface $entityManager, UnitOfWork $unitOfWork): void
     {
-        foreach ($entities as $customer) {
-            if (!$customer instanceof CustomerInterface) {
+        foreach ($entities as $entity) {
+            if (!$entity instanceof ShopUserInterface && !$entity instanceof CustomerInterface) {
                 continue;
             }
 
-            $user = $customer->getUser();
-            if (null === $user) {
+            if ($entity instanceof ShopUserInterface) {
+                $user = $entity;
+                $customer = $user->getCustomer();
+            } else {
+                $customer = $entity;
+                $user = $customer->getUser();
+            }
+
+            if (!$customer || !$user) {
                 continue;
             }
 
--- a/features/user/managing_customers/adding_customer_account.feature
+++ b/features/user/managing_customers/adding_customer_account.feature
@@ -26,5 +26,6 @@ Feature: Adding a new customer account
         And I specify their password as "killSauron"
         And I save my changes
         Then I should be notified that it has been successfully edited
+        And I should not see create account option
         And the customer "f.baggins@example.com" should appear in the store
         And this customer should have an account created


src/Sylius/Bundle/CoreBundle/EventListener/DefaultUsernameORMListener.php
```diff
@@ -35,13 +36,20 @@ public function onFlush(OnFlushEventArgs $onFlushEventArgs)
 
     private function processEntities(array $entities, EntityManagerInterface $entityManager, UnitOfWork $unitOfWork): void
     {
-        foreach ($entities as $customer) {
-            if (!$customer instanceof CustomerInterface) {
+        foreach ($entities as $entity) {
+            if (!$entity instanceof ShopUserInterface && !$entity instanceof CustomerInterface) {
```
lchrusciel: Can you maybe split this condition into two separate ones? The logic below could be extracted to private method with two arguments I suppose (`user` + `customer`), it should increase readability. Also, can you update specs?
```suggestion
            if ($entity instanceof ShopUserInterface) {
                $this->updateUserName($entity, $entity->getCustomer());
            }
            
            if ($entity instanceof CustomerInterface) {
                $this->updateUserName($entity->getUser(), $entity);
            }
```

hatem20: @lchrusciel 
so you think this is more readable ?
as i should also pass `$entityManager` and `$unitOfWork` to updateUserName
in addition to @var tag to bypass phpStan
```
private function processEntities(array $entities, EntityManagerInterface $entityManager, UnitOfWork $unitOfWork): void
    {
        foreach ($entities as $entity) {
            if ($entity instanceof ShopUserInterface) {
                /** @var CustomerInterface $customer */
                $customer = $entity->getCustomer();
                $this->updateUsername($entity, $customer, $entityManager, $unitOfWork);
            } elseif($entity instanceof CustomerInterface) {
                /** @var ShopUserInterface $user */
                $user = $entity->getUser();
                $this->updateUsername($user, $entity, $entityManager, $unitOfWork);
            }
        }
    }

private function updateUsername(?ShopUserInterface $user, ?CustomerInterface $customer, EntityManagerInterface $entityManager, UnitOfWork $unitOfWork): void
    {
        if (!$user || !$customer) {
            return;
        }

        if ($customer->getEmail() === $user->getUsername() && $customer->getEmailCanonical() === $user->getUsernameCanonical()) {
            return;
        }

        $user->setUsername($customer->getEmail());
        $user->setUsernameCanonical($customer->getEmailCanonical());

        /** @var ClassMetadata $userMetadata */
        $userMetadata = $entityManager->getClassMetadata(get_class($user));
        $unitOfWork->recomputeSingleEntityChangeSet($userMetadata, $user);
    }
```

lchrusciel: Yes, I would say, that it is easier to follow this logic because we have fewer conditions here. Don't you agree?

src/Sylius/Bundle/CoreBundle/EventListener/DefaultUsernameORMListener.php
```diff
@@ -35,26 +36,34 @@ public function onFlush(OnFlushEventArgs $onFlushEventArgs)
...
     {
-        foreach ($entities as $customer) {
-            if (!$customer instanceof CustomerInterface) {
-                continue;
+        foreach ($entities as $entity) {
+            if ($entity instanceof ShopUserInterface) {
+                /** @var CustomerInterface $customer */
+                $customer = $entity->getCustomer();
+                $this->updateUsername($entity, $customer, $entityManager, $unitOfWork);
+            } elseif ($entity instanceof CustomerInterface) {
```
lchrusciel:
```suggestion
                 continue;   
            } 

            if ($entity instanceof CustomerInterface) {
```

lchrusciel:
Hey @hatem20. The build is red and it seems, that it may be related to your PR. Can you rebase to the newest master and fix the build if it will not resolve all issues?

hatem20:
@lchrusciel failed again, and not related to PR

lchrusciel:
Thanks, Hatem! :1st_place_medal:

lchrusciel:
Hey Hatem. Can you append your work with spec class? Your PR is not working as expected: https://master.demo.sylius.com/admin/customers/ however it seems, that what you have proposed is a useful fix. I would prefer to have at least spec for that. AFAIR, we are not allowing for shop user email update, so you won't be able to test it with behat.

If you will not cover it with spec, I will have to revert it before the nearest tag.

hatem20:
@lchrusciel sure will add spec thanks

lchrusciel:
yes, please :)

hatem20:
@lchrusciel I added spec as requested, but the PR is working as expected,
I think you are testing it against master while the PR is merged to 1.7 branch.
also, behat scenario is added.
I hope you correct me if I miss something

lchrusciel:
It seems, that you've spotted my mistake! I didn't up merge changes to the master, my fault. Nonetheless, the spec is a good addition to your fix. Thanks a lot for your work ;)

PR #11052, opened 2020-01-22
Disabled all buttons from the column after click

kulczy:
Nagranie z ekranu 2020-01-22 o 12 49 32

works the same on all index pages with a "transition" buttons (payments, shipments, product reviews...)


pamil:
Thanks, Szymon! :tada:

PR #11057, opened 2020-01-23
Move "php-http/guzzle6-adapter" to "require-dev"

pamil:
It is required only by "hwi/oauth-bundle" which is in "require-dev" as well.


lchrusciel:
Shouldn't we merge it to 1.6? I would say it is a bug fix

lchrusciel:
Thank you, Kamil! :tada:

Issue #10997, opened 2020-01-02
New template events architecture


pamil:
_Extracted the following description from #10924 - let's keep track of this feature in this issue._

We aim to replace our current template events system using Sonata Block bundle and improve both the developer experience using it and Sylius extendability, especially in the light of [Sylius Store](https://sylius.com/plugins/) which has been recently opened.

## Keeping track of changes

 - [x] Initial implementation (started with #10924, refactored in #10988, added more functionality in #10991, refactored further in #11036, and another in #11059)
 - [x] Replace existing `sonata_block_render_event` method calls with `sylius_template_event`(started with topbar in #10991, continued with homepage in #11014)
    - [x] Register blocks that will keep backwards compatibility and call `sonata_block_render_event`
    - [x] Merge existing `*.before_something` and `*.after_something` events into `*.something` and use them for core functionalities as well (so that one can add a block in the middle of core ones or disable it)
 - [x] Write more extensive documentation for template events
 - [x] Consider creating [ADR (Architecture Decision Record)](https://github.com/joelparkerhenderson/architecture_decision_record) detailing the reasoning, alternatives and pros/cons of this solution
 - [x] Create _best practices_ guide (what priorities to use in your application / your plugin) so that Sylius ecosystem can adopt it without struggles
   - Initial convention: https://github.com/Sylius/Sylius/pull/10991/files#r362234318
 - [x] Increase discoverability of events and blocks:
    - [x] Debugging commands for CLI (show all registered events, show all blocks for an event) (#10989)
    - [x] HTML comments with event / block names before rendered content (#10988, #11036)
    - [x] Symfony Profiler / Data Collector (#10990)
 - [x] Create issues in downstream repositories making use of current events implementation

## Current implementation

 - We use Sonata Block bundle to create extension points in Sylius, which are recommended to be used by our plugins.

 - An extension point is created by calling `sonata_block_render_event` Twig function:

    ```twig
    {{ sonata_block_render_event('sylius.admin.order.show.before_summary', {'resource': resource}) }}

    <div class="ui segment">
        {% include '@SyliusAdmin/Order/Show/_summary.html.twig' %}
    </div>

    {{ sonata_block_render_event('sylius.admin.order.show.after_summary', {'resource': resource}) }}
    ``` 

 - A block can be injected by creating a template and registering a event listener in DIC:

    ```yaml
    services:
        my.service.name.rendering.block:
            class: Sylius\Bundle\UiBundle\Block\BlockEventListener
[...]

PR #11059, opened 2020-01-23
Rendering multiple dependent template events at once + admin-side CRUD events

pamil:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | a bit
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | -
| Related tickets | related to #10997 
| License         | MIT

_Please review the code commit by commit, it'll be much easier._

Lots of the code added in the Sylius UI bundle configuration is to provide backwards compatibility with Sonata blocks.

<hr/>

This PR adds an ability to render multiple dependent template events at once. What does it mean?

Apart from usual call to render template event using `sylius_template_event('event')`, it is now possible to pass an array with two or more events using `sylius_template_event('first_event', 'second_event')`.

The call mentioned above will merge blocks defined in both `first_event` and `second_event` and render them all. In case the same block is defined, the first event's block configuration overrides the second event's block configuration. 

<hr/>

### Example: Admin CRUD template events

This allows defining two events for generic templates - one event which name is variable, and another one which name is constant. The following example comes from generic CRUD templates in AdminBundle, which was modified in this PR as well:

```twig
{% set event_prefix = metadata.applicationName ~ '.admin.' ~ metadata.name ~ '.create' %}

{{ sylius_template_event([event_prefix, 'sylius.admin.create'], _context) }}
```

In this case, you can manage blocks for all `create` pages:

```yaml
sylius_ui:
  events:
    sylius.admin.create:
      blocks:
        new_block: template.html.twig
```

Or manage blocks only for product create page:

```yaml
sylius_ui:
  events:
    sylius.admin.product.create:
      blocks:
        new_block: template.html.twig
```

Screenshot 2020-01-23 at 23 51 24

<hr/>

### Example: Equivalent configuration (one event / multiple events)

Given the following basic configuration:

```yaml
# First config file
sylius_ui:
  events:
    event:
      blocks:
        block:
          template: template.html.twig
          priority: 10
          context:
            foo: bar

# Second config file
sylius_ui:
  events:
    event:
      blocks:
        block:
          priority: -5
```

```twig
{{ sylius_template_event('event') }}
```

The equivalent configuration using dependent template events:

```yaml
sylius_ui:
  events:
    first_event:
      blocks:
        block:
          template: template.html.twig
          priority: 10
          context:
            foo: bar
    second_event:
      blocks:
        block:
          priority: -5
```

```twig
{{ sylius_template_event(['second_event', 'first_event']) }}
```

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/views/DataCollector/templateBlock.html.twig
+++ b/src/Sylius/Bundle/UiBundle/Resources/views/DataCollector/templateBlock.html.twig
@@ -66,12 +66,19 @@
 
             {% for event in collector.renderedEvents %}
                 <tr>
-                    <td>{{ event.name }}</td>
+                    <td>
+                        {% for name in event.names %}
+                            {% if loop.first %}<strong>{% endif %}
+                            {{ name }}{% if not loop.last %}, {% endif %}
+                            {% if loop.first %}</strong>{% endif %}
+                        {% endfor %}
+                    </td>
                     <td>{{ '%.0f'|format(event.time * 1000) }}ms</td>
                     <td>
                         {% if event.blocks|length > 0 %}
                             <table>
                                 <tr>
+                                    {% if event.names|length > 1 %}<th>Origin event</th>{% endif %}
                                     <th>Name</th>
                                     <th>Template</th>
                                     <th>Duration</th>
@@ -79,6 +86,7 @@
                                 </tr>
                                 {% for block in event.blocks %}
                                     <tr>
+                                        {% if event.names|length > 1 %}<td>{{ block.definition.eventName }}</td>{% endif %}
                                         <td>{{ block.definition.name }}</td>
                                         <td>{{ block.definition.template }}</td>
                                         <td>{{ '%.0f'|format(block.time * 1000) }}ms</td>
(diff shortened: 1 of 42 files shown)


src/Sylius/Bundle/UiBundle/Resources/views/DataCollector/templateBlock.html.twig
```diff
@@ -66,19 +66,27 @@
...
+                            {{ name }}{% if not loop.last %}, {% endif %}
+                            {% if loop.first %}</strong>{% endif %}
+                        {% endfor %}
+                    </td>
                     <td>{{ '%.0f'|format(event.time * 1000) }}ms</td>
                     <td>
                         {% if event.blocks|length > 0 %}
                             <table>
                                 <tr>
+                                    {% if event.names|length > 1 %}<th>Origin event</th>{% endif %}
```
pamil: Origin event is only needed if there are multiple events rendered at the same time.

src/Sylius/Bundle/UiBundle/Tests/Functional/TemplateEventTest.php
```diff
@@ -34,7 +34,7 @@ public function it_renders_template_events_blocks(): void
     {
         // See Kernel.php for the configuration resulting in those lines
         $expectedLines = [
-            'First block (for event: "first_event")',
```
pamil: Removed `_event` variable - it wasn't used anywhere but in the tests and after introducing multiple events it would be quite misleading.

lchrusciel:

PR #11065, opened 2020-01-26
user provider fix exception

### LINKED ISSUE #10879: [BUG] UserProvider should throw UnsupportedUserException and not UnexpectedValueException
**Sylius version affected**: 1.6.2

**Description**  
When using different user providers with multiple firewall contexts and some are outside the Sylius-User Provider scope then a `refreshUser` can throw (depending on which UserProviders are registered first and if they provide a user or not) an UnexpectedValueException as Symfony checks all UserProviders when calling `refreshUser`.

If the user is not supported, Symfony expects a `UnsupportedUserException` where they would try the next provider and they throw their own exception if no provider supports the user.
See also: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Http/Firewall/ContextListener.php#L174

**Steps to reproduce**  
see above.

**Possible Solution**  
Instead of `Assert::instanceOf` at https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/UserBundle/Provider/AbstractUserProvider.php#L71
(refreshUser of the AbstractUserProvider).
Do an instanceof check and throw an `\Symfony\Component\Security\Core\Exception\UnsupportedUserException` instead.

oallain:
| Q               | A
| --------------- | -----
| Branch?         | 1.5 and more
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10879 
| License         | MIT

UserProvider should throw UnsupportedUserException and not UnexpectedValueException.

### DIFF
--- a/src/Sylius/Bundle/UserBundle/Provider/AbstractUserProvider.php
+++ b/src/Sylius/Bundle/UserBundle/Provider/AbstractUserProvider.php
@@ -19,7 +19,6 @@ use Sylius\Component\User\Repository\UserRepositoryInterface;
 use Symfony\Component\Security\Core\Exception\UnsupportedUserException;
 use Symfony\Component\Security\Core\Exception\UsernameNotFoundException;
 use Symfony\Component\Security\Core\User\UserInterface;
-use Webmozart\Assert\Assert;
 
 abstract class AbstractUserProvider implements UserProviderInterface
 {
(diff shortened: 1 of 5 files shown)


src/Sylius/Bundle/UserBundle/Provider/AbstractUserProvider.php
```diff
@@ -67,8 +66,11 @@ public function loadUserByUsername($username): UserInterface
      */
     public function refreshUser(UserInterface $user): UserInterface
     {
-        /** @var SyliusUserInterface $user */
-        Assert::isInstanceOf($user, SyliusUserInterface::class);
+        if(!$user instanceof SyliusUserInterface) {
```
lchrusciel: I'm wondering if it shouldn't be part of `supportsClass` function

oallain: I don't believe, `supportsClass` does not have the same purpose.

lchrusciel:
Should be targeted to 1.6 I suppose. 

We've dropped support for 1.5

Zales0123:
may become messed-up.**

execute `git fetch origin && git reset "user-provider-fix-exception"` to update your local branch.
 
lchrusciel:
Thanks, @oallain! :tada:

PR #11075, opened 2020-01-28
Fix testing migrations

loic425:
| Q               | A
| --------------- | -----
| Branch?         | master (1.7)
| Bug fix?        | yes (CI only)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

### DIFF
--- a/etc/travis/suites/application/script.sh
+++ b/etc/travis/suites/application/script.sh
@@ -10,6 +10,7 @@ commands=(
     validate-twig
     validate-yaml-files
     validate-yarn-packages
+    test-doctrine-migrations
     test-phpstan
     test-psalm
     test-phpspec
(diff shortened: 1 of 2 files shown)


Zales0123:
👍 but the build failed, so probably there is something wrong with the migrations... can you take a look at it @loic425?

loic425:
@Zales0123 it proves the changes work :smile: 
I'll have a look soon.

Zales0123:
Indeed 😄 Thank you!

loic425:
@Zales0123 I check it out. This is not really a migration problem cause database cannot be reverted until first version due to data unique on sylius product translation unique index (slug, locale).

I think we have to empty database before or add this test before testing fixtures or behat features.

loic425:
Empty database could hide some problems... But the only other solution is to remove duplicated to match the old unique key and this is not acceptable too...

pamil:
Thanks, Loïc! :tada:

PR #11064, opened 2020-01-26
fix code generation with too long code length

oallain:
| Q               | A
| --------------- | -----
| Branch?         | 1.6 and after
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10912 
| License         | MIT

Fix error when amount of code combination is greater than the value of PHP_INT_MAX.

PR #11095, opened 2020-02-05
[Promotion] Generating long coupons

### LINKED ISSUE #10912: [Promotion] Unable to generate coupons

**Description**  
```
Invalid coupons code length or coupons amount. It is not possible to generate 50 unique coupons with code length equals 16. Possible generate amount is -9223372036854775808.
```

**Steps to reproduce**  
See screenshot: https://www.dropbox.com/s/x4fbw8wohwnweom/Screenshot%202019-12-09%2019.28.31.png?dl=0

Another one: https://www.dropbox.com/s/na4xbazvadlpz7d/Screenshot%202019-12-09%2019.33.07.png?dl=0

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.6
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10912, based on #11064 and #11097
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Promotion/Generator/PercentageGenerationPolicy.php
+++ b/src/Sylius/Component/Promotion/Generator/PercentageGenerationPolicy.php
@@ -68,11 +68,11 @@ final class PercentageGenerationPolicy implements GenerationPolicyInterface
             $instruction->getSuffix()
         );
 
-        $codeCombination = 16 ** $expectedCodeLength;
-        if ($codeCombination > PHP_INT_MAX) {
-            $codeCombination = 0;
+        $codeCombination = 16 ** $expectedCodeLength * $this->ratio;
+        if ($codeCombination >= PHP_INT_MAX) {
+            return PHP_INT_MAX - $generatedAmount;
         }
 
-        return (int) floor($codeCombination * $this->ratio - $generatedAmount);
+        return (int) $codeCombination - $generatedAmount;
     }
 }
--- a/src/Sylius/Component/Promotion/Generator/PercentageGenerationPolicy.php
+++ b/src/Sylius/Component/Promotion/Generator/PercentageGenerationPolicy.php
@@ -68,6 +68,11 @@ final class PercentageGenerationPolicy implements GenerationPolicyInterface
             $instruction->getSuffix()
         );
 
-        return (int) floor((16 ** $expectedCodeLength) * $this->ratio - $generatedAmount);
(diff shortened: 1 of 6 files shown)


src/Sylius/Component/Promotion/Generator/PercentageGenerationPolicy.php
```diff
@@ -68,6 +68,11 @@ private function calculatePossibleGenerationAmount(PromotionCouponGeneratorInstr
             $instruction->getSuffix()
         );
 
-        return (int) floor((16 ** $expectedCodeLength) * $this->ratio - $generatedAmount);
+        $codeCombination = 16 ** $expectedCodeLength * $this->ratio;
+        if ($codeCombination >= PHP_INT_MAX) {
+            return PHP_INT_MAX - $generatedAmount;
+        }
+
+        return (int) $codeCombination - $generatedAmount;
```
GSadee: Example:
When: `$expectedAmount = 20;`
Then: `$codeCombination = {float} 6.0446290980731E+23;`
And: `$codeCombination >= PHP_INT_MAX` is true
So the value `PHP_INT_MAX - $generatedAmount` will be returned

IMO this fix (without changing current behaviour and without BC breaks) makes sense, but WDYT /cc @Sylius/core-team ? 😃

lchrusciel: In PromotionCouponGenerator we are allowing for coupons up to 40 characters. If we can process with calculation, maybe we should increase this amount? 

Also, how much of precision will we lost when we will convert to float?

GSadee: The problem is that we cannot convert to float because of BC Policy, we need to return here an integer value, so `PHP_INT_MAX` is the max value that could be returned. IMO this fix resolves the bug, we are able to generate coupons even for 40 characters, but there will be a limit to `PHP_INT_MAX` coupons that can be generated. Increasing this value could be also a problem for form (`PromotionCouponGeneratorInstructionType`) where is an integer field for the amount.

pamil: Thanks for the explanation!

pamil: Thanks for the explanation!

pamil:
Thank you, Grzegorz! :tada:

PR #11104, opened 2020-02-07
[Behat] Rework testing customer orders to use test attributes

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Page/Shop/Account/Order/IndexPage.php
+++ b/src/Sylius/Behat/Page/Shop/Account/Order/IndexPage.php
@@ -71,14 +71,14 @@ class IndexPage extends SymfonyPage implements IndexPageInterface
 
     public function openLastOrderPage(): void
     {
-        $this->getElement('last_order')->click();
+        $this->getElement('last_order')->clickLink('Show');
     }
 
     protected function getDefinedElements(): array
     {
         return array_merge(parent::getDefinedElements(), [
-            'customer_orders' => 'table',
-            'last_order' => 'table tbody tr:last-child a:contains("Show")',
+            'customer_orders' => '[data-test-grid-table]',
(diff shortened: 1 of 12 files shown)


src/Sylius/Behat/Page/Shop/Account/Order/IndexPage.php
```diff
@@ -86,14 +69,16 @@ public function isItPossibleToChangePaymentMethodForOrder(OrderInterface $order)
...
+        $this->getElement('last_order')->click();
+    }
+
     protected function getDefinedElements(): array
     {
         return array_merge(parent::getDefinedElements(), [
-            'customer_orders' => 'table',
-            'last_order' => 'table tbody tr:last-child a:contains("Show")',
+            'customer_orders' => '[data-test-grid-table]',
+            'last_order' => '[data-test-grid-table-body] tr:last-child a:contains("Show")',
```
lchrusciel: /cc @kulczy can we do it better?

kulczy: we can find the last element in the loop and add a new attribute, it seems a bit better solution

Tomanhez: Maybe we should add `{ sylius_test_html_attribute('grid-table-row') }` in: https://github.com/Sylius/Sylius/blob/ae812d3fc91f34fae353e106e68c65a70c089f80/src/Sylius/Bundle/UiBundle/Resources/views/Macro/table.html.twig#L25 and replace `tr` to `['data-test-grid-table-row']`
It would be more consistent with other changes

lchrusciel:

PR #10986, opened 2019-12-25
WIP: fix(entity): make ShopBillingData extendable

Gounlaf:
| Q               | A
| --------------- | -----
| Branch?         | 1.5
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10563
| License         | MIT

PR #11108, opened 2020-02-07
Make ShopBillingData a resource

### LINKED ISSUE #10563: Cannot override ShopBillingData as it's not a mapped-superclass
Map ShopBillingData to be a MappedSuperclass as any other entity, this way we can override its properties.

I tried to override both Channel and ShopBIllingData xml mapping, but I couldn't get the bundle ShopBillingData not to create its own table, so that I had two shop_billing_data tables in the db.

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10563, replaces #10986
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/DependencyInjection/Configuration.php
+++ b/src/Sylius/Bundle/CoreBundle/DependencyInjection/Configuration.php
@@ -16,6 +16,7 @@ namespace Sylius\Bundle\CoreBundle\DependencyInjection;
 use Sylius\Bundle\CoreBundle\Controller\ProductTaxonController;
 use Sylius\Bundle\CoreBundle\Doctrine\ORM\AvatarImageRepository;
 use Sylius\Bundle\CoreBundle\Form\Type\Product\ChannelPricingType;
+use Sylius\Bundle\CoreBundle\Form\Type\ShopBillingDataType;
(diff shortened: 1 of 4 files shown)


app/migrations/Version20200207131742.php
```diff
@@ -0,0 +1,36 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20200207131742 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+
+        $this->addSql('ALTER TABLE sylius_channel DROP FOREIGN KEY FK_16C8119EB5282EDF');
+        $this->addSql('DROP TABLE sylius_shop_billing_data');
```
Zales0123: I believe we don't want to drop `sylius_shop_billing_data` table 😄 Probably the table will not be kept if we don't register a `ShopBillingData` as a resource

Zales0123:

loevgaard:
Very nice!

PR #11111, opened 2020-02-10
[Docs] Recommend trait usage in Plugin Development Guide

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.6
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

For a long time already, we provide all entities [overridden in Sylius-Standard](https://github.com/Sylius/Sylius-Standard/tree/master/src/Entity). This approach enforces annotations mapping by default but allows Sylius to use multiple plugins extending the same entity at the same time. 

We should definitely recommend this approach in our **Plugin Development Guide**.

### DIFF
--- a/docs/plugin-development-guide/implementation.rst
+++ b/docs/plugin-development-guide/implementation.rst
@@ -16,59 +16,27 @@ take a look at :doc:`customizing models</customization/model>`, :doc:`form</cust
 Model
 *****
 
-The only field we need to add is an additional ``$availableOnDemand`` boolean. We should start with the unit tests (written with
-PHPSpec, PHPUnit, or any other unit testing tool):
+The only field we need to add is an additional ``$availableOnDemand`` boolean. To allow plugin's user to use multiple plugins extending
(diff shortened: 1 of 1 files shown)


JoppeDC:
Thanks for the further information @Zales0123 , this definetly clears up some misunderstanding.

lchrusciel:
PS. Nice PR number 🎉

docs/plugin-development-guide/implementation.rst
```diff
@@ -86,54 +54,23 @@ PHPSpec, PHPUnit, or any other unit testing tool):
...
-            availableOnDemand:
-                type: boolean
-
-Then our new entity should be configured as a resource model:
-
-.. code-block:: yaml
-
-    # src/Resources/config/config.yml
-
-    sylius_product:
```
lchrusciel: What do you think about recommending this configuration in the test folder, so the entity will be checked if wired up properly (and preferably will be a good introduction for testing)

lchrusciel:
Thank you, Mateusz! :tada:

PR #11114, opened 2020-02-11
Fixed wrongly calculated order-item subtotal

### LINKED ISSUE #11112: [Admin] Order Item Subtotal wrongly calculated in order view
**Sylius version affected**: 1.6.*

**Description**  
When an order has order-level and order-item(-unit) level promotion adjustments applied to it, that end up having an uneven adjustment amount per unit, the current `@SyliusAdminBundle/Resources/views/Order/Show/Summary/_item.html.twig` template displays the wrong subtotal.

**Steps to reproduce**  
Create two promotions.
1) No rules, fixed order discount 10 *currency*
2) No rules, fixed item discount 5 *currency*

Order something with an uneven amount of items. On my system, I had one item 5 times and a second one 2 times in my order. The **Distributed order discount** column shows one amount (which should be two, because it is distributed unevenly over the item units).

This can also be checked in the database with this query:

```sql
SELECT label, amount FROM sylius_adjustment WHERE order_item_unit_id IN (SELECT id FROM sylius_order_item_unit soiu WHERE order_item_id IN (your_two, item_ids_from_your_order));
```
Check for all amounts that correspond to your previously setup promotions and you should see they are uneven (this is just to confirm, uneven adjustments are correct here. F*ck rounding issues :wink:)

**Possible Solution**  
In the mentioned template, line 11 is the culprit.

```twig
{% set subtotal = item.quantity * (item.unitPrice + item.units.first.adjustmentsTotal(unitPromotionAdjustment) + item.units.first.adjustmentsTotal(orderPromotionAdjustment)) %}
```
This only takes the first units' promotion adjustments into account when calculating the item subtotal.

A proposed fix (with some annotations from myself :wink: ) could look something like this:

```twig
{# This does not respect uneven distributed discounts from order-level #}
{# {% set subtotal = item.quantity * (item.unitPrice + item.units.first.adjustmentsTotal(unitPromotionAdjustment) + item.units.first.adjustmentsTotal(orderPromotionAdjustment)) %} #}

{% set aggregatedUnitPromotionAdjustments = 0 %}
{% for unit in item.units %}
    {% set aggregatedUnitPromotionAdjustments = aggregatedUnitPromotionAdjustments + unit.adjustmentsTotal(unitPromotionAdjustment) + unit.adjustmentsTotal(orderPromotionAdjustment) %}
{% endfor %}
{% set subtotal = (item.unitPrice * item.quantity) + aggregatedUnitPromotionAdjustments %}

{# Why not just use the item total here? #}
{# {% set subtotal = item.total %} #}
```

peterukena:
Since adjustments can have different value, picking just the first one may yield wrong results. Hence, I use the recursively fetched adjustments. Also, I added a ~ before the distributed amount to signify its variance. See #11112 for more details

| Q               | A
| --------------- | -----
| Branch?         | 1.6 or master 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no/
| Related tickets | fixes #11112
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/Summary/_item.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/Summary/_item.html.twig
@@ -25,7 +25,7 @@
         {{ money.format(item.units.first.adjustmentsTotal(unitPromotionAdjustment), order.currencyCode) }}
     </td>
     <td class="right aligned unit-order-discount">
-        <span style="font-style: italic;">~{{ money.format(item.units.first.adjustmentsTotal(orderPromotionAdjustment), order.currencyCode) }}</span>
+        <span style="font-style: italic;">~ {{ money.format(item.units.first.adjustmentsTotal(orderPromotionAdjustment), order.currencyCode) }}</span>
     </td>
     <td class="right aligned discounted-unit-price">
         {{ money.format(item.fullDiscountedUnitPrice, order.currencyCode) }}
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/Summary/_item.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/Summary/_item.html.twig
@@ -7,7 +7,9 @@
 
 {% set variant = item.variant %}
 {% set product = variant.product %}
-{% set subtotal = item.quantity * (item.unitPrice + item.units.first.adjustmentsTotal(unitPromotionAdjustment) + item.units.first.adjustmentsTotal(orderPromotionAdjustment)) %}
+
+{% set aggregatedUnitPromotionAdjustments = item.getAdjustmentsTotalRecursively(unitPromotionAdjustment) + item.getAdjustmentsTotalRecursively(orderPromotionAdjustment) %}
+{% set subtotal = (item.unitPrice * item.quantity) + aggregatedUnitPromotionAdjustments %}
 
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/Summary/_item.html.twig
```diff
@@ -23,7 +25,7 @@
         {{ money.format(item.units.first.adjustmentsTotal(unitPromotionAdjustment), order.currencyCode) }}
     </td>
     <td class="right aligned unit-order-discount">
-        <span style="font-style: italic;">{{ money.format(item.units.first.adjustmentsTotal(orderPromotionAdjustment), order.currencyCode) }}</span>
+        <span style="font-style: italic;">~{{ money.format(item.units.first.adjustmentsTotal(orderPromotionAdjustment), order.currencyCode) }}</span>
```
GSadee: This value is already displayed with `font-style: italic;` and I guess that adding `~` should be reverted

peterukena: I talked to @CoderMaggie - she thinks its a nice way to signify the variance this specific value might have. You can see [here](https://github.com/Sylius/Sylius/issues/11112#issuecomment-584628423), why. I can fix the tests by adding the ~ to the `ShowPage` so it gets trimmed away nicely.

Or I can remove it - your call, I am not opposed to any solution :slightly_smiling_face:

GSadee:
Thanks, Peter! :tada:

PR #11125, opened 2020-02-17
Add metatags event in layout's block

maximehuran:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | no
| License         | MIT

Add event to be able to append some data in metatags block.

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/layout.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/layout.html.twig
@@ -10,6 +10,7 @@
     <meta content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no" name="viewport">
 
     {% block metatags %}
+        {{ sonata_block_render_event('sylius.shop.layout.metatags') }}
     {% endblock %}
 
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/ShopBundle/Resources/views/layout.html.twig
```diff
@@ -10,6 +10,7 @@
     <meta content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no" name="viewport">
 
     {% block metatags %}
+        {{ sonata_block_render_event('sylius.shop.layout.metatags') }}
```
lchrusciel: This should rather be part of `{{ sylius_template_event('sylius.shop.layout') }}` I suppose /cc @pamil

Tomanhez: Its true, we replace old `sonata_block_render_event` to `sylius_template_event`, look in this PR : https://github.com/Sylius/Sylius/pull/11126

PR #11136, opened 2020-02-24
[API] Product options (with values) creation and index

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | api
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

PR consists of API for creation and indexing product options (with product option values as sub-resources):

Zrzut ekranu 2020-02-25 o 12 36 00

It also contains a PoC of the Behat extension, that does not allow to run a javascript session for API scenarios with multiple tags (like `@ui @javascript @api`) 💃

### DIFF
--- a/src/Sylius/Behat/Context/Api/Admin/NotificationContext.php
+++ b/src/Sylius/Behat/Context/Api/Admin/NotificationContext.php
@@ -32,6 +32,6 @@ final class NotificationContext implements Context
      */
     public function iShouldBeNotifiedThatSomethingHappened(): void
     {
-        Assert::true($this->client->isCreationSuccessful(), 'Currency could not be created');
+        Assert::true($this->client->isCreationSuccessful(), 'Resource could not be created');
     }
 }
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/ProductOption.xml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/ProductOption.xml
@@ -33,4 +33,5 @@
         <property name="values" readable="true" writable="true" />
         <property name="translations" readable="true" writable="true" />
     </resource>
+
 </resources>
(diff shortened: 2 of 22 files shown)


src/Sylius/Behat/Context/Api/Admin/NotificationContext.php
```diff
@@ -32,6 +32,6 @@ public function __construct(ApiClientInterface $client)
      */
     public function iShouldBeNotifiedThatSomethingHappened(): void
     {
-        Assert::true($this->client->isCreationSuccessful(), 'Currency could not be created');
+        Assert::true($this->client->isCreationSuccessful(), 'Resource could not be created');
     }
```
Tomanhez:
```
  /**
     * @Then I should be notified that it has been successfully created
     */
    public function iShouldBeNotifiedThatSomethingHappened(): void
    {
        Assert::true($this->client->isCreationSuccessful(), 'Currency could not be created');
        Assert::true($this->client->isCreationSuccessful(), 'Resource could not be created');
    }
```

`Is successfully created` or `could not be created` ? 😄

lchrusciel: This is an exception message, so reversed information is expected here.

Zales0123: I would work on the method name, as it's too generic 😄 but fix for the next PR imo 🖖

pamil:

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/ProductOption.xml
```diff
@@ -0,0 +1,39 @@
...
+        </collectionOperations>
+        <itemOperations>
+            <itemOperation name="get" />
+        </itemOperations>
+        <property name="id" identifier="false" writable="false" />
+        <property name="code" identifier="true" required="true" />
+        <property name="createdAt" writable="false" />
+        <property name="updatedAt" writable="false" />
+        <property name="values" readable="true" writable="true">
+            <subresource resourceClass="Sylius\Component\Product\Model\ProductOptionValue" collection="true" />
```
pamil: What about not using subresources for values but IRIs instead?

PR #11144, opened 2020-02-26
[API] Product options - update vol.1

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | api
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

TO-DO:

- [x] ~handle code immutability properly - even though serialization groups are defined properly, so code change is impossible, we should still get a validation error instead of just not having a code changed 💃 - see [this line](https://github.com/Sylius/Sylius/pull/11144/files#diff-4c44808bf3d11ace4ee02fe004283aaeR168)~ (EDIT: we decided that the current behavior is good, at least for now 🐎)
- [ ] start refactoring of `ApiPlatformClient` to not make it too big 😄 (but I suppose it would be better to do it in a separate PR)

### DIFF
--- a/features/product/managing_product_options/editing_product_option.feature
+++ b/features/product/managing_product_options/editing_product_option.feature
@@ -19,7 +19,7 @@ Feature: Editing product options
         Then I should be notified that it has been successfully edited
         And this product option name should be "T-Shirt color"
 
-    @ui
-    Scenario: Seeing disabled code field while editing product option
+    @ui @api
+    Scenario: Not being able to edit code of an existing product option
         Given I want to modify the "T-Shirt size" product option
-        Then the code field should be disabled
+        Then I should not be able to edit its code
--- a/features/product/managing_product_options/editing_product_option.feature
+++ b/features/product/managing_product_options/editing_product_option.feature
@@ -9,7 +9,7 @@ Feature: Editing product options
         And the store has a product option "T-Shirt size" with a code "t_shirt_size"
         And I am logged in as an administrator
 
-    @ui
+    @ui @api
     Scenario: Renaming the product option
         Given this product option has the "S" option value with code "t_shirt_size_s"
         And this product option has also the "M" option value with code "t_shirt_size_m"
--- a/src/Sylius/Behat/Client/ApiClientInterface.php
+++ b/src/Sylius/Behat/Client/ApiClientInterface.php
@@ -48,4 +48,6 @@ interface ApiClientInterface
     public function hasValue(string $key, string $value): bool;
 
     public function hasItemWithValue(string $key, string $value): bool;
+
+    public function hasItemWithTranslation(string $locale, string $key, string $translation): bool;
 }
--- a/src/Sylius/Behat/Client/ApiClientInterface.php
+++ b/src/Sylius/Behat/Client/ApiClientInterface.php
@@ -17,15 +17,23 @@ interface ApiClientInterface
 {
     public function index(string $resource): void;
 
+    public function show(string $resource, string $id): void;
+
     public function subResourceIndex(string $resource, string $subResource, string $id): void;
 
     public function buildCreateRequest(string $resource): void;
+
+    public function buildUpdateRequest(string $resource, string $id): void;
 
     public function addRequestData(string $key, string $value): void;
 
     public function addCompoundRequestData(array $data): void;
 
+    public function updateRequestData(array $data): void;
+
     public function create(): void;
+
+    public function update(): void;
 
     public function countCollectionItems(): int;
 
@@ -34,6 +42,10 @@ interface ApiClientInterface
     public function getError(): string;
 
     public function isCreationSuccessful(): bool;
+
+    public function isUpdateSuccessful(): bool;
+
+    public function hasValue(string $key, string $value): bool;
 
     public function hasItemWithValue(string $key, string $value): bool;
 }
--- a/src/Sylius/Behat/Client/ApiPlatformClient.php
+++ b/src/Sylius/Behat/Client/ApiPlatformClient.php
@@ -35,6 +35,11 @@ final class ApiPlatformClient implements ApiClientInterface
         $this->client->request('GET', '/new-api/'.$resource, [], [], ['HTTP_ACCEPT' => 'application/ld+json']);
     }
 
+    public function show(string $resource, string $id): void
+    {
+        $this->client->request('GET', sprintf('/new-api/%s/%s', $resource, $id), [], [], ['HTTP_ACCEPT' => 'application/ld+json']);
+    }
+
     public function subResourceIndex(string $resource, string $subResource, string $id): void
     {
         $url = sprintf('/new-api/%s/%s/%s', $resource, $id, $subResource);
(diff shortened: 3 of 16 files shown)


src/Sylius/Behat/Context/Api/Admin/ManagingProductOptionsContext.php
```diff
@@ -119,6 +141,15 @@ public function theProductOptionShouldAppearInTheRegistry(ProductOptionInterface
         Assert::true($this->client->hasItemWithValue('name', $productOption->getName()));
     }
 
+    /**
+     * @Then /^(this product option) name should be "([^"]+)"$/
+     */
+    public function thisProductOptionNameShouldBe(ProductOptionInterface $productOption, string $name): void
+    {
+        $this->client->show('product_options', $productOption->getCode());
```
AdamKasp:
```suggestion
        $this->client->show('product_options', $productOption->getId());
```
?

Zales0123: We defined the code as identifier of currency (check [here](https://github.com/Sylius/Sylius/blob/api/src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Currency.xml#L20)), so I've done the same with product options. It's another decision we need to take. We use codes in Sylius (ui routes), but, on the other hand, it's not reliable as it's always just and autogenerated integer ⚖️ I'm for using codes, but cc @pamil @lchrusciel

AdamKasp: okey I was confused because in `apiClient` we have `public function buildUpdateRequest(string $resource, string $id): void`  so in this functions we should have `$code` instead `$id` if i understand it correctly 💃

Zales0123: To be honest... I'm not sure 😄 For us, in Sylius, it's a code indeed, but for ApiPlatform it's an identifier (even if you take a look at the URL there is `{id}`, not `{code}`). So it still makes sense to me to keep it as it is 🚀 But I'm not forcing it if it's confusing for some more people as well 🖖

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/ProductOptionTranslation.xml
```diff
@@ -0,0 +1,20 @@
...
+        class="Sylius\Component\Product\Model\ProductOptionTranslation"
+        shortName="ProductOptionTranslation"
+    >
+        <attribute name="validation_groups">sylius</attribute>
+        <collectionOperations />
+        <itemOperations>
+            <itemOperation name="get" />
+        </itemOperations>
+        <property name="id" identifier="true" writable="false" />
+        <property name="name" identifier="true" required="true" />
```
Tomanhez: Why we have 2 identifier? It seems that we can get resource by name or id ?

Zales0123: It should not be an identifier for sure 🎉 🚀

src/Sylius/Behat/Client/ApiPlatformClient.php
```diff
@@ -35,6 +35,11 @@ public function index(string $resource): void
         $this->client->request('GET', '/new-api/'.$resource, [], [], ['HTTP_ACCEPT' => 'application/ld+json']);
     }
 
+    public function show(string $resource, string $id): void
+    {
+        $this->client->request('GET', sprintf('/new-api/%s/%s', $resource, $id), [], [], ['HTTP_ACCEPT' => 'application/ld+json']);
```
GSadee: Maybe we can use `ApiPlatform\Core\Api\IriConverterInterface` to generate an iri in this client?

src/Sylius/Behat/Client/ApiClientInterface.php
```diff
@@ -35,5 +43,11 @@ public function getError(): string;
 
     public function isCreationSuccessful(): bool;
 
+    public function isUpdateSuccessful(): bool;
```
GSadee: I'm not sure, but this method name is not consistent with the one above 😄

src/Sylius/Behat/Client/ApiPlatformClient.php
```diff
@@ -110,4 +160,15 @@ private function getResponseContentValue(string $key)
 
         return $content[$key];
     }
+
+    private function mergeArraysUniquely(array $array1, array $array2): array
```
GSadee:
```suggestion
    private function mergeArraysUniquely(array $firstArray, array $secondArray): array
```
? 😃

src/Sylius/Behat/Context/Api/Admin/ManagingProductOptionsContext.php
```diff
@@ -52,20 +49,45 @@ public function iWantToCreateANewProductOption(): void
         $this->client->buildCreateRequest('product_options');
     }
 
+    /**
+     * @Given I want to modify the :productOption product option
```
GSadee:
```suggestion
     * @When I want to modify the :productOption product option
```
🙏

src/Sylius/Behat/Context/Api/Admin/ManagingProductOptionsContext.php
GSadee: Maybe we should use an IRI as the unique identifier in API Platform (`ApiPlatform\Core\Api\IriConverterInterface`)?

features/product/managing_product_options/editing_product_option.feature
```diff
@@ -19,7 +19,7 @@ Feature: Editing product options
         Then I should be notified that it has been successfully edited
         And this product option name should be "T-Shirt color"
 
-    @ui
-    Scenario: Seeing disabled code field while editing product option
+    @ui @api
+    Scenario: Not being able to edit code of an existing product option
         Given I want to modify the "T-Shirt size" product option
-        Then the code field should be disabled
+        Then I should not be able to edit its code
```
pamil: 👍

src/Sylius/Behat/Client/ApiClientInterface.php
Zales0123: I believe `update` is also a noun, not only a verb

src/Sylius/Behat/Client/ApiPlatformClient.php
Zales0123: I agree :+1: However, do you think it can be a part of next PR, that would refactor `ApiPlatformClient` a little bit? It's needed, nevertheless

src/Sylius/Behat/Context/Api/Admin/ManagingProductOptionsContext.php
```diff
@@ -119,15 +149,74 @@ public function theProductOptionShouldAppearInTheRegistry(ProductOptionInterface
         Assert::true($this->client->hasItemWithValue('name', $productOption->getName()));
     }
 
+    /**
+     * @Then the product option with :element :value should not be added
+     */
+    public function theProductOptionWithElementValueShouldNotBeAdded(string $element, string $value): void
+    {
+        $this->client->index('product_options');
+        Assert::false($this->client->hasItemWithValue($element, $value));
```
AdamKasp: Please add some custom error message boolean assertions

src/Sylius/Bundle/ApiBundle/Resources/config/serializer/ProductOptionTranslation.xml
```diff
@@ -11,9 +11,11 @@
         <attribute name="name">
             <group>product_option:read</group>
             <group>product_option:write</group>
+            <group>product_option:update</group>
```
GSadee:
```suggestion
            <group>product_option:create</group>
            <group>product_option:update</group>
```
?

GSadee:

lchrusciel:
Part of #11250

PR #11164, opened 2020-02-29
[typography] return function

DurandSacha:


lchrusciel:
Thanks, Sacha! :tada:

PR #11053, opened 2020-01-22
[Admin][Channel] Revert #10860

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | reverts #10860
| License         | MIT

PR #11167, opened 2020-03-02
Fix UPGRADE-1.7 file

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

I believe we should also delete [this migration file](https://github.com/Sylius/Sylius-Standard/blob/1.7/src/Migrations/Version20191119131635.php) due to https://github.com/Sylius/Sylius/pull/11053

### DIFF
--- a/UPGRADE-1.7.md
+++ b/UPGRADE-1.7.md
@@ -3,7 +3,7 @@
 1. Require upgraded Sylius version using Composer:
 
     ```bash
-    composer require sylius/sylius:~1.7.0
+    composer require sylius/sylius:~1.7.0 --update-with-dependencies
     ```
    
    You might need to adjust your `config.platform.php` setting in `composer.json`, because Sylius 1.7 requires PHP 7.3 or higher.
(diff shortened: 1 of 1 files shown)


UPGRADE-1.7.md
```diff
@@ -50,7 +52,6 @@
    
 6. Copy migration files into `src/Migrations`:
 
-    - [Version20191119131635.php](https://raw.githubusercontent.com/Sylius/Sylius-Standard/1.7/src/Migrations/Version20191119131635.php)
```
pamil: This will make the other migration fail while trying to revert the first one: https://github.com/Sylius/Sylius-Standard/blob/1.7/src/Migrations/Version20200301170604.php#L29

There are two possible solutions:
1. Generate one migration instead of those two
2. Leave these two migrations

If we go with 1., all projects on v1.7.0-ALPHA.2 will be much harder to upgrade.

Zales0123: 👍 I didn't notice this tiny `DROP type` 😄

lchrusciel:
Thank you, Mateusz! :tada:

PR #11210, opened 2020-03-11
[API] Refactor ApiPlatformClient ver.1

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | api
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Changes:
- extraction of `Request` object, that knows about all technical details of the sent request (headers, method, URL, optionally content)
- splitting the `ApiPlatformClient` into two services: one sending requests (still named `ApiPlatformClient`), the other for checking the response and getting its values (named `ResponseChecker`)
- setting the resource per `ApiPlatformClient` to not being forced to specify it on each request

Known issues:
- [filter](https://github.com/Sylius/Sylius/compare/api...Zales0123:refactor-api-client?expand=1#diff-00c4f46caee98d6e5ba3c593d125e36cR83) method still operates kinda _old way_ (we can consider merging it with the index or separating another `Request` type)
- `ApiSecurityService` is still not refactored (the question is - should it)

It should also allow us to operate on multiple responses easier 🖖

I'm open for the discussion regarding naming and proposed solutions - we should make this service as perfect as possible 💃


src/Sylius/Behat/Client/ApiPlatformClient.php
```diff
@@ -38,214 +40,100 @@ public function __construct(AbstractBrowser $client, SharedStorageInterface $sha
...
-        $content = json_decode($this->client->getResponse()->getContent(), true);
-
-        Assert::keyExists($content, $key);
-
-        return $content[$key];
+        return $this->client->getResponse();
     }
 
-    private function mergeArraysUniquely(array $firstArray, array $secondArray): array
+    private function request(Request $request): void
```
lchrusciel: I'm wondering, if it won't be useful to provide an interface for our `Request` object, in case someone would like to change something in the default behaviour.

src/Sylius/Behat/Context/Api/Admin/ManagingTaxCategoriesContext.php
```diff
@@ -23,17 +24,22 @@ final class ManagingTaxCategoriesContext implements Context
...
     private $client;
 
-    public function __construct(ApiClientInterface $client)
+    /** @var ResponseCheckerInterface */
+    private $responseChecker;
+
+    public function __construct(ApiClientInterface $client, ResponseCheckerInterface $responseChecker)
     {
         $this->client = $client;
+        $this->client->setResource('tax_categories');
```
lchrusciel: I would prefer to create clients per request instead of setting it in the constructor. Right now, class API wouldn't be obvious for me. It is not possible to use this service without this configuration, still, it is super easy to omit this.

src/Sylius/Behat/Client/ApiPlatformClient.php
```diff
@@ -38,214 +40,100 @@ public function __construct(AbstractBrowser $client, SharedStorageInterface $sha
...
 
-    public function filter(string $resource): void
+    public function filter(): void
     {
         $query = http_build_query($this->filters, '', '&', PHP_QUERY_RFC3986);
-        $path = sprintf('/new-api/%s?%s', $resource, $query);
-
-        $this->request('GET', $path, ['HTTP_ACCEPT' => 'application/ld+json']);
-    }
+        $path = sprintf('/new-api/%s?%s', $this->resource, $query);
```
pamil: Could we include a dedicated method in `Request` object?

src/Sylius/Behat/Client/Request.php
```diff
@@ -0,0 +1,148 @@
...
+    {
+        $this->content = $this->mergeArraysUniquely($this->content, $newValues);
+    }
+
+    public function addSubResource(string $key, array $subresource): void
+    {
+        $this->content[$key][] = $subresource;
+    }
+
+    private function mergeArraysUniquely(array $firstArray, array $secondArray): array
```
pamil: What's the difference from `array_merge_recursive`? Would be cool to mention.

src/Sylius/Behat/Client/ApiClientInterface.php
```diff
@@ -13,65 +13,42 @@
...
-    public function buildFilter(array $filters): void;
-
-    /** @param string|int $value */
-    public function addRequestData(string $key, $value): void;
-
-    public function addCompoundRequestData(array $data): void;
+    public function subResourceIndex(string $subResource, string $id): void;
 
-    public function updateRequestData(array $data): void;
+    public function show(string $id): void;
```
pamil: It'd be cool if it could return a response.

src/Sylius/Behat/Client/ApiClientInterface.php
```diff
@@ -13,65 +13,42 @@
...
 
-    /** @param string|float $value */
-    public function hasItemWithValue(string $key, $value): bool;
+    public function addSubResourceData(string $key, array $data): void;
 
-    public function hasItemOnPositionWithValue(int $position, string $key, string $value): bool;
+    public function updateRequestData(array $data): void;
 
-    public function hasItemWithTranslation(string $locale, string $key, string $translation): bool;
+    public function getResponse(): Response;
```
pamil: `getLastResponse()`?

src/Sylius/Behat/Context/Api/Admin/ManagingCustomerGroupsContext.php
```diff
@@ -23,17 +24,22 @@ final class ManagingCustomerGroupsContext implements Context
...
     private $client;
 
-    public function __construct(ApiClientInterface $client)
+    /** @var ResponseCheckerInterface */
+    private $responseChecker;
+
+    public function __construct(ApiClientInterface $client, ResponseCheckerInterface $responseChecker)
     {
         $this->client = $client;
+        $this->client->setResource('customer_groups');
```
pamil: Is the `$client` instance different for every context? Or is it shared?

If it's shared, there might be a case in which two context constructors will be called, so that only the last resource will stay on the client instance.

src/Sylius/Behat/Context/Api/Admin/ManagingTaxCategoriesContext.php
Zales0123: That's the next step I would like to do 👌

src/Sylius/Behat/Client/ApiPlatformClient.php
Zales0123: Yes, I've mentioned it in the PR description 👍

src/Sylius/Behat/Context/Api/Admin/ManagingCustomerGroupsContext.php
Zales0123: Indeed 👌 Right not it's not a problem, as current API scenarios use steps only from one context 💃 I was thinking about not registering the ApiPlatformClient in the container explicitly but injecting to each context anonymous service with the $resource property already defined. What do you think about such an approach?

src/Sylius/Behat/Context/Api/Admin/ManagingTaxCategoriesContext.php
Zales0123: See [this discusion](https://github.com/Sylius/Sylius/pull/11210/files#r391062258)

src/Sylius/Behat/Context/Api/Admin/ManagingCustomerGroupsContext.php
Tomanhez: Its good question, we can also add optional resource parameter in methods like `buildCreateRequest()` -> `buildCreateRequest(string resource = null)`, but this case make you to write more code for checking existing parameter

src/Sylius/Behat/Client/ResponseChecker.php
```diff
@@ -0,0 +1,110 @@
...
+                $resource['translations'][$locale][$key] === $translation
+            ) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    private function getResponseContentValue(Response $response, string $key)
```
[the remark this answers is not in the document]
Zales0123: Return type is mixed 💃

src/Sylius/Behat/Client/Request.php
Zales0123: It does not duplicate values for the same keys. But yes, it should be in the comment 👌

src/Sylius/Behat/Context/Api/Admin/ManagingCustomerGroupsContext.php
Zales0123: Probably we can firstly go with multiple `ApiClient`s registered for each resource (with `$resource` property injected) and then refactor it to register them manually. WDYT?

src/Sylius/Behat/Context/Api/Admin/ManagingProductOptionsContext.php
```diff
@@ -203,11 +210,14 @@ public function theProductOptionWithElementValueShouldNotBeAdded(string $element
...
     {
-        $this->client->index('product_options');
-        $itemsCount = $this->client->countCollectionItems();
+        $this->client->index();
+        $itemsCount = $this->responseChecker->countCollectionItems($this->client->getLastResponse());
 
-        Assert::eq(1, $this->client->countCollectionItems(), sprintf('Expected 1 product options, but got %d', $itemsCount));
-        Assert::true($this->client->hasItemWithValue($element, $value));
+        Assert::eq(
+            1, $this->responseChecker->countCollectionItems($this->client->getLastResponse()),
```
lchrusciel:
```suggestion
            1,
            $this->responseChecker->countCollectionItems($this->client->getLastResponse()),
```

src/Sylius/Behat/Context/Api/Admin/NotificationContext.php
```diff
@@ -1,53 +0,0 @@
...
-
-declare(strict_types=1);
-
-namespace Sylius\Behat\Context\Api\Admin;
-
-use Behat\Behat\Context\Context;
-use Sylius\Behat\Client\ApiClientInterface;
-use Webmozart\Assert\Assert;
-
-final class NotificationContext implements Context
```
lchrusciel: Maybe it wasn't a bad idea to have these checks in one place?

GSadee: I agree, but I have no idea, how it should be implemented since we have api client per resource 😞

Zales0123: It wasn't, but we also would have to register this context with each api_platform_client, as there is a requirement to have the `resource` property set. In theory - it doesn't matter, as we only check the status. But we still need to get the response from the client :/ So in a result - it's worth to be refactored, but probably not now 🖖

src/Sylius/Behat/Context/Api/Admin/ManagingTaxCategoriesContext.php
```diff
@@ -201,13 +221,37 @@ public function taxCategoryWithNamedElementShouldNotBeAdded(string $element, str
...
-        Assert::same($this->client->countCollectionItems(), 1);
+        Assert::same($this->responseChecker->countCollectionItems($this->client->getLastResponse()), 1);
+    }
+
+    /**
+     * @Then I should be notified that it has been successfully created
+     */
+    public function iShouldBeNotifiedThatItHasBeenSuccessfullyCreated(): void
+    {
+        Assert::true($this->responseChecker->isCreationSuccessful($this->client->getLastResponse()));
```
lchrusciel: There are a few `true/false` assertion without any comment, which will end up with an awesome error: "Expected true, got false"

We should do something about it in the next iteration

lchrusciel:

Zales0123:
And Grzegorz! 🐎

lchrusciel:
Part of #11250

PR #11214, opened 2020-03-11
Use https instead of http for Channel's hostname in form

jacquesbh:
| Q               | A
| --------------- | -----
| Branch?         | ^1.0 I think, master otherwise 
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

I really prefer to show `https://` instead of `http://` in the Channel form, it's a way to say "Hey, keep the web secure!".

And it doesn't change that much.

It could be `http(s)://` as well…

As you prefer!


lchrusciel:
Thank you, Jacques! :tada:

PR #11226, opened 2020-03-13
Initializing adjustments before the order item is processed

mamazu:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no (kinda)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | ----
| License         | MIT

This pull request moves the initialization of the object's property before the logic that this object executes. The reason for this change is if the methods that are called because of the `addUnit` function require the adjustments to be set, then those will crash because the property has not been initialized yet.


mamazu:
Sould I rebase it to an older version?

pamil:
Thank you, @mamazu! :tada:

PR #11243, opened 2020-03-17
[API] Browsing shipping methods

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | api
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT


src/Sylius/Behat/Context/Api/Admin/ManagingShippingMethodsContext.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $this->client->index();
+    }
+
+    /**
+     * @Then I should see :count shipping methods in the list
+     */
+    public function iShouldSeeShippingMethodsInTheList(int $count): void
+    {
+        Assert::count($this->responseChecker->getCollection($this->client->index()), $count);
```
Zales0123: Maybe it should be `getLastResponse()`? I think we should use it in the steps when the previous `index` call is natural ("in the list", "in the registry" etc.), wdyt?

lchrusciel:
Part of #11250

PR #11248, opened 2020-03-18
[API][Order] Fulfilling orders

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Part of https://github.com/Sylius/Sylius/issues/11250

### DIFF
--- a/src/Sylius/Behat/Client/Request.php
+++ b/src/Sylius/Behat/Client/Request.php
@@ -100,7 +100,11 @@ final class Request implements RequestInterface
 
     public static function custom(string $url, string $method, string $token): RequestInterface
     {
-        return new self($url, $method, ['HTTP_Authorization' => 'Bearer ' . $token]);
+        return new self(
+            $url,
+            $method,
+            ['CONTENT_TYPE' => 'application/merge-patch+json', 'HTTP_Authorization' => 'Bearer ' . $token]
+        );
     }
 
     public function url(): string
--- a/src/Sylius/Behat/Context/Api/Admin/ManagingOrdersContext.php
+++ b/src/Sylius/Behat/Context/Api/Admin/ManagingOrdersContext.php
@@ -21,12 +21,20 @@ use Sylius\Behat\Service\SharedStorageInterface;
 use Sylius\Component\Core\Model\CustomerInterface;
 use Sylius\Component\Core\Model\OrderInterface;
 use Sylius\Component\Order\OrderTransitions;
+use Sylius\Component\Payment\PaymentTransitions;
+use Sylius\Component\Shipping\ShipmentTransitions;
 use Webmozart\Assert\Assert;
 
 final class ManagingOrdersContext implements Context
 {
     /** @var ApiClientInterface */
     private $client;
+
+    /** @var ApiClientInterface */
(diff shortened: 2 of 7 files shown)


src/Sylius/Behat/Context/Api/Admin/ManagingOrdersContext.php
```diff
@@ -76,6 +77,30 @@ public function iCancelThisOrder(OrderInterface $order): void
...
+        );
+    }
+
+    /**
+     * @When /^I ship (this order)$/
+     */
+    public function iShipThisOrder(OrderInterface $order): void
+    {
+        $this->client->customEndPoint(
+            sprintf('/shipments/%s/ship', (string) $order->getShipments()->first()->getId()),
```
lchrusciel: That's why we need some possibility to gather links and available actions :(

lchrusciel: Another question is, do we want to expose these actions also as sub resources of orders (what we are doing in admin panel). Perhaps not?

lchrusciel:
Part of #11250

src/Sylius/Behat/Context/Api/Admin/ManagingOrdersContext.php
Zales0123: Perhaps... I don't know 🤷‍♂. On the one hand, It seems reasonable to access shipments via orders. You check out the orders list (`orders/`), then the specific order (`orders/000111`), you want to see its shipments (`orders/000111/shipments`), especially one of them (`orders/000111/shipments/12`), and to ship it (`orders/000111/shipments/12/dancer`) 💃 But it seems like a long way and probably an improvement to the base as it is right now 🖖

lchrusciel:
part of #11250

src/Sylius/Behat/Client/Request.php
```diff
@@ -100,7 +100,11 @@ public static function transition(string $resource, string $id, string $transiti
 
     public static function custom(string $url, string $method, string $token): RequestInterface
     {
-        return new self($url, $method, ['HTTP_Authorization' => 'Bearer ' . $token]);
+        return new self(
```
lchrusciel: Probably not relevant to this PR

src/Sylius/Behat/Context/Api/Admin/ManagingOrdersContext.php
```diff
@@ -76,6 +89,28 @@ public function iCancelThisOrder(OrderInterface $order): void
         );
     }
 
+    /**
+     * @When /^I mark (this order) as paid$/
+     */
+    public function iMarkThisOrderAsAPaid(OrderInterface $order): void
+    {
+        $this->paymentsClient->applyTransition(
+            (string) $order->getLastPayment()->getId(),
```
lchrusciel: Something that we should consider is to test these actions when we will access them from the order show response

lchrusciel:
Thanks, Adam! :1st_place_medal:

Issue #11250, opened 2020-03-18
New Sylius API on API Platform


lchrusciel:
This issue is created to coordinate efforts and have one space to reference to all of the PR's aiming to provide API platform support for Sylius.

## Sylius API Architecture

Every time you will contribute to Sylius API please, take into account the following decisions (ADRs):
- [API - Product Option Value](https://github.com/Sylius/Sylius/blob/master/adr/2020_03_03_api_product_option_values.md)
- [API - Translations](https://github.com/Sylius/Sylius/blob/master/adr/2020_03_03_api_translations.md)
- [Feature Coverage in API](https://github.com/Sylius/Sylius/blob/master/adr/2020_03_03_feature_coverage_in_api.md)

If you would like to add something there feel free to open a new proposal :) 

**This effort is being done in the current Sylius `master` branch. Please help us with testing what's been done, by using the API locally** 🚀

## TODO for the community - Admin API coverage

If you can't wait to see Sylius with the full API Platform support, feel more than welcome to help us with it. All of these will require to have behat coverage (@ui tagged scenarios run also tagged as @api) to merge them to master. 

- [x]  **Locales** #11273 
All the features from `features/locale/managing_locales`
2 scenarios / 3 files
Create an index action only (show may be required as well)

- [x] **Association Types** #11319 
All the features from: `features/product/managing_product_association_types`
12 scenarios / 8 files
Full CRUD for options and values + validation and filtering

- [x] **Countries** #11278, #11431
All the features from `features/addressing/managing_countries`
15 scenarios / 8 files 
Full CRUD for countries and provinces + validation

- [x] **Zones** #11372
All the features from `features/addressing/managing_zones`
20 scenarios, 7 files
Full CRUD
Ensure proper zone type validation

- [ ] **Taxonomies**
All the features from `features/taxonomy`
47 scenarios, 14 files
Full CRUD
Only features with `@ui` and without `@todo` tag

- [ ] **Channels** #11417, #11486
All the features from `features/user/managing_channels`
35 scenarios, 19 files (16 scenarios/4 files already done)
Full CRUD

- [x] **Tax rates** #14811
All the features from `features/taxation/managing_tax_rates`
18 scenarios, 6 files
Full CRUD

- [ ] **Modifying Order's address**
All the features from `features/order/managing_orders/modifying_billing_address` and `features/order/managing_orders/modifying_shipping_address`
18 scenarios, 10 files

- [ ] **Filtering on the Orders list**
[...]

PR #11254, opened 2020-03-19
[API] Adding shipping methods

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | parts of #11250 
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Client/ResponseChecker.php
+++ b/src/Sylius/Behat/Client/ResponseChecker.php
@@ -69,6 +69,12 @@ final class ResponseChecker implements ResponseCheckerInterface
     }
 
     /** @param string|int $value */
+    public function hasValueInCollection(Response $response, string $key, $value): bool
+    {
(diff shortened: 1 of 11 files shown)


src/Sylius/Behat/Client/ResponseChecker.php
```diff
@@ -68,6 +68,12 @@ public function hasValue(Response $response, string $key, $value): bool
         return $this->getResponseContentValue($response, $key) === $value;
     }
 
+    /** @param string|int $value */
+    public function hasValueInCollection(Response $response, string $key, $value): bool
```
GSadee: We should definitely refactor names of these methods in future, as I thought at the first moment that this is the same method as `hasItemWithValue`

GSadee:

Issue #8864, opened 2017-10-17
how to upload image via api product creation


sherif807:
can anyone give an example on how to upload image file via the api when creating a product.?

the documentation doesn't show it. I am using guzzle. any help please?

sherif807:
no matter what file I send it, it keeps saying this

"Expected argument of type "SplFileInfo or null", "array" given"

lasotaartur:
Someone can help?

sherif807:
left sylius to woo api. never got an answer anywhere.

bruno-ds:
Hello,
I just looked at the doc, it seem to be explain : http://docs.sylius.org/en/latest/api/products.html
> Note
>
> The images (files) should be passed in an array as an attribute of request. See how it is done in Sylius here. https://github.com/Sylius/Sylius/blob/master/tests/Controller/ProductApiTest.php

gorkalaucirica:
Any cURL example? Can't make it work

salamichel:
Ok Bruno, I understand how to push image to sylius with API with they're examples https://github.com/Sylius/Sylius/blob/master/tests/Controller/ProductApiTest.php#L483

But I do not understand how to push image without using UploadedFile which is a symfony component... so there is a way to push image with tranditionnal php libs ? or equivalent ?

bruno-ds:
Probably, it's role is just to serialize the file, you should look at the raw request sent (there should be a method like `$this->client->getRequest()`).
But since Sylius rely on Symfony, using UploadedFile seems the easiest way...

dimaip:
Am I correct to assume that currently it's not possible to upload images into Sylius purely using its REST API, without creating any Symfony plugins?

My usecase is that I'm trying to upload images into Sylius from my GO app, and I don't want to create any plugins for it.

kaankilic:
Does it still not possible in 2020? It's really hard to upload photos via API in some cases impossible

vvasiloi:
@kaankilic There's an workaround: https://github.com/Sylius/SyliusAdminApiBundle/issues/31

Krishan29:
Hi @bruno-ds 
I am sending the images in the same format in the request data in sylius API as given in the file https://github.com/Sylius/Sylius/blob/master/tests/Controller/ProductApiTest.php#L483
but the image is not creating at sylius.

Can anyone help me?
Actually I am sending as below:

 [
  "code" => "test1"
  "productTaxons" => ""
  "channels" => array:1 [
[...]

PR #11255, opened 2020-03-19
[API][Administrator] Implement managing avatar features

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | part of #11250
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Api/Admin/ManagingAdministratorsContext.php
+++ b/src/Sylius/Behat/Context/Api/Admin/ManagingAdministratorsContext.php
@@ -13,10 +13,14 @@ declare(strict_types=1);
 
 namespace Sylius\Behat\Context\Api\Admin;
 
+use ApiPlatform\Core\Api\IriConverterInterface;
 use Behat\Behat\Context\Context;
 use Sylius\Behat\Client\ApiClientInterface;
 use Sylius\Behat\Client\ResponseCheckerInterface;
+use Sylius\Behat\Service\SharedStorageInterface;
+use Sylius\Component\Core\Formatter\StringInflector;
 use Sylius\Component\Core\Model\AdminUserInterface;
+use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Webmozart\Assert\Assert;
 
 final class ManagingAdministratorsContext implements Context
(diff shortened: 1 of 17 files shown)


src/Sylius/Behat/Context/Api/Admin/ManagingAdministratorsContext.php
```diff
@@ -129,13 +155,41 @@ public function iDeleteAdministratorWithEmail(AdminUserInterface $adminUser): vo
...
     {
         $this->client->buildUpdateRequest((string) $adminUser->getId());
     }
 
+    /**
+     * @When /^I (?:|upload|update) the "([^"]+)" image as (my) avatar$/
+     */
+    public function iUploadTheImageAsMyAvatar(string $avatar, AdminUserInterface $administrator): void
+    {
+        $response = $this->client->upload(
```
Zales0123: Shouldn't we also use `avatarImageClient` for the avatar uploading?

src/Sylius/Behat/Context/Api/Admin/ManagingAdministratorsContext.php
```diff
@@ -129,13 +155,41 @@ public function iDeleteAdministratorWithEmail(AdminUserInterface $adminUser): vo
...
+            '/new-api/avatar-images',
+            ['owner' => $this->iriConverter->getIriFromItem($administrator)],
+            ['file' => new UploadedFile($this->minkParameters['files_path'] . $avatar, basename($avatar))]
```
Zales0123: I'm wondering, is it a proper approach 🤔 It's not resolving the issue like [this one](https://github.com/Sylius/Sylius/issues/8864) when one wants to just use the API (not call the endpoint in the code). Maybe we should just pass the URL to the image which is temporarily somewhere else? I don' know, just thinking out loud 😄

GSadee: Probably I don't understand you in 100%, the case is to upload an image, not to pass a URL to the image... I have no other idea how to approach this.

src/Sylius/Behat/Context/Api/Admin/ManagingAdministratorsContext.php
GSadee: To be honest, there is no difference as I pass entire url here. Do you suggest to refactor upload method?

Zales0123: It's an option, but for me, the main problem is confusion, that we upload an avatar with `$client`, but remove it with the `avatarImageClient` 🐎

src/Sylius/Behat/Context/Api/Admin/ManagingAdministratorsContext.php
lchrusciel: If I understood correctly, there is an ongoing problem in Sylius documentation on how to handle files upload through API since the Admin API. Problem is that what you have used here (and is used in the whole Sylius codebase) works because we are operating inside of a project. One could not simply use `UploadedFile` in js.

src/Sylius/Bundle/ApiBundle/Controller/UploadAvatarImageAction.php
```diff
@@ -0,0 +1,80 @@
...
+        $ownerIri = $request->get('owner');
+        Assert::notEmpty($ownerIri);
+
+        /** @var ResourceInterface|AdminUserInterface $owner */
+        $owner = $this->iriConverter->getItemFromIri($ownerIri);
+        Assert::isInstanceOf($owner, AdminUserInterface::class);
+
+        $oldImage = $owner->getImage();
+        if ($oldImage !== null) {
+            $this->avatarImageRepository->remove($oldImage);
```
lchrusciel: I'm wondering if it will remove a file from disk as well

src/Sylius/Behat/Context/Api/Admin/ManagingAdministratorsContext.php
lchrusciel: I would vote for doing it the same way with the rest of our resources. Right now upload function is behaving differently than other client methods, so +1 for refactor

lchrusciel:
Thank you, Grzegorz! :tada:

src/Sylius/Bundle/ApiBundle/Controller/UploadAvatarImageAction.php
GSadee: Yup, to be 100% sure, I've checked that and `ImagesRemoveListener` will remove the file

PR #11267, opened 2020-03-20
[API] Sorting shipping method by code and position

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | part of https://github.com/Sylius/Sylius/issues/11250
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/services/filters.xml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/services/filters.xml
@@ -39,6 +39,13 @@
             <tag name="api_platform.filter" />
         </service>
 
+        <service id="sylius.api.shipping_method_order_filter" parent="api_platform.doctrine.orm.order_filter">
+            <argument type="collection">
+                <argument key="code"/>
+            </argument>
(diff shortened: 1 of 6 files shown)


src/Sylius/Bundle/ApiBundle/Resources/config/services/filters.xml
```diff
@@ -39,6 +39,13 @@
             <tag name="api_platform.filter" />
         </service>
 
+        <service id="sylius.api.shipping_method_order_filter" parent="api_platform.doctrine.orm.order_filter">
+            <argument type="collection">
+                <argument key="code"/>
```
lchrusciel: Is it working ok, if we don't have position in this collection?

Zales0123: Yes 😄 In fact, in our _sorting by position_ scenarios we don't sort shipping methods 💃 We just check does the default sorting by position works well :rock

lchrusciel:

Issue #11301, opened 2020-03-29
Support for JMS serializer 3.0


reyostallenberg:
**Description**
To be able to install other libraries on my installation I want to have support for JMS Serializer 3.0, see https://github.com/schmittjoh/serializer/blob/master/UPGRADING.md#from-2x-to-300

Is it possible? Or are there issues mentioned as "deeper branch group exclusion strategy" has a different behaviour, the latest group is used instead of falling back to "Default" ?

I'm aware of https://github.com/Sylius/Sylius/issues/9288 but I think it'll take longer to implement and has a totally different path.

lchrusciel:
Hey Reyo,

I don't remember usage of `deeper branch group exclusion strategy` from the top of my head. Can you open a PR with support for the new version? While we will see a build, we will know how much work is needed to finish it. 

Also, I would recommend dropping support for JMS serializer 2.0

reyostallenberg:
Dropping support makes upgrade paths harder and is not needed for Sylius if `deeper branch group exclusion strategy` is not used. In that case it will be able to support both

lchrusciel:
The problem with the support for 2.0 is that we should have it tested and it would mean, that we need another build for that. I would vote for the removal of 2.0. @Sylius/core-team does anyone have some opinion here?

leflings:
As for `deeper branch group exclusion strategy` I'm assuming [this](http://jmsyst.com/libs/serializer/master/cookbook/exclusion_strategies#overriding-groups-of-deeper-branches-of-the-graph) is what's referred to.

Also mentioned in #10586

PR #11302, opened 2020-03-30
Also allow jms/serializer-bundle 3.0 as both 2.0 and 3.0 are compatible with Sylius

reyostallenberg:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/11301
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
         "friendsofsymfony/rest-bundle": "^2.1",
         "fzaninotto/faker": "^1.6",
         "gedmo/doctrine-extensions": "^2.4.12",
-        "jms/serializer-bundle": "^2.0",
+        "jms/serializer-bundle": "^2.0 || ^3.0",
         "knplabs/gaufrette": "^0.8",
         "knplabs/knp-gaufrette-bundle": "^0.7",
         "knplabs/knp-menu": "^3.1",


lchrusciel:
Sure, thanks for your effort! ;) 

As you can see in #10928, we will need to update them anyway. It may happen, that because of dependencies constraints, we will support both versions to enable Symfony 4.4 and Symfony 5.0 support.

composer.json
```diff
@@ -97,8 +97,8 @@
         "webmozart/assert": "^1.1",
         "white-october/pagerfanta-bundle": "^1.0.8",
         "willdurand/hateoas": "^2.12",
-        "willdurand/hateoas-bundle": "^1.2",
-        "winzou/state-machine-bundle": "^0.3",
+        "willdurand/hateoas-bundle": "^1.2 || ^2.0",
+        "winzou/state-machine-bundle": "^0.3 || ^4.0",
```
pamil, line 101: I guess this change can be reverted, there's no `winzou/state-machine-bundle` in `^4.0`.

pamil: I removed the last commit from your PR, so we can get it merged faster 🎉

pamil:
ResourceBundle needs to be updated first, right now the next major versions are not installed and tested.

reyostallenberg:
I've made a mistake, I meant

https://github.com/winzou/StateMachineBundle/tree/0.4.2

What requirement is needed for ResourceBundle?

pamil:
@reyostallenberg as for the next minor of WinzouStateMachineBundle, it would require some changes to provide backwards-compatibility - there are end-user Sylius applications that will break when we upgrade, the names of the services have changed in that release.

reyostallenberg:
All right, that's not what we want. But for my current setup in an existing sf app I only got it working by requiring:
`"winzou/state-machine-bundle": "0.4.2 as 0.3.1"`

I need some time to sort out what we need to do with this dependency hell. Only allowing `"jms/serializer-bundle": "^2.0 || ^3.0"` is not enought to get all stuff working/installable.

esserj:
@pamil what needs to happen here still? Because we ran into the same issue when trying to combine sylius with our bundles, I see the resource bundle does this:

```php
final class WinzouStateMachinePass implements CompilerPassInterface
{
    /**
     * {@inheritdoc}
     */
    public function process(ContainerBuilder $container): void
    {
        $container->setAlias('sm.factory', FactoryInterface::class);
        $container->setAlias('sm.callback_factory', CallbackFactoryInterface::class);
        $container->setAlias('sm.callback.cascade_transition', CascadeTransitionCallback::class);

        foreach (['sm.factory', 'sm.callback_factory', 'sm.callback.cascade_transition'] as $id) {
            try {
                $container->findDefinition($id)->setPublic(true);
            } catch (ServiceNotFoundException $exception) {
            }
        }
    }
}
```

Is this maybe something we can add to all bundles that support .3 and .4?

esserj:
any feedback?

reyostallenberg:
I think after all upgrades done this year, this issue can and should be closed

PR #11314, opened 2020-04-01
Add translation key

loevgaard:
Used in Sylius\Bundle\LocaleBundle\Form\Type\LocaleChoiceType

| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/LocaleBundle/Resources/translations/messages.en.yml
+++ b/src/Sylius/Bundle/LocaleBundle/Resources/translations/messages.en.yml
@@ -2,3 +2,4 @@ sylius:
     form:
         locale:
             name: Name
+            select: Select a locale...


Zales0123:
Hello Joachim! Thank you for your fix... but I'm wondering why it was not detected before 😄 To be honest, I see no places where this placeholder is visible in the Admin panel - did you encounter this problem somewhere or just saw a missing translation? :) If it's not needed, maybe we should just remove a `placeholder` attribute from the `LocaleChoiceType` 💃

loevgaard:
I used the locale choice type in my own project and then I saw the missing translation key. From my point of view it would be no problem to delete although I really like placeholders :D You decide, @Zales0123 ;)

loevgaard:
Here is a screenshot so you can see it from my pov

image

loevgaard:
You are very welcome 🎉

PR #11347, opened 2020-04-10
[RFC][DX] Open consts holder classes

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Final classes, which holds consts has been changed to interfaces. Therefore, still, no one is able to initialize them, while we are open to multi inheritance.  This should remove the problem of adding new values to them in plugins and end applications.

### DIFF
--- a/src/Sylius/Bundle/UserBundle/UserEvents.php
+++ b/src/Sylius/Bundle/UserBundle/UserEvents.php
@@ -13,7 +13,7 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\UserBundle;
 
-final class UserEvents
+interface UserEvents
 {
     public const REQUEST_RESET_PASSWORD_TOKEN = 'sylius.user.password_reset.request.token';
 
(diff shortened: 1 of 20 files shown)


src/Sylius/Bundle/UserBundle/UserEvents.php
```diff
@@ -13,7 +13,7 @@
 
 namespace Sylius\Bundle\UserBundle;
 
-final class UserEvents
+interface UserEvents
```
lchrusciel, line 16: I assume, that in this case, we should treat it as a BC Break :( /cc @pamil

vvasiloi: Only if someone, for an unknown reason, instantiated these classes in their code...

pamil: I'm OK as long as we mention that in the UPGRADE file.

Zales0123:
@lchrusciel a small change in psalm tests and it's good to go 👍

Zales0123:
Thanks, Łukasz! :1st_place_medal:

PR #11367, opened 2020-04-16
Check different shipping address field and display shipping address form

### LINKED ISSUE #11348: Going back to Addresses step is not fully functionnal
**Sylius version affected**: 1.7.0

**Description**  
If you fill the billing AND shipping addresses in the checkout, then you go to Shipping method, then you go back to the addresses: the checkbox "Use different address for shipping?" isn't checked anymore.

address_bug

**Steps to reproduce**  

See above.

**Possible Solution**  

I don't have any, I think it's because of the switch of the billing and shipping addresses.

vvasiloi:
When shipping and billing addresses are different

| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | fixes #11348
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Page/Shop/Checkout/AddressPage.php
+++ b/src/Sylius/Behat/Page/Shop/Checkout/AddressPage.php
@@ -301,7 +301,7 @@ class AddressPage extends SymfonyPage implements AddressPageInterface
             'login_password' => '[data-test-password-input]',
             'login_validation_error' => '[data-test-login-validation-error]',
             'next_step' => '[data-test-next-step]',
-            'shipping_address' => '#sylius-shipping-address',
+            'shipping_address' => '[data-test-shipping-address]',
             'shipping_address_book' => '[data-test-shipping-address] [data-test-address-book]',
             'shipping_city' => '[data-test-shipping-city]',
             'shipping_country' => '[data-test-shipping-country]',
--- a/src/Sylius/Behat/Page/Shop/Checkout/AddressPage.php
+++ b/src/Sylius/Behat/Page/Shop/Checkout/AddressPage.php
@@ -16,6 +16,7 @@ namespace Sylius\Behat\Page\Shop\Checkout;
 use Behat\Mink\Driver\Selenium2Driver;
 use Behat\Mink\Element\NodeElement;
 use Behat\Mink\Exception\ElementNotFoundException;
+use Behat\Mink\Exception\UnsupportedDriverActionException;
 use Behat\Mink\Session;
 use DMore\ChromeDriver\ChromeDriver;
 use FriendsOfBehat\PageObjectExtension\Page\SymfonyPage;
@@ -66,6 +67,21 @@ class AddressPage extends SymfonyPage implements AddressPageInterface
         );
 
         $billingAddressSwitch->check();
+    }
+
+    public function isDifferentShippingAddressChecked(): bool
+    {
+        return $this->getElement('different_shipping_address')->isChecked();
+    }
+
+    public function isShippingAddressVisible(): bool
+    {
+        try {
+            return $this->getElement('shipping_address')->isVisible();
+        } catch (UnsupportedDriverActionException $e) {
+            // it's visible by default and is being hidden with JS
+            return true;
+        }
     }
 
     public function checkInvalidCredentialsValidation(): bool
@@ -285,6 +301,7 @@ class AddressPage extends SymfonyPage implements AddressPageInterface
             'login_password' => '[data-test-password-input]',
             'login_validation_error' => '[data-test-login-validation-error]',
             'next_step' => '[data-test-next-step]',
+            'shipping_address' => '#sylius-shipping-address',
             'shipping_address_book' => '[data-test-shipping-address] [data-test-address-book]',
             'shipping_city' => '[data-test-shipping-city]',
             'shipping_country' => '[data-test-shipping-country]',
--- a/src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php
+++ b/src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php
@@ -40,7 +40,7 @@ final class AddressType extends AbstractResourceType
         if (null === $addressComparator) {
             @trigger_error(
                 sprintf(
-                    'Not passing an $addressComparator to %s constructor is deprecated since Sylius 1.7 and will be removed in Sylius 2.0.',
+                    'Not passing an $addressComparator to "%s" constructor is deprecated since Sylius 1.8 and will be impossible in Sylius 2.0.',
                     __CLASS__,
                 ),
                 \E_USER_DEPRECATED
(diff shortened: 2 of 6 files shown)


vvasiloi:
I started it from master, but then I noticed the issue has the "bug" label, so I switched to 1.7.

src/Sylius/Behat/Page/Shop/Checkout/AddressPage.php
```diff
@@ -284,6 +300,7 @@ protected function getDefinedElements(): array
             'login_password' => '[data-test-password-input]',
             'login_validation_error' => '[data-test-login-validation-error]',
             'next_step' => '[data-test-next-step]',
+            'shipping_address' => '#sylius-shipping-address',
```
pamil, line 303:
```suggestion
            'shipping_address' => '[data-test-shipping-address]',
```

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php
```diff
@@ -29,32 +30,61 @@
...
+    private $addressComparator;
+
+    public function __construct(string $dataClass, array $validationGroups = [], ?AddressComparatorInterface $addressComparator = null)
+    {
+        parent::__construct($dataClass, $validationGroups);
+
+        if (null === $addressComparator) {
+            @trigger_error(
+                sprintf(
+                    'Not passing an $addressComparator to %s constructor is deprecated since Sylius 1.7 and will be removed in Sylius 2.0.',
```
pamil, line 43:
```suggestion
                    'Not passing an $addressComparator to "%s" constructor is deprecated since Sylius 1.8 and will be impossible in Sylius 2.0.',
```

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php
```diff
@@ -29,32 +30,61 @@
...
                 $order = $event->getData();
-
                 /** @var OrderInterface $order */
                 Assert::isInstanceOf($order, OrderInterface::class);
 
                 $channel = $order->getChannel();
 
+                $differentBillingOrShippingAddressChecked = null !== $this->addressComparator
+                    && null !== $order->getBillingAddress() && null !== $order->getShippingAddress()
+                    && !$this->addressComparator->equal($order->getBillingAddress(), $order->getShippingAddress());
```
pamil, line 69: Could you extract this conditional to a private method? It's really tough to understand the flow now.

src/Sylius/Bundle/CoreBundle/Resources/config/services/checkout.xml
```diff
@@ -39,6 +39,7 @@
         <service id="sylius.form.type.checkout_address" class="Sylius\Bundle\CoreBundle\Form\Type\Checkout\AddressType">
             <argument>%sylius.model.order.class%</argument>
             <argument>%sylius.form.type.checkout_address.validation_groups%</argument>
+            <argument type="service" id="sylius.address_comparator"/>
```
pamil, line 42:
```suggestion
            <argument type="service" id="sylius.address_comparator" />
```

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php
```diff
@@ -29,32 +30,61 @@
...
+
                 $form
+                    ->add('differentBillingAddress', CheckboxType::class, [
+                        'mapped' => false,
+                        'required' => false,
+                        'label' => 'sylius.form.checkout.addressing.different_billing_address',
+                        'attr' => [
+                            'checked' => $differentBillingOrShippingAddressChecked,
```
pamil, line 77: I wonder whether we can just set the data here like:

```php
$data = $event->getData();
$data['differentBillingAddress'] = $data['differentShippingAddress'] = $differentBillingOrShippingAddressChecked;
$event->setData($data);
```

Could you check it out? This way we could leave these two fields outside of the event listener and achieve better backwards compatibility.

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php
```diff
@@ -29,32 +30,61 @@
...
-                'required' => false,
-                'label' => 'sylius.form.checkout.addressing.different_billing_address',
-            ])
-            ->add('differentShippingAddress', CheckboxType::class, [
-                'mapped' => false,
-                'required' => false,
-                'label' => 'sylius.form.checkout.addressing.different_shipping_address',
-            ])
-            ->addEventListener(FormEvents::PRE_SET_DATA, static function (FormEvent $event): void {
+            ->addEventListener(FormEvents::PRE_SET_DATA, function (FormEvent $event): void {
```
pamil, line 59: To be reverted:

```suggestion
            ->addEventListener(FormEvents::PRE_SET_DATA, static function (FormEvent $event): void {
```

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php
```diff
@@ -29,32 +30,61 @@
...
                 $form = $event->getForm();
                 $order = $event->getData();
-
```
pamil, line 51: To be reverted.

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php
vvasiloi: The data here holds the value of the checkbox, i.e. `value` attribute, not the `checked` attribute/property, so it won't work.

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php
vvasiloi: It can't be made static, because it uses `$this`

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php
vvasiloi: There will be a CS check error if I do that.

vvasiloi:
```
src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php:63
 ----------------------------------------------------------------------------------------------------------------
 Missing variable $order before or after the documentation comment.

(SlevomatCodingStandard\Sniffs\Commenting\InlineDocCommentDeclarationSniff.MissingVariable)
```

vvasiloi: I moved it before the `$order` is set.

vvasiloi: Now I get psalm errors https://psalm.dev/docs/running_psalm/issues/RedundantConditionGivenDocblockType/
The only way I found to not get any is to remove the docblock.
Is there another way?

pamil: I guess I'm OK with CS check error, we should disable that sniff anyway, it's more important to get Psalm passing.

pamil: So moving this docblock just before `Assert` would be the best solution here.

vvasiloi: Still get that psalm error.

vvasiloi: The docblock is here for intelligent code completion and the assert if for the actual type checking.
The only way I can think of to achieve both is this:
```
                if (!($order instanceof OrderInterface)) {
                    // or just return;
                    throw new \InvalidArgumentException(
                        'Expected an instance of %2$s. Got: %s',
                        OrderInterface::class,
                        get_class($order)
                    );
                }
```

lchrusciel:
Hey Victor :wave:

Could you rebase your fix to the newest master? Also, there is a small issue with the psalm checks.

vvasiloi:
@lchrusciel done!

lchrusciel:
Thanks, Victor! :tada:

Issue #10321, opened 2019-04-17
[Admin] Ability to override breadcrumbs


igormukhingmailcom:
Hi, guys.

At admin templates, we have `@SyliusAdmin/Macro/breadcrumb.html.twig` macro which is pretty cool and save much time. 

But how about make ability to override `_breadcrumbs.html.twig` in a way when breadcrumb items from parent template will be available in all children templates?

So for breadcrumbs `1 > 2 > 3 > 4` and `1 > 2 > 3 > 4 > 5` - I can make some base `_breadcrumb.html.twig` with `1 > 2 > 3 > 4` defined, and for second one - another which override first one with just one additional item `> 5`.

**Describe the proposed solution**
For now I have no practical/working example how to implement that...

igormukhingmailcom:
Any ideas?

Zales0123:
@kulczy maybe you have some thought about that? The idea itself sounds nice 🖖

vvasiloi:
Was looking into the macros for breadcrumbs and noticed an issue in :
https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/UiBundle/Resources/views/Macro/breadcrumb.html.twig
Parameter `root_path` is defined, but `root_url` is used.

Now back to the actual issue.
The easiest way to achieve what you described @igormukhingmailcom is to allow passing breadcrumbs to the templates where it is built:
```twig
{% include '@SyliusAdmin/Order/History/_breadcrumb.html.twig' with {'breadcrumbs': [{ label: 'app.ui.thing'|trans }]} %}
```
```twig
{% import '@SyliusAdmin/Macro/breadcrumb.html.twig' as breadcrumb %}

{% set breadcrumbs = [
        { label: 'sylius.ui.administration'|trans, url: path('sylius_admin_dashboard') },
        { label: 'sylius.ui.orders'|trans, url: path('sylius_admin_order_index') },
        { label: '#'~order.number, url: path('sylius_admin_order_show', {'id': order.id}) },
        { label: 'sylius.ui.history'|trans }
    ]|merge(breadcrumbs|default([]))
%}

{{ breadcrumb.crumble(breadcrumbs) }}
```

igormukhingmailcom:
@Zales0123, how about great idea proposed by @vvasiloi?

I will make PR based on `master` branch where `|merge(breadcrumbs|default([]))` will be added at every Core `_breadcrumb.html.twig`.

vvasiloi:
[...]

PR #11373, opened 2020-04-19
[WIP] Overridable admin breadcrumbs

igormukhingmailcom:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially #10321
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Crud/_breadcrumb.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Crud/_breadcrumb.html.twig
@@ -1,0 +1,23 @@
+{% import '@SyliusAdmin/Macro/breadcrumb.html.twig' as breadcrumb %}
+
+{% if breadcrumbs|default([])|length > 0 %}
+    {% set index_url = path(
+            configuration.vars.index.route.name|default(configuration.getRouteName('index')),
+            configuration.vars.index.route.parameters|default(configuration.vars.route.parameters|default({}))
+        )
+    %}
+
+    {% set breadcrumbs = [
+            { label: 'sylius.ui.administration'|trans, url: path('sylius_admin_dashboard') },
+            { label: (metadata.applicationName~'.ui.'~metadata.pluralName)|trans, url: index_url },
+        ]|merge(breadcrumbs|default([]))
+    %}
+{% else %}
+    {% set breadcrumbs = [
+            { label: 'sylius.ui.administration'|trans, url: path('sylius_admin_dashboard') },
+            { label: (metadata.applicationName~'.ui.'~metadata.pluralName)|trans },
+        ]|merge(breadcrumbs|default([]))
+    %}
+{% endif %}
+
+{{ breadcrumb.crumble(breadcrumbs) }}
--- a/src/Sylius/Bundle/UiBundle/Resources/views/Macro/breadcrumb.html.twig
+++ b/src/Sylius/Bundle/UiBundle/Resources/views/Macro/breadcrumb.html.twig
@@ -11,7 +11,7 @@
                 <i class="right chevron icon divider"></i>
             {% endif %}
         {% else %}
-            <a href="{{ root_url|default('/') }}" class="section">{{ root_label|default('sylius.ui.root'|trans) }}</a>
+            <a href="{{ root_path|default('/') }}" class="section">{{ root_label|default('sylius.ui.root'|trans) }}</a>
         {% endfor %}
     </div>
 {% endmacro %}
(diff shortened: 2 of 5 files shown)


igormukhingmailcom:
@Zales0123 Could someone give some feedback? 
If it looks good (and fit to be merged to master later) - I will add breadcrumbs for remaining admin pages.

PS I've used the same technique at private apps and it:
- looks clean
- saving tons of time
- decrease templates copy-paste
- decrease possibility to make typos on same crumbs but at different pages
- adding ability to extend breadcrumbs at plugins, where we need to add some new sub-pages to core admin pages

src/Sylius/Bundle/UiBundle/Resources/views/Macro/breadcrumb.html.twig
```diff
@@ -11,7 +11,7 @@
                 <i class="right chevron icon divider"></i>
             {% endif %}
         {% else %}
-            <a href="{{ root_url|default('/') }}" class="section">{{ root_label|default('sylius.ui.root'|trans) }}</a>
+            <a href="{{ root_path|default('/') }}" class="section">{{ root_label|default('sylius.ui.root'|trans) }}</a>
```
lchrusciel, line 14: Why have you changed this?

src/Sylius/Bundle/AdminBundle/Resources/views/Crud/_breadcrumb.html.twig
```diff
@@ -0,0 +1,23 @@
+{% import '@SyliusAdmin/Macro/breadcrumb.html.twig' as breadcrumb %}
+
+{% if breadcrumbs|default([])|length > 0 %}
```
lchrusciel, line 3: I'm not convinced, that this is the best condition, to append index if there will be anything in breadcrumbs. Won't it make some problems in the long term? Is there any possibility to simplify these if statements?

lchrusciel:
Seems nice and easier to extend indeed. I would vote in favor of merging it :) Just left two comments for clarification

src/Sylius/Bundle/UiBundle/Resources/views/Macro/breadcrumb.html.twig
vvasiloi: @lchrusciel  https://github.com/Sylius/Sylius/issues/10321#issuecomment-513726559

src/Sylius/Bundle/AdminBundle/Resources/views/Crud/_breadcrumb.html.twig
igormukhingmailcom: I agree. 

Better solution will be like this:

```twig
{# .../Macro/breadcrumb.html.twig #}
{% macro crumble(crumbs = {}) %}
    <div class="breadcrumbs-wrapper">
        <div class="container-fluid">
            <div class="breadcrumbs">
                <ul itemscope="" itemtype="http://schema.org/BreadcrumbList">
                    {% for crumb in crumbs %}
                        <li itemprop="itemListElement" itemscope="" itemtype="http://schema.org/ListItem">
                            {% if loop.last or crumb.url|default(null) is null %} {# if url is null - this should be rendered as label rather than link #}
                                <span itemprop="item">
                                    <span itemprop="name">{{ crumb.label }}</span>
                                </span>
                            {% else %}
                                <a href="{{ crumb.url }}" title="{{ crumb.label }}" itemprop="item">
                                    <span itemprop="name">{{ crumb.label }}</span>
                                </a>
                            {% endif %}
                            <meta itemprop="position" content="{{ loop.index }}">
                        </li>
                    {% endfor %}
                </ul>
            </div>
        </div>
    </div>
{% endmacro %}
```

With that little change, at macro, _breadcrumb templates now looks much cleaner:

```twig
{# .../shop/Account/_breadcrumb.html.twig #}
{% include 'shop/_breadcrumb.html.twig' with {
    'breadcrumbs': [
        { label: 'sylius.ui.my_account'|trans, url: breadcrumbs|default([])|length > 0 ? path('sylius_shop_account_dashboard') : null }
    ]|merge(breadcrumbs|default([]))
} %}
```

lchrusciel: Seems nicer. Is it somehow compatible with #11607?

igormukhingmailcom: Yes, current PR and 11607 is not conflicting as mine is about admin, when 11607 is about shop

igormukhingmailcom: I've added breadcrumbs example from shop in this thread, but just to show how to simplify if condition.

vvasiloi:
I remember this one... @igormukhingmailcom, would you like to finish it? I can help.

Rafikooo:
Hi Igor, thank you for the contribution. 
Please note that we have recently adopted a new approach using Sylius Template Events, which has replaced the previous twig-based solution. While we appreciate your effort in contributing this pull request, we have decided not to merge it as it is no longer necessary for the current version of Sylius. We encourage you to explore the Sylius Template Events feature and consider submitting pull requests that use this approach instead. Thank you for your understanding and continued support 🚀

PR #11378, opened 2020-04-20
Reorder order's sidebar in admin

jacquesbh:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

According to our experience in e-commerce payment and shipping are more important than the addresses when we look at an order.

Now we can see these informations first.
image

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/app/config.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/app/config.yml
@@ -670,39 +670,39 @@ sylius_ui:
             blocks:
                 before_customer_information_legacy:
                     template: "@SyliusUi/Block/_legacySonataEvent.html.twig"
-                    priority: 55
+                    priority: 90
                     context:
                         event: sylius.admin.order.show.before_customer_information
                 customer:
                     template: "@SyliusAdmin/Order/Show/_customer.html.twig"
-                    priority: 50
-                before_addresses_legacy:
-                    template: "@SyliusUi/Block/_legacySonataEvent.html.twig"
-                    priority: 45
-                    context:
-                        event: sylius.admin.order.show.before_addresses
-                addresses:
-                    template: "@SyliusAdmin/Order/Show/_addresses.html.twig"
-                    priority: 40
+                    priority: 80
                 before_payments_legacy:
                     template: "@SyliusUi/Block/_legacySonataEvent.html.twig"
-                    priority: 35
+                    priority: 70
                     context:
                         event: sylius.admin.order.show.before_payments
                 payments:
                     template: "@SyliusAdmin/Order/Show/_payments.html.twig"
-                    priority: 30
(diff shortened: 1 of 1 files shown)


CoderMaggie:
Hi @jacquesbh!
We will be reworking the Order show page UX in the coming weeks (months?) and we are planning to move the payments and shipments sections to the left (wider) column. That's why I would wait with your proposal until we make a decision on the improved UX with ou frontend magician @kulczy 🧙 

Is that fine for you? :) If you have any other remarks on this specific page I'd be more than welcome to hear it 🚀

jacquesbh:
No problem! Catch me on Slack if you want my opinion on this.
After 10 years in e-commerce, I may have ideas :p

jacquesbh:
@CoderMaggie Maybe we can merge this since there is no special move in 1.8 on this part? It's an UX improvement and I think it's great to have this in the meantime.

What do you think?

jacquesbh:
Ping @CoderMaggie @Zales0123 

It needs a rebase yes. Bit this is a real UI improvement.
It could have been merged long time ago.

What do we do?

Thanks!

Zales0123 (approved):
Seems good for me 👍 Would also like to see @lchrusciel @pamil opinion

src/Sylius/Bundle/AdminBundle/Resources/config/app/config.yml
```diff
@@ -670,39 +670,39 @@ sylius_ui:
             blocks:
                 before_customer_information_legacy:
                     template: "@SyliusUi/Block/_legacySonataEvent.html.twig"
-                    priority: 55
```
lchrusciel, line 673: Legacy events had `x5` priority in purpose, to make a distinction from the normal events IIRC. What is more, such a change would rearrange the whole panel for people who already injected some UI blocks into the existing structure. I'm ok with your proposal, but I would prefer to stay with old priorities (just change which block is assigned to the given priority,

jacquesbh: You mean for this it should be `95` and not `90`?

lchrusciel: I mean it should stay 55. We should revert your adjustment to change priorities range from 90 to 10 and bring back old one (55 to 5). After that, you may apply your changes just adjusting existing priorities instead of rebuilding the whole structure.

jacquesbh: ok! I'll do that.

jakubtobiasz:
Thanks @jacquesbh!
Due to differences between 1.12 and 1.13 I decided it's easier to open new PR. Thanks for your contribution :).

PR #11388, opened 2020-04-22
[Shipment] Remove shipment index and add show shipment from order button

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Screenshot 2020-04-24 at 11 44 14

Screenshot 2020-04-22 at 10 32 22

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/_shipment.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/_shipment.html.twig
@@ -1,5 +1,6 @@
 {% import '@SyliusUi/Macro/labels.html.twig' as label %}
 
+{% set cart = constant('Sylius\\Component\\Shipping\\Model\\Shipment::STATE_CART') %}
 {% set shipped = constant('Sylius\\Component\\Shipping\\Model\\Shipment::STATE_SHIPPED') %}
 
 <div class="item">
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/AdminBundle/Resources/views/Shipment/Show/_breadcrumb.html.twig
```diff
@@ -4,7 +4,6 @@
     { label: 'sylius.ui.administration'|trans, url: path('sylius_admin_dashboard') },
     { label: 'sylius.ui.orders'|trans, url: path('sylius_admin_order_index') },
     { label: '#' ~ shipment.order.number, url: path('sylius_admin_order_show', {'id': shipment.order.id}) },
-    { label: 'sylius.ui.shipments'|trans, url: path('sylius_admin_shipment_index') },
```
lchrusciel, line 7: Should be brought back. We need just to remove a link to shipment index

src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/_shipment.html.twig
```diff
@@ -27,7 +27,11 @@
             <p>{{ shipment.tracking }}</p>
         </div>
     {% endif %}
-
+    {% if shipment.state == shipped %}
```
lchrusciel, line 30: Should this button be available for shipments that are ready to be shipped? IMHO, yes, it should. /cc @CoderMaggie 

```suggestion
    {% if shipment.state is not cart %}
```

Or we can even consider usage of const, but that would be precedent.

src/Sylius/Bundle/AdminBundle/Resources/views/Order/Show/_shipment.html.twig
```diff
@@ -27,7 +27,11 @@
...
+        <div class="ui icon labeled tiny fluid button" style="margin-top: 10px;">
+            <i class="icon search"></i><a style="color: gray;" href="{{ path('sylius_admin_shipment_show', {'id': shipment.id}) }}" {{ sylius_test_html_attribute('shipment-show-button') }}>{{ 'sylius.ui.show'|trans }}</a>
+        </div>
```
GSadee, line 33: Did you check how it works? Because I guess that there will be a problem that only the text is clickable, not entire button

lchrusciel:
Thanks, Tomasz! :tada:

PR #11389, opened 2020-04-22
Switch to Doctrine/Migrations 3.0

pamil:
### Customisations
 - [x] Decorated migration factory to support migrations implementing `ContainerAwareInterface` - done in CoreBundle
 - [x] Added topology-based version comparator in CoreBundle to achieve the following order:
   - Sylius/Sylius
   - Plugins (Sylius/PluginSkeleton: Sylius/InvoicingPlugin, Sylius/RefundPlugin)
   - Sylius/Plus 
   - End-user application (Sylius/Sylius-Standard)

### Concerns
 - [x] **Q:** Currently, migrations are doubled in `app/migrations` and `src/Sylius/Bundle/CoreBundle/Migrations` - could we safely delete the ones in `app/migrations`?

    **A:** It should be possible to delete them in the next minor release - while migrating migrations database, it correctly assigns FQCN of the migration by its timestamp. However, we need to figure out how to upgrade existing apps, especially after we recommended copying migrations to the end-user's project before.

 - [x] **Q:** All the migrations exist in CoreBundle, even though some of them contain SQL queries related to AdminApiBundle - what's the best way to tackle that (one might not use AdminApiBundle but only CoreBundle)? Should we split them into two bundles? If so, how to do it and allow for a seamless upgrade?

    **A:** They can be safely extracted to AdminApiBundle as long as there are no migrations mixing both CoreBundle and AdminApiBundle. In that case, I'm not sure what's the best solution - probably skipping some of them if AdminApiBundle is not installed.

### Up next for upgrade to migrations 3.0-alpha
 - [ ] Sylius-Standard
 - [ ] PluginSkeleton
 - [x] InvoicingPlugin - https://github.com/Sylius/InvoicingPlugin/pull/175
 - [x] RefundPlugin - https://github.com/Sylius/RefundPlugin/pull/205
 - [x] Plus (requiring InvoicingPlugin and RefundPugin) - https://github.com/Sylius/Plus/pull/575

### TODO

1. [x] Extract related utilities (comparator and factory) to a standalone package (SyliusLabs/DoctrineMigrationsExtraBundle?) with a following configuration schema:
   ```yaml
   sylius_labs_doctrine_migrations_extra:
      migrations:
         "Sylius\InvoicingPlugin\Migrations": ["Sylius\Bundle\CoreBundle\Migrations"]
   ```
   Tests need to be added for both those services, configuration and DI extension.

2. [x] Use bundle notation for specifying migrations instead of prepending `doctrine_migrations` with relative paths, see doctrine/DoctrineMigrationsBundle#318.

3. [x] Decouple CoreBundle migrations from AdminApiBundle (either conditionally add them or move somehow to AdminApiBundle, while maintaining BC - after running all migrations on `master` branch, checking in this branch should result in no additional migrations)

4. [ ] Spread those changes to all the repositories mentioned above

### References
 - https://www.goetas.com/blog/multi-namespace-migrations-with-doctrinemigrations-30/
 - https://www.goetas.com/blog/released-doctrinemigrations-30-alpha/
 - https://github.com/doctrine/migrations/issues/967
 - https://github.com/doctrine/DoctrineMigrationsBundle/issues/314


src/Sylius/Bundle/CoreBundle/DependencyInjection/SyliusCoreExtension.php
```diff
@@ -70,24 +73,43 @@ public function prepend(ContainerBuilder $container): void
...
     {
-        if (!$container->hasExtension('hwi_oauth')) {
+        if (!$container->hasExtension('doctrine_migrations')) {
             return;
         }
 
-        $loader->load('services/integrations/hwi_oauth.xml');
+        $container->prependExtensionConfig('doctrine_migrations', [
+            'migrations_paths' => [
+                'Sylius\Bundle\CoreBundle\Migrations' => __DIR__ . '/../Migrations',
```
goetas, line 111: as of https://github.com/doctrine/DoctrineMigrationsBundle/pull/318 you can use also bundle relative paths

lchrusciel: Is it the feature above the main reason why you have added these mapping dynamically? Cleanup of this class could be merged anyway

src/Sylius/Bundle/CoreBundle/Migrations/Utility/TopologyBasedVersionComparator.php
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\Migrations\Utility;
```
lchrusciel, line 14:
```suggestion
namespace Sylius\Bundle\CoreBundle\Migrations\Comparator;
```
or
```suggestion
namespace Sylius\Bundle\CoreBundle\Doctrine\Migrations\Comparator;
```

The `utility` namespace can be a pandora box 📦

lchrusciel:
Anyway, that it looks really promising 👍 Can we proceed with the migration movement right now, to start preparation? 

BTW. Huge kudos to doctrine team 🎉

src/Sylius/Bundle/CoreBundle/DependencyInjection/SyliusCoreExtension.php
pamil: Yeah, it could be moved to regular config files we provide with CoreBundle now.

src/Sylius/Bundle/CoreBundle/Migrations/Utility/TopologyBasedVersionComparator.php
pamil: The idea is to either extract it to a standalone package or push those changes upstream.

pamil:
I guess we first need to figure out if we can move migrations to another directory (+ change their namespace) safely and make sure that we don't reference the moved migrations in some upgrade instructions.

pamil:
There's a new library on the market: https://github.com/SyliusLabs/DoctrineMigrationsExtraBundle

GSadee (approved):

Tomanhez (approved):

PR #11392, opened 2020-04-23
Add title on most pages

dunglas:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | yes 
| Related tickets | n/a
| License         | MIT

Having "Sylius" displayed in the `<title>` of every pages by default is weird and can be bad for SEO. I suggest to display the name of the current channel instead.


kulczy:
thanks @dunglas, good point, but it would be even better to add the title of the current page before the channel name

dunglas:
@kulczy I agree, but it can be done in other PRs right?

dunglas:
@kulczy actually, I added a new commit!

kulczy:
great work! thank you

lchrusciel:
Thank you, Kévin! :1st_place_medal:

PR #11371, opened 2020-04-17
[WIP] Add time ranges to dashboard statistics

kulczy:
bez nazwy

JSON structure:
```
{
    "summary": {
      "sales": "$10000",
      "paidOrders": "56",
      "customers": "78",
      "averageOrderValue": "$5500"
    },
    "chart": {
      "labels": [
        "01.01",
        "02.01"
      ],
      "values": [
        "344.26",
        "123.00"
      ]
    },
    "currency": {
      "prefix": "$",
      "suffix": ""
    }
  }
```

In `_statistics.html.twig` fill  `data-stats-url=""` with specyfic url's

and it should work

PR #11401, opened 2020-04-27
[Backend] Feeding sales statistics graph with date from chosen period

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    |yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets?   | #11371 
| License         | MIT

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -10,5 +10,11 @@ doctrine:
         driver: 'pdo_mysql'
         server_version: '5.7'
         charset: UTF8
-
         url: '%env(resolve:DATABASE_URL)%'
+    orm:
+        dql:
+            string_functions:
+                DAY: DoctrineExtensions\Query\Mysql\Day
+                HOUR: DoctrineExtensions\Query\Mysql\Hour
+                MONTH: DoctrineExtensions\Query\Mysql\Month
+                YEAR: DoctrineExtensions\Query\Mysql\Year
--- a/src/Sylius/Bundle/AdminBundle/Controller/DashboardController.php
+++ b/src/Sylius/Bundle/AdminBundle/Controller/DashboardController.php
@@ -76,10 +76,12 @@ final class DashboardController
         $data = ['statistics' => $statistics, 'channel' => $channel];
 
         if ($this->salesDataProvider !== null) {
-            $startDate = $request->query->get('startDate');
-            $endDate = $request->query->get('endDate');
-            $interval = $request->query->get('interval');
-            $dateFormat = $request->query->get('dateFormat');
+            // this data will be getting from UI after improve graph on dashboard
+            $startDate = (new \DateTime('first day of next month last year'));
+            $endDate = (new \DateTime('last day of this month'));
+            $interval = 'month';
+            $dateFormat = 'm';
+
             $data['sales_summary'] = $this
                 ->salesDataProvider
                 ->getSalesSummary($startDate, $endDate, $interval, $channel, $dateFormat)
--- a/src/Sylius/Bundle/AdminBundle/Controller/DashboardController.php
+++ b/src/Sylius/Bundle/AdminBundle/Controller/DashboardController.php
@@ -80,7 +80,7 @@ final class DashboardController
             $startDate = (new \DateTime('first day of next month last year'));
             $endDate = (new \DateTime('last day of this month'));
             $interval = 'month';
-            $dateFormat = 'm';
+            $dateFormat = 'n';
 
             $data['sales_summary'] = $this
                 ->salesDataProvider
--- a/src/Sylius/Bundle/AdminBundle/Controller/DashboardController.php
+++ b/src/Sylius/Bundle/AdminBundle/Controller/DashboardController.php
@@ -76,7 +76,14 @@ final class DashboardController
         $data = ['statistics' => $statistics, 'channel' => $channel];
 
         if ($this->salesDataProvider !== null) {
-            $data['sales_summary'] = $this->salesDataProvider->getLastYearSalesSummary($channel);
+            $startDate = $request->query->get('startDate');
+            $endDate = $request->query->get('endDate');
+            $interval = $request->query->get('interval');
+            $dateFormat = $request->query->get('dateFormat');
+            $data['sales_summary'] = $this
+                ->salesDataProvider
+                ->getSalesSummary($startDate, $endDate, $interval, $channel, $dateFormat)
+            ;
             $data['currency'] = $channel->getBaseCurrency()->getCode();
         }
 
--- a/src/Sylius/Component/Core/Dashboard/IntervalsConverter.php
+++ b/src/Sylius/Component/Core/Dashboard/IntervalsConverter.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\Dashboard;
+
+final class IntervalsConverter implements IntervalsConverterInterface
+{
+    public function getIntervals(
+        \DateTimeInterface $startDate,
+        \DateTimeInterface $endDate,
+        string $intervalName
+    ): \DatePeriod {
+        $supportedIntervals = [
+            'hour' => '1 hour',
+            'month' => '1 month',
+            'year' => '1 year',
+        ];
+
+        if (!array_key_exists($intervalName, $supportedIntervals)) {
+            throw new \InvalidArgumentException(sprintf('%s is a not supported interval', $intervalName));
+        }
+        if ($startDate >= $endDate) {
+            throw new \InvalidArgumentException('endDate should be later then startDate');
+        }
+
+        return new \DatePeriod($startDate, \DateInterval::createFromDateString($supportedIntervals[$intervalName]), $endDate);
+    }
+}
--- a/src/Sylius/Component/Core/Dashboard/SalesDataProvider.php
+++ b/src/Sylius/Component/Core/Dashboard/SalesDataProvider.php
@@ -26,11 +26,16 @@ final class SalesDataProvider implements SalesDataProviderInterface
     /** @var OrderRepositoryInterface|EntityRepository */
     private $orderRepository;
 
-    public function __construct(OrderRepositoryInterface $orderRepository)
+    /** @var IntervalsConverterInterface */
+    private $intervalsConverter;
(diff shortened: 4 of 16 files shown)


tests/Functional/SalesDataProviderTest.php
```diff
@@ -54,34 +54,84 @@ protected function setUp(): void
...
     /** @test */
-    public function it_provides_simple_year_sales_summary_grouped_per_month(): void
+    public function it_provides_year_sales_summary_grouped_by_year(): void
     {
-        $salesSummary = $this->getSummaryForChannel('CHANNEL');
-        $expectedMonths = $this->getExpectedMonths();
+        /** @var \DateTimeInterface $startDate */
+        $startDate = new \DateTime('2019-12-27 00:00:01');
+
+        /** @var \DateTimeInterface $startDate */
```
Tomanhez, line 62:
```suggestion
        /** @var \DateTimeInterface $endDate */
```

oallain: (This change has to be made several times lower in the file.)

src/Sylius/Component/Core/Dashboard/SalesDataProvider.php
```diff
@@ -29,35 +29,41 @@ public function __construct(EntityManagerInterface $entityManager)
...
+                DATE_FORMAT(checkout_completed_at, '%y.%m.%d %H') AS hour,
+                DATE_FORMAT(checkout_completed_at, '%y') AS year,
+            SUM(total) as total
             FROM sylius_order
             WHERE (channel_id = $channelId)
-            AND (checkout_completed_at BETWEEN '$startDate' AND '$endDate')
+            AND (checkout_completed_at BETWEEN '$formattedStartDate' AND '$formattedEndDate')
             AND (payment_state = 'paid')
-            GROUP BY date;"
+            GROUP BY '$interval', checkout_completed_at;"
```
pamil, line 56: `$interval` and other injected variables directly to this SQL query should be replaced with prepared statements to mitigate SQL injection risks.

tests/Functional/SalesDataProviderTest.php
pamil: Why is it even needed? Any IDE should guess that `\DateTime` implements `\DateTimeInteface`.

src/Sylius/Component/Core/Dashboard/SalesDataProvider.php
lchrusciel: There is a refactor proposed already in #11405. We should keep original commits if only possible

src/Sylius/Bundle/AdminBundle/Controller/DashboardController.php
```diff
@@ -76,7 +76,14 @@ public function indexAction(Request $request): Response
         $data = ['statistics' => $statistics, 'channel' => $channel];
 
         if ($this->salesDataProvider !== null) {
-            $data['sales_summary'] = $this->salesDataProvider->getLastYearSalesSummary($channel);
+            $startDate = $request->query->get('startDate');
+            $endDate = $request->query->get('endDate');
+            $interval = $request->query->get('interval');
+            $dateFormat = $request->query->get('dateFormat');
```
lchrusciel, line 82: Is it needed for the first iteration?

config/packages/doctrine.yaml
```diff
@@ -10,5 +10,11 @@ doctrine:
         driver: 'pdo_mysql'
         server_version: '5.7'
         charset: UTF8
-
         url: '%env(resolve:DATABASE_URL)%'
+    orm:
+        dql:
+            string_functions:
+                DAY: DoctrineExtensions\Query\Mysql\Day
+                HOUR: DoctrineExtensions\Query\Mysql\Hour
```
lchrusciel, line 18: Why do we need this precision in the provider? 

Also, can we put them in some order?
```suggestion
                HOUR: DoctrineExtensions\Query\Mysql\Hour
                DAY: DoctrineExtensions\Query\Mysql\Day
```

src/Sylius/Component/Core/Dashboard/SalesDataProvider.php
```diff
@@ -13,51 +13,60 @@
...
         $data = [];
-        foreach ($result as $item) {
-            $data[$item['date']] = (int) $item['total'];
+
+        foreach ($ordersTotals as $item) {
+            $data[(string) str_pad($item[$interval],2,"0",STR_PAD_LEFT)] = 0;
+        }
+        foreach ($ordersTotals as $item) {
+            $data[(string) str_pad($item[$interval],2,"0",STR_PAD_LEFT)] += (int) $item['total'];
         }
```
lchrusciel, line 68: This should provide the same end result: 
```suggestion
        foreach ($ordersTotals as $item) {
            $data[(string) str_pad($item[$interval],2,"0",STR_PAD_LEFT)] = (int) $item['total'];
        }
```

src/Sylius/Component/Core/Dashboard/SalesSummary.php
```diff
@@ -19,39 +19,41 @@
...
     public function __construct(
         \DateTimeInterface $startDate,
         \DateTimeInterface $endDate,
-        array $salesData
+        string $interval,
+        array $salesData,
+        string $dateFormat
     ) {
-        $period = new \DatePeriod($startDate, \DateInterval::createFromDateString('1 month'), $endDate);
+        $dateIntervals = new \DatePeriod($startDate, \DateInterval::createFromDateString(sprintf('1 %s', $interval)), $endDate);
```
lchrusciel, line 31: Do we restrict what kind of intervals do we support? Maybe we should have some kind of enum for them?

src/Sylius/Component/Core/Dashboard/SalesSummary.php
```diff
@@ -19,39 +19,41 @@
...
 
         /** @var \DateTimeInterface $date */
-        foreach ($period as $date) {
-            $periodName = $date->format('m.y');
+        foreach ($dateIntervals as $date) {
+            $periodName =$date->format($dateFormat);
```
lchrusciel, line 35:
```suggestion
            $periodName = $date->format($dateFormat);
```

config/packages/doctrine.yaml
AdamKasp: because our new graph proposed by @kulczy needs data aggregates by these intervals,
And also I have put it in alphabetical order (your suggestion was my second choice) - I am not sure what version is better :D

src/Sylius/Component/Core/Dashboard/SalesDataProvider.php
AdamKasp: i am afraid it is not the best suggestion :(
because your solution overriding the previous value, my solution is adding the next value to the previous one,
but you are right it should be refactored :)

lchrusciel: now I see it. `array_reduce` could do something like this but we need to filter data by intervals first anyway

src/Sylius/Component/Core/Dashboard/IntervalsConverter.php
```diff
@@ -0,0 +1,38 @@
...
+namespace Sylius\Component\Core\Dashboard;
+
+final class IntervalsConverter implements IntervalsConverterInterface
+{
+    public function getIntervals(
+        \DateTimeInterface $startDate,
+        \DateTimeInterface $endDate,
+        string $intervalName
+    ): \DatePeriod {
+        $supportedIntervals = [
```
lchrusciel, line 23: Could be defined as private const

src/Sylius/Component/Core/Dashboard/SalesDataProvider.php
```diff
@@ -13,51 +13,67 @@
...
+use Sylius\Component\Core\Repository\OrderRepositoryInterface;
 
 /**
  * @experimental
  */
 final class SalesDataProvider implements SalesDataProviderInterface
 {
-    /** @var EntityManagerInterface */
-    private $entityManager;
+    /** @var OrderRepositoryInterface|EntityRepository */
```
lchrusciel, line 26: Do we need both definitions because of `createQueryBuilder`? Maybe we should declare make it available through the main interface? But that would couple us with Doctrine ORM probably. /cc @pamil @Zales0123

Zales0123: If it's allowed by our BC policy then I suppose 👍

lchrusciel: Ref. https://docs.sylius.com/en/latest/book/organization/backward-compatibility-promise.html?%20policy#repositories-repository-interfaces

lchrusciel:
Thanks, Adam! :tada:

src/Sylius/Component/Core/Dashboard/SalesDataProvider.php
```diff
@@ -13,51 +13,67 @@
...
+    private $orderRepository;
 
-    public function __construct(EntityManagerInterface $entityManager)
+    /** @var IntervalsConverterInterface */
+    private $intervalsConverter;
+
+    public function __construct(OrderRepositoryInterface $orderRepository, IntervalsConverterInterface $intervalsConverter)
```
lchrusciel, line 32: It needs to be reflected in UPGRADE.md

src/Sylius/Component/Core/Dashboard/SalesSummary.php
```diff
@@ -19,39 +19,37 @@
 final class SalesSummary implements SalesSummaryInterface
 {
     /** @psalm-var array<string, string> */
-    private $monthsSalesMap = [];
+    private $intervalsSalesMap = [];
 
     public function __construct(
```
lchrusciel, line 24: Needs to be reflected in UPGRADE.md

src/Sylius/Component/Core/Dashboard/SalesDataProviderInterface.php
```diff
@@ -17,5 +17,11 @@
 
 interface SalesDataProviderInterface
 {
-    public function getLastYearSalesSummary(ChannelInterface $channel): SalesSummaryInterface;
+    public function getSalesSummary(
```
lchrusciel, line 20: Needs to be reflected in UPGRADE.md

src/Sylius/Component/Core/Dashboard/SalesSummaryInterface.php
```diff
@@ -15,7 +15,7 @@
 
 interface SalesSummaryInterface
 {
-    public function getMonths(): array;
+    public function getIntervals(): array;
```
lchrusciel, line 18: Needs to be reflected in UPGRADE.md

PR #11418, opened 2020-05-03
set default value for Channel taxCalculationStrategy

### LINKED ISSUE #11414: Channel class, default taxCalculationStrategy value
**Sylius version affected**: 1.6+

**Description**  
Field `taxCalculationStrategy` of `Sylius\Component\Core\Model\Channel` :
* has no default value in class [Channel.php#L36](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Model/Channel.php#L36)
* getter and setter uses `string|null` [Channel.php#L143](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Model/Channel.php#L143)
* field is **not** nullable in ORM [Channel.orm.xml#L21](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/CoreBundle/Resources/config/doctrine/model/Channel.orm.xml#L21)

**Steps to reproduce**  
Creation from the BO is not a problem since the field is of type select, the first value `order_items_based` is therefore used.
But creation by Factory or Api call, throw error.

**Possible Solution**  
Change in  [Channel.php#L36](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/Model/Channel.php#L36)
```diff
-    protected $taxCalculationStrategy;
+    protected $taxCalculationStrategy = 'order_items_based';
```

What do you think about that ?

I can do the PR to fix it :wink:

oallain:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #11414 
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Core/Factory/ChannelFactory.php
+++ b/src/Sylius/Component/Core/Factory/ChannelFactory.php
@@ -13,7 +13,7 @@ declare(strict_types=1);
 
 namespace Sylius\Component\Core\Factory;
 
-use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Channel\Model\ChannelInterface;
 use Sylius\Component\Resource\Factory\FactoryInterface;
 
 class ChannelFactory implements ChannelFactoryInterface
@@ -35,7 +35,7 @@ class ChannelFactory implements ChannelFactoryInterface
      */
     public function createNew(): ChannelInterface
     {
-        /** @var ChannelInterface $channel */
+        /** @var \Sylius\Component\Core\Model\ChannelInterface $channel */
         $channel = $this->decoratedFactory->createNew();
         $channel->setTaxCalculationStrategy($this->defaultCalculationStrategy);
 
--- a/src/Sylius/Component/Core/Factory/ChannelFactory.php
+++ b/src/Sylius/Component/Core/Factory/ChannelFactory.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\Factory;
+
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Resource\Factory\FactoryInterface;
+
+class ChannelFactory implements ChannelFactoryInterface
+{
+    /** @var FactoryInterface */
+    private $decoratedFactory;
+
+    /** @var string */
+    private $defaultCalculationStrategy;
+
+    public function __construct(FactoryInterface $decoratedFactory, string $defaultCalculationStrategy)
+    {
+        $this->decoratedFactory = $decoratedFactory;
+        $this->defaultCalculationStrategy = $defaultCalculationStrategy;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createNew(): ChannelInterface
+    {
+        /** @var ChannelInterface $channel */
+        $channel = $this->decoratedFactory->createNew();
+        $channel->setTaxCalculationStrategy($this->defaultCalculationStrategy);
+
+        return $channel;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createNamed(string $name): ChannelInterface
+    {
+        $channel = $this->createNew();
+        $channel->setName($name);
+
+        return $channel;
+    }
+}
--- a/src/Sylius/Component/Core/Factory/ChannelFactoryInterface.php
+++ b/src/Sylius/Component/Core/Factory/ChannelFactoryInterface.php
@@ -1,0 +1,22 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\Factory;
+
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Resource\Factory\FactoryInterface;
+
+interface ChannelFactoryInterface extends FactoryInterface
+{
+    public function createNamed(string $name): ChannelInterface;
+}
--- a/src/Sylius/Component/Core/Factory/ChannelFactoryInterface.php
+++ b/src/Sylius/Component/Core/Factory/ChannelFactoryInterface.php
@@ -13,10 +13,10 @@ declare(strict_types=1);
 
 namespace Sylius\Component\Core\Factory;
 
-use Sylius\Component\Core\Model\ChannelInterface;
-use Sylius\Component\Resource\Factory\FactoryInterface;
+use Sylius\Component\Channel\Factory\ChannelFactoryInterface as BaseChannelFactoryInterface;
+use Sylius\Component\Channel\Model\ChannelInterface;
 
-interface ChannelFactoryInterface extends FactoryInterface
+interface ChannelFactoryInterface extends BaseChannelFactoryInterface
 {
     public function createNamed(string $name): ChannelInterface;
 }
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ChannelExampleFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Factory/ChannelExampleFactory.php
@@ -16,7 +16,7 @@ namespace Sylius\Bundle\CoreBundle\Fixture\Factory;
 use Sylius\Bundle\CoreBundle\Fixture\OptionsResolver\LazyOption;
 use Sylius\Component\Addressing\Model\Scope as AddressingScope;
 use Sylius\Component\Addressing\Model\ZoneInterface;
-use Sylius\Component\Channel\Factory\ChannelFactoryInterface;
+use Sylius\Component\Core\Factory\ChannelFactoryInterface;
 use Sylius\Component\Core\Formatter\StringInflector;
 use Sylius\Component\Core\Model\ChannelInterface;
 use Sylius\Component\Core\Model\Scope;
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
@@ -119,6 +119,11 @@
             <argument type="service" id="sylius.custom_factory.address.inner" />
         </service>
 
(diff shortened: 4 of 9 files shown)


pamil:
Oh, I've just seen #11414. I'm not feeling comfortable with hardcoding the strategy, we should find some way to at least rely on some constraint or just use the first strategy that is available.

lchrusciel:
Neither do I, the current solution is a no-go right now. 

Maybe we should provide such config through our factory service? This is the default method of creating new objects in Sylius.

oallain:
Hi @lchrusciel @pamil ,

I create a ChannelFactory in Core, tell me if this is what you expected?

src/Sylius/Component/Core/Factory/ChannelFactory.php
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createNamed(string $name): ChannelInterface
+    {
+        /** @var ChannelInterface $channel */
+        $channel = $this->decoratedFactory->createNew();
+        $channel->setTaxCalculationStrategy('order_items_based');
+        $channel->setName($name);
```
lchrusciel, line 49: Not needed as it is already covered by the inner function

src/Sylius/Component/Core/Factory/ChannelFactory.php
```diff
@@ -0,0 +1,53 @@
...
+
+        return $channel;
+    }
```
mamazu, line 52:
```suggestion
class ChannelFactory implements ChannelFactoryInterface
{
    /** @var FactoryInterface */
    private $decoratedFactory;
    
    /** @var string */
    private $defaultCalculationStrategy;

    public function __construct(FactoryInterface $decoratedFactory, string $defaultCalculationStrategy)
    {
        $this->decoratedFactory = $decoratedFactory;
        $this->defaultCalculationStrategy = $defaultCalculationStrategy;
    }

    /**
     * {@inheritdoc}
     */
    public function createNew(): ChannelInterface
    {
        /** @var ChannelInterface $channel */
        $channel = $this->decoratedFactory->createNew();
        $channel->setTaxCalculationStrategy('order_items_based');

        return $channel;
    }

    /**
     * {@inheritdoc}
     */
    public function createNamed(string $name): ChannelInterface
    {
        $channel = $this->createNew();
        $channel->setName($name);

        return $channel;
    }
```

mamazu: That is probably what @pamil meant. And I agree with that.

lchrusciel: @mamazu proposal makes sense. However, I need to ask one more thing. Do we need `createNamed` in the first place? It seems to be unused in this scope and therefore we should remove it in the first place IMHO. Other than this, its good to go.

mamazu: Agreed. Because then we also don't need an extra interface for it.

oallain: Thanks for feedbacks, i'll change it :)

oallain: Hi,

I use `defaultCalculationStrategy` as parameter of ChannelFactory as proposed by @mamazu but i have keep function `createNamed()` because she is use in

https://github.com/Sylius/Sylius/blob/2875e9e2549eca13d53bc3cd3b6617da7d44ac04/src/Sylius/Component/Core/spec/Test/Services/DefaultChannelFactorySpec.php#L68

What do you think ?

mamazu: Well the `createNamed` doesn't really fit the style of Sylius. But other than that it looks good.

src/Sylius/Component/Core/Factory/ChannelFactory.php
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function createNamed(string $name): ChannelInterface
+    {
+        /** @var ChannelInterface $channel */
+        $channel = $this->decoratedFactory->createNew();
+        $channel->setTaxCalculationStrategy($this->defaultCalculationStrategy);
```
lchrusciel, line 52:
```suggestion
        $channel = $this->createNew();
```

src/Sylius/Component/Core/Factory/ChannelFactoryInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\Factory;
+
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Resource\Factory\FactoryInterface;
+
+interface ChannelFactoryInterface extends FactoryInterface
```
lchrusciel, line 19: Is this interface even needed, as we have the interface from component namespace? https://github.com/Sylius/Sylius/blob/7901f9a3c65f626a6b40fa9b7ffa8c3a2df18ab3/src/Sylius/Component/Channel/Factory/ChannelFactoryInterface.php#L21

If so, it should be extended rather than replaced.

src/Sylius/Component/Core/Factory/ChannelFactory.php
```diff
@@ -0,0 +1,57 @@
...
+
+class ChannelFactory implements ChannelFactoryInterface
+{
+    /** @var FactoryInterface */
+    private $decoratedFactory;
+
+    /** @var string */
+    private $defaultCalculationStrategy;
+
+    public function __construct(FactoryInterface $decoratedFactory, string $defaultCalculationStrategy)
```
lchrusciel, line 27: IMHO current ChannelFactory should be injected
```suggestion
    public function __construct(Sylius\Component\Channel\Factory\ ChannelFactoryInterface $decoratedFactory, string $defaultCalculationStrategy)
```

lchrusciel:
Sorry for pushing it back and forth, but not spotting the current factory implementation was my mistake from the early beginning

src/Sylius/Component/Core/Factory/ChannelFactoryInterface.php
oallain: IMHO, we need it while we don't have the covariance provided by PHP 7.4.

src/Sylius/Component/Core/Factory/ChannelFactory.php
oallain: I try but i got :

>   [TypeError]                                                                  
  Argument 1 passed to Sylius\Component\Core\Factory\ChannelFactory::__constr  
  uct() must implement interface Sylius\Component\Core\Factory\ChannelFactory  
  Interface, instance of Sylius\Component\Resource\Factory\Factory given, cal  
  led in /home/travis/build/Sylius/Sylius/var/cache/test_cached/ContainerJNMx  
  Xga/getChannelFactoryInterfaceService.php on line 12  

:disappointed:

oallain:
Hi @lchrusciel 

All looks good, but Travis failed.
Can you (or anyone) can help me ?

Thanks

mamazu:
The reason some of your tests may fail is because currently the master is broken. So just wait for them to fix it and then rebase again. (Or fix it yourself and make a separate pull request)

lchrusciel:
We can try to rebase now, it should work ;)

src/Sylius/Component/Core/Factory/ChannelFactoryInterface.php
lchrusciel: However, all of the current implementations, that are based on the previous interface cannot be changed, due to BC Policy. If we want to extend the previous implementation we need to extend the previous interface

oallain:
Hi @lchrusciel ,

Rebase is done, and "application" is green on Travis, but "doc" still red :disappointed:

mamazu:
Can you rebase again now?

oallain:
@mamazu ,

It's worse :wink:

mamazu:
Hm, but the master is green. Very odd.

oallain:
My target is Sylius:1.7, should i change the target?

mamazu:
As this is a new feature the master branch would be great.

oallain:
@mamazu 

I change target to master and it's green now :) 

Thanks !

oallain:
Hello @lchrusciel @mamazu 

I rebase (again) and now Travis is green :smile: (But Symfony/Insight is Pending :joy: )

Regards

mamazu:
Hm, someone from Sylius needs to restart this job.

lchrusciel:
Hey Olivier 👋 

I didn't find time to take over this PR, however, I won't merge it without change to https://github.com/Sylius/Sylius/pull/11418#discussion_r441566110. Right now this PR contains BC Break.

oallain:
Hello,

I am facing a dilemma and I don't know what/how to do :

1) Either the new `Sylius\Component\Core\Factory\ChannelFactoryInterface` extends `Sylius\Component\Channel\Factory\ChannelFactoryInterface` and in this case, the return types should be different but this is only possible with the covariance provided by PHP 7.4 - so BC with PHP < 7.4
2) Either the new `Sylius\Component\Core\Factory\ChannelFactoryInterface` return `Sylius\Component\Channel\Model\ChannelInterface` and not `Sylius\Component\Core\Model\ChannelInterface`.
2) Either keep this PR like this, however, all of the current implementations, that are based on the previous interface cannot be changed, like say Lukasz in https://github.com/Sylius/Sylius/pull/11418#discussion_r441566110 - so BC with older implementations

What is the best solution ?

oallain:
I just push with solution 2) :wink: 

Looking forward to your feedback.

mamazu:
I would suggest option 1 with the return type not being typehinted but just annotated in the doc type. Then psalm or phpstan or any other static code analysis tool can pick it up and it doesn't break things. (And this is also the best way to go in my opinion)

Option 2 is not great. Because the interface of the component is smaller than the interface of the core bundle (as far as I know) so this would also be sub-optimal.

Option 3: I don't know if there is already a branch in Sylius that is for 2.0 where those BC breaks will probably be bundled into. I also made a breaking change as a PR which also still has no place to go really.

lchrusciel:
Option 3 is a no-go for us. Option 2 is what we are doing for the last few years and this is what I would recommend. However, as @mamazu stated, probably improved option 1 (with docblock) is the best solution right now.

oallain:
Hello, @mamazu 

This rebase is :green_heart: (green) :tada:

lchrusciel:
may become messed-up.**

execute `git fetch origin && git reset "tax-calculation-strategy"` to update your local branch.
 
lchrusciel (approved):

lchrusciel:
Thanks, @oallain! :1st_place_medal:

PR #11200, opened 2020-03-10
[Shop][Product] Display product original price on product page

JulienLoison:
- [ ] Document the BC breaks

| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/8258
| License         | MIT

When original price is greater than the actual price of the product, it is displayed on the product page. 

image

PR #11423, opened 2020-05-05
[Shop][Product] Display product original price on product page (no bc break)

### LINKED ISSUE #8258: Original price not displayed in product listing

FlashPanther:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | fixes #8258
| License         | MIT

Follows up pull request #11200 by @JulienLoison , fixing backward compatibility break

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
+++ b/src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
@@ -13,17 +13,17 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\CoreBundle\Templating\Helper;
 
-use Sylius\Component\Core\Calculator\ProductVariantPriceCalculatorInterface;
+use Sylius\Component\Core\Calculator\ProductVariantPricesCalculatorInterface;
 use Sylius\Component\Core\Model\ProductVariantInterface;
 use Symfony\Component\Templating\Helper\Helper;
 use Webmozart\Assert\Assert;
 
 class PriceHelper extends Helper
 {
-    /** @var ProductVariantPriceCalculatorInterface */
+    /** @var ProductVariantPricesCalculatorInterface */
     private $productVariantPriceCalculator;
 
-    public function __construct(ProductVariantPriceCalculatorInterface $productVariantPriceCalculator)
+    public function __construct(ProductVariantPricesCalculatorInterface $productVariantPriceCalculator)
     {
         $this->productVariantPriceCalculator = $productVariantPriceCalculator;
     }
--- a/src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
+++ b/src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\CoreBundle\Templating\Helper;
 
+use Sylius\Component\Core\Calculator\ProductVariantPriceCalculatorInterface;
 use Sylius\Component\Core\Calculator\ProductVariantPricesCalculatorInterface;
 use Sylius\Component\Core\Model\ProductVariantInterface;
 use Symfony\Component\Templating\Helper\Helper;
@@ -20,10 +21,10 @@ use Webmozart\Assert\Assert;
 
 class PriceHelper extends Helper
 {
-    /** @var ProductVariantPricesCalculatorInterface */
+    /** @var ProductVariantPriceCalculatorInterface */
     private $productVariantPriceCalculator;
 
-    public function __construct(ProductVariantPricesCalculatorInterface $productVariantPriceCalculator)
+    public function __construct(ProductVariantPriceCalculatorInterface $productVariantPriceCalculator)
     {
         $this->productVariantPriceCalculator = $productVariantPriceCalculator;
     }
@@ -51,6 +52,7 @@ class PriceHelper extends Helper
     public function getOriginalPrice(ProductVariantInterface $productVariant, array $context): int
     {
         Assert::keyExists($context, 'channel');
+        Assert::isInstanceOf($this->productVariantPriceCalculator, ProductVariantPricesCalculatorInterface::class);
 
         return $this
             ->productVariantPriceCalculator
@@ -65,6 +67,7 @@ class PriceHelper extends Helper
     public function hasDiscount(ProductVariantInterface $productVariant, array $context): bool
     {
         Assert::keyExists($context, 'channel');
+        Assert::isInstanceOf($this->productVariantPriceCalculator, ProductVariantPricesCalculatorInterface::class);
 
         return $this->getOriginalPrice($productVariant, $context) > $this->getPrice($productVariant, $context);
     }
--- a/src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
+++ b/src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
@@ -45,6 +45,32 @@ class PriceHelper extends Helper
 
     /**
      * {@inheritdoc}
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function getOriginalPrice(ProductVariantInterface $productVariant, array $context): int
+    {
+        Assert::keyExists($context, 'channel');
+
+        return $this
+            ->productVariantPriceCalculator
+            ->calculateOriginal($productVariant, $context);
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function hasDiscount(ProductVariantInterface $productVariant, array $context): bool
+    {
+        Assert::keyExists($context, 'channel');
+
+        return $this->getOriginalPrice($productVariant, $context) > $this->getPrice($productVariant, $context);
+    }
+
+    /**
+     * {@inheritdoc}
      */
     public function getName(): string
     {
--- a/src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
+++ b/src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
@@ -17,7 +17,7 @@ use Sylius\Component\Core\Exception\MissingChannelConfigurationException;
 use Sylius\Component\Core\Model\ProductVariantInterface;
 use Webmozart\Assert\Assert;
 
-final class ProductVariantPriceCalculator implements ProductVariantPriceCalculatorInterface
+final class ProductVariantPriceCalculator implements ProductVariantPricesCalculatorInterface
 {
     /**
      * {@inheritdoc}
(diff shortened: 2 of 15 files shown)


vvasiloi:
I think there are ways of adding this feature without deprecating anything or adding any BC breaks.
1. Use the current `ProductVariantPriceCalculatorInterface` for a new implementation which will return the original price.
2. Introduce a new interface `ProductVariantOriginalPriceCalculatorInterface` and it's implementation.
3. Introduce a new interface `ProductVariantChannelPricingResolver` which will return an instance of `ChannelPricingInterface` for a given product variant and channel.
This can be used to complement 1st and/or 2nd option(s). It can be used directly in the view layer (through Twig extension), similar to `sylius_resolve_variant`.

I find the 3rd option more practical and flexible. An instance of `ChannelPricingInterface` can be easily used in the view layer for all kinds of calculations. 

Also, there's `ChannelPricingInterface::isPriceReduced` which can be used for/instead of `sylius_has_discount`

src/Sylius/Component/Core/Provider/ProductVariantsPricesProvider.php
```diff
@@ -13,18 +13,18 @@
...
 use Sylius\Component\Product\Model\ProductOptionValueInterface;
 
 final class ProductVariantsPricesProvider implements ProductVariantsPricesProviderInterface
 {
-    /** @var ProductVariantPriceCalculatorInterface */
+    /** @var ProductVariantPricesCalculatorInterface */
     private $productVariantPriceCalculator;
 
-    public function __construct(ProductVariantPriceCalculatorInterface $productVariantPriceCalculator)
+    public function __construct(ProductVariantPricesCalculatorInterface $productVariantPriceCalculator)
```
pamil, line 27: This is a BC break 🚨

src/Sylius/Component/Core/Promotion/Filter/PriceRangeFilter.php
```diff
@@ -13,16 +13,16 @@
...
 use Webmozart\Assert\Assert;
 
 final class PriceRangeFilter implements FilterInterface
 {
-    /** @var ProductVariantPriceCalculatorInterface */
+    /** @var ProductVariantPricesCalculatorInterface */
     private $productVariantPriceCalculator;
 
-    public function __construct(ProductVariantPriceCalculatorInterface $productVariantPriceCalculator)
+    public function __construct(ProductVariantPricesCalculatorInterface $productVariantPriceCalculator)
```
pamil, line 25: This is a BC break 🚨

src/Sylius/Component/Core/OrderProcessing/OrderPricesRecalculator.php
```diff
@@ -22,10 +22,10 @@
 
 final class OrderPricesRecalculator implements OrderProcessorInterface
 {
-    /** @var ProductVariantPriceCalculatorInterface */
+    /** @var ProductVariantPricesCalculatorInterface */
     private $productVariantPriceCalculator;
 
-    public function __construct(ProductVariantPriceCalculatorInterface $productVariantPriceCalculator)
+    public function __construct(ProductVariantPricesCalculatorInterface $productVariantPriceCalculator)
```
pamil, line 28: This is a BC break 🚨

pamil:
@vvasiloi 2nd idea looks like a good solution! It'd not introduce any more new abstractions, so the complexity would stay almost the same.

FlashPanther:
Thanks for your feedback @vvasiloi , @pamil .

I followed @lchrusciel comment : [here](https://github.com/Sylius/Sylius/pull/11200/files#r396510427)

**Idea 1 :** @JulienLoison proposed this in pull request #11200 , right ? 
**Idea 2 :** Yes, but some code will be duplicated ... 
**Idea 3 :** Maybe more complicated ... @vvasiloi "P.S. I can help with the implementation this presents interest." ( I appreciate !)

Which solution do I have to choose ? 

(Sorry, first pull request in this project). :smile:

vvasiloi:
> Idea 1 : @JulienLoison proposed this in pull request #11200 , right ?

@FlashPanther No, there's a new method in the current interface, instead of a new interface.

> @vvasiloi 2nd idea looks like a good solution! It'd not introduce any more new abstractions, so the complexity would stay almost the same.

@pamil It is actually a new abstraction, since it adds a new interface. Perhaps you were referring to the 1st, which reuses the current interface.

patrickallaert:
> > Idea 1 : @JulienLoison proposed this in pull request #11200 , right ?
> 
> @FlashPanther No, there's a new method in the current interface, instead of a new interface.

Correct, but that was flagged as a BC Break, hence the new proposal of @FlashPanther which was based on the commit of @JulienLoison introduced in #11200.

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
```diff
@@ -17,7 +17,7 @@
 use Sylius\Component\Core\Model\ProductVariantInterface;
 use Webmozart\Assert\Assert;
 
-final class ProductVariantPriceCalculator implements ProductVariantPriceCalculatorInterface
+final class ProductVariantPriceCalculator implements ProductVariantPricesCalculatorInterface
```
lchrusciel, line 20: We need to maintain both interfaces, while `ProductVariantPricesCalculatorInterface` should extend (and maybe duplicate) `ProductVariantPriceCalculatorInterface`. `ProductVariantPriceCalculatorInterface` should marked as deprecated.

Then everywhere `ProductVariantPriceCalculatorInterface` is used, we should check if we have `ProductVariantPriceCalculatorInterface` or `ProductVariantPricesCalculatorInterface` and allow use only functions that are allowed. 

Another possibility is to completely rewrite some classes, but still, we need to maintain current contracts and behaviors in old part of code. Probably, you should go with the solution proposed by @pamil here: https://github.com/Sylius/Sylius/pull/11423#issuecomment-624504678

JulienLoison: If `ProductVariantPricesCalculatorInterface` extends `ProductVariantPriceCalculatorInterface` and provide the `calculateOriginal`, old contracts should be maintained ? In that case why should we check usages ? 

If `ProductVariantPriceCalculator` implements `ProductVariantPricesCalculatorInterface` that extends `ProductVariantPriceCalculatorInterface`, would there be any BC breaks left ?

lchrusciel: Current interfaces(ProductVariantPriceCalculatorInterface) need to be left untouched. You can create a new one(ProductVariantPricesCalculatorInterface) and extend the current one. Therefore end implementation (if the previous contract is maintained, so the old interface is still preserved there) can have new functionality. 

Same, everywhere when the old, deprecated interface is used we can check if we are using the new one, and if so make new logic available. Every place like this needs a separate deprecation note. 

That's why @pamil proposition will be the easiest one probably.

src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
```diff
@@ -13,17 +13,17 @@
...
 use Webmozart\Assert\Assert;
 
 class PriceHelper extends Helper
 {
-    /** @var ProductVariantPriceCalculatorInterface */
+    /** @var ProductVariantPricesCalculatorInterface */
     private $productVariantPriceCalculator;
 
-    public function __construct(ProductVariantPriceCalculatorInterface $productVariantPriceCalculator)
+    public function __construct(ProductVariantPricesCalculatorInterface $productVariantPriceCalculator)
```
lchrusciel, line 26:
```suggestion
    public function __construct(ProductVariantPriceCalculatorInterface $productVariantPriceCalculator)
```

src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
```diff
@@ -13,17 +13,17 @@
...
-use Sylius\Component\Core\Calculator\ProductVariantPriceCalculatorInterface;
+use Sylius\Component\Core\Calculator\ProductVariantPricesCalculatorInterface;
 use Sylius\Component\Core\Model\ProductVariantInterface;
 use Symfony\Component\Templating\Helper\Helper;
 use Webmozart\Assert\Assert;
 
 class PriceHelper extends Helper
 {
-    /** @var ProductVariantPriceCalculatorInterface */
+    /** @var ProductVariantPricesCalculatorInterface */
```
lchrusciel, line 23:
```suggestion
    /** @var ProductVariantPriceCalculatorInterface */
```

src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
```diff
@@ -43,6 +43,32 @@ public function getPrice(ProductVariantInterface $productVariant, array $context
...
 
+    /**
+     * {@inheritdoc}
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function getOriginalPrice(ProductVariantInterface $productVariant, array $context): int
+    {
+        Assert::keyExists($context, 'channel');
+
```
lchrusciel, line 54:
```suggestion
        Assert::isNotInstanceOf($this->productVariantPriceCalculator, ProductVariantPricesCalculatorInterface::class); 
        
```

lchrusciel: Same below

src/Sylius/Component/Core/Calculator/ProductVariantPricesCalculatorInterface.php
```diff
@@ -0,0 +1,26 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\Calculator;
+
+use Sylius\Component\Core\Exception\MissingChannelConfigurationException;
+use Sylius\Component\Core\Model\ProductVariantInterface;
+
+interface ProductVariantPricesCalculatorInterface extends ProductVariantPriceCalculatorInterface
+{
+
```
lchrusciel, line 21:
```suggestion
```

src/Sylius/Bundle/CoreBundle/Templating/Helper/PriceHelper.php
FlashPanther: Didn't you meant :
`Assert::isInstanceOf($this->productVariantPriceCalculator, ProductVariantPricesCalculatorInterface::class); `
?

lchrusciel: Yes, of course :)

FlashPanther:
Of course ... Psalm analysis triggers errors because of deprecated class ... 
Am I wrong ? Do I have to change something ?

FlashPanther:
@pamil : ok for you ? (requested for review)

lchrusciel:
Thanks, tdevos! :tada:

lchrusciel:
If anyone would like to improve this feature, even more, we could add original prices to our fixtures, so they will be visible on demo/testing installation just after fixtures reload ;)

lchrusciel:
There is an additional case, that I didn't spot during code review - the price is not updated when the variant is changed. It means also, that it will not work if the first variant doesn't have an original price (even if the previous problem was fixed). @FlashPanther would you like to give it a try to fix these cases?

PR #11424, opened 2020-05-05
[Admin][Product] Ajax product search fails

### LINKED ISSUE #9364: Product association : ajax product search fails

Prometee:
| Q               | A
| --------------- | -----
| Branch?         |  master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9364
| License         | MIT

### DIFF
--- a/UPGRADE-1.8.md
+++ b/UPGRADE-1.8.md
@@ -85,3 +85,5 @@
 
 1. Service alias `Sylius\Component\Channel\Context\ChannelContextInterface` was changed from `sylius.context.channel.composite` to `sylius.context.channel`.
 The later is being decorated by `sylius.context.channel.cached` which caches the channel per request and reduces the amount of database queries.
+
+1. A `serialization_group` has been added to the route `sylius_admin_ajax_product_index` to avoid infinite loop or time out during this ajax request.
--- a/UPGRADE-1.8.md
+++ b/UPGRADE-1.8.md
@@ -86,4 +86,5 @@
 1. Service alias `Sylius\Component\Channel\Context\ChannelContextInterface` was changed from `sylius.context.channel.composite` to `sylius.context.channel`.
 The later is being decorated by `sylius.context.channel.cached` which caches the channel per request and reduces the amount of database queries.
 
-1. A `serialization_group` has been added to the route `sylius_admin_ajax_product_index` to avoid infinite loop or time out during this ajax request.
+1. A serialization group has been added to the route `sylius_admin_ajax_product_index` to avoid an infinite loop, or a
+time out during this ajax request (previously no serialization group was defined on this route).
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product.yml
@@ -41,5 +41,6 @@ sylius_admin_ajax_product_index:
         _controller: sylius.controller.product:indexAction
         _format: json
         _sylius:
+            serialization_groups: [Default]
             permission: true
             grid: sylius_admin_product
--- a/tests/Controller/AdminProductAjaxTest.php
+++ b/tests/Controller/AdminProductAjaxTest.php
@@ -1,0 +1,71 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Tests\Controller;
+
+use ApiTestCase\JsonApiTestCase;
+use Exception;
+use Symfony\Component\BrowserKit\Cookie;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
+
+class AdminProductAjaxTest extends JsonApiTestCase
+{
+    /**
+     * @test
+     *
+     * @throws Exception
+     */
+    public function it_denies_access_to_a_products_list_for_not_authenticated_user()
+    {
+        $this->client->request('GET', '/admin/ajax/products/');
+
+        $response = $this->client->getResponse();
+        $this->assertTrue($response->isRedirection());
+    }
+
+    /**
+     * @test
+     *
+     * @throws Exception
+     */
+    public function it_allows_to_get_a_products_list()
+    {
+        $this->loadFixturesFromFile('authentication/administrator.yml');
+        $this->loadFixturesFromFile('resources/products.yml');
+        $this->loadFixturesFromFile('resources/many_products.yml');
+
+        $this->authenticateAdminUser();
+
+        $this->client->request('GET', '/admin/ajax/products/');
+
+        $response = $this->client->getResponse();
+        $this->assertResponse($response, 'ajax/product/index_response', Response::HTTP_OK);
+    }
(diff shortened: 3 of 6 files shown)


tests/Controller/AdminProductAjaxTest.php
```diff
@@ -0,0 +1,71 @@
...
+        $session = self::$container->get('session');
+        $firewallName = 'admin';
+        $firewallContext = 'admin';
+        $token = new UsernamePasswordToken($user, null, $firewallName, $user->getRoles());
+        $session->set(sprintf('_security_%s', $firewallContext), serialize($token));
+        $session->save();
+
+        $cookie = new Cookie($session->getName(), $session->getId());
+        $this->client->getCookieJar()->set($cookie);
+    }
```
Prometee, line 70: This method could be refactored to an abstract class, this could help making more Ajax Controllers test classes in the futur.

tests/DataFixtures/ORM/authentication/administrator.yml
```diff
@@ -0,0 +1,8 @@
+Sylius\Component\Core\Model\AdminUser:
+    admin:
+        plainPassword: sylius
+        roles: [ROLE_ADMINISTRATION_ACCESS]
+        enabled: true
+        email: admin@sylius.com
+        username: sylius
+        localeCode: en_US
```
Prometee, line 8: If `en` is set then no products are retrieved

tests/Responses/ajax/product/index_response.json
```diff
@@ -0,0 +1,160 @@
...
+        },
+        "first": {
+            "href": "\/admin\/ajax\/products\/?page=1&limit=10"
+        },
+        "last": {
+            "href": "\/admin\/ajax\/products\/?page=3&limit=10"
+        },
+        "next": {
+            "href": "\/admin\/ajax\/products\/?page=2&limit=10"
+        }
```
Prometee, line 18: The only differences with `tests/Responses/product/index_response.json` are those lines, I don't know if there is a way to avoid creating two nearly same files.

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product.yml
```diff
@@ -41,5 +41,6 @@ sylius_admin_ajax_product_index:
         _controller: sylius.controller.product:indexAction
         _format: json
         _sylius:
+            serialization_groups: [Default]
```
pamil, line 44: Could you add an upgrade note to `UPGRADE-1.8.md` mentioning this change and its effects?

Prometee: Added it 😉

tests/Controller/AdminProductAjaxTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Sylius\Tests\Controller;
+
+use ApiTestCase\JsonApiTestCase;
+use Exception;
+use Symfony\Component\BrowserKit\Cookie;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
+
+class AdminProductAjaxTest extends JsonApiTestCase
```
lchrusciel, line 22: Could be finalized
```suggestion
final class AdminProductAjaxTest extends JsonApiTestCase
```

Prometee:
To fail the test it require more than just adding product related to another, it require also to change the php config : `memory_limit` or `max_execution_time` to a lower values.

Prometee:
I will try fail it with products related to another, because in my case I have only 20 products and the bug occurred.

lchrusciel:
Would be awesome to have failing test for that, to prevent regression in the future

Prometee:
I was not enable to set a timeout on the request but the test will fail if the `serialization_groups` is changed because the json result will be different.

For the record here is the failing test :
image

tests/Responses/ajax/product/index_response.json
```diff
@@ -0,0 +1,160 @@
...
+                "name": "Breaking bad mug",
+                "code": "MUG_BB",
+                "options": [],
+                "averageRating": 0,
+                "images": [],
+                "_links": {
+                    "self": {
+                        "href": "\/api\/v1\/products\/MUG_BB"
+                    }
+                }
```
lchrusciel, line 33: If we are allowing for a BC break here, I would go further, and reduce amount of data to the bare minimum. (perhaps only name and code?)

UPGRADE-1.8.md
```diff
@@ -85,3 +85,5 @@
 
 1. Service alias `Sylius\Component\Channel\Context\ChannelContextInterface` was changed from `sylius.context.channel.composite` to `sylius.context.channel`.
 The later is being decorated by `sylius.context.channel.cached` which caches the channel per request and reduces the amount of database queries.
+
+1. A `serialization_group` has been added to the route `sylius_admin_ajax_product_index` to avoid infinite loop or time out during this ajax request.
```
lchrusciel, line 89: Can you add a description how to bring old behavior?

tests/Responses/ajax/product/index_response.json
Prometee: To do this, the `serialization_groups` will have to be changed to a new one. So all Ajax requests should get a new `AjaxRequest` `validation_groups` to only retrieve the strict minimum info. Another PR should be done to target this thing I suppose.

Prometee: I have to retest with the `serialization_group` named `Autocomplete`, I already tested it a while ago and there was more problem using it, I will see if I can take care of it.

Prometee: If I try to change the `serialization_groups ` to `Autocomplete ` then the json result is `{}` because the `Hateoas\Representation\PaginatedRepresentation` owning the data is serializable only with the serialization group `Default`. Other ajax routes override the repository method so the data retrieved is a simple array and not a `PaginatedRepresentation`.

So if I add a repository method like `findAll` to the route definition it will returns a simple json `array` of `Product` without other info like `total`, `pages`, etc. And then the javascript [use to display those results](https://github.com/Sylius/Sylius/tree/master/src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-product-auto-complete.js) have to be deleted in favor of the wider [`jQuery.autoComplete`](https://github.com/Sylius/Sylius/tree/master/src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-auto-complete.js) function use with the other autocomplete, and maybe more than this change is needed. Less code but this will be a BC break.

So @lchrusciel, @pamil do I have to introduce this BC break ?

pamil: I'd stay with the current solution then, WDYT @lchrusciel?

pamil:
Thank you, Francis! :tada:

Issue #8984, opened 2017-11-24
Cannot capture payment when the en_US locale is disabled for the channel


aadmathijssen:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.0.4

When placing an order in the shop, it is not possible to capture a payment when the `en_US` locale is not assigned to the channel for which the order is placed. Example channel locale configuration which triggers this issue:

![sylius_admin_channels_locale_settings](https://user-images.githubusercontent.com/3796971/33204302-65232706-d103-11e7-977c-ad2baab60f5b.png)

When trying to place an order, the user is redirected to the `payment/capture` route, which results in a `NotFoundHttpException` being thrown with the folllowing message:

```
The "en_US" locale is unavailable in this channel.
```

The error occurs on line [77 of `src/Sylius/Bundle/ShopBundle/EventListener/NonChannelLocaleListener.php`](https://github.com/Sylius/Sylius/blob/v1.0.4/src/Sylius/Bundle/ShopBundle/EventListener/NonChannelLocaleListener.php#L77).

aadmathijssen:
Full stack trace:

```
Symfony\Component\HttpKernel\Exception\NotFoundHttpException:
The "en_US" locale is unavailable in this channel.

  at vendor/sylius/sylius/src/Sylius/Bundle/ShopBundle/EventListener/NonChannelLocaleListener.php:77
  at Sylius\Bundle\ShopBundle\EventListener\NonChannelLocaleListener->restrictRequestLocale(object(GetResponseEvent), 'kernel.request', object(TraceableEventDispatcher))
  at call_user_func(array(object(NonChannelLocaleListener), 'restrictRequestLocale'), object(GetResponseEvent), 'kernel.request', object(TraceableEventDispatcher))
     (vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/Debug/WrappedListener.php:104)
  at Symfony\Component\EventDispatcher\Debug\WrappedListener->__invoke(object(GetResponseEvent), 'kernel.request', object(ContainerAwareEventDispatcher))
  at call_user_func(object(WrappedListener), object(GetResponseEvent), 'kernel.request', object(ContainerAwareEventDispatcher))
     (vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php:212)
[...]

PR #7650, opened 2017-03-01
Make Payum routes locale-independent

pamil:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | yes |
| Related tickets | closes #7609 |
| License         | MIT |

PR #7609, opened 2017-02-24
[Payum] Remove locale attribute from payum routes

Arminek:
| Q               | A
| --------------- | ---
| Bug fix?        | no |
| New feature?    | no |
| BC breaks?      | no |
| License         | MIT |

PR #11426, opened 2020-05-06
Made: Payum links locale-based

### LINKED ISSUE #10942: [Payum] Authorize link not contain locale and gives an error

**Description**  
When we have multi-channel store (say `US` and `DK`) and second channel don't have default locale (say have only `da_DK`, but haven't `en_US`), then after pressing pay button and redirecting to `http://127.0.0.1:8000/payment/authorize/BMhXXkSNJ9vxR8umzBTv8uWmNQzD_ISbcJPxZ1YmM4s`, we see: `The "en_US" locale is unavailable in this channel.`

**Possible Solution**  
- Make payum links locale-based (prepend `/locale/`)?

Also, I believe default fixtures should have multiple channels out of the box - that way more miltichannel-related bugs will be visible at demo => they will be fixed.

igormukhingmailcom:
| Q               | A
| --------------- | -----
| Branch?         | 1.5
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10942
| License         | MIT

### DIFF
--- a/config/routes/sylius_shop.yaml
+++ b/config/routes/sylius_shop.yaml
@@ -6,6 +6,9 @@ sylius_shop:
 
 sylius_shop_payum:
     resource: "@SyliusShopBundle/Resources/config/routing/payum.yml"
+    prefix: /{_locale}
+    requirements:
+        _locale: ^[A-Za-z]{2,4}(_([A-Za-z]{4}|[0-9]{3}))?(_([A-Za-z]{2}|[0-9]{3}))?$
 
 sylius_shop_default_locale:
     path: /


Zales0123:
may become messed-up.**

execute `git fetch origin && git reset "patch-4"` to update your local branch.
 
lchrusciel (approved):
I'm wondering however, if we can test it somehow to avoid regression in the future

pamil (commented):
It looks like we had some reason for Payum links not being localised, that's why they're in `config/routes/sylius_shop.yaml` instead of inside SyliusShopBundle. Does anyone remember /cc @Zales0123 @lchrusciel @GSadee?

lchrusciel:
No, I don't. And I regret that we didn't write ADR back then :(

lchrusciel:
another question, would if it is a BC-Break, and probably it is, isn't it?

igormukhingmailcom:
Do not stale

vvasiloi:
The Payum routes were localized a few years ago, so you probably had a reason to remove the locale.
Here are some references, unfortunately there is not much context:
https://github.com/Sylius/Sylius/pull/7609
https://github.com/Sylius/Sylius/pull/7650

However, there's evidence that it causes issues without locale:
https://github.com/Sylius/Sylius/issues/10942
https://github.com/Sylius/Sylius/issues/8984

So maybe the problem is somewhere else?
I can't think of a reason these routes should be localized right now, but there might be some use-cases, so it should work both ways, just like the whole shop: https://docs.sylius.com/en/latest/cookbook/shop/disabling-localised-urls.html

I propose to move the routes back to shop bundle and make sure it works with and without locale.
I will write some scenarios later this week, if someone doesn't do it before.

jakubtobiasz:
Hi @igormukhingmailcom!
Thanks for contributing. As it's easy to adjust this part, we decided not to include it in Sylius. Anyway, when we merge this PR it won't affect end-projects as they have their routing config.

PR #11435, opened 2020-05-08
[Api]Add new-api regexp to shop firewall

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

If you would turn off new-api firewall, shop firewall catch new-api routing

### DIFF
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -1,7 +1,7 @@
 parameters:
     sylius.security.admin_regex: "^/admin"
     sylius.security.api_regex: "^/api"
-    sylius.security.shop_regex: "^/(?!admin|api/.*|api$|media/.*)[^/]++"
+    sylius.security.shop_regex: "^/(?!admin|api|new-api/.*|api$|media/.*)[^/]++"
     sylius.security.new_api_route: "/new-api"
     sylius.security.new_api_regex: "^%sylius.security.new_api_route%"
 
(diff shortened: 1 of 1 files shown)


config/packages/security.yaml
```diff
@@ -1,7 +1,7 @@
 parameters:
     sylius.security.admin_regex: "^/admin"
     sylius.security.api_regex: "^/api"
-    sylius.security.shop_regex: "^/(?!admin|api/.*|api$|media/.*)[^/]++"
+    sylius.security.shop_regex: "^/(?!admin|api|new-api/.*|api$|media/.*)[^/]++"
```
AdamKasp, line 4: this regex is related to our old API (shop and admin) imo if it is necessary we shoud add new parameter for new API, instead of your approach.
```suggestion
    sylius.security.new_api_regex: "^/(?!new-api/.*|api$|media/.*)[^/]++"
```

lchrusciel, line 4:
```suggestion
    sylius.security.shop_regex: "^/(?!admin|new-api|api/.*|api$|media/.*)[^/]++"
```

pamil: Btw. it feels like we should write down why shop regex should match `/api`, `/api/something` but not `/api/` - I've already forgotten.

pamil:

PR #11445, opened 2020-05-11
[Docs] Update installation requirements & add upgrade instructions for Plus

pamil:


docs/book/installation/sylius_plus_installation.rst
```diff
@@ -346,6 +346,12 @@ your application's ``.env`` file:
 
 **Phew! That's all, you can now run the application just like you usually do with Sylius (using Symfony Server for example).**
 
+Upgrading Sylius Plus
```
CoderMaggie, line 349: I already have a PR prepared for this "paragraph", a bit more extensive, as a separate chapter, similarily to upgrading chapter for OS.

So please omit it.

pamil: Is it already proposed? If not, I'd still go with this one and remove this section in your PR - better done than perfect.

CoderMaggie: nope, not proposed.

We can go then :D

PR #11458, opened 2020-05-12
[POC] Oauth supporting

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

This is first POC iteration of Oauth support,I borrowed huge part of logic from admin-api because it has Oauth implementation, In future we should think about extracting this logic to common place, Duplicate logic is redundant.

SwaggerUi configuration is not completed, not work properly yet
What we have now: 
Get authorization token by SwaggerUi:
Screenshot 2020-05-12 at 12 44 35
Screenshot 2020-05-12 at 12 44 50
When we are not login:
Screenshot 2020-05-12 at 12 45 18

For test we can test it by curl
Create Oauth client:
```
bin/console sylius:new-api:oauth-server:create-client \
    --grant-type="password" \
    --grant-type="refresh_token" \
    --grant-type="token"
```
You should get public id and secret:
```
A new client with public id 4x6s19u7vjeog4k80wo48444o04sw8cg4kswsk0w8gwk0csc0o, secret 4bww7ef520ysww8kcw4g0s8cwwogsgwo4cgg840soowgw80800 has been added
```
Next you must request token:
```
curl -X POST http://127.0.0.1:8000/api/oauth/v2/token \
    -d "client_id"=4x6s19u7vjeog4k80wo48444o04sw8cg4kswsk0w8gwk0csc0o \
    -d "client_secret"=4bww7ef520ysww8kcw4g0s8cwwogsgwo4cgg840soowgw80800 \
    -d "grant_type"=password \
    -d "username"=api@example.com \
    -d "password"=sylius-api
```
After get access token:
```
{"access_token":"NzA4NzhkM2RiZmUxY2UxYWIxZGJiNTQxZGM5YWI4YmRjNzQ1NDczOGVjNTkwMDczYzJkNTBlMWY4MDU1NGYxOQ",
"expires_in":3600,
"token_type":"bearer",
"scope":"api_access",
"refresh_token":"NzFkZWYwM2RjNWY5ZGQ3MDAyZTY2Njg2ZDMxMWVjYjM5Zjg4MGI2NGU4Zjc1MTM0ZjMzN2E5ZjE1ODExZDRhYw"}%
```
Now you can get for example administrators:
```
curl -X GET "http://127.0.0.1:8000/new-api/administrators?page=1" -H "accept: application/ld+json" -H "Authorization: Bearer NzA4NzhkM2RiZmUxY2UxYWIxZGJiNTQxZGM5YWI4YmRjNzQ1NDczOGVjNTkwMDczYzJkNTBlMWY4MDU1NGYxOQ"
```
```
{"@context":"\/new-api\/contexts\/Administrator","@id":"\/new-api\/administrators","@type":"hydra:Collection","hydra:member":[{"@id":"\/new-
...
```
If you bearer token was wrong:
```
{"error":"access_denied","error_description":"OAuth2 authentication required"}%
```
If you try get resource without token like:
```
curl -X GET "http://127.0.0.1:8000/new-api/administrators?page=1" -H "accept: application/ld+json"
```
You get:
```
{"error":"access_denied","error_description":"OAuth2 authentication required"}%
``` 

**Now we don't have separate login for admin and shop yet**


src/Sylius/Bundle/ApiBundle/Command/CreateClientCommand.php
```diff
@@ -0,0 +1,78 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class CreateClientCommand extends Command
+{
+    /** @var ClientManagerInterface|null */
+    private $clientManager;
+
+    public function __construct(?string $name = null, ClientManagerInterface $clientManager = null)
```
Zales0123, line 28: We should not allow to `null` the `$clientManager`. Moreover, we could take only this argument and pass the command name in `parent::__construct('name')` or (as it's done now) in the `configure()` method

src/Sylius/Bundle/ApiBundle/Command/CreateClientCommand.php
```diff
@@ -0,0 +1,78 @@
...
+        parent::__construct($name);
+
+        $this->clientManager = $clientManager;
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->setName('sylius:new-api:oauth-server:create-client')
+            ->setDescription('Creates a new client')
```
Zales0123, line 39: `Creates a new OAuth client for Sylius API`?

config/packages/security.yaml
```diff
@@ -54,44 +54,42 @@ security:
...
+#            json_login:
+#                check_path: "%sylius.security.new_api_route%/shop-user-authentication-token"
+#                username_path: email
+#                password_path: password
+#                success_handler: lexik_jwt_authentication.handler.authentication_success
+#                failure_handler: lexik_jwt_authentication.handler.authentication_failure
+#            guard:
+#                authenticators:
+#                    - lexik_jwt_authentication.jwt_token_authenticator
+#
```
Zales0123, line 86: Commented code is evil 🔥

src/Sylius/Bundle/ApiBundle/Command/CreateClientCommand.php
pamil:
```suggestion
    public function __construct(ClientManagerInterface $clientManager)
```

src/Sylius/Bundle/ApiBundle/Resources/config/app/config.yaml
```diff
@@ -15,6 +15,21 @@ api_platform:
         SM\SMException: 422
         Sylius\Bundle\ApiBundle\Exception\CannotRemoveCurrentlyLoggedInUser: 422
         Sylius\Bundle\ApiBundle\Exception\ShippingMethodCannotBeRemoved: 422
+# Configurations for Swagger still doesn't work properly
+    oauth:
+        enabled: true
+        clientId: '5sxomk2lm680w80w00s0kssc4so4cccg0ksokks4csc88skgo0'
```
lchrusciel, line 21: This values are hardcoded, however, should they be fetched from the results of the command you've written above? Is it the main problem, why OAuth is not working properly?

src/Sylius/Bundle/ApiBundle/Resources/config/app/config.yaml
```diff
@@ -15,6 +15,21 @@ api_platform:
...
+        type: 'oauth2'
+        flow: 'password'
+        tokenUrl: '/api/oauth/v2/token'
+        authorizationUrl: '/api/oauth/v2/auth'
+        scopes: [api_access]
+        
+fos_oauth_server:
+    service:
+        options:
+            supported_scopes: api_access
```
lchrusciel, line 32: Have you checked admin and shop scopes?

src/Sylius/Bundle/ApiBundle/Resources/config/app/config.yaml
Tomanhez: In next iteration it should be like global variables? for example`%oauth_client_id%` that would be assigned after running the command, I think problem was not here

PR #11463, opened 2020-05-13
[Admin] Use findOneBy() instead of findAll() to get the first channel

pamil:


lchrusciel:

PR #11482, opened 2020-05-17
[WIP] Default shipping and payment method resolver

vvasiloi:
Default shipping method resolver and default payment method resolver should use the corresponding collection resolver to get available methods to choose from.

Right now the logic is duplicated and in case a custom collection resolver uses a different logic it might happen that the default resolver will return a method which is not available.

Need suggestions for tests and on how to deprecate the current logic, especially the constructor arguments. Also, should I split it 2 PRs?

TODO:
- [x] Tests
- [x] Deprecation of current logic, including constructor arguments
- [ ] UPGRADE-*.md

### DIFF
--- a/src/Sylius/Component/Core/Resolver/DefaultPaymentMethodResolver.php
+++ b/src/Sylius/Component/Core/Resolver/DefaultPaymentMethodResolver.php
@@ -13,13 +13,13 @@ declare(strict_types=1);
 
 namespace Sylius\Component\Core\Resolver;
 
-use Sylius\Component\Core\Model\ChannelInterface;
 use Sylius\Component\Core\Model\PaymentInterface;
 use Sylius\Component\Core\Repository\PaymentMethodRepositoryInterface;
 use Sylius\Component\Payment\Exception\UnresolvedDefaultPaymentMethodException;
 use Sylius\Component\Payment\Model\PaymentInterface as BasePaymentInterface;
 use Sylius\Component\Payment\Model\PaymentMethodInterface;
 use Sylius\Component\Payment\Resolver\DefaultPaymentMethodResolverInterface;
+use Sylius\Component\Payment\Resolver\PaymentMethodsResolverInterface;
 use Webmozart\Assert\Assert;
 
 class DefaultPaymentMethodResolver implements DefaultPaymentMethodResolverInterface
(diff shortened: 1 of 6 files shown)


src/Sylius/Component/Core/Resolver/EligibleDefaultShippingMethodResolver.php
```diff
@@ -73,6 +114,8 @@ public function getDefaultShippingMethod(ShipmentInterface $shipment): ShippingM
     }
 
     /**
+     * @deprecated
```
lchrusciel, line 117: Not needed as the method is private

src/Sylius/Component/Core/Resolver/EligibleDefaultShippingMethodResolver.php
```diff
@@ -38,30 +38,71 @@ final class EligibleDefaultShippingMethodResolver implements DefaultShippingMeth
...
+        if (empty($shippingMethods)) {
+            throw new UnresolvedDefaultShippingMethodException();
+        }
+
+        return $shippingMethods[0];
+    }
 
-        $shippingMethods = $this->getShippingMethods($channel, $order->getShippingAddress());
+    /**
+     * @deprecated
```
lchrusciel, line 98: Not needed as the method is private

lchrusciel (commented):
Hey Victor! 

thanks for your contribution. 

I would say, we are fine with specs only. Another possibility is to write more complicated functional tests, but it may be tricky, as we will need to wire it up manually. /cc @Sylius/core-team

vvasiloi:
@lchrusciel any updates on this?

lchrusciel:
Let's go with this PR as it is, without splitting. Looking right now at the code, specs are a must. Here we need to cover at least changes to the logic. Also, we could easily test that some argument is not needed. From a high-level perspective, we could add one scenario for each case, where the resolver will return different objects than the current solution. If you have an idea for PHPUnit(which I mentioned previously) go for it, but it would just skip this part.

This is a nice addition, and I would like to merge it. Is there anything else I can help you with to push it forward?

vvasiloi:
@lchrusciel Thank you! There's one thing that I'm not sure how to do and I need help.
How do I deprecate passing of the 1st argument (the repository) and also deprecate not passing the second one without a BC break?
Considering that the 1st parameter's type and name must remain unchanged so it won't break service arguments auto-wiring.

lchrusciel:
I missed this case, and to be sure I wasn't aware about issue you've raised. However, I cannot see too many possibilities other than just message, that only the second argument can be passed and it will become the first one.

I'm considering also, that maybe some bind option could provide backward compatibility in case you raised. Then, we can remove typehint from the constructor, inject new service, bind this argument for autowiring and then in code just handle different behavior depending on injected service.

vvasiloi:
Don't stale.

src/Sylius/Component/Core/Resolver/EligibleDefaultShippingMethodResolver.php
vvasiloi: I added it to know what to remove in the next major version.

vvasiloi:
@lchrusciel can't believe that it's been more than a year since I opened this.
Time to get back to work.
Could you please take a look at it again?
I added a deprecation notice when passing the repository, a check for neither the repository nor the resolver  is passed and the specs.

src/Sylius/Component/Core/Resolver/DefaultPaymentMethodResolver.php
```diff
@@ -13,37 +13,59 @@
...
 
 class DefaultPaymentMethodResolver implements DefaultPaymentMethodResolverInterface
 {
     /** @var PaymentMethodRepositoryInterface */
     protected $paymentMethodRepository;
 
-    public function __construct(PaymentMethodRepositoryInterface $paymentMethodRepository)
-    {
+    /** @var PaymentMethodsResolverInterface|null */
+    private $paymentMethodsResolver;
```
stloyd, line 31: New code should use PHP 7.4
```suggestion
    private ?PaymentMethodsResolverInterface $paymentMethodsResolver;
```

src/Sylius/Component/Core/Resolver/DefaultPaymentMethodResolver.php
```diff
@@ -13,37 +13,59 @@
...
-        /** @var ChannelInterface $channel */
-        $channel = $subject->getOrder()->getChannel();
+        if (null !== $this->paymentMethodsResolver) {
+            $paymentMethods = $this->paymentMethodsResolver->getSupportedMethods($payment);
+        } else {
+            $paymentMethods = $this->paymentMethodRepository->findEnabledForChannel($channel);
+        }
 
-        $paymentMethods = $this->paymentMethodRepository->findEnabledForChannel($channel);
         if (empty($paymentMethods)) {
```
stloyd, line 69:
```suggestion
        if (!$paymentMethods) {
```

src/Sylius/Component/Core/Resolver/EligibleDefaultShippingMethodResolver.php
```diff
@@ -17,48 +17,104 @@
...
 
-        $shippingMethods = $this->getShippingMethods($channel, $order->getShippingAddress());
+    /**
+     * @throws UnresolvedDefaultShippingMethodException
+     */
+    private function getFromResolver(CoreShipmentInterface $shipment): ShippingMethodInterface
+    {
+        $shippingMethods = $this->shippingMethodsResolver->getSupportedMethods($shipment);
+
+        if (empty($shippingMethods)) {
```
stloyd, line 104:
```suggestion
        if (!$shippingMethods) {
```

src/Sylius/Component/Core/Resolver/DefaultPaymentMethodResolver.php
vvasiloi: Thanks! I forgot about this file.

src/Sylius/Component/Core/Resolver/DefaultPaymentMethodResolver.php
vvasiloi: Maybe `count($paymentMethods) === 0`?
Or use `reset($paymentMethods)` and check if it's `false` instead of relying on the numeric index (`$paymentMethods[0]`).

src/Sylius/Component/Core/Resolver/EligibleDefaultShippingMethodResolver.php
vvasiloi: Same as https://github.com/Sylius/Sylius/pull/11482/files#r688957860

jakubtobiasz:
Hi @vvasiloi!
Any plans to continue work on it? PR is old, and we'd like to close it if there are no further plans to finish it :D.

vvasiloi:
The question is what does it take for it to be finished?

jakubtobiasz:
@vvasiloi I dunno, you're the author 😂. I see there's a WIP in the title that's why I'm asking :D. Also there're conflicts.

vvasiloi:
Well, now I have to resolve the conflict, so yeah, it's WIP.

PR #11488, opened 2020-05-19
[Docs] Unify installation sections from Getting Started and Book

pamil:
Makes our docs more like Symfony docs, with articles grouped around topics.

CleanShot 2020-05-19 at 15 33 53@2x


docs/book/setup/installing-sylius-plus.rst
```diff
@@ -1,8 +1,7 @@
-
 .. rst-class:: plus-doc
 
-Sylius Plus Installation
-========================
+How to Install Sylius Plus?
```
CoderMaggie, line 3: why are you changing this?
We are in the book, we should not use the "How to" format, this is for cookbooks :D

And if you have to then at least:
```suggestion
How to install Sylius Plus?
```

docs/book/setup/upgrading-sylius-plus.rst
```diff
@@ -1,7 +1,7 @@
 .. rst-class:: plus-doc
 
-Upgrading Sylius Plus
-=====================
+How to Upgrade Sylius Plus Application?
```
CoderMaggie, line 3: I'm against this change, as above.

And if you have to then...
```suggestion
How to upgrade Sylius Plus?
```

docs/book/setup/upgrading-sylius.rst
```diff
@@ -1,8 +1,5 @@
-.. index::
-   single: Upgrading
-
-Upgrading
-=========
+How to Upgrade Sylius Application?
```
CoderMaggie, line 1: :(

docs/book/index.rst
```diff
@@ -16,18 +16,17 @@ Introduction aims to describe the philosophy of Sylius. It will also teach you a
...
 
-The installation chapter is of course a comprehensive guide to installing Sylius on your machine, but it also provides
+The setup chapter is of course a comprehensive guide to installing Sylius on your machine, but it also provides
 a general instruction on upgrading Sylius in your project.
 
 .. toctree::
-    :hidden:
-
-    installation/index
+    :titlesonly:
```
CoderMaggie, line 26: is it necessary?

docs/book/setup/installing-sylius-plus.rst
pamil: Mostly to make it look more like a cookbook - it's a how-to article, step-by-step guide for Sylius Plus installation, without explaining the details of the process. I'm OK with reverting it and I'll do that (so that we can tackle it independently and once for all in the future).

As for capitalising, I follow title capitalisation rules (https://capitalizemytitle.com/).

docs/book/setup/upgrading-sylius.rst
pamil: As mentioned above, I'll revert this change. However, "Upgrading" doesn't say much. I'll go with "Upgrading Sylius Application" - it states its purpose better.

docs/book/index.rst
pamil: Without it, we'd get sections like "Before installation" and "Learn more" listed.

docs/book/setup/installing-sylius-plus.rst
pamil: Anyway, my end goal is to get rid of cookbooks and embed them in topic-related book articles. Eg. looking at "Resource models" article and having "How to customise a model?", "How to make model translatable?" articles linked at the bottom.

CoderMaggie: But these are sentences not "titles" like for blog posts. And sentences are written like sentences, it's a common practice. We have never used title-case ever before and I would not introduce it just because we found out it exists :).

pamil: That's why I reverted this change - not to add irrelevant changes to the docs unification PR.

I could argue whether it is a title or not (well, it's an article after all) and whether never using title case is a good enough argument to refrain from using them. 

Our docs contain a recommendation for section names, https://docs.sylius.com/en/latest/book/contributing/documentation/standards.html#language-standards, so we should define it there.

PR #11497, opened 2020-05-22
[API][ShippingMethods] Api authorization tests for ShippingMethods

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Api paths:

- [x] GET `/new-api/shipping-methods`
- [x] POST `/new-api/shipping-methods`
- [x] GET `/new-api/shipping-methods/{id}`
- [x] PUT `/new-api/shipping-methods/{id}`
- [x] DELETE `/new-api/shipping-methods/{id}`
- [x] PATCH `/new-api/shipping-methods/{id}/archive`
- [x] PATH `/new-api/shipping-methods/{id}/restore`


tests/Controller/ShippingMethodsNewApiTest.php
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    private function getShippingMethodUrl(?ShippingMethodInterface $shippingMethod = null): string
+    {
+        $url = '/new-api/shipping-methods';
+
+        return $shippingMethod !== null ? sprintf('%s/%s', $url, $shippingMethod->getCode()) : $url;
+    }
+
+    private function loadFixturesAndGetShippingMethod(string $shippingMethodName): ShippingMethodInterface
```
AdamKasp, line 285: this method and next one do more than one thing, IMO they aren't the true SOLID method, maybe two separately methods will be better?

tests/Controller/ShippingMethodsNewApiTest.php
```diff
@@ -0,0 +1,311 @@
...
+
+use ApiTestCase\JsonApiTestCase;
+use Sylius\Component\Core\Model\AdminUserInterface;
+use Sylius\Component\Core\Model\ShippingMethodInterface;
+use Symfony\Component\HttpFoundation\Response;
+
+final class ShippingMethodsNewApiTest extends JsonApiTestCase
+{
+    static private $authorizedHeaderWithDenied = [
+        'HTTP_Authorization' => 'Bearer ' . 'wrong_token',
```
AdamKasp, line 24: why do you use concatenation here instead of single string?

tests/Controller/ShippingMethodsNewApiTest.php
Tomanhez: I think that in this simple tests, Joining this two methods reduces the amount of redundant lines. Maybe others give opinion? @lchrusciel @GSadee  ?

lchrusciel: From the perspective of the method alone I think it would be ok to leave it as it is. However taking into account line https://github.com/Sylius/Sylius/pull/11497/files#diff-f11056c5b09a5942258aa228edeb658dR56 or https://github.com/Sylius/Sylius/pull/11497/files#diff-f11056c5b09a5942258aa228edeb658dR145 I would say, that separation of concerns may be a good idea. So in the first method we should load fixtures data (which is covered by ApiTestCase, so maybe new method is not needed at all), and the second should fetch the shipping method from database.

tests/Controller/ShippingMethodsNewApiTest.php
```diff
@@ -0,0 +1,396 @@
...
+            'resources/shipping_methods.yml',
+        ]);
+
+        return $shippingMethods[$shippingMethodName];
+    }
+
+    private function logInAsAdminAndGetToken(): string
+    {
+        /** @var AdminUserInterface $admin */
+        $admin = $this->loadFixturesFromFile('authentication/api.yml')['admin'];
```
lchrusciel, line 383: This could be a part of the loading data step

PR #11499, opened 2020-05-22
[UI] added display of product variant original price on product page

JulienLoison:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

Follows up @lchrusciel's comment on #11423 about variant prices not updating on the product page or variant not displaying discount if the first variant of the product has no discount.

### DIFF
--- a/src/Sylius/Behat/Context/Setup/ProductContext.php
+++ b/src/Sylius/Behat/Context/Setup/ProductContext.php
@@ -879,6 +879,18 @@ final class ProductContext implements Context
     }
 
     /**
+     * @Given /^the ("[^"]+" product variant) has original price at ("[^"]+")$/
+     */
+    public function productVariantHasOriginalPrice(ProductVariantInterface $productVariant, int $price): void
+    {
+        /** @var ChannelInterface $channel */
+        $channel = $this->sharedStorage->get('channel');
+
+        $productVariant->getChannelPricingForChannel($channel)->setOriginalPrice($price);
+        $this->objectManager->flush();
+    }
+
+    /**
      * @Given the store has a product :productName in channel :channel
      * @Given the store also has a product :productName in channel :channel
      */
--- a/src/Sylius/Behat/Context/Ui/Shop/ProductContext.php
+++ b/src/Sylius/Behat/Context/Ui/Shop/ProductContext.php
@@ -306,7 +306,16 @@ final class ProductContext implements Context
      */
     public function iShouldSeeTheProductOriginalPrice($price)
     {
+        Assert::true($this->showPage->isOriginalPriceVisible());
         Assert::same($this->showPage->getOriginalPrice(), $price);
+    }
+
+    /**
+     * @Then I should not see any original price
+     */
+    public function iShouldNotSeeTheProductOriginalPrice(): void
+    {
+        Assert::false($this->showPage->isOriginalPriceVisible());
     }
 
     /**
--- a/src/Sylius/Component/Core/Provider/ProductVariantsPricesProvider.php
+++ b/src/Sylius/Component/Core/Provider/ProductVariantsPricesProvider.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sylius\Component\Core\Provider;
 
 use Sylius\Component\Core\Calculator\ProductVariantPriceCalculatorInterface;
+use Sylius\Component\Core\Calculator\ProductVariantPricesCalculatorInterface;
 use Sylius\Component\Core\Model\ChannelInterface;
 use Sylius\Component\Core\Model\ProductInterface;
 use Sylius\Component\Core\Model\ProductVariantInterface;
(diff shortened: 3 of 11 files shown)


src/Sylius/Behat/Context/Setup/ProductContext.php
```diff
@@ -878,6 +878,18 @@ public function productHasVariant(ProductInterface $product, string $productVari
         $this->createProductVariant($product, $productVariantName, $priceValue, $code, $channel, null, true, $currentStock);
     }
 
+    /**
+     * @Given /^the ("[^"]+" product variant) has original price at ("[^"]+")$/
+     */
+    public function productVariantHasOriginalPrice(ProductVariantInterface $productVariant, int $price)
```
Zales0123, line 884: Missing `: void`

src/Sylius/Behat/Context/Ui/Shop/ProductContext.php
```diff
@@ -306,9 +306,18 @@ public function iShouldSeeTheProductPrice($price)
...
     public function iShouldSeeTheProductOriginalPrice($price)
     {
+        Assert::true($this->showPage->isOriginalPriceVisible());
         Assert::same($this->showPage->getOriginalPrice(), $price);
     }
 
+    /**
+     * @Then I should not see any original price
+     */
+    public function iShouldNotSeeTheProductOriginalPrice()
```
Zales0123, line 316: Here as well

src/Sylius/Component/Core/spec/Provider/ProductVariantsPricesProviderSpec.php
```diff
@@ -45,7 +46,7 @@ function it_provides_array_containing_product_variant_options_map_with_correspon
         ProductVariantInterface $blackSmallTShirt,
         ProductVariantInterface $whiteLargeTShirt,
         ProductVariantInterface $whiteSmallTShirt,
-        ProductVariantPriceCalculatorInterface $productVariantPriceCalculator
+        ProductVariantPricesCalculatorInterface $productVariantPriceCalculator
```
Zales0123, line 49: We should also change it in the `let` method (it fails the build, as far as I see)

src/Sylius/Component/Core/Provider/ProductVariantsPricesProvider.php
```diff
@@ -53,7 +55,15 @@ private function constructOptionsMap(ProductVariantInterface $variant, ChannelIn
             $optionMap[$option->getOptionCode()] = $option->getCode();
         }
 
-        $optionMap['value'] = $this->productVariantPriceCalculator->calculate($variant, ['channel' => $channel]);
+        $price = $this->productVariantPriceCalculator->calculate($variant, ['channel' => $channel]);
+        $optionMap['value'] = $price;
+
+        Assert::isInstanceOf($this->productVariantPriceCalculator, ProductVariantPricesCalculatorInterface::class);
```
Zales0123, line 61: Maybe it would be safer to use `if` rather than assertion here? **Technically** it's still possible to pass the old `ProductVariantPriceCalculatorInterface` in the constructor (for example if someone overrode the service declaration)

src/Sylius/Component/Core/Provider/ProductVariantsPricesProvider.php
```diff
@@ -53,7 +54,16 @@ private function constructOptionsMap(ProductVariantInterface $variant, ChannelIn
...
+        $price = $this->productVariantPriceCalculator->calculate($variant, ['channel' => $channel]);
+        $optionMap['value'] = $price;
+
+        if ($this->productVariantPriceCalculator instanceof ProductVariantPricesCalculatorInterface) {
+            $originalPrice = $this->productVariantPriceCalculator->calculateOriginal($variant, ['channel' => $channel]);
+
+            if ($originalPrice > $price) {
+                $optionMap['original-price'] = $originalPrice;
+            }
+        }
```
Zales0123, line 66: As I don't like nested `if` statements, I would go with:

```php
if (!$this->productVariantPriceCalculator instanceof ProductVariantPricesCalculatorInterface) {
    return $optionMap;
}

if ($originalPrice > $price) {
    $optionMap['original-price'] = $originalPrice;
}

return $optionMap;
```

but it's not a blocker for me 🖖

lchrusciel:
Thanks, Julien! :1st_place_medal:

src/Sylius/Component/Core/Provider/ProductVariantsPricesProvider.php
JulienLoison: Has @lchrusciel mentioned on the original PR, I'll try to add some original prices fixtures. I'll fix those nested `if` as well with the same PR. 

Thanks for the review !

PR #11500, opened 2020-05-24
[Maintenance] Bump up minor versions of requirements

pamil:
Bumping up `sylius/sylius` requirements, it should speed up resolving dependencies. Also, locking Psalm to a specific version to avoid unrelated build failures.


Zales0123:
Thank you, Kamil! :tada:

PR #11526, opened 2020-05-28
Upgrade order fields from integer to bigint

### LINKED ISSUE #9639: items_total in sylius_order table column causes 500 internal error
**Sylius version affected**: 1.1

**Description**  

the items_total field has limited size int(11) , thus cause 500 internal error with high values
**Steps to reproduce**  

add 70 unit of some item to cart  with unit price 15000000 through sylius resftul api

**Possible Solution**  

items_total field type or value should be changed or if you have restrictions, there should be validation rather than internal error.

arti0090:
| Q               | A
| --------------- | -----
| Branch?         |  master 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #9639
| License         | MIT

### DIFF
--- a/app/migrations/Version20200528103816.php
+++ b/app/migrations/Version20200528103816.php
@@ -1,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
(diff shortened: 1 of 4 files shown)


arti0090:
Note : I would also consider changing ID fields for bigint but this is for another PR

pamil:
Does it really work? We'd need some test for those huge prices. IIRC `bigint` database type casts to `string` PHP type, while `int` database type casts to `int` PHP type, which might cause issues.

PR #11529, opened 2020-05-28
[RFC] Using tabs for channel pricing

mamazu:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

## Proposal
Channel prices in a simple product get displayed on the product page under the list of channels. So I propose that they will be made into a tabbed view so that they are more easy to manage. If you have a lot of channels (where the product is not active for example) you have a lot of values to scroll through.

### Before
screenshot_before

### After
screenshot_after


src/Sylius/Bundle/AdminBundle/Resources/views/Product/Tab/_details.html.twig
```diff
@@ -29,13 +29,39 @@
...
             <div class="column">
                 {{ form_row(form.channels) }}
-
-                {% if product.simple %}
-                    <h4 class="ui dividing header">{{ 'sylius.ui.pricing'|trans }}</h4>
-                    {{ form_row(form.variant.channelPricings, {'label': false}) }}
-                {% endif %}
             </div>
         </div>
+        {% if product.simple %}
```
lchrusciel, line 34: Should be duplicated on variant page as well. Or we can try to extract common logic from this file and add it to variants as well.

mamazu: The variant page doesn't have a form though.

mamazu: Going to do that as well.

mamazu: I extracted it as a template. But I am not sure where to put it. I just for now put it in the Product folder, but I am open for suggestions.

lchrusciel:
Hey Max,

can you rebase once again? Tests are probably failing due to bc break at Symfony end (see #11542)

src/Sylius/Behat/Context/Ui/Admin/ManagingProductVariantsContext.php
```diff
@@ -129,17 +130,17 @@ public function iEnableItsTracking()
...
      */
-    public function iSetItsPriceTo(?string $price = null, $channelName = null)
+    public function iSetItsPriceTo(?string $price = null, ?ChannelInterface $channel = null)
     {
-        $this->createPage->specifyPrice($price ?? '', $channelName ?? (string) $this->sharedStorage->get('channel'));
+        $this->createPage->specifyPrice($price ?? '', $channel ?? $this->sharedStorage->get('channel'));
     }
 
     /**
      * @When /^I set its original price to "(?:€|£|\$)([^"]+)" for "([^"]+)" channel$/
```
lchrusciel, line 139:
```suggestion
     * @When /^I set its original price to "(?:€|£|\$)([^"]+)" for ("([^"]+)" channel)$/
```

src/Sylius/Behat/Context/Ui/Admin/ManagingProductVariantsContext.php
```diff
@@ -129,17 +130,17 @@ public function iEnableItsTracking()
      * @When /^I set its(?:| default) price to "(?:€|£|\$)([^"]+)" for "([^"]+)" channel$/
```
lchrusciel, line 130:
```suggestion
     * @When /^I set its(?:| default) price to "(?:€|£|\$)([^"]+)" for ("([^"]+)" channel)$/
```

mamazu:
I rebased the feature on a working master. So if it breaks now this should be my fault.

mamazu:
One thing I am not sure about is the message about the product being not active in this channel. Which would in turn also solve the issue #11559 .

lchrusciel:
Thanks, @mamazu! :tada:

lchrusciel:
According to your doubts, let's iterate over it on the master. It already looks much better  :D

PR #11535, opened 2020-05-30
Resolve awaiting_payment state in resolver

kayue:
OrderPaymentStateResolver should resolve awaiting_payment state when there is payment under processing. This is needed to handle asynchronous payment confirmation.

> For asynchronous payment methods, it can take up to several days to confirm whether the payment has been successful. During this time, the payment cannot be guaranteed. The status of the payment’s Charge object is initially set to pending, until the payment has been confirmed as successful or failed. ACH debits are an example of an asynchronous method: with these debits, it takes a few days to confirm that the payment has succeeded.

https://stripe.com/docs/sources#synchronous-or-asynchronous-confirmation

| Q               | A
| --------------- | -----
| Branch?         | 1.6
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT


pamil:
Thank you, Ka! :1st_place_medal:

PR #11545, opened 2020-06-03
Add priority to payment gateways

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

`sylius.gateway_factories` parameter stays the same, priority is only used to order gateways - and therefore to display them in proper order in the admin panel.

Zrzut ekranu 2020-06-3 o 16 45 41

If priority is not set, it's `0` by default.

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/DependencyInjection/Compiler/RegisterGatewayConfigTypePass.php
+++ b/src/Sylius/Bundle/PayumBundle/DependencyInjection/Compiler/RegisterGatewayConfigTypePass.php
@@ -56,11 +56,11 @@ final class RegisterGatewayConfigTypePass implements CompilerPassInterface
             return $secondGateway['priority'] - $firstGateway['priority'];
         });
 
+        $sortedGatewayFactories = [];
         foreach ($gatewayFactories as $key => $factory) {
-            $gatewayFactories[$factory['type']] = $factory['label'];
-            unset($gatewayFactories[$key]);
+            $sortedGatewayFactories[$factory['type']] = $factory['label'];
         }
 
-        $container->setParameter('sylius.gateway_factories', $gatewayFactories);
+        $container->setParameter('sylius.gateway_factories', $sortedGatewayFactories);
     }
 }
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/PayumBundle/DependencyInjection/Compiler/RegisterGatewayConfigTypePass.php
```diff
@@ -47,8 +51,15 @@ public function process(ContainerBuilder $container): void
...
+        usort($gatewayFactories, function (array $firstGateway, array $secondGateway): int {
+            return $secondGateway['priority'] - $firstGateway['priority'];
+        });
+
+        foreach ($gatewayFactories as $key => $factory) {
+            $gatewayFactories[$factory['type']] = $factory['label'];
+            unset($gatewayFactories[$key]);
+        }
 
         $container->setParameter('sylius.gateway_factories', $gatewayFactories);
```
pamil, line 64: Could we add an integration test for this compiler pass as we have similar ones for other passes? This way we can ensure it's sorted correctly.

src/Sylius/Bundle/PayumBundle/DependencyInjection/Compiler/RegisterGatewayConfigTypePass.php
```diff
@@ -47,8 +51,15 @@ public function process(ContainerBuilder $container): void
...
-        $gatewayFactories = array_merge($gatewayFactories, ['offline' => 'sylius.payum_gateway_factory.offline']);
-        ksort($gatewayFactories);
+
+        usort($gatewayFactories, function (array $firstGateway, array $secondGateway): int {
+            return $secondGateway['priority'] - $firstGateway['priority'];
+        });
+
+        foreach ($gatewayFactories as $key => $factory) {
+            $gatewayFactories[$factory['type']] = $factory['label'];
+            unset($gatewayFactories[$key]);
```
pamil, line 61: This is kinda hard to follow, would be easier if it would fill the data to another variable instead of filling and deleting from the same one at the same time.

pamil:
Thanks, Mateusz! :tada:

PR #11329, opened 2020-04-05
Make taxon comparable

oallain:
| Q               | A
| --------------- | -----
| Branch?         | 1.6
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #10797 
| License         | MIT

Is this good for you?
Do you think I should do the same for all resources (as said in the issue commands)?

PR #11580, opened 2020-06-15
make product variant comparable

### LINKED ISSUE #11487: Nesting level too deep fatal error when changing variant position
**Sylius version affected**: 1.7.3

**Description**  
When I programmatically change the position of an existing product variant, I get a fatal error "Nesting level too deep - recursive dependency? in vendor/gedmo/doctrine-extensions/lib/Gedmo/Sortable/SortableListener.php on line 471"

**Steps to reproduce**  
(pseudocode)
```
$product = productRepository->findOneBy(['code' => 'some_product']);
$variant = productVariantRepository->findOneBy(['code' => 'some_product_variant']);
$variant->setPosition(2);
$product->addVariant($variant);
productRepository->add($product);
```
The add() and position change actually does work, but the fatal error prevents further action.

**Possible Solution**
This appears similar to a issue recently reported and fixed regarding taxons:
https://github.com/Sylius/Sylius/issues/10797

Perhaps variants should implement Doctrine\Common\Comparable as well?

oallain:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #11487, like as #11329
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Core/Model/ProductVariant.php
+++ b/src/Sylius/Component/Core/Model/ProductVariant.php
@@ -15,11 +15,12 @@ namespace Sylius\Component\Core\Model;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
+use Doctrine\Common\Comparable;
 use Sylius\Component\Product\Model\ProductVariant as BaseVariant;
 use Sylius\Component\Shipping\Model\ShippingCategoryInterface;
 use Sylius\Component\Taxation\Model\TaxCategoryInterface;
 
-class ProductVariant extends BaseVariant implements ProductVariantInterface
+class ProductVariant extends BaseVariant implements ProductVariantInterface, Comparable
 {
     /** @var int */
     protected $version = 1;
@@ -322,5 +323,13 @@ class ProductVariant extends BaseVariant implements ProductVariantInterface
         if ($this->hasImage($image)) {
             $this->images->removeElement($image);
         }
+    }
+
+    /**
+     * @inheritDoc
+     */
+    public function compareTo($other): int
+    {
+        return $this->code === $other->getCode() ? 0 : 1;
     }
 }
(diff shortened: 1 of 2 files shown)


lchrusciel:
Hey Olivier,

can you rebase to newest 1.7? It may fix the build.

oallain:
Hi @lchrusciel ,

Rebase is done, and "application" is green on Travis, but "doc" still red :disappointed:

lchrusciel:
Our 1.7 branch is red, we will try to fix it in the nearest days

src/Sylius/Component/Core/Model/ProductVariant.php
```diff
@@ -444,4 +445,12 @@ public function removeImage(ProductImageInterface $image): void
             $this->images->removeElement($image);
         }
     }
+
+    /**
+     * @inheritDoc
+     */
+    public function compareTo($other): int
+    {
+        return $this->code === $other->getCode() ? 0 : 1;
```
dudemelo, line 454: > * This method should not check for identity using ===, only for semantical equality for example
> * when two different DateTime instances point to the exact same Date + **TZ.**
> https://github.com/doctrine/common/blob/master/lib/Doctrine/Common/Comparable.php#L20-L21

I would also like to suggest using another interface because this 0 for :milk_glass:  and 1 for :coffee:  is really tricky.

lchrusciel: If I understood correctly, for us it is indeed semantical equality when the codes are identical. Strong cooperation is used here more because of our standards rather than real need (as both codes will be strings)

lchrusciel:
Can you rebase again to 1.7? We could fix it there, I assume.

lchrusciel:
may become messed-up.**

execute `git fetch origin && git reset "product-variant-comparable"` to update your local branch.
 
lchrusciel:
Thank you, @oallain! :tada:

src/Sylius/Component/Core/spec/Model/ProductVariantSpec.php
```diff
@@ -273,4 +280,16 @@ function it_returns_images_by_type(ProductImageInterface $image, Product $produc
         $this->addImage($image);
         $this->getImagesByType('thumbnail')->shouldBeLike(new ArrayCollection([$image->getWrappedObject()]));
     }
+
+    function it_is_comparable(): void
+    {
+        $this->setCode('test');
+
+        $otherTaxon = new ProductVariant();
```
lchrusciel, line 288:
```suggestion
        $otherVariant = new ProductVariant();
```

PR #11590, opened 2020-06-17
Added expression language support to template event system

4c0n:
Added expression language support to template event system block contexts

| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no (Unless you count passing values that start with `@=` in the context values)
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/UiBundle/DependencyInjection/SyliusUiExtension.php
+++ b/src/Sylius/Bundle/UiBundle/DependencyInjection/SyliusUiExtension.php
@@ -19,6 +19,7 @@ use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Definition;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
+use Symfony\Component\ExpressionLanguage\Expression;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Zend\Stdlib\SplPriorityQueue;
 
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/UiBundle/DependencyInjection/SyliusUiExtension.php
```diff
@@ -59,6 +60,14 @@ private function loadEvents(array $eventsConfig, ContainerBuilder $container): v
             }
 
             foreach ($blocksPriorityQueue->toArray() as $details) {
+                if (\is_array($details['context'])) {
+                    foreach ($details['context'] as $key => &$value) {
+                        if (\is_string($value) && 0 === strpos($value, '@=')) {
```
pamil, line 65: I wonder if there's any other way to mark expressions - something like `!php/const` in Symfony's YAML config.

4c0n: I basically took this from: https://github.com/symfony/dependency-injection/blob/bb3e33b2c67b021b72e1200c4b1319af935b4f28/Loader/YamlFileLoader.php#L869
It's the way it's implemented in Symfony atm, not the nicest I agree.
So it's also likely that it won't work with other types of configuration, `xml` for example.

4c0n: The conversion to `Expression` objects here is also not strictly necessary, as the `ExpressionLanguage` should also be able to evaluate straight from `string`. However doing it here should be slightly more efficient, as it will hopefully parse the expression during compile time and cache it.

4c0n:
Couldn't update the base branch of the PR, so let's continue the conversation here: https://github.com/Sylius/Sylius/pull/11591

Issue #11605, opened 2020-06-27
Show available variants in grid if such exists


dbalabka:
Currently, the grid contains any variants sorted by position. IMO customer is interested to see available variants if such exists. To make it possible, available variants sorted by position should show first otherwise not available variant should be shown.

**Additional context**
It is important if you would like to customize the grid and add product availability information based on variant information. If a variant that has position "0" is not available, such product will be marked as not available.

**Alternative solution**
Summary of variants `on hand` and `on hold` information should be stored on product level as well.

TheMilek:
Hi @dbalabka!
I'm closing this issue in ref to https://github.com/Sylius/Sylius/pull/11606#issuecomment-1420848222

PR #11606, opened 2020-06-27
Show available variants first (#11605)

dbalabka:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #11605 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
@@ -95,11 +95,23 @@ class ProductRepository extends BaseProductRepository implements ProductReposito
         // Grid hack, we do not need to join these if we don't sort by price
         if (isset($sorting['price'])) {
             // Another hack, the subquery to get the first position variant
-            $subQuery = $this->createQueryBuilder('m')
-                 ->select('min(v.position)')
-                 ->innerJoin('m.variants', 'v')
-                 ->andWhere('m.id = :product_id')
+            $allVariants = $this->createQueryBuilder('m1')
+                 ->select('min(v1.position)')
+                 ->innerJoin('m1.variants', 'v1')
+                 ->andWhere('m1.id = :product_id')
+                 ->andWhere('v1.enabled = :enabled')
+                ->getDQL()
              ;
+            // ... show available variant if such exists
+            $availableVariants = $this->createQueryBuilder('m2')
+                ->select('min(v2.position)')
+                ->innerJoin('m2.variants', 'v2')
+                ->andWhere('m2.id = :product_id')
+                ->andWhere('v2.enabled = :enabled')
+                ->getDQL()
+            ;
+
+            [$allVariants, $availableVariants] = str_replace(':product_id', 'o.id', [$allVariants, $availableVariants]);
 
             $queryBuilder
                 ->addSelect('variant')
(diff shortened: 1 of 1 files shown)


dbalabka:
@lchrusciel could you please update if there is a chance to review these changes and get feedback?

lchrusciel:
Hey @dbalabka I may reopen the issue and send it over to PO, but do not guarantee, that something will happen to it

lchrusciel:
After reading #11605, I would say, that it is a matter of new filtration options on the variant list rather than a required change in our behavior. You have issues with your custom implementation, not with Sylius per see. I'm not convinced if it should land in the core. It complicates query and makes the business logic not obvious.

New filter (probably turned on by default) seems to be much better approach here

dbalabka:
@lchrusciel I see what you mean. Unfortunately, it is complicated to implement any kind of advanced filtration using Doctrine and MySQL due to DQL limitations (e.g. there no [hierarchical queries](https://en.wikipedia.org/wiki/Hierarchical_and_recursive_queries_in_SQL) support). There are some workarounds but they provide poor DX and increase maintainability costs (e.g. [native SQL mapping](https://www.doctrine-project.org/projects/doctrine-orm/en/2.8/reference/native-sql.html) )

It would be great to have a simple and developer-friendly approach to customize product grids :)

vvasiloi:
@dbalabka 
I understand 2 different problems from #11605.
One is resolving the default product variant, which could be solved with a new implementation of `Sylius\Component\Product\Resolver\ProductVariantResolverInterface`.
Another is sorting variants by availability in a grid.
The repository method you modified, by default, is only used in the shop's product grid (sylius_shop_product) which doesn't display variants.
Could you clarify it please?

dbalabka:
@vvasiloi you probably right. I wasn't aware of the resolver and it might be a viable and easier solution. Still, most of the stores keeping unorderable products on the last pages by default. It will be useful to have such sorting.

vvasiloi:
@dbalabka that is indeed very common and would be a useful addition to Sylius.
However, your proposed implementation doesn't do that.
It will only work when the products are sorted by price and will change which variant's price is used for sorting.
Before it used the first one and with your implementation it will use the first enabled.
Also, the queries for `$allVariants` and `$availableVariants` are identical, I believe you intended the first one to be without `->andWhere('v1.enabled = :enabled')`.
I will think about it and try to propose a better solution.

vvasiloi:
@dbalabka do you want to continue with this or should I close it?

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ProductRepository.php
```diff
@@ -95,24 +95,31 @@ public function createShopListQueryBuilder(
...
+                 ->andWhere('m1.id = :product_id')
+                 ->andWhere('v1.enabled = :enabled')
+                ->getDQL()
              ;
+            // ... show available variant if such exists
+            $availableVariants = $this->createQueryBuilder('m2')
+                ->select('min(v2.position)')
+                ->innerJoin('m2.variants', 'v2')
+                ->andWhere('m2.id = :product_id')
+                ->andWhere('v2.enabled = :enabled')
```
dbalabka, line 110: missing line `->andWhere('(v2.onHand - v2.onHold) > 0')`

dbalabka:
@vvasiloi seems, there is not much interest in this feature from the community. Also, the solution looks very hacky. I don't think that it is worth adding additional complexity. IMO it is much easier to implement such complex sorting using Elastsearch instead of MySQL.

jakubtobiasz:
Hi @dbalabka!
Thanks for your contribution, but we've decided we prefer to left is as a project-level decision. If we change our mind, we'll back here :).

dbalabka:
@dbalabka , I respect your decision. Still, the problem is that overriding takes work, and the code quickly becomes messy. Please consider providing more convenient extension points.

PR #11607, opened 2020-06-28
Schema.org compatible breadcrumbs

jacquesbh:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

I find it not very useful to have multiple places to look out for the
breadcrumbs in the templates.

So I've added here a new macro `breadcrumb` and I've updated the templates
so it can be used properly.

Also, the macro changes a little bit the HTML because I've added the corresponding
microdata for the breadcrumb to be compatible with rich results.  
It can be verified here: https://search.google.com/test/rich-results with the content of the page.

In order to have correct rich results I had to remove the name of the root taxon.  
Since we just display the name of the root taxon and it doesn't have a link, it cannot
be used as an item in the middle of the breadcrumb.  
And I find it better this way since in all our projects we had to remove it.

**Before:**  
image

**After:**  
image

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/Common/Macro/breadcrumb.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/Common/Macro/breadcrumb.html.twig
@@ -1,0 +1,31 @@
+{% macro breadcrumb(crumbs = {}, prependHome = true) %}
+    {% from _self import breadcrumb_current_crumb, breadcrumb_crumb %}
+    {% if prependHome %}
+        {% set crumbs = [
+            {
+                url: path('sylius_shop_homepage'),
+                name: 'sylius.ui.home'|trans
+            }
+        ]|merge(crumbs) %}
+    {% endif %}
+    {% if crumbs|length %}
+    <div class="ui breadcrumb" itemscope itemtype="https://schema.org/BreadcrumbList">
+        {% for crumb in crumbs %}
+            {% if loop.last %}
+                <span itemprop="itemListElement" itemscope itemtype="http://schema.org/ListItem">
+                    <strong class="active section" itemprop="name">{{ crumb.name }}</strong>
+                    <meta itemprop="position" content="{{ loop.index }}">
+                </span>
+            {% else %}
+                <span itemprop="itemListElement" itemscope itemtype="http://schema.org/ListItem">
+                    <a class="section" href="{{ crumb.url }}" itemprop="item">
+                        <span itemprop="name">{{ crumb.name }}</span>
+                    </a>
+                    <meta itemprop="position" content="{{ loop.index }}">
+                </span>
+                <div class="divider"> / </div>
+            {% endif %}
+        {% endfor %}
+    </div>
+    {% endif %}
+{% endmacro %}
--- a/src/Sylius/Bundle/ShopBundle/Resources/views/Product/Show/_breadcrumb.html.twig
+++ b/src/Sylius/Bundle/ShopBundle/Resources/views/Product/Show/_breadcrumb.html.twig
@@ -1,21 +1,23 @@
-<div class="ui breadcrumb">
-    <a href="{{ path('sylius_shop_homepage') }}" class="section">{{ 'sylius.ui.home'|trans }}</a>
-    <div class="divider"> / </div>
-    {% if product.mainTaxon is not null %}
-        {% set taxon = product.mainTaxon %}
-        {% set ancestors = taxon.ancestors|reverse %}
+{% from '@SyliusShop/Common/Macro/breadcrumb.html.twig' import breadcrumb %}
+{% set crumbs = [] %}
+{% if product.mainTaxon is not null %}
+    {% set taxon = product.mainTaxon %}
+    {% set ancestors = taxon.ancestors|reverse %}
 
-        {% for ancestor in ancestors %}
-            {% if ancestor.isRoot() %}
-                <div class="section">{{ ancestor.name }}</div>
-            {% else %}
-                <a href="{{ path('sylius_shop_product_index', {'slug': ancestor.slug, '_locale': ancestor.translation.locale}) }}" class="section">{{ ancestor.name }}</a>
-            {% endif %}
-            <div class="divider"> / </div>
-        {% endfor %}
-
-        <a href="{{ path('sylius_shop_product_index', {'slug': taxon.slug, '_locale': taxon.translation.locale}) }}" class="section">{{ taxon.name }}</a>
-        <div class="divider"> / </div>
-    {% endif %}
-    <div class="active section">{{ product.name }}</div>
-</div>
+    {% for ancestor in ancestors %}
+        {% if not ancestor.isRoot() %}
+            {% set crumbs = crumbs|merge([{
+                url: path('sylius_shop_product_index', {'slug': ancestor.slug, '_locale': ancestor.translation.locale}),
+                name: ancestor.name
+            }]) %}
+        {% endif %}
+    {% endfor %}
+    {% set crumbs = crumbs|merge([{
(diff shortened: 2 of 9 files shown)


vvasiloi:
Hey @jacquesbh! Good job! Have you considered doing it like in [Admin](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/AdminBundle/Resources/views/Macro/breadcrumb.html.twig#L1) / [UI](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/UiBundle/Resources/views/Macro/breadcrumb.html.twig#L1) bundles?

jacquesbh:
> Hey @jacquesbh! Good job! Have you considered doing it like in [Admin](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/AdminBundle/Resources/views/Macro/breadcrumb.html.twig#L1) / [UI](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/UiBundle/Resources/views/Macro/breadcrumb.html.twig#L1) bundles?

No I have not.
But I will now.

lchrusciel:
Hey Jacques,

what would you say about extracting breadcrumb creation logic to PHP (like building an array of them) and minimize the amount of logic in twig? We can use some twig functions for this. Such solution would be much easier to customize and maintain in the long term.

jacquesbh:
@lchrusciel What do you mean? To fill the crumbs in PHP? Or to display the breadcrumb?
Like using a twig method `add_crumb()` or something like that?
Then we display the crumb where we want using a `breadcrumb()` method as well?

Can you clarify the idea here? please :)

jacquesbh:
I'm not sure having it in PHP would help a lot. It's pretty easy to understand and the data can come from so many places. Having in in Twig is easier I think.

Also, the macro is not complicated.

And using merges to form the crumbs is fine to me, it exists for this reason.
And it keeps the flexibility. At the end it's PHP anyway ;).

src/Sylius/Bundle/ShopBundle/Resources/views/Product/Show/_breadcrumb.html.twig
```diff
@@ -1,21 +1,23 @@
...
+                url: path('sylius_shop_product_index', {'slug': ancestor.slug, '_locale': ancestor.translation.locale}),
+                name: ancestor.name
+            }]) %}
+        {% endif %}
+    {% endfor %}
+    {% set crumbs = crumbs|merge([{
+        url: path('sylius_shop_product_index', {'slug': taxon.slug, '_locale': taxon.translation.locale}),
+        name: taxon.name
+    }]) %}
+{% endif %}
```
lchrusciel, line 19: IMHO this logic could be extracted to Twig function

src/Sylius/Bundle/ShopBundle/Resources/views/Taxon/_breadcrumb.html.twig
```diff
@@ -1,15 +1,15 @@
...
-</div>
+{% set crumbs = [] %}
+{% for ancestor in ancestors %}
+    {% if not ancestor.isRoot() %}
+        {% set crumbs = crumbs|merge([{
+            url: path('sylius_shop_product_index', {'slug': ancestor.slug, '_locale': ancestor.translation.locale}),
+            name: ancestor.name
+        }]) %}
+    {% endif %}
+{% endfor %}
```
lchrusciel, line 11: If above logic will be extracted to Twig function, here it could be reused probably

src/Sylius/Bundle/ShopBundle/Resources/views/Common/Macro/breadcrumb.html.twig
```diff
@@ -0,0 +1,31 @@
+{% macro breadcrumb(crumbs = {}, prependHome = true) %}
+    {% from _self import breadcrumb_current_crumb, breadcrumb_crumb %}
+    {% if prependHome %}
+        {% set crumbs = [
+            {
+                url: path('sylius_shop_homepage'),
+                name: 'sylius.ui.home'|trans
+            }
+        ]|merge(crumbs) %}
+    {% endif %}
```
lchrusciel, line 10: This could be perhaps be extracted to twig filter/function? So we will enrich crumbs with homepage, but you would be able to customize it in php

lchrusciel:
My main point is that crumbling taxons is not that easy. With PHP we can test it with phpspec at least. What is more, we are able to reuse logic, instead of duplicating it. Adding the homepage is just a minor improvement for me. 

Nonetheless, thanks a lot for the proposal and implementation. Macro used on my account website looks really decent 🎉

src/Sylius/Bundle/ShopBundle/Resources/views/Taxon/_breadcrumb.html.twig
jacquesbh: WDYT if it comes to something like this (take the idea, not the implementation itself):

```suggestion
{% set crumbs = getCrumbsFromTaxons(ancestors) %}
```

Would it be ok to you?

src/Sylius/Bundle/ShopBundle/Resources/views/Common/Macro/breadcrumb.html.twig
jacquesbh: WDYT if it comes to something like this (take the idea, not the implementation itself):

```suggestion
{% set crumbs = prepareCrumbs(crumbs, prependHome) %}
```

src/Sylius/Bundle/ShopBundle/Resources/views/Product/Show/_breadcrumb.html.twig
jacquesbh: WDYT if it comes to something like this (take the idea, not the implementation itself):

```twig
{% set crumbs = getCrumbsFromTaxons(ancestors)|merge([{
        url: path('sylius_shop_product_index', {'slug': taxon.slug, '_locale': taxon.translation.locale}),
        name: taxon.name
    }]) %}
```

For me, having a twig function doing a merge is useless since the `merge` filter exists.
I understand for the loop on the taxons, it makes sens, but not for the second part.

lchrusciel: I was thinking about something like 
```twig
{% set crumbs = crumbTaxon(taxon)|merge([{
        url: path('sylius_shop_product_index', {'slug': taxon.slug, '_locale': taxon.translation.locale}),
        name: taxon.name
    }]) %}
```
Where it would return ready to use an array with URL and name. However, the main point here is, that this logic is duplicated and it is not just a merge, but it is a merge in a if statement in a for loop.

jacquesbh: Ok, I'll improve this.

lchrusciel: thank you :)

gabiudrescu:
we decided to go with JSON-LD instead of itemprops. Google handles this format better: https://www.searchenginejournal.com/google-structured-data-preference/297479/#close

also, I remember I once read that Google will retire the support for itemprops, but I cannot seem to find this now.

also, using JSON-LD instead of itemprops makes this functionality available for API as well, not only twig logic that cannot be used somewhere else.

jacquesbh:
Yes, actually that's why I haven't came back on this PR yet. I think we will provide something more interesting with JSON-LD.
So, let's put this PR as a draft and I'll come back to it later!

PR #11624, opened 2020-07-06
Fix migration

### LINKED ISSUE #11592: An existing migration is blocking future migrations
**Sylius version affected**: **since 1.1.0**

**Description**  
If you add a field using annotations to the ProductAttribute entity then you end up with an error during migrations.
Here I've added a `filterable` field and I get this:

image

This is caused because we ask doctrine to load the attributes during a migration:
https://github.com/Sylius/Sylius/blob/f08c3bf8f70e8f13da1ca3efa8440846297f0c76/app/migrations/Version20171003103916.php

**Steps to reproduce**  

Add a field to your entity, generate the migration and migrate the database from the beginning.

**Possible Solution**  

Using a specific method of the Repository, limiting the fields to select.

```php
    public function findByTypeForMigration(string $type)
    {
        return $this
            ->createQueryBuilder('o')
            ->select([
                'o.id',
                'o.code',
                'o.type',
                'o.storageType',
                'o.configuration',
                'o.createdAt',
                'o.updatedAt',
                'o.position'
            ])
            ->where('o.type = :type')
            ->setParameter('type', $type)
            ->getQuery()
            ->getResult()
        ;
    }
```

jacquesbh:
_Previous title: Add method findTypeForMigration to ProductAttributeRepository_

| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #11592
| License         | MIT

I don't know how to test these except by running the database migrations.  
So…

I've added the repository in Core because I think it is where it belongs since
the Migrations are in the main App.

### DIFF
--- a/app/migrations/Version20171003103916.php
+++ b/app/migrations/Version20171003103916.php
@@ -37,7 +37,18 @@ class Version20171003103916 extends AbstractMigration implements ContainerAwareI
         $defaultLocale = $this->container->getParameter('locale');
         $productAttributeRepository = $this->container->get('sylius.repository.product_attribute');
 
-        $productAttributes = $productAttributeRepository->findBy(['type' => SelectAttributeType::TYPE]);
+        $productAttributes = $productAttributeRepository
+            ->createQueryBuilder('o')
+                ->select([
+                    'o.id',
+                    'o.configuration',
+                ])
+                ->where('o.type = :type')
+                ->setParameter('type', SelectAttributeType::TYPE)
+                ->getQuery()
+            ->getResult()
+        ;
+
         /** @var ProductAttributeInterface $productAttribute */
         foreach ($productAttributes as $productAttribute) {
             $configuration = $productAttribute->getConfiguration();


app/migrations/Version20171003103916.php
```diff
@@ -37,7 +37,7 @@ public function up(Schema $schema): void
         $defaultLocale = $this->container->getParameter('locale');
         $productAttributeRepository = $this->container->get('sylius.repository.product_attribute');
 
-        $productAttributes = $productAttributeRepository->findBy(['type' => SelectAttributeType::TYPE]);
+        $productAttributes = $productAttributeRepository->findByTypeForMigration(SelectAttributeType::TYPE);
```
lchrusciel, line 40: Despite the typehint, from the repository, you can fetch query builder. Therefore, I would say it would be better to just write custom method only in this migration file rather than place it in generally available place

jacquesbh: Yep. Much easier, you perfectly right. I've done the change.

lchrusciel:
may become messed-up.**

execute `git fetch origin && git reset "fix-migration"` to update your local branch.
 
lchrusciel:
Thank you, Jacques! :1st_place_medal:

PR #11639, opened 2020-07-14
Replace usage of abandoned ZF libraries in favour of Laminas

stloyd:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | yes(?)
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | partially #10928 (?)
| License         | MIT

Zend Framework packages are abandoned & should be replaced with Laminas ones.

PR #11640, opened 2020-07-14
Update "sylius/theme-bundle" to latest which allows ZF update

stloyd:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | yes(?)
| New feature?    | yes(?)
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially #10928
| License         | MIT

Without this update, merge of #11639 is blocked.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -62,7 +62,7 @@
         "sylius/registry": "^1.5",
         "sylius/resource": "1.6.*",
         "sylius/resource-bundle": "1.6.*",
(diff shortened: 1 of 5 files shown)


stloyd:
This is still valid.

lchrusciel (approved):

Zales0123:
Even though the fix is valid... It seems as it will not work without Pagerfanta upgrade, which requires the ResourceBundle upgrade 💃

PR #11642, opened 2020-07-15
Add field where You can verify shop user

### LINKED ISSUE #9582: Cannot verify shop user account when account was created in the backend
**Sylius version affected**: <= 1.2.2

**Description**  

**My use case is not to allow unverified shop users to authenticate.** (I have a `UserChecker` that verifies that)

As an admin I created a shop user. I checked that user is enabled, can login to the shop and assigned a password. However this shop user's account is not verified (`verified_at` is null). I "recovered the password" successfully, clicked on the link to recover the password, set a new one, but still the account isn't verified. Is it possible to verify the account then under any circumstance?

**Steps to reproduce**  
Simple as is.

**Possible Solution**  
1. Add `verified` checkbox to the user profile form in the admin's backend, so the admin can mark the account as verified by himself.
2. Verify the account if any e-mail link is clicked (like a reset password link)

arti0090:
| Q               | A
| --------------- | -----
| Branch?         |  master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | not sure
| Deprecations?   | not sure
| Related tickets | fixes #9582
| License         | MIT

Zrzut ekranu 2020-07-14 o 13 50 21

Field is disabled when user was verified before and shows only verification date

### DIFF
--- a/src/Sylius/Bundle/UserBundle/Form/Type/UserType.php
+++ b/src/Sylius/Bundle/UserBundle/Form/Type/UserType.php
@@ -15,10 +15,13 @@ namespace Sylius\Bundle\UserBundle\Form\Type;
 
 use Sylius\Bundle\ResourceBundle\Form\Type\AbstractResourceType;
 use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
+use Symfony\Component\Form\Extension\Core\Type\DateTimeType;
 use Symfony\Component\Form\Extension\Core\Type\EmailType;
 use Symfony\Component\Form\Extension\Core\Type\PasswordType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
 use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\Form\FormEvent;
+use Symfony\Component\Form\FormEvents;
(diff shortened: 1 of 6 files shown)


src/Sylius/Bundle/UserBundle/Form/Type/UserType.php
```diff
@@ -38,8 +41,29 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+                'widget' => 'single_text',
+                'label' => 'sylius.form.user.verified_date',
+                'required' => false,
             ])
         ;
+
+        $builder->addEventListener(FormEvents::PRE_SET_DATA, function(FormEvent $event){
+            $data = $event->getData();
+            if(null !== $data) {
+                if($data->isVerified()) {
```
lchrusciel, line 56:
```suggestion
            if(null !== $data && $data->isVerified()) {
```

arti0090: It works when editing user, but when creating new one $data is null so calling function isVerified() on it creates error.

lchrusciel:

PR #11664, opened 2020-07-20
Use env(SYLIUS_ADMIN_ROUTING_PATH_NAME) as admin's route name

jacquesbh:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes, kind of
| BC breaks?      | I'm not sure. This PR doesn't change the routes but it allows it…
| Deprecations?   | no
| Related tickets |
| License         | MIT

We always change the admin route URL. By default `/admin` is unsecure, it's too
easy to guess.

We wanted to ease this change by using an environment variable.

All plugins using `/admin` hard coded will have to use the dynamic `/%sylius_admin.path_name%` value instead.


lchrusciel:
Thank you, Jacques! :1st_place_medal:

Issue #11622, opened 2020-07-06
Testing themes in browser tests requires clearing the cache, but normal users cannot do that


4c0n:
**Describe the proposed solution**
In browser based tests like the ones we run in Behat, currently when we'd like to test the UI with a custom theme, before we can actually see the theme we need to clear the cache. Clearing the cache is an issue, because it is also required by https://github.com/FriendsOfBehat/SymfonyExtension to for example be able to inject dependencies.
A solution would be to be able to select a theme and have it present itself without having to clear the cache, maybe there are different solutions to achieve this goal.

**Describe alternatives you've considered**
I haven't a clue.

**Additional context**
Admins cannot typically clear/reset the cache, they would have to ask their system admin or someone that has access to the CLI, which is not very user friendly.

4c0n:
For example:

https://github.com/nedac-sorbo/SyliusTemporarilyOutOfStockPlugin/blob/b8151b50ee255456515ab00936b1967872ca7ca7/run-behat.sh#L4

PR #11669, opened 2020-07-21
[WIP] Add option to set theme via parameters

anthid:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially #11622
| License         | MIT

With the current implementation, the only option to set themes is selecting channel theme via admin panel. This is not optimal when we want to test themes (set theme; clear cache; run tests).

One possible solution is to set a `default_theme` via parameters. This `default_theme` will only be loaded if there is no channel theme defined. It is like a fallback option and its default value should be `null`. That way we can install a new theme, set as `default_theme`, clear cache and run tests against it.

The **[WIP]** label is because there are many ways to implement this functionality and at the end it is a core decission. For example `default_theme` probably should belong to `sylius_theme` and not `parameters`, but I put it here to make it understandable.

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -2,3 +2,4 @@
 # https://symfony.com/doc/current/best_practices/configuration.html#application-related-configuration
 parameters:
     locale: en_US
+    default_theme: null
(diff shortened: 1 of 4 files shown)


lchrusciel:
Hey @anthid, thanks for your proposal and welcome to the Sylius community.

From my perspective, the default theme functionality is indeed more related to ThemeBundle rather than to the Core logic itself. 

However, if I understood your request correctly, it is already possible to achieve what you've asked for. During CLI installation, you can select a theme with Sylius Fixtures, with the Channel Fixture in particular. So now, what would you need to do is to define proper fixture suites with the chosen themes of your choice. Potential improvement to it could be achieved with the env variable, which would allow anyone to select a proper theme during installation. Would it work for you?

PR #11668, opened 2020-07-21
[Admin] Added collapse menu

TefDesign:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

Add collapse on admin sidebar 

sylius_admin_menu

PR #11674, opened 2020-07-23
Revert "[Admin] Added collapse menu"

CoderMaggie:
Reverts Sylius/Sylius#11668

Although it's nice to see new contributions and I am very grateful for the community input, this PR has to be reverted. 😢 

Nagranie z ekranu 2020-07-23 o 08 34 23

The behaviour of this "menu collapsing JS" highly lowers the UX of the admin menu. 
First of all, it works slowly. Secondly, it is collapsed by default, so after each redirect, we end up with a menu collapsed again even though we've been using it a moment ago. And finally, it was merged pretty quick, without enough consulting in the community and in the Core Team.

Final thought, for anyone who'd like to rework that: The admin menu should be collapsible, not collapsed. 🚀


pamil:
Thanks, Magdalena! :tada:

TefDesign:
Hello @CoderMaggie and @pamil , I'm the dev of this plugin. I tested again this morning and I have no performance loss I don't understand. In addition, when clicking on a menu, the selected menu remains open when loading.
I am interested for any suggestion :)
Thanks :)

CoderMaggie:
Hi @TefDesign 👋 
We've been testing it on the public demo, and it was lagging slightly (it's visible on the gif I've posted), maybe it's caused by the environment, but we need to have a solution that works smoothly on every environment.

The main suggestion is: the menu should not be collapsed by default, as it increases the number of clicks an admin needs to go anywhere (from 1, to at least 2 if you remember where the subsection should be). 

When it comes to losing the active menu element, as far as I know, in the current architecture it is not possible to hold an active element, while we are browsing pages deeper than the ones available in the menu. That's why currently you also do not have the "Products" subsection of "Catalog" bolded, while you are editing products, it is bolded only until you are browsing the products index.

We could consider the feature to collapse a section when one wants to, but I do not see a particular benefit of having it, while we already have the search bar on the admin menu (on its very top).

TefDesign:
@CoderMaggie Ok it's noted thank you for your feedback !

PR #11682, opened 2020-07-24
[Taxon] Enhance delete taxon message while in use as main taxon by listing products

### LINKED ISSUE #11277: Taxon without product isn't removable
**Sylius version affected**: v1.4.4

**Description**  

Sylius says the category is used and cannot be removed, when we try to remove an empty category.

We figured out that we had some products with the taxon as "Main Taxon".

**Steps to reproduce**  

1. Add a taxon, keep it empty of products.
2. Update a product to use the newly created taxon as Main taxon.
3. Try to remove the taxon.

**Possible Solution**  

I suggest to set to `NULL` the `main_taxon_id` column for those products, but this may not be the best solution.

Adding a warning could be a good idea, like "This taxon is used as main taxon for some products.".
But here the issue is "Which products???".

So, maybe a solution could be a warning and a link or a button to "Remove the taxon anyway".

I don't know. The usability of this is quite tricky :/.

hatem20:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | fixes #11277 
| License         | MIT

I followed the last suggesting in https://github.com/Sylius/Sylius/issues/11277#issuecomment-645868992

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/EventListener/TaxonDeletionListener.php
+++ b/src/Sylius/Bundle/CoreBundle/EventListener/TaxonDeletionListener.php
@@ -17,6 +17,7 @@ use Sylius\Bundle\CoreBundle\Provider\FlashBagProvider;
 use Sylius\Component\Channel\Repository\ChannelRepositoryInterface;
 use Sylius\Component\Core\Model\TaxonInterface;
 use Sylius\Component\Core\Promotion\Updater\Rule\TaxonAwareRuleUpdaterInterface;
+use Sylius\Component\Core\Repository\ProductRepositoryInterface;
 use Symfony\Component\EventDispatcher\GenericEvent;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface;
@@ -31,6 +32,7 @@ final class TaxonDeletionListener
     public function __construct(
         private SessionInterface|RequestStack $requestStackOrSession,
         private ChannelRepositoryInterface $channelRepository,
+        private ProductRepositoryInterface $productRepository,
         TaxonAwareRuleUpdaterInterface ...$ruleUpdaters,
     ) {
         $this->ruleUpdaters = $ruleUpdaters;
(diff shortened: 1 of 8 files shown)


src/Sylius/Bundle/UiBundle/Resources/translations/flashes.en.yml
```diff
@@ -28,3 +28,4 @@ sylius:
         update_rules: 'Some rules of the promotions with codes %codes% have been updated.'
     taxon:
         menu_taxon_delete: 'You cannot delete a menu taxon of any channel.'
+        main_taxon_delete: 'You cannot delete this taxon, it is used as main taxon for products with codes %codes%'
```
lchrusciel, line 31:
```suggestion
        main_taxon_delete: 'You cannot delete this taxon, it is used as the main taxon for at least products with codes %codes% (max. 20)'
```

lchrusciel (commented):
Hey Hatem,

thanks a lot for your contribution. Can you cover your feature with the behat scenario?

src/Sylius/Bundle/UiBundle/Resources/translations/flashes.en.yml
hatem20: thanks @lchrusciel sure I will add behat scenario
but `(max. 20)` do you think it adds value?
and if yes, at least it should be dynamic.

lchrusciel: Let's make it dynamic, that's a good idea. I think so, especially for newcomers. As you wanted to delete taxon and then you received information, that you need to remove it from 20 products. You did it and still the same message, while you could expect, that that's it. In the future, we could add a count of all products.

src/Sylius/Bundle/CoreBundle/EventListener/TaxonDeletionListener.php
```diff
@@ -57,6 +63,35 @@ public function protectFromRemovingMenuTaxon(GenericEvent $event): void
         }
     }
 
+    public function protectFromRemovingProductMainTaxon(GenericEvent $event): void
```
lchrusciel, line 66: This class becoming more and more like a controller. We should split it into separate parts, but that may be a case for separate PR.

lchrusciel (commented):
Hey Hatem,

could you rebase to the newest master, so we can check if the build is green?

src/Sylius/Bundle/CoreBundle/EventListener/TaxonDeletionListener.php
hatem20: agreed

hatem20:
@lchrusciel I rebased twice to pass. I do not actually understand why it fails one time and passes another without changing anything. also when it passes I notice errors in the build.

hatem20:
@lchrusciel can you please review, thanks.

TheMilek:
Hi @hatem20!

Thank you for your PR 🥇 
We have already addressed this issue in https://github.com/Sylius/Sylius/pull/14994. 
However, based on your contribution, we are planning to add some confirmation to inform the user about removing the taxon that serves as the main taxon for certain products 🎉

PR #11653, opened 2020-07-16
[Behat] Refactor cart transformer

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

PR #11697, opened 2020-07-30
[Behat] Refactor cart transformer

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | replaced #11653 
| License         | MIT


src/Sylius/Behat/Context/Transform/CartContext.php
```diff
@@ -0,0 +1,62 @@
...
+        if ($this->sharedStorage->has('cart')) {
+            return $this->sharedStorage->get('cart');
+        }
+
+        $this->commandBus->dispatch(new PickupCart());
+
+        /** @var OrderInterface|null $cart */
+        $cart = $this->repository->findLatestCart();
+
+        $this->sharedStorage->set('cart', $cart);
```
Tomanhez, line 58: We shared $cart object, but `$cart` can be changed after certain request, then this `$cart` will be incorrect. I think that `$token` is more safety an less error prone.

PR #11706, opened 2020-07-31
[API][Cart] Maintaining cart after authorization

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

I've added `TokenValueBasedCartContext` with higher priority than other contexts to assign cart to a logged in customer by token value in `CartBlamerListener`

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Context/TokenValueBasedCartContext.php
+++ b/src/Sylius/Bundle/ApiBundle/Context/TokenValueBasedCartContext.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Context;
+
+use Sylius\Component\Core\Repository\OrderRepositoryInterface;
+use Sylius\Component\Order\Context\CartContextInterface;
+use Sylius\Component\Order\Context\CartNotFoundException;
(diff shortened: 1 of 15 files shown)


src/Sylius/Bundle/ApiBundle/Context/TokenValueBasedCartContext.php
```diff
@@ -0,0 +1,62 @@
...
+    public function __construct(RequestStack $requestStack, OrderRepositoryInterface $orderRepository)
+    {
+        $this->requestStack = $requestStack;
+        $this->orderRepository = $orderRepository;
+    }
+
+    public function getCart(): OrderInterface
+    {
+        $request = $this->getMasterRequest();
+        $tokenValue = $request->attributes->get('id');
```
Zales0123, line 40: It was a little bit confusing at first sight for me why we're looking for the `id` instead of `tokenValue`. I understand it's the default parameter name in API Platform, but maybe it would be good to explain that :) Also, I was wondering can we make this service working only in API context. I know it would still break at line 45, but I feel uncomfortable that we still always look for cart by the `id`s passed in the URL, before we try another way of resolving cart 😄

GSadee: I've added simple check if the master request is an API request that this service works only in API context

symfony.lock
```diff
@@ -62,7 +62,7 @@
         "version": "2.2.3"
     },
     "php": {
-        "version": "7.4"
+        "version": "7.3"
```
Zales0123, line 65: 😢

Zales0123:
Thanks, Grzegorz! :tada:

PR #11710, opened 2020-08-03
[Hot-fix] Handle extra docblocks in Psalm

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

I've removed two unneeded docblocks, but most of the problems were caused by the `Assert::` methods usage. We could, in theory, remove the docblocks there as well, but it would result in the PHPStorm (or some other IDE) not detecting the types properly... That's why it's a hot-fix 💃

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -78,7 +78,7 @@
     </projectFiles>
 
     <issueHandlers>
-        <!-- level 3 issues - slightly lazy code writing, but provably low false-negatives -->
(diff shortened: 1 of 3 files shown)


GSadee:
Thanks, Mateusz! :tada:

PR #11735, opened 2020-08-13
[Api][Checkout] ShippingMethod on Checkout

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

PR #11745, opened 2020-08-17
[Api][Checkout] Modify ShippingMethod and PaymentMethod paths on Checkout

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Its continue: #11735

Screenshot 2020-08-20 at 12 31 59

Screenshot 2020-09-02 at 12 24 38

Screenshot 2020-09-02 at 16 45 33

### DIFF
--- a/src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
@@ -238,12 +238,15 @@ final class CheckoutContext implements Context
     {
         $this->client->request(
             Request::METHOD_PATCH,
-            \sprintf('/new-api/orders/%s/shipments/{shipmentId}', $this->sharedStorage->get('cart_token')),
+            \sprintf(
+                '/new-api/orders/%s/shipments/%s',
+                $this->sharedStorage->get('cart_token'),
+                (string) $this->iriConverter->getItemFromIri($this->getCart()['shipments'][0])->getId()
+            ),
             [],
             [],
             $this->getHeaders(),
             json_encode([
-                'shipmentId' => (string) $this->iriConverter->getItemFromIri($this->getCart()['shipments'][0])->getId(),
                 'shippingMethodCode' => $shippingMethod->getCode(),
             ], \JSON_THROW_ON_ERROR)
         );
--- a/src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
@@ -238,13 +238,13 @@ final class CheckoutContext implements Context
     {
         $this->client->request(
             Request::METHOD_PATCH,
-            \sprintf('/new-api/orders/%s/select-shipping-methods', $this->sharedStorage->get('cart_token')),
+            \sprintf('/new-api/orders/%s/shipments/{shipmentId}', $this->sharedStorage->get('cart_token')),
             [],
             [],
             $this->getHeaders(),
             json_encode([
-                'shipmentIdentifier' => 0,
-                'shippingMethod' => $shippingMethod->getCode(),
+                'shipmentId' => (string) $this->iriConverter->getItemFromIri($this->getCart()['shipments'][0])->getId(),
+                'shippingMethodCode' => $shippingMethod->getCode(),
             ], \JSON_THROW_ON_ERROR)
         );
     }
--- a/src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
@@ -617,11 +617,11 @@ final class CheckoutContext implements Context
         }
 
         $this->client->request(
-            Request::METHOD_PUT,
-            \sprintf('/new-api/orders/%s', $this->sharedStorage->get('cart_token')),
+            Request::METHOD_PATCH,
+            \sprintf('/new-api/orders/%s/address', $this->sharedStorage->get('cart_token')),
             [],
             [],
-            $this->getHeaders(['HTTP_ACCEPT' => 'application/ld+json', 'CONTENT_TYPE' => 'application/ld+json']),
+            $this->getHeaders(),
             json_encode($content, \JSON_THROW_ON_ERROR)
         );
     }
--- a/src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
@@ -271,13 +271,16 @@ final class CheckoutContext implements Context
     {
         $this->client->request(
             Request::METHOD_PATCH,
-            \sprintf('/new-api/orders/%s/select-payment-methods', $this->sharedStorage->get('cart_token')),
+            \sprintf(
+                '/new-api/orders/%s/payments/%s',
+                $this->sharedStorage->get('cart_token'),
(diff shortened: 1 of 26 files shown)


src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -206,13 +206,13 @@ public function iProceededWithShippingMethod(ShippingMethodInterface $shippingMe
     {
         $this->client->request(
             Request::METHOD_PATCH,
-            \sprintf('/new-api/orders/%s/select-shipping-methods', $this->sharedStorage->get('cart_token')),
+            \sprintf('/new-api/orders/%s/shipments/{shipmentId}', $this->sharedStorage->get('cart_token')),
```
GSadee, line 209:
```suggestion
            \sprintf(
                '/new-api/orders/%s/shipments/%s', 
                $this->sharedStorage->get('cart_token'),
                (string) $this->iriConverter->getItemFromIri($this->getCart()['shipments'][0])->getId()
            )
```

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -206,13 +206,13 @@ public function iProceededWithShippingMethod(ShippingMethodInterface $shippingMe
...
             [],
             [],
             $this->getHeaders(),
             json_encode([
-                'shipmentIdentifier' => 0,
-                'shippingMethod' => $shippingMethod->getCode(),
+                'shipmentId' => (string) $this->iriConverter->getItemFromIri($this->getCart()['shipments'][0])->getId(),
```
GSadee, line 214: And this will not be needed I guess

lchrusciel:
image
Why is this `shipmentId` duplicated? Can we have it only on a path?

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Order.xml
```diff
@@ -97,7 +97,7 @@
                     <attribute name="summary">Select shipping methods for particular shipment</attribute>
                 </attribute>
                 <attribute name="method">PATCH</attribute>
-                <attribute name="path">/orders/{id}/select-shipping-methods</attribute>
+                <attribute name="path">/orders/{id}/shipments/{shipmentId}</attribute>
```
lchrusciel, line 100: We may be forced to move this endpoint to Shipment.xml class. There we can perhaps define 
```suggestion
                <attribute name="path">/orders/{token}/shipments/{id}</attribute>
```
and execute the command based mostly on `id` related to the shipment

Tomanhez: I think that changing `id` name will be not easy: https://github.com/api-platform/core/pull/2126

src/Sylius/Bundle/ApiBundle/Command/CommandAwareDataTransformerInterface.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Command;
+
+interface CommandAwareDataTransformerInterface
```
Tomanhez, line 16: I added new tag interface, because some `DataTransformer` can be executed only one in certain object, if we have, for example two `DataTransformers` where `supportsTransformation (...)` equals to us conditions, will be executed first in line .

GSadee (approved):

src/Sylius/Bundle/ApiBundle/DataTransformer/CommandAwareInputDataTransformer.php
```diff
@@ -0,0 +1,75 @@
...
+
+        return $object;
+    }
+
+    private function assignSubresourceId(SubresourceIdAwareInterface $object, string $to, array $context = [])
+    {
+        $attributes = $this->requestStack->getCurrentRequest()->attributes->all();
+
+        /** @var string|null $subresourceId */
+        $subresourceId = $attributes[$object->getSubresourceIdAttributeKey()] ?? null;
```
lchrusciel, line 67: Instead of `all()` function, we should use `get` method with the fallback to default value.

src/Sylius/Bundle/ApiBundle/Command/Checkout/ChooseShippingMethod.php
```diff
@@ -50,4 +44,19 @@ public function setOrderTokenValue(?string $orderTokenValue): void
...
+    }
+
+    public function setSubresourceId(?string $subresourceId): void
+    {
+        $this->shipmentId = $subresourceId;
+    }
+
+    public function getSubresourceIdAttributeKey(): string
+    {
+        return 'shipmentId';
```
lchrusciel, line 60: IMHO, the command should not be aware of path definition details. An additional interface is ok, but I would prefer to make more detailed (get/setShipmentId), make additional transformer, where `shipmentId` will be injected into the constructor. 

Also, can you check if we can use parameters in route definition?

Tomanhez: I added `getSubresourceIdAttributeKey` instead of details like get/setShipmentId, because i would use it in next cases like payment, out of the box without duplicated boilerplate code.

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Order.xml
```diff
@@ -79,8 +79,8 @@
                 <attribute name="openapi_context">
                     <attribute name="summary">Address cart to given location, logged in Customer does not have to provide an email</attribute>
                 </attribute>
-                <attribute name="method">PATCH</attribute>
-                <attribute name="path">/orders/{id}/address</attribute>
+                <attribute name="method">PUT</attribute>
+                <attribute name="path">/orders/{id}</attribute>
```
Tomanhez, line 83: I'm not thrilled this solution, All checkout endpoints are individual and this changes is non consistent for others, It looks like `PUT` is for all order, not only for address

src/Sylius/Bundle/ApiBundle/DependencyInjection/Compiler/CommandDataTransformerPass.php
```diff
@@ -0,0 +1,40 @@
...
+        $commandDataTransformersChainDefinition = new Definition(CommandDataTransformersChain::class);
+
+        $taggedServices = $container->findTaggedServiceIds('sylius.api.command_data_transformer');
+
+        foreach ($taggedServices as $key => $value) {
+            $commandDataTransformersChainDefinition->addArgument(new Reference($key));
+        }
+
+        $container->setDefinition(
+            'sylius_bundle_api.command_data_transformers_chain',
```
GSadee, line 36:
```suggestion
            'sylius_api.command_data_transformers_chain',
```

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Order.xml
GSadee: I wonder if it wouldn't be better to leave `/address` in the path, it would be clearer for users and still consistent with other endpoints

GSadee (approved):

GSadee:
Thank you, Tomasz! :tada:

PR #11755, opened 2020-08-21
Avoid getting the theme if we have it already

jacquesbh:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

I've always had a pretty slow development environment.
So I decided to run a blackfire profile on it and found out that the Debug Toolbar was running a tons of SQL requests (thousands).

After a moment into the blackfire's graph, I found out this simple solution to avoid all these SQL requests (one per twig template, which leads to A LOT of SQL requests).

### On my homepage in dev mode

image

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Theme/ChannelBasedThemeContext.php
+++ b/src/Sylius/Bundle/CoreBundle/Theme/ChannelBasedThemeContext.php
@@ -28,28 +28,34 @@ final class ChannelBasedThemeContext implements ThemeContextInterface
     /** @var ThemeRepositoryInterface */
     private $themeRepository;
 
+    /** @var null|false|ThemeInterface */
+    private $theme = false;
+
     public function __construct(ChannelContextInterface $channelContext, ThemeRepositoryInterface $themeRepository)
     {
         $this->channelContext = $channelContext;
         $this->themeRepository = $themeRepository;
     }
 
+    /**
+     * @psalm-suppress InvalidReturnType
+     */
     public function getTheme(): ?ThemeInterface
     {
-        try {
-            /** @var ChannelInterface $channel */
-            $channel = $this->channelContext->getChannel();
-            $themeName = $channel->getThemeName();
-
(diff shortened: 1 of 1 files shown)


pamil (commented):
Great improvement! Could you implement it as a decorator? This way we could cache any implementation of a channel context.

jacquesbh:
Actually @pamil I would like to discuss this because I don't see the actual benefit of using a decorator.

We have 3 implementations right now and only this one is really resource consuming.

I agree for other implementations in general, but I don't think this behavior has been often changed. So for me using a decorator lead to more code and more complexity. And there is no need for it.

Could we talk about it on Slack maybe?

I reopen this for review in case you finally want to merge it.

🙃

lchrusciel (approved):

lchrusciel:
However, this logic should be cover by spec. Also, I would consider adding it to 1.8

jacquesbh:
> However, this logic should be cover by spec. Also, I would consider adding it to 1.8

Could you define "by spec"? :p. phpspec? something else?

Zales0123:
> > However, this logic should be cover by spec. Also, I would consider adding it to 1.8
> 
> Could you define "by spec"? :p. phpspec? something else?

Sorry @jacquesbh, it's been a long time :D I believe @lchrusciel meant PHPSpec modification, as the changed class is already covered with `ChannelBasedThemeContextSpec` 🖖

jacquesbh:
Ok @Zales0123 I'll do that :-)

jacquesbh:
Do not stale. I'll arrange the tests asap 😬

jakubtobiasz:
Hi 🙋🏼‍♂️!
Thank you very much for you efforts. As you might notice @NoResponseMate finished this PR in #14894, so I feel we can close this one.

Once again, thanks 🤙🏼!

PR #11772, opened 2020-08-28
Introduce `DashboardStatisticsInterface`

alanpoulain:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes (excluded from the BC promise?)
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

To allow modifying the `DashboardStatistics` model, for instance to add more statistics.
The interface was missing to do so.

Potentially a BC break if the `DashboardStatisticsProvider` has been extended (but excluded from the BC promise?).

### DIFF
--- a/src/Sylius/Component/Core/Dashboard/DashboardStatisticsProvider.php
+++ b/src/Sylius/Component/Core/Dashboard/DashboardStatisticsProvider.php
@@ -33,7 +33,10 @@ class DashboardStatisticsProvider implements DashboardStatisticsProviderInterfac
         $this->customerRepository = $customerRepository;
     }
 
-    public function getStatisticsForChannel(ChannelInterface $channel): DashboardStatistics
+    /**
+     * {@inheritdoc}
+     */
+    public function getStatisticsForChannel(ChannelInterface $channel)
     {
         return new DashboardStatistics(
             $this->orderRepository->getTotalPaidSalesForChannel($channel),
@@ -43,11 +46,14 @@ class DashboardStatisticsProvider implements DashboardStatisticsProviderInterfac
         );
     }
 
+    /**
+     * {@inheritdoc}
+     */
     public function getStatisticsForChannelInPeriod(
         ChannelInterface $channel,
         \DateTimeInterface $startDate,
         \DateTimeInterface $endDate
-    ): DashboardStatistics {
+    ) {
         return new DashboardStatistics(
             $this->orderRepository->getTotalPaidSalesForChannelInPeriod($channel, $startDate, $endDate),
             $this->orderRepository->countPaidForChannelInPeriod($channel, $startDate, $endDate),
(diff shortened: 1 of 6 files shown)


alanpoulain:
Test failing is not related.

lchrusciel (commented):
Hey Alan,

thanks a lot for your contribution. Interface here is truly a missing thing. We should add it. However, what you have proposed will work only since PHP 7.4 and we are still supporting 7.3. What we can do, is to remove the return typehint, and describe it in docblock above the method. Maybe with a note, that interface will be brought back once we will drop PHP 7.3 support. 

I've also restarted your build.

What is more, welcome to our community. It is nice to have you here ;)

alanpoulain:
Hello @lchrusciel and thank you.
I think you are talking about covariance in PHP 7.4, aren't you?
I don't think there is an issue with covariance here: I'm only replacing a class typehinting with an interface one.

lchrusciel:
Isn't replacing class typehint with an interface a widening of contract? Without covariance it may end up with an error, assuming, that one would implement custom `DashboardStatisticsProvider ` already as it was available since Sylius 1.7.

src/Sylius/Component/Core/Dashboard/DashboardStatisticsProvider.php
```diff
@@ -33,7 +33,7 @@ public function __construct(
         $this->customerRepository = $customerRepository;
     }
 
-    public function getStatisticsForChannel(ChannelInterface $channel): DashboardStatistics
+    public function getStatisticsForChannel(ChannelInterface $channel): DashboardStatisticsInterface
```
lchrusciel, line 36: Also, why didn't you adjust the second method from this class as well?

alanpoulain: It's totally an oversight, sorry.

alanpoulain:
Oh you're talking about the BC break then. Sure, it will be a BC break if the user has implemented the `DashboardStatisticsProvider` and if PHP is < 7.4.

But from [this page](https://docs.sylius.com/en/latest/book/organization/backward-compatibility-promise.html), it says:
> In order to fulfill the constant Sylius’ need to evolve, model interfaces are excluded from this BC promise.

That's why I was wondering if it was important to do like you say.

lchrusciel:
That's correct, but this class didn't have an interface in the first place. Adding it wouldn't be a BC break. However, changing the returned type hint is. Your PR is a valid request, but in this case, it would be better to move the return typehint to the docblock rather than breaking someone's app.

alanpoulain:
Typehint replaced by PHPDoc :slightly_smiling_face:

Zales0123 (approved):

jakubtobiasz:
This PR will be no more valid after #15598, so I'm closing it. Thanks anyway.

PR #11774, opened 2020-08-31
[User] Remove usage of deprecated `AdvancedUserInterface`

stloyd:
That class was remove in Symfony 5, re-added its method into `UserInterface` instead to prevent fatal errors.

| Q               | A
| --------------- | -----
| Branch?         | 1.8(?)
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no(?)
| Deprecations?   | no
| Related tickets | related to #10928 
| License         | MIT

### DIFF
--- a/UPGRADE-1.8.md
+++ b/UPGRADE-1.8.md
@@ -274,6 +274,8 @@ time out during this ajax request (previously no serialization group was defined
 in some admin URLs you can now replace `/admin` by `/%sylius_admin.path_name%`.  
 Also the route is now dynamic. You can change the `SYLIUS_ADMIN_ROUTING_PATH_NAME` environment variable to custom the admin's URL.
 
+1. Due to its deprecation, we removed the usage of `Symfony\Component\Security\Core\User\AdvancedUserInterface` in `Sylius\Component\User\Model\UserInterface` (the missing methods were added to the latter). If you target the `AdvancedUserInterface` anywhere in your code, extend it again in your interface or switch the type hint.
+
 ## Special attention
 
 ### Migrations
--- a/src/Sylius/Bundle/CoreBundle/Security/UserChecker.php
+++ b/src/Sylius/Bundle/CoreBundle/Security/UserChecker.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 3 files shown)


lchrusciel (commented):
If we would move to 5.0 only, then we could easily remove this class. However, now it is kind of bc-break. Maybe some polifil or conditional implementation would be a better solution here? 

or do we accept this BC-Break (and then it should be added to Upgrade file)? /cc @Sylius/core-team

lchrusciel:
After a second thought, let's just add a note to the UPGRADE.md file.

We should note that if one was targeting this interface, they should either add this interface to their classes or target our interface.

stloyd:
This is still valid unfortunately.

UPGRADE-1.8.md
```diff
@@ -224,6 +224,8 @@ time out during this ajax request (previously no serialization group was defined
 in some admin URLs you can now replace `/admin` by `/%sylius_admin.path_name%`.  
 Also the route is now dynamic. You can change the `SYLIUS_ADMIN_ROUTING_PATH_NAME` environment variable to custom the admin's URL.
 
+1. Remove usage of deprecated `Symfony\Component\Security\Core\User\AdvancedUserInterface` in `Sylius\Component\User\Model\UserInterface`, to prevent BC break, added methods from the removed interface into existing one.
```
Zales0123, line 227: Maybe something more explainable:

1. Due to its deprecation, we removed the usage of `Symfony\Component\Security\Core\User\AdvancedUserInterface` in `Sylius\Component\User\Model\UserInterface` (the missing methods were added to the latter). If you target the `AdvancedUserInterface` anywhere in your code, extend it again in your interface or switch the type hint.

?

Zales0123:
Ok, as far as I see there is one more problem. The `UserChecker` service requires the user entity to implement `AdvancedUserInterface` (see [this](https://github.com/symfony/security-core/blob/4.4/User/UserChecker.php#L31)). So we should somehow duplicate this behaviour if we want to remove this interface and keep the existing features.

pamil:
may become messed-up.**

execute `git fetch origin && git reset "patch-4"` to update your local branch.
 

PR #11781, opened 2020-08-31
Add missing new method into form extensions

stloyd:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | related #10928 
| License         | MIT

Error with newer Symfony:
```bash
PHP Fatal error:  Class Sylius\Bundle\CoreBundle\Form\Extension\ChannelTypeExtension contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Symfony\Component\Form\FormTypeExtensionInterface::getExtendedTypes)
```

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Form/Extension/CartItemTypeExtension.php
+++ b/src/Sylius/Bundle/CoreBundle/Form/Extension/CartItemTypeExtension.php
@@ -70,4 +70,9 @@ final class CartItemTypeExtension extends AbstractTypeExtension
     {
         return CartItemType::class;
     }
+
+    public static function getExtendedTypes(): iterable
+    {
+        return [CartItemType::class];
+    }
 }
--- a/src/Sylius/Bundle/CoreBundle/Form/Extension/CartTypeExtension.php
+++ b/src/Sylius/Bundle/CoreBundle/Form/Extension/CartTypeExtension.php
@@ -54,4 +54,9 @@ final class CartTypeExtension extends AbstractTypeExtension
     {
         return CartType::class;
     }
+
+    public static function getExtendedTypes(): iterable
+    {
+        return [CartType::class];
+    }
 }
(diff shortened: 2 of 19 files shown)


src/Sylius/Bundle/CoreBundle/Form/Extension/CartItemTypeExtension.php
```diff
@@ -70,4 +70,9 @@ public function getExtendedType(): string
     {
         return CartItemType::class;
     }
+
+    public static function getExtendedTypes(): iterable
+    {
+        return [CartItemType::class];
```
lchrusciel, line 76: What would you say for:
```suggestion
        return [$this->getExtendedType()];
```
or
```suggestion
        yield $this->getExtendedType();
```
to reuse still existing method?

lchrusciel (commented):
Hey Józef,

thanks for contribution, this change will be needed for sure 👍  I left one proposal to improve this BC layer

src/Sylius/Bundle/CoreBundle/Form/Extension/CartItemTypeExtension.php
stloyd: @lchrusciel Wrong button... this cannot be done like this, cause new method is `static`, thus it's not called on object (no `$this` etc.), but statically like `ClassType:: getExtendedType()`.

lchrusciel: You are right 👍

lchrusciel:
Nonetheless, the build is red and it seems to be related to your changes, isn't it?

lchrusciel (approved):

lchrusciel:
Thank you, Joseph! :1st_place_medal:

PR #11782, opened 2020-08-31
[Maintenance][HotFix] Conflict with Symofny/Form 4.4.12

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Just a tmp solution, as this is coming back into us for the forth time.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -193,7 +193,7 @@
         "nikic/php-parser":"^4.7",
         "doctrine/inflector": "^1.4",
         "sylius/grid-bundle": "1.7.4",
-        "symfony/form": "4.4.9|4.4.10|4.4.11",
+        "symfony/form": "4.4.9|4.4.10|4.4.11|4.4.12",
         "symfony/symfony": "3.4.7",
         "twig/twig": "2.6.1",
         "phpspec/prophecy": "1.11.0"


Tomanhez (approved):

stloyd:
@lchrusciel What was the issue? Cause it's now kinda lost in the history of commits :(

lchrusciel:
I don't remember, what was a bug for the previous 3 versions, I can describe the current behavior only. 

At the moment bug can be spotted on https://travis-ci.org/github/Sylius/Sylius/jobs/722721300#L1758 (two scenarios are failing). 

The problem is as follows: if you try to review product without any data, then validation will work as expected: 
image
with following validation calls:
image
If you, however, fill review related data (but omit email address) and submit form:
image
you will stay on the same page, but without any errors displayed. Probably due to different validation calls:
image

There is a clear difference in the validated field resolution(`[author]` vs `children[author]`), where we could clearly expect exactly the same behavior. 

The issue is probably related to changes introduced somewhere here: https://github.com/symfony/form/compare/v4.4.8...v4.4.9 or https://github.com/symfony/form/commit/fa833d8726c11ced06b51e0d4350f5192cb92270#diff-d14286dfb8590b9989ecc626176b2759

AdamKasp:
Thank you, Łukasz! :tada:

PR #11783, opened 2020-08-31
Reset credentials expiration on password reset

Jeroen-G:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no/yes
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

I am not sure _if_ you want this, or where, but in our use case we want customers to reset their password when their credentials have expired. And therefore, when the password is updated, this expiration date should be removed. Let me know what you think!

### DIFF
--- a/src/Sylius/Bundle/UserBundle/EventListener/PasswordUpdaterListener.php
+++ b/src/Sylius/Bundle/UserBundle/EventListener/PasswordUpdaterListener.php
@@ -60,5 +60,9 @@ class PasswordUpdaterListener
         if (null !== $user->getPlainPassword()) {
             $this->passwordUpdater->updatePassword($user);
         }
+        
+        if (null !== $user->getCredentialsExpireAt()) {
+            $user->setCredentialsExpireAt(null);
+        }
     }
 }


lchrusciel:
Calls you added here are not used anywhere, so I would prefer not to add it to the core. However, we can, of course, wait for another opinion about it. Let's request for comments ;)

vvasiloi:
@lchrusciel, that value is used for reading, but not for writing.
It's not obvious, because the setter and getter are from the Sylius User interface and are not used anywhere.
However, the property value is read directly in `isCredentialsNonExpired`, which used to be part of Symfony Adranced User Interface, but now it's in the polyfill from SyliusLabs: `SyliusLabs\Polyfill\Symfony\Security\Core\User\AdvancedUserInterface::isCredentialsNonExpired`.
It is then used in `SyliusLabs\Polyfill\Symfony\Security\Core\User\UserChecker::checkPostAuth`, which is called in `Sylius\Bundle\UserBundle\Security\UserLogin::login`, which by default is only used in `\Sylius\Bundle\ShopBundle\EventListener\UserRegistrationListener::enableAndLogin`.

In my opinion, the whole security layer in Sylius must be refactored, because it is already far behind the Symfony's Security.
Everything related to credential expiration feature was deprecated in Symfony, but we could implement it from scratch in Sylius.

Rafikooo:
Hello @Jeroen-G, thank you for submitting this pull request.
After reviewing the proposed solution, we have decided that it is rather project specific and does not align with our goals for the Sylius core. Therefore, we will not be merging this pull request.
Regarding the security refactoring, we agree that it is an important area that needs to be addressed. However, we are currently unable to make changes to the security implementation until Sylius 2.0. For the time being, we have to stay with the existing implementation. We appreciate your effort in contributing this pull request and encourage you to submit more in the future.
Best regards 🚀

PR #11785, opened 2020-08-31
[WIP] Bump ResourceBundle to 1.7, update related dependencies

mbabker:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

This updates the resource bundle to the 1.7 branch (currently master).  As a result, it also makes the following updates to various dependencies:

- Bump FOSRestBundle to 3.0 (new major)
- Bump JMSSerializerBundle to 3.0 (new major)
- Bump the HATEOAS library to 3.0 (new major)
- Bump BazingaHateoasBundle to 2.0 (new major)
- Allow winzouStateMachineBundle 0.4
- Replace WhiteOctoberPagerfantaBundle with BabDevPagerfantaBundle


mbabker:
BTW if anyone's got any ideas on why the tests blew up, I'm all eyes.  Some of it seems to be a changed exception format for errors (does that need a B/C layer somewhere?), some of it is tied to a now open PR in ResourceBundle, but some of it seems like one of the updated bundles aren't fully upgraded and the configuration just exploded as a result.

mbabker:
I'm tapping out on this PR because after two-plus days with several hours of troubleshooting, I'm getting nowhere and the issue is boiling down to an integration I'm not all that familiar with.

Aside from the alias visibility problem (which there is a PR on ResourceBundle to fix), the other big issue that is in trying to serialize the `Hateoas\Representation\PaginatedRepresentation` object after `Sylius\Bundle\ResourceBundle\Controller\ResourcesCollectionProvider` has collected the resources, something is trying to serialize the entity manager and it ends up crawling up the class chain from there to a bunch of classes that shouldn't even be involved in the serialization context in the first place.  Looking at the metadata that gets dumped into cache, it eventually gets up to the Container instance and is trying to walk through objects such as the ORM mapping drivers, Bundle classes, cache adapters, and the kernel.  I regularly use most of the packages included in Sylius minus the `willdurand/hateoas` and `willdurand/hateoas-bundle` packages and can honestly say I've not seen an error anything like this in all my time working with them, and after going over all of the upgrade guides for all of the packages upgraded with this pull request (and my own experiences with upgrading in general), I can't figure out if the issue is because something wasn't upgraded correctly, there's a breaking change not accounted for, or some other thing I'm just overlooking.  And at this point, I can't keep spinning my wheels on the problem.

I'll leave this branch on my fork if anyone's interested, but I don't think I can offer much else of value here.

lchrusciel:
Hey Michel, thanks for your proposal anyway. I hope someone will have space to pick your work from here.

PR #11797, opened 2020-09-02
Fixed Twig translate extension not working with `null` values

stloyd:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | related #10928 
| License         | MIT

@lchrusciel I think this is one of most unexpected BC breaks.

Fatal error due to type-hint change:
```
# Before
/**
 * @param string $id
 */
public function trans($id, array $parameters = [], string $domain = null, string $locale = null);
# Now
public function trans(string $id, array $parameters = [], string $domain = null, string $locale = null);
```


lchrusciel:
Hey Józef,

your change makes sense and is valid according to the docblock: 
image

Let's merge it to the master. I've restarted one job that have failed.

lchrusciel:
Thank you, Joseph! :1st_place_medal:

Issue #11689, opened 2020-07-27
Non-translatable attributes


mamazu:
**Describe the proposed solution**
There are some attributes that need to be translated: like material (of a shirt) or color but there are also some attributes which don't need to be translated like "date of manufacturing", "delivery time in days" etc.

**Describe alternatives you've considered**
The current solution we are using is the translated attributes with all translations being the same. This however produces a very large amount of data.

Eg. We have about 15k products in our shop with 30 attributes in 5 locales. This makes about 1 million entries in the translation table. But all of them are the same.

**Additional context**
Implementation ideas:
* A: Make the `locale` of the attribute value translations null and if they are null it's a universal translation. This would make migrating to them easier. (Structure is the same but might break somewhere else)
* B: Create a new property on the product with attributes (a bigger change would be necessary to display them.)

Zales0123:
I believe this issue would be solved if we finally integrate this new, awesome attributes UI https://github.com/Sylius/Sylius/pull/10543

mamazu:
Not really, the UI makes creating these values easier (as you can use them for all locales directly). 
But it doesn't solve the problem that those values are redundant.

Zales0123:
Maybe I've just presumed, that together with changing the UI we will change the whole attributes structure as well 💃 🚀

mamazu:
That would be great. If you already have an implementation approach for it in mind let me know and I would be happy to help.

pamil:
For the read logic, could we make `localeCode` property nullable on `AttributeValue` and then fall back to attribute value which does not have locale code assigned?

This way we could support the current behaviour but allow for non-translatable attributes at the same time. Might be worth to prepare a proof of concept first to see whether it's a suitable solution.

mamazu:
So option A it is. :D Sounds like a good idea. I will try to start to work on a proof of concept hopefully next week.

mamazu:
Poke

lchrusciel:
I guess we can finally close this issue

PR #11826, opened 2020-09-09
[AttributeBundle] Non translatable attributes

JakobTolkemit:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #11689
| License         | MIT

Adds a `translatable` flag to attributes. 
Non translatable attributes have *one* value for all locales.

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Product/Attribute/attributesCollection.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Product/Attribute/attributesCollection.html.twig
@@ -6,7 +6,7 @@
 
     <div class="ui top attached tabular menu">
 
-        <a class="item active" data-tab="__not_translatable__">Not translatable</a>
+        <a class="item active" data-tab="__not_translatable__">{{ 'sylius.ui.not_translatable'|trans }}</a>
         {% for localeCode, translationForm in attr.translations %}
             <a class="item" data-tab="{{ localeCode }}">
                 {{ flags.fromLocaleCode(localeCode) }} {{ localeCode|sylius_locale_name }}
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Product/Attribute/attributesCollection.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Product/Attribute/attributesCollection.html.twig
@@ -5,22 +5,34 @@
     {% import _self as self %}
 
     <div class="ui top attached tabular menu">
+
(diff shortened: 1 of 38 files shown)


src/Sylius/Bundle/AdminBundle/Resources/views/Product/Attribute/attributesCollection.html.twig
```diff
@@ -5,22 +5,34 @@
     {% import _self as self %}
 
     <div class="ui top attached tabular menu">
+
```
GSadee, line 8:
```suggestion
```

src/Sylius/Bundle/CoreBundle/Migrations/Version20171003103916.php
```diff
@@ -37,7 +37,18 @@ public function up(Schema $schema): void
         $defaultLocale = $this->container->getParameter('locale');
         $productAttributeRepository = $this->container->get('sylius.repository.product_attribute');
 
-        $productAttributes = $productAttributeRepository->findBy(['type' => SelectAttributeType::TYPE]);
+        $productAttributes = $productAttributeRepository
```
GSadee, line 40: The existing migrations shouldn't be changed

JakobTolkemit: The problem is, this migration is blocking changes to the attribute mapping as described in: https://github.com/Sylius/Sylius/issues/11592

JakobTolkemit: https://github.com/Sylius/Sylius/pull/11624 was actually supposed to have fixed that.

GSadee:
I've pulled your branch and checked these attributes and to be honest, it doesn't look like it's working. When I try to add a new not translatable attribute to the product, It doesn't appear in the list of attributes. Could you provide some behat scenarios for adding a not translatable attribute to the existing product? Similar to [these ones](https://github.com/Sylius/Sylius/blob/master/features/product/managing_products/adding_text_attributes_to_existing_product.feature)

src/Sylius/Bundle/CoreBundle/Migrations/Version20171003103916.php
JakobTolkemit: This is now fixed in the master

Tomanhez:
This PR is continue on: https://github.com/Sylius/Sylius/pull/11976

PR #11827, opened 2020-09-09
[API] - Add product association endpoint

GalloisLuca:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

API feature, we can now get product associations


CoderMaggie:
Hello Luca!

Sylius already has a Product Associations feature, where you can create several types of associations (like: similar products, accessories, other customers also bought etc.), then having these association types, on a product you can choose which products will be associated with it in the chosen association type.

Docs: https://docs.sylius.com/en/latest/book/products/product_associations.html
Demo: [Association Types](https://demo.sylius.com/admin/product-association-types/) and on any [Product's](https://demo.sylius.com/admin/products/) edit page the "Associations" tab. 

Isn't it the same thing you are trying to achieve with your PR?

If so, then you should be covering the existing scenarios.

If you've meant to prepare a new feature, as you have in the scenarios ([here](https://github.com/Sylius/Sylius/blob/master/features/product/managing_products/adding_product_with_associations.feature), [here](https://github.com/Sylius/Sylius/blob/master/features/product/managing_products/adding_associations_to_existing_product.feature) and [there](https://github.com/Sylius/Sylius/blob/master/features/product/managing_products/changing_associations_of_product.feature)), so:
- adding a new association not from the product but separate endpoint;
- getting a list of all associations, not only the chosen product's associations;
- and in general: operation on associations themself not on the product and then on its associations

then it is a new concept of which I am not sure if we'd want to have it in the Sylius core. We'd rather have the current behavior covered with API.

GalloisLuca:
Hello @CoderMaggie, I changed some of my scenarios to adapted the existent ones with the @api, I kept the browsing_product_association.feature because there is no similar scenarios. I couldn't adapt this [test](https://github.com/Sylius/Sylius/blob/master/features/product/managing_products/adding_product_with_associations.feature) with the API. Waiting for your reply :)

vvasiloi:
This might be useful for an implementation in the [new API ](https://github.com/Sylius/Sylius/issues/11250).

jakubtobiasz:
Done in #15639. Thanks for your help!

PR #11894, opened 2020-09-30
[Order] Do not stop cancelling unpaid orders on exception during canc…

### LINKED ISSUE #11768: sylius:cancel-unpaid-orders halts if exception thrown on single order
**Sylius version affected**: 1.3.x and up

The command `sylius:cancel-unpaid-orders` is being used in our shop regularly (cron scheduled) to cancel old orders. In one of the event listeners to this event, we send the customer an email. All fine, but I just noticed someone entered `gmail..com` in their email address. This results in an exception on the sending of that email, which occurs as soon as the state machine kicks in to cancel the order.

Now one could reason that the email sending logic should handle this (the event listener in this case), which makes full sense. But next to this, any exception that happens inside a single order will make this command stop instantly and not process the other orders. This can keep blocking forever, making the list of unpaid orders big and they never get cancelled. 

From that perspective it might be good to catch exceptions separately and let the cancel updater continue to the next order.

### LINKED ISSUE #11945: Cancel unpaid orders command should process next orders upon error
The command "sylius:cancel-unpaid-orders" should be atomic on each order cancellation, so that if an order fails to be canceled, andit throws an error, next orders can still be processed for cancellation. This could be done easily with a try/catch and a writeln to output or a message in the log.

twojtylak:
…elling a single order

| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #11768 & fixes #11945
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
@@ -186,6 +186,7 @@
             <argument type="service" id="sylius.repository.order" />
             <argument type="service" id="sm.factory" />
             <argument>%sylius_order.order_expiration_period%</argument>
+            <argument type="service" id="monolog.logger" />
         </service>
         <service id="Sylius\Component\Core\Updater\UnpaidOrdersStateUpdaterInterface" alias="sylius.unpaid_orders_state_updater" />
 
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
@@ -186,7 +186,7 @@
             <argument type="service" id="sylius.repository.order" />
             <argument type="service" id="sm.factory" />
             <argument>%sylius_order.order_expiration_period%</argument>
-            <argument type="service" id="monolog.logger" />
+            <argument type="service" id="logger" />
         </service>
         <service id="Sylius\Component\Core\Updater\UnpaidOrdersStateUpdaterInterface" alias="sylius.unpaid_orders_state_updater" />
 
--- a/src/Sylius/Component/Core/Updater/UnpaidOrdersStateUpdater.php
+++ b/src/Sylius/Component/Core/Updater/UnpaidOrdersStateUpdater.php
@@ -62,7 +62,7 @@ final class UnpaidOrdersStateUpdater implements UnpaidOrdersStateUpdaterInterfac
             try {
                 $this->cancelOrder($expiredUnpaidOrder);
             } catch (\Exception $e) {
-                if ($this->logger instanceof LoggerInterface) {
+                if (null !== $this->logger) {
                     $this->logger->error(
                         sprintf('An error occurred while cancelling unpaid order #%s', $expiredUnpaidOrder->getId()),
                         ['exception' => $e, 'message' => $e->getMessage()]
--- a/src/Sylius/Component/Core/Updater/UnpaidOrdersStateUpdater.php
+++ b/src/Sylius/Component/Core/Updater/UnpaidOrdersStateUpdater.php
@@ -30,7 +30,7 @@ final class UnpaidOrdersStateUpdater implements UnpaidOrdersStateUpdaterInterfac
     /** @var string */
     private $expirationPeriod;
 
-    /** @var LoggerInterface */
+    /** @var LoggerInterface|null */
     private $logger;
 
     /**
@@ -40,11 +40,18 @@ final class UnpaidOrdersStateUpdater implements UnpaidOrdersStateUpdaterInterfac
         OrderRepositoryInterface $orderRepository,
         Factory $stateMachineFactory,
         $expirationPeriod,
-        LoggerInterface $logger
+        LoggerInterface $logger = null
     ) {
         $this->orderRepository = $orderRepository;
         $this->stateMachineFactory = $stateMachineFactory;
         $this->expirationPeriod = $expirationPeriod;
+        if (null === $logger) {
+            @trigger_error(
+                'Not passing a logger is deprecated since 1.7',
+                \E_USER_DEPRECATED
+            );
+        }
+
         $this->logger = $logger;
     }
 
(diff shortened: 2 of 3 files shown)


src/Sylius/Component/Core/Updater/UnpaidOrdersStateUpdater.php
```diff
@@ -29,24 +30,36 @@ final class UnpaidOrdersStateUpdater implements UnpaidOrdersStateUpdaterInterfac
...
+
     /**
      * @param string $expirationPeriod
      */
     public function __construct(
         OrderRepositoryInterface $orderRepository,
         Factory $stateMachineFactory,
-        $expirationPeriod
+        $expirationPeriod,
+        LoggerInterface $logger
```
loevgaard, line 43:
```suggestion
        LoggerInterface $logger = null
```

You need to allow this to be null, else it's a BC break and can't be in 1.x. You could (should) trigger a deprecation warning if this is null

src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
```diff
@@ -186,6 +186,7 @@
             <argument type="service" id="sylius.repository.order" />
             <argument type="service" id="sm.factory" />
             <argument>%sylius_order.order_expiration_period%</argument>
+            <argument type="service" id="monolog.logger" />
```
loevgaard, line 189: Isn't there a service just called `logger`?

loevgaard (changes requested):

loevgaard:
Also please add to your PR description: `Fixes #11945`

src/Sylius/Component/Core/Updater/UnpaidOrdersStateUpdater.php
twojtylak: Done

src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
twojtylak: Done. `logger` is an alias for "monolog.logger", so I think it wont make any difference

loevgaard:
Your `Fixes` statement only closes 11768 right now. I think you should prepend `Fixes` to each issue number to have it closed :) Just a heads up ;)

twojtylak:
> Your `Fixes` statement only closes 11768 right now. I think you should prepend `Fixes` to each issue number to have it closed :) Just a heads up ;)

Alright I added `fixes` in the pr description 😄

src/Sylius/Component/Core/Updater/UnpaidOrdersStateUpdater.php
```diff
@@ -29,24 +30,43 @@ final class UnpaidOrdersStateUpdater implements UnpaidOrdersStateUpdaterInterfac
...
 
     public function cancel(): void
     {
         $expiredUnpaidOrders = $this->orderRepository->findOrdersUnpaidSince(new \DateTime('-' . $this->expirationPeriod));
         foreach ($expiredUnpaidOrders as $expiredUnpaidOrder) {
-            $this->cancelOrder($expiredUnpaidOrder);
+            try {
+                $this->cancelOrder($expiredUnpaidOrder);
+            } catch (\Exception $e) {
+                $this->logger && $this->logger->error(
```
lchrusciel, line 65: Can we make explicit if statement to improve readability?

src/Sylius/Component/Core/spec/Updater/UnpaidOrdersStateUpdaterSpec.php
```diff
@@ -24,9 +25,9 @@
 
 final class UnpaidOrdersStateUpdaterSpec extends ObjectBehavior
 {
-    function let(OrderRepositoryInterface $orderRepository, Factory $stateMachineFactory): void
+    function let(OrderRepositoryInterface $orderRepository, Factory $stateMachineFactory, LoggerInterface $logger): void
     {
-        $this->beConstructedWith($orderRepository, $stateMachineFactory, '10 months');
+        $this->beConstructedWith($orderRepository, $stateMachineFactory, '10 months', $logger);
```
lchrusciel, line 30: Can we make a unit test for behavior without logger?

lchrusciel (commented):
Hey @twojtylak,

thanks for the contribution. I think it would be worthy to have it merged in the Core. Can you make two more improvements in your code?

src/Sylius/Component/Core/Updater/UnpaidOrdersStateUpdater.php
twojtylak: Hey @lchrusciel,

I just added an explicit `instance of` check.

src/Sylius/Component/Core/spec/Updater/UnpaidOrdersStateUpdaterSpec.php
twojtylak: Hey @lchrusciel,

I added the `it_wont_stop_cancelling_unpaid_orders_on_exception_for_a_single_order_and_skips_logging_if_logger_is_not_set` (I am not sure if the name is meaningful enough or too long) test. I hope that's okay. Cheers

src/Sylius/Component/Core/Updater/UnpaidOrdersStateUpdater.php
```diff
@@ -29,24 +30,45 @@ final class UnpaidOrdersStateUpdater implements UnpaidOrdersStateUpdaterInterfac
...
 
     public function cancel(): void
     {
         $expiredUnpaidOrders = $this->orderRepository->findOrdersUnpaidSince(new \DateTime('-' . $this->expirationPeriod));
         foreach ($expiredUnpaidOrders as $expiredUnpaidOrder) {
-            $this->cancelOrder($expiredUnpaidOrder);
+            try {
+                $this->cancelOrder($expiredUnpaidOrder);
+            } catch (\Exception $e) {
+                if ($this->logger instanceof LoggerInterface) {
```
loevgaard, line 65:
```suggestion
                if (null !== $this->logger) {
```

src/Sylius/Component/Core/spec/Updater/UnpaidOrdersStateUpdaterSpec.php
lchrusciel: Hey @twojtylak, it is understandable enough ;)

lchrusciel:
Hey @twojtylak, I've restarted also your build.

loevgaard (approved):

lchrusciel:
Thanks, @twojtylak! :1st_place_medal:

LucaGallinari:
Thanks!

Issue #11864, opened 2020-09-18
On 1.8, bin/console doctrine:migrations:diff generates migrations in `vendor/sylius/sylius/src/Sylius/Bundle/AdminApiBundle/Migrations/` dir


armellarcier:
**Sylius version affected**: 1.8.0

**Description**  
When I try to generate a new migration, it gets generated in `vendor/sylius/sylius/src/Sylius/Bundle/AdminApiBundle/Migrations/` instead of `src/Migrations/`

Is that ok?

**Steps to reproduce**  
Migrate to 1.8, install bitbag CMS and start `bin/console doctrine:migrations:diff`

I'll try with just an entity and not a plugin

**Possible Solution**  
Don't remove some lines in `config/packages/doctrine_migrations.yaml` as stated in `UPGRADE-1.8.md` ?

I know this is not the solution with doctrine migrations 3…

```
-    dir_name: "%kernel.project_dir%/src/Migrations"
-
-    # Namespace is arbitrary but should be different from App\Migrations as migrations classes should NOT be autoloaded
-    namespace: DoctrineMigrations
```

stloyd:
Have a look at this: https://github.com/SyliusLabs/DoctrineMigrationsExtraBundle/pull/5

chrissnyder2337:
I agree with @armellarcier. I would recommend, at the very least, updating `./UPGRADE-1.8.md` to at least note in which situations you should not remove those lines. 

Making the UPGRADE documentation as complete as possible for most situations improves DX.

chrissnyder2337:
One more thought. If `./UPGRADE-1.8.md` is changed to note that if you do keep your own migrations by adding the following to 
`./config/packages/doctrine_migrations.yaml`

```yaml
    migrations_paths:
        'DoctrineMigrations': 'src/Migrations'
```

you may also need to run the script below. This was true for me because I started the project using [Sylius-Standard 1.7](https://github.com/Sylius/Sylius-Standard) which starts you off with a set of migrations, causing duplicate migrations.

```bash
#!/usr/bin/env bash

rm -f 'src/Migrations/Version20170912085504.php'
rm -f 'src/Migrations/Version20170913125128.php'
rm -f 'src/Migrations/Version20180102140039.php'
rm -f 'src/Migrations/Version20190109095211.php'
rm -f 'src/Migrations/Version20190109160409.php'
rm -f 'src/Migrations/Version20190508083953.php'
[...]

PR #11898, opened 2020-10-01
Improve 1.8 migrations upgrade instructions

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/11864
| License         | MIT

This PR fixes the problem with upgrade to Sylius 1.8. I believe it can be done better (_duplicated_ migrations should be detected automatically), but at least it allows to update quite smoothly. The main goal is to delete redundant migrations provided previously in Sylius-Standard (as they're **indeed** not needed anymore) and mark already executed migrations as *migrated*. 

The change in extension is also fixing the bug, that migrations are generated in vendors instead of `src/Migrations` directory.

Thank you for all the people involved in fixing this problem, especially for @mmenozzi, @Jibbarth, @chrissnyder2337, Martjn... I hope I didn't miss anyone :) It would be great if you can follow the instruction and check out can you upgrade your store with them. 

I also believe we would be able to provide some more elegant solution to this problem as soon as possible 🖖

### DIFF
--- a/UPGRADE-1.8.md
+++ b/UPGRADE-1.8.md
@@ -147,6 +147,12 @@ time out during this ajax request (previously no serialization group was defined
 in some admin URLs you can now replace `/admin` by `/%sylius_admin.path_name%`.  
 Also the route is now dynamic. You can change the `SYLIUS_ADMIN_ROUTING_PATH_NAME` environment variable to custom the admin's URL.
 
+## Special attention
+
+### Migrations
+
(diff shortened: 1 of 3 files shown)


GSadee (approved):

lchrusciel (approved):

lchrusciel:
Thanks, Mateusz! :tada:

PR #11911, opened 2020-10-06
Removing typehint for Symfony 4.4 compatibility

mamazu:
In the [symfony/property-info 4.4](https://github.com/symfony/property-info/blob/v4.4.15/PropertyListExtractorInterface.php) the interface does not have a string typehint so Sylius does nolonger work with 4.4

| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no (fixing one)
| Deprecations?   | no
| Related tickets | -
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -86,7 +86,7 @@
         "symfony/polyfill-php80": "^1.17",
         "symfony/process": "^4.4",
         "symfony/property-access": "^4.4",
-        "symfony/property-info": "4.4.*",
+        "symfony/property-info": "^4.4",
         "symfony/proxy-manager-bridge": "^4.4",
         "symfony/routing": "^4.4",
         "symfony/security-bundle": "^4.4",
--- a/composer.json
+++ b/composer.json
@@ -86,6 +86,7 @@
         "symfony/polyfill-php80": "^1.17",
         "symfony/process": "^4.4",
         "symfony/property-access": "^4.4",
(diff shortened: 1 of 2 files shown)


stloyd:
This should be merged into `1.8` as bugfix.

lchrusciel:
And we should force this lib to be installed in v4.4, to ensure that it works as expected

mamazu:
It also works with 5 as the signature has parameter wider than the interface.

lchrusciel:
I'm aware of that, but we do not test it at all. And still, we do not support Symfony 5.0 anyway

mamazu:
Okay, then should I add this?

lchrusciel:
If you can, yes please.

composer.json
```diff
@@ -86,6 +86,7 @@
         "symfony/polyfill-php80": "^1.17",
         "symfony/process": "^4.4",
         "symfony/property-access": "^4.4",
+        "symfony/property-info": "4.4.*",
```
lchrusciel, line 89:
```suggestion
        "symfony/property-info": "^4.4",
```
It won't change much, but I would prefer to keep it consistent with others constraints.

lchrusciel:
Thank you, @mamazu! :1st_place_medal:

PR #11922, opened 2020-10-12
Replace http with https in src/ directory

### LINKED ISSUE #9601: Channels urls use `http`
**Sylius version affected**: master

**Description**  
Channels can have only `http` urls - not `https`. Example: https://github.com/Sylius/Sylius/blob/281d33c1c7b3b83e9ac420e7be917e4632cb2593/src/Sylius/Bundle/ShopBundle/Resources/views/Email/verification.html.twig#L6. It would be nice if it was  possible to configure https.

loevgaard:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | Some would say since we are in 2020 and Google and others **really** encourages the use of https over http
| New feature?    | no
| BC breaks?      | In very rare cases I guess
| Deprecations?   | no
| Related tickets | fixes #9601
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Email/layout.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Email/layout.html.twig
@@ -1,12 +1,12 @@
 {% block body %}
     {% autoescape false %}
-        {% set logo = channel.hostname is not null ? 'http://' ~ channel.hostname ~ asset('assets/admin/img/logo.png') : absolute_url(asset('assets/admin/img/logo.png')) %}
+        {% set logo = channel.hostname is not null ? 'https://' ~ channel.hostname ~ asset('assets/admin/img/logo.png') : absolute_url(asset('assets/admin/img/logo.png')) %}
 
         <div style="font-family: sans-serif; background: #eee; padding-top: 30px; padding-bottom: 30px; font-size: 16px; color: #333; line-height: 1.5">
             <div style="max-width: 600px; margin: 20px auto;">
                 <div style="margin-bottom: 30px; text-align: center;">
                     {% if sylius_bundle_loaded_checker('SyliusShopBundle') %}
-                        {% set url = channel.hostname is not null ? 'http://' ~ channel.hostname ~ path('sylius_shop_homepage', {'_locale': localeCode}) : url('sylius_shop_homepage', {'_locale': localeCode}) %}
+                        {% set url = channel.hostname is not null ? 'https://' ~ channel.hostname ~ path('sylius_shop_homepage', {'_locale': localeCode}) : url('sylius_shop_homepage', {'_locale': localeCode}) %}
                         <a href="{{ url|raw }}">
                             <img src="{{ logo }}" alt="Sylius" style="width: 170px">
                         </a>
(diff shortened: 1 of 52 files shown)


Prometee:
@loevgaard I would go for a `channel.scheme` or another getter getting the scheme and the hostname when we need it in a twig template. But it's a very good start ;)

loevgaard:
Yeah, but that would definitely make this a `1.9` change. I was hoping for a 1.7.x and 1.8.x change, but it depends on what the core team says of course :)

src/Sylius/Bundle/AdminBundle/Resources/views/Email/layout.html.twig
```diff
@@ -1,12 +1,12 @@
 {% block body %}
     {% autoescape false %}
-        {% set logo = channel.hostname is not null ? 'http://' ~ channel.hostname ~ asset('assets/admin/img/logo.png') : absolute_url(asset('assets/admin/img/logo.png')) %}
+        {% set logo = channel.hostname is not null ? 'https://' ~ channel.hostname ~ asset('assets/admin/img/logo.png') : absolute_url(asset('assets/admin/img/logo.png')) %}
```
lchrusciel, line 3: Only relevant changes are done in twig templates. What do you think about global twig config, which will toggle `http` and `https`? The default configuration should use https, but we will provide an easy way for backward compatibility.

loevgaard: Not sure what you mean by global twig config. You mean adding a global twig variable containing this information?

Prometee: Adding a `scheme` attribute to the `Channel` entity could be also an idea. But it will require to add it to the admin form and default value should be `http` to ensure backward compatibility.

vvasiloi: Hoping that it's not too late...
Having it configurable per channel doesn't make much sense to me, because why would one have some channels on http and others on https? Besides that, it requires a lot more extra effort than adding a twig global variable.
I suggest adding an environment variable and use it's value for a Twig global variable.
Maybe also define an intermediate container parameter...

vvasiloi:
@loevgaard would you give it another try, please? 😄

loevgaard:
Right now I don't have time to recreate this PR, but my opinion is that a quick search replace for `http => https` should be enough. in 2022 I don't see a scenario where we don't want `https`. If anybody else wants to make a twig global variable I am fine with that of course, I just don't see the need

vvasiloi:
I think it's mostly for development environments without SSL.
Maybe this should be split in 2 PR, one for dynamic URLs and another for static.

loevgaard:
But don't people have ssl for dev env in 2022? The symfony binary provides it out of the box. Symfonys official docker image provides it out of the box

vvasiloi:
They should... but are you going to answer all the complains of those who don't? 😄 
Adding a parameter for that is safer and not that difficult.

loevgaard:
Yeah yeah, Victor, I agree :D

vvasiloi:
I created two new PRs, one for [static URLs](https://github.com/Sylius/Sylius/pull/13438) (ready for review) and one for [dynamic URLs](https://github.com/Sylius/Sylius/pull/13437) (WIP, but would appreciate some early feedback).

PR #11924, opened 2020-10-13
[Api] Change payment method after checkout

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Related tickets | part of #11250 
| License         | MIT


features/checkout/paying_for_order/changing_payment_method_after_order_confirmation.feature
```diff
@@ -32,3 +32,13 @@ Feature: Changing the method after order confirmation
         When I go to the change payment method page
         And I try to pay with "Offline" payment method
         Then I should see the thank you page
+
+    @api
+    Scenario: Changing the payment method for placed order
```
GSadee, line 37: I guess that you can adjust the existing scenarios for UI to be more business and the use them also for API context. What is more, it seems that in this api scenario you are missing some important behaviour from ui as paying by the changed payment method.

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Payment.xml
```diff
@@ -53,6 +53,19 @@
                     <attribute name="summary">Completes Payment</attribute>
                 </attribute>
             </itemOperation>
+
+            <itemOperation name="shop_change_payment_method">
```
GSadee, line 57: IMO you should use the same endpoint as in checkout, there is no need for adding the another one

GSadee (changes requested):

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -301,6 +302,7 @@ public function iProvideAdditionalNotesLike(string $notes): void
     /**
      * @Given I confirmed my order
      * @When I confirm my order
+     * @When I have confirmed order
```
GSadee, line 305:
```suggestion
     * @Given I have confirmed order
```

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -339,6 +341,7 @@ public function iConfirmMyOrder(): void
      * @Given /^the (?:visitor|customer) has proceeded ("[^"]+" shipping method)$/
      * @When /^the visitor try to proceed with ("[^"]+" shipping method) in the customer cart$/
      * @When I try to change shipping method to :shippingMethod
+     * @When I have proceeded selecting :shippingMethod shipping method
```
GSadee, line 344:
```suggestion
     * @Given I have proceeded selecting :shippingMethod shipping method
```

src/Sylius/Behat/Context/Api/Shop/OrderContext.php
```diff
@@ -257,6 +259,21 @@ public function theShouldHaveUnitPriceDiscountedFor(ProductInterface $product, i
...
     }
 
+    /**
+     * @Then I should have chosen :paymentMethod payment method
+     */
+    public function iShouldHaveChosenPaymentMethodForMyOrder(PaymentMethod $paymentMethod): void
+    {
+        $payment = $this->iriConverter->getItemFromIri(
+            $this->responseChecker->getValue($this->client->show($this->sharedStorage->get('cart_token')), 'payments')[0]
+        );
```
GSadee, line 269: I would be for getting payment method by an API call instead of translating payment to an item

src/Sylius/Behat/Context/Ui/Shop/Checkout/CheckoutThankYouContext.php
```diff
@@ -57,6 +63,16 @@ public function iProceedToTheRegistration(): void
         $this->thankYouPage->createAccount();
     }
 
+    /**
+     * @When I change payment method to :paymentMethodName
+     */
+    public function iChoosePaymentMethod(string $paymentMethodName): void
```
GSadee, line 69: IMO this method could be in `CheckoutOrderDetailsContext`

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -387,6 +390,7 @@ public function iCompleteTheShippingStepWithFirstShippingMethod(): void
      * @When /^the (?:customer|visitor) proceed with ("[^"]+" payment)$/
      * @Given /^the (?:customer|visitor) has proceeded ("[^"]+" payment)$/
      * @When I try to change payment method to :paymentMethod payment
+     * @When I change payment method to :paymentMethod
```
GSadee, line 393: As I wrote in one of the comments in an earlier review iteration, it seems that in the API context some of the important behaviour is missing. [In UI](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ShopBundle/Resources/config/routing/order.yml#L42), after changing the payment method, payment takes place, while in API only the payment is changed.

src/Sylius/Behat/Page/Shop/Order/ShowPage.php
```diff
@@ -59,6 +59,19 @@ public function getAmountOfItems(): int
         return count($paymentItems);
     }
 
+    public function getChosenPaymentMethod(): string
+    {
+        $paymentItems = $this->getDocument()->findAll('css', '[data-test-payment-item]');
```
GSadee, line 64:
```suggestion
        $paymentMethodItems = $this->getDocument()->findAll('css', '[data-test-payment-item]');
```

src/Sylius/Bundle/ApiBundle/CommandHandler/Checkout/ChoosePaymentMethodHandler.php
```diff
@@ -73,8 +68,29 @@ public function __invoke(ChoosePaymentMethod $choosePaymentMethod): OrderInterfa
         $payment = $this->paymentRepository->findOneByOrderId($choosePaymentMethod->paymentId, $cart->getId());
         Assert::notNull($payment, 'Can not find payment with given identifier.');
 
+        if (in_array(
+            $cart->getCheckoutState(),
+            [OrderCheckoutStates::STATE_SHIPPING_SELECTED, OrderCheckoutStates::STATE_SHIPPING_SKIPPED, OrderCheckoutStates::STATE_PAYMENT_SELECTED],
+            true)) {
```
GSadee, line 74:
```suggestion
        if (in_array(
            $cart->getCheckoutState(),
            [OrderCheckoutStates::STATE_SHIPPING_SELECTED, OrderCheckoutStates::STATE_SHIPPING_SKIPPED, OrderCheckoutStates::STATE_PAYMENT_SELECTED],
            true)
        ) {
```

src/Sylius/Bundle/ApiBundle/Doctrine/QueryItemExtension/OrderMethodsItemExtension.php
```diff
@@ -57,15 +57,27 @@ public function applyToItem(
...
 
     private function applyUserRulesToItem(
         ?UserInterface $user,
         QueryBuilder $queryBuilder,
         string $rootAlias,
-        string $httpRequestMethodType
+        string $httpRequestMethodType,
+        string $operationName
     ): void {
+        if ($user === null && $operationName === 'shop_select_payment_method') {
```
GSadee, line 70: Why only a user that is not logged in can change the payment method?

src/Sylius/Behat/Page/Shop/Order/ShowPage.php
```diff
@@ -59,6 +59,19 @@ public function getAmountOfItems(): int
         return count($paymentItems);
     }
 
+    public function getChosenPaymentMethod(): string
+    {
+        $paymentMethodItems = $this->getDocument()->findAll('css', '[data-test-payment-item]');
+
+        foreach ($paymentMethodItems as $method) {
+            if ($method->find('css', '[data-test-payment-method-select]')->hasAttribute('checked')) {
+                return $method->find('css', 'a')->getText();
```
lchrusciel, line 68: Shouldn't we provide a custom test data attribute for a link as well?

src/Sylius/Bundle/ApiBundle/Doctrine/QueryItemExtension/OrderMethodsItemExtension.php
```diff
@@ -57,15 +57,20 @@ public function applyToItem(
...
 
     private function applyUserRulesToItem(
         ?UserInterface $user,
         QueryBuilder $queryBuilder,
         string $rootAlias,
-        string $httpRequestMethodType
+        string $httpRequestMethodType,
+        string $operationName
     ): void {
+        if ($operationName === 'shop_select_payment_method') {
```
Tomanhez, line 70: Why you disabled this method in all cases? If it is disabled should be removed because this statement block it in all cases.

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -300,6 +300,7 @@ public function iProvideAdditionalNotesLike(string $notes): void
 
     /**
      * @Given I confirmed my order
+     * @Given I have confirmed order
```
lchrusciel, line 303: I would prefer to do order management based on commands if possible instead of API calls

src/Sylius/Bundle/ApiBundle/CommandHandler/Checkout/ChoosePaymentMethodHandler.php
```diff
@@ -73,8 +68,30 @@ public function __invoke(ChoosePaymentMethod $choosePaymentMethod): OrderInterfa
         $payment = $this->paymentRepository->findOneByOrderId($choosePaymentMethod->paymentId, $cart->getId());
         Assert::notNull($payment, 'Can not find payment with given identifier.');
 
+        if (in_array(
+            $cart->getCheckoutState(),
+            [OrderCheckoutStates::STATE_SHIPPING_SELECTED, OrderCheckoutStates::STATE_SHIPPING_SKIPPED, OrderCheckoutStates::STATE_PAYMENT_SELECTED],
+            true)
+        ) {
```
lchrusciel, line 75: This check seems wrong. Instead of extending the logic of `choosing payment method` we should create a separate handler, which will handle changing payment methods after checkout is completed.

src/Sylius/Bundle/ApiBundle/Doctrine/QueryItemExtension/OrderMethodsItemExtension.php
AdamKasp: sorry, but I am not sure what do you mean. my intention was return order always if `shop_select_payment_method` was called. What do you mean by "disabled method in all cases"?

GSadee: But after adding this `if`, the order will be returned even if the order doesn't belong to the current user

src/Sylius/Behat/Page/Shop/Order/ShowPage.php
AdamKasp: here i am iterating over the shipping method and IMO my approach is more readable, this link is also a label form shipping method.

features/checkout/paying_for_order/changing_offline_payment_method_after_order_confirmation.feature
```diff
@@ -0,0 +1,44 @@
...
+        And the store ships everywhere for free
+
+    @ui
+    Scenario: Retrying the payment with different offline payment
+        Given I added product "PHP T-Shirt" to the cart
+        When I complete addressing step with email "john@example.com" and "United States" based billing address
+        And I have proceeded selecting "Free" shipping method
+        And I have proceeded selecting "Cash on delivery" payment method
+        And I have confirmed order
+        And I change payment method to "Offline" after checkout
```
GSadee, line 21: This scenarios looks the same as the one with `@api` tag in this feature, and as the real thing that is happening in UI is changing the payment method and then paying I would be for changing this step to sth like:
`I retry the payment with "Offline" payment method`
or go back to what was before:
`I try to pay with with "Offline" payment method`

src/Sylius/Behat/Context/Ui/Shop/Checkout/CheckoutOrderDetailsContext.php
```diff
@@ -23,40 +24,54 @@ final class CheckoutOrderDetailsContext implements Context
...
+    public function iChangePaymentMethodTo(string $paymentMethodName): void
     {
         $this->orderDetails->choosePaymentMethod($paymentMethodName);
         $this->orderDetails->pay();
     }
 
+    /**
+     * @When I change payment method to :paymentMethodName after checkout
+     */
+    public function iChoosePaymentMethod(string $paymentMethodName): void
```
GSadee, line 56:
```suggestion
    public function iChangePaymentMethodAfterCheckout(string $paymentMethodName): void
```

src/Sylius/Behat/Context/Ui/Shop/Checkout/CheckoutThankYouContext.php
GSadee: Why did you have this method in two different UI contexts?

src/Sylius/Bundle/ApiBundle/Doctrine/QueryItemExtension/OrderMethodsItemExtension.php
```diff
@@ -57,15 +57,36 @@ public function applyToItem(
...
 
     private function applyUserRulesToItem(
         ?UserInterface $user,
         QueryBuilder $queryBuilder,
         string $rootAlias,
-        string $httpRequestMethodType
+        string $httpRequestMethodType,
+        string $operationName
     ): void {
+        if ($user instanceof ShopUserInterface && $operationName === 'shop_select_payment_method') {
```
GSadee, line 70: I suggest refactoring the existing code to apply on query builder
```
->andWhere(sprintf('%s.state = :state', $rootAlias))
->setParameter('state', OrderInterface::STATE_CART)
```
only if there is an endpoint for checkout.
IMO it could simplify the logic 😃

src/Sylius/Behat/Context/Ui/Shop/Checkout/CheckoutOrderDetailsContext.php
```diff
@@ -23,40 +24,56 @@ final class CheckoutOrderDetailsContext implements Context
...
-    public function iChangePaymentMethodTo($paymentMethodName)
+    public function iChangePaymentMethodTo(string $paymentMethodName): void
     {
         $this->orderDetails->choosePaymentMethod($paymentMethodName);
         $this->orderDetails->pay();
     }
 
+    /**
+     * @When I change payment method to :paymentMethodName after checkout
+     * @When I change payment method to :paymentMethodName
```
GSadee, line 55: I guess that these two steps are not necessary anymore
```suggestion
```

GSadee (approved):

Tomanhez (approved):

GSadee:
Thank you, Adam! :1st_place_medal:

PR #11931, opened 2020-10-14
[API] Refactor SubresourceIdAwareCommandDataTransformer into two more precise implementation

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bugfix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | 
| License         | MIT

This PR aims to define corresponding fields of request in service instead of command, as it seems to break the encapsulation of data. The command is aware of some external object, while service without proper command is incomplete in its business logic.


jakubtobiasz:
I'm wondering about closing this PR without merging as:
- We've been living with that through several Sylius versions
- We're slowly forced to add a support for APIP3 (`1.13`) and dropping APIP2 at all (`2.0`), where there are no transformers
- It's BC (I know the API is experimental, but the two previous points make this one considerable

WDYT? @Sylius/core-team

lchrusciel:
Removal of transformers convinved me 👍

PR #11932, opened 2020-10-14
[Minor][API] Add missing messenger handler interfaces

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

`\Symfony\Component\Messenger\Handler\MessageHandlerInterface` is just a marker interface, so Symfony can autowire these classes. We have it implemented in all handlers despite the ones from the Checkout folder. Maybe it doesn't make sense to add them, but I would prefer to keep it consistent and in case just remove all these interfaces, or add missing as proposed in this PR.


Zales0123 (approved):

Zales0123:

PR #11937, opened 2020-10-15
[Cart] Change logic for maintaining cart after logging in

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no?
| Deprecations?   | 
| Related tickets | 
| License         | MIT

Now, when the customer didn't have or had an empty cart and their had a cart in session, the one from session will be assigned to them.

### DIFF
--- a/src/Sylius/Behat/Client/ApiPlatformClient.php
+++ b/src/Sylius/Behat/Client/ApiPlatformClient.php
@@ -221,8 +221,8 @@ final class ApiPlatformClient implements ApiClientInterface
         return $this->getLastResponse();
     }
 
-    private function getToken(): string
+    private function getToken(): ?string
     {
-        return $this->sharedStorage->has('token') ? $this->sharedStorage->get('token') : '';
+        return $this->sharedStorage->has('token') ? $this->sharedStorage->get('token') : null;
     }
 }
--- a/src/Sylius/Behat/Context/Api/Shop/CartContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/CartContext.php
@@ -75,6 +75,8 @@ final class CartContext implements Context
     public function iClearMyCart(string $tokenValue): void
     {
         $this->cartsClient->delete($tokenValue);
+
+        $this->sharedStorage->set('cart_token', null);
     }
 
     /**
--- a/src/Sylius/Behat/Context/Api/Shop/CartContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/CartContext.php
@@ -80,7 +80,7 @@ final class CartContext implements Context
     }
 
     /**
-     * @When /^I see the summary of my (cart)$/
+     * @When /^I see the summary of my ((?:|previous )cart)$/
      * @When /^the visitor try to see the summary of ((?:visitor|customer)'s cart)$/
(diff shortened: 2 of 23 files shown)


src/Sylius/Bundle/CoreBundle/Context/CustomerAndChannelBasedCartContext.php
```diff
@@ -55,7 +55,7 @@ public function getCart(): OrderInterface
             throw new CartNotFoundException('Sylius was not able to find the cart, as there is no logged in user.');
         }
 
-        $cart = $this->orderRepository->findLatestCartByChannelAndCustomer($channel, $customer);
+        $cart = $this->orderRepository->findLatestNotEmptyCartByChannelAndCustomer($channel, $customer);
```
GSadee, line 58: To be honest, I'm not 100% sure if this change will not be a little BC Break so I would like to mark this place for review 😃

src/Sylius/Behat/Client/ApiPlatformClient.php
```diff
@@ -203,7 +203,8 @@ public function getLastResponse(): Response
         return $this->client->getResponse();
     }
 
-    private function request(RequestInterface $request): Response
+    private function
+    request(RequestInterface $request): Response
```
AdamKasp, line 207:
```suggestion
    private function request(RequestInterface $request): Response
```

src/Sylius/Behat/Context/Api/Shop/CartContext.php
```diff
@@ -362,6 +380,10 @@ private function putProductToCart(ProductInterface $product, string $tokenValue,
 
     private function putProductVariantToCart(ProductVariantInterface $productVariant, string $tokenValue, int $quantity = 1): void
     {
+        if ($tokenValue === null) {
```
lchrusciel, line 381: cannot be true

src/Sylius/Bundle/CoreBundle/Context/CustomerAndChannelBasedCartContext.php
lchrusciel: I don't have a strong opinion, but we could leave the previous implementation and check object emptiness. On the other hand, we are saving object hydration, and it seems to be a cleaner solution. So 👍 from my side

lchrusciel (approved):

src/Sylius/Behat/Context/Api/Shop/CartContext.php
```diff
@@ -357,11 +377,16 @@ private function putProductToCart(ProductInterface $product, string $tokenValue,
             'quantity' => $quantity,
         ]);
 
-        $this->cartsClient->executeCustomRequest($request);
+        $res = $this->cartsClient->executeCustomRequest($request);
+        $res = $res;
```
arti0090, line 381: `
        $res = $res;
`
Is it needed? :D

arti0090: And is $res needed to be honest? :)

AdamKasp (approved):

AdamKasp:
Thanks, Grzegorz! :tada:

PR #11939, opened 2020-10-16
[API]Seeing shipping cost of order

arti0090:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT


src/Sylius/Behat/Context/Api/Shop/CartContext.php
```diff
@@ -117,6 +117,7 @@ public function iAddOfThemToMyCart(int $quantity, ProductInterface $product, str
     }
 
     /**
+     * @Given /^I have ("[^"]+" variant of this product) in the (cart)$/
```
lchrusciel, line 120: I see, that it is not always true, but IMHO steps starting with `given` should be part of the setup context. Preferably, they shouldn't make any requests at all. We already made this mistake in UI contexts, but it is harder there to maintain the same separation.

src/Sylius/Component/Core/Formatter/StringInflector.php
```diff
@@ -29,7 +30,7 @@ public static function nameToCode(string $value): string
 
     public static function nameToSlug(string $value): string
     {
-        return str_replace(['_'], '-', self::nameToLowercaseCode($value));
+        return str_replace(['_'], '-', self::nameToLowercaseCode(Transliterator::transliterate($value)));
```
AdamKasp, line 33: a change like this does not be a BC break?

src/Sylius/Behat/Context/Api/Shop/CartContext.php
```diff
@@ -347,6 +348,19 @@ public function myCartShouldHaveItems(int $quantity, ProductInterface $product):
...
+     * @Then /^my cart shipping total should be "(?:€|£|\$)([^"]+)"$/
+     * @Then I should not see shipping total for my cart
+     */
+    public function myCartShippingFeeShouldBe(string $shippingTotal = '0'): void
+    {
+        $response = $this->cartsClient->getLastResponse();
+
+        Assert::same(
+            $this->responseChecker->getValue($response, 'shippingTotal'),
+            (int) str_replace('.','',$shippingTotal));
```
AdamKasp, line 361: you propably can use in suit `sylius.behat.context.transform.lexical` which transform string with a price to `int`

src/Sylius/Behat/Context/Api/Shop/CartContext.php
```diff
@@ -392,15 +406,15 @@ private function getProductForItem(array $item, bool $shopSection = true): Respo
         }
 
         $variantIri = $shopSection ? $this->adminToShopIriConverter->convert($item['variant']) : $item['variant'];
-        $response = $this->cartsClient->executeCustomRequest(Request::custom($variantIri, HttpRequest::METHOD_GET));
+        $response = $this->cartsClient->executeCustomRequest(Request::custom(urldecode($variantIri), HttpRequest::METHOD_GET));
```
AdamKasp, line 409: can you use here function `showByIri`?

src/Sylius/Component/Core/Formatter/StringInflector.php
arti0090: @GSadee @lchrusciel would you have a look at it? I did it because we are not really resolving slug sometimes. In the case I have found, I had special characters in my string ( '!' in my PR ) which was not erased from it as it should be. Returned value should be from "string!_string!" to "string-string" but it returns "string!-string!". When data provider is looking for it, it is not available to find it with this slug.

GSadee (approved):

lchrusciel (approved):

lchrusciel:
Thank you, @arti0090! :tada:

PR #11946, opened 2020-10-19
Changed behavior of string inflector

arti0090:
| Q               | A
| --------------- | -----
| Branch?         | 1.8 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | possible
| Deprecations?   | no
| License         | MIT

When we change strings with special characters ( f.e. '!' ) String inflector doesn't change it as should. Now it will delete those characters.

### DIFF
--- a/src/Sylius/Component/Core/Formatter/StringInflector.php
+++ b/src/Sylius/Component/Core/Formatter/StringInflector.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sylius\Component\Core\Formatter;
 
+use Behat\Transliterator\Transliterator;
 use function Symfony\Component\String\u;
 
 final class StringInflector
@@ -24,7 +25,7 @@ final class StringInflector
 
     public static function nameToSlug(string $value): string
     {
-        return str_replace(['_'], '-', self::nameToLowercaseCode($value));
+        return str_replace(['_'], '-', self::nameToLowercaseCode(Transliterator::transliterate($value)));
     }
 
     public static function nameToLowercaseCode(string $value): string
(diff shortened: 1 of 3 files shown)


AdamKasp (approved):

GSadee (approved):

src/Sylius/Component/Core/Formatter/StringInflector.php
```diff
@@ -13,6 +13,7 @@
 
 namespace Sylius\Component\Core\Formatter;
 
+use Behat\Transliterator\Transliterator;
```
GSadee, line 16: I'm not sure if it is ok to use service from test dependency (Behat) in the Core component, wdyt @lchrusciel ?

arti0090: So in this case should we use same Transliterator in src/Sylius/Component/Product/Generator/SlugGenerator.php class? 🤔

lchrusciel: We already have this library in required in composer, so it is ok.
image

src/Sylius/Component/Core/Test/Tests/StringInflectorTest.php
```diff
@@ -0,0 +1,68 @@
...
+use Sylius\Component\Core\Formatter\StringInflector;
+
+final class StringInflectorTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function nameToCode(): void
+    {
+        self::assertEquals('Test_value', StringInflector::nameToCode('Test value'));
```
lchrusciel, line 26: Can we cover more cases? We are missing examples with special characters.

lchrusciel (approved):

GSadee:
Thank you, @arti0090! :1st_place_medal:

PR #11971, opened 2020-10-28
[API] Seeing promotions total in cart

arti0090:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Api/Shop/CartContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/CartContext.php
@@ -186,20 +186,6 @@ final class CartContext implements Context
     }
 
     /**
-     * @Then /^my discount should be ("[^"]+")$/
-     * @Then there should be no discount
-     */
-    public function myDiscountShouldBe(int $discount = 0): void
-    {
-        $discountTotal = $this->responseChecker->getValue(
-            $this->cartsClient->show($this->sharedStorage->get('cart_token')),
-            'orderPromotionTotal'
-        );
-
-        Assert::same($discount, (int) $discountTotal);
-    }
-
-    /**
      * @Then /^my (cart) should be empty$/
      */
     public function myCartShouldBeEmpty(string $tokenValue): void
--- a/src/Sylius/Behat/Context/Api/Shop/CartContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/CartContext.php
@@ -51,9 +51,6 @@ final class CartContext implements Context
     /** @var ProductVariantResolverInterface */
     private $productVariantResolver;
 
-    /** @var Response */
-    private $response;
-
     public function __construct(
         ApiClientInterface $cartsClient,
         ApiClientInterface $productsClient,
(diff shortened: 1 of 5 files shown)


src/Sylius/Behat/Resources/config/suites/api/promotion/receiving_discount.yml
```diff
@@ -0,0 +1,30 @@
...
+                - sylius.behat.context.setup.product
+                - sylius.behat.context.setup.product_taxon
+                - sylius.behat.context.setup.promotion
+                - sylius.behat.context.setup.shop_security
+                - sylius.behat.context.setup.shipping
+                - sylius.behat.context.setup.taxonomy
+                    
+                - sylius.behat.context.api.shop.cart
+
+
```
AdamKasp, line 28:
```suggestion
```

src/Sylius/Behat/Context/Api/Shop/CartContext.php
```diff
@@ -173,15 +176,27 @@ public function myCartShouldBeCleared(): void
 
     /**
      * @Then /^my (cart)'s total should be ("[^"]+")$/
+     * @Then /^my (cart) total should be ("[^"]+")$/
      */
     public function myCartSTotalShouldBe(string $tokenValue, int $total): void
     {
-        $response = $this->cartsClient->show($tokenValue);
+        $this->response = $this->cartsClient->show($tokenValue);
```
AdamKasp, line 183: IMO this change is unnecessary, you should leave here previous code, and in function `myDiscountShouldBe` use `cartsClient->getLastResponse` or get cart token from `sharedStorage` and make show call again.

AdamKasp (changes requested):

AdamKasp (approved):

GSadee (approved):

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -682,11 +682,21 @@ public function iShouldHaveProductsInTheCart(int $quantity, string $productName)
...
 
     /**
+     * @Then /^my discount should be ("[^"]+")$/
      * @Then there should be no discount
      */
-    public function thereShouldBeNoDiscount(): void
+    public function myDiscountShouldBe(int $discount = 0): void
     {
-        Assert::same($this->responseChecker->getValue($this->client->getResponse(), 'orderPromotionTotal'), 0);
+        if ($this->sharedStorage->has('cart_token')) {
```
lchrusciel, line 690: Does this if makes sense? I would assume that cart_token is always set (as we always have cart declared).

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -682,11 +682,21 @@ public function iShouldHaveProductsInTheCart(int $quantity, string $productName)
...
+            $discountTotal = $this->responseChecker->getValue(
+                $this->ordersClient->show($this->sharedStorage->get('cart_token')),
+                'orderPromotionTotal'
+            );
+
+            Assert::same($discount, (int) $discountTotal);
+        } else {
```
lchrusciel, line 697: Instead of else, can we make early return?

GSadee (approved):

GSadee:

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
arti0090: Yes, because it depends on context where it was started. If steps are all in checkout then they are using `$this->client->getResponse()` but some steps from cartContext are finished here and those steps are using `sharedStorage`.

PR #11981, opened 2020-10-30
Adjust services to symfony/serializer:^4.4

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

It's required for our ApiPlatform denormalizer implementations. With `symfony/serializer` in version **4.4**, the following error occurs:

Zrzut ekranu 2020-10-30 o 16 47 20


composer.json
```diff
@@ -94,6 +94,7 @@
         "symfony/security-csrf": "^4.4",
         "symfony/security-guard": "^4.4",
         "symfony/security-http": "^4.4",
+        "symfony/serializer": "^5.0",
```
lchrusciel, line 97: I would rather go with "symfony/serializer" 4.4, as this is the official supported version for Symfony packages. We do not have support for 5.0 for any library, so we can migrate all these libraries together later.

Zales0123: 👍 but then we need to change the denormalizers implementation 💃

lchrusciel (approved):

lchrusciel:

PR #11992, opened 2020-11-02
[Api] Renaming CommandAwareDataTransformerInterface to better express…

lchrusciel:
… its purpose

| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Proposal of renaming, to better express intention behind the interface. `CommandAwareDataTransformerInterface` is placed on commands, that will be passed to our data transformers, so we can put additional data into them, like order token, or subresource id. There is nothing to be aware of because this interface is placed on command in the first place. 

In addition, I've added Command suffix to command interfaces.

If we would like to merge it, I should probably with UPGRADE note for this change.

### DIFF
--- a/UPGRADE-1.9.md
+++ b/UPGRADE-1.9.md
@@ -147,7 +147,7 @@
 
 1. Unified API registration path in shop has been changed from `/new-api/shop/register` to `/new-api/shop/customers/`. 
  
-1. Identifier needed to retrieve a product in shop API endpoint (`/new-api/shop/products/{id}`) has been changed from `slug` to `code`. 
+1. Identifier needed to retrieve a product in shop API endpoint (`/new-api/shop/products/{id}`) has been changed from `slug` to `code`.
 
 1. Replace and add new keys in `config/packages/dev/jms_serializer.yaml`:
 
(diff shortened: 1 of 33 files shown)


Zales0123 (commented):
It would be good to have a note in the UPGRADE file 👌

arti0090 (approved):

Rafikooo:
Hello @lchrusciel,
We encourage you to rebase and update this PR as it is still valid. It would be nice if you would also update the name of the PR, as it differs from the actual content inside. 🚀

jakubtobiasz:
Similarly to #11931

> I'm wondering about closing this PR without merging as:
>
> We've been living with that through several Sylius versions
> We're slowly forced to add a support for APIP3 (1.13) and dropping APIP2 at all (2.0), where > there are no transformers
> It's BC (I know the API is experimental, but the two previous points make this one considerable
> WDYT? https://github.com/orgs/Sylius/teams/core-team

lchrusciel:
Removal of transformers convinved me 👍

PR #11993, opened 2020-11-02
[API] Explicitly define bus to all handlers

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no (kind of)
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Symfony messenger component will register handlers into all busses if we won't specify it. I would say, that additional busses are common enough, that we should not register our handlers in different busses by default. 

Ref. https://symfony.com/doc/current/messenger/multiple_buses.html#restrict-handlers-per-bus


oallain (approved):
It fix an issue from my side, thanks !

oallain (approved):

Zales0123 (approved):

Zales0123:
Thanks, Łukasz! :tada:

PR #12010, opened 2020-11-05
[Hot-fix] Allow to use svg images without filtration

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

The problem is, even though we're able to upload svg files as images, they cannot be filtered properly by LiipImagineBundle filters. Is it the perfect solution? For sure no - but allows handling as very common case, which, definitely, using an svg file is. Probably the best option would be to fix it in LiipImagineBundle itself.

I would like to see the opinion about the implementation. Are we able to define svg image path in other way? We could create an additional route for rendering not filtered images, but it's a lot of additional code and configuration. Maybe the decoration of the default extension would be better?


lchrusciel (approved):

src/Sylius/Bundle/AdminBundle/Resources/views/Product/_mainImage.html.twig
```diff
@@ -5,5 +5,8 @@
 {% else %}
     {% set path = '//placehold.it/50x50' %}
 {% endif %}
-
-<img src="{{ path }}" alt="" class="ui bordered image" />
+{% if product.images.first.path|split('.')|last is same as 'svg' %}
+    <img src="{{ path }}" alt="" class="ui bordered image" style="max-height: 100px; max-width: 150px" />
+{% else %}
+    <img src="{{ path }}" alt="" class="ui bordered image" />
+{% endif %}
```
GSadee, line 12: Instead of this `if`, you could maybe change css classes for grids and other places with uploaded images

src/Sylius/Bundle/AdminBundle/Resources/views/Product/_mainImage.html.twig
```diff
@@ -5,5 +5,8 @@
 {% else %}
     {% set path = '//placehold.it/50x50' %}
 {% endif %}
-
-<img src="{{ path }}" alt="" class="ui bordered image" />
+{% if product.images.first.path|split('.')|last is same as 'svg' %}
```
Zales0123, line 8: It would fail if there are no images in product :/

src/Sylius/Bundle/AdminBundle/Resources/views/Product/_mainImage.html.twig
AdamKasp: I checked all places with images and if i am right this grid is the only one without proper sizing provide by semanticUI

Tomanhez (approved):

src/Sylius/Bundle/CoreBundle/Twig/FilterExtension.php
```diff
@@ -0,0 +1,35 @@
...
+        if (!$this->canImageBeFiltered($path)) {
+            return $this->imagesPath.$path;
+        }
+
+        return parent::filter($path, $filter, $config, $resolver);
+    }
+
+    private function canImageBeFiltered(string $path): bool
+    {
+        return substr($path, -3) !== 'svg';
```
SirDomin, line 33: I guess we should check for last 4 characters, this method can return true for any file that ends with svg, eg. file.asvg.
We should check for '.svg' as it determines extension type🥦

GSadee (approved):

GSadee:

Zales0123:
And Adam! And Dominik! 😄 💃

PR #12018, opened 2020-11-07
[Order] Order cancelation fix

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | kind of(as I reduced the scope of exceptions, that are caught)
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Catching all exceptions may lead to flushing partially changed data. If after transition some service will change some part of data and after that, an exception will occur, we would silence this exception and proceed further without compensating partially changed data.

However, the release with these changes was published on Thursday, so the impact of it is negligible  (comparing to the problems it can create).


Zales0123 (approved):

Zales0123:

PR #12050, opened 2020-11-17
Refactor mailing

SirDomin:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | not sure
| Related tickets | refactoring the way we handle emails implemented in https://github.com/Sylius/Sylius/pull/11754
| License         | MIT

PR #12054, opened 2020-11-18
[ADR] Sending emails via api

SirDomin:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #12050
| License         | MIT

### DIFF
--- a/adr/2020_11_18_sending_emails_via_api.md
+++ b/adr/2020_11_18_sending_emails_via_api.md
@@ -1,0 +1,53 @@
+# API - Sending emails
+
+* Status: proposed
+* Date: 2020-11-18
+
+## Context and Problem Statement
+
+To achieve 100% api coverage, we need to handle emails by api.
+
+## Considered Options
+
+### Using events
+
+Using events allow us to trigger any handler once operation is completed. 
+
+* Sylius\ApiBundle\Event\OrderCompleted
+
+Once this event is triggered we handle it with ```src/Sylius/Bundle/ApiBundle/EventHandler/OrderCompletedHandler.php```
+
+Then command ```src/Sylius/Bundle/ApiBundle/Command/SendOrderConfirmation.php ``` is dispatched, which triggers command handler (```src/Sylius/Bundle/ApiBundle/CommandHandler/SendOrderConfirmationHandler.php```)
+ 
+SendOrderConfirmationHandler is sending email based on its destination (in this case OrderConfirmation)
+
+```
+$this->emailSender->send(
+    Emails::ORDER_CONFIRMATION_RESENT,
+    [$order->getCustomer()->getEmail()],
+    [
+        'order' => $order,
+        'channel' => $order->getChannel(),
+        'localeCode' => $order->getLocaleCode(),
+    ]
+);
+```
+
+* Good, because its sending email based on triggered event, once understood it's simple to add new handlers.
+* Bad, because adds a lot of abstraction, can be hard to understand.
+
+### Direct email call
+
+in ```src/Sylius/Bundle/ApiBundle/CommandHandler/Checkout/CompleteOrderHandler.php```
+we create direct call to email
+
+```php
+$this->emailManager->sendConfirmationEmail($cart);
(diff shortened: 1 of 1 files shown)


adr/2020_11_18_sending_emails_via_api.md
```diff
@@ -0,0 +1,43 @@
...
+
+# API - Sending emails
+
+In power from version: master (?)
+
+Related PRs: #11754, #12050
+
+### Problem to solve
+
+How to send email using api
```
AdamKasp, line 11:
```suggestion
How to send email using api.
```

adr/2020_11_18_sending_emails_via_api.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Context
+
+With more and more api coverage we had to decide how we should handle sending emails.
```
AdamKasp, line 15: some operations in Sylius as the final step send emails to the customer, for full API coverage we need to handle it also by API

adr/2020_11_18_sending_emails_via_api.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Possible solutions
```
AdamKasp, line 17: You described only one solution, so this title is misleading.

adr/2020_11_18_sending_emails_via_api.md
```diff
@@ -0,0 +1,43 @@
...
+        'order' => $order,
+        'channel' => $order->getChannel(),
+        'localeCode' => $order->getLocaleCode(),
+    ]
+);
+```
+
+### Decision and reasoning
+
+This adds alot of new abstraction, but using events we can trigger any email handler without digging into code. Only thing we need to do is handle triggered event (or create one if none is dispatched), send it to emailSender and we are good to go.
```
AdamKasp, line 43:
```suggestion
This adds a lot of new abstraction, but using events we can trigger any email handler without digging into code. Only thing we need to do is handle triggered event (or create one if none is dispatched), send it to emailSender and we are good to go.
```

adr/2020_11_18_sending_emails_via_api.md
AdamKasp: or sth like this :D

adr/2020_11_18_sending_emails_via_api.md
```diff
@@ -0,0 +1,53 @@
...
+    [$order->getCustomer()->getEmail()],
+    [
+        'order' => $order,
+        'channel' => $order->getChannel(),
+        'localeCode' => $order->getLocaleCode(),
+    ]
+);
+```
+
+* Good, because its sending email based on triggered event, once understood it's simple to add new handlers.
```
lchrusciel, line 36:
```suggestion
* Good, because it's easy to integrate other logic after the initial command will succeed.
* Good, because it's easy to trigger the same operation several times.
* Good, because it's easier to handle mailing server downs gracefully
```

adr/2020_11_18_sending_emails_via_api.md
```diff
@@ -0,0 +1,53 @@
+# API - Sending emails
+
+* Status: proposed
+* Date: 2020-11-18
+
+## Context and Problem Statement
+
+To achieve 100% api coverage, we need to handle emails by api.
```
GSadee, line 8:
```suggestion
To achieve 100% API coverage, we need to handle emails by API.
```

GSadee (approved):

GSadee:
Thank you, @SirDomin! :tada:

adr/2020_11_18_sending_emails_via_api.md
```diff
@@ -0,0 +1,57 @@
+# API - Sending emails
+
+* Status: proposed
```
oallain, line 3: @SirDomin @GSadee 

Shouldn't the status be "accepted" now?

PR #12055, opened 2020-11-18
Change route prefixes in security.yaml

jacquesbh:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | no, not really
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

Since we had the same change in the `access_control`, I've done it in the firewalls as well.

😘

_I'm not sure about the UPGRADE-1.8.md actually. Changing the version number is maybe not the best thing to do._

### DIFF
--- a/UPGRADE-1.8.md
+++ b/UPGRADE-1.8.md
@@ -1,10 +1,19 @@
-# UPGRADE FROM `v1.8.4` TO `v1.8.5`
+# UPGRADE FROM `v1.8.4` TO `v1.8.6`
 
 1. Change configuration of new ApiBundle in your `config/packages/security.yaml` file:
 
     ```diff
-        security:
-            access_control:
+         security:
+             firewalls:
+                 new_api_admin_user:
+                     json_login:
+    -                    check_path: "%sylius.security.new_api_route%/admin/authentication-token"
+    +                    check_path: "%sylius.security.new_api_admin_route%/authentication-token"
+                 new_api_shop_user:
+                     json_login:
+    -                    check_path: "%sylius.security.new_api_route%/shop/authentication-token"
+    +                    check_path: "%sylius.security.new_api_shop_route%/authentication-token"
+             access_control:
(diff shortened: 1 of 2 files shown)


lchrusciel (approved):

UPGRADE-1.8.md
```diff
@@ -1,10 +1,19 @@
...
+             firewalls:
+                 new_api_admin_user:
+                     json_login:
+    -                    check_path: "%sylius.security.new_api_route%/admin/authentication-token"
+    +                    check_path: "%sylius.security.new_api_admin_route%/authentication-token"
+                 new_api_shop_user:
+                     json_login:
+    -                    check_path: "%sylius.security.new_api_route%/shop/authentication-token"
+    +                    check_path: "%sylius.security.new_api_shop_route%/authentication-token"
+             access_control:
```
Zales0123, line 16: Maybe it would be better to introduce it in `1.8.5` -> `1.8.6` upgrade note? But I don't know isn't it too finicky 💃

Zales0123 (approved):

UPGRADE-1.8.md
jacquesbh: Yeah, I thought about it… and had the same hesitation 😅

jacquesbh: This should be in 1.8.5 actually… it's a mistake not to have this change in the previous PR :p.

That's why I hesitate. Let me know about it and I'll do the change if needed. It's not complicated :p

jacquesbh: @Zales0123
Do I change? Or not? What do you want me to do?

Zales0123:
Thanks, Jacques! :1st_place_medal:

UPGRADE-1.8.md
Zales0123: Sorry for missing this comment, let it be as it is 😄

PR #12057, opened 2020-11-18
Switch to Psalm 4

pamil:


composer.json
```diff
@@ -146,7 +146,7 @@
         "symfony/dotenv": "^4.4",
         "symfony/flex": "^1.7",
         "symfony/web-profiler-bundle": "^4.4",
-        "vimeo/psalm": "^3.11.6"
+        "vimeo/psalm": "4.1.1"
```
pamil, line 149: Locking Psalm at a specific version so that we won't get unexpected build failures.

lchrusciel (approved):

lchrusciel:
Thanks, Kamil! :tada:

PR #12076, opened 2020-11-21
[API][Behat] fixed steps for quantity-change api test

arti0090:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| License         | MIT

Step was badly implemented, it was adding the items, not changing quantity. It is fixed here.

### DIFF
--- a/src/Sylius/Behat/Context/Api/Shop/CartContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/CartContext.php
@@ -325,7 +325,7 @@ final class CartContext implements Context
      */
     public function iShouldBeInformedThatCartItemsAreNoLongerAvailable(): void
     {
-        $response = $this->cartsClient->getLastResponse();
+       $response = $this->sharedStorage->get('response') ?? $this->cartsClient->getLastResponse();
 
         Assert::same($response->getStatusCode(), 404);
 
@@ -502,6 +502,8 @@ final class CartContext implements Context
         $request->updateContent(['orderItemId' => $orderItemId, 'newQuantity' => $quantity]);
 
         $this->cartsClient->executeCustomRequest($request);
+
+        $this->sharedStorage->set('response', $this->cartsClient->getLastResponse());
     }
 
     private function hasItemWithNameAndQuantity(Response $response, string $productName, int $quantity): bool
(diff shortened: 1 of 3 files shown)


src/Sylius/Behat/Context/Api/Shop/CartContext.php
```diff
@@ -497,10 +506,10 @@ private function changeQuantityOfOrderItem(string $orderItemId, int $quantity, s
 
     private function hasItemWithNameAndQuantity(Response $response, string $productName, int $quantity): bool
     {
-        $items = json_decode($response->getContent(), true)['hydra:member'];
+        $items = $this->responseChecker->getCollection($response);
 
         foreach ($items as $item) {
-            if ($item['productName'] === $productName && $item['quantity'] === $quantity) {
+            if ($item['quantity'] === $quantity && $item['productName'] === $productName) {
```
GSadee, line 512: Why did you change the order?

src/Sylius/Behat/Context/Api/Shop/CartContext.php
```diff
@@ -319,6 +320,18 @@ public function iShouldSeeWithQuantityInMyCart(string $productName, int $quantit
         $this->checkProductQuantity($this->cartsClient->getLastResponse(), $productName, $quantity);
     }
 
+    /**
+     * @Then I should be informed that cart items are no longer available
+     */
+    public function iShouldBeInformedThatCartItemsAreNoLongerAvailable(): void
+    {
+        $response = $this->cartsClient->getLastResponse() ?? $this->sharedStorage->get('response');
```
GSadee, line 328: Why are you getting a response from shared storage here? If I see it correctly, you are setting this response only after complete step, so it is not associated with the scenarios, that you are changing in this PR

arti0090: some steps are called on CheckoutContext, and there it is using AbstractBrowser class to make requests. I used there a sharedStorage to send objects between contexts. It is needed to make those steps work for now. But soon I will work on refactor of both of these contexts, so it won't be such a confussion.

src/Sylius/Behat/Context/Api/Shop/CartContext.php
arti0090: Comparing integer is faster 😄  but I can revert this change if it creates confusion 😉

GSadee: But there is a greater chance that quantity of different products could be the same and there will be more checks in the result

GSadee (approved):

GSadee:

PR #12088, opened 2020-11-25
[Behat] Remove conflict with "lcobucci/jwt" and error reporting for E_USER_DEPRECATED

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

After conflict with `lcobucci/jwt`

### DIFF
--- a/src/Sylius/Behat/Service/ApiSecurityService.php
+++ b/src/Sylius/Behat/Service/ApiSecurityService.php
@@ -39,6 +39,8 @@ final class ApiSecurityService implements SecurityServiceInterface
 
     public function logIn(UserInterface $user): void
     {
+        error_reporting(error_reporting() & ~E_USER_DEPRECATED);
+
         $this->client->request(
             'POST',
             sprintf('/new-api/%s/authentication-token', $this->loginEndpoint),
--- a/CONFLICTS.md
+++ b/CONFLICTS.md
@@ -9,15 +9,6 @@ refereneces related issues.
    
    References: https://github.com/doctrine/inflector/issues/147
    
- - `lcobucci/jwt:^3.4`:
- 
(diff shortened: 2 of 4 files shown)


src/Sylius/Behat/Service/ApiSecurityService.php
```diff
@@ -39,6 +39,13 @@ public function __construct(AbstractBrowser $client, SharedStorageInterface $sha
 
     public function logIn(UserInterface $user): void
     {
+        /**
+         * Hot-fix: Added since conflict with 'lcobucci/jwt' in version 3.4
+         * Problem occurs when in calls is using 'trigger_error' with error level E_USER_DEPRECATED
+         * Then run code is interrupt and error is throwing in up
+         */
+        error_reporting(~E_USER_DEPRECATED);
```
pamil, line 47: This will change the global state of the application - I'd rather see it inside some bootstrap file just for the test environment.

pamil (commented):
Could you rebase it to 1.8 and remove the conflict with lcobucci/jwt in composer.json there?

src/Sylius/Behat/Service/ApiSecurityService.php
Tomanhez: I can add config to our behat.yaml : 
`calls:`
      `error_reporting: 16383`
Then error doesn't crash our test but all error stack trace is visible: 
Screenshot 2020-11-26 at 20 24 29

This problem is known in the past:
https://stackoverflow.com/questions/31050716/how-to-run-behat-tests-when-there-are-errors-of-level-e-user-deprecated

pamil:
Fixed by #12101.

PR #12089, opened 2020-11-25
[CI] Switch packages build from Travis to GitHub Actions

pamil:
This will create two jobs for every package that we have, making use of 20 parallel jobs GitHub Actions provides.


lchrusciel (approved):

.github/workflows/packages.yml
```diff
@@ -0,0 +1,97 @@
...
+                    restore-keys: |
+                        ${{ runner.os }}-php-${{ matrix.php }}-package-${{ matrix.package }}-composer-
+                        ${{ runner.os }}-php-${{ matrix.php }}-package-
+
+            -   
+                name: Restrict Symfony version
+                if: matrix.symfony != ''
+                run: |
+                    composer global require --no-progress --no-scripts --no-plugins "symfony/flex:^1.10"
+                    composer config extra.symfony.require "${{ matrix.symfony }}"
```
stof, line 81: shouldn't this have `working-directory` too, to set this `extra.symfony.require` config in the right composer.json ?

pamil: Good catch, thanks!

PR #12090, opened 2020-11-26
[CI] Switch application build from Travis to GitHub Actions

pamil:
The last PR of the three-part series (#12087, #12089) - switching the application build from Travis CI to GitHub Actions.

Non-JS Behat take around 20% longer (from ~10-11 minutes to ~12-14 minutes) and PHPUnit around 50% longer (from ~1.5m to ~2.5m), I've got no idea what is it caused by, but we've much better parallelization of the builds, so we should be able to get feedback much faster anyway.


pamil:
The execution time will be improved by #12093. Also, there might be some rough edges, we'll probably need to add some conditional waits to our JS Behat suite, but it'll make it less fragile in the end.

oallain (commented):
It's a good job :muscle: , here are some questions.

I think PHPSpec and PHPUnit should be run separately from Behat, WDYT ?

The check by Easy Coding Standard isn't made by the CI?

Regards :handshake:

pamil:
@oallain we can put PHPSpec into static checks, although the name isn't perfect, and it's just a few seconds. PHPUnit is bundled with Behat cause both needs fully set-up application with the database schema created as well (and it takes some time).

As for ECS, we apply coding standard fixes manually from time to time. We don't want to interrupt contributors with failures related to the coding standards.

oallain:
Hello @pamil ,

Thanks for your answer, now I understand better and therefore I share your opinion :wink: 

So the PR is good for me, just one last idea. You can export Behat logs as artifact to help debug, like :

```yaml
name: Upload Behat logs as artifact
uses: actions/upload-artifact@v2
if: failure()
with:
    name: Behat logs
    path: etc/build/
    if-no-files-found: ignore
```

Regards

pamil:
@oallain thanks, it's a good idea, got it in Sylius-Standard PR (https://github.com/Sylius/Sylius-Standard/pull/443), will send a PR here soon as well! :)

.github/workflows/application.yml
```diff
@@ -0,0 +1,343 @@
...
+                name: Validate Yaml files
+                run: bin/console lint:yaml src
+                
+            -
+                name: Run Psalm
+                run: vendor/bin/psalm --show-info=false --output-format=github
+                
+            -
+                name: Run PHPStan
+                run: vendor/bin/phpstan analyse -c phpstan.neon -l 1 src
```
stof, line 93: As you run several independent test commands in the same job, there is a trick to make them run even if the others are failing (while still making the job fail so not using `continue-on-error`), so that you get the result from all of them in a single execution:

```diff
             -
                 name: Install PHP dependencies
                 run: composer update --no-interaction --no-scripts
+                id: end-of-setup

             -
                 name: Check for security vulnerabilities
                 run: symfony security:check
+                if: always() && steps.end-of-setup.outcome == 'success'

             -
                 name: Validate Doctrine mapping
                 run: bin/console doctrine:schema:validate --skip-sync -vvv
+                if: always() && steps.end-of-setup.outcome == 'success'
```

the `always()` part of the condition is necessary because any `if` expression which does not include one of the status check function receives an implicit `success()` check. The other part is about checking that the setup was successful

pamil: Wow, that's cool! I'll open a PR soon.

stof: that was exactly my reaction when I discovered it :smile:

PR #12091, opened 2020-11-26
[WIP][API] placing order on multiple channels

arti0090:
| Q               | A
| --------------- | -----
| Branch?         |  master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| License         | MIT


features/checkout/placing_order_on_multi_channel/placing_an_order_on_multiple_channels_with_different_currency.feature
```diff
@@ -5,8 +5,8 @@ Feature: Placing an order on multiple channels with different currency
     I want user to place orders in channel’s base currency
 
     Background:
-        Given the store operates on a channel named "United States" in "USD" currency
-        And the store operates on another channel named "Colombia" in "COP" currency
+        Given the store operates on a channel named "United States" in "USD" currency with hostname "usa-store"
```
AdamKasp:
```suggestion
        Given the store operates on a channel named "United States" in "USD" currency with hostname "usa-store.com"
```

src/Sylius/Behat/Context/Api/Admin/ManagingOrdersContext.php
```diff
@@ -258,4 +258,12 @@ public function theCustomerServiceShouldKnowAboutThisAdditionalNotes(
 
         Assert::same($notes, $orderNotes);
     }
+
+    /**
+     * @Then /^(the administrator) should see that (order placed by "[^"]+") has "([^"]+)" currency$/
+     */
+    public function theAdministratorShouldSeeThatThisOrderHasBeenPlacedIn(): void
+    {
+        // Intentionally left blank, we do not return currency in API at the moment
```
GSadee, line 267: So IMO we should change the serialization of order to return currency code and to be able to check that

src/Sylius/Behat/Context/Setup/Api/CartContext.php
```diff
@@ -0,0 +1,128 @@
+<?php
+
```
AdamKasp, line 2: missing doc bloc

And I am not sure that this context is necessary, Setup should not depend on UI or API context.

arti0090: I would like to add @GSadee @lchrusciel for our little conversation 🍵

src/Sylius/Behat/Context/Setup/Api/CartContext.php
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /**
+     * @Given /^I (?:have|had) (product "[^"]+") in the (cart)$/
+     * @Given /^I (?:have|had) (product "[^"]+") added to the (cart)$/
+     * @Given /^I added (product "[^"]+") to the (cart)$/
+     * @Given /^the (?:customer|visitor) has (product "[^"]+") in the (cart)$/
+     * @When /^the (?:customer|visitor) try to add (product "[^"]+") in the customer (cart)$/
+     */
+    public function iAddedProductToTheCart(ProductInterface $product, ?string $tokenValue): void
```
GSadee, line 52: In my opinion, in setup, we should dispatch commands instead of making requests as it was already done. Why did you change that in this PR?

src/Sylius/Behat/Context/Setup/Api/CartContext.php
GSadee: I've noticed the same thing in a comment below, @arti0090 could you tell us why you added this context and changed the current setup cart context? Because I agree with @AdamKasp that the setup shouldn't depend on UI or API and we should dispatch proper commands there.

src/Sylius/Behat/Context/Setup/Api/CartContext.php
arti0090: Dispatching commands was not setting correct hostname to make requests by API. They were called (API calls) to lets say 'default.com' not 'usa-store.com' hence it generate the problem of creating order in wrong channel.

src/Sylius/Behat/Context/Api/Admin/ManagingOrdersContext.php
arti0090: We have a task for it in backlog, so should I change serialization here? or we change it in another task and put a note that this step should be implemented?

src/Sylius/Behat/Context/Setup/Api/CartContext.php
arti0090: By utilizing old setup/CartContext there are few problems:

1.  We are mixing contexts so the behat doesn't f.e. choose correct channel and then when going from setup to api doesn't set this channel to order
2. Before creating order we need to choose on which hostname (channel for API) we are operating, setting this hostname is crucial for next steps
3. In behat we have those two steps `Given I changed my current channel to "United States"` and `And I have product "PHP T-Shirt" in the cart` when I created them with setup/CartContext on the execution there was an exception 
Zrzut ekranu 2020-11-27 o 09 06 49

src/Sylius/Behat/Context/Api/Admin/ManagingOrdersContext.php
GSadee: In my opinion, we should cover all steps from scenarios that we are implementing in API context without leaving them working confusingly

src/Sylius/Behat/Context/Setup/Api/CartContext.php
lchrusciel: We are hitting the problems of stateful commands. ChannelContext is taking the default channel (or not at all), so our pickup-cart logic begun to be broken. Removing coupling of PickupCartHandler with ChannelContext and moving channel itself to the command would allow us to keep the current channel in shared storage and use it to pick a cart. On the command implementation, I think it should be part of data transformers (similar to order token value, but here we can use context)

src/Sylius/Behat/Context/Setup/Api/CartContext.php
lchrusciel: I would move back setup to command dispatching only and refactor of pickup cart command, so we can add channel to it

PR #12093, opened 2020-11-26
Use sha512 to encode passwords in test environments

pamil:
See #11599.


Tomanhez (approved):

Zales0123 (approved):

Zales0123:
Thank you, Kamil! :tada:

Kocal:
Why not using `plaintext` as encoder? That would be faster no?

Issue #12095, opened 2020-11-27
bad use custom separator on taxon fullname 


jbcr:
**Sylius version affected**: 1.4 to 1.8

**Description**  

when you call `getFullname` on taxon with custom separator, the custom separator is not used on parent taxon full name-calling.

**Steps to reproduce**  

```php
$root = new Taxon();
$root->setName('Root');

$level1 = new Taxon();
$level1->setName('Level 1');
$level1->setParent($root);
$root->addChild($level1);

$level2 = new Taxon();
$level2->setName('Level 2');
$level2->setParent($level1);
$level1->addChild($level2);

$level2->getFullname(' -- ');
```
Expected: `Root -- Level 1 -- Level2`

Result: `Root / Level 1 -- Level2`

**Possible Solution**  

When a taxon has a parent taxon, call `getFullname` method with `$pathDelimiter` in the parameter.

jbcr:
Solved in Sylius 1.9

PR #12096, opened 2020-11-27
fix separator propagation

jbcr:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #12095
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Taxonomy/spec/Model/TaxonSpec.php
+++ b/src/Sylius/Component/Taxonomy/spec/Model/TaxonSpec.php
@@ -130,6 +130,18 @@ final class TaxonSpec extends ObjectBehavior
 
     function its_fullname_prepends_with_parents_fullname(TaxonInterface $root): void
     {
+        $root->getFullname()->willReturn('Category');
+        $this->setName('T-shirts');
+
+        $root->addChild($this)->shouldBeCalled();
+        $this->setParent($root);
+
+        $this->getFullname()->shouldReturn('Category / T-shirts');
+        $this->getFullname(' -- ')->shouldReturn('Category -- T-shirts');
+    }
+    
+    function its_fullname_prepends_with_parents_fullname_with_custom_separator(TaxonInterface $root): void
+    {
         $root->getFullname()->withArguments([' / '])->willReturn('Category');
         $root->getFullname()->withArguments([' -- '])->willReturn('Category');
         $this->setName('T-shirts');
--- a/src/Sylius/Component/Taxonomy/spec/Model/TaxonSpec.php
+++ b/src/Sylius/Component/Taxonomy/spec/Model/TaxonSpec.php
@@ -15,6 +15,7 @@ namespace spec\Sylius\Component\Taxonomy\Model;
 
 use Doctrine\Common\Collections\Collection;
 use PhpSpec\ObjectBehavior;
+use Sylius\Component\Taxonomy\Model\Taxon;
 use Sylius\Component\Taxonomy\Model\TaxonInterface;
 
 final class TaxonSpec extends ObjectBehavior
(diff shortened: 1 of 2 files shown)


src/Sylius/Component/Taxonomy/spec/Model/TaxonSpec.php
```diff
@@ -130,7 +130,8 @@ function its_fullname_equal_name_if_no_parent(): void
 
     function its_fullname_prepends_with_parents_fullname(TaxonInterface $root): void
     {
-        $root->getFullname()->willReturn('Category');
```
rimas-kudelis, line 133: I would suggest to keep the original test (a call with no arguments), but add the second one as you did.

jbcr: Changed

src/Sylius/Component/Taxonomy/spec/Model/TaxonSpec.php
```diff
@@ -139,6 +139,19 @@ function its_fullname_prepends_with_parents_fullname(TaxonInterface $root): void
...
+    function its_fullname_prepends_with_parents_fullname_with_custom_separator(TaxonInterface $root): void
+    {
+        $root->getFullname()->withArguments([' / '])->willReturn('Category');
+        $root->getFullname()->withArguments([' -- '])->willReturn('Category');
+        $this->setName('T-shirts');
+
+        $root->addChild($this)->shouldBeCalled();
+        $this->setParent($root);
+
+        $this->getFullname()->shouldReturn('Category / T-shirts');
```
rimas-kudelis, line 152: I guess you wanted that custom argument here too?

jbcr: I want to test the default value.

jbcr:
I have fixed the original test. And add a middle taxon in a specific test for this fix.

src/Sylius/Component/Taxonomy/spec/Model/TaxonSpec.php
rimas-kudelis: But the default value is tested in the preceding test.

rimas-kudelis:
No offense, but I almost feel like you're just committing random changes now, such as totally irrelevant (in this bug) changes to `AddressType`.

As for tests, they still need some polishing. I could help with that if you want.

jbcr:
> No offense, but I almost feel like you're just committing random changes now, such as totally irrelevant (in this bug) changes to `AddressType`.

The change into `AddressType` is coming from another fix. I have removed it. Thank you.

> 
> As for tests, they still need some polishing. I could help with that if you want.
> 
> P.S. just in case, I'm not a team member, so all of my comments are just an opinion.

This change fixes the separator propagation. In the test, I need a root taxon, one child, and one grandchild.

Please, can you suggest a better code to test my change?

rimas-kudelis:
I would revert the changed line in the first test. There's no point in testing that function with the explicit `' / '`, since it's the default value anyway, and the default must be tested anyway, and you're already testing with a non-default value next.

Also, consider injecting `$middle` the same way `$root` is injected into your test.

I think it would even make sense to either merge both tests into one, or at least remove duplicate assertions from the second test. There's no need to test what `$root->getFullName()` returns twice (or even four times, in this case). Again, I would consider merging the tests into one, and just testing that method on `$root` (first-level taxon) and `$this`(third-level taxon).

jbcr:
If I don't add `->withArguments(...)` in the test `its_fullname_prepends_with_parents_fullname`, the test fail.

To test the separator propagation, the middle taxon must be a real object, not a mock.

Can you suggest a code? Because the instructions are subject to a wrong understanding.

To prevent the side effects, I test again the default behavior.

Merge the test `its_fullname_prepends_with_parents_fullname` and `its_fullname_prepends_with_parents_fullname_with_custom_separator` does not make sense because, I want cover a bug. Now, I prefer to wait for the Core team's opinion on this point.

src/Sylius/Component/Taxonomy/spec/Model/TaxonSpec.php
jbcr: I test another behavior from the previous test

rimas-kudelis:
> If I don't add `->withArguments(...)` in the test `its_fullname_prepends_with_parents_fullname`, the test fail.

Oh wait, that was a `willReturn()`, not a `shouldReturn()`. My bad. 

Now that I think about it, perhaps you don't even need the middle taxon in the second test. What you need is to check that $root is always called with the separator argument. So only the changed first line is necessary in the first test (i'd also inject `->shouldBeCalled()` before `->willReturn()` there), and the second test would be basically the same, but test with a custom separator.

What I'm saying is actually the existing test was wrong. I suppose it would fail if left unchanged and the code was fixed, no?

> Now, I prefer to wait for the Core team's opinion on this point.

Of course.

jbcr:
Solved in Sylius 1.9

PR #11547, opened 2020-06-04
Create new scenario for issue #11530

deStrO:
Adding a product to the cart with a promotion of 100% on shipping and 100% on the product produces an error

| Q               | A
| --------------- | -----
| Branch?         | 1.7 (tested on 1.7.4 and 1.7.3)
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #11530
| License         | MIT

PR #12110, opened 2020-12-01
Add behat scenario with free order

clem21:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | #11547
| License         | MIT

This is continuation of #11547

### DIFF
--- a/features/checkout/adding_product_with_promotion_that_makes_the_order_free.feature
+++ b/features/checkout/adding_product_with_promotion_that_makes_the_order_free.feature
@@ -1,6 +1,8 @@
 @checkout
-Feature: Adding product with promotion that makes the order free
-    I want to buy a product with a promotion that gives me free shipping and free order
+Feature: Buying product with promotion that makes the order free
+    In order to buy a product that is free
+    As a Customer
+    I should be able to purchase a full discounted order
 
     Background:
         Given the store operates on a single channel in "United States"
@@ -13,12 +15,12 @@ Feature: Adding product with promotion that makes the order free
         And I am a logged in customer
 
     @ui
-    Scenario: Adding a simple product to the cart
-        Given I have product "T-shirt banana" in the cart
-        And I am at the checkout addressing step
-        When I specified the billing address as "Ankh Morpork", "Frost Alley", "90210", "United States" for "Jon Snow"
+    Scenario: Buying product with promotion that makes the order free
+        When I add product "T-shirt banana" to the cart
+        And I go to the checkout addressing step
+        And I specified the billing address as "Ankh Morpork", "Frost Alley", "90210", "United States" for "Jon Snow"
         And I proceed with "DHL" shipping method
         Then I should be on the checkout summary step
         And "Holiday promotion" should be applied to my order shipping
-        Then my order total should be "$0.00"
+        And my order total should be "$0.00"
         And I confirm my order
--- a/features/checkout/adding_product_with_promotion_that_makes_the_order_free.feature
+++ b/features/checkout/adding_product_with_promotion_that_makes_the_order_free.feature
@@ -11,16 +11,12 @@ Feature: Buying product with promotion that makes the order free
         And there is a promotion "Holiday promotion"
         And the promotion gives "100%" discount on shipping to every order
         And the promotion gives "100%" discount to every order
-        And the store allows paying offline
         And I am a logged in customer
 
     @ui
     Scenario: Buying product with promotion that makes the order free
-        When I add product "T-shirt banana" to the cart
-        And I go to the checkout addressing step
+        Given I have product "T-shirt banana" in the cart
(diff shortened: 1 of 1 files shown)


features/checkout/adding_product_with_promotion_that_makes_the_order_free.feature
```diff
@@ -0,0 +1,24 @@
+@checkout
+Feature: Adding product with promotion that makes the order free
+    I want to buy a product with a promotion that gives me free shipping and free order
```
AdamKasp, line 3: features should be described according to our convention.

features/checkout/adding_product_with_promotion_that_makes_the_order_free.feature
```diff
@@ -0,0 +1,24 @@
...
+        And there is a promotion "Holiday promotion"
+        And the promotion gives "100%" discount on shipping to every order
+        And the promotion gives "100%" discount to every order
+        And the store allows paying offline
+        And I am a logged in customer
+
+    @ui
+    Scenario: Adding a simple product to the cart
+        Given I have product "T-shirt banana" in the cart
+        And I am at the checkout addressing step
```
AdamKasp, line 18: IMO `Given` steps should be in the `Background`

features/checkout/adding_product_with_promotion_that_makes_the_order_free.feature
```diff
@@ -0,0 +1,24 @@
...
+        When I specified the billing address as "Ankh Morpork", "Frost Alley", "90210", "United States" for "Jon Snow"
+        And I proceed with "DHL" shipping method
+        Then I should be on the checkout summary step
+        And "Holiday promotion" should be applied to my order shipping
+        Then my order total should be "$0.00"
```
AdamKasp, line 23:
```suggestion
        And my order total should be "$0.00"
```

features/checkout/adding_product_with_promotion_that_makes_the_order_free.feature
```diff
@@ -0,0 +1,26 @@
...
+    @ui
+    Scenario: Adding a simple product to the cart
+        When I add product "T-shirt banana" to the cart
+        And I go to the checkout addressing step
+        And I specified the billing address as "Ankh Morpork", "Frost Alley", "90210", "United States" for "Jon Snow"
+        And I proceed with "DHL" shipping method
+        Then I should be on the checkout summary step
+        And "Holiday promotion" should be applied to my order shipping
+        And my order total should be "$0.00"
+        And I confirm my order
```
lchrusciel, line 26: This is not an assertion, but rather an action. It should be part of the when step. Maybe we need scenarios for that? Or let's just make a scenario, that checks if placing a free order is possible - we should remove most of these assertions as they are not a clue.

features/checkout/adding_product_with_promotion_that_makes_the_order_free.feature
```diff
@@ -0,0 +1,26 @@
...
+        And the store has a product "T-shirt banana" priced at "$10.00"
+        And the store has "DHL" shipping method with "$50.00" fee
+        And there is a promotion "Holiday promotion"
+        And the promotion gives "100%" discount on shipping to every order
+        And the promotion gives "100%" discount to every order
+        And the store allows paying offline
+        And I am a logged in customer
+
+    @ui
+    Scenario: Adding a simple product to the cart
```
lchrusciel, line 18: Is it a proper scenario name?
```suggestion
    Scenario: Placing free order
```
or
```suggestion
    Scenario: Buying product with promotion that makes the order free
```

features/checkout/adding_product_with_promotion_that_makes_the_order_free.feature
```diff
@@ -0,0 +1,26 @@
+@checkout
+Feature: Adding product with promotion that makes the order free
```
lchrusciel, line 2:
```suggestion
Feature: Buing product with promotion that makes the order free
```
file name should be adjusted as well

Zales0123: `Buying` would be even better 💃

features/checkout/adding_product_with_promotion_that_makes_the_order_free.feature
clem21: I am not sure I understand what you are saying, should I ask another scenario that checks if placing a free order is possible and then remove the assertions in this existing file ? Or should I just change this scenario ?

lchrusciel (approved):

lchrusciel:
Thank you, Clémence! :1st_place_medal:

PR #12114, opened 2020-12-02
[Behat]Javascript tests helper

arti0090:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no

Changes related to JS tests
Partially based on @pamil 's #12090

### DIFF
--- a/src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
+++ b/src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
@@ -438,16 +438,12 @@ final class ManagingProductsContext implements Context
         $this->sharedStorage->set('product', $product);
 
         if ($product->isSimple()) {
-            $this->testHelper->waitUntilPageOpens(3, function() use ($product): void {
-                $this->updateSimpleProductPage->open(['id' => $product->getId()]);
-            });
+            $this->testHelper->waitUntilPageOpens($this->updateSimpleProductPage, ['id' => $product->getId()]);
 
             return;
         }
 
-        $this->testHelper->waitUntilPageOpens(3, function() use ($product): void {
-            $this->updateSimpleProductPage->open(['id' => $product->getId()]);
-        });
+        $this->testHelper->waitUntilPageOpens($this->updateSimpleProductPage, ['id' => $product->getId()]);
     }
 
     /**
--- a/src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
+++ b/src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
@@ -27,6 +27,8 @@ use Sylius\Behat\Page\Admin\ProductReview\IndexPageInterface as ProductReviewInd
 use Sylius\Behat\Page\Admin\ProductVariant\CreatePageInterface as VariantCreatePageInterface;
 use Sylius\Behat\Page\Admin\ProductVariant\GeneratePageInterface;
 use Sylius\Behat\Page\Admin\ProductVariant\UpdatePageInterface as VariantUpdatePageInterface;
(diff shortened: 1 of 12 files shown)


src/Sylius/Behat/JavaScriptTestHelper.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\Behat;
```
GSadee, line 5:
```suggestion
namespace Sylius\Behat\Service\Helper;
```
?

src/Sylius/Behat/JavaScriptTestHelper.php
```diff
@@ -0,0 +1,50 @@
...
+    public function waitUntilNotificationPopups(int $timeout, callable $assertion): void
+    {
+        $start = microtime(true);
+        $end = $start + $timeout;
+
+        do {
+            try {
+                $assertion();
+            } catch (ElementNotFoundException $exception) {
+                usleep(100000);
```
GSadee, line 20: Maybe this value could be passed to this service as an argument in constructor

GSadee (approved):

arti0090:
If this will help with our JS tests i would think about some changes to it as logic is almost same for every function. Only the type of exception is different for those tests.

src/Sylius/Behat/Context/Ui/Admin/ManagingProductsContext.php
```diff
@@ -431,12 +438,16 @@ public function iWantToModifyAProduct(ProductInterface $product): void
         $this->sharedStorage->set('product', $product);
 
         if ($product->isSimple()) {
-            $this->updateSimpleProductPage->open(['id' => $product->getId()]);
+            $this->testHelper->waitUntilPageOpens(3, function() use ($product): void {
+                $this->updateSimpleProductPage->open(['id' => $product->getId()]);
+            });
```
pamil, line 443: If we have such a specific method, what do you think about passing the page instance and route options instead of the callable?

```suggestion
            $this->testHelper->waitUntilPageOpens(3, $this->updateSimpleProductPage, ['id' => $product->getId()]);
```

And we could even define a default timeout and use it as the last argument so that it could be even shorter:

```suggestion
            $this->testHelper->waitUntilPageOpens($this->updateSimpleProductPage, ['id' => $product->getId()]);
```

src/Sylius/Behat/Service/Helper/JavaScriptTestHelper.php
```diff
@@ -0,0 +1,78 @@
...
+namespace Sylius\Behat\Service\Helper;
+
+use Behat\Mink\Exception\ElementNotFoundException;
+use FriendsOfBehat\PageObjectExtension\Page\UnexpectedPageException;
+
+final class JavaScriptTestHelper implements JavaScriptTestHelperInterface
+{
+    private $sleepTime;
+
+    public function __construct(int $sleepTime)
```
pamil, line 14: `$microsecondsInterval`? `$usleepInterval`? It'd be cool to have some descriptive name for this argument.

pamil:
As for opening pages, a nice solution to consider would be building it in `Page` class, so that it would wait automatically. This would reduce the complexity and use this feature for every page. We could enable it conditionally for `Selenium2Driver` and `ChromeDriver` only.

arti0090:
I was able to extract logics to one private function, so there is no duplication. Also the timeout is now as argument of service (default) or we can use it as function argument. I still don't like look of `$callable = function() use` blocks 🙃

GSadee (approved):

GSadee:

PR #12117, opened 2020-12-02
Remove pamil/prophecy-common from require-dev dependencies

pamil:
Now I can abandon that library :tada:

### DIFF
--- a/src/Sylius/Component/Channel/spec/Context/CachedPerRequestChannelContextSpec.php
+++ b/src/Sylius/Component/Channel/spec/Context/CachedPerRequestChannelContextSpec.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 
 namespace spec\Sylius\Component\Channel\Context;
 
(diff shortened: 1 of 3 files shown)


src/Sylius/Component/Channel/spec/Context/CachedPerRequestChannelContextSpec.php
```diff
@@ -116,7 +115,30 @@ function it_does_not_cache_channel_not_found_exceptions_for_null_master_requests
...
+                    $this->channel = $channel;
+                }
+
+                /** @throws ChannelNotFoundException */
+                public function __invoke(): ChannelInterface
+                {
+                    $currentCounter = $this->counter;
+                    ++$this->counter;
+
+                    if ($currentCounter === 0) {
```
pamil, line 135: Yeah I know it could be made smarter by:

```suggestion
                    if ($this->counter++ === 0) {
```

But it feels better to have more descriptive code.

Issue #12111, opened 2020-12-01
Only validate configuration of promotions for active channel


mamazu:
**The issue**
When you have a promotion (like the item fixed discount) which is only activated in one channel, you have to still fill the configuration for all channels.

**Describe the proposed solution**
Remove validation for all channels and make the template similar to the channel pricing in the product (tab view).

**Describe alternatives you've considered**
One other hand you can just default the values (like the discount being 0%). I am not sure if this is a good idea as this would probably cause an adjustment to be created with amount being 0.

laurentdabbb:
Hi,
+1 for first proposed solution

mamazu:
This needs to be re-evaluated in 2.0. I think that should already be the case.

PR #12119, opened 2020-12-02
Introducing a unified template for configuration per channel

mamazu:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | related to #11529
| License         | MIT

The same as in #11529 now all instances that inherit from the `ChannelCollectionType` look the same. 

Here an example how it looks on the shipping method edit box:
image

## Known issues with this PR.
* This still however poses a question about empty configuration as mentioned in the issue #12111. Which is now becoming more relevant as you can't see the whole configuration at a glance.
* This maybe hinders a fast entry of data as you can't easily cycle through the fields with the tab key anymore. But it looks cool. I am open for suggestions on how to improve on this.

### DIFF
--- a/src/Sylius/Behat/Page/Admin/ProductVariant/GeneratePage.php
+++ b/src/Sylius/Behat/Page/Admin/ProductVariant/GeneratePage.php
@@ -67,8 +67,8 @@ class GeneratePage extends SymfonyPage implements GeneratePageInterface
     {
         return array_merge(parent::getDefinedElements(), [
             'code' => '#sylius_product_generate_variants_variants_%position%_code',
-            'price' => '#sylius_product_generate_variants_variants_%position%_channelPricings input[id*="%channelCode%"]',
-            'prices_validation_message' => '#sylius_product_generate_variants_variants_%position%_channelPricings ~ .sylius-validation-error',
+            'price' => '#sylius_product_generate_variants_variants_%position%_channelPricings_%channelCode% input[id*="%channelCode%"]',
+            'prices_validation_message' => 'div[data-form-collection-index="%position%"] div.tabular.menu ~ .sylius-validation-error',
         ]);
     }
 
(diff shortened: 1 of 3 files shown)


vvasiloi:
@mamazu would optionally allowing to expand the form and doing so automatically when there are validation errors help with the issues you mentioned?
Also, the second one could probably be solved with some JS, that will go to next tab and focus the first field when Tab is pressed.

mamazu:
No, the problem is that you have required fields that are hidden. Let's say for example if you have 3 channels in your shop and you want to create a promotion. You enable the promotion in the first channel. Then you add an action like for example order fixed amount: There the configuration **requires you** to configure it in all channels. (But you don't see that because the other channels are hidden in the other tabs.) It will still bubble the error and you will get error messages but as mentioned in the related issue we don't need to fill them in the first place.

About the Javascript, noted. I will try that sometime. :D

vvasiloi:
@mamazu You could perhaps show the "*" on the tab, WDYT?

mamazu:
I don't think that this is a good idea, especially for larger forms this is can be a huge computational problem (and I don't even know if twig supports that). Or can Symfony bubble the "require" up?

mamazu:
I don't have an idea why this pull request fails? Any help?

lchrusciel (commented):
Hey Max,

your proposal seems pretty nice. I'm in favor of merging it, while Victor's proposals could be added later. 

According to the build, it is happening due to random build fails. Can you rebase to the newest master?

mamazu:
Let's hope this fixes it.

mamazu:
Nope still failing for some odd reasons. I didn't even touch the PagerFantaBundle. Very strange.

lchrusciel:
#12140 should made build green. Can you rebase again?

Zales0123 (approved):

mamazu:
Yay, it went through. :D

lchrusciel:
may become messed-up.**

execute `git fetch origin && git reset "unified_channel_collection"` to update your local branch.
 
lchrusciel:
may become messed-up.**

execute `git fetch origin && git reset "unified_channel_collection"` to update your local branch.
 
mamazu:
@lchrusciel Looks like the pipeline is still failing because of Behat.

mamazu:
@GSadee could you check what is up with Behat? I don't get what the problem seems to be.

mamazu:
Closing as it has been replaced by #14158

PR #12149, opened 2020-12-09
Improve retrying Behat and reduce failure rate

pamil:
This change probably increases the execution time for JS Behat tests, but drastically reduces the failure rate.

I did a benchmark running `features/product/managing_products/changing_images_of_product.feature` 50 times and the results are:

Pre-change:

Passed: 34
Failed: 16

Post-change:

Passed: 49
Failed: 1

### DIFF
--- a/src/Sylius/Behat/Service/Setter/CookieSetter.php
+++ b/src/Sylius/Behat/Service/Setter/CookieSetter.php
@@ -83,7 +83,7 @@ final class CookieSetter implements CookieSetterInterface
             return true;
         }
 
-        if ($driver instanceof ChromeDriver && $driver->isStarted() === false) {
+        if ($driver instanceof ChromeDriver) {
             return true;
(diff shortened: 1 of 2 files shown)


Tomanhez (approved):

src/Sylius/Behat/Service/Setter/CookieSetter.php
```diff
@@ -83,7 +83,7 @@ private function shouldMinkSessionBePrepared(Session $session): bool
             return true;
         }
 
-        if ($driver instanceof ChromeDriver && $driver->isStarted() === false) {
```
pamil, line 86: What it means: every time administrator is logging in while testing JS scenarios - refresh the browser.

pamil:
And compared with some other GitHub Actions runs, it's not significantly slower, so it's a double win.

PR #12154, opened 2020-12-10
[API] show possible shipping methods for order

arti0090:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/serialization/Order.xml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/serialization/Order.xml
@@ -110,6 +110,7 @@
 
         <attribute name="localeCode">
             <group>shop:order:account:read</group>
+            <group>order:read</group>
             <group>checkout:read</group>
         </attribute>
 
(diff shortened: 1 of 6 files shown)


src/Sylius/Bundle/ApiBundle/Resources/config/serialization/Order.xml
```diff
@@ -110,6 +110,7 @@
 
         <attribute name="localeCode">
             <group>shop:order:account:read</group>
+            <group>order:read</group>
```
lchrusciel, line 113: Why do we need it here?

lchrusciel (approved):

src/Sylius/Bundle/ApiBundle/Resources/config/serialization/Order.xml
arti0090: The steps I have implemented or refactored in this PR are now calling for completed order with GET orders/id. Because there is no localeCode in them I have added this serialization group in here. Well to be honest this is not part of PR ( as title says) but now the steps are much better ( not performance wise as I need to make a one call per step more), but we have current state of the cart, not state from the step before.
Also I think this group should be here ( so we can see what locale is added to the cart/order).
Well You are right that this should not be here, but otherwise I would need to change step `And my order's locale should be "English (United States)"` and when we add this serialization in other PR we would need to make a refactor of it again.

arti0090: I think we should go with it here, If we make another PR for this one serialization change, it will give us more work in future 😢

lchrusciel:
Thank you, @arti0090! :tada:

PR #12203, opened 2020-12-29
[Admin,Ui] Added collapse menu

vasilvestre:
| Q               | A
| --------------- | -----
| Branch?         | master (can change to 1.7)
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

This PR goal is to update #11668 and take rework why it was reverted in #11674
- [x] Fix transition performance (but testing it on your local would be appreciated !)
- [x] Don't collapse all menu

##### Remember uncollapsed menu
Should we keep information about the open menus between page navigation?

##### Sub-menu second level support
I would like to discuss which level of sub-menu should be collapsed. In my opinion, starting from the second level of the sub-menu, they should be collapsed.

Note: this isn't relevant as Sylius does not allow more than a level of sub-menu. Why? I would like to fix it.

#### Performance

I've fixed performance using this: https://github.com/Sylius/Sylius/pull/11668#discussion_r520640746
Performance is now way better, I couldn't reproduce the laggy menu on local but using Firefox Performance tool, I saw that it requires 4-5x less CPU computing. **My procedure was that one: uncollapsing all menus, from the lower one to the highest.** 
I'm not good at CSS but it seems like the "all" directive requires the browser to redraw elements, pixels per pixel in all directions. Translate let you use Hardware acceleration and somehow manage to perform way better.

Before :
before

After :
after

We can clearly see here that opening the menu has a negative impact on performance. With translate usage, it doesn't.
My average FPS is still low as I'm running IDE + the browser in WSL2, it will work better on the user machine.

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
+++ b/src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
@@ -26,7 +26,6 @@ import './sylius-notification';
 import './sylius-product-images-preview';
 import './sylius-product-slug';
 import './sylius-taxon-slug';
-import './sylius-chart';
 import customCollapse from './sylius-custom-collapse';
 
 import StatisticsComponent from './sylius-statistics';
--- a/src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
+++ b/src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
@@ -26,6 +26,8 @@ import './sylius-notification';
 import './sylius-product-images-preview';
 import './sylius-product-slug';
 import './sylius-taxon-slug';
+import './sylius-chart';
+import customCollapse from './sylius-custom-collapse';
 
 import StatisticsComponent from './sylius-statistics';
 import SyliusTaxonomyTree from './sylius-taxon-tree';
@@ -119,6 +121,8 @@ $(document).ready(() => {
   const dashboardStatistics = new StatisticsComponent(document.querySelector('.stats'));
 
   $('.sylius-admin-menu').searchable('.sylius-admin-menu-search-input');
+
+  customCollapse('#sidebar .item .header', '#sidebar .item', 'open', false);
 });
 
 window.$ = $;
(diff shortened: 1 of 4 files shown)


src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
```diff
@@ -26,6 +26,8 @@ import './sylius-notification';
 import './sylius-product-images-preview';
 import './sylius-product-slug';
 import './sylius-taxon-slug';
+import './sylius-chart';
```
vvasiloi, line 29: What is this?

src/Sylius/Bundle/AdminBundle/Resources/private/sass/_ui.scss
```diff
@@ -97,6 +96,61 @@ a {
...
+                background: white;
+            }
+            &:after {
+                transform: rotate(90deg);
+                transition: .3s;
+            }
+        }
+        &.open,
+        &.current_ancestor {
+            > .menu{
```
vvasiloi, line 140:
```suggestion
            > .menu {
```

src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
vasilvestre: Idk.. this was in the first merged PR

lchrusciel: May you try to remove it and check if it is still needed?

lchrusciel (commented):
Two quick thoughts - 
1. usage of the search box and collapsable is misleading in my opinion. Once something is searched and collapsed only collapsed super menu is displayed - I would expect to uncollapse at this moment 
2. We are leaving `-` sign on active menu, while it does not work. I would say, it would be better to just hide this sign

and a little bit controversial, but yet another - 
once we move to another website menu is no longer collapsed. That was one of the points for reversal, so may not be relevant here, but just wanted to point it out.

src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
vasilvestre: I've remove it, I should try it again locally

PR #12226, opened 2021-01-06
Use CustomerProvider in OAuth UserProvider

kayue:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | -
| License         | MIT

---

The current OAuth UserProvider doesn't assign customer pool when create new customer. 
Using `CustomerProvider` will fix it.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
+++ b/src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
@@ -17,10 +17,10 @@ use Doctrine\Common\Persistence\ObjectManager;
 use HWI\Bundle\OAuthBundle\Connect\AccountConnectorInterface;
 use HWI\Bundle\OAuthBundle\OAuth\Response\UserResponseInterface;
 use HWI\Bundle\OAuthBundle\Security\Core\User\OAuthAwareUserProviderInterface;
+use Sylius\Bundle\ApiBundle\Provider\CustomerProviderInterface;
 use Sylius\Bundle\UserBundle\Provider\UsernameOrEmailProvider as BaseUserProvider;
 use Sylius\Component\Core\Model\CustomerInterface;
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -62,7 +62,7 @@ public function __construct(
         RepositoryInterface $oauthRepository,
         ObjectManager $userManager,
         CanonicalizerInterface $canonicalizer,
-        CustomerRepositoryInterface $customerRepository
+        CustomerProviderInterface $customerProvider
```
Zales0123, line 65: Unfortunately, changing the type of argument is a BC break :/ We should allow using both of them and introduce some check, together with deprecation info trigger

Zales0123 (commented):
Also, as you've mentioned customer pools, which are a Sylius Plus feature, maybe it can be fixed firstly on the Sylius Plus itself? cc @GSadee @lchrusciel

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
kayue: @Zales0123 will this be fixed in 1.9? good time to introduce BC break :)

AdamKasp: @kayue we can't merge this solution in this way.
We Shouldn't couple API bundle with core bundle

kayue: @AdamKasp I am aware. Feel free to fix this in a future release, just reporting bugs here. :)

PR #12247, opened 2021-01-14
Add independent error page template

kulczy:
Zrzut ekranu 2021-01-14 o 14 14 39


GSadee (approved):

Zales0123 (approved):
👍 but it would be great if we can use it already somewhere - to not have any dead code 🖖

lchrusciel (commented):
We could put it here: https://github.com/Sylius/Sylius/blob/master/templates/bundles/TwigBundle/Exception/error404.html.twig (and in other error templates) in the array of includes. This is exactly, what we should do in Sylius-Standard and will allow testing this behavior as well.

@kulczy that's all I've needed. Thanks 👍 

Needs to be open to 1.8

PR #12263, opened 2021-01-18
[Translations] Fix wrong select field in German

aleho:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Just a few updates to select fields.
Two basic issues: The naming of the `select` type should not be "Bundesland auswählen" and not every selection should be titled as such.

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Resources/translations/messages.de.yml
+++ b/src/Sylius/Bundle/AddressingBundle/Resources/translations/messages.de.yml
@@ -23,7 +23,7 @@ sylius:
         province:
             name: Name
             abbreviation: Abkürzung
-            select: Bundesland auswählen
+            select: Provinz auswählen
         zone:
             add_member: Mitglied hinzufügen
             members: Mitglieder
@@ -36,7 +36,7 @@ sylius:
             scope: Umfang
             scopes:
                 all: Alle
-            select: Bundesland auswählen
+            select: Mitglied auswählen
             select_scope: Bereich wählen
         zone_member:
-            select: Bundesland auswählen
+            select: Mitglied auswählen
--- a/src/Sylius/Bundle/AttributeBundle/Resources/translations/messages.de.yml
+++ b/src/Sylius/Bundle/AttributeBundle/Resources/translations/messages.de.yml
@@ -18,7 +18,7 @@ sylius:
             datetime: Datum / Zeit
             integer: Integer
             percent: Prozent
-            select: Bundesland auswählen
+            select: Auswahl
             text: Text
             textarea: Textfeld
             title: Attributtyp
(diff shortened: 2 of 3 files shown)


jbcr:
All translations are managed by Cowdin

https://docs.sylius.com/en/latest/book/contributing/translations/index.html

Zales0123:
Hello @aleho! Indeed, on the main repository, we manage only English translations. For everything else, please, use out [Crowdin](https://crowdin.com/project/sylius) platform - as mentioned by @jbcr 🖖 Thanks!

aleho:
I looked everywhere before opening this PR, but wasn't sure, because I couldn't find any source related to the wrong translations.
Now I'm pretty confident that the Crowdin data is "broken" somehow.

For example the `messages.en.yml` doesn't contain all the entries from the file for `AddressingBundle`.
AddressingBundle/messages.en.yml

If you look at the general translation of `select` (in all strings selection) there's even a comment from 3(!) years ago stating that "Bundesland auswählen" is apparently entirely wrong for a select field (he is correct in that statement):

all strings

This gave me the impression that the Crowdin translations might be outdated / deprecated (or else why the wait for 3 years?).

So how do you actually manage / update the translations and what do I have to do to suggest a similar approach to this PR?

The following actions would be required in my opinion:

- Allow `select` in bundles to be translated separately (or at least change them to "Auswahl" as well, which is not entirely correct but better)
- Approve the correction of "Bundesland auswählen" to "Auswahl" for general select fields / types?

Can I help with that in any way?

pamil:
Hello Alexander!

CrowdIn is our source of truth when it comes to translations and it's still used - however, we don't have anyone in Sylius team speaking German, so we can't really verify translations (that's why there's not much activity).

All translations have to be approved before they are synchronised with the repository - which is done by proofreaders on CrowdIn. I've just added you (`a.hofbauer`) as a proofreader for German and German, Switzerland (in case there's some shared issue). Now you should be able to approve translations yourself.

After the translation is approved, it's synchronized with GitHub on Monday morning (around 4-5am). Then the pull request is created by the bot. Same happens when the new translation keys are added to GitHub, they're synced with CrowdIn every Monday morning as well.

I've just approved the translation you suggested ("Aushwahl" for all select types) and forced the sync (#12269). This should solve the issue for now. However, it seems like it only updates the `master` branch, but not `1.8`, which is still supported - I'll dig into that and try to make it work for `1.8`.

Apart from that, I've also changed the setting to allow auto-translating same translation keys (like select) when introduced, but also allow changing them later to more specific ones. Right now you should be able to propose and approve specific select translations like "Provinz auswählen".

Please let me know if I could help more with this or any other translation-related issues. Thank you for contributing :)

aleho:
Thanks! That looks much better now!

PR #12293, opened 2021-01-31
Use authorization checker in data provider

kayue:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

`$user->getRoles()` is not the right way to check user role.


kayue:
Actually why is there a APIUserContext? does API use a different user other than the current user?

Rafikooo:
Hello @kayue,
Thank you for submitting the PR. Your proposal is valid, but this problem occurs in more places. Therefore, we decided to correct it everywhere at once, and we will close this PR 🚀 
Best regards

PR #12299, opened 2021-02-01
[PROMOTION][RULE]Add comparison operator for Cart Quantity and Item Total rules

Prometee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

Add a comparison operator to be able to make more complex rules instead of having only >= comparison operator available.
This PR allow to choose among basic comparison operators : ===, !==, <, >, <= or >= (the default legacy one)

image

### DIFF
--- a/src/Sylius/Bundle/PromotionBundle/Form/Type/Rule/CartQuantityConfigurationType.php
+++ b/src/Sylius/Bundle/PromotionBundle/Form/Type/Rule/CartQuantityConfigurationType.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sylius\Bundle\PromotionBundle\Form\Type\Rule;
 
 use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\IntegerType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Validator\Constraints\NotBlank;
@@ -29,6 +30,20 @@ final class CartQuantityConfigurationType extends AbstractType
                 'constraints' => [
                     new NotBlank(['groups' => ['sylius']]),
                     new Type(['type' => 'numeric', 'groups' => ['sylius']]),
+                ],
+            ])
+            ->add('comparison_operator', ChoiceType::class, [
+                'label' => 'sylius.form.promotion_rule.cart_quantity_configuration.comparison_operator.label',
+                'choices' => [
+                    'sylius.form.promotion_rule.cart_quantity_configuration.comparison_operator.choices.greater_than_equal' => '>=',
+                    'sylius.form.promotion_rule.cart_quantity_configuration.comparison_operator.choices.equal' => '===',
+                    'sylius.form.promotion_rule.cart_quantity_configuration.comparison_operator.choices.different' => '!==',
+                    'sylius.form.promotion_rule.cart_quantity_configuration.comparison_operator.choices.lower_than' => '<',
+                    'sylius.form.promotion_rule.cart_quantity_configuration.comparison_operator.choices.lower_than_equal' => '<=',
+                    'sylius.form.promotion_rule.cart_quantity_configuration.comparison_operator.choices.greater_than' => '>',
+                ],
+                'constraints' => [
+                    new NotBlank(['groups' => ['sylius']]),
                 ],
             ])
         ;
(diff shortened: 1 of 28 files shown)


src/Sylius/Bundle/PromotionBundle/Resources/translations/messages.en.yml
```diff
@@ -50,12 +50,30 @@ sylius:
...
                 amount: Amount
+                comparison_operator:
+                    label: Comparison operator
+                    choices:
+                        greater_than_equal: Greater than or equal (>=)
+                        equal: Equal (===)
+                        different: Different (!==)
+                        lower_than: Lower than (<)
+                        lower_than_equal: Lower than or equal (<=)
+                        greater_than: Greater than (>)
```
AdamKasp, line 76: Do we need this duplication?
I think that a common translation key for both item and cart will be enough.

AdamKasp:
And Behats are appreciated. :)

Prometee:
Do not stale

lchrusciel (commented):
Hey Francis,

thanks a lot for contribution. I've requested @CoderMaggie to decide if we want to have it in the codebase from the feature side.

From my side, we cannot merge this PR at this state for following reasons:
* Lack of behat tests (to cover these functionalities from the business perspective) or phpunit test (to cover these functionalities from the operating perspective)
* Lack of specs 
* Code duplication - the choice type could be extracted to separate class. In addition comparison logic could be extracted as well - just for sack of simplicity of testing each case
But before you will do any of that, lets see if we want to have it in our codebase

CoderMaggie:
it took some time for me to get there, sorry 😓 

Of course, we do want to have this feature in Core 🙏🏻 . @Prometee will you be able to continue working on this? We're awaiting tests as @lchrusciel mentioned above.

Prometee:
@CoderMaggie Yes sure, I will make some refactoring as @lchrusciel suggested.

src/Sylius/Bundle/PromotionBundle/Form/Type/Rule/CartQuantityConfigurationType.php
```diff
@@ -13,23 +13,52 @@
...
 use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\IntegerType;
 use Symfony\Component\Form\FormBuilderInterface;
 
 final class CartQuantityConfigurationType extends AbstractType
 {
+    public function __construct(private ?ComparisonOperatorMatcherInterface $comparisonOperatorMatcher = null)
+    {
+    }
```
GSadee, line 36: Deprecate not passing this matcher and note it in the UPGRADE file, the same comment applies to the other form type and checkers

src/Sylius/Component/Core/Factory/PromotionRuleFactoryInterface.php
```diff
@@ -23,9 +23,9 @@
  */
 interface PromotionRuleFactoryInterface extends FactoryInterface
 {
-    public function createCartQuantity(int $count): PromotionRuleInterface;
+    public function createCartQuantity(int $count, string $comparisonOperator = '>='): PromotionRuleInterface;
 
-    public function createItemTotal(string $channelCode, int $amount): PromotionRuleInterface;
+    public function createItemTotal(string $channelCode, int $amount, string $comparisonOperator = '>='): PromotionRuleInterface;
```
GSadee: There are two ways to resolve this: either actually extend these methods in the interface, but document this correctly in the UPGRADE file, although this may still be considered a BC break, or do not extend these methods in the interface, but instead add additional optional arguments to the classes + document that these methods will change in 3.0.

GSadee: What is more, the default comparison operator could be extracted to const

SzymonFilipek:
Hi @Prometee!
I've done the QA for this feature and almost everything seems fine.
The only thing I've found was in the API. 
When creating/editing a promotion, there is no validation for the "comparison_operator". It can lead to situation like this:
image
Could you take a look at this? 
Thank you in advance!

SzymonFilipek:
Hi @Prometee!
QA was done after fixes and the validation works fine now. 
Thank you!

Prometee:
@GSadee I fix the API and made you required changes 😉

src/Sylius/Bundle/PromotionBundle/Validator/Constraints/ComparisonOperatorExists.php
```diff
@@ -0,0 +1,57 @@
...
+                static::class,
+            );
+
+            $this->message = $options['message'] ?? $this->message;
+            $groups ??= $options['groups'] ?? null;
+            $payload ??= $options['payload'] ?? null;
+        }
+
+        parent::__construct(groups: $groups, payload: $payload);
+    }
```
GSadee, line 46: As this is newly created constraint, there is no need to maintain options and trigger deprecation here

```suggestion
    #[HasNamedArguments]
    public function __construct(
        public string $message = 'sylius.promotion_rule.comparison_operator.invalid',
        ?array $groups = null,
        mixed $payload = null,
    ) {
        parent::__construct(groups: $groups, payload: $payload);
    }
```

GSadee (approved):

GSadee:
Thank you @Prometee! 🎉

PR #12312, opened 2021-02-05
[CI] Ignore changes in "adr" directory

pamil:
We don't need to trigger application build if we only change ADRs.


lchrusciel (approved):

lchrusciel:
Thanks, Kamil! :tada:

PR #12324, opened 2021-02-09
Removing hard timeout on edit while load selected values

christopherhero:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

In the js file, `src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-auto-complete.js` is set timeout = 5 sec to wait for an async response from ajax autocomplete action.

possibly this was added because someone calculated that this time would be enough to load mutation observers. However, when we add more than one field of type `ResourceAutocompleteChoiceType` it loads with a long delay.
In this PR I removed timeout and add the `refresh` action on the dropdown with 'set selected' action after the response receiving.


pamil (approved):

pamil:
Thank you, Krzysztof! :1st_place_medal:

PR #12333, opened 2021-02-11
[API] remove promotion coupon

arti0090:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Setup/CartContext.php
+++ b/src/Sylius/Behat/Context/Setup/CartContext.php
@@ -86,7 +86,9 @@ final class CartContext implements Context
      */
     public function iHaveVariantOfProductInTheCart(ProductVariantInterface $productVariant, ?string $tokenValue): void
     {
-        $tokenValue = $this->pickupCart($tokenValue);
+        if ($tokenValue === null) {
+            $tokenValue = $this->pickupCart();
+        }
 
         $this->commandBus->dispatch(AddItemToCart::createFromData(
             $tokenValue,
@@ -107,7 +109,9 @@ final class CartContext implements Context
         string $productOptionValue,
         ?string $tokenValue
     ): void {
-        $tokenValue = $this->pickupCart($tokenValue);
+        if ($tokenValue === null) {
+            $tokenValue = $this->pickupCart();
+        }
 
         $this->commandBus->dispatch(AddItemToCart::createFromData(
             $tokenValue,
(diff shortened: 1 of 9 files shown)


AdamKasp (approved):

src/Sylius/Behat/Context/Setup/CartContext.php
```diff
@@ -126,8 +123,22 @@ public function iAddThisProductWithToTheCart(
...
+    public function thisCartHasCouponAppliedWithCode(?string $tokenValue, string $couponCode): void
+    {
+        $tokenValue = $this->pickupCart($tokenValue);
+
+        $this->commandBus->dispatch(ApplyCouponToCart::createFromData($tokenValue, $couponCode));
+    }
+
+    private function pickupCart(?string $tokenValue = null): string
     {
+        if ($tokenValue !== null) {
```
GSadee, line 138: Why did you refactor that? IMO the previous behaviour makes more sense, when there is a method for pickuping the cart independently from passed token value and using that method only if needed.

src/Sylius/Bundle/ApiBundle/spec/CommandHandler/Cart/ApplyCouponToCartHandlerSpec.php
```diff
@@ -84,4 +84,19 @@ function it_throws_exception_if_promotion_coupon_is_not_found(
...
+
+    function it_removes_coupon_if_promotion_coupon_is_not_found(
+        OrderRepositoryInterface $orderRepository,
+        PromotionCouponRepositoryInterface $promotionCouponRepository,
+        OrderProcessorInterface $orderProcessor,
+        OrderInterface $cart
+    ): void {
+        $orderRepository->findCartByTokenValue('cart')->willReturn($cart);
+
+        $promotionCouponRepository->findOneBy(['code' => null])->willReturn(null);
```
GSadee, line 96: If I see correctly, it is not true. This method will not be used if the coupon code is null according to above implementation. 

This spec should be probably renamed to sth like `it_removes_coupon_if_passed_promotion_coupon_code_is_null`. 

There is also missing `$cart->setPromotionCoupon(null)->shouldBeCalled;`

src/Sylius/Behat/Context/Setup/CartContext.php
arti0090: It was used in every method of this Context where `pickupCart` was used. I moved it here so we don't rly duplicate this code.

src/Sylius/Bundle/ApiBundle/spec/CommandHandler/Cart/ApplyCouponToCartHandlerSpec.php
arti0090: Fixed, also I added this missing step for the rest of specs

src/Sylius/Behat/Context/Setup/CartContext.php
GSadee: As I wrote above, for me, the previous behaviour was more clear, now I am a little confused because there is method `pickupCart` and I can pass there a `$tokenValue`, so the expected behaviour is pickup the cart with given token, but it behaves differently. I would be for reverting this minor refactor.

GSadee (approved):

GSadee:
Thank you, @arti0090! :tada:

Issue #12298, opened 2021-02-01
The road to PHP 8.0


pamil:
It's high time to take care of PHP 8.0 support for Sylius, here's the list of dependencies that need to be upgraded first:

- [x] `friends-of-behat/mink`
- [x] `friends-of-behat/mink-browserkit-driver`
- [x] `friends-of-behat/mink-extension`
- [x] `friends-of-behat/symfony-extension`
- [x] `friends-of-behat/variadic-extension`
- [x] `friendsofsymfony/oauth-server-bundle` (https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/issues/672, not needed after AdminApiBundle is moved out of the core)
- [x] `friendsofsymfony/oauth2-php` (https://github.com/FriendsOfSymfony/oauth2-php/issues/123, not needed after AdminApiBundle is moved out of the core)
- [x] `lchrusciel/api-test-case` (https://github.com/lchrusciel/ApiTestCase/pull/179)
- [x] `liip/imagine-bundle` (https://github.com/liip/LiipImagineBundle/issues/1354)
- [x] `payum/payum`
- [x] `payum/payum-bundle` (https://github.com/Payum/PayumBundle/pull/524 - waiting for tag)
- [x] `php-http/guzzle6-adapter`
- [x] `polishsymfonycommunity/symfony-mocker-container` (https://github.com/PolishSymfonyCommunity/SymfonyMockerContainer/pull/15: it's a test-only dependency, could we replace it?)
- [x] `sspooky13/yaml-standards`
- [x] `stof/doctrine-extensions-bundle` (#12371)
- [x] `sylius-labs/association-hydrator`
- [x] `sylius-labs/coding-standard`
- [x] `sylius-labs/doctrine-migrations-extra-bundle`
- [x] `sylius/grid-bundle`
- [x] `sylius/registry`
- [x] `sylius/resource-bundle`
- [x] `sylius/theme-bundle`
- [x] `willdurand/hateoas-bundle` (https://github.com/willdurand/BazingaHateoasBundle/pull/95)
- [x] `winzou/state-machine`
- [x] `winzou/state-machine-bundle` (https://github.com/winzou/StateMachineBundle/pull/62)
- [x] `zendframework/zend-hydrator`
- [x] `zendframework/zend-stdlib`

goetas:
FYI https://github.com/willdurand/BazingaHateoasBundle/releases/tag/2.2.0 with php8 support

[...]

PR #12356, opened 2021-02-16
Replace php-http/guzzle6-adapter with php-http/guzzle7-adapter

pamil:
See #12298.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
         "liip/imagine-bundle": "^2.3",
         "payum/payum": "^1.6",
         "payum/payum-bundle": "^2.4",
-        "php-http/guzzle6-adapter": "^2.0",
+        "php-http/guzzle7-adapter": "^0.1.1",
         "polishsymfonycommunity/symfony-mocker-container": "^1.0",
         "ramsey/uuid": "^3.9",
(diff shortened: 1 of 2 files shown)


alexander-schranz:
Maybe it would make sense to use here the `symfony/http-client` instead of guzzle, maybe this is more future-proof as you build on top of symfony. See as example: https://github.com/FriendsOfSymfony/FOSHttpCacheBundle/pull/561/files

alexander-schranz:
Also the guzzle6 adapter was now merged with PHP 8 support. https://github.com/php-http/guzzle6-adapter/pull/86 and tagged as [2.0.2](https://github.com/php-http/guzzle6-adapter/releases/tag/v2.0.2). Still would go here with symfony/http-client :)

pamil:
Thanks for the info, @alexander-schranz! Symfony's HTTP Client would be the next step, but this is good enough for now :)

gomcodoctor:
Can we move to php-http/guzzle7-adapter instead of php-http/guzzle6-adapter till we waiting for Symfony's HTTP Client?

Keimpy:
Any update on this? Is it possible to update to guzzle 7?

PR #12358, opened 2021-02-16
Filter on product's name as well

jacquesbh:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

The variant's name is optional.
Meaning that we won't find any product with an empty variant name if we filter on the name.

Now we filter on the product's name as well so we can find what we are searching for!


jacquesbh:
Do not stale.

ping @Sylius/core-team

PR #12359, opened 2021-02-16
Avoid error if no enabled variant

jacquesbh:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

It seems we have an issue here if the product has no enabled variant.
This fixes the issue by just avoiding displaying the price (since no variant, no price!)


lchrusciel:
Hey Jacques,

thanks for your contribution. Can you cover this case with a simple behat scenario?

jakubtobiasz:
We've checked this and everything seems to work as expected. If you use it anywhere, you should check if there are any enabled variants. Displaying “nothing” might be misleading.

If you have any objections, please let me know, for now, I'm closing this PR. Thanks anyway :)!

PR #12376, opened 2021-02-22
Pass path delimiter up to parent taxon

Jeroen-G:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

This fixes a bug where this following taxon tree:

* A
    * B
        * C

would lead to a full name of `A  / B > C` instead of `A > B > C` if the call was `$taxonC->getFullname('>');`

### DIFF
--- a/src/Sylius/Component/Taxonomy/Model/Taxon.php
+++ b/src/Sylius/Component/Taxonomy/Model/Taxon.php
@@ -182,7 +182,7 @@ class Taxon implements TaxonInterface
 
         return sprintf(
             '%s%s%s',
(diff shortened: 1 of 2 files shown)


Zales0123 (approved):
👍 it would be perfect to have it checked also with PHPUnit or Behat... but it can be overkill :D 🚀

pamil (approved):
I've added a test for that case :)

pamil:
Thanks, JeroenG! :tada:

PR #12378, opened 2021-02-23
[Bug-fix] Fix default form problem with _token field

Zales0123:
Discussion
----------

| Q               | A
| --------------- | -----
| Branch?         | 1.9 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

After upgrade to Symfony 5, there is a problem with the default form for custom resources. Already rendered `_token` with `form_widget(form)` throws an exception when rendered again (previously it was just not rendered again).

Zrzut ekranu 2021-02-23 o 12 55 30

Probably it would be nice to have some test for a custom resource with default templates 💃 Write now it can be easily reproduced by removing `vars->all->templates->form` from one of the resources (e.g. `AdminUser`).

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/Crud/Create/_content.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/Crud/Create/_content.html.twig
@@ -9,6 +9,7 @@
     {% include '@SyliusAdmin/Crud/form_validation_errors_checker.html.twig' %}
     {% if configuration.vars.templates.form is defined %}
         {% include configuration.vars.templates.form %}
+        {{ form_row(form._token) }}
     {% else %}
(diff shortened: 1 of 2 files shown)


GSadee (approved):

pamil (approved):

lchrusciel (approved):

pamil:

maximehuran:
Why do we have to keep the ` {{ form_row(form._token) }}` when `configuration.vars.templates.form` is defined ?
I have an issue on one of our plugin because we customize the template form and I have the same error : 
```
An exception has been thrown during the rendering of a template ("Field "_token" has already been rendered, save the result of previous render call to a variable and output that instead.").
```

PR #12386, opened 2021-02-25
[API] Fix admin API filtering has no effect

Destronom:
Add grid config to customer API index routes. Otherwise if we want to filter the customers over criterias like: https://demo.sylius.com/api/v1/customers?criteria[search][type]=contains&criteria[search][value]=test it will have no effect on the expected result.

I think the grids must also be set on other _index routes in the AdminAPIBundle.

The filtering for the orders works because the grid is defined in order.yml:
sylius_admin_api_order:
    resource: |
        alias: sylius.order
        path: ''
        section: admin_api
        only: [show, index]
        grid: sylius_admin_order
        serialization_version: $version
    type: sylius.resource_api

| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | none found
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminApiBundle/Resources/config/routing/customer.yml
+++ b/src/Sylius/Bundle/AdminApiBundle/Resources/config/routing/customer.yml
@@ -11,6 +11,7 @@ sylius_admin_api_customer_index:
             serialization_groups: [Default]
             paginate: $limit
             sortable: true
+            grid: sylius_admin_customer
             sorting:
                 id: desc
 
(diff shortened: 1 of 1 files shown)


Rafikooo:
Hi @Destronom ,

Thank you for bringing this to our attention. However, we would like to let you know that the API V1 is no longer part of the Sylius Core. Instead, it is now developed in the [SyliusAdminApiBundle](https://github.com/Sylius/SyliusAdminApiBundle) repository. Therefore, we kindly ask you to open a pull request there instead of here. 
Thank you for your contribution and understanding.

Best regards! 🚀

PR #12391, opened 2021-03-01
[API] reseting password with validation

arti0090:
| Q               | A
| --------------- | -----
| Branch?         |master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

PR #12396, opened 2021-03-01
[API] Make Sylius compatible with api-platform v2.6

arti0090:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes, i think so 😄 
| License         | MIT

- [x] Add api-platform 2.6 in composer and remove conflict
- [x] Prepare behats that test Validation to use 400 and 422 http status code as correct
- [x] Fix file PasswordReset.xml ( from 2.6 it shows error when there is no property to identify by)
- [x] Find the solution to fix proper IRI resolution
- [ ] Adapt solution for `red` scenarios

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "ext-exif": "*",
         "ext-fileinfo": "*",
         "ext-gd": "*",
-        "api-platform/core": "^2.5",
+        "api-platform/core": "^2.6",
         "babdev/pagerfanta-bundle": "^2.5",
         "behat/transliterator": "^1.3",
         "doctrine/collections": "^1.6",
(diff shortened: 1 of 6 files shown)


arti0090:
I have found that using `<attribute name="identifiedBy">code</attribute>` on routes, can solve issues with identifiers for generating IRI. Btw `id` is now a default identifier in api-platform so If resource is using 2 identifiers, we should treat id as default one and 2nd identifier should be added for route as attribute (see Product.xml file as example)

arti0090:
When this PR will be resolved task  #12391 should be reverted to commit https://github.com/Sylius/Sylius/pull/12391/commits/2149e879c5ac8011801c3cf829c012bfdc33b0b5

pamil:
I believe API Platform 2.6 compatibility has been merged in #12510, thanks for contributing!

PR #12397, opened 2021-03-01
Do not fail when mb_detect_encoding returns false

stefandoorn:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

In case `mb_detect_encoding` returns false, the call to `mb_convert_case` results in a `TypeError` as it has to be a nullable string as input.


pamil (approved):

pamil:
Thank you, Stefan! :1st_place_medal:

PR #12406, opened 2021-03-02
Updated migrations path

TheDevilOnLine:
migrations have moved from vendor/sylius/plus/migrations to vendor/sylius/plus/src/Migrations

| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

### DIFF
--- a/docs/book/installation/sylius_plus_installation.rst
+++ b/docs/book/installation/sylius_plus_installation.rst
@@ -295,7 +295,7 @@ and apply them to your database:
 .. code-block:: bash
 
     bin/console doctrine:database:create --if-not-exists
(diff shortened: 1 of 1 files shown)


pamil (approved):

pamil (changes requested):
I think with newer versions of Sylius Plus (based on Sylius 1.8+), copying those should not be needed since they're autodiscovered, could you remove this line?

GSadee:
If I see correctly, this section has been already removed from docs

PR #12413, opened 2021-03-03
`JSON ` data type in `Version20201204071301` migration file causing compatibility issues with some MariaDB versions

Gogopex:
| Q               | A
| --------------- | -----
| Branch?         | 1.7, 1.8 or master 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | no
| License         | MIT

While creating a brand new Sylius project on 1.9, I encountered the following error during the step 2 of `bin/console sylius:install`:

```shell
  An exception occurred while executing 'ALTER TABLE sylius_adjustment ADD details JSON NOT NULL':

  SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the man
  ual that corresponds to your MariaDB server version for the right syntax to use near 'JSON NOT NULL' at lin
  e 1
```

For context, here's my MariaDB version:
```shell
mysq -V
mysql  Ver 15.1 Distrib 10.1.48-MariaDB, for debian-linux-gnu (x86_64) using readline 5.2
```

JSON itself as a data type is just an alias for LONGTEXT in MariaDB:

```
JSON is an alias for LONGTEXT introduced for compatibility reasons with MySQL's JSON data type. MariaDB implements this as a LONGTEXT rather, as the JSON data type contradicts the SQL standard, and MariaDB's benchmarks indicate that performance is at least equivalent.
```
source: https://mariadb.com/kb/en/json-data-type/

However the JSON_VALID function can only be used on this aliased data type:
```
The JSON_VALID function can be used as a CHECK constraint. This constraint is automatically included for types using the JSON alias from MariaDB 10.4.3.
```
source: https://mariadb.com/kb/en/json-data-type/

As I couldn't see any specific requirements for which min version of MariaDB to use, and the environment I'm in is also within the requirements listed here: https://docs.sylius.com/en/latest/book/installation/requirements.html, I decided to open this PR.

As I imagine this cannot be merged for several reasons, I've also created a patch which can be included in your composer.json if, like me, your version of MariaDB does not support the JSON alias yet, but you cannot update it.

```shell
composer require cweagans/composer-patches
```

and, in the "extra" part of your composer.json:
```json
        "patches": {
            "sylius/sylius": {
                "Changes JSON data type to LONGTEXT in a migration file to avoid breaking backwards compatilibity with MariaDB versions": "https://gist.githubusercontent.com/Gogopex/b82f68c91c916374ac8d2145e68da086/raw/ec97d11dcb4da6dd54765822d98ec3345bb197c0/.patch"
            }
        }
```

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Migrations/Version20201204071301.php
+++ b/src/Sylius/Bundle/CoreBundle/Migrations/Version20201204071301.php
@@ -30,7 +30,7 @@ final class Version20201204071301 extends AbstractMigration
             return;
         }
 
-        $this->addSql('ALTER TABLE sylius_adjustment ADD details JSON NOT NULL');
+        $this->addSql('ALTER TABLE sylius_adjustment ADD details LONGTEXT NOT NULL');
     }
 
     public function down(Schema $schema): void
--- a/src/Sylius/Bundle/CoreBundle/Migrations/Version20201204071301.php
+++ b/src/Sylius/Bundle/CoreBundle/Migrations/Version20201204071301.php
@@ -30,7 +30,7 @@ final class Version20201204071301 extends AbstractMigration
             return;
         }
 
-        $this->addSql('ALTER TABLE sylius_adjustment ADD details LONGTEXT NOT NULL');
+        $this->addSql('ALTER TABLE sylius_adjustment ADD details LONGTEXT NOT NULL COMMENT \'(DC2Type:json)\'');
     }
 
     public function down(Schema $schema): void


src/Sylius/Bundle/CoreBundle/Migrations/Version20201204071301.php
```diff
@@ -30,7 +30,7 @@ public function up(Schema $schema): void
             return;
         }
 
-        $this->addSql('ALTER TABLE sylius_adjustment ADD details JSON NOT NULL');
+        $this->addSql('ALTER TABLE sylius_adjustment ADD details LONGTEXT NOT NULL');
```
mbabker, line 33:
```suggestion
        $this->addSql('ALTER TABLE sylius_adjustment ADD details LONGTEXT NOT NULL COMMENT \'(DC2Type:json)\'');
```

Include the Doctrine type comment so values can be reverse engineered correctly.

Gogopex: Thank you. 👍

jacquesbh (approved):

pamil (commented):
Doesn't MariaDB support JSON data type alias from version 10.2.7 which was released in July 2017? 

Sources:
 - https://mariadb.com/kb/en/json-data-type/
 - https://mariadb.com/kb/en/mariadb-1027-release-notes/
 
Officially we only support MySQL ^5.7 and ^8.0, it would be amazing if we could add MariaDB on this list soon as well. In this case, we'd probably require at least that 10.2.7 version, so that we don't have to maintain different schemas for different databases.

This change would complicate maintaining MySQL migrations as well, projects which already upgraded their database would have JSON column type and the ones that will do it after the change would end up with LONGTEXT column type, which will cause issues.

Gogopex:
Hi @pamil,

Thanks for you answer ! You're entirely right about 10.2.7 being the first version to implement the JSON DataType, and also about the potential for issues if x% of people have a JSON column type and y% have a LONGTEXT column type.

I supposed the PR still had some use: we got to clarify a few things! 👍 🚀 

I'll close this PR, and if anyone with an <10.2.7 version  of MariaDB is having problems installing Sylius, you can still use the composer-patches method:

```
 I've also created a patch which can be included in your composer.json [...]

composer require cweagans/composer-patches
and, in the "extra" part of your composer.json:

        "patches": {
            "sylius/sylius": {
                "Changes JSON data type to LONGTEXT in a migration file to avoid breaking backwards compatilibity with MariaDB versions": "https://gist.githubusercontent.com/Gogopex/b82f68c91c916374ac8d2145e68da086/raw/ec97d11dcb4da6dd54765822d98ec3345bb197c0/.patch"
            }
        }
```

pamil:
Thanks a lot, @Gogopex!

PR #12420, opened 2021-03-04
[API] Adjust reset password requests

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Continuation of #12391

### DIFF
--- a/src/Sylius/Behat/Context/Api/Shop/LoginContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/LoginContext.php
@@ -61,7 +61,7 @@ final class LoginContext implements Context
      */
     public function iWantToResetPassword(): void
     {
-        $this->request = Request::create('shop', 'request-reset-password', 'Bearer');
+        $this->request = Request::create('shop', 'reset-password-requests', 'Bearer');
     }
 
     /**
@@ -81,7 +81,7 @@ final class LoginContext implements Context
     public function iFollowLinkOnMyEmailToResetPassword(ShopUserInterface $user): void
     {
         $this->request = Request::custom(
-            sprintf('api/v2/shop/reset-password/%s', $user->getPasswordResetToken()),
+            sprintf('api/v2/shop/reset-password-requests/%s', $user->getPasswordResetToken()),
             HttpRequest::METHOD_PATCH
         );
     }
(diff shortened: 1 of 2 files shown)


src/Sylius/Behat/Context/Api/Shop/LoginContext.php
```diff
@@ -61,7 +61,7 @@ public function iWantToLogIn(): void
      */
     public function iWantToResetPassword(): void
     {
-        $this->request = Request::create('shop', 'request-reset-password', 'Bearer');
+        $this->request = Request::create('shop', 'reset-password-requests', 'Bearer');
```
arti0090, line 64: Well I would call here also @GSadee as, first version was `reset-password-request`, then it was suggested to change to `reqest-reset-password` and PATCH `reset-password/{token}` now it is changed to `reset-password-requests` (plural form), and from what I see in our task there was suggested to use singular `reset-password-request`

GSadee: To be honest, I got a little lost in what these endpoints should look like, the current form is consistent with endpoints like `apply-coupon` or `change-quantity`, but these ones should be based on the API resource, so this is a good change.

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/ResetPassword.xml
```diff
@@ -15,15 +15,15 @@
            xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
            xsi:schemaLocation="https://api-platform.com/schema/metadata https://api-platform.com/schema/metadata/metadata-2.0.xsd"
 >
-    <resource class="Sylius\Bundle\ApiBundle\Command\ResetPassword" shortName="ResetPassword">
+    <resource class="Sylius\Bundle\ApiBundle\Command\ResetPassword" shortName="ResetPasswordRequest">
```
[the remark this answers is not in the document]
GSadee: I would be also for changing the command name, now it is quite confusing

GSadee (approved):

GSadee:

PR #12429, opened 2021-03-08
Use 4-Byte UTF-8 Unicode Encoding

jacquesbh:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -9,6 +9,6 @@ doctrine:
     dbal:
         driver: 'pdo_mysql'
         server_version: '5.7'
-        charset: UTF8
+        charset: utf8mb4
 
         url: '%env(resolve:DATABASE_URL)%'


pamil (commented):
What are the implications of this change for existing databases? Does it change the behaviour for columns length?

jacquesbh:
@pamil For me, no impact on column length or indices after 5.7.7. According to my searches.

pamil:
Indeed, looks like a change that we should do then! `utf8` is an alias for `utf8mb3` which is deprecated in favour of `utf8mb4`. I'm just wondering how to upgrade existing projects to `uft8mb4`, considering these concerns here: https://serversforhackers.com/c/mysql-utf8-and-indexing

pamil:
Indeed, looks like a change that we should do then! `utf8` is an alias for `utf8mb3` which is deprecated in favour of `utf8mb4`. I'm just wondering how to upgrade existing projects to `uft8mb4`, considering these concerns here: https://serversforhackers.com/c/mysql-utf8-and-indexing

jacquesbh:
The issue withe the indices is gone after 5.7.7 :).

config/packages/doctrine.yaml
```diff
@@ -9,6 +9,6 @@ doctrine:
     dbal:
         driver: 'pdo_mysql'
         server_version: '5.7'
-        charset: UTF8
+        charset: UTF8mb4
```
pamil, line 12:
```suggestion
        charset: utf8mb4
```

That looks like an official spelling, according to MySQL documentation.

pamil (commented):
Should we generate some migrations for this change to take place for existing data?

jacquesbh:
> Should we generate some migrations for this change to take place for existing data?

I don't think so… maybe a warning in the UPGRADE?

pamil (approved):

pamil:
Thanks, Jacques! :tada:

PR #12441, opened 2021-03-11
[UserBundle] Allow user password to be null for SSO

loic425:
| Q               | A
| --------------- | -----
| Branch?         | 1.7, 1.8 or master 
| Bug fix?        | no/yes
| New feature?    | no/yes
| BC breaks?      | no/yes
| Deprecations?   | no/yes 
| License         | MIT

When using a SSO such as Keycloak, you don't need a password. It's better to allow null instead of setting plain password with a random value.

Moreover, password will be decoupled from user interface on Symfony security user.
https://github.com/symfony/symfony/pull/40267

### DIFF
--- a/src/Sylius/Bundle/AdminApiBundle/Migrations/Version20210311142134.php
+++ b/src/Sylius/Bundle/AdminApiBundle/Migrations/Version20210311142134.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\AdminApiBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20210311142134 extends AbstractMigration
+{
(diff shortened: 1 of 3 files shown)


pamil (approved):

src/Sylius/Bundle/AdminApiBundle/Migrations/Version20210311142134.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\AdminApiBundle\Migrations;
```
loic425, line 5: @pamil I think this is a wrong namespace (automatically generated).
Sylius\Bundle\UserBundle\Migrations ?

pamil: Oh, that's right! Let's move it to CoreBundle, we only provide migrations there.

pamil (changes requested):

pamil (approved):

pamil:
Thank you, Loïc! :tada:

PR #12459, opened 2021-03-19
Adding more properties to ShippingMethod fixture

mamazu:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

Adding a way in the fixtures to set the shipping method's position and category requirements.


lchrusciel (commented):
Hey Max,

can you add usage of these new options to the current fixtures?

mamazu:
Sure, what should I use as a default value for them? 
all?

mamazu:
Does anybody know what this error is?

mamazu:
Hm, now I it stays that I have a cyclic dependency. But that makes no sense since the `category_requirements` is a string.

PR #12469, opened 2021-03-22
Extract sylius/admin-api-bundle out of Sylius Core (but keep it installed by default in existing minor releases)

pamil:
**Why?**

 - Two packages required by `sylius/admin-api-bundle` are blocking Sylius from upgrading to PHP 8.
 - We're rolling out v2 of the API, for both storefront and admin usage, which are getting more and more mature.
 - This bundle will be extracted as for Sylius v1.8 (so that we can still include bugfixes) and removed from the default installation in Sylius v1.10. However, it will be relatively simple to add it back to v1.10 if you're still using it - composer require and that's it.
 - Admin API is not as widely used as Shop API anyway.

Reference: #12298


lchrusciel (approved):

pamil:
It almost works, but it requires adding `COMPOSER_ROOT_VERSION=dev-master` with `composer update` for any other branch that `master` (so `1.8` and `1.9` as well). 

If not added, resolving dependencies fails with the following message:
```
  Problem 1
    - sylius/admin-api-bundle[v1.8.0, ..., v1.9.1] require sylius/core-bundle ^1.6 -> found sylius/core-bundle[v1.6.0, ..., v1.9.1] but these were not loaded, likely because it conflicts with another require.
    - Root composer.json requires sylius/admin-api-bundle ^1.8 -> satisfiable by sylius/admin-api-bundle[v1.8.0, ..., v1.9.1].
```

I'm looking for ideas to make it work by default without having to add this env variable; it hurts DX a lot.

pamil:
Changing branch names helps, renaming `1.8` to `1.8.x` fixes the problem, but I'm not sure how it would work out for new branches based on those.

pamil:
Two more potential fixes:
 - explicitly specifying `version` in `composer.json` (prone to human error while tagging)
 - replacing packages for the whole minor version, like `1.8.*`, instead of `self.version` (but then eg. `sylius/sylius` v1.8.15 could be installed even if `sylius/core-bundle` v1.8.20 is required and some dependency prevents from installing more recent version of `sylius/sylius`)

pamil:
I guess I'll go with explicitly setting version in composer.json, we already do that in `Kernel` for Sylius version - it seems like the best possible solution.

pamil:
When upmerging, make sure all AdminApiBundle files are removed and the required version is bumped to `~1.9.0` for Sylius 1.9 and `^1.10` for Sylius 1.10 (dev-master).

GSadee (approved):

src/Sylius/Bundle/CoreBundle/Fixture/OptionsResolver/LazyOption.php
```diff
@@ -137,4 +137,35 @@ function (Options $options, $previousValue) use ($repository, $field, $criteria)
             }
         ;
     }
+
+    public static function getOneBy(RepositoryInterface $repository, string $field, array $criteria = []): \Closure
```
lchrusciel, line 141: Where is it used?

PR #12485, opened 2021-03-27
Use route name instead of a hardcoded '/' to redirect user on logout

rimas-kudelis:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

On logout, shop users are currently being redirected to a hardcoded `/` path. However, this isn't always desired. For example, if localized routes are involved, this effectively switches the shop to its default locale on logout.

This change passes `sylius_shop_homepage` route name instead of a `/` path to the LogoutHandler, meaning the actual path will be generated when the event occurs.


pamil (approved):

pamil:
Thank you, Rimas! :1st_place_medal:

PR #12486, opened 2021-03-28
[ApiBundle] Fix normalization of resource Order

floriangosse:
| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT


src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Order.xml
```diff
@@ -330,6 +343,9 @@
                 <attribute name="denormalization_context">
                     <attribute name="groups">admin:cart:update</attribute>
```
pamil, line 344: Not related to changes in this PR, but it looks like it should be `shop`, not `admin` here.

floriangosse:
@pamil I've a questions about the API responses:

1. Is there a guideline how nesting should be handled? Is the project fine with that or should it be avoided?
2. If nested is fine, can that be applied to the admin endpoints as well?

PR #12487, opened 2021-03-28
[Api][POC]Convert iri to code in command

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT


src/Sylius/Behat/Context/Api/Shop/ProductReviewContext.php
```diff
@@ -75,7 +75,7 @@ public function iAddIt(): void
     public function iWantToReviewProduct(ProductInterface $product): void
     {
         $this->client->buildCreateRequest();
-        $this->client->addRequestData('productCode', $product->getCode());
+        $this->client->addRequestData('productCode', $this->iriConverter->getIriFromItem($product));
```
GSadee, line 78:
```suggestion
        $this->client->addRequestData('product', $this->iriConverter->getIriFromItem($product));
```
but keeping the `productCode` on the command, could you investigate is it possible?

src/Sylius/Bundle/ApiBundle/Converter/CommandFieldItemIriToIdConverter.php
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Converter;
+
+final class CommandFieldItemIriToIdConverter implements CommandFieldItemIriToIdConverterInterface
```
GSadee, line 16:
```suggestion
final class CommandFieldItemIriToIdentifierConverter implements CommandFieldItemIriToIdentifierConverterInterface
```
to make it looks more generic

src/Sylius/Bundle/ApiBundle/Converter/ItemIriToIdConverter.php
```diff
@@ -0,0 +1,64 @@
...
+namespace Sylius\Bundle\ApiBundle\Converter;
+
+use ApiPlatform\Core\DataProvider\OperationDataProviderTrait;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+use ApiPlatform\Core\Exception\InvalidIdentifierException;
+use ApiPlatform\Core\Util\AttributesExtractor;
+use Symfony\Component\Routing\Exception\ExceptionInterface as RoutingExceptionInterface;
+use Symfony\Component\Routing\RouterInterface;
+
+final class ItemIriToIdConverter implements ItemIriToIdConverterInterface
```
GSadee, line 23:
```suggestion
final class ItemIriToIdentifierConverter implements ItemIriToIdentifierConverterInterface
```

src/Sylius/Bundle/ApiBundle/Converter/ItemIriToIdConverter.php
```diff
@@ -0,0 +1,64 @@
...
+        $attributes = AttributesExtractor::extractAttributes($parameters);
+
+        try {
+            $identifiers = $this->extractIdentifiers($parameters, $attributes);
+        } catch (InvalidIdentifierException $e) {
+            throw new InvalidArgumentException($e->getMessage(), (int) $e->getCode(), $e);
+        }
+
+        if (is_array($identifiers)) {
+            throw new InvalidArgumentException('ItemIriToIdConverter do not support subresources');
```
GSadee, line 59:
```suggestion
            throw new InvalidArgumentException(sprintf('%s does not support subresources', self::class));
```

src/Sylius/Bundle/ApiBundle/Converter/ItemIriToIdConverterInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Converter;
+
+interface ItemIriToIdConverterInterface
+{
+    /** @return string */
+    public function getId(string $iri): string ;
```
GSadee, line 19:
```suggestion
    public function getIdentifier(string $iri): string;
```

src/Sylius/Bundle/ApiBundle/Serializer/CommandFieldItemIriToIdDenormalizer.php
```diff
@@ -0,0 +1,89 @@
...
+
+namespace Sylius\Bundle\ApiBundle\Serializer;
+
+use Sylius\Bundle\ApiBundle\Converter\CommandFieldItemIriToIdConverterInterface;
+use Sylius\Bundle\ApiBundle\Converter\ItemIriToIdConverter;
+use Symfony\Component\DependencyInjection\ServiceLocator;
+use Symfony\Component\Serializer\Normalizer\ContextAwareDenormalizerInterface;
+use Symfony\Component\Serializer\Normalizer\DenormalizerInterface;
+
+final class CommandFieldItemIriToIdDenormalizer implements ContextAwareDenormalizerInterface
```
GSadee, line 22: As above, `identifier` instead of `id`

src/Sylius/Bundle/ApiBundle/Converter/CommandFieldItemIriToIdConverter.php
GSadee: What is more, this service is not a converter, it is an object to store command and field

AdamKasp:
@Tomanhez, before potentially merge, you should add specs for it, and all classes should be `@experimental`

AdamKasp:
And probably you should update the description of this PR :)
And it is a new and custom logic, so I suggest adding a doc with an explanation of how it works and how to use it (maybe ADR?

src/Sylius/Bundle/ApiBundle/Command/AddProductReview.php
```diff
@@ -50,13 +50,13 @@ public function __construct(
         string $title,
         int $rating,
         string $comment,
-        string $productCode,
+        string $product,
```
pamil, line 53: This is definitely something we should investigate more about - currently we need to consider the constructor's variable names as a part of BC promise.

src/Sylius/Bundle/ApiBundle/Converter/CommandFieldItemIriToIdentifierStorage.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Converter;
+
+final class CommandFieldItemIriToIdentifierStorage implements CommandFieldItemIriToIdentifierStorageInterface
+{
+    /** @var string */
+    private $commandType;
```
pamil, line 19:
```suggestion
    private $commandClass;
```

It's a FQCN that is stored there.

src/Sylius/Bundle/ApiBundle/Converter/CommandFieldItemIriToIdentifierStorage.php
```diff
@@ -0,0 +1,39 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Converter;
+
+final class CommandFieldItemIriToIdentifierStorage implements CommandFieldItemIriToIdentifierStorageInterface
+{
+    /** @var string */
```
pamil, line 18:
```suggestion
    /** 
     * @var string 
     * @psalm-var class-string
     */
```

src/Sylius/Bundle/ApiBundle/Converter/CommandFieldItemIriToIdentifierStorageInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Converter;
+
+interface CommandFieldItemIriToIdentifierStorageInterface
```
pamil, line 16: As internal storage for data, I don't think we need an interface for this data structure.

src/Sylius/Bundle/ApiBundle/Converter/ItemIriToIdentifierConverter.php
```diff
@@ -0,0 +1,64 @@
...
+namespace Sylius\Bundle\ApiBundle\Converter;
+
+use ApiPlatform\Core\DataProvider\OperationDataProviderTrait;
+use ApiPlatform\Core\Exception\InvalidArgumentException;
+use ApiPlatform\Core\Exception\InvalidIdentifierException;
+use ApiPlatform\Core\Util\AttributesExtractor;
+use Symfony\Component\Routing\Exception\ExceptionInterface as RoutingExceptionInterface;
+use Symfony\Component\Routing\RouterInterface;
+
+final class ItemIriToIdentifierConverter implements ItemIriToIdentifierConverterInterface
```
pamil, line 23: I'm missing a comment that says that this class logic is basically API Platform's `IriConverter` without fetching object logic.

src/Sylius/Bundle/ApiBundle/Resources/config/serialization/Commands/AddProductReview.xml
```diff
@@ -28,7 +28,7 @@
             <group>shop:product_review:create</group>
             <group>shop:product_review:read</group>
         </attribute>
-        <attribute name="productCode">
+        <attribute name="product">
```
pamil, line 31: I'm not sure whether it would work, but have you tried `serialized-name`?
```suggestion
        <attribute name="productCode" serialized-name="product">
```

src/Sylius/Bundle/ApiBundle/Resources/config/services.xml
```diff
@@ -164,5 +164,23 @@
...
+            <argument type="service" id="Sylius\Bundle\ApiBundle\Converter\ItemIriToIdentifierConverter" />
+            <argument type="service" id="sylius.api.data_transformer.command_aware_input_data_transformer" />
+            <argument type="tagged_locator" tag="sylius.api.command_field_item_iri_to_identifier_storage" />
+            <tag name="serializer.normalizer" />
+        </service>
+
+        <service id="AddProductReviewCommandFieldItemIriToIdentifierStorage" class="Sylius\Bundle\ApiBundle\Converter\CommandFieldItemIriToIdentifierStorage">
+            <argument type="string">Sylius\Bundle\ApiBundle\Command\AddProductReview</argument>
+            <argument type="string">product</argument>
+            <tag name="sylius.api.command_field_item_iri_to_identifier_storage" />
```
pamil, line 183: This is my main concern about this implementation: this configuration would be much better off when either included in the command itself (XML metadata for API Platform / Serializer) or at least configurable in `sylius_api` bundle.

Having it defined as tagged services makes it really hard to see how and why does this conversion happens.

src/Sylius/Bundle/ApiBundle/Serializer/CommandFieldItemIriToIdentifierDenormalizer.php
```diff
@@ -0,0 +1,103 @@
...
+                ->commandFieldIriToIdentifierStorages
+                ->get($commandFieldItemIriToIdentifierStorageName)
+            ;
+
+            if ($this->getInputClass($context) === $commandFieldItemIriToIdentifierConverter->getCommandType()) {
+                return true;
+            }
+        }
+
+        return false;
```
pamil, line 64: This basically looks like:

```suggestion
        foreach ($this-commandFieldIriToIdentifierStorages as $name => $storage) {
            if ($this->getInputClass($context) === $storage->getCommandType()) {
                return true;
            }
        }

        return false;
```

src/Sylius/Bundle/ApiBundle/Serializer/CommandFieldItemIriToIdentifierDenormalizer.php
```diff
@@ -0,0 +1,103 @@
...
+
+        /** @var string $commandFieldIriToIdentifierStorageName */
+        foreach ($this->getCommandFieldIriToIdStoragesNames() as $commandFieldIriToIdentifierStorageName) {
+            /** @var CommandFieldItemIriToIdentifierStorageInterface $commandFieldItemIriToIdentifierStorage */
+            $commandFieldItemIriToIdentifierStorage = $this
+                ->commandFieldIriToIdentifierStorages
+                ->get($commandFieldIriToIdentifierStorageName)
+            ;
+
+            if ($inputClass === $commandFieldItemIriToIdentifierStorage->getCommandType()) {
```
pamil, line 79: This logic looks duplicated (look for the comment above) - I'd wrap it at least in a private method `provideCommandFieldName(string $inputClass): string` that throws an exception if it can't be found.

src/Sylius/Bundle/ApiBundle/Serializer/CommandFieldItemIriToIdentifierDenormalizer.php
```diff
@@ -0,0 +1,103 @@
...
+        if ($this->commandAwareInputDataTransformer->supportsTransformation($denormalizedInput, $type, $context)) {
+            return $this->commandAwareInputDataTransformer->transform($denormalizedInput, $type, $context);
+        }
+
+        return $denormalizedInput;
+    }
+
+    private function getInputClass(array $context): ?string
+    {
+        return isset($context['input']['class']) ? $context['input']['class'] : null;
```
pamil, line 96:
```suggestion
        return $context['input']['class'] ?? null;
```

src/Sylius/Bundle/ApiBundle/Serializer/CommandFieldItemIriToIdentifierDenormalizer.php
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private function getCommandFieldIriToIdStoragesNames(): array
```
pamil, line 99: If the comment above gets applied, this is no longer needed.

src/Sylius/Bundle/ApiBundle/Resources/config/services.xml
pamil: Or we can get rid of the whole storage class concept by just injecting a map of classes to fields directly to `CommandFieldItemIriToIdentifierDenormalizer` (or creating a new class for that). This way we don't have to create multiple services holding just two strings, there's a dedicated service for managing those instead.

```
class CommandIriArgumentToIdenitiferMap {
  /** @psalm-var array<class-string, string> Command FQCN as key, field as value */
  public function __construct(array $map) {}

  /** 
   * @psalm-var class-string $commandClass
   * @throws CommandNotFound
   */
  public function get(string $commandClass): string {}

  public function has(string $commandClass): bool {}
}
```

src/Sylius/Bundle/ApiBundle/spec/Map/CommandItemIriArgumentToIdentifierMapSpec.php
```diff
@@ -0,0 +1,50 @@
...
+    function it_is_command_item_iri_argument_to_identifier_map(): void
+    {
+        $this->shouldImplement(CommandItemIriArgumentToIdentifierMapInterface::class);
+    }
+
+    function it_get_element(): void
+    {
+        $map = new CommandItemIriArgumentToIdentifierMap(['Sylius\Bundle\ApiBundle\Command\Cart\PickupCart' => 'token']);
+        Assert::same('token', $map->get('Sylius\Bundle\ApiBundle\Command\Cart\PickupCart'));
+    }
```
GSadee, line 37: I guess that this should work:
```suggestion
    function it_gets_element(): void
    {
        $this->get('Sylius\Bundle\ApiBundle\Command\Cart\PickupCart'))->shouldReturn('token');
    }
```
Similar implementation could be implemented in other tests in this class

adr/2021_04_15_using_iri_instead_of_code_id.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Context and Problem Statement
+
+ApiPlatform provides and recommends using IRI as identifier. That identifier is more clearly than classic id and 
+contain more information. On resources, ApiPlatform handles IRI out of the box.
+During creation our new API in many places we decided to use commands while creating our endpoints. For certain cases
+this solution is more flexible than classic approach, but it doesn't handle transformation IRI to proper `code`/`id`.
+In the past we decided to use `code`/`id` instead of IRI. Now we try to unify API and replace this approach for solution with using IRI.
+Main problem is sent request with field as IRI and creation command without IRI, but with `id`/`code`. 
+It's needed for proper handling behaviour in command handler.
```
CoderMaggie, line 14:
```suggestion
API Platform recommends using IRI as an identifier. That identifier gives more clarity than an id because it contains more information - a full endpoint path to the resource and its unique identifier. On resources, API Platform handles IRI out of the box.
While we were designing Sylius new API, we have decided to use commands in many endpoints. 
In some cases, this solution is more flexible than the default creation approach (as we have control over what we are processing), but it doesn't handle the transformation of IRI to proper `code`/`id`.

In the past, we have been using `code`/`id` instead of IRI, then we have been using both approaches. 
Now we are trying to unify the new API and replace codes and ids with IRI everywhere.

The main challenge is the usage of IRI in requests, where we want to have IRI in the request but `id`/`code`in its commands and command handlers.
```

adr/2021_04_15_using_iri_instead_of_code_id.md
```diff
@@ -0,0 +1,46 @@
...
+## Considered Options
+
+### Handling and transforming IRI to `id`/`code`
+
+For handling IRI we created `Sylius\Bundle\ApiBundle\Serializer\CommandFieldItemIriToIdentifierDenormalizer` and
+`Sylius\Bundle\ApiBundle\Map\CommandItemIriArgumentToIdentifierMap`. `CommandFieldItemIriToIdentifierDenormalizer` 
+transforms and denormalizes our command using `CommandItemIriArgumentToIdentifierMap` as a bag with definition for
+supported denormalize commands. All you have to do for adding command support is add command FQCN as a key 
+and field name as a field that will be transformed from IRI to `code`/`id`. All definition are registered in 
+`Sylius\Bundle\ApiBundle\Map\CommandItemIriArgumentToIdentifierMap` service:
```
CoderMaggie, line 25:
```suggestion
For handling IRI we have created `Sylius\Bundle\ApiBundle\Serializer\CommandFieldItemIriToIdentifierDenormalizer` and
`Sylius\Bundle\ApiBundle\Map\CommandItemIriArgumentToIdentifierMap`. `CommandFieldItemIriToIdentifierDenormalizer` 
transforms and denormalizes the command using `CommandItemIriArgumentToIdentifierMap` as a bag with a definition for
the supported denormalizing commands. All one has to do for adding command support is to add the command FQCN as a key and the field name as a field that will be transformed from IRI to `code`/`id`. All definitions are registered in 
`Sylius\Bundle\ApiBundle\Map\CommandItemIriArgumentToIdentifierMap` service: 
```

adr/2021_04_15_using_iri_instead_of_code_id.md
```diff
@@ -0,0 +1,46 @@
...
+                <argument key="NewCommandFQCN">NewCommandFieldName</argument><-- Example -->
+            </argument>
+        </service>
+````
+
+* Good, because it unifies API structure
+* Good, because it causes API easier to use
+* Good, because it allows creating request with field on command as IRI
+* Good, because it leaves command field without IRI but with `id`/`code`
+* Bad, because it imposes new abstraction layer for commands
```
CoderMaggie, line 42:
```suggestion
* Good, because it unifies the API structure
* Good, because it makes the API easier to use
* Good, because it allows having `id`/`code` transformed from IRI in the command field
* Bad, because it imposes a new abstraction layer for commands
```

CoderMaggie (commented):
The 🇺🇸  language police at your service 🖖🏼

src/Sylius/Bundle/ApiBundle/Converter/ItemIriToIdentifierConverterInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Converter;
+
+interface ItemIriToIdentifierConverterInterface
+{
+    public function getIdentifier(string $iri): string ;
```
arti0090, line 18:
```suggestion
    public function getIdentifier(string $iri): string;
```

arti0090 (approved):

adr/2021_04_15_using_iri_instead_of_code_id.md
```diff
@@ -0,0 +1,46 @@
+# Using iri instead of code/id in requests
```
GSadee, line 1:
```suggestion
# Using IRI as API resource identifier in requests instead of code/id
```

adr/2021_04_15_using_iri_instead_of_code_id.md
```diff
@@ -0,0 +1,46 @@
...
+
+ApiPlatform provides and recommends using IRI as identifier. That identifier is more clearly than classic id and 
+contain more information. On resources, ApiPlatform handles IRI out of the box.
+During creation our new API in many places we decided to use commands while creating our endpoints. For certain cases
+this solution is more flexible than classic approach, but it doesn't handle transformation IRI to proper `code`/`id`.
+In the past we decided to use `code`/`id` instead of IRI. Now we try to unify API and replace this approach for solution with using IRI.
+Main problem is sent request with field as IRI and creation command without IRI, but with `id`/`code`. 
+It's needed for proper handling behaviour in command handler.
+
+## Considered Options
```
GSadee, line 16: I would be for adding the second option to use ids and codes to this consideration section and describing its pros and cons

adr/2021_04_15_using_iri_instead_of_code_id.md
```diff
@@ -0,0 +1,46 @@
...
+                <argument key="Sylius\Bundle\ApiBundle\Command\Checkout\ChoosePaymentMethod">paymentMethod</argument>
+                <argument key="Sylius\Bundle\ApiBundle\Command\Account\ChangePaymentMethod">paymentMethod</argument>
+                <argument key="NewCommandFQCN">NewCommandFieldName</argument><-- Example -->
+            </argument>
+        </service>
+````
+
+* Good, because it unifies API structure
+* Good, because it causes API easier to use
+* Good, because it allows creating request with field on command as IRI
```
GSadee, line 40:
```suggestion
```

src/Sylius/Bundle/ApiBundle/spec/Map/CommandItemIriArgumentToIdentifierMapSpec.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $this->beConstructedWith(['Sylius\Bundle\ApiBundle\Command\Cart\PickupCart' => 'token']);
+    }
+
+    function it_is_command_item_iri_argument_to_identifier_map(): void
+    {
+        $this->shouldImplement(CommandItemIriArgumentToIdentifierMapInterface::class);
+    }
+
+    function it_get_element(): void
```
GSadee, line 31:
```suggestion
    function it_gets_element(): void
```

src/Sylius/Bundle/ApiBundle/spec/Map/CommandItemIriArgumentToIdentifierMapSpec.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $this->get('Sylius\Bundle\ApiBundle\Command\Cart\PickupCart')->shouldReturn('token');
+    }
+
+    function it_has_element(): void
```
GSadee, line 36:
```suggestion
    function it_has_an_element(): void
```

src/Sylius/Bundle/ApiBundle/spec/Map/CommandItemIriArgumentToIdentifierMapSpec.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $this->has('Sylius\Bundle\ApiBundle\Command\Cart\PickupCart')->shouldReturn(true);
+    }
+
+    function it_has_not_element(): void
```
GSadee, line 41:
```suggestion
    function it_has_no_element(): void
```

GSadee (commented):
Could you list the rest of the commands and fields that should be changed to the new approach?

adr/2021_04_15_using_iri_instead_of_code_id.md
```diff
@@ -0,0 +1,53 @@
...
+API Platform recommends using IRI as an identifier. That identifier gives more clarity than an id because it contains more information - a full endpoint path to the resource and its unique identifier. On resources, API Platform handles IRI out of the box.
+While we were designing Sylius new API, we have decided to use commands in many endpoints.
+In some cases, this solution is more flexible than the default creation approach (as we have control over what we are processing), but it doesn't handle the transformation of IRI to proper `code`/`id`.
+In the past, we have been using `code`/`id` instead of IRI, then we have been using both approaches.
+Now we are trying to unify the new API and replace codes and ids with IRI everywhere.
+The main challenge is the usage of IRI in requests, where we want to have IRI in the request but `id`/`code`in its commands and command handlers.
+
+## Considered Options
+
+### Using `id`/`code` instead of iri
```
GSadee, line 17:
```suggestion
### Using `id`/`code` instead of IRI
```
And in other places

adr/2021_04_15_using_iri_instead_of_code_id.md
```diff
@@ -0,0 +1,53 @@
...
+
+Using `code`/`id` instead of iri is easier implementable, but it is not consistence with most cases.
+ApiPlatform handles it out of the box. Main problem is that for getting fe. `code` we need call "extra" request only for 
+getting code from this object. Iri are available as a collections element or field on certain object. Is many examples
+when we don't need to call next request for getting `code`/`id`.
```
GSadee, line 22:
```suggestion
Using `code`/`id` instead of IRI is easier to implement, but it is not consistent with API Platform and its default behaviour.
```
I don't understand the second part of this section, what additional request are you talking about?

adr/2021_04_15_using_iri_instead_of_code_id.md
```diff
@@ -0,0 +1,53 @@
...
+
+* Good, because it is easier implementable
```
GSadee, line 24:
```suggestion
* Good, because it is easier to implement
```

adr/2021_04_15_using_iri_instead_of_code_id.md
```diff
@@ -0,0 +1,53 @@
...
+* Bad, because it not consistence with others endpoints
```
GSadee, line 25:
```suggestion
* Bad, because it not consistent with others endpoints
```

adr/2021_04_15_using_iri_instead_of_code_id.md
```diff
@@ -0,0 +1,53 @@
+# Using IRI as API resource identifier in requests instead of code/id
```
GSadee, line 1: The name of the file could be also adjusted

adr/2021_04_15_using_iri_as_api_resource_identifier_in_request_instead_of_code_id.md
```diff
@@ -0,0 +1,50 @@
...
+and field name as a field that will be transformed from IRI to `code`/`id`. All definition are registered in
+`Sylius\Bundle\ApiBundle\Map\CommandItemIriArgumentToIdentifierMap` service:
+
+````
+        <service id="Sylius\Bundle\ApiBundle\Map\CommandItemIriArgumentToIdentifierMap">
+            <argument type="collection">
+                <argument key="Sylius\Bundle\ApiBundle\Command\AddProductReview">product</argument>
+                <argument key="Sylius\Bundle\ApiBundle\Command\Checkout\ChoosePaymentMethod">paymentMethod</argument>
+                <argument key="Sylius\Bundle\ApiBundle\Command\Account\ChangePaymentMethod">paymentMethod</argument>
+                <argument key="NewCommandFQCN">NewCommandFieldName</argument><-- Example -->
```
GSadee, line 39:
```suggestion
                <argument key="NewCommandFQCN">newCommandFieldName</argument>
```

GSadee (approved):

src/Sylius/Bundle/ApiBundle/Resources/config/services.xml
```diff
@@ -160,10 +160,24 @@
...
+            <argument type="collection">
+                <argument key="Sylius\Bundle\ApiBundle\Command\AddProductReview">product</argument>
+                <argument key="Sylius\Bundle\ApiBundle\Command\Checkout\ChoosePaymentMethod">paymentMethod</argument>
+                <argument key="Sylius\Bundle\ApiBundle\Command\Account\ChangePaymentMethod">paymentMethod</argument>
+                <argument key="Sylius\Bundle\ApiBundle\Command\RequestResetPasswordToken">locale</argument>
+                <argument key="Sylius\Bundle\ApiBundle\Command\ResendVerificationEmail">locale</argument>
+                <argument key="Sylius\Bundle\ApiBundle\Command\Cart\PickupCart">locale</argument>
+                <argument key="Sylius\Bundle\ApiBundle\Command\Cart\AddItemToCart">productVariant</argument>
+                <argument key="Sylius\Bundle\ApiBundle\Command\Checkout\ChooseShippingMethod">shippingMethod</argument>
+            </argument>
```
lchrusciel, line 173: That's a neat idea, but I'm afraid about this list expandability. One would have to override this service definition, therefore they will not receive changes to this collection that we will do in the later releases. We need some refactoring here. Probably configuration tree would be the best choice. Maybe second class for client-specific mappings?

src/Sylius/Bundle/ApiBundle/Serializer/CommandFieldItemIriToIdentifierDenormalizer.php
```diff
@@ -13,46 +13,53 @@
...
-        $product = $this->iriConverter->getItemFromIri($data['product']);
+        /** @psalm-var class-string $inputClassName */
+        $inputClassName = $this->getInputClassName($context);
+
+        $fieldName = $this->commandItemIriArgumentToIdentifierMap->get($inputClassName);
 
-        $data['product'] = $product->getCode();
+        if (array_key_exists($fieldName, $data)) {
+            $data[$fieldName] = $this->itemIriToIdentifierConverter->getIdentifier($data[$fieldName]);
+        }
```
lchrusciel, line 62: Also, improvement for the second iteration, but what is blocking us from checking if the field consists IRI and execute this logic on all fields like this? This way, we could remove the whole map from the concept

src/Sylius/Bundle/ApiBundle/Tests/Converter/ItemIriToIdentifierConverterTest.php
```diff
@@ -0,0 +1,105 @@
...
+            '_api_item_operation_name' => 'get',
+            '_api_identifiers' => ['id', 'nextId'],
+            'id' => 3,
+            'nextId' => 5
+        ])->shouldBeCalledTimes(1);
+
+        $identifierConverter->convert(['id' => 3, 'nextId' => 5], AddProductReview::class)->willReturn(['3', '5']);
+        $converter = new ItemIriToIdentifierConverter($router->reveal(), $identifierConverter->reveal());
+
+        $this->assertSame('3', $converter->getIdentifier('/users/3/nexts/5'));
```
lchrusciel, line 82:
```suggestion
        $converter->getIdentifier('/users/3/nexts/5');
```

lchrusciel:

PR #12501, opened 2021-04-01
[BUG] Product options null value returns 400 instead of 500

SirDomin:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Product/Model/ProductVariant.php
+++ b/src/Sylius/Component/Product/Model/ProductVariant.php
@@ -108,9 +108,9 @@ class ProductVariant implements ProductVariantInterface
         }
     }
 
-    public function hasOptionValue(ProductOptionValueInterface $optionValue): bool
+    public function hasOptionValue(?ProductOptionValueInterface $optionValue): bool
     {
-        return $this->optionValues->contains($optionValue);
+        return $optionValue === null ? true : $this->optionValues->contains($optionValue);
     }
 
     public function getProduct(): ?ProductInterface
--- a/src/Sylius/Component/Product/Model/ProductVariantInterface.php
+++ b/src/Sylius/Component/Product/Model/ProductVariantInterface.php
@@ -45,7 +45,7 @@ interface ProductVariantInterface extends
 
     public function removeOptionValue(ProductOptionValueInterface $optionValue): void;
 
-    public function hasOptionValue(ProductOptionValueInterface $optionValue): bool;
+    public function hasOptionValue(?ProductOptionValueInterface $optionValue): bool;
 
     public function getProduct(): ?ProductInterface;
 
--- a/tests/Controller/ProductVariantApiTest.php
+++ b/tests/Controller/ProductVariantApiTest.php
@@ -17,7 +17,6 @@ use ApiTestCase\JsonApiTestCase;
 use Sylius\Component\Core\Model\ProductInterface;
 use Sylius\Component\Core\Model\ProductVariantInterface;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\VarDumper\VarDumper;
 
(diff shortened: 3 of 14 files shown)


arti0090 (approved):

tests/Controller/ProductVariantApiTest.php
```diff
@@ -17,6 +17,7 @@
 use Sylius\Component\Core\Model\ProductInterface;
 use Sylius\Component\Core\Model\ProductVariantInterface;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\VarDumper\VarDumper;
```
GSadee, line 20:
```suggestion
```

src/Sylius/Component/Product/Model/ProductVariant.php
```diff
@@ -108,9 +108,9 @@ public function removeOptionValue(ProductOptionValueInterface $optionValue): voi
         }
     }
 
-    public function hasOptionValue(ProductOptionValueInterface $optionValue): bool
+    public function hasOptionValue(?ProductOptionValueInterface $optionValue = null): bool
```
GSadee, line 111:
```suggestion
    public function hasOptionValue(?ProductOptionValueInterface $optionValue): bool
```

src/Sylius/Component/Product/Model/ProductVariantInterface.php
```diff
@@ -45,7 +45,7 @@ public function addOptionValue(ProductOptionValueInterface $optionValue): void;
 
     public function removeOptionValue(ProductOptionValueInterface $optionValue): void;
 
-    public function hasOptionValue(ProductOptionValueInterface $optionValue): bool;
+    public function hasOptionValue(?ProductOptionValueInterface $optionValue): bool;
```
GSadee, line 48: I wonder if this change is not a BC Break, although it expands the accepted argument, wdyt @lchrusciel

src/Sylius/Component/Product/Model/ProductVariant.php
```diff
@@ -108,9 +108,9 @@ public function removeOptionValue(ProductOptionValueInterface $optionValue): voi
         }
     }
 
-    public function hasOptionValue(ProductOptionValueInterface $optionValue): bool
+    public function hasOptionValue(?ProductOptionValueInterface $optionValue): bool
     {
-        return $this->optionValues->contains($optionValue);
+        return $optionValue === null ? true : $this->optionValues->contains($optionValue);
```
lchrusciel, line 113: Why true instead of false?

lchrusciel (approved):

lchrusciel:
Thank you, @SirDomin! :1st_place_medal:

src/Sylius/Component/Product/Model/ProductVariant.php
SirDomin: this is abit tricky, because by returning true we force it ProductVariant to dont allow null as argument and handle it as bad request.

src/Sylius/Component/Product/Model/ProductVariantInterface.php
pamil: 🚨 That's a BC break, all classes implementing this interface have to change their method signature.

src/Sylius/Component/Product/Model/ProductVariant.php
pamil: It is tricky because it feels wrong to check whether a product variant has an option value `null`. I think we should fix it upstream and do not let `null` to be passed to this method in the first place.

PR #12502, opened 2021-04-01
Recalculate order adjustments total after adjustment is added or remo…

kayue:
| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

In the current code (1.8 at least) adjustment total is not being updated after shipping fee is added to order. For example in the code below:

https://github.com/Sylius/Sylius/blob/f486b804e3e460847c90e29698698a8c88357d55/src/Sylius/Component/Core/OrderProcessing/ShippingChargesProcessor.php#L51-L59

Recalculation is designed to happen when `$adjustment->setAmount()` is called, however the Order object is only added after. (in the last line) 

Proposed fixed is to trigger `recalculateAdjustmentsTotal()` whenever adjustment is added (or removed) to order.

### DIFF
--- a/tests/Model/OrderAdjustmentsRecalculationTest.php
+++ b/tests/Model/OrderAdjustmentsRecalculationTest.php
@@ -1,0 +1,118 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Sylius Sp. z o.o.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Tests\Model;
+
+use PHPUnit\Framework\TestCase;
+use Sylius\Component\Core\Model\AdjustmentInterface;
+use Sylius\Component\Order\Model\Order;
+use Sylius\Component\Order\Model\OrderItem;
+use Sylius\Component\Order\Model\OrderItemInterface;
+use Sylius\Component\Order\Model\OrderItemUnit;
+
+final class OrderAdjustmentsRecalculationTest extends TestCase
+{
+    private Order $order;
+    private OrderItemInterface $item;
+
+    protected function setUp(): void
+    {
+        $neutralAdjustment = $this->createAdjustment(AdjustmentInterface::ORDER_PROMOTION_ADJUSTMENT, -150, true);
+
+        $this->order = new Order();
+        $this->item = new OrderItem();
+        $this->item->setUnitPrice(1000);
+        $this->unitNumberOne = new OrderItemUnit($this->item);
+        $this->unitNumberTwo = new OrderItemUnit($this->item);
+        $this->order->addItem($this->item);
+        $this->order->addAdjustment($neutralAdjustment);
+    }
+
+    /** @test */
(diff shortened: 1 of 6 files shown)


Zales0123 (commented):
Generally, I think it's a good idea 👍 2 notes from me:

- we should probably add such a recalculation in every adjustable object (`Order`, `OrderItem` and `OrderItemUnit`)
- I would start with changing the spec for the proper objects (e.g. `it_adds_adjustments_properly` for the `Order`) and call `$this->getAdjustmentsTotal()->shouldReturn(XXX);` at the end

🖖

SirDomin:
may become messed-up.**

execute `git fetch origin && git reset "patch-6"` to update your local branch.
 
Zales0123 (dismissed):

Zales0123:
OR 😄 in fact I'm not 100% sure should we merge it without adding the same feature on `OrderItem`, `OrderItemUnit` and `Shipment` level - currently no adjustment is added to `Order` directly, so we *cannot* be sure it does not break anything, as these change would not be covered by any Behat scenario 💃

tests/Model/OrderAdjustmentsRecalculationTest.php
```diff
@@ -0,0 +1,178 @@
...
+    /** @test */
+    public function it_recalculates_order_properly_while_tax_adjustment_removed(): void
+    {
+        $adjustment = $this->createAdjustment(AdjustmentInterface::TAX_ADJUSTMENT, 90);
+        $this->order->addAdjustment($adjustment);
+        $this->order->removeAdjustment($adjustment);
+
+        $this->assertEquals(2000, $this->order->getTotal());
+        $this->assertEquals(0, $this->order->getAdjustmentsTotal());
+    }
```
jakubtobiasz: I don't see a significant value coming from testing every type of adjustment. Let's create a simple test case where you have some adjustment already, you check if the returned value is correct, then you add 2–3 adjustments (both a negative and a positive) and again check if it's recalculated. We don't need anything more.

At the same time, you don't have such tests for `OrderItem` and `OrderItemUnit`.

jakubtobiasz (changes requested):

tests/Model/OrderAdjustmentsRecalculationTest.php
TheMilek: There are tests for `OrderItem` `OrderItemUnit` as  it is part of order
https://github.com/Sylius/Sylius/pull/12502/files#diff-e1c288738071c3d116f3735eead129fe90f38e7d8fd51e28f7bad8a738f03d0aR69-R78
https://github.com/Sylius/Sylius/pull/12502/files#diff-e1c288738071c3d116f3735eead129fe90f38e7d8fd51e28f7bad8a738f03d0aR39-R46

TheMilek: I would be after modifying this tests by adding adjustments that already applies on order that im creating in SetUp, WDYT?

GSadee (approved):

GSadee:
Thanks, @kayue! :1st_place_medal:

PR #12546, opened 2021-04-19
[Minor] Use explicit bus instead of the default one

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/services/event_handlers.xml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/services/event_handlers.xml
@@ -17,7 +17,7 @@
 >
     <services>
         <service id="Sylius\Bundle\ApiBundle\EventHandler\OrderCompletedHandler">
-            <argument type="service" id="messenger.default_bus" />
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/ApiBundle/Resources/config/services/event_handlers.xml
```diff
@@ -17,7 +17,7 @@
 >
     <services>
         <service id="Sylius\Bundle\ApiBundle\EventHandler\OrderCompletedHandler">
-            <argument type="service" id="messenger.default_bus" />
+            <argument type="service" id="sylius_default.bus" />
```
pamil, line 20: Looks like a good change - just a note that we probably should work on the naming of these buses. It doesn't follow convention (`sylius_default.bus` instead of `sylius.default_bus` or `sylius.bus.default`) and it's also a default bus for API Bundle (and it's defined there).

pamil (approved):

pamil:

PR #12547, opened 2021-04-20
Remove Admin API Bundle from the default Sylius installation

pamil:
See #12469.

`UPGRADE` instructions missing, I'll do them once we get this merged while customising Sylius-Standard to those changes.


tests/Responses/ajax/product/index_response.json
```diff
@@ -25,25 +25,15 @@
                 "code": "MUG_BB",
                 "options": [],
                 "averageRating": 0,
-                "images": [],
-                "_links": {
-                    "self": {
-                        "href": "\/api\/v1\/products\/MUG_BB"
```
lchrusciel, line 31: Isn't it a BC? Do we have any possibility to sustain it? 

On the other hand, this filed should not be hire in the first place

config/packages/security.yaml
```diff
@@ -74,13 +70,6 @@ security:
                 authenticators:
                     - lexik_jwt_authentication.jwt_token_authenticator
 
-        api:
-            pattern: "%sylius.security.api_regex%/.*"
```
lchrusciel, line 78: Is this parameter defined in AdminApi bundle?

tests/Responses/ajax/product/index_response.json
pamil: Without Admin API Bundle it's impossible to generate those URLs. If you have Admin API Bundle installed, it will have those links included.

It feels like it should be mentioned in the upgrade instructions though. Most likely no one uses those AJAX responses in any other way than for autovompletes.

config/packages/security.yaml
pamil: Yes, it is - there's also no need for the firewall if the Admin API Bundle is not enabled.

lchrusciel: I just wanted to ensure, that we are not leaving its definition in the core 👍

lchrusciel (approved):

lchrusciel:

PR #12549, opened 2021-04-20
[Product Variant] Price Calculator fixed to return always int

lruozzi9:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | no
| License         | MIT

This Pull Request will avoid returning null on int only return type allowed function. If getPrice is null for the channel the same exception is thrown.

### DIFF
--- a/src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
+++ b/src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
@@ -25,7 +25,7 @@ final class ProductVariantPriceCalculator implements ProductVariantPricesCalcula
 
         $channelPricing = $productVariant->getChannelPricingForChannel($context['channel']);
 
-        if (null === $channelPricing) {
+        if (null === $channelPricing || $channelPricing->getPrice() === null) {
             $message = sprintf('Channel %s has no price defined for product variant', $context['channel']->getName());
 
             if ($productVariant->getName() !== null) {
@@ -58,6 +58,18 @@ final class ProductVariantPriceCalculator implements ProductVariantPricesCalcula
         }
 
         if (null === $channelPricing->getOriginalPrice()) {
+            if ($channelPricing->getPrice() === null) {
+                $message = sprintf('Channel %s has no price defined for product variant', $context['channel']->getName());
+
+                if ($productVariant->getName() !== null) {
+                    $message .= sprintf(' %s (%s)', $productVariant->getName(), $productVariant->getCode());
+                } else {
+                    $message .= sprintf(' with code %s', $productVariant->getCode());
+                }
+
+                throw new MissingChannelConfigurationException($message);
+            }
+
             return $channelPricing->getPrice();
         }
 
--- a/src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
+++ b/src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
@@ -14,7 +14,6 @@ declare(strict_types=1);
 namespace Sylius\Component\Core\Calculator;
 
 use Sylius\Component\Core\Exception\MissingChannelConfigurationException;
-use Sylius\Component\Core\Model\ChannelInterface;
 use Sylius\Component\Core\Model\ProductVariantInterface;
 use Webmozart\Assert\Assert;
 
@@ -27,7 +26,7 @@ final class ProductVariantPriceCalculator implements ProductVariantPricesCalcula
         $channelPricing = $productVariant->getChannelPricingForChannel($context['channel']);
 
         if (null === $channelPricing || $channelPricing->getPrice() === null) {
-            throw new MissingChannelConfigurationException(MissingChannelConfigurationException::getMissingChannelProductVariantPriceMessage($context['channel'], $productVariant));
+            throw MissingChannelConfigurationException::createForProductVariantChannelPricing($productVariant, $context['channel']);
         }
 
         return $channelPricing->getPrice();
@@ -43,7 +42,7 @@ final class ProductVariantPriceCalculator implements ProductVariantPricesCalcula
         $channelPricing = $productVariant->getChannelPricingForChannel($context['channel']);
 
         if (null === $channelPricing) {
-            throw new MissingChannelConfigurationException(MissingChannelConfigurationException::getMissingChannelProductVariantPriceMessage($context['channel'], $productVariant));
+            throw MissingChannelConfigurationException::createForProductVariantChannelPricing($productVariant, $context['channel']);
         }
 
         if (null !== $channelPricing->getOriginalPrice()) {
@@ -54,6 +53,6 @@ final class ProductVariantPriceCalculator implements ProductVariantPricesCalcula
             return $channelPricing->getPrice();
         }
 
-        throw new MissingChannelConfigurationException(MissingChannelConfigurationException::getMissingChannelProductVariantPriceMessage($context['channel'], $productVariant));
+        throw MissingChannelConfigurationException::createForProductVariantChannelPricing($productVariant, $context['channel']);
     }
 }
--- a/src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
+++ b/src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sylius\Component\Core\Calculator;
 
 use Sylius\Component\Core\Exception\MissingChannelConfigurationException;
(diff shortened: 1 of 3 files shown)


vvasiloi:
Hey @lruozzi9, thank you for your contribution!
Could you add some [tests ](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Component/Core/spec/Calculator/ProductVariantPriceCalculatorSpec.php)to cover these changes?

lruozzi9:
Sure 👍🏻

vvasiloi:
@lruozzi9 Great! Thank you!
WDYT about changing the conditions to `null !== X` and returning early, then throw an exception at the end.
Also, the logic to construct the error message is duplicated.
Maybe extract it to a factory method on the exception class, something like `createForVariant`?

lruozzi9:
I agree, it is more readable! I will change it.

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
```diff
@@ -50,29 +43,30 @@ public function calculateOriginal(ProductVariantInterface $productVariant, array
...
         if (null === $channelPricing) {
-            throw new MissingChannelConfigurationException(sprintf(
-                'Channel %s has no price defined for product variant %s',
-                $context['channel']->getName(),
-                $productVariant->getName()
-            ));
+            throw new MissingChannelConfigurationException($this->getMissingChannelProductVariantPriceMessage($context['channel'], $productVariant));
         }
 
         if (null === $channelPricing->getOriginalPrice()) {
```
vvasiloi, line 49: This can also return early.
```suggestion
        if (null !== $channelPricing->getOriginalPrice()) {
            return $channelPricing->getOriginalPrice();
        }
```

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
```diff
@@ -50,29 +43,30 @@ public function calculateOriginal(ProductVariantInterface $productVariant, array
...
-                $message = sprintf('Channel %s has no price defined for product variant', $context['channel']->getName());
-
-                if ($productVariant->getName() !== null) {
-                    $message .= sprintf(' %s (%s)', $productVariant->getName(), $productVariant->getCode());
-                } else {
-                    $message .= sprintf(' with code %s', $productVariant->getCode());
-                }
-
-                throw new MissingChannelConfigurationException($message);
+            if ($channelPricing->getPrice() !== null) {
```
vvasiloi, line 50: Then this will move one level up.

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
```diff
@@ -50,29 +43,30 @@ public function calculateOriginal(ProductVariantInterface $productVariant, array
...
+                return $channelPricing->getPrice();
             }
 
-            return $channelPricing->getPrice();
+            throw new MissingChannelConfigurationException($this->getMissingChannelProductVariantPriceMessage($context['channel'], $productVariant));
```
vvasiloi, line 54: And this will replace the return at the end of the method.

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
```diff
@@ -50,29 +43,30 @@ public function calculateOriginal(ProductVariantInterface $productVariant, array
...
         }
 
         return $channelPricing->getOriginalPrice();
     }
+
+    private function getMissingChannelProductVariantPriceMessage(ChannelInterface $channel, ProductVariantInterface $productVariant): string
```
vvasiloi, line 60: Can you move it to a static factory method inside the MissingChannelConfigurationException?
So it can also be used downstream in custom calculators.
It should look like this:
```php
throw new MissingChannelConfigurationException::createForProductVariantAndChannel($productVariant, $channel);
```
There's also a practice of creating separate factory classes, like `MissingChannelConfigurationExceptionFactory` for these methods, but it's probably too much for this case.

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
```diff
@@ -50,29 +43,30 @@ public function calculateOriginal(ProductVariantInterface $productVariant, array
...
+    {
+        $message = sprintf('Channel %s has no price defined for product variant', $channel->getName());
+
+        if ($productVariant->getName() !== null) {
```
vvasiloi, line 64: You can remove the `else` here and return early, like:
```suggestion
        if ($productVariant->getName() !== null) {
            return $message . sprintf(' %s (%s)', $productVariant->getName(), $productVariant->getCode());
        }

        return $message . sprintf(' with code %s', $productVariant->getCode());
```

vvasiloi (changes requested):
I would like you to make a few more changes if you don't mind.
I promise it's the last one from me.

lruozzi9:
Hey @vvasiloi! Sorry, I was a little busy these days! I will check your comments!

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
vvasiloi: @lruozzi9 why you didn't do it like in the example I gave?

vvasiloi: I see there's a mistake in my example, it should not have the `new ` keyword.

lruozzi9: Sorry, I misunderstood the comment. If I have to make a method that instantiates the class inside itself then I like most the idea of having an exception factory as you wrote.

src/Sylius/Component/Core/Exception/MissingChannelConfigurationException.php
```diff
@@ -13,10 +13,23 @@
...
     }
+
+    public static function getMissingChannelProductVariantPriceMessage(ChannelInterface $channel, ProductVariantInterface $productVariant): string
+    {
+        $message = sprintf('Channel %s has no price defined for product variant', $channel->getName());
+        if ($productVariant->getName() !== null) {
+            return $message . sprintf(' %s (%s)', $productVariant->getName(), $productVariant->getCode());
+        }
+
+        return $message . sprintf(' with code %s', $productVariant->getCode());
```
vvasiloi, line 33: I noticed only now that the message is not very appropriate and it can be improved, which will also reduce the amount of code here.

```suggestion
        return new self(sprintf('Product variant %s has no price defined for channel %s', $productVariant->getDescriptor(), $channel->getName()))
```

vvasiloi (changes requested):

src/Sylius/Component/Core/Calculator/ProductVariantPriceCalculator.php
vvasiloi: Unfortunately I have to ask you to make some changes again because of this misunderstanding.
There's also a small improvement regarding the exception message I spotted this time.

lruozzi9: NP, what do you think of this Factory class https://github.com/Sylius/Sylius/pull/12549/commits/089788f52e5a34352ec7a3c69403ca64e0901bcb?

vvasiloi: If you want to go with an exception factory, then you can just apply my suggestions first then extract the method to a factory class and change the `self` with the exception class.
In case you need some examples, you can find a lot here https://github.com/nelmio/alice/tree/master/src/Throwable/Exception.

lruozzi9: I couldn't reapply your comments because I had already made some changes. Now they should all be applied.

lruozzi9:
I have a doubt about the location of the factory exception folder. Perhaps it is more correct to create a Factory folder inside Exception?

vvasiloi:
There's really no right or wrong here, it's all a matter of opinion, especially because there's no precedent for it in Sylius' codebase.
Let's have the factory in the same namespace as the exception class, that's what I saw others practice and it seems convenient.

vvasiloi (approved):

vvasiloi:
@lruozzi9 I just noticed that you target branch 1.8 🤦‍♂️. Can you rebase your branch and change the target to 1.10?

lchrusciel (approved):

GSadee (approved):

GSadee:
Thanks, Lorenzo! :tada:

PR #12555, opened 2021-04-21
Fix misleading signing in the admin panel feature description

Mateuszry:
| Q               | A
| --------------- | -----
| Branch?         | 1.7
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

The test didn't check the described scenario, because it contained incorrect customer login data, and the scenario assumes logging in with correct customer data, but to administration panel instead of customer panel in shop.

### DIFF
--- a/features/admin/signing_in_to_administration_panel.feature
+++ b/features/admin/signing_in_to_administration_panel.feature
@@ -1,8 +1,8 @@
 @administrator_login
-Feature: Sign in to the store
-    In order to view my orders
-    As a Visitor
-    I want to be able to log in to the store
+Feature: Signing in to the administration panel
+    In order to manage the store
+    As an Administrator
+    I want to be able to log in to the administration panel
 
     Background:
         Given the store operates on a single channel in "United States"
@@ -29,7 +29,7 @@ Feature: Sign in to the store
     @ui @api
     Scenario: Sign in using customer account
         When I want to log in
-        And I specify the username as "bear@example.com"
+        And I specify the username as "ted@example.com"
         And I specify the password as "bear"
         And I log in
         Then I should be notified about bad credentials


features/admin/sign_in.feature
```diff
@@ -1,8 +1,8 @@
 @administrator_login
-Feature: Sign in to the store
-    In order to view my orders
-    As a Visitor
-    I want to be able to log in to the store
+Feature: Sign in to the administration
```
loic425, line 2: Maybe "admin panel" instead.

Mateuszry: Changed to `administration panel`, because this name already appears in other tests

loic425 (dismissed):

features/admin/sign_in.feature
```diff
@@ -1,8 +1,8 @@
 @administrator_login
-Feature: Sign in to the store
-    In order to view my orders
-    As a Visitor
-    I want to be able to log in to the store
+Feature: Sign in to the administration panel
+    In order to access administration panel
```
Zales0123, line 3: We should avoid descriptions `In order to X... I want to X` 💃 Accessing the administrator panel is a way to fulfil some goal. Maybe sth like `In order to be able to manage my store`?

Zales0123 (commented):
👍 but it should be opened to branch `1.8` as the lowest supported right now. Thanks! 🚀

features/admin/sign_in.feature
Mateuszry: You should update contributing documentation then, because everywhere you mention current version as 1.7 https://docs.sylius.com/en/latest/book/contributing/code/patches.html#choose-the-right-base-branch
Should I open a new pull request to 1.8 now?

If we are talking about feature description I just tried to follow your standards, because the old description didn't describe the sign in to the administration panel, but was a copy of the sign in to the customer panel (https://github.com/Sylius/Sylius/blob/master/features/account/signing_in.feature)

jakubtobiasz (approved):

TheMilek (approved):

TheMilek:
Thank you, @Mateuszry! :tada:

PR #12561, opened 2021-04-23
Bugfix | Don't declare `default_bus` for Messenger & always use Sylius Bus

stloyd:
| Q               | A
| --------------- | -----
| Branch?         | 1.10-dev, but possibly lower as well
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Having that `framework.messenger.default_bus` being set, makes impossible to declared own one & forces to use `sylius_default.bus` as bus in our project.

### DIFF
--- a/.github/workflows/application.yml
+++ b/.github/workflows/application.yml
@@ -86,17 +86,17 @@ jobs:
 
             -
                 name: Validate Doctrine mapping
-                run: bin/console doctrine:schema:validate --skip-sync -vvv
+                run: bin/console doctrine:schema:validate --skip-sync -vvv --env=test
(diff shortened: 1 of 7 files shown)


stloyd:
Okey, seems that one change is wrong, second is correct. Will adjust the PR during weekend.

stloyd:
@pamil @GSadee The failed test is due too timeout, all other passed. Can you check & give me feedback, thanks :)

stloyd:
@pamil @lchrusciel Guys as you soon want to release `1.10` officially, I guess this is must have as it otherwise will broke any project using already messenger.

PR #12570, opened 2021-04-27
Bugfix | Don't use "json" native column in `Adjustment` entity

stloyd:
| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no(?)
| Deprecations?   | no
| License         | MIT

Using: `MariabDB:10.4.12`

Error while running migration:
```
In DBALException.php line 282:

Unknown column type " json" requested. Any Doctrine type that you use has t
o be registered with \Doctrine\DBAL\Types\Type::addType(). You can get a li  
st of all the known types with \Doctrine\DBAL\Types\Type::getTypesMap(). If  
this error occurs during database introspection then you might have forgot  
ten to register all database types for a Doctrine Type. Use AbstractPlatfor  
m#registerDoctrineTypeMapping() or have your custom types implement Type#ge
tMappedDatabaseTypes(). If the type name is empty you might have a problem   
with the cache or forgot some mapping information.    
```

Setting this doesn't help ;)
```yaml
doctrine:
    dbal:
        types:
            json: Doctrine\DBAL\Types\JsonType

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Migrations/Version20201204071301.php
+++ b/src/Sylius/Bundle/CoreBundle/Migrations/Version20201204071301.php
@@ -30,7 +30,7 @@ final class Version20201204071301 extends AbstractMigration
             return;
         }
 
-        $this->addSql('ALTER TABLE sylius_adjustment ADD details JSON NOT NULL');
+        $this->addSql('ALTER TABLE sylius_adjustment ADD details LONGTEXT NOT NULL COMMENT \'(DC2Type:json_array)\'');
     }
 
     public function down(Schema $schema): void
--- a/src/Sylius/Bundle/OrderBundle/Resources/config/doctrine/model/Adjustment.orm.xml
+++ b/src/Sylius/Bundle/OrderBundle/Resources/config/doctrine/model/Adjustment.orm.xml
@@ -28,7 +28,7 @@
         <field name="neutral" column="is_neutral" type="boolean" />
         <field name="locked" column="is_locked" type="boolean" />
         <field name="originCode" column="origin_code" type="string" nullable="true" />
-        <field name="details" type="json" />
+        <field name="details" column="details" type="json_array" />
 
         <field name="createdAt" column="created_at" type="datetime">
             <gedmo:timestampable on="create"/>

### CODE REVIEW

**Comment by vvasiloi** [comment]
@stloyd here's how I tested it and the result:
```bash
composer create-project sylius/sylius-standard
cd sylius-standard
docker run -p 127.0.0.1:33060:3306  --name sylius-mariadb -e MYSQL_ROOT_PASSWORD=sylius -e MYSQL_USER=sylius -e MYSQL_PASSWORD=sylius -e MYSQL_DATABASE=sylius_dev -d mariadb:10.4.12
echo "DATABASE_URL=mysql://sylius:sylius@127.0.0.1:33060/sylius_%kernel.environment%?serverVersion=mariadb-10.4.12" > .env.local
bin/console sylius:install --no-interaction
bin/console doctrine:migrations:list | grep 20201204071301

| Sylius\Bundle\CoreBundle\Migrations\Version20201204071301     | migrated | 2021-04-28 21:20:28 | 0.608s         | Add details to adjustment                            |
```


And the column type is `longtext`.
```bash
bin/console doctrine:query:sql "select data_type from information_schema.columns where table_schema = 'sylius_dev' and table_name = 'sylius_adjustment' and column_name = 'details'"
```

```php
array(1) {
  [0]=>
  array(1) {
    ["data_type"]=>
    string(8) "longtext"
  }
}
```



According to the documentation Doctrine will fallback to a text type if the engine doesn't support json natively: https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#json

I don't know if you set the server version, but for MariaDB it's required according to documentation:
https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#automatic-platform-version-detection

> If you are running a MariaDB database, you should prefix the serverVersion with mariadb- (ex: mariadb-10.2.12).

stloyd:
@vvasiloi We have `?serverVersion=mariadb-10.4.12` in our DSN url for Doctrine.

Problem is that all you mentioned is Doctrine fallback, which means that if you have `@Column(type="json")` it will fallback, to what you quote.

Our issue is that this migration contains `JSON` as **database** type, not Doctrine, thus it fails for us (tested & with `LONGTEXT` works like charm).

vvasiloi:
@stloyd the steps I followed also execute that migration and it still works. Did I miss something? Can you describe a scenario to reproduce it?

oleg-andreyev:
jfyi https://mariadb.com/kb/en/json-data-type/

> The JSON alias was added in MariaDB 10.2.7. This was done to make it possible to use JSON columns in statement based replication from MySQL to MariaDB and to make it possible for MariaDB to read mysqldumps from MySQL.

so it has nothing to do with MariaDB itself.

PR #12607, opened 2021-05-06
Maintain | Exclude tests, specs & behat from zip archives

stloyd:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no*
| Deprecations?   | no
| License         | MIT

Please read for Symfony decision why this is good idea: https://github.com/symfony/symfony/pull/33579

This also improves our 🌍 impact as developers cause smaller zips, mean less traffic, which reduces can a little reduce amount of pollution 🏭  !

Current `master` | My PR
------------ | -------------
Zip ~32M | Zip ~18M

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,14 @@
+/adr export-ignore
+/docs export-ignore
+/features export-ignore
+/phpspec.yml.dist export-ignore
+/phpstan.neon.dist export-ignore
+/phpunit.xml.dist export-ignore
+/psalm.xml export-ignore
+/src/Sylius/Behat export-ignore
+/tests export-ignore
+/.env export-ignore
+/.env.test export-ignore
+/.env.test_cached export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
--- a/src/Sylius/Bundle/AddressingBundle/.gitattributes
+++ b/src/Sylius/Bundle/AddressingBundle/.gitattributes
@@ -1,0 +1,9 @@
+/Behat export-ignore
+/fixtures export-ignore
+/spec export-ignore
+/test export-ignore
+/Tests export-ignore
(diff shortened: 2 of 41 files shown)


oallain:
Hello,

Thanks for that ❤️ 
Have you been able to do a benchmark to see the benefit of this change?

oallain:
> @oallain
> 
> Current `master`	My PR
> Zip ~32M	Zip ~18M
> So I guess that reducing size by almost half is quite good result

Yes, this is a good result 💪🏼

oallain (approved):

src/Sylius/Bundle/AddressingBundle/.gitattributes
```diff
@@ -0,0 +1,9 @@
+/Behat export-ignore
```
pamil, line 1: I think this directory exists only in ApiBundle.

src/Sylius/Bundle/AddressingBundle/.gitattributes
```diff
@@ -0,0 +1,9 @@
...
+/fixtures export-ignore
```
pamil, line 2: Not sure about this one, but I'm not sure if any bundle has this directory.

.gitattributes
```diff
@@ -0,0 +1,14 @@
+/adr export-ignore
+/docs export-ignore
+/features export-ignore
+/phpspec.yml.dist export-ignore
+/phpstan.neon.dist export-ignore
+/phpunit.xml.dist export-ignore
+/psalm.xml export-ignore
+/src/Sylius/Behat export-ignore
```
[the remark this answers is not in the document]
victor-v-bv: Just run composer install with `--prefer-source`.
Perhaps a note somewhere will help.

PR #12622, opened 2021-05-10
Refactor UpdatePaymentStateExtension 

Prometee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

The actual `UpdatePaymentStateExtension` is trying to please some different use cases for specific gateways, but it doesn't comply with all gateways.
If you get back to Payum library and read how a `Payment` entity is managed, then you will found that this extension : [`Payum\Core\Extension\StorageExtension`](https://github.com/Payum/Payum/blob/master/src/Payum/Core/Extension/StorageExtension.php)  is simply building a list of entities to update  `onPreExecute` and then `onPostExecute` of the first Payum `Request` (meaning after all `Request` have been done) the `Payment` entities will be updated.

This PR allow to save all `Payment` entities `onPreExecute` and then `onPostExecute` will decide to trigger the state machine if it's required. This way all previous gateways which needed to trigger the state machine will be working as expected, but now all gateways with specific cases won't need to make extra code to handle it.

It also allows Notify safe and unsafe one to be able to save the updated `Payment` to the database.

### DIFF
--- a/src/Sylius/Bundle/PayumBundle/DependencyInjection/Compiler/PayumStoragePaymentAliasPass.php
+++ b/src/Sylius/Bundle/PayumBundle/DependencyInjection/Compiler/PayumStoragePaymentAliasPass.php
@@ -1,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\PayumBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+
+final class PayumStoragePaymentAliasPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
+        $paymentClass = $container->getParameter('sylius.model.payment.class');
+        $serviceIds = $container->findTaggedServiceIds('payum.storage');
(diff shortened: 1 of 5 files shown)


src/Sylius/Bundle/PayumBundle/DependencyInjection/Compiler/PayumStoragePaymentAliasPass.php
```diff
@@ -0,0 +1,38 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\PayumBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Definition;
+
+final class PayumStoragePaymentAliasPass implements CompilerPassInterface
```
Prometee, line 11: This `CompilerPass` is needed here because the service id will change depending on the `Payment`'s fqdn.
So it will search for the right service and add an alias to it named `sylius.payum.storage.payment`

Prometee:
Ping @SirDomin talks about Payum during the SyliusCon after party 😉

TheMilek:
Hi!

I'm closing this PR after some private consultation 😅

PR #12648, opened 2021-05-18
[Api] Prevent product more than in stock

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Validator/Constraints/AddingEligibleProductVariantToCartValidator.php
+++ b/src/Sylius/Bundle/ApiBundle/Validator/Constraints/AddingEligibleProductVariantToCartValidator.php
@@ -15,7 +15,6 @@ namespace Sylius\Bundle\ApiBundle\Validator\Constraints;
 
 use Sylius\Bundle\ApiBundle\Command\Cart\AddItemToCart;
 use Sylius\Component\Core\Model\OrderInterface;
-use Sylius\Component\Core\Model\OrderItemInterface;
 use Sylius\Component\Core\Model\ProductInterface;
 use Sylius\Component\Core\Model\ProductVariantInterface;
 use Sylius\Component\Core\Repository\OrderRepositoryInterface;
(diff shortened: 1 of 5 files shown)


src/Sylius/Bundle/ApiBundle/Validator/Constraints/AddingEligibleProductVariantToCartValidator.php
```diff
@@ -85,7 +86,31 @@ public function validate($value, Constraint $constraint)
...
+
+                /** @var ProductVariantInterface */
+                $variant = $orderItem->getVariant();
+
+                $code = $variant->getCode();
+
+                if ($value->productVariantCode = $code) {
+                    ++$sameOrderItemsCounter;
+                }
+            }
```
AdamKasp, line 110: I am wondering about the performance of this solution. In cart, we can have a really big amount of item units, and every addition to the cart will iterate over the entire cart...

src/Sylius/Bundle/ApiBundle/Validator/Constraints/AddingEligibleProductVariantToCartValidator.php
```diff
@@ -85,7 +86,31 @@ public function validate($value, Constraint $constraint)
...
+        }
+
+        if (!$this->availabilityChecker->isStockSufficient($productVariant, $value->quantity + $sameOrderItemsCounter)) {
```
GSadee, line 113: You can create a private method for providing a quantity of given product in the cart, sth like:
```
    private function getExistingCartItemQuantityFromCart(OrderInterface $cart, ProductVariantInterface $productVariant): int
    {
        foreach ($cart->getItems() as $existingCartItem) {
            if ($existingCartItem-> getVariant()->equals($productVariant)) {
                return $existingCartItem->getQuantity();
            }
        }

        return 0;
    }
```
and then
```suggestion
        if (!$this->availabilityChecker->isStockSufficient($productVariant, $value->quantity + $this->getExistingCartItemQuantityFromCart($cart, $productVariant))) {
```

src/Sylius/Bundle/ApiBundle/Validator/Constraints/AddingEligibleProductVariantToCartValidator.php
```diff
@@ -107,4 +111,15 @@ public function validate($value, Constraint $constraint)
             );
         }
     }
+
+    private function getExistingCartItemQuantityFromCart(OrderInterface $cart, ProductVariantInterface $productVariant): int
+    {
+        foreach ($cart->getItems() as $existingCartItem) {
```
Zales0123, line 117: I'm wondering, maybe we can use `OrderInterface::getItemUnitsByVariant($variant)` there? After calling `->count()` method we would have the number of item units for a specific variant - which indeed is a _cart item quantity_ 💃 Maybe we would even be able to resign from the additional private function:

```php
if (!$this-&gt;availabilityChecker-&gt;isStockSufficient(
    $productVariant,
    $value->quantity + $cart->getItemUnitsByVariant($variant)->count()
)) {
```

src/Sylius/Bundle/ApiBundle/spec/Validator/Constraints/AddingEligibleProductVariantToCartValidatorSpec.php
```diff
@@ -131,21 +134,91 @@ function it_adds_violation_if_product_variant_is_disabled(
...
+
+        $executionContext
+            ->addViolation('sylius.product_variant.not_sufficient', ['%productVariantCode%' => 'productVariantCode'])
+            ->shouldBeCalled()
+        ;
+
+        $this->validate(
+            $command,
+            new AddingEligibleProductVariantToCart()
+        );
```
Zales0123, line 183: Would fit in one line 🚀

Zales0123 (approved):
Btw, this spec file is absurdly long 😄 We should definitely think about some validator simplification, as _maybe_ it does a little bit too much 🐎

Zales0123:

PR #12649, opened 2021-05-19
[API][Product] Expose default variant on product show and index

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Serializer/ProductNormalizer.php
+++ b/src/Sylius/Bundle/ApiBundle/Serializer/ProductNormalizer.php
@@ -40,7 +40,8 @@ final class ProductNormalizer implements ContextAwareNormalizerInterface, Normal
         $context[self::ALREADY_CALLED] = true;
 
         $data = $this->normalizer->normalize($object, $format, $context);
-        $data['defaultVariant'] = $this->iriConverter->getIriFromItem($this->defaultProductVariantResolver->getVariant($object));
+        $variant = $this->defaultProductVariantResolver->getVariant($object);
+        $data['defaultVariant'] = $variant === null ? null : $this->iriConverter->getIriFromItem($variant);
 
         return $data;
     }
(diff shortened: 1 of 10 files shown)


src/Sylius/Bundle/ApiBundle/Serializer/ProductNormalizer.php
```diff
@@ -0,0 +1,57 @@
...
+    ) {
+        $this->defaultProductVariantResolver = $defaultProductVariantResolver;
+        $this->iriConverter = $iriConverter;
+    }
+
+    public function normalize($object, $format = null, array $context = [])
+    {
+        Assert::isInstanceOf($object, ProductInterface::class);
+
+        $context[self::ALREADY_CALLED] = true;
```
lchrusciel, line 40: Perhaps we should add assert of early return in case this class is already called indeed?

src/Sylius/Bundle/ApiBundle/Serializer/ProductNormalizer.php
```diff
@@ -0,0 +1,57 @@
...
+        $data = $this->normalizer->normalize($object, $format, $context);
+        $variant = $this->defaultProductVariantResolver->getVariant($object);
+        $data['defaultVariant'] = $variant === null ? null : $this->iriConverter->getIriFromItem($variant);
+
+        return $data;
+    }
+
+    public function supportsNormalization($data, $format = null, $context = []): bool
+    {
+        if (isset($context[self::ALREADY_CALLED])) {
```
lchrusciel, line 51: I would add one more spec for testing of this logic

src/Sylius/Bundle/ApiBundle/Serializer/ProductNormalizer.php
GSadee: IMO there is no need, because this check is in `supportsNormalization` method, but I can add it to be in 100% secured

lchrusciel (approved):

PR #12654, opened 2021-05-20
[Api] Check stock sufficient while cart update

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/spec/Validator/Constraints/ChangingItemQuantityInCartValidatorSpec.php
+++ b/src/Sylius/Bundle/ApiBundle/spec/Validator/Constraints/ChangingItemQuantityInCartValidatorSpec.php
@@ -1,0 +1,282 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace spec\Sylius\Bundle\ApiBundle\Validator\Constraints;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\Bundle\ApiBundle\Command\Cart\ChangeItemQuantityInCart;
+use Sylius\Bundle\ApiBundle\Command\Checkout\CompleteOrder;
+use Sylius\Bundle\ApiBundle\Validator\Constraints\AddingEligibleProductVariantToCart;
+use Sylius\Bundle\ApiBundle\Validator\Constraints\ChangingItemQuantityInCart;
+use Sylius\Component\Core\Model\ChannelInterface;
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Model\OrderItemInterface;
+use Sylius\Component\Core\Model\ProductInterface;
+use Sylius\Component\Core\Model\ProductVariantInterface;
+use Sylius\Component\Core\Repository\OrderRepositoryInterface;
+use Sylius\Component\Inventory\Checker\AvailabilityCheckerInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidatorInterface;
+use Symfony\Component\Validator\Context\ExecutionContextInterface;
+
+final class ChangingItemQuantityInCartValidatorSpec extends ObjectBehavior
+{
+    function let(
+        RepositoryInterface $orderItemRepository,
+        OrderRepositoryInterface $orderRepository,
+        AvailabilityCheckerInterface $availabilityChecker
+    ): void {
+        $this->beConstructedWith($orderItemRepository, $orderRepository, $availabilityChecker);
+    }
+
+    function it_is_a_constraint_validator(): void
+    {
+        $this->shouldImplement(ConstraintValidatorInterface::class);
+    }
+
+    function it_throws_an_exception_if_value_is_not_an_instance_of_change_item_quantity_in_cart(): void
+    {
+        $this
+            ->shouldThrow(\InvalidArgumentException::class)
+            ->during('validate', [new CompleteOrder(), new AddingEligibleProductVariantToCart()]);
+    }
+
+    function it_throws_an_exception_if_constraint_is_not_an_instance_of_changing_item_quantity_in_cart(): void
+    {
+        $this
+            ->shouldThrow(\InvalidArgumentException::class)
+            ->during('validate', [
+                new ChangeItemQuantityInCart( 2),
+                new class() extends Constraint {}
+            ])
+        ;
+    }
+
+    function it_adds_violation_if_product_variant_does_not_exist(
+        RepositoryInterface $orderItemRepository,
+        ExecutionContextInterface $executionContext,
+        OrderItemInterface $orderItem
+    ): void {
+        $this->initialize($executionContext);
+
+        $orderItemRepository->findOneBy(['id' => '11'])->willReturn($orderItem);
+        $orderItem->getVariant()->willReturn(null);
+        $orderItem->getVariantName()->willReturn('MacPro');
+
+        $executionContext
+            ->addViolation('sylius.product_variant.with_name_not_exist', ['%productVariantName%' => 'MacPro'])
+            ->shouldBeCalled()
+        ;
+
+        $this->validate(
+            ChangeItemQuantityInCart::createFromData('token', '11', 2),
+            new ChangingItemQuantityInCart()
+        );
+    }
+
+    function it_adds_violation_if_product_is_disabled(
+        RepositoryInterface $orderItemRepository,
+        ExecutionContextInterface $executionContext,
+        OrderItemInterface $orderItem,
+        ProductVariantInterface $productVariant,
+        ProductInterface $product
+    ): void {
+        $this->initialize($executionContext);
+
+        $orderItemRepository->findOneBy(['id' => '11'])->willReturn($orderItem);
+        $orderItem->getVariant()->willReturn($productVariant);
+        $orderItem->getVariantName()->willReturn('Variant Name');
+
+        $productVariant->getProduct()->willReturn($product);
+        $productVariant->getCode()->willReturn('VARIANT_CODE');
+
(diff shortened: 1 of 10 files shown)


src/Sylius/Bundle/ApiBundle/Validator/Constraints/EligibleProductVariantToCart.php
```diff
@@ -16,7 +16,7 @@
 use Symfony\Component\Validator\Constraint;
 
 /** @experimental */
-final class AddingEligibleProductVariantToCart extends Constraint
+final class EligibleProductVariantToCart extends Constraint
```
GSadee, line 19:
```suggestion
final class EligibleProductVariantInCart extends Constraint
```

You should mention this change in UPGRADE file as this constraint has been added in `1.9`

src/Sylius/Bundle/ApiBundle/Validator/Constraints/EligibleProductVariantToCartValidator.php
```diff
@@ -14,59 +14,74 @@
...
 use Sylius\Component\Inventory\Checker\AvailabilityCheckerInterface;
 use Sylius\Component\Product\Repository\ProductVariantRepositoryInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\ConstraintValidator;
 use Webmozart\Assert\Assert;
 
 /** @experimental */
-final class AddingEligibleProductVariantToCartValidator extends ConstraintValidator
+final class EligibleProductVariantToCartValidator extends ConstraintValidator
```
GSadee, line 31:
```suggestion
final class EligibleProductVariantInCartValidator extends ConstraintValidator
```

You should mention this change in UPGRADE file as this service has been added in `1.9`

src/Sylius/Bundle/ApiBundle/Validator/Constraints/EligibleProductVariantToCartValidator.php
```diff
@@ -14,59 +14,74 @@
...
         $this->orderRepository = $orderRepository;
         $this->availabilityChecker = $availabilityChecker;
     }
 
     public function validate($value, Constraint $constraint)
     {
-        Assert::isInstanceOf($value, AddItemToCart::class);
+        if (!$value instanceof ChangeItemQuantityInCart && !$value instanceof AddItemToCart) {
+            throw new \InvalidArgumentException;
+        }
```
GSadee, line 67:
```suggestion
        Assert::isInstanceOf($value, [AddItemToCart::class, ChangeItemQuantityInCart::class]);
```

src/Sylius/Bundle/ApiBundle/Validator/Constraints/EligibleProductVariantToCartValidator.php
```diff
@@ -107,4 +122,23 @@ public function validate($value, Constraint $constraint)
             );
         }
     }
+
+    private function assignProductVariantAndProductVariantCode($value): void
```
GSadee, line 126: I wonder if it would be better to add the second validator for `ChangeItemQuantityInCart` command 🤔

src/Sylius/Bundle/ApiBundle/Validator/Constraints/EligibleProductVariantToCartValidator.php
Tomanhez: This service was added few days ago
Screenshot 2021-05-20 at 12 35 46

src/Sylius/Bundle/ApiBundle/Validator/Constraints/EligibleProductVariantToCartValidator.php
Tomanhez: I thought about it, but IMO both validators are responsible for validation adding items with quantity. Is easier remember about one validator if would be something to fix. Maybe I should to add a trait or not final class with related common part.

src/Sylius/Bundle/ApiBundle/Validator/Constraints/EligibleProductVariantToCartValidator.php
lchrusciel: It is there: https://github.com/Sylius/Sylius/blob/1.9/src/Sylius/Bundle/ApiBundle/Validator/Constraints/AddingEligibleProductVariantToCartValidator.php, but not tagged yet https://github.com/Sylius/Sylius/blob/v1.9.3/src/Sylius/Bundle/ApiBundle/Validator/Constraints/AddingEligibleProductVariantToCartValidator.php

src/Sylius/Bundle/ApiBundle/Validator/Constraints/EligibleProductVariantToCartValidator.php
```diff
@@ -14,59 +14,74 @@
...
     private $orderRepository;
 
     /** @var AvailabilityCheckerInterface  */
     private $availabilityChecker;
 
+    /** @var ProductVariantInterface|null */
+    private $productVariant;
+
+    /** @var string */
+    private $productVariantCode = '';
```
lchrusciel, line 49: We should try to avoid storing a state in class fields and defining not meaningful default values. I don't see any valid reason to extract this field. If you need this `assignProductVariantAndProductVariantCode` function, it would be better to return an array with two fields and assign them instantly to two variables. A better solution would be to even split this method by half and execute similar logic twice. Another step forward would be to add VO with both values. But the best and simplest solution would be to unpack this function into the main function flow and just forget about this new function

src/Sylius/Bundle/ApiBundle/Validator/Constraints/EligibleProductVariantToCartValidator.php
Tomanhez: I checked it, only object/string is possible
Screenshot 2021-05-20 at 13 36 23

src/Sylius/Bundle/ApiBundle/Validator/Constraints/EligibleProductVariantToCartValidator.php
lchrusciel: On the other hand, if one would like to customize behavior of adding to cart validator, but not the change items quantity, they would have a hard time. 
Why not inverse the logic? Right now I see ~10 paths going throughout this class. It's way too much for regular service. On the other hand, there are only three types of validation errors. If you would extract common checking logic to some services like:
ProductVariantPurchaseAvailibilityChecker - to check if stock is available
ProductVariantEnebilityChecker - to check if variant exist/is turned on
ProductEnebilityChecker - to check if product exist/is turned on

The first two could have even the same interface, but that would mean that we are diffusing the meaning of each service - so I would rather even consider packing them together and leaving only two types of communication from here. Or just stick to the interface for each of the above-mentioned checkers.

In addition, it will be waaaay easier to test it with spec. Another side effect? Adding new checking logic will be piece of cake

src/Sylius/Bundle/ApiBundle/spec/Validator/Constraints/EligibleProductVariantToCartValidatorSpec.php
```diff
@@ -76,7 +83,7 @@ function it_adds_violation_if_product_variant_does_not_exist(
 
         $this->validate(
             new AddItemToCart( 'productVariantCode', 1),
```
lchrusciel, line 85: Talking about the amount of paths that go through this service, in case you would like to keep both constraints in one validator you need to duplicate most of the tests here with the change of the constraint and its behavior - ChangeItemQuantityInCart

Plus, can you remove this space before `'` as you already touched this code?

src/Sylius/Bundle/ApiBundle/Validator/Constraints/EligibleProductVariantToCartValidator.php
```diff
@@ -14,59 +14,74 @@
...
 
-        /** @var ProductVariantInterface|null $productVariant */
-        $productVariant = $this->productVariantRepository->findOneBy(['code' =>$value->productVariantCode]);
+        /** @var EligibleProductVariantToCart $constraint */
+        Assert::isInstanceOf($constraint, EligibleProductVariantToCart::class);
 
-        if ($productVariant === null) {
+        if ($this->productVariant === null) {
             $this->context->addViolation(
                 $constraint->productVariantNotExistMessage,
```
lchrusciel, line 61: Also, is it really good validation message? We already had this variant in a cart and during the update it disappeared? No longer available seems like a better wording

src/Sylius/Bundle/ApiBundle/Validator/Constraints/ChangingItemQuantityInCart.php
```diff
@@ -0,0 +1,38 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+final class ChangingItemQuantityInCart extends Constraint
```
GSadee, line 18:
```suggestion
final class ChangedItemQuantityInCart extends Constraint
```
would be more consistent with other constraints

src/Sylius/Bundle/ApiBundle/spec/Validator/Constraints/ChangingItemQuantityInCartValidatorSpec.php
```diff
@@ -0,0 +1,282 @@
...
+            ->shouldThrow(\InvalidArgumentException::class)
+            ->during('validate', [new CompleteOrder(), new AddingEligibleProductVariantToCart()]);
+    }
+
+    function it_throws_an_exception_if_constraint_is_not_an_instance_of_changing_item_quantity_in_cart(): void
+    {
+        $this
+            ->shouldThrow(\InvalidArgumentException::class)
+            ->during('validate', [
+                new ChangeItemQuantityInCart( 2),
```
arti0090, line 60:
```suggestion
                new ChangeItemQuantityInCart(2),
```
😱

GSadee (approved):

GSadee:

PR #12696, opened 2021-06-07
WIP Add Plugin Creator Command

tonicospinelli:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

It's just a lightweight proposal to automate the creation of the Sylius Plugin.
This work in progress is to onboard new developers to create their sylius plugins.

image

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Command/PluginCreateCommand.php
+++ b/src/Sylius/Bundle/CoreBundle/Command/PluginCreateCommand.php
@@ -1,0 +1,84 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\Command;
+
+use Sylius\Bundle\CoreBundle\Installer\Plugin\ComposerSetup;
+use Sylius\Bundle\CoreBundle\Installer\Plugin\ConfigurationSetup;
+use Sylius\Bundle\CoreBundle\Installer\Plugin\ExtensionSetup;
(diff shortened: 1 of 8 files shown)


tonicospinelli:
if you understand this command makes sense, please, give me a feedback and let's improve it 🤗

loevgaard:
It shouldn't be in this repository. I have tried to something similar in this skeleton: https://github.com/Setono/SyliusPluginSkeleton with [init file](https://github.com/Setono/SyliusPluginSkeleton/blob/master/init).

vvasiloi:
This should probably be in https://github.com/Sylius/PluginSkeleton.

tonicospinelli:
@vvasiloi @loevgaard

Giving my rationale on the subject...

The core is responsible for the Sylius resources and ecosystem. Therefore, it is fair that you provide any tool to extend your capabilities.

In my point of view, the plugin is an alternative to extend or add any new resource to the Sylius platform. Therefore, this repo may be responsible for providing tools and instructions for extending them.

With that in mind, any tools used to extend or add resources, or command lines, or any instructions for creating or installing a new Sylius plugin or resources should be here.

For example, the symfony framework provides an empty [skeleton] (https://github.com/symfony/skeleton) to create a new project.

Also, it can be an opportunity to start an initiative to install or generate a new plugin based on existing or new features.

Does it make sense?

vvasiloi:
@tonicospinelli 
It makes sense, however I have some counter arguments.
The Symfony Skeleton is the equivalent of Sylius Standard, not the plugins.
I see what you're trying to do more as an optional dev dependency like [Symfony Maker](https://symfony.com/doc/current/bundles/SymfonyMakerBundle/index.html)
Also, I think it's better to be avoid with adding more stuff to the core, when possible, otherwise it will be harder and harder to maintain.
What do you think?

jakubtobiasz:
Closing as this functionality is covered in https://github.com/Sylius/PluginTemplate

PR #12698, opened 2021-06-08
Fix the build with Symfony 5.3

pamil:


.github/workflows/application.yml
```diff
@@ -32,7 +32,7 @@ jobs:
             fail-fast: false
             matrix:
                 php: [7.3, 7.4]
-                symfony: [^4.4, 5.2.*]
+                symfony: [^4.4, ^5.2]
```
lchrusciel, line 35: Shouldn't we explicitly test 5.2 and 5.3? Or bump the minimal version to 5.3 directly?

lchrusciel (approved):

.github/workflows/application.yml
pamil: Bumping in a patch release is discouraged - considering price/quality ratio for the builds, we should be quite safe with Symfony 5.2 since Symfony 4.4 build prevents us from using Symfony 5.3 features.

GSadee (approved):

PR #12686, opened 2021-06-02
[DOCS] Cookbook for customizing refund process

arti0090:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

PR #12702, opened 2021-06-09
[Documentation] Fix Refund Process Cookbook

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | after #12686 
| License         | MIT


docs/cookbook/index.rst
```diff
@@ -149,13 +150,3 @@ Frontend
...
 
 .. include:: /cookbook/frontend/map.rst.inc
-
-Refund
-------
-
-.. toctree::
-    :hidden:
-
-    refund/refund-process.rst
```
arti0090, line 159: i am a bit concerned that payments will mix up with refunds in this case, but still customers can use `search` and find what they are looking for (very low priority comment imo) 🙂

arti0090 (approved):

docs/cookbook/index.rst
GSadee: I wanted to be consistent with our current docs structure, where we have cookbook for customizing Credit Memo in Paymets section, IMO these two cookbooks should belong to the same section

Tomanhez (approved):

Zales0123 (approved):

Zales0123:

PR #12713, opened 2021-06-17
[API] Use serializer instead of custom view for shipping methods

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | replaces https://github.com/Sylius/Sylius/pull/12392
| License         | MIT

PR #12716, opened 2021-06-17
[API][ShippingMethod] Add contract test for available shipping methods endpoint

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | based on #12713 
| License         | MIT


src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -669,7 +669,7 @@ public function myOrdersShippingMethodShouldBe(ShippingMethodInterface $shipping
         $shippingMethodName = $shippingMethod->getName();
 
         foreach ($shippingMethods as $method) {
-            if ($method['shippingMethod']['translations']['en_US']['name'] === $shippingMethodName) {
+            if ($method['translations']['en_US']['name'] === $shippingMethodName) {
```
pamil, line 672: I think we should mention that in UPGRADE file that the response schema has changed.

GSadee: Added 👌🏻

src/Sylius/Bundle/ApiBundle/Serializer/ShippingMethodNormalizer.php
```diff
@@ -0,0 +1,90 @@
+<?php
+
```
AdamKasp, line 2: missing license block

AdamKasp (approved):

src/Sylius/Bundle/ApiBundle/spec/Serializer/ShippingMethodNormalizerSpec.php
```diff
@@ -0,0 +1,134 @@
...
+use Sylius\Component\Core\Model\ShippingMethodInterface;
+use Sylius\Component\Core\Repository\OrderRepositoryInterface;
+use Sylius\Component\Core\Repository\ShipmentRepositoryInterface;
+use Sylius\Component\Registry\ServiceRegistryInterface;
+use Sylius\Component\Shipping\Calculator\CalculatorInterface;
+use Symfony\Component\Serializer\Normalizer\NormalizerInterface;
+
+final class ShippingMethodNormalizerSpec extends ObjectBehavior
+{
+    function let(
```
SirDomin, line 28: public before function? same for all methods in this file

SirDomin (commented):
image
Do we need these 2 empty files?

GSadee:
> image
> Do we need these 2 empty files?

@SirDomin these files are not empty, just no changes were made to them except the filename

src/Sylius/Bundle/ApiBundle/Serializer/ShippingMethodNormalizer.php
```diff
@@ -0,0 +1,99 @@
...
+        $subresourceIdentifiers = $context['subresource_identifiers'] ?? null;
+
+        return
+            $data instanceof ShippingMethodInterface &&
+            $this->isNotAdminGetOperation($context) &&
+            isset($subresourceIdentifiers['tokenValue'], $subresourceIdentifiers['shipments'])
+        ;
+    }
+
+    private function isNotAdminGetOperation(array $context): bool
```
Zales0123, line 95: The second normalizer where we use such a function, maybe it would worth to extract it into some separate service in the future 🖖 💃

Zales0123 (approved):

arti0090 (approved):

AdamKasp:

tests/Api/Responses/Expected/shop/get_shipping_methods_response.json
```diff
@@ -0,0 +1,44 @@
+{
+    "@context": "\/api\/v2\/contexts\/ShippingMethod",
+    "@id": "\/api\/v2\/shop\/orders\/nAWw2jewpA\/shipments\/@integer@\/methods",
+    "@type": "hydra:Collection",
+    "hydra:member": [
+        {
+            "@id": "\/api\/v2\/shop\/shipping-methods\/UPS",
+            "@type": "ShippingMethod",
+            "id": @integer@,
```
lchrusciel, line 9:
```suggestion
            "id": "@integer@",
```
From the matcher's point of view, such a notation would be equal to the current one. The change is, that GitHub nor phpstorm would complain about invalid json shema. On the other hand, it would look like there is a string at the integer field... Just a proposal. It is worth it to put all matchers into `"`, but I would love to hear your opinion about it as well

PR #12719, opened 2021-06-18
[Api] Unnecessary addresses should not be saved to the database while addressing an order

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT


features/checkout/addressing_order/changing_address.feature
```diff
@@ -0,0 +1,21 @@
+@checkout
+Feature: Changing address
+    In order to change address in the cart
+    As a Visitor
+    I want to be able to change address in the cart
```
GSadee, line 5:
```suggestion
Feature: Changing address during checkout
    In order to place an order with the correct address
    As a Visitor
    I want to be able to change the address during checkout
```

features/checkout/addressing_order/changing_address.feature
```diff
@@ -0,0 +1,21 @@
...
+        Given the store operates on a single channel in "United States"
+        And the store has a product "T-shirt banana" priced at "$12.54"
+
+    @api
+    Scenario: Changing address
+        Given the visitor has product "T-shirt banana" in the cart
+        And the visitor has specified the email as "jon.snow@example.com"
+        And the visitor has specified address as "Los Angeles", "Frost Alley", "90210", "United States" for "Jon Snow"
+        And the visitor has completed the addressing step
+        When the visitor change the billing address as "Los Angeles", "Frost Alley", "90210", "United States" for "Jon Snow"
```
GSadee, line 17:
```suggestion
        When the visitor change the billing address to "Los Angeles", "Frost Alley", "90210", "United States" for "Jon Snow"
```

GSadee: What is more, if I see correctly, this address is exactly the same as the one you are changing, so this test checks nothing... You should use here a different address

src/Sylius/Behat/Context/Api/AddressContext.php
```diff
@@ -0,0 +1,48 @@
...
+
+    public function __construct(AddressRepositoryInterface $addressRepository)
+    {
+        $this->addressRepository = $addressRepository;
+    }
+
+    /**
+     * @Then /^store should contain only ("[^"]*", "[^"]*", "[^"]*", "[^"]*", "[^"]*" address)$/
+     */
+    public function storeShouldContainOnlyAddress(AddressInterface $address): void
```
GSadee, line 33: It is a domain/application test rather than api. Maybe the better way will be adding a unit test instead of behat to check the behaviour of one handler 🤔

src/Sylius/Bundle/ApiBundle/spec/CommandHandler/Checkout/AddressOrderHandlerSpec.php
```diff
@@ -187,6 +195,50 @@ function it_handles_addressing_an_order_for_not_logged_in_shop_user(
         $this($addressOrder);
     }
 
+    function it_does_not_create_new_addresses_for_already_addressed_order(
```
GSadee, line 198: This is not 100% true, you are creating new addresses, but also removing the old ones.

src/Sylius/Bundle/ApiBundle/CommandHandler/Checkout/AddressOrderHandler.php
```diff
@@ -76,6 +77,19 @@ public function __invoke(AddressOrder $addressOrder): OrderInterface
...
+        /** @var AddressInterface|null $shippingAddress */
+        $shippingAddress = $order->getShippingAddress();
+
+        if ($billingAddress !== null) {
+            $this->manager->remove($billingAddress);
+        }
+
+        if ($shippingAddress !== null) {
+            $this->manager->remove($shippingAddress);
+        }
```
GSadee, line 91: Instead of removing, couldn't we update the existing addresses?

src/Sylius/Behat/Context/Api/AddressContext.php
Tomanhez: I thought about, I think both options are good, UnitTest need more work. We have similar context for emails that is using with api tests, I know that its mix api/domain test but is more readable that UnitTest

features/checkout/addressing_order/changing_address.feature
pamil:
```suggestion
        When the visitor changes the billing address as "Los Angeles", "Frost Alley", "90210", "United States" for "Jon Snow"
```

features/checkout/addressing_order/changing_address.feature
```diff
@@ -0,0 +1,21 @@
...
+        And the visitor complete the addressing step
```
pamil, line 18:
```suggestion
        And the visitor completes the addressing step
```

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -191,7 +191,7 @@ public function iSpecifyTheEmailAs(?string $email): void
 
     /**
      * @When /^I specify the billing (address as "([^"]+)", "([^"]+)", "([^"]+)", "([^"]+)" for "([^"]+)")$/
-     * @When /^the (?:customer|visitor) specify the billing (address as "([^"]+)", "([^"]+)", "([^"]+)", "([^"]+)" for "([^"]+)")$/
+     * @When /^the (?:customer|visitor) (?:specify|change) the billing (address as "([^"]+)", "([^"]+)", "([^"]+)", "([^"]+)" for "([^"]+)")$/
```
pamil, line 194:
```suggestion
     * @When /^the (?:customer|visitor) (?:specifies|changes) the billing (address as "([^"]+)", "([^"]+)", "([^"]+)", "([^"]+)" for "([^"]+)")$/
```

src/Sylius/Bundle/ApiBundle/CommandHandler/Checkout/AddressOrderHandler.php
Tomanhez: Do you have any idea how? I try do it without using setters for all fields on adress object but it creates new object on database and leaves old. Using getters from new object and setters for old is less customizable IMO. All added new fields by whose who customize address will be not supported.

src/Sylius/Behat/Context/Api/AddressContext.php
Zales0123: The described feature definitely does not benefits a feature's actor (a Visitor). We should not use byways and focus on implementing the feature in a proper way. Therefore, as mentioned by @GSadee, this definitely should be an application/domain scenario or as a functional test.

src/Sylius/Bundle/ApiBundle/CommandHandler/Checkout/AddressOrderHandler.php
Zales0123: Do you say, that if you set new values on the existing shipping/billing address (and flushes the manager), the new database entry is created? How is it possible? :) It should definitely not work this way, doctrine take care of the upgrades well for other entities

Zales0123 (changes requested):

arti0090 (approved):

src/Sylius/Bundle/ApiBundle/spec/CommandHandler/Checkout/AddressOrderHandlerSpec.php
```diff
@@ -187,6 +198,53 @@ function it_handles_addressing_an_order_for_not_logged_in_shop_user(
         $this($addressOrder);
     }
 
+    function it_deletes_old_order_address_while_creating_new_addresses_for_already_addressed_order(
```
GSadee, line 201: I'm not convinced if it is true 🤔

GSadee (approved):

src/Sylius/Bundle/ApiBundle/CommandHandler/Checkout/AddressOrderHandler.php
```diff
@@ -76,8 +83,29 @@ public function __invoke(AddressOrder $addressOrder): OrderInterface
...
+        /** @var AddressInterface|null $shippingAddress */
+        $shippingAddress = $order->getShippingAddress();
+
+        if ($billingAddress !== null) {
+            $order->setBillingAddress($this->addressMapper->mapExisting($billingAddress, $newBillingAddress));
+        } else {
+            $order->setBillingAddress($newBillingAddress);
+        }
+
+        $newShippingAddress = $newShippingAddress ?? clone $newBillingAddress;
```
Zales0123, line 102: Cannot we do it in line 89?
```
$newShippingAddress = $addressOrder->shippingAddress ?? clone $newBillingAddress;
```

src/Sylius/Bundle/ApiBundle/Mapper/AddressMapperInterface.php
```diff
@@ -0,0 +1,21 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Mapper;
+
+use Sylius\Component\Core\Model\AddressInterface;
+
+interface AddressMapperInterface
+{
+    public function mapExisting(AddressInterface $targetAddress, AddressInterface $giverAddress): AddressInterface;
```
Zales0123, line 20: `giver` 💃 🔫 

maybe `$currentAddress` and `$targetAddress`?

src/Sylius/Bundle/ApiBundle/Tests/CommandHandler/ChangeAddressOrderHandlerTest.php
```diff
@@ -0,0 +1,195 @@
...
+
+        /** @var ObjectManager $manager */
+        $manager = $container->get('doctrine.orm.default_entity_manager');
+
+        /** @var StateMachineFactoryInterface $stateMachineFactory */
+        $stateMachineFactory = $container->get('sm.factory');
+
+        /** @var AddressMapperInterface $addressMapper */
+        $addressMapper = $container->get('Sylius\Bundle\ApiBundle\Mapper\AddressMapperInterface');
+
```
Zales0123, line 72: This setup is so big 😢 Is there any chance we could use some fixtures for that? Nelmio maybe?

src/Sylius/Bundle/ApiBundle/spec/Mapper/AddressMapperSpec.php
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace spec\Sylius\Bundle\ApiBundle\Mapper;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Core\Model\AddressInterface;
+
+final class AddressMapperSpec extends ObjectBehavior
```
Zales0123, line 19: Spec for this class looks hideous 😆 It can stay, but I believe a PHPUnit test would be more suitable and nicer to read 🖖

features/checkout/addressing_order/changing_address_during_checkout.feature
```diff
@@ -0,0 +1,20 @@
...
+    In order to place an order with the correct address
+    As a Visitor
+    I want to be able to change the address during checkout
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "T-shirt banana" priced at "$12.54"
+
+    @api
+    Scenario: Changing address
```
Zales0123, line 12: As this scenario is not checking an implemented improvement, are we sure we do not have a similar scenario, maybe not implemented in `@api` context? I just want to avoid the duplication

GSadee: This one looks similar: `features/checkout/addressing_order/returning_to_addressing_step_and_changing_address_data.feature`

src/Sylius/Bundle/ApiBundle/Mapper/AddressMapper.php
```diff
@@ -0,0 +1,38 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Mapper;
+
+use Sylius\Component\Core\Model\AddressInterface;
+
+final class AddressMapper implements AddressMapperInterface
+{
+    public function mapExisting(AddressInterface $targetAddress, AddressInterface $giverAddress): AddressInterface
```
SirDomin, line 20: another giver 🔫

src/Sylius/Bundle/ApiBundle/Tests/CommandHandler/ChangeAddressOrderHandlerTest.php
```diff
@@ -0,0 +1,195 @@
...
+                $addressFromDatabase->getCompany() !== $newBillingAddress->getCompany() ||
+                $addressFromDatabase->getStreet() !== $newBillingAddress->getStreet() ||
+                $addressFromDatabase->getCity() !== $newBillingAddress->getCity() ||
+                $addressFromDatabase->getPostcode() !== $newBillingAddress->getPostcode() ||
+                $addressFromDatabase->getPhoneNumber() !== $newBillingAddress->getPhoneNumber() ||
+                $addressFromDatabase->getCountryCode() !== $newBillingAddress->getCountryCode() ||
+                $addressFromDatabase->getProvinceName() !== $newBillingAddress->getProvinceName() ||
+                $addressFromDatabase->getProvinceCode() !== $newBillingAddress->getProvinceCode()
+            ) {
+                throw new \InvalidArgumentException('The store has other address than should');
```
SirDomin, line 164:
```suggestion
                throw new \InvalidArgumentException('The store has other address than it should');
```
i think it looks better :D

src/Sylius/Bundle/ApiBundle/CommandHandler/Checkout/AddressOrderHandler.php
```diff
@@ -76,8 +83,27 @@ public function __invoke(AddressOrder $addressOrder): OrderInterface
             $order->setCustomer($this->provideCustomerByEmail($addressOrder->email));
         }
 
-        $order->setBillingAddress($addressOrder->billingAddress);
-        $order->setShippingAddress($addressOrder->shippingAddress ?? clone $addressOrder->billingAddress);
+        /** @var AddressInterface $newBillingAddress */
+        $newBillingAddress = $addressOrder->billingAddress;
+        /** @var AddressInterface|null $newShippingAddress */
+        $newShippingAddress = $addressOrder->shippingAddress ?? clone $newBillingAddress;
```
stloyd, line 89: Both assigns are pointless to be done here, and should be done in `if`'s below not before them IMHO.

src/Sylius/Bundle/ApiBundle/CommandHandler/Checkout/AddressOrderHandler.php
```diff
@@ -76,8 +83,27 @@ public function __invoke(AddressOrder $addressOrder): OrderInterface
             $order->setCustomer($this->provideCustomerByEmail($addressOrder->email));
         }
 
-        $order->setBillingAddress($addressOrder->billingAddress);
-        $order->setShippingAddress($addressOrder->shippingAddress ?? clone $addressOrder->billingAddress);
+        /** @var AddressInterface $newBillingAddress */
+        $newBillingAddress = $addressOrder->billingAddress;
+        /** @var AddressInterface|null $newShippingAddress */
```
stloyd, line 88:
```suggestion
        /** @var AddressInterface $newShippingAddress */
```
It cannot be `null` as there is `??` and even method call requires object not allowing to pass `null` into it.

src/Sylius/Bundle/ApiBundle/Tests/config/fixtures/channel.yaml
```diff
@@ -0,0 +1,13 @@
...
+        name: "Web Channel"
+        hostname: "localhost"
+        description: "Lorem ipsum"
+        baseCurrency: "@currency1"
+        defaultLocale: "@locale"
+        locales: ["@locale"]
+        color: "black"
+        enabled: true
+        taxCalculationStrategy: "order_items_based"
+
```
Zales0123, line 13: 💃

src/Sylius/Bundle/ApiBundle/Tests/CommandHandler/ChangeAddressOrderHandlerTest.php
```diff
@@ -0,0 +1,176 @@
...
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase;
+
+final class ChangeAddressOrderHandlerTest extends KernelTestCase
+{
+    /**
+     * @test
+     */
+    public function it_changes_address_order_without_duplication_in_database(): void
+    {
+        $container = self::bootKernel()->getContainer();
```
Zales0123, line 39: https://github.com/Sylius/Sylius/pull/12719#discussion_r656928647

Did you take it into consideration?

src/Sylius/Bundle/ApiBundle/CommandHandler/Checkout/AddressOrderHandler.php
Zales0123: I believe the reason is that we need to type-hint them with `Core/AddressInterface`... Why these properties are type-hinted with `Addressing/AddressInterface` in the command?

features/checkout/addressing_order/changing_address_during_checkout.feature
GSadee: @Tomanhez From what I can see you deleted your scenario and didn't implement the one mentioned by me, why?

GSadee (approved):

features/checkout/addressing_order/changing_address_during_checkout.feature
Tomanhez: I understood that this scenario is redundant in favor of above scenario

arti0090 (approved):

features/checkout/addressing_order/changing_address_during_checkout.feature
GSadee: But the above scenario is not implemented in `@api` context

features/checkout/addressing_order/changing_address_during_checkout.feature
```diff
@@ -0,0 +1,33 @@
...
+        Given the visitor has product "T-shirt banana" in the cart
+        And the visitor has specified the email as "jon.snow@example.com"
+        And the visitor has specified address as "Los Angeles", "Frost Alley", "90210", "United States" for "Jon Snow"
+        And the visitor has completed the addressing step
+        When the visitor changes the billing address to "Los Angeles", "Avenue", "90210", "United States" for "Jon Snow"
+        And the visitor completes the addressing step
+        Then the visitor should has "Los Angeles", "Avenue", "90210", "United States", "Jon Snow" specified as billing address
+        Then the visitor should has "Los Angeles", "Avenue", "90210", "United States", "Jon Snow" specified as shipping address
```
GSadee, line 33:
```suggestion
        And the visitor should has "Los Angeles", "Avenue", "90210", "United States", "Jon Snow" specified as shipping address
```

GSadee (approved):

GSadee:

PR #12737, opened 2021-06-21
Force PHP 8.0.5 to avoid PHP 8.0.7 which might cause segfaults

pamil:


.github/workflows/application.yml
```diff
@@ -31,8 +31,8 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: ["7.4", "8.0"]
-                symfony: ["^4.4", "^5.2"]
+                php: ["7.4", "8.0.5"]
+                symfony: ["^4.4", "5.3"]
```
GSadee, line 35: Shouldn't be:
```suggestion
                symfony: ["^4.4", "^5.3"]
```
?

PR #12739, opened 2021-06-21
Translate AJAX login error messages

### LINKED ISSUE #12738: "Invalid credentials" message is not translated during AJAX login
**Sylius version affected**: reproduced on 1.8.x but I'm sure all version are affected

**Description**  
During the login via AJAX request the "Invalid credentials" message is not translated in the user locale language.

**Steps to reproduce**  
* Go to https://demo.sylius.com/admin and login to the admin panel
* Go to Channels -> Fashion Web Store
* Set the channel locale to a locale for which there's a translation of the Symfony's "Invalid credentials" message (for example German).
* Go to the frontend https://demo.sylius.com
* Add a product to cart
* Go to the checkout addressing page
* Enter shop@example.com in the Email field
* Enter a wrong password in the Password field
* Click login
* You'll see "Invalid credentials" but you should see the German translation of that message (Fehlerhafte Zugangsdaten.)

**Possible Solution**  
The problem is here:

https://github.com/Sylius/Sylius/blob/ec917067c5bc7ab38f119edac76566a88a7f5aa6/src/Sylius/Bundle/UserBundle/Authentication/AuthenticationFailureHandler.php#L27

The custom AuthenticationFailureHandler does not translate the message.

I'll provide a PR for that.

mmenozzi:
| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | fixes #12738 
| License         | MIT

### DIFF
--- a/UPGRADE-1.9.md
+++ b/UPGRADE-1.9.md
@@ -1,3 +1,7 @@
+# UPGRADE FROM `v1.9.4` TO `v1.9.5`
+
+1. `Sylius\Bundle\UserBundle\Authentication\AuthenticationFailureHandler` now needs the `translator`. Make sure to inject it using setter injection if you customized the `sylius.authentication.failure_handler` service definition.
+
 # UPGRADE FROM `v1.9.3` TO `v1.9.4`
 
 1. `Sylius\Bundle\ApiBundle\DataProvider\OrderCollectionDataProvider` has been removed and the same logic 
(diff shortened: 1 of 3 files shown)


Rafikooo:
Hello @mmenozzi, thank you for your contribution, however the solution you proposed is more like your special use case (which you could decorate for example) than the solution we want in Sylius. Wish you all the best! 🎄

mmenozzi:
> Hello @mmenozzi, thank you for your contribution, however the solution you proposed is more like your special use case (which you could decorate for example) than the solution we want in Sylius. Wish you all the best! 🎄 

Hi @Rafikooo, what do you mean? Logging in is not a special use case of mine. It's a feature of Sylius and there's a bug in this feature (see #12738). This PR fixes it. Period.

If, in your opinion, this fix has any problem tell me the problem. I can fix it.

I know I can decorate the service in my store if I want to fix this bug. But this is not a good reason to not fix it.

PR #12746, opened 2021-06-23
[API] handle shipping without addressing

SirDomin:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
@@ -420,6 +420,18 @@ final class CheckoutContext implements Context
     }
 
     /**
+     * @Then I should be notified that the order should be addressed first
+     */
+    public function iShouldBeNotifiedThatTheOrderShouldBeAddressedFirst()
+    {
+        Assert::same($this->ordersClient->getLastResponse()->getStatusCode(), Response::HTTP_UNPROCESSABLE_ENTITY);
+
+        $content = json_decode($this->ordersClient->getLastResponse()->getContent(), true);
+
+        Assert::same($content['message'], 'Order should be addressed first.');
+    }
+
+    /**
      * @Then I should be informed that shipping method with code :code does not exist
      */
     public function iShouldBeInformedThatShippingMethodWithCodeDoesNotExist(string $code): void
--- a/src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
@@ -422,11 +422,13 @@ final class CheckoutContext implements Context
     /**
      * @Then I should be notified that the order should be addressed first
      */
-    public function iShouldBeNotifiedThatTheOrderShouldBeAddressedFirst()
+    public function iShouldBeNotifiedThatTheOrderShouldBeAddressedFirst(): void
     {
-        Assert::same($this->ordersClient->getLastResponse()->getStatusCode(), Response::HTTP_UNPROCESSABLE_ENTITY);
+        $response = $this->ordersClient->getLastResponse();
 
-        $content = json_decode($this->ordersClient->getLastResponse()->getContent(), true);
+        Assert::same($response->getStatusCode(), Response::HTTP_UNPROCESSABLE_ENTITY);
+
+        $content = json_decode($response->getContent(), true);
 
         Assert::same($content['message'], 'Order should be addressed first.');
     }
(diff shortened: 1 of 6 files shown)


src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -419,6 +419,18 @@ public function iTryToSelectPaymentMethod(string $paymentMethodCode): void
         $this->ordersClient->executeCustomRequest($request);
     }
 
+    /**
+     * @Then I should be notified that the order should be addressed first
+     */
+    public function iShouldBeNotifiedThatTheOrderShouldBeAddressedFirst()
+    {
+        Assert::same($this->ordersClient->getLastResponse()->getStatusCode(), Response::HTTP_UNPROCESSABLE_ENTITY);
```
Zales0123, line 427: We could extract `$this->ordersClient->getLastResponse()` to the variable

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -419,6 +419,18 @@ public function iTryToSelectPaymentMethod(string $paymentMethodCode): void
         $this->ordersClient->executeCustomRequest($request);
     }
 
+    /**
+     * @Then I should be notified that the order should be addressed first
+     */
+    public function iShouldBeNotifiedThatTheOrderShouldBeAddressedFirst()
```
Zales0123, line 425: Missing `: void`

src/Sylius/Bundle/ApiBundle/Exception/OrderCannotBeShippedWithoutAddressing.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Exception;
+
+class OrderCannotBeShippedWithoutAddressing extends \RuntimeException
```
Zales0123, line 16: I suppose it could be final

src/Sylius/Bundle/ApiBundle/CommandHandler/Checkout/ChooseShippingMethodHandler.php
```diff
@@ -66,6 +67,10 @@ public function __invoke(ChooseShippingMethod $chooseShippingMethod): OrderInter
 
         $stateMachine = $this->stateMachineFactory->get($cart, OrderCheckoutTransitions::GRAPH);
 
+        if ($cart->getShippingAddress() === null) {
+            throw new OrderCannotBeShippedWithoutAddressing();
```
Zales0123, line 71: I wonder, is it something we would like to do with a custom exception? Or maybe it should be a guard on state machine? 🤔 We're checking the possibility of transitioning order below nonetheless. cc @GSadee @lchrusciel

SirDomin: state machine is check in the next lines, it throws 500 error if state cant be applied, this way we capture error we expect before and throw custom error to tell what caused this error.

Zales0123: I totally understand that 😄 I just wonder shouldn't we keep the order validation in one place rather than have it done with two fashions (custom exception + state machine checking). Still, it's just an improvement 🖖

Zales0123 (approved):

Zales0123:
Thank you, @SirDomin! :1st_place_medal:

src/Sylius/Bundle/ApiBundle/CommandHandler/Checkout/ChooseShippingMethodHandler.php
lchrusciel: Validation should be handled mostly with the validation component. That should be possible with the command, that we are having here. I'm wondering if we should start moving some protecting rules into the state machine, as we don't have any yet. Exception itself is not a big deal in the current implementation tho

SirDomin: https://github.com/Sylius/Sylius/pull/12746 okay, added another approach (this time with validations)

PR #12752, opened 2021-06-24
[ADR] Strategy for resource internationalization in API

### LINKED ISSUE #11412: Unified API: JSON-LD and internationalisation
# What is JSON-LD?

JSON-LD stands for “JSON for Linking Data”. It’s developed by W3C and used to build interoperable Web services.

JSON-LD 1.1 provides a non-normative specification for string internationalisation.

JSON-LD is meant to be used in the server responses, providing data that can be processed and normalised with JSON-LD processor, which behaviour can be explored further using JSON-LD Playground.

# How can JSON-LD be used for internationalisation?

JSON-LD internationalisation support can be applied to the whole document, or to specific fields only. JSON-LD processor will normalise both the internationalised fields and the non-internationalised fields:

1. Field `jobTitle` is set to be in the en language
2. After processing, `jobTitle` field is a list of maps, each map containing the information of the language and the translated value
3. For the non-internationalised fields like name, they will be processed similarly, containing a list of maps, each map containing the value, but no language data, as it’s not specified in the context

<hr />

The way JSON-LD processor works allows us to add more translations for a given field without sacrificing backwards compatibility, by using `@container` in the context:

1. By setting `@container` as `@language` for `jobTitle` field, we tell the processor that `jobTitle` field will be a map, containing translated values indexed by the language.
2. `jobTitle` changes its form from just a simple string in a previous example to a map mentioned above
3. Even though `jobTitle` structure has changed, the processed JSON-LD returns a backwards-compatible structure. `jobTitle` field is still a list containing a map of the language to the translated value, but containing two elements now.

# Could we apply JSON-LD i18n to the existing structure?

*TL;DR: No.*

Having translations as a map of languages to a map of translations is not compatible with JSON-LD specification.

jsonld3

# How to determine which languages to include in the response?

To prevent overfetching, improve the performance and maintain consistent behaviour across our existing store frontend and Shop API, it is required to define a clear way to determine which languages should be included in the server response. 

Some inspiration can be found in Locastic/ApiPlatformTranslationBundle and Accept-Language header spec.

I propose to use Accept-Language header and extend its usage to a GET parameter if needed. Accept-Language header accepts either a list of accepted languages or `*` as a wildcard for all of them:
 - If we get a single language like `en` - return translations only for this language
- If we get a list like `en, pl, de` - return translations in those three languages OR prefer `en` and fallback to `pl` and then to `de`
 - If we get `*` - return all the possible translations

If the request is made in shop context (eg. user is not logged in / logged in as a shop user), we should filter out the list of languages using the list of languages defined on the channel, accepting only the ones existing in both of those lists.

# Okay, what about requests to the server: POST/PUT/PATCH?

It’s time for things to get trickier - setting translations while adding a new translatable resource or updating one.

In the current implementation, the responses from GET requests are almost symmetrical to the payload for POST/PUT requests. If the GET response gets to conform with JSON-LD internationalisation, this would no longer be the case. However, there are two main schema designs to explore:

### Translations as the main container for any translatable value

This solution is the same as the currently used - all the translatable data is stored in the translations field, all the non-translatable data is stored at the root level.

```json
{
    "code": "JOHN_SMITH", 
    "translations": {
        "en": {
            "name": "John Smith",
            "title": "Blackmith"
        },
        "pl": {
            "name": "Jan Kowalski",
            "title": "Kowal"
        }
    }
}
```

### Translations embedded in every translatable field

This solution is more similar to the JSON-LD one, with language maps embedded for translatable properties, next to non-translatable ones.

```json
{
    "code": "JOHN_SMITH",
    "name": {
        "en": "John Smith",
        "pl": "Jan Kowalski"
    },
    "title": {
        "en": "Blacksmith",
        "pl": "Kowal"
    }
}
```

# Conclusion?

Further R&D work is needed to define the input schema in our API. JSON-LD looks like a good solution for the output schema in our API.

PoC needs to be made to make sure there are no hidden blocking issues. Collaboration with the API Platform team could be a way to explore more possible solutions, as well as to get some advice regarding the implementation details.

# References

- JSON-LD 1.1 specification: https://www.w3.org/TR/json-ld11/
- JSON-LD Playground (live processor): https://json-ld.org/playground/
- JSON-LD benefits & introduction: https://rollout.io/blog/json-ld-building-meaningful-data-apis/
- JSON-LD for developers (practical info): https://datalanguage.com/blog/publishing-json-ld-for-developers
- JSON-LD origin story & critique: http://manu.sporny.org/2014/json-ld-origins-2/

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.12
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | closes #11412
| License         | MIT

### DIFF
--- a/adr/2021_06_24_api_content_internationalization.md
+++ b/adr/2021_06_24_api_content_internationalization.md
@@ -1,0 +1,61 @@
+# API - Customization internationalization
+
+* Status: proposed
+* Date: 2021-06-24
+
+## Context and Problem Statement
+
+A big chunk of resources available in Sylius have some part locale aware. Product descriptions, names of payment or shipping methods
+should be translated to the customer to provide the best user experience. Right now most of the resources are returned with 
+all available translations in the store and proper localization of content is left to the frontend developers. What is more,
+this solution is not consistent, as some endpoints already embeds some part of translatables in them. We should provide clear,
+easy to use and consistent way of handling customer locales. 
+
+## Considered Options
+
+### Returning all possible translations to the frontend
+
+We can update all not coherent at the moment resources to always returning all possible translations and leave it to the fronted to render proper one.
+
+* Good, because it is simplest and fastest solution at the moment of writing this document.
+* Good, because it leaves decision of proper content localization to the frontend developer. This guarantees flexibility.
+* Bad, because of data over-fetching. Most of the locales in the most to the cases won't be needed.
+* Bad, because it leaves decision of proper content localization to the frontend developer. Additional work is required on the frontend
+  App in order to provide localization functionality.
+
+### Taking advantage of JSON-LD specification for strings internationalization
+
+As researched by @pamil, we can take advantage of JSON-LD processor, return data according to its specification and leave 
+proper data structure to the JSON-LD processor.
+
+* Good, because it follows some external standard. This may allow us to take an advantage of the standard ecosystem.
+* Good, because it solves over-fetching problem, yet leaves the flexibility to the developer if they decide to fetch more than one locale in a single request.
+* Good, because it is suggested as a good direction for content internationalization by the API Platform core team. 
+* Bad, because we are bounded to only one standard. It may not solve internationalization issues with other formats than JSON-LD.
+* Bad, because it is not part of API Platform yet and significant effort would need to be done in order to provide proper solution.
+
+Ref.
+* https://github.com/Sylius/Sylius/issues/11412
+* https://github.com/api-platform/core/issues/127
+
+### Flatting down translatable parts of resources to the structure of the main resource
+
+This approach goes the most side to side with the default Sylius architecture. This is the way, how we treat Sylius resources 
+internally for most of the cases. It is supported out-of-the-box and can be easily achieved. By default all resources are 
+served in a default locale of a channel (or Administrator in terms of admin panel). This may fallback to app wise default locale.
+
+* Good, because it follows our current behaviour.
+* Good, because it solves over-fetching problem.
+* Good, because it is straight-forward to implement.
+* Bad, because it does not allow fetching more than one locale.
+* Bad, because it will become our internal standard only.
+
+## Decision Outcome
+
+The first approach should be used in admin section of API, while the third one should be used in shop section. This way, we
+will receive best of both worlds with the minimal effort. It is worth mentioning that Sylius have two distinguish entry points
+to API - admin and shop. They are concerned about different things and resolves different problems. Unified view of a product
+in the shop will make it easy and straightforward to build product pages. At the same time all translation approach should be used in
+admin to provide full control over "translation" resource - as we should treat them. Therefore, every resource translation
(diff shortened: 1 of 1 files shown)


adr/2021_06_24_api_content_internationalization.md
```diff
@@ -0,0 +1,61 @@
...
+
+* Good, because it is simplest and fastes solution at the moment of writing this document.
+* Good, because it leaves decision of proper content localization to the frontend developer. This guarantees flexibility.
+* Bad, because of data over-fetching. Most of the locales in the most to the cases won't be needed.
+* Bad, because it leaves decision of proper content localization to the frontend developer. Additional work is required on the frontend
+  App in order to provide localization functionality.
+
+### Taking advantage of JSON-LD specification for strings internationalization
+
+As researched by @pamil, we can take advanatge of JSON-LD processor, return data according to its specification and leave
```
GSadee, line 28:
```suggestion
As researched by @pamil, we can take advantage of JSON-LD processor, return data according to its specification and leave 
```

adr/2021_06_24_api_content_internationalization.md
```diff
@@ -0,0 +1,61 @@
...
+  App in order to provide localization functionality.
+
+### Taking advantage of JSON-LD specification for strings internationalization
+
+As researched by @pamil, we can take advanatge of JSON-LD processor, return data according to its specification and leave 
+proper data structure to the JSON-LD processor.
+
+* Good, because it follows some external standard. This may allow us to take an advantage of the standard ecosystem.
+* Good, because it solves over-fetching problem, yet leaves the flexibility to the developer if they decide to fetch more than one locale in a single request.
+* Good, because it is suggested as a good direction for content internationalization by the Api Platform core team.
```
GSadee, line 33:
```suggestion
* Good, because it is suggested as a good direction for content internationalization by the API Platform core team. 
```

adr/2021_06_24_api_content_internationalization.md
```diff
@@ -0,0 +1,61 @@
...
+
+Ref.
+* https://github.com/Sylius/Sylius/issues/11412
+* https://github.com/api-platform/core/issues/127
+
+### Flatting down translatable parts of resources to the structure of the main resource
+
+This approach goes the most side to side with the default Sylius architecture. This is the way, how we treat Sylius resources 
+internally for most of the cases. It is supported out-of-the-box and can be easily achieved. By default all resources are 
+served in a default locale of a channel(or Administrator in terms of admin panel). This may fallback to app wise default locale.
```
GSadee, line 45:
```suggestion
served in a default locale of a channel (or Administrator in terms of admin panel). This may fallback to app wise default locale.
```

GSadee (approved):

jakubtobiasz (approved):

jakubtobiasz:
Thank you, @lchrusciel!

PR #12758, opened 2021-06-28
Fetch shipping methods ordered all the time

Nek-:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | no/yes
| New feature?    | no/yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none, I just open this PR
| License         | MIT

This was an issue because depending on if the user has an account and an address or not, the behavior is different. Actually, in most cases the position will not be used, and it's pretty sad. :)


lchrusciel (commented):
Hey Nek-! Thanks a lot for your contribution. It seems that it is a really nice addition. I would consider it a bug fix. Can you reproduce this behaviour with behat scenarios and rebase your PR to 1.11? I would like to have tests to avoid regression. And these are only tests for our repositories at the moment

Nek-:
It seems like it has been fix already.

PR #12759, opened 2021-06-28
Use phpspec ^7.1 and avoid custom error reporting level in its configuration

pamil:


Zales0123 (approved):

Zales0123:

PR #12773, opened 2021-07-03
[API][FIX] Remove code attribute of OrderItem in API serialization

oallain:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | 
| License         | MIT

OrderItem has no attribut `code` but this attribute is in Serialisation configuration.
This throw an error when we Normalize Object.

### DIFF
--- a/UPGRADE-1.11.md
+++ b/UPGRADE-1.11.md
@@ -5,6 +5,9 @@
 `Sylius\Component\Channel\Context\RequestBased\HostnameBasedRequestResolver::findChannel` will start selecting only a channel from a range
 of enabled channels.
 
+2. The `code` field was removed from OrderItem serialization (in `src/Sylius/Bundle/ApiBundle/Resources/config/serialization/OrderItem.xml`)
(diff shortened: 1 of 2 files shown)


oallain:
Ping @lchrusciel

oallain:
Ping @Sylius/core-team ;)

timching:
facing the same issue, also found that same issue in the SyliusPlus in this file `vendor/sylius/plus/src/Resources/config/serialization/OrderItem.xml`

AdamKasp (approved):

Zales0123 (approved):
👍 but I wonder, shouldn't it be merged into 1.11 (as it's definitely a bug fix). And we definitely need to add a note into the UPGRADE file (_in theory_ one could already have a `code` property added in the `OrderItem` entity and count on this serialization configuration 🔥)

Zales0123:
Thanks, @oallain and @lchrusciel! :tada:

PR #12798, opened 2021-07-14
[WIP][PoC][API] Liip imagine image handling

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

A sample implementation of liip imagines handling. Each product image may be requested with additional filter query parameters and in return, a 302 redirect to the image on the server will be returned. It will leave up to the frontend developer to crop it on their own in the frontend or use one of backend prepared cached images: 
image

Enum value should be propagated automatically from the liip config.

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Filter/ImageFilters.php
+++ b/src/Sylius/Bundle/ApiBundle/Filter/ImageFilters.php
@@ -1,0 +1,21 @@
+<?php
+
+
+namespace Sylius\Bundle\ApiBundle\Filter;
+
+
+class ImageFilters
+{
+    /** @var array */
+    private $filters;
+
+    public function __construct($filters)
+    {
+        $this->filters = $filters;
+    }
+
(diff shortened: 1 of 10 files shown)


arti0090:
Before clicking `try it out` UI view is well... not nice. 
image

And that how it looks with all filters: 
image

Still if no filter is provided it fails to return image with 500 error.

If this will be merged, maybe we should consider dropping controler?
https://github.com/Sylius/Sylius/pull/12847

src/Sylius/Bundle/ApiBundle/Filter/ImageFilters.php
```diff
@@ -0,0 +1,21 @@
+<?php
```
arti0090, line 1: Not needed, left after concepting

PR #12831, opened 2021-07-22
[RFC] [UPGRADE] Split API and non API upgrade files

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

The idea came from our slack and I thought it may make sense. I'm proposing a split of upgrade files, as most people do not have to care about our API changes at the moment.


UPGRADE-1.10.md
```diff
@@ -1,12 +1,3 @@
-# UPGRADE FROM `v1.10.0` TO `v1.10.1`
-
-1. API is disabled by default, to enable it you need to set flag to ``true`` in ``app/config/packages/_sylius.yaml``:
```
stloyd, line 3: This should be left here IMHO as it's big BC break and should be widely visible.

GSadee (approved):

PR #12870, opened 2021-08-07
Remove shipments if no shipping method available to the order

kayue:
| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

**Description**  
- If a shipment already contain shipment previously, and system no longer have any default shipping method available, Sylius will return error 500.

**Steps to reproduce**  
- Create a cart and assign it with a shipping method
- Disable all shipping methods in system
- OrderShipmentProcessor will do nothing to that order, and then ShippingChargesProcessor will try to calculate price with the invalid shipping method and could cause error.

**Possible Solution**  
- Remove all shipments if no shipping method is available to this order.

### DIFF
--- a/src/Sylius/Component/Core/OrderProcessing/OrderShipmentProcessor.php
+++ b/src/Sylius/Component/Core/OrderProcessing/OrderShipmentProcessor.php
@@ -70,7 +70,8 @@ final class OrderShipmentProcessor implements OrderProcessorInterface
             $shipment = $this->getExistingShipmentWithProperMethod($order);
 
             if (null === $shipment) {
-                return;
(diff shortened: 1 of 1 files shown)


Arminek:
@kayue, thank you very much for your contribution. Could you provide a user story that will reproduce this bug?

Arminek:
Your fix also affects this story: ``` Scenario: Being alerted about no shipping method available ```. Could you take a look?

GSadee:
Closing in favour of #15170

PR #12877, opened 2021-08-10
[API] add original price to product variant

SirDomin:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Context/Api/Shop/ProductContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/ProductContext.php
@@ -166,7 +166,7 @@ final class ProductContext implements Context
     {
         $response = $this->responseChecker->getResponseContent($this->client->getLastResponse());
 
-        Assert::same($response['original_price'], null);
+        Assert::same($response['original_price'], $response['price']);
     }
 
     /**
--- a/src/Sylius/Behat/Context/Api/Shop/ProductContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/ProductContext.php
@@ -147,6 +147,29 @@ final class ProductContext implements Context
     }
 
     /**
+     * @Then /^I should see the product original price ("[^"]+")$/
+     */
+    public function iShouldSeeTheProductOriginalPrice(int $originalPrice): void
+    {
+        Assert::true(
+            $this->hasProductWithOriginalPrice(
+                [$this->responseChecker->getResponseContent($this->client->getLastResponse())],
+                $originalPrice
+            )
+        );
+    }
+
+    /**
+     * @Then I should not see any original price
+     */
+    public function iShouldNotSeeAnyOriginalPrice(): void
+    {
+        $response = $this->responseChecker->getResponseContent($this->client->getLastResponse());
+
+        Assert::same($response['original_price'], null);
+    }
+
+    /**
      * @Then /^I should see the (product "[^"]+") with price ("[^"]+")$/
      */
     public function iShouldSeeTheProductWithPrice(ProductInterface $product, int $price): void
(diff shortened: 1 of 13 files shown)


src/Sylius/Behat/Context/Api/Shop/ProductContext.php
```diff
@@ -146,6 +146,27 @@ public function iShouldSeeTheProductPrice(int $price): void
...
+                [$this->responseChecker->getResponseContent($this->client->getLastResponse())],
+                $originalPrice
+            )
+        );
+    }
+
+    /**
+     * @Then I should not see any original price
+     */
+    public function iShouldNotSeeTheProductOriginalPrice(): void
```
arti0090, line 165:
```suggestion
    public function iShouldNotSeeAnyOriginalPrice(): void
```

arti0090 (approved):

src/Sylius/Bundle/ApiBundle/Serializer/ProductVariantNormalizer.php
```diff
@@ -71,6 +71,12 @@ public function normalize($object, $format = null, array $context = [])
             unset($data['price']);
         }
 
+        $originalPrice = $this->priceCalculator->calculateOriginal($object, ['channel' => $this->channelContext->getChannel()]);
+
+        if ($originalPrice !== $data['price']) {
+            $data['original_price'] = $originalPrice;
+        }
```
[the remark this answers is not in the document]
AdamKasp: And do we really want to do it in this way? IMO JSON should always look the same way, if `originalPrice` is configured it should show this value if not it should show maybe `null`. From a frontend perspective JSON with a "random" body will cause unexpected errors (I always check the value of the field instead the existence of this field)

AdamKasp (changes requested):

src/Sylius/Behat/Context/Api/Shop/ProductVariantContext.php
```diff
@@ -0,0 +1,65 @@
...
+use Behat\Behat\Context\Context;
+use Sylius\Behat\Client\ApiClientInterface;
+use Sylius\Behat\Client\ResponseCheckerInterface;
+use Sylius\Component\Core\Model\ProductVariantInterface;
+use Webmozart\Assert\Assert;
+
+final class ProductVariantContext implements Context
+{
+    /** @var ApiClientInterface */
+    private $client;
```
Arminek, line 25: Can we start using typed properties?

SirDomin: sounds good to me

src/Sylius/Bundle/ApiBundle/Serializer/ProductVariantNormalizer.php
```diff
@@ -67,8 +67,13 @@ public function normalize($object, $format = null, array $context = [])
 
         try {
             $data['price'] = $this->priceCalculator->calculate($object, ['channel' => $this->channelContext->getChannel()]);
+            $originalPrice = $this->priceCalculator->calculateOriginal($object, ['channel' => $this->channelContext->getChannel()]);
+
+            $data['original_price'] = ($data['price'] === $originalPrice ? null : $originalPrice);
```
AdamKasp, line 72: but if `originalPrice` is equal to `price` we should show 2 same value for `price` and `originalPrice` with this code, you will overwrite `originalPrice` :(

what if you will only leave ` data['original_price'] = $originalPrice`

AdamKasp (changes requested):

AdamKasp (approved):

Tomanhez (approved):

AdamKasp:
Thanks, @SirDomin! :1st_place_medal:

src/Sylius/Behat/Context/Api/Shop/ProductContext.php
```diff
@@ -343,6 +366,17 @@ private function hasProductWithName(array $products, string $name): bool
         return false;
     }
 
+    private function hasProductWithOriginalPrice(array $products, int $originalPrice): bool
+    {
+        foreach ($products as $product) {
```
lchrusciel, line 371: Why do we have an array of products here? Also, this loop just checks if there is any product with `original_price`, not the particular, which seems bad

PR #12894, opened 2021-08-14
Make CheckoutResolver fail safe

vvasiloi:
| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Some time ago I encountered several issues with the CheckoutResolver failing because of some assumptions that were not met by my customizations. Unfortunately, I can't remember the context.
This PR makes it fail safe in case those assumptions are not met and covers it with PHPSpec tests.

### DIFF
--- a/UPGRADE-1.12.md
+++ b/UPGRADE-1.12.md
@@ -4,6 +4,9 @@
     constructor argument has been made optional and is `null` by default, subsequently the first argument of
     `sylius.form.type.attribute_type.select.choices_collection` has been removed.
 
+1. The default checkout resolving route pattern has been changed from `/checkout/.+` to
+   `%sylius.security.shop_regex%/checkout/.+` to reduce the probability of conflicts with other routes.
+
 # UPGRADE FROM `v1.12.9` TO `v1.12.10`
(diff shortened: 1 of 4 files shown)


vvasiloi:
Any ideas on how to test the checkout resolver pattern behavior?

vvasiloi:
I changed the type-hints, but 1.9 is still on PHP 7.3, so I changed the base to 1.10.
Should I leave it like that or change back to 1.9 and revert the type-hints?

jakubtobiasz:
Maybe we should add some logging 🤔. Usually when something goes wrong we throw an exception, but in this case when we want to be silent, let's let know the developer (or administrator) something happened 🤔?

WDYT @vvasiloi? Do you want to finish this PR (I promise to review it quicker this time 🤣), or should we take it over?

GSadee (approved):

GSadee:
Thanks, Victor and Jan! :1st_place_medal:

PR #12970, opened 2021-08-17
[DX] Add sylius prefixes to normalizers ALREADY_CALLED consts values to avoid name collision

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | kind of
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

While adding custom normalizer, it may be tempting to use the same name for the flag. This may lead to skipping part of normalization. It would be better to prefix our config keys with `sylius` and make official recommendation, that such keys in end projects should be prefixed with `app`


AdamKasp (approved):

GSadee (approved):

GSadee:
Thanks, Łukasz! :tada:

PR #12968, opened 2021-08-17
[API] AddressOrderHandler refactored to use customer provider

lchrusciel:
instead of logic duplication

| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | partially 
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | 
| License         | MIT

PR #12974, opened 2021-08-18
[Maintenance] Make all API Bundle services public and unify public notation

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/services/api_payment_methods.xml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/services/api_payment_methods.xml
@@ -16,6 +16,8 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd"
 >
     <services>
+        <defaults public="true" />
+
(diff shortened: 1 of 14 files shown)


stloyd:
Hmmmmmmm, will all the DI autowiring etc. why such change? It's against best practices of Symfony IIRC.

Arminek:
@lchrusciel Could you give more background on why we want to have this?

lchrusciel:
Hey @stloyd & @Arminek.

You are right, this PR is against Symfony Best practices. However, our internal rule is to make all services public, due to Sylius Resource Controller and testing with PHPUnit for example. In all these scenarios we are fetching services directly from the container. In the resource controller, the services are fetched to perform regular logic. In the case of PHPUnit, I just had a failing test at #12968. IIRC, Behat also requires us to have public services, but I'm not 100% sure about it. 

These changes should make us coherent with something, that we introduced in #9366. All our services should be public by default, unless we will write a new ADR

Arminek (approved):

GSadee (approved):

GSadee:
Thank you, Łukasz! :tada:

PR #12986, opened 2021-08-19
[CatalogPromotion] Creating simple catalog promotion only with code and name

Arminek:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | #12799 
| License         | MIT


features/promotion/managing_catalog_promotions/managing_catalog_promotions.feature
```diff
@@ -0,0 +1,15 @@
+@managing_catalog_promotions
+Feature: Managing catalog promotions
```
Arminek, line 2: Temporary, I want to have a single feature file.

src/Sylius/Behat/Context/Application/Admin/ManagingCatalogPromotionsContext.php
```diff
@@ -0,0 +1,62 @@
...
+        $this->client->addRequestData('name', $name);
+        $this->client->create();
+    }
+
+    /**
+     * @Then there should be :amount new catalog promotion on the list
+     */
+    public function thereShouldBeNewCatalogPromotionOnTheList(int $amount): void
+    {
+        Assert::count($this->messageBus->getDispatchedMessages(), $amount);
```
Arminek, line 49: Checking if the message has been dispatched during resource creation. @arti0090 You need to figure out how to hook into API platform event cycle to dispatch CatalogPromotionCreated. 🎉  Look for something like post persist [here](https://api-platform.com/docs/core/events/)

src/Sylius/Component/Promotion/Model/CatalogPromotion.php
```diff
@@ -0,0 +1,51 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Promotion\Model;
+
+use Sylius\Component\Resource\Model\ResourceInterface;
+
+class CatalogPromotion implements ResourceInterface
```
Zales0123, line 18: I believe, we should have a `CatalogPromotionInterface` for this class 🖖

src/Sylius/Component/Promotion/Model/CatalogPromotion.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    /** @psalm-suppress PropertyNotSetInConstructor */
+    protected int $id;
+    /** @psalm-suppress PropertyNotSetInConstructor */
+    protected string $name;
+    /** @psalm-suppress PropertyNotSetInConstructor */
```
Zales0123, line 24: I totally understand we need it, but in my perfect world I would like to keep PHP classes as configuration-free as possible 😄 Maybe putting it in the psalm.xml file would work? (on the other hand, it could result in a giga-file, impossible to manage 🤷‍♂️ )

src/Sylius/Component/Promotion/Model/CatalogPromotion.php
```diff
@@ -0,0 +1,51 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Component\Promotion\Model;
+
+use Sylius\Component\Resource\Model\ResourceInterface;
+
+class CatalogPromotion implements ResourceInterface
+{
+    /** @psalm-suppress PropertyNotSetInConstructor */
+    protected int $id;
```
Zales0123, line 21: To keep it consistent with other resources, it should probably be mixed type

src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius/sylius_promotion.yml
```diff
@@ -14,3 +14,6 @@ sylius_promotion:
             classes:
                 model: Sylius\Component\Core\Model\PromotionCoupon
                 repository: Sylius\Bundle\PromotionBundle\Doctrine\ORM\PromotionCouponRepository
+        catalog_promotion:
+            classes:
+                model: Sylius\Component\Promotion\Model\CatalogPromotion
```
Zales0123, line 19: Do we need it? It's already set in the `Configuration` file in the `PromotionBundle` and is not (yet) overridden in the `CoreBundle`

src/Sylius/Behat/Context/Application/Admin/ManagingCatalogPromotionsContext.php
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * @Then it should have :code code and :name name
+     */
+    public function itShouldHaveCodeAndName(string $code, string $name): void
+    {
+        /** @var CatalogPromotionCreated $actualMessage */
+        $actualMessage = $this->messageBus->getDispatchedMessages()[0];
+        Assert::eq($actualMessage->code, $code);
+        Assert::eq($actualMessage->name, $name);
```
Zales0123, line 60: As far as I see, we already have a `get` endpoint for Catalog Promotions [here](https://github.com/Sylius/Sylius/pull/12986/files#diff-664b75ee2acc7eb49590a963ef9b3beafe2cc3d8410ebb154182e152fae76f82R24), so we should probably be able to check this data via API?

Zales0123 (commented):
Good to see it started 💪

src/Sylius/Behat/Context/Application/Admin/ManagingCatalogPromotionsContext.php
Arminek: Yup, check `Api/Admin/ManagingCatalogPromotionsContext` this is just a quick idea of how to check that the message has been dispatched to unlock other tasks but it could be moved to PHPUnit as well.

src/Sylius/Bundle/CoreBundle/Resources/config/app/sylius/sylius_promotion.yml
Arminek: Yup, this is not needed

src/Sylius/Component/Promotion/Model/CatalogPromotion.php
Arminek: Yup, I agree we need to think about how to make it better.

src/Sylius/Behat/Context/Application/Admin/ManagingCatalogPromotionsContext.php
AdamKasp, line 60:
```suggestion
        Assert::same($actualMessage->code, $code);
        Assert::same($actualMessage->name, $name);
```
we should use a stronger comparison if it is possible

Zales0123: I wonder, maybe we should have a step `And this catalog promotion should be usable` (or sth similar) in which we would check the message dispatching? Because having a CP with code and name is not related to this message at all (as it's not created by consuming this event). But for sure, checking is the message dispatched gives us a huge value - ensuring, that it can be handled by some (yet) mysterious _catalog promotions system_ 🪨

AdamKasp (approved):

src/Sylius/Component/Promotion/Message/CatalogPromotionCreated.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Promotion\Message;
```
AdamKasp, line 14:
```suggestion
namespace Sylius\Component\Promotion\Event;
```
IMHO it should be in namespace `Event`, it will be more precise what it is.

src/Sylius/Component/Promotion/Model/CatalogPromotion.php
lchrusciel: Isn't nulling a way to go? Also, as this is resource model, these fields will be nullable anyway

src/Sylius/Behat/Context/Application/Admin/ManagingCatalogPromotionsContext.php
lchrusciel: One option would be to check even on the application level, while API response in API context. The problem with it, is how would we like to create catalog promotion on an application level

Zales0123: Currently this application context is not _de facto_ application context, as it uses the API client to throw API calls 💃 Maybe we should stand with only `@api` context and there have a step for checking the created CP data (API index) and _readiness for applications (check if event is thrown)

src/Sylius/Bundle/PromotionBundle/Subscriber/CatalogPromotionSubscriber.php
```diff
@@ -0,0 +1,56 @@
...
+        $method = $event->getRequest()->getMethod();
+
+        if (!$catalogPromotion instanceof CatalogPromotionInterface || Request::METHOD_POST !== $method) {
+            return;
+        }
+
+        $this->messageBus->dispatch(new CatalogPromotionCreated(
+            $catalogPromotion->getId(),
+            $catalogPromotion->getName(),
+            $catalogPromotion->getCode()
```
AdamKasp, line 53: do we need these 3 datas?  just `code` should be enough :)

src/Sylius/Behat/Context/Application/Admin/ManagingCatalogPromotionsContext.php
lchrusciel: The last problem with checking it in two steps for me is that checking for event and API call mixes our perspectives. I would prefer to go with a separate context. But that may be also a case for refactor

src/Sylius/Component/Promotion/Event/CatalogPromotionCreated.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Promotion\Event;
+
+final class CatalogPromotionCreated
```
AdamKasp, line 16:
```suggestion
class CatalogPromotionCreated
```
in sylius commands and events are public :)

src/Sylius/Behat/Context/Application/Admin/ManagingCatalogPromotionsContext.php
Zales0123: Then we either have a redundant context (application) that does not provide us with any real value, or the API context with one step that is not strictly API-related. In my opinion, the second option is less harmful and easier/faster to refactor in the future. Nevertheless - this event-related discussion is not valid in this PR any more 😄

src/Sylius/Behat/Resources/config/suites.yml
```diff
@@ -36,6 +36,7 @@ imports:
     - suites/api/promotion/applying_promotion_coupon.yml
     - suites/api/promotion/applying_promotion_rules.yml
     - suites/api/promotion/managing_promotions.yml
+    - suites/api/promotion/managing_catalog_promotions.yml
```
GSadee, line 39: Should be in an alphabetical order

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/CatalogPromotion.xml
```diff
@@ -0,0 +1,43 @@
...
+            <collectionOperation name="admin_get">
+                <attribute name="method">GET</attribute>
+                <attribute name="normalization_context">
+                    <attribute name="groups">admin:catalog_promotion:read</attribute>
+                </attribute>
+            </collectionOperation>
+
+            <collectionOperation name="admin_post">
+                <attribute name="method">POST</attribute>
+                <attribute name="denormalization_context">
```
GSadee, line 33: I wonder if there should be also `normalization_context` defined

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/CatalogPromotion.xml
```diff
@@ -0,0 +1,43 @@
...
+                    <attribute name="groups">admin:catalog_promotion:create</attribute>
+                </attribute>
+            </collectionOperation>
+        </collectionOperations>
+
+        <property name="id" identifier="true" writable="false" />
+        <property name="name" readable="true" writable="true" />
+        <property name="code" readable="true" writable="true" />
```
GSadee, line 41: Shouldn't code be an identifier for catalog promotions?

src/Sylius/Component/Promotion/Event/CatalogPromotionCreated.php
GSadee, line 16: Should this event be a part of this PR? It is not used anywhere

AdamKasp: it shouldn't, @Tomanhez moved event-related logic but probably forgot about this class :(

src/Sylius/Component/Promotion/Model/CatalogPromotion.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    /**
+     * @psalm-suppress PropertyNotSetInConstructor
+     * @var mixed
+     */
+    protected $id;
+    /** @psalm-suppress PropertyNotSetInConstructor */
+    protected string $name;
+    /** @psalm-suppress PropertyNotSetInConstructor */
+    protected string $code;
```
lchrusciel, line 26:
```suggestion
    /** @var mixed */
    protected $id;
    
    protected ?string $name = null;
    
    protected ?string $code = null;
```

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/CatalogPromotion.xml
Tomanhez: I think this is not needed here.

src/Sylius/Bundle/PromotionBundle/DependencyInjection/Configuration.php
```diff
@@ -90,6 +93,23 @@ private function addResourcesSection(ArrayNodeDefinition $node): void
...
                         ->end()
+                        ->arrayNode('catalog_promotion')
+                            ->addDefaultsIfNotSet()
+                            ->children()
+                                ->variableNode('options')->end()
+                                ->arrayNode('classes')
+                                    ->addDefaultsIfNotSet()
+                                    ->children()
+                                        ->scalarNode('model')->defaultValue(CatalogPromotion::class)->cannotBeEmpty()->end()
+                                        ->scalarNode('interface')->defaultValue(ResourceInterface::class)->cannotBeEmpty()->end()
```
vvasiloi, line 104: Wasn't `interface` deprecated?

src/Sylius/Component/Promotion/Model/CatalogPromotion.php
```diff
@@ -0,0 +1,49 @@
...
+    protected ?string $name = null;
+
+    protected ?string $code = null;
+
+    public function getId()
+    {
+        return $this->id;
+    }
+
+    public function getName(): string
```
vvasiloi, line 30:
```suggestion
    public function getName(): ?string
```

src/Sylius/Component/Promotion/Model/CatalogPromotion.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        return $this->name;
+    }
+
+    public function setName(string $name): void
```
vvasiloi, line 35:
```suggestion
    public function setName(?string $name): void
```

src/Sylius/Component/Promotion/Model/CatalogPromotion.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        $this->name = $name;
+    }
+
+    public function getCode(): string
```
vvasiloi, line 40:
```suggestion
    public function getCode(): ?string
```

src/Sylius/Component/Promotion/Model/CatalogPromotion.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        return $this->code;
+    }
+
+    public function setCode(string $code): void
```
vvasiloi, line 45:
```suggestion
    public function setCode(?string $code): void
```

src/Sylius/Component/Promotion/Model/CatalogPromotionInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Sylius\Component\Promotion\Model;
+
+use Sylius\Component\Resource\Model\ResourceInterface;
+
+interface CatalogPromotionInterface extends ResourceInterface
+{
+    public function getId();
+
+    public function getName(): string;
```
vvasiloi, line 22:
```suggestion
    public function getName(): ?string;
```

src/Sylius/Component/Promotion/Model/CatalogPromotionInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+    public function setName(string $name): void;
```
vvasiloi, line 24:
```suggestion
    public function setName(?string $name): void;
```

src/Sylius/Component/Promotion/Model/CatalogPromotionInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+    public function getCode(): string;
```
vvasiloi, line 26:
```suggestion
    public function getCode(): ?string;
```

src/Sylius/Component/Promotion/Model/CatalogPromotionInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+    public function setCode(string $code): void;
```
vvasiloi, line 28:
```suggestion
    public function setCode(?string $code): void;
```

arti0090 (approved):

src/Sylius/Behat/Context/Api/Admin/ManagingCatalogPromotionsContext.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->client->addRequestData('name', $name);
+        $this->client->create();
+    }
+
+    /**
+     * @Then there should be :amount new catalog promotion on the list
+     */
+    public function thereShouldBeNewCatalogPromotionOnTheList(int $amount): void
+    {
+        $x = $this->client->index();
```
Zales0123, line 48: 😱

src/Sylius/Behat/Context/Api/Admin/ManagingCatalogPromotionsContext.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * @Then it should have :code code and :name name
+     */
+    public function itShouldHaveCodeAndName(string $code, string $name): void
+    {
+        $expectedValues = ['code' => $code, 'name' => $name];
+        Assert::true(
+            $this->responseChecker->hasItemWithValues($this->client->index(), $expectedValues),
+            sprintf('Cannot find catalog promotions with %s in the list', implode(',', $expectedValues))
```
Zales0123, line 62: No need to implode the array that we create 3 lines above:

```php
Assert::true(
    $this->responseChecker->hasItemWithValues($this->client->index(), ['code' => $code, 'name' => $name]),
    sprintf('Cannot find catalog promotions with code "%s" and name "%s" in the list', $code, $name)
);
```

src/Sylius/Bundle/PromotionBundle/DependencyInjection/Configuration.php
Zales0123: Indeed https://github.com/Sylius/SyliusResourceBundle/pull/97 🖖 However, to keep consistency within the bundle, I think we can leave it and remove it altogether with other's when we upgrade to ResourceBundle 2.0

Zales0123: But still, it should be `CatalogPromotionInterace::class` 💃

tests/Api/DataFixtures/ORM/catalog_promotion.yaml
```diff
@@ -0,0 +1,4 @@
+Sylius\Component\Promotion\Model\CatalogPromotion:
+    catalog_promotion:
+        name: "catalog promotion"
+        code: "catalog_promotion"
```
Zales0123, line 4: We could come up with something more fancy and _real_ 😄

tests/Api/Admin/CatalogPromotionsTest.php
```diff
@@ -0,0 +1,46 @@
...
+            'GET',
+            sprintf('/api/v2/admin/catalog-promotions'),
+            [],
+            [],
+            $header
+        );
+
+        $response = $this->client->getResponse();
+
+        $this->assertResponse($response, 'admin/get_catalog_promotions_admin_response', Response::HTTP_OK);
```
GSadee, line 44:
```suggestion
        $this->assertResponse($response, 'admin/get_catalog_promotions_response', Response::HTTP_OK);
```

GSadee (approved):

GSadee:
Thank you, Arkadiusz! :1st_place_medal:

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/CatalogPromotion.xml
lchrusciel: But without it, we are skipping serialization groups in return, therefore we may expose some data by mistake. Should be added

tests/Api/Admin/CatalogPromotionsTest.php
```diff
@@ -0,0 +1,46 @@
...
+use Sylius\Tests\Api\JsonApiTestCase;
+use Sylius\Tests\Api\Utils\AdminUserLoginTrait;
+use Symfony\Component\HttpFoundation\Response;
+
+final class CatalogPromotionsTest extends JsonApiTestCase
+{
+    use AdminUserLoginTrait;
+
+    /** @test */
+    public function it_allows_admin_to_get_catalog_promotions(): void
```
lchrusciel, line 25: Can we also add a test for catalog promotion creation?

GSadee: It will be added here: https://github.com/Sylius/Sylius/pull/12994

PR #13019, opened 2021-08-27
[Bug] Fix default sylius config file path in Sylius 1.9

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

There is no such file as `app/config/packages/_sylius.yaml`. App prefix is not needed, because related configuration should be placed in: https://github.com/Sylius/Sylius-Standard/blob/1.10/config/packages/_sylius.yaml

This PR contains fixes to 1.8 as well, as I'm not sure if we want to merge then on previous branch

### DIFF
--- a/UPGRADE-1.8.md
+++ b/UPGRADE-1.8.md
@@ -1,6 +1,6 @@
 # UPGRADE FROM `v1.8.4` TO `v1.8.6`
 
-1. API is disabled by default, to enable it you need to set flag to ``true`` in ``app/config/packages/_sylius.yaml``:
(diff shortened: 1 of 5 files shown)


CoderMaggie (dismissed):

GSadee (approved):

GSadee:
Thanks, Łukasz! :tada:

PR #13022, opened 2021-08-27
[Bug] Fix default sylius config file path in Sylius master branch

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

There is no such file as `app/config/packages/_sylius.yaml`. App prefix is not needed, because related configuration should be placed in: https://github.com/Sylius/Sylius-Standard/blob/1.10/config/packages/_sylius.yaml

### DIFF
--- a/UPGRADE-API-1.11.md
+++ b/UPGRADE-API-1.11.md
@@ -15,7 +15,7 @@
         } 
    ```
    
-   To change the prefix you need to set parameter in ``app/config/packages/_sylius.yaml``:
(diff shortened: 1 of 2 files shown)


GSadee (approved):

GSadee:
Thanks, Łukasz! :tada:

PR #13042, opened 2021-09-01
[WIP][CustomValidator] Create base custom validator

Arminek:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Validator/Constraints/CustomConfiguration.php
+++ b/src/Sylius/Bundle/CoreBundle/Validator/Constraints/CustomConfiguration.php
@@ -1,0 +1,27 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+final class CustomConfiguration extends Constraint
+{
+    /** @var string */
+    public $message = 'sylius.custom_configuration';
+
+    public function validatedBy(): string
+    {
+        return 'sylius_custom_configuration';
+    }
+}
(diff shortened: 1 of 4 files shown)


src/Sylius/Bundle/CoreBundle/Validator/Constraints/CustomConfigurationValidator.php
```diff
@@ -0,0 +1,36 @@
...
+    private ConstraintResolver $constraintResolver;
+    private ValidatorInterface $validator;
+
+    public function __construct(ConstraintResolver $constraintResolver, ValidatorInterface $validator)
+    {
+        $this->constraintResolver = $constraintResolver;
+        $this->validator = $validator;
+    }
+
+    public function validate($value, Constraint $constraint)
```
AdamKasp, line 31:
```suggestion
    public function validate($value, Constraint $constraint): void
```

src/Sylius/Bundle/CoreBundle/Validator/Constraints/CustomConfiguration.php
```diff
@@ -0,0 +1,27 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+
+final class CustomConfiguration extends Constraint
+{
+    /** @var string */
+    public $message = 'sylius.custom_configuration';
```
GSadee, line 21:
```suggestion
    public string $message = 'sylius.custom_configuration';
```

src/Sylius/Bundle/CoreBundle/Validator/Constraints/AtLeastOneOf.php
```diff
@@ -0,0 +1,58 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\Validator\Constraints;
+
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\Constraints\Composite;
+
+/**
+ * @internal
```
Zales0123, line 20: I would definitely explain here (and in the validator) that these are services copied from Symfony, so we can use them while still supporting Symfony 4.4. Let's also make it explicit that they would be removed when dropping Symfony 4.4 support. This information is IMO crucial to merge it and avoid any confusion :) 🖖

Zales0123 (approved):

PR #13047, opened 2021-09-02
Chinese customer might have single word name

kayue:
| Q               | A
| --------------- | -----
| Branch?         | 1.9, 1.10 or master 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

Fix a rule that might not apply to Chinese customer


Arminek:
@kayue Could you add this case here features/account/registering_validation.feature to ensure that this will work?

jakubtobiasz:
Hi!
Thanks for rising this issue. We've found this PR doesn't cover all places where we validate the name length, and also this PR seems abandoned. We've created a task for it in our internal tracking system. If anybody wants to pick it and finish – feel free to do it.

PR #13071, opened 2021-09-08
[CatalogPromotion][API] Add validation for creating catalog promotion with actions

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | based on #13037
| License         | MIT

### DIFF
--- a/tests/Api/Admin/CatalogPromotionsTest.php
+++ b/tests/Api/Admin/CatalogPromotionsTest.php
@@ -161,6 +161,67 @@ final class CatalogPromotionsTest extends JsonApiTestCase
     }
 
     /** @test */
+    public function it_does_not_create_a_catalog_promotion_with_invalid_actions(): void
+    {
+        $this->loadFixturesFromFiles(['authentication/api_administrator.yaml', 'channel.yaml', 'product_variant.yaml']);
+        $header = $this->getLoggedHeader();
+
+        $this->client->request(
+            'POST',
+            '/api/v2/admin/catalog-promotions',
+            [],
+            [],
+            $header,
+            json_encode([
+                'name' => 'T-Shirts discount',
+                'code' => 'tshirts_discount',
+                'channels' => [
+                    '/api/v2/admin/channels/WEB',
+                ],
+                'actions' => [
+                    [
+                        'type' => 'invalid_type',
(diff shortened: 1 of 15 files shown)


tests/Api/Admin/CatalogPromotionsTest.php
```diff
@@ -160,6 +160,67 @@ public function it_does_not_create_a_catalog_promotion_with_taken_code(): void
...
+        $header = $this->getLoggedHeader();
+
+        $this->client->request(
+            'POST',
+            '/api/v2/admin/catalog-promotions',
+            [],
+            [],
+            $header,
+            json_encode([
+                'name' => 'T-Shirts discount',
```
arti0090, line 176: I think we should consider something for our contract tests that will allow us to use json files with request body. Because those tests are getting enormously huge with just this funky 💃 code.

tests/Api/Admin/CatalogPromotionsTest.php
```diff
@@ -160,6 +160,67 @@ public function it_does_not_create_a_catalog_promotion_with_taken_code(): void
...
+                    'label' => 'T-Shirts discount',
+                    'description' => '50% discount on every T-Shirt',
+                ]]
+            ], JSON_THROW_ON_ERROR)
+        );
+
+        $response = $this->client->getResponse();
+
+        $this->assertResponse(
+            $response,
```
arti0090, line 218:
```suggestion
        $this->assertResponse(
             $this->client->getResponse(),
```

arti0090 (approved):

lchrusciel (approved):

tests/Api/Admin/CatalogPromotionsTest.php
GSadee: We can probably use json files without problems, but then these tests will become less readable, so I'm not 100% convinced. We should think about that and probably refactor our contracts tests in a separate PR

tests/Api/Admin/CatalogPromotionsTest.php
GSadee: I'll change that in another PR

PR #13092, opened 2021-09-13
[CatalogPromotion][Behat] Add scenarios for reapplying catalog promotions after editing its channels

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/features/promotion/applying_catalog_promotions/reapplying_catalog_promotions_after_editing_its_channels.feature
+++ b/features/promotion/applying_catalog_promotions/reapplying_catalog_promotions_after_editing_its_channels.feature
@@ -12,23 +12,24 @@ Feature: Reapplying catalog promotions after editing its channels
         And "PHP T-Shirt" variant priced at "$30.00" in "Web-GB" channel
         And this product is available in "Web-US" channel and "Web-GB" channel
         And there is a catalog promotion "Winter sale" available in "Web-US" channel that reduces price by "30%" and applies on "PHP T-shirt" variant
+        And I am logged in as an administrator
 
     @todo
     Scenario: Removing applied catalog promotion after removing its channel
         When I make this catalog promotion unavailable in the "Web-US" channel
-        Then the visitor should see "$20.00" as the price of the product "T-Shirt" in the "Web-US" channel
-        And the visitor should see "$20.00" as the original price of the product "T-Shirt" in the "Web-US" channel
+        Then the visitor should see "$20.00" as the price of the "T-Shirt" product in the "Web-US" channel
+        And the visitor should see "$20.00" as the original price of the "T-Shirt" product in the "Web-US" channel
 
     @todo
     Scenario: Reapplying catalog promotion after adding new channel to them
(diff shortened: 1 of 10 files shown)


features/promotion/applying_catalog_promotions/reapplying_catalog_promotions_after_editing_its_channels.feature
```diff
@@ -0,0 +1,39 @@
...
+        And the store has a "T-Shirt" configurable product
+        And this product has "PHP T-Shirt" variant priced at "$20.00" in "Web-US" channel
+        And "PHP T-Shirt" variant priced at "$30.00" in "Web-GB" channel
+        And this product is available in "Web-US" channel and "Web-GB" channel
+        And there is a catalog promotion "Winter sale" available in "Web-US" channel that reduces price by "30%" and applies on "PHP T-shirt" variant
+        And I am currently in the "Web-US" channel
+
+    @todo
+    Scenario: Removing applied catalog promotion after removing its channel
+        Given this catalog promotion is no longer available in "Web-US" channel
```
GSadee, line 19: Won't the behaviour be better covered if this step looks like:
```suggestion
        Given the administrator removed "Web-US" channel from the "Winter sale" catalog promotion
```
?
And then it would be performed in the API/UI instead of removing the channel from entity in setup context. Wdyt?

SirDomin (approved):

features/promotion/applying_catalog_promotions/reapplying_catalog_promotions_after_editing_its_channels.feature
AdamKasp: for me it is ok, but IMHO this new step should be `when` because, channel removing is an action that causes changes in the system, so `when admin remove a channel, then promotions are reapplied

Zales0123: From the technical PoV, it makes sense to conduct an actual **action** rather than programmatically removing the channel from the entity.

*But* on the other hand, these scenarios are written from the **Visitor** perspective. So the visitor does not care *how* did it happen that the promotion is no longer available (either the administrator edits it, there is a problem in the system or whatever else). They just see the result of the action. additionally, it would be tricky (but, of course, possible) to switch between Admin (to conduct an action) and Visitor (to see the result) context, in both UI and API implementation.

So, summing this up: I'm sure Visitor should have no knowledge about the administrator's actions, so I would leave it as it is right know 🖖

GSadee: After small discussion, I decided to change the actor of these scenarios

AdamKasp (approved):

Tomanhez (approved):

AdamKasp:
Thank you, Grzegorz! :tada:

PR #13098, opened 2021-09-14
[CatalogPromotions] Add product to cart with discounted catalog price

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/features/cart/shopping_cart/adding_product_with_discounted_catalog_price_to_cart.feature
+++ b/features/cart/shopping_cart/adding_product_with_discounted_catalog_price_to_cart.feature
@@ -1,0 +1,22 @@
+@shopping_cart
+Feature: Adding a simple product with discounted catalog price to the cart
+    In order to select products with proper price
+    As a Visitor
+    I want to be able to add simple products with discounted catalog price to the cart
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "Mug" priced at "$40"
+        And the store has a product "T-Shirt" priced at "$20"
+        And there is a catalog promotion "Winter sale" that reduces price by "25%" and applies on "T-Shirt" variant
+
+    @todo
+    Scenario: Adding a simple product with discounted catalog price to the cart
+        When I add product "T-Shirt" to the cart
+        And I add product "Mug" to the cart
+        Then I should be on my cart summary page
+        And I should be notified that the product has been successfully added
+        And I should see "T-Shirt" with unit price "$15.00" in my cart
+        And I should see "T-Shirt" with original price "$20.00" in my cart
+        And I should see also "Mug" with unit price "$40.00" in my cart
+        And I should see "Mug" with original price "$40.00" in my cart
--- a/features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
+++ b/features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
@@ -1,0 +1,18 @@
+@shopping_cart
+Feature: Adding a product with selected variant with discounted catalog price to the cart
+    In order to select products with proper price
+    As a Visitor
+    I want to be able to add products with selected variants to cart
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "T-Shirt"
+        And this product has "PHP T-Shirt" variant priced at "$20"
+        And there is a catalog promotion "Winter sale" that reduces price by "25%" and applies on "PHP T-Shirt" variant
+
+    @todo
+    Scenario: Adding a product with multiple variants with discounted catalog price to the cart
+        When I add "PHP T-Shirt" variant of this product to the cart
+        Then I should be on my cart summary page
+        And I should be notified that the product has been successfully added
+        And I should see "T-Shirt Banana" with unit price "$15.00" in my cart
--- a/features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
+++ b/features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
@@ -8,11 +8,14 @@ Feature: Adding a product with selected variant with discounted catalog price to
         Given the store operates on a single channel in "United States"
         And the store has a product "T-Shirt"
         And this product has "PHP T-Shirt" variant priced at "$20"
+        And the store has a product "Keyboard"
+        And this product has "RGB Keyboard" variant priced at "$40"
         And there is a catalog promotion "Winter sale" that reduces price by "25%" and applies on "PHP T-Shirt" variant
 
     @todo
-    Scenario: Adding a product with multiple variants with discounted catalog price to the cart
-        When I add "PHP T-Shirt" variant of this product to the cart
-        Then I should be on my cart summary page
-        And I should be notified that the product has been successfully added
-        And I should see "T-Shirt Banana" with unit price "$15.00" in my cart
+    Scenario: Adding multiple product variants with discounted price by catalog promotion catalog to the cart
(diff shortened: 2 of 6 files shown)


features/cart/shopping_cart/adding_product_with_discounted_catalog_price_to_cart.feature
```diff
@@ -0,0 +1,28 @@
...
+
+    @todo
+    Scenario: Adding a simple product with discounted catalog price to the cart
+        When I add this product to the cart
+        Then I should be on my cart summary page
+        And I should be notified that the product has been successfully added
+        And I should see "PHP T-Shirt" with unit price "$15.00" in my cart
+
+    @todo
+    Scenario: Adding a simple product with discounted catalog price to the cart as a logged in customer
```
Zales0123, line 23: I'm wondering if this scenario is needed 🤔 It does not have another result than the previous one, so does not bring a real value checking it

features/cart/shopping_cart/adding_product_with_discounted_catalog_price_to_cart.feature
```diff
@@ -0,0 +1,28 @@
...
+        And the store has a "Mug" configurable product
+        And this product has "PHP Mug" variant priced at "$5.00"
+        And there is a catalog promotion "Winter sale" that reduces price by "25%" and applies on "PHP T-Shirt" variant
+
+    @todo
+    Scenario: Adding a simple product with discounted catalog price to the cart
+        When I add this product to the cart
+        Then I should be on my cart summary page
+        And I should be notified that the product has been successfully added
+        And I should see "PHP T-Shirt" with unit price "$15.00" in my cart
```
Zales0123, line 20: Shouldn't we check do we have both original and discounted price shown in the UI/API response?

Zales0123: And probably it would be good to also add "PHP Mug" in this scenario to see if it **does not** have the promotion applied

features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
```diff
@@ -0,0 +1,18 @@
+@shopping_cart
+Feature: Adding a product with selected variant with discounted catalog price to the cart
+    In order to select products with proper price
+    As a Visitor
+    I want to be able to add products with selected variants  to cart
```
Zales0123, line 5: Doubled space before `to cart`

features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
```diff
@@ -0,0 +1,18 @@
...
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "T-Shirt"
+        And this product has "PHP T-Shirt" variant priced at "$20"
```
Zales0123, line 10: To make it essentially different from the previous feature, I would add another variant here - to point out the discount is applied only on one of them

features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
```diff
@@ -0,0 +1,18 @@
...
+        And there is a catalog promotion "Winter sale" that reduces price by "25%" and applies on "PHP T-Shirt" variant
+
+    @todo
+    Scenario: Adding a product with multiple variants with discounted catalog price to the cart
```
Zales0123, line 14: `**multiple** variants` see above 😄

features/product/viewing_products/viewing_different_discounted_price_for_different_product_variants.feature
```diff
@@ -23,3 +23,11 @@ Feature: Viewing different discounted price for different product variants
...
         Then the product variant price should be "$12.55"
         And the product original price should be "$20.00"
+
+    @todo
+    Scenario: Viewing a detailed page with product's discounted catalog price for different variant
+        Given there is a catalog promotion "For apple drinks" that reduces price by "30%" and applies on "Wyborowa Apple" variant
+        When I view product "Wyborowa Vodka"
+        And I select "Wyborowa Apple" variant
+        Then the product variant price should be "$8,79"
+        And the product original price should be "$20.00"
```
Zales0123, line 33: I wonder, should we use `price/original price` words within the shop context. For a Customer/Visitor it's more like a `discounted price/price`, wdyt?

features/product/viewing_products/viewing_different_discounted_price_for_different_product_variants.feature
```diff
@@ -23,3 +23,11 @@ Feature: Viewing different discounted price for different product variants
         And I select "Wyborowa Apple" variant
         Then the product variant price should be "$12.55"
         And the product original price should be "$20.00"
+
+    @todo
+    Scenario: Viewing a detailed page with product's discounted catalog price for different variant
```
AdamKasp, line 28: The task related to this behats is about prices in cart not on show page.

features/product/viewing_products/viewing_different_discounted_price_for_different_product_variants.feature
arti0090: IMO the most important thing here is still `Price`, and this is something that really bothers Mr. Customer. The way how we say what was the price before is not so important - does the customer cares if it was called original? I don't think so. And in this case I would think about API first - where we are returning original price 🎉

features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
arti0090: I have just added the second product and variant (not discounted) and add it to checks 🎉

features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
```diff
@@ -0,0 +1,21 @@
...
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a product "T-Shirt"
+        And this product has "PHP T-Shirt" variant priced at "$20"
+        And the store has a product "Keyboard"
+        And this product has "RGB Keyboard" variant priced at "$40"
+        And there is a catalog promotion "Winter sale" that reduces price by "25%" and applies on "PHP T-Shirt" variant
+
+    @todo
+    Scenario: Adding multiple product variants with discounted variant by catalog promotion catalog to the cart
```
AdamKasp, line 16:
```suggestion
    Scenario: Adding multiple product variants with discounted price by catalog promotion catalog to the cart
```

AdamKasp (approved):

features/product/viewing_products/viewing_different_discounted_price_for_different_product_variants.feature
Zales0123: Yes.... but! 😄 API-first is not an argument, as Behat scenarios are and should be always context-agnostic. What we should do in Behat scenarios is imitating the user behaviour. So, does the Customer sees the crossed-out price with a new one beside and thinks "the original price is X, but a price is now Y!"? Or it's more like "the price before was X, but a new, discounted price is Y!" 🚀 🐎 

Of course, it's not a blocker for merge, but I believe we should really focus on Customer's behaviour and their **vocabulary**, to write perfect Behat scenarios :) 🖖

Zales0123 (approved):

Zales0123:

PR #13103, opened 2021-09-14
[UI][Catalog Promotion] Add discounted product to cart

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         |  master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| License         | MIT

Screenshot 2021-09-15 at 11 22 09

### DIFF
--- a/features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
+++ b/features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
@@ -16,8 +16,8 @@ Feature: Adding a product with selected variant with discounted catalog price to
 
     @ui
     Scenario: Adding multiple product variants with discounted price by catalog promotion catalog to the cart
-        Given I add "PHP T-Shirt" variant of product "T-Shirt" to the cart
+        When I add "PHP T-Shirt" variant of product "T-Shirt" to the cart
         And I add "RGB Keyboard" variant of product "Keyboard" to the cart
-        When I check details of my cart
+        And I check details of my cart
         Then I should see "T-Shirt" with unit price "$15.00" in my cart
         And I should see "Keyboard" with unit price "$40.00" in my cart
--- a/features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
+++ b/features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
@@ -6,16 +6,18 @@ Feature: Adding a product with selected variant with discounted catalog price to
 
     Background:
         Given the store operates on a single channel in "United States"
-        And the store has a product "T-Shirt"
+        And the store has a "T-Shirt" configurable product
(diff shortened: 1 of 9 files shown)


features/cart/shopping_cart/adding_product_with_variants_with_discounted_catalog_price_to_cart.feature
```diff
@@ -6,16 +6,18 @@ Feature: Adding a product with selected variant with discounted catalog price to
...
         And this product has "RGB Keyboard" variant priced at "$40"
+        And this product has "Pink Keyboard" variant priced at "$300"
         And there is a catalog promotion "Winter sale" that reduces price by "25%" and applies on "PHP T-Shirt" variant
 
-    @todo
+    @ui
     Scenario: Adding multiple product variants with discounted price by catalog promotion catalog to the cart
-        Given I add product "PHP T-Shirt" to the cart
-        And I add product "RGB Keyboard" to the cart
+        Given I add "PHP T-Shirt" variant of product "T-Shirt" to the cart
```
GSadee, line 19:
```suggestion
        When I add "PHP T-Shirt" variant of product "T-Shirt" to the cart
```

GSadee (approved):
Can you add an additional scenario to check the price of item in the cart when there is both a catalog and a cart promotion?

src/Sylius/Bundle/ShopBundle/Resources/views/Cart/Summary/_item.html.twig
```diff
@@ -1,12 +1,16 @@
...
 
 {% set product_variant = item.variant %}
+{% set pricing_for_active_channel = product_variant.getChannelPricingForChannel(cart.channel) %}
 
 <tr {{ sylius_test_html_attribute('cart-product-row', item.productName) }}>
     <td class="single line" {{ sylius_test_html_attribute('cart-item', loop_index|default(null) ) }}>
         {% include '@SyliusShop/Product/_info.html.twig' with {'variant': product_variant} %}
     </td>
     <td class="right aligned">
+        {% if pricing_for_active_channel.originalPrice is not null %}
```
GSadee, line 11: You should check here also if there is a catalog promotion applied (I'm not sure because the administrator can set also the original price manually) or if the original price is greater than unit price

AdamKasp: as you said admin also can change the original price, so I do not see a reason to check if CP is applied, I can check If the original price is greater than the unit price but IMO this check should be validated on `originalPrice` input in the admin Panel

GSadee: I partially agree, but displaying the crossed-out original price only makes sense if it is greater than the current price. For now, if I'm correct, we don't have validation for the original price, so we should cover it here, especially in the case when the original price === price, what is a correct situation but has no sense to crossed-out it here.

GSadee (approved):

PR #13112, opened 2021-09-15
[Catalog Promotion] Cart catalog promotion becomes ineligible

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         |  master 
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| License         | MIT


features/checkout/seeing_order_summary/seeing_current_prices_of_products_after_catalog_promotion_becomes_ineligible.feature
```diff
@@ -0,0 +1,23 @@
...
+        And this product has "PHP T-Shirt" variant priced at "$20.00" in "United States" channel
+        And the store ships everywhere for free
+        And the store allows paying offline
+        And there is a catalog promotion "Winter sale" available in "United States" channel that reduces price by "25%" and applies on "PHP T-Shirt" variant
+        And I am a logged in customer
+
+    @ui @api
+    Scenario: Processing order with valid prices after catalog promotion becomes ineligibly
+        Given I have "PHP T-Shirt" variant of this product in the cart
+        When the administrator makes this catalog promotion unavailable in the "United States" channel
```
Zales0123, line 19: I'm still more for something like `And the "Winter sale" catalog promotion is no longer available`, as the Customer has no info about neither administrator's actions nor the concept of _current channel_ 🚀

features/checkout/seeing_order_summary/seeing_current_prices_of_products_after_catalog_promotion_becomes_ineligible.feature
```diff
@@ -0,0 +1,23 @@
+@checkout
+Feature: Seeing current prices of products after catalog promotion becomes ineligible
+    In order to buy products in its correct prices
+    As a Customer
+    I want to have products with its current prices in the cart
```
Zales0123, line 5: `I want to have products with their original prices in the cart`?

features/checkout/seeing_order_summary/seeing_current_prices_of_products_after_catalog_promotion_becomes_ineligible.feature
```diff
@@ -0,0 +1,23 @@
...
+        And I am a logged in customer
+
+    @ui @api
+    Scenario: Processing order with valid prices after catalog promotion becomes ineligibly
+        Given I have "PHP T-Shirt" variant of this product in the cart
+        When the administrator makes this catalog promotion unavailable in the "United States" channel
+        And I specified the billing address as "Ankh Morpork", "Frost Alley", "90210", "United States" for "Jon Snow"
+        And I proceed through checkout process
+        Then I should be on the checkout summary step
+        And I should see product "T-Shirt" with unit price "$20.00"
```
Zales0123, line 23: Cannot we check it before the final stap? Even on the cart itself?

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -1073,6 +1080,18 @@ public function iTryToChangeQuantityToOfProductFromTheCart(int $quantity, Produc
         $this->putProductToCart($product, $tokenValue, $quantity);
     }
 
+    /**
+     * @When /^the administrator makes (this catalog promotion) unavailable in the ("([^"]+)" channel)$/
+     */
+    public function theAdministratorMakesThisCatalogPromotionUnavailableInTheChannel(
```
Zales0123, line 1086: It's not an API action, so should be placed in `Setup` context

Zales0123: Or be changed into API-related implementation 💃

features/checkout/seeing_order_summary/seeing_current_prices_of_products_after_catalog_promotion_becomes_ineligible.feature
AdamKasp: TBH yes we can :D, we need to run `order Processor` after admin changed something so each step after addressing is valid, going through the entire checkout was easy in behat and also it checks what I want :)

features/checkout/seeing_order_summary/seeing_current_prices_of_products_after_catalog_promotion_becomes_ineligible.feature
AdamKasp: the current sentence has more sense for me because in this scenarios I want to have a valid and current price, for customer, it is irrelevant if it is original or discounted (by different catalog promotion for example)

SirDomin (approved):

features/checkout/seeing_order_summary/seeing_current_prices_of_products_after_catalog_promotion_becomes_ineligible.feature
Zales0123: I was referring that the price that Customer sees is *always* current for they :) but it's not relevant for know, I would only change `its` to `their` 🖖

features/checkout/seeing_order_summary/seeing_current_prices_of_products_after_catalog_promotion_becomes_ineligible.feature
Zales0123: Maybe we could just update the cart? It's not a blocker for merge, but with all these steps it seems like it's necessary to go through the checkout to see the result. And there is `in the cart` in the scenario's description 💃

SirDomin:
Thank you, Adam! :tada:

PR #13117, opened 2021-09-16
[API] Remove phoneNumber from Customer command and handler

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no (kind of)
| Deprecations?   | no 
| Related tickets | 
| License         | MIT


Zales0123 (commented):
It would be great to explain in the PR description **why** we do that 💃

lchrusciel:
The `phoneNumber` field was never exposed via serialization, so it was never used from the API perspective. What is more, we do not have any behat scenario covering this value. As a result, it is dead code for me, thats why I'm proposing removal of it.

Zales0123 (approved):

Zales0123:
Thank you, Łukasz! :tada:

PR #13126, opened 2021-09-20
Fix #13123, Only use products total in OrderTotalRuleChecker to select shipping methods

### LINKED ISSUE #13123: Incoherent total price in OrderTotalRuleChecker
**Sylius version affected**: 1.8, 1.9, 1.0

**Description**  
The rules `OrderTotalGreaterThanOrEqualRuleChecker` and `OrderTotalGreaterThanOrEqualRuleChecker` use the order total to check if a shipping method is available. Depending on when the rule is called, the total can be different for the same cart. For example:

* When computing the list of shipping method for the checkout form the price is the complete total (products + shipping + taxes + promotions).
* During the order processing steps, the price is the total without the adjustements (only products without taxes) since it runs just after OrderAdjustementsClearer.

It results in incoherent shipping method throughout the checkout.

**Steps to reproduce**  
1. Create one shipping method A available if total is bellow 100 with a price of 10.
1. Create a second shipping method B available if total is above 100.
1. Create a cart where items total is bellow 100 but above 91, default shipping method should therefore be A.
1. On the shipping method selection form, only shipping method B is available (items total + estimated shipping cost of A > 100). Only A should be available in this form.

**Possible Solution**  
User `$order->getItemsTotal()` instead of `$order->getTotal()`.

I'm not sure if only products should contribute to the shipping method rule checker. Maybe taxes and promotion should contribute too. But at least shipping method cost should not contribute to shipping method rule checker.

Maybe also update the priority of the steps of order processors to recompute products price and taxes before selecting the shipping method as mentioned in https://github.com/Sylius/Sylius/issues/12554.

mdevlamynck:
Only use products total in OrderTotalRuleChecker to select shipping methods. Not sure if it should be a breaking change or not.

| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no?
| Deprecations?   | no
| Related tickets | fixes #13123
| License         | MIT

### DIFF
--- a/features/checkout/consistent_shipping_method_selection.feature
+++ b/features/checkout/consistent_shipping_method_selection.feature
@@ -1,0 +1,30 @@
+@checkout
+Feature: Selected shipping method stays consistent throughout checkout
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And the store has a tax category "tax category"
+        And the store has "VAT" tax rate of 20% for "tax cageroy" within the "United States" zone
+        And the store has a product "Mantis blade" priced at "$91"
+        And the product "Mantis blade" belongs to "tax category" tax category
+        And the store has "A" shipping method with "$10" fee
+        And this shipping method is only available for orders under or equal to "$100"
+        And the store has "B" shipping method with "$0" fee
+        And this shipping method is only available for orders over or equal to "$100"
+        And the store allows paying offline
+        And there is a customer "John Doe" identified by an email "john@example.com" and a password "secret"
+
+    @ui
+    Scenario: Method A should be available and not method B
+        Given I have product "Mantis blade" in the cart
+        When I complete addressing step with email "john@example.com" and "United States" based billing address
+        Then I should see "A" shipping method
+        And I should not see "B" shipping method
+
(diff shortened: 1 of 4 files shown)


Zales0123:
Hello Matthias and thank you very much for this fix! One more question before we can merge it. Do you think you would be able to construct a Behat scenario reproducing this bug? We would then be sure that it's red before and green after this change :) You've written so detailed reproduction steps in the [related issue](https://github.com/Sylius/Sylius/issues/13123), that they almost seem like a scenario 😄 🚀

mdevlamynck:
@Zales0123 Done :)

Snowbaha:
Hello, when do you plan to merge it?

Snowbaha (approved):
I confirm the fix!

Snowbaha:
Hello, Any news from the fix? Why it is not merge yet?

stloyd:
@lchrusciel @Zales0123 This seems correct, please check it.

CoderMaggie:
Replaced by #13870

PR #13154, opened 2021-09-30
[API] Do not allow to address empty cart

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

After addressing the empty cart and adding an item to it, there is a cart with some products but payment_skipped as payment state :dancer: *In theory*, it's not a problem, as after completion payment state is correct (awaiting payment), but if anything happens about payment before checkout completion, it can lead to some nasty errors. And, conceptually, addressing empty cart does not seems as something correct 🚀

### DIFF
--- a/features/checkout/addressing_order/order_address_validation.feature
+++ b/features/checkout/addressing_order/order_address_validation.feature
@@ -16,7 +16,7 @@ Feature: Order address validation
         When the visitor specify the email as "jon.snow@example.com"
         And the visitor try to specify the incorrect billing address as "Ankh Morpork", "Frost Alley", "90210", "United Russia" for "Jon Snow"
         And the visitor complete the addressing step
-        Then I should be notified that "United Russia" country does not exist
+        Then they should be notified that "United Russia" country does not exist
 
     @api
     Scenario: Trying to add address without country to the cart by the visitor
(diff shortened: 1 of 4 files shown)


features/checkout/addressing_order/order_address_validation.feature
```diff
@@ -16,12 +16,20 @@ Feature: Order address validation
         When the visitor specify the email as "jon.snow@example.com"
         And the visitor try to specify the incorrect billing address as "Ankh Morpork", "Frost Alley", "90210", "United Russia" for "Jon Snow"
         And the visitor complete the addressing step
-        Then I should be notified that "United Russia" country does not exist
+        Then they should be notified that "United Russia" country does not exist
```
arti0090, line 19: Why `they`? When the context changed from singular to plural?

arti0090 (approved):

features/checkout/addressing_order/order_address_validation.feature
Zales0123: It's [singular they](https://en.wikipedia.org/wiki/Singular_they) 💃

GSadee (approved):
But the steps in these scenarios should be refactored from `the visitor` to `I` perspective

GSadee:
Thanks, Mateusz! :tada:

PR #13160, opened 2021-09-30
Remove types in public and protected properties in not final classes

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

It went too smooth :smile: With the v1.10.3 tag we released code transformed to use PHP7.4 syntax (typed properties). **However** there is one problem with such an approach. If the property is overridable (so is either public or protected and its class is not final), it has to have a type defined also in the child class. Even though overriding properties is much less common than overriding functions (and most of our classes are final :dancer:), it's technically possible these changes could break the applications after upgrade to Sylius v1.10.3.

I then revert this change for non-private properties from not final classes. These are mostly models from components, validator constraints and API commands.

Live long and prosper 🖖

### DIFF
--- a/src/Sylius/Component/Core/Model/Order.php
+++ b/src/Sylius/Component/Core/Model/Order.php
@@ -29,50 +29,86 @@ use Webmozart\Assert\Assert;
 
 class Order extends BaseOrder implements OrderInterface
 {
-    protected ?CustomerInterface $customer = null;
+    /**
+     * @var \Sylius\Component\Core\Model\CustomerInterface|null
+     */
+    protected $customer;
 
-    protected ?ChannelInterface $channel = null;
+    /**
+     * @var \Sylius\Component\Core\Model\ChannelInterface|null
+     */
+    protected $channel;
 
-    protected ?AddressInterface $shippingAddress = null;
+    /**
+     * @var AddressInterface|null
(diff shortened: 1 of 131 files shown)


src/Sylius/Component/Core/Model/Order.php
```diff
@@ -29,50 +29,86 @@
 
 class Order extends BaseOrder implements OrderInterface
 {
-    protected ?CustomerInterface $customer = null;
+    /**
+     * @var \Sylius\Component\Core\Model\CustomerInterface|null
```
lchrusciel, line 33: We should simplify these imports

src/Sylius/Component/Currency/Model/Currency.php
```diff
@@ -23,7 +23,10 @@ class Currency implements CurrencyInterface
     /** @var mixed */
     protected $id;
 
-    protected ?string $code = null;
+    /**
+     * @var string|null
```
lchrusciel, line 27: We have mixed notation for brought backed properties

lchrusciel (approved):

PR #13166, opened 2021-10-04
fix(translation): Useful `%stock%` value for translated message

chadyred:
Useful `%stock%` value for translated message

| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | no/yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT


chadyred:
Need review plz :D

vvasiloi:
@chadyred can you give an example where this it's useful?
I think there might be a confusion here, because the value you pass is not the stock quantity, as the name suggests, but the requested quantity that's validated against the stock quantity.

vvasiloi:
@chadyred you can probably  use `$stockable->getOnHand()` instead.
I just wanted to know an example use case so I can suggest a proper solution. 😄

PR #13200, opened 2021-10-13
[WIP][Catalog Promotion] Price to original price migration

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes?
| BC breaks?      | probably no
| License         | MIT

### DIFF
--- a/docs/book/plugins/guide/specification.rst
+++ b/docs/book/plugins/guide/specification.rst
@@ -26,7 +26,7 @@ Let's start with describing how **marking a product variant available on demand*
         Background:
             Given the store operates on a single channel in "United States"
             And the store has a "Iron Man Suite" configurable product
-            And the product "Iron Man Suite" has a "Mark XLVI" variant priced at "$400000"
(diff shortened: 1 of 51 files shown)


loevgaard:
@Arminek: I am not sure I agree. The word 'original' means something 'that was' in this case. If you have a price of `€100`, then it _wasn't_ `€100`, it _is_ `€100`. In my opinion it's definitional wrong.

If you will continue down this path, please consider the user experience in this PR also. When setting prices, people are used to setting the `price` only, now they have to set two fields. Could we have a form listener that sets the original price to price if original price is null?

Arminek:
Sure we could consider adding 3rd price ```DiscountedPrice``` for migration time. But I am afraid of having three prices.

Arminek:
@AdamKasp We do not need to do it after a conversation with @loevgaard on slack. We should set the original price during catalog promotion processing only for prices that do not have the original price set. 🎉  Thank you very much for participation @loevgaard @Prometee @igormukhingmailcom @stloyd @Roshyo you are awesome!

I was blind now I see xD.

PR #13237, opened 2021-10-22
BUGFIX: Sylius#12225 Update routing and refactor parameter name

### LINKED ISSUE #12225: Admin routing for orders misses /orders prefix for two routes
https://github.com/Sylius/Sylius/blob/e6c9cb6318bc806474427e0fca84c25515a4802d/src/Sylius/Bundle/AdminBundle/Resources/config/routing/order.yml#L123

```diff
    --- path: /{id}/resend-confirmation-email
    +++ path: /orders/{id}/resend-confirmation-email
```

https://github.com/Sylius/Sylius/blob/e6c9cb6318bc806474427e0fca84c25515a4802d/src/Sylius/Bundle/AdminBundle/Resources/config/routing/order.yml#L98

```diff
    --- path: /{id}/ship
    +++ path: /orders/{id}/ship
```

Otherwise you will get routes like this
image

alex-boehm:
| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #12225
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Action/ResendOrderConfirmationEmailAction.php
+++ b/src/Sylius/Bundle/AdminBundle/Action/ResendOrderConfirmationEmailAction.php
@@ -54,7 +54,7 @@ final class ResendOrderConfirmationEmailAction
 
     public function __invoke(Request $request): Response
     {
-        $orderId = $request->attributes->get('id');
+        $orderId = $request->attributes->get('orderId');
 
         if (!$this->csrfTokenManager->isTokenValid(new CsrfToken($orderId, (string) $request->query->get('_csrf_token')))) {
             throw new HttpException(Response::HTTP_FORBIDDEN, 'Invalid csrf token.');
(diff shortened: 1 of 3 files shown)


coldic3:
Hey @alex-boehm,
thanks a lot for your contribution :dancer: Unfortunately, it cannot be done that way because your proposed change breaks our BC policy. We would like to keep the current route path and add an alias to the current routes.

diimpp:
Hi @coldic3, I don't understand how route alias will help, we need different path for the same controller, not different route name, like so

```patch
sylius_admin_order_resend_confirmation_email:
    path: /{id}/resend-confirmation-email
    methods: [GET]
    defaults:
        _controller: Sylius\Bundle\AdminBundle\Action\ResendOrderConfirmationEmailAction
+++sylius_admin_order_resend_confirmation_email_v2:
+++    path: /orders/{id}/resend-confirmation-email
+++    methods: [GET]
+++    defaults:
+++        _controller: Sylius\Bundle\AdminBundle\Action\ResendOrderConfirmationEmailAction
```

If I'm reading [docs](https://symfony.com/doc/current/routing.html#deprecating-route-aliases) right, then symfony doesn't have a mechanic to mark two routes as deprecation of one another.

While routes are covered with sylius BC promise, I think best solution is just to introduce one. :) I can't imagine anybody even using those routes aside from built-in buttons.

jakubtobiasz:
Hi!
Thanks for contributing. Due to lack of activity and being unable to push to this PR I've created a replacement PR #14801. Thanks for this base :).

PR #13241, opened 2021-10-22
BUGFIX: Sylius#11975 Added check for existing admin user id before…

### LINKED ISSUE #11975: When I create an administrator user
**Sylius version affected**: 1.8.2 and probably lower

**Description**  
When I create a new administrator user with an image, but required field not set, I have a TypeError.
If I don't add an image, form validation is correct.

```
Argument 1 passed to Symfony\Component\Security\Csrf\CsrfToken::__construct() must be of the type string, null given, called in /Users/mhuran/Sites/sylius/vendor/symfony/security-csrf/CsrfTokenManager.php on line 80
```

**Steps to reproduce**  
- Create a new administrator
- Add an image
- Keep other required fields empty
- Save

alex-boehm:
… rendering Avatar delete button

| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #11975
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/AdminUser/_form.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/AdminUser/_form.html.twig
@@ -26,7 +26,7 @@
                 <label>{{ 'sylius.ui.avatar'|trans }}</label>
                 {{ form_row(form.avatar, {'label': false}) }}
             </div>
-            {% if admin_user.avatar is not null %}
+            {% if admin_user.avatar is not null and app.request.attributes.get('id') > 0 %}
                 <button formaction="{{ path('sylius_admin_admin_user_remove_avatar', {'id': app.request.attributes.get('id'), '_csrf_token': csrf_token(app.request.attributes.get('id'))}) }}" type="submit" class="ui icon red labeled button">
                     <i class="icon trash"></i> {{ 'sylius.ui.delete'|trans }}
                 </button>


vvasiloi:
Hey @alex-boehm, thank you for your contribution!
Can you explain the reason behind this change? Why would it be necessary?

alex-boehm:
Hi @vvasiloi,
the corresponding bug is https://github.com/Sylius/Sylius/issues/11975.

How to reproduce the issue:
1. Go to Administrator Section in Backend
2. Start to create a new Administrator
3. !! Only add an image to the administrator, leave all other (required) fields empty
4. Save

Result:
CSRF Token error.

Reason:
The code for rendering the Delete-Button for the avatar image requires the record-Id to build the link ('id': app.request.attributes.get('id')) .
In this case we don't have an existing record because we left all other fields empty, BUT we have an avatar image, so the check " admin_user.avatar is not null" is true here and building the link for the delete button fails.

src/Sylius/Bundle/AdminBundle/Resources/views/AdminUser/_form.html.twig
```diff
@@ -26,7 +26,7 @@
                 <label>{{ 'sylius.ui.avatar'|trans }}</label>
                 {{ form_row(form.avatar, {'label': false}) }}
             </div>
-            {% if admin_user.avatar is not null %}
+            {% if admin_user.avatar is not null and app.request.attributes.get('id') > 0 %}
                 <button formaction="{{ path('sylius_admin_admin_user_remove_avatar', {'id': app.request.attributes.get('id'), '_csrf_token': csrf_token(app.request.attributes.get('id'))}) }}" type="submit" class="ui icon red labeled button">
```
vvasiloi, line 30:
```suggestion
            {% if admin_user.id is not null and admin_user.avatar is not null %}
                <button formaction="{{ path('sylius_admin_admin_user_remove_avatar', {'id': admin_user.id, '_csrf_token': csrf_token(admin_user.id)}) }}" type="submit" class="ui icon red labeled button">
```

vvasiloi (changes requested):
Thank your! I'm sorry, I didn't notice it references an issue.
I believe this is supposed to work only on the edit page.
Considering that, I suggested a small improvement.

vvasiloi:
Also, please rebase you branch an change the PR base branch to 1.10.

jakubtobiasz:
Closing as this PR has been replaced by #15148. Of course, @alex-boehm you're added as a co-author of the new commit :).

PR #13253, opened 2021-10-27
Run CI on Symfony 5.2 and 5.3

4c0n:
| Q               | A
| --------------- | -----
| Branch?         | 1.10 master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Run CI on both Symfony 5.2 and 5.3.

### DIFF
--- a/.github/workflows/application.yml
+++ b/.github/workflows/application.yml
@@ -122,19 +122,7 @@ jobs:
                 php: ["7.4", "8.0"]
                 symfony: ["^4.4", "5.2.*", "5.3.*"]
                 node: ["14.x"]
-                mysql: ["8.0"]
-
-                include:
-                    -
-                        php: "8.0"
-                        symfony: "5.2.*"
-                        node: "14.x"
-                        mysql: "5.7"
-                    -
-                        php: "8.0"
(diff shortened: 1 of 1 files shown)


4c0n:
@lchrusciel  Yes the conflict is gone, so the road to Symfony 5.3 and php 8 is open again.

1.9 doesn't support php 8, right? And it still has the Symfony ^5.2 constraint. So I don't think we can/should target the 1.9 branch. It'll test against 5.3 already and won't use php 8.

Would you prefer to test 1.10 and up against 5.2 and 5.3 or is just 5.3 enough (and of course keep 4.4)?

I'm fine with bumping the Node version.

lchrusciel:
Oh, you are correct, 1.9 is already tested with 5.3. So let's keep it with branch 1.10. And if you can, please change node version.

I would stay with testing v4.4, v5.2, and v5.3

4c0n:
@lchrusciel This should be good to go now, please review it :smiley:

lchrusciel (approved):

lchrusciel:
And merged! Thanks Youri!

PR #13270, opened 2021-11-03
Fixed logic at NthOrderRuleChecker

### LINKED ISSUE #13269: NthOrderRuleChecker has wrong logic
**Sylius version affected**: 1.10

**Description** 
At these lines we expect that the Customer is always attached to the Order, otherwise, it always returns false:
https://github.com/Sylius/Sylius/blame/dc9a42a66e04d55d555dcdac3698ded1ca112f61/src/Sylius/Component/Core/Promotion/Checker/Rule/NthOrderRuleChecker.php#L47-L52

But correct logic is:
```php
...
        $customer = $subject->getCustomer();

        // Eligible if it is first order of guest and the promotion is on first order
        if (null === $customer) {
            return 1 === $configuration['nth'];
        }
...
```

**Steps to reproduce**  

- Load fixture like this:
```yaml
sylius_fixtures:
    suites:
        default:
            fixtures:
                promotions:
                    name: 'promotion'
                    options:
                        custom:
                            new_10:
                                code: "new_10"
                                name: "New 10%"
                                description: "10% OFF on first order"
                                coupon_based: true
                                exclusive: true
                                rules:
                                    -
                                        type: "nth_order"
                                        configuration:
                                            nth: 1
                                actions:
                                    - ...
                                coupons:
                                    -
                                        code: 'NEW10'
                                channels:
                                    - 'FASHION_WEB'
```
- Submit `NEW10` promotion at cart. You will see `Coupon code is invalid` error message.

**Possible Solution**  
Described already.

igormukhingmailcom:
So, now `nth_order` rule is eligible not only for logged-in users.

| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #13269
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Core/Promotion/Checker/Rule/NthOrderRuleChecker.php
+++ b/src/Sylius/Component/Core/Promotion/Checker/Rule/NthOrderRuleChecker.php
@@ -45,12 +45,9 @@ final class NthOrderRuleChecker implements RuleCheckerInterface
         }
 
         $customer = $subject->getCustomer();
-        if (null === $customer) {
-            return false;
-        }
 
-        //eligible if it is first order of guest and the promotion is on first order
-        if (null === $customer->getId()) {
+        // Eligible if it is first order of guest and the promotion is on first order
+        if (null === $customer) {
             return 1 === $configuration['nth'];
         }
 


Roshyo (approved):

Roshyo:
Shouldn't we add some Behat scenario on that ? To make sure the same email does not enjoy twice the promotion.

lchrusciel:
Hey Igor, please add behat scenario to cover this case

jakubtobiasz:
Closing with the same reason as described in #13269. Thanks for your work :)!

PR #13281, opened 2021-11-08
fix: the parameter  period error was wrong

Snowbaha:
| Q               | A
| --------------- | -----
| Branch?         | 1.9 and more
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/OrderBundle/Command/RemoveExpiredCartsCommand.php
+++ b/src/Sylius/Bundle/OrderBundle/Command/RemoveExpiredCartsCommand.php
@@ -27,7 +27,7 @@ class RemoveExpiredCartsCommand extends ContainerAwareCommand
     protected function configure(): void
     {
         $this
-            ->setDescription('Removes carts that have been idle for a period set in `sylius_order.expiration.cart` configuration key.')
+            ->setDescription('Removes carts that have been idle for a period set in `sylius_order.cart_expiration_period` configuration key.')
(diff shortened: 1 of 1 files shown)


vvasiloi:
Hey @Snowbaha, thank you for the contribution.
Unfortunately, I cannot accept it and here's why.
This change is probably based on a confusion between configuration values and container parameters.
The command description is correct, because it refers to the configuration key path, for example one that you put in `config/packages/_sylius.yaml`:
```yaml
sylius_order:
    expiration:
        cart: '2 days'
```
This configuration's key path is `sylius_order.expiration.cart` and it's defined [here](https://github.com/Sylius/Sylius/blob/e87f9cfbe813bddfc369c7cd868aa6f113fd29f5/src/Sylius/Bundle/OrderBundle/DependencyInjection/Configuration.php#L152-L165) and when it's processed [here](https://github.com/Sylius/Sylius/blob/e87f9cfbe813bddfc369c7cd868aa6f113fd29f5/src/Sylius/Bundle/OrderBundle/DependencyInjection/SyliusOrderExtension.php#L38-L39), a corresponding container parameter named `sylius_order.cart_expiration_period` is defined .

Each processed value can be inspected with the debug commands:
```
php bin/console debug:config sylius_order expiration.cart
php bin/console debug:container --parameter=sylius_order.cart_expiration_period
```

PR #13288, opened 2021-11-11
[1.10] Maintain | Relax conflict rule with "liip/imagine-bundle"

stloyd:
| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | no?
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | related to #13261 
| License         | MIT

The conflict rule was to wide, as it also prevented to install fixed version (`2.7.1`).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -174,7 +174,7 @@
         "symfony/var-exporter": "^6.0",
         "symfony/property-info": "4.4.22 || 5.2.7",
         "symfony/serializer": "4.4.19 || 5.2.2",
-        "liip/imagine-bundle": "^2.7"
+        "liip/imagine-bundle": "2.7.0"
     },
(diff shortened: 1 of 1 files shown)


mbabker:
As a quick fix, add the deprecation to the Psalm configuration.  The deprecation is a result of a feature in the bundle's 2.7 release that allows configuring to use a version of the Twig extension with a runtime service, and they deprecated the version of the extension without the runtime service integration.

Long term, a new decorator is going to have to be created around the new lazy extension/runtime classes in the bundle to apply the same SVG fix.

stloyd:
Fixed in https://github.com/Sylius/Sylius/pull/13866

Issue #12832, opened 2021-07-22
ApiBundle requires API Platform to have Swagger enabled


mbabker:
**Sylius version affected**: 1.10.0

**Description**  
If you disable API Platform's Swagger integration (or, as in my case, the documentation as a whole; we just aren't using the API and there's no reason to expose the docs for it in any way), it does not load a handful of services.  The ApiBundle unconditionally registers a number of services that depend on ApiPlatformBundle parameters or decorate services that do not exist when Swagger is disabled, leading to an error along the lines of these:

```sh
The service "sylius.api.swagger_admin_authentication_documentation_normalizer" has a dependency on a non-existent service "api_platform.swagger.normalizer.documentation".
```

```sh
You have requested a non-existent parameter "api_platform.enable_swagger_ui". Did you mean one of these: "api_platform.enable_entrypoint", "api_platform.enable_docs"?  
 ```

**Steps to reproduce**  
- Install Sylius 1.10 with ApiBundle enabled
- Add the following to `config/packages/api_platform.yaml`

```yaml
api_platform:
    enable_swagger: false
    enable_swagger_ui: false
    enable_re_doc: false
    enable_entrypoint: false
    enable_docs: false
```

- Try to build the container

**Possible Solution**  
Can't say I have one offhand, there isn't anything from the ApiPlatformBundle which exposes the `enable_swagger` key in a way that other container extensions could use without re-parsing the configuration, and that is the config node that will cause the most problems here (there is the `api_platform.enable_swagger_ui` and `api_platform.enable_re_doc` parameters, but those are only set when `enable_swagger: true`).

For now, my workaround is a compiler pass removing these services defined in the ApiBundle:

- api_platform.swagger.action.ui (overrides the service defined by ApiPlatformBundle)
- sylius.api.swagger_admin_authentication_documentation_normalizer
- sylius.api.swagger_shop_authentication_documentation_normalizer
- sylius.api.swagger_product_documentation_normalizer
- sylius.api.swagger_product_variant_documentation_normalizer

devfive:
@mbabker thanks, this worked for us!

coldic3:
Hi @mbabker, thanks for rising this issue – it should be fixed now :dancer:

PR #13320, opened 2021-11-22
Remove swagger decorators when docs are disabled

paullla:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | https://github.com/Sylius/Sylius/issues/12832
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/DependencyInjection/SyliusApiExtension.php
+++ b/src/Sylius/Bundle/ApiBundle/DependencyInjection/SyliusApiExtension.php
@@ -15,6 +15,7 @@ namespace Sylius\Bundle\ApiBundle\DependencyInjection;
 
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 
@@ -29,5 +30,11 @@ final class SyliusApiExtension extends Extension
         $container->setParameter('sylius_api.enabled', $config['enabled']);
 
         $loader->load('services.xml');
+
+        if (!$container->hasParameter('enable_swagger') || !$container->getParameter('enable_swagger')) {
+            return;
+        }
+
+        $loader->load('integrations/swagger.xml');
     }
 }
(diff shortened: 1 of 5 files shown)


src/Sylius/Bundle/ApiBundle/DependencyInjection/SyliusApiExtension.php
```diff
@@ -30,5 +31,15 @@ public function load(array $configs, ContainerBuilder $container): void
...
 
         $loader->load('services.xml');
+
+        try {
+            if (!$container->getParameter('enable_swagger')) {
+                return;
+            }
+        } catch (ParameterNotFoundException $exception) {
+            return;
+        }
```
lchrusciel, line 41:
```suggestion
        if (!$container->hasParameter('enable_swagger') || !$container->getParameter('enable_swagger')) {
            return;
        }
```

src/Sylius/Bundle/ApiBundle/DependencyInjection/SyliusApiExtension.php
```diff
@@ -30,5 +31,11 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->setParameter('sylius_api.product_image_prefix', $config['product_image_prefix']);
 
         $loader->load('services.xml');
+
+        if (!$container->hasParameter('enable_swagger') || !$container->getParameter('enable_swagger')) {
+            return;
+        }
+
+        $loader->load('swagger/swagger.xml');
```
lchrusciel, line 39: Almost all of our integrations are placed in `services/integrations/*.xml` files. Yet, placing it there forbids us from usage of import wildcards, so my proposal is to adjust path to:
```suggestion
        $loader->load('integartions/swagger.xml');
```

paullla: @lchrusciel this is a special case because we shouldn't load it if swagger is turned off. So if we use wildcard here it could be loaded when it's off too and make the same bug.

lchrusciel:
Thanks, Paula! :1st_place_medal:

src/Sylius/Bundle/ApiBundle/DependencyInjection/SyliusApiExtension.php
```diff
@@ -29,5 +30,11 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->setParameter('sylius_api.enabled', $config['enabled']);
 
         $loader->load('services.xml');
+
+        if (!$container->hasParameter('enable_swagger') || !$container->getParameter('enable_swagger')) {
+            return;
+        }
+
```
pptasinski-mytheresa, line 37: It should be api_platform.enable_swagger or it should not be in api_platform namespace in config.yaml modified in this PR.

PR #13354, opened 2021-12-02
Refactor order serialization groups

paullla:
…for both normalization and denormalization groups

| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | 
| License         | MIT


src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Order.xml
```diff
@@ -38,6 +41,12 @@
                 <attribute name="denormalization_context">
                     <attribute name="groups">shop:order:create</attribute>
                 </attribute>
+                <attribute name="normalization_context">
+                    <attribute name="groups">
+                        <attribute>shop:order:item</attribute>
+                        <attribute>shop:order:list</attribute>
```
lchrusciel, line 47: I was thinking about something like this:
```suggestion
                        <attribute>shop:order:read:item</attribute>
                        <attribute>shop:order:read:list</attribute>
```
what do you think?

AdamKasp:
should we mention it in `UPGRADE_API-1.11.md`?

src/Sylius/Bundle/ApiBundle/Controller/DeleteOrderItemAction.php
```diff
@@ -37,6 +37,6 @@ public function __invoke(Request $request): Response
 
         $this->commandBus->dispatch($command);
 
-        return new JsonResponse();
+        return new JsonResponse(null, Response::HTTP_NO_CONTENT);
```
paullla, line 40: @lchrusciel Api platform returns this kind of response for DELETE endpoints instead of returning 200 with {} in content. I noticed this while adding contract test. Let me know if you are ok with it

UPGRADE-API-1.11.md
```diff
@@ -314,3 +314,11 @@
 24. `Sylius\Bundle\ApiBundle\Command\Cart\ApplyCouponToCart` and `Sylius\Bundle\ApiBundle\Command\Checkout\AddressOrder` commands have been replaced with `Sylius\Bundle\ApiBundle\Command\Checkout\UpdateCart`.
 
 25. `Sylius\Bundle\ApiBundle\CommandHandler\Cart\ApplyCouponToCartHandler` and `Sylius\Bundle\ApiBundle\CommandHandler\Checkout\AddressOrderHandler` command handlers have been replaced with `Sylius\Bundle\ApiBundle\CommandHandler\Checkout\UpdateCartHandler`.
+
+26. Following changes were made for normalization groups in `Sylius\Component\Core\Model\Order`:
+    * Removed `shop:cart:read` and `shop:cart:read` groups
+    * New groups added for item endpoints are `shop:order:read:item` and `shop:order:read:item`
```
lchrusciel, line 320:
```suggestion
    * New groups added for item endpoints are `admin:order:read:item` and `shop:order:read:item`
```

UPGRADE-API-1.11.md
```diff
@@ -314,3 +314,11 @@
...
+    * New groups added for collection endpoints are `shop:order:read:list` and `shop:order:read:list`
```
lchrusciel, line 321:
```suggestion
    * New groups added for collection endpoints are `admin:order:read:list` and `shop:order:read:list`
```

UPGRADE-API-1.11.md
```diff
@@ -314,3 +314,11 @@
...
+    * Removed following fields from `Order` collection response:
+        * `channel`, `customer`, `payments`, `shipments`, `items` and  `billingAddress` fully removed
+        * from `shippingAddress` field following fields removed: `city`, `street`, `postCode`, `provinceCode`, `provinceName`
```
lchrusciel, line 324: billing address is the main required address, and this one should be used to provide details about person who bought items
```suggestion
        * `channel`, `customer`, `payments`, `shipments`, `items` and  `shippingAddress` fully removed
        * from `billingAddress` field following fields removed: `city`, `street`, `postCode`, `provinceCode`, `provinceName`
```

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Order.xml
```diff
@@ -57,13 +66,22 @@
...
             </itemOperation>
 
             <itemOperation name="shop_get">
                 <attribute name="method">GET</attribute>
                 <attribute name="path">/shop/orders/{tokenValue}</attribute>
                 <attribute name="normalization_context">
-                    <attribute name="groups">shop:cart:read</attribute>
+                    <attribute name="groups">
+                        <attribute>shop:order:read:item</attribute>
+                        <attribute>shop:order:list</attribute>
```
lchrusciel, line 83:
```suggestion
                        <attribute>shop:order:read:list</attribute>
```

same for other occurrence

src/Sylius/Bundle/ApiBundle/Resources/config/serialization/Commands/Cart/AddItemToCart.xml
```diff
@@ -17,10 +17,10 @@
 >
     <class name="Sylius\Bundle\ApiBundle\Command\Cart\AddItemToCart">
         <attribute name="productVariantCode" serialized-name="productVariant">
-            <group>shop:cart:add_item</group>
+            <group>shop:order:add_item</group>
```
lchrusciel, line 20: in the next iteration it should probably be changed to 
```suggestion
            <group>shop:order_item:create</group>
```

src/Sylius/Bundle/ApiBundle/Resources/config/serialization/Commands/Cart/ChangeItemQuantityInCart.xml
```diff
@@ -17,7 +17,7 @@
 >
     <class name="Sylius\Bundle\ApiBundle\Command\Cart\ChangeItemQuantityInCart">
         <attribute name="quantity">
-            <group>shop:cart:change_quantity</group>
+            <group>shop:order:change_quantity</group>
```
lchrusciel, line 20: in the next iteration it should probably be changed to
```suggestion
            <group>shop:order_item:update</group>
```

src/Sylius/Bundle/ApiBundle/Resources/config/serialization/Commands/Cart/RemoveItemFromCart.xml
```diff
@@ -17,7 +17,7 @@
 >
     <class name="Sylius\Bundle\ApiBundle\Command\Cart\RemoveItemFromCart">
         <attribute name="orderItemId">
-            <group>shop:cart:remove_item</group>
+            <group>shop:order:remove_item</group>
         </attribute>
     </class>
```
lchrusciel, line 22: I'm wondering if this file is even needed. Delete endpoint does not have any body
image

src/Sylius/Bundle/ApiBundle/Resources/config/serialization/Commands/Checkout/ChoosePaymentMethod.xml
```diff
@@ -17,7 +17,7 @@
 >
     <class name="Sylius\Bundle\ApiBundle\Command\Checkout\ChoosePaymentMethod">
         <attribute name="paymentMethodCode" serialized-name="paymentMethod">
-            <group>shop:cart:select_payment_method</group>
+            <group>shop:order:select_payment_method</group>
```
lchrusciel, line 20: following the example of order item, it seems to me, that 
```suggestion
            <group>shop:payment:update</group>
```
would be a better group for this endpoint

src/Sylius/Bundle/ApiBundle/Resources/config/serialization/Commands/Checkout/ChooseShippingMethod.xml
```diff
@@ -17,7 +17,7 @@
 >
     <class name="Sylius\Bundle\ApiBundle\Command\Checkout\ChooseShippingMethod">
         <attribute name="shippingMethodCode" serialized-name="shippingMethod">
-            <group>shop:cart:select_shipping_method</group>
+            <group>shop:order:select_shipping_method</group>
```
lchrusciel, line 20: following the example of order item, it seems to me, that
```suggestion
            <group>shop:shipment:update</group>
```

 would be a better group for this endpoint

lchrusciel (dismissed):

tests/Api/Responses/Expected/shop/get_orders_response.json
```diff
@@ -0,0 +1,31 @@
+{
+  "@context": "\/api\/v2\/contexts\/Order",
+  "@id": "\/api\/v2\/shop\/orders",
+  "@type": "hydra:Collection",
+  "hydra:member": [
+    {
+      "@id": "\/api\/v2\/shop\/orders\/nAWw2jewpA",
+      "@type": "Order",
+      "shippingAddress": {
```
lchrusciel, line 9:
```suggestion
      "billingAddress": {
```

stloyd:
@lchrusciel @paullla This is huge BC break even if API is still marked as experimental, yet there is no task, nor description giving people notes why all those changes are required/applied.

lchrusciel:
Hey Joseph,

you are correct, this PR should have bigger explanation. We are aiming to stabelize new API and remove deprecation flags. Therefore, this is the last moment, we can introduce BC-Breaks and they won't be possible(or at least shouldn't happen) afterwards. 

This PR aims to solve a few issues:
* We've introduced a new serialization group `no_serialization`, to not leak any data by accident. If anyone has a better idea, how to configure symfony serializer to have exclusion by default policy, I'm open for that. Nontheless, it aims to protect end users from not expose some data by accident. Previously, default order serialization was `admin:order:read`. If anyone would add new endpoint in their app, they could leak some data, even tho the endpoint was exposed through `shop` prefix. Even now, in our repo this serialization group was added to `shop_post` and previously on order update. In the long term, this may lead to unexpected behaviours and issues. This was not detected during review nor through contract tests because we didn't have them. 
* We've split item and collection serialization for the better DX in the long term. It is normal, that people may want to have more date on item endpoints. In addition, we've already introduce practice that list collection will be always taken into account on item endpoint
* Following endpoints: 
```
    * Renamed `shop:order:add_item` to `shop:order_item:create`
    * Renamed `shop:order:select_payment_method` to `shop:payment:update`
    * Renamed `shop:order:select_shipping_method` to `shop:shipment:update`
    * Renamed `shop:cart:change_quantity` to `shop:order_item:update`
```
has been adjusted to follow CRUD perspective over the paths, that are using these groups. This also aims to improve DX and make serialization groups predictable and easier to describe. 
* Probably the change of `cart` to `order` is controversial, but I found it more coherent. I'm open for discussion if this should be reverted. 

These changes should be propagated to all API resources. Perhaps, I will create aggregation issue for that. 

To reduce the pain of this change, perhaps some migration script should be added.

vvasiloi:
> If anyone has a better idea, how to configure symfony serializer to have exclusion by default policy, I'm open for that.
@lchrusciel You means something like this: https://jmsyst.com/libs/serializer/master/cookbook/exclusion_strategies#general-exclusion-strategies?

stloyd:
Or if we speak about Symfony Serializer: https://symfony.com/doc/4.4/components/serializer.html#ignoring-attributes

lchrusciel:
@vvasiloi yes, something similar in Symfony Serializer would be awesome

@stloyd this is great lead, but I don't see, how to use it on all fields, on all resources by default.

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Order.xml
```diff
@@ -38,6 +41,12 @@
                 <attribute name="denormalization_context">
                     <attribute name="groups">shop:order:create</attribute>
                 </attribute>
+                <attribute name="normalization_context">
+                    <attribute name="groups">
+                        <attribute>shop:order:read:item</attribute>
+                        <attribute>shop:order:read:list</attribute>
```
diimpp, line 47: `show`, `index` is more inline with sylius terminology.

```suggestion
                        <attribute>shop:order:show</attribute>
                        <attribute>shop:order:index</attribute>
```

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Adjustment.xml
```diff
@@ -39,12 +39,12 @@
         <subresourceOperations>
             <subresourceOperation name="api_orders_adjustments_get_subresource">
                 <attribute name="normalization_context">
-                    <attribute name="groups">shop:cart:read</attribute>
+                    <attribute name="groups">shop:order:read:item</attribute>
```
diimpp, line 42: `cart` and `order` separation are good one and most likely should be kept as is.

NoResponseMate:
Hi @paullla 👋 

Thanks for the PR and the idea by itself.
We've incorporated the split in 1.13 via #15770 with a [BC layer](https://github.com/Sylius/Sylius/pull/15780).
For now, the old groups will still be in place and will be completely removed in 2.0.

Cheers 🍻

PR #13361, opened 2021-12-06
[ADMIN] Fix products routes id requirements

### LINKED ISSUE #13359: [Admin] Products routes id requirements
**Sylius version affected**: 1.9

**Description**  
In products admin routes, the id parameter has no requirements.

**Steps to reproduce**  
```
sylius_admin_product_update:
    path: /products/{id}/edit
    methods: [GET, PUT, PATCH]
    defaults:
        _controller: sylius.controller.product:updateAction
        _sylius:
            section: admin
            permission: true
            redirect: referer
            template: "@SyliusAdmin/Crud/update.html.twig"
            vars:
                subheader: sylius.ui.manage_your_product_catalog
                icon: cube
                templates:
                    form: "@SyliusAdmin/Product/_form.html.twig"
                    toolbar: "@SyliusAdmin/Product/Update/_toolbar.html.twig"

sylius_admin_product_show:
    path: /products/{id}
    methods: [GET]
    defaults:
        _controller: sylius.controller.product:showAction
        _sylius:
            section: admin
            permission: true
            template: "@SyliusAdmin/Product/show.html.twig"
```

**Possible Solution**  
```
sylius_admin_product_update:
    path: /products/{id}/edit
    methods: [GET, PUT, PATCH]
    requirements:
        id: '\d+'
    defaults:
        _controller: sylius.controller.product:updateAction
        _sylius:
            section: admin
            permission: true
            redirect: referer
            template: "@SyliusAdmin/Crud/update.html.twig"
            vars:
                subheader: sylius.ui.manage_your_product_catalog
                icon: cube
                templates:
                    form: "@SyliusAdmin/Product/_form.html.twig"
                    toolbar: "@SyliusAdmin/Product/Update/_toolbar.html.twig"

sylius_admin_product_show:
    path: /products/{id}
    methods: [GET]
    requirements:
        id: '\d+'
    defaults:
        _controller: sylius.controller.product:showAction
        _sylius:
            section: admin
            permission: true
            template: "@SyliusAdmin/Product/show.html.twig"
```

bigboss86:
| Q               | A
| --------------- | -----
| Branch?         | 1.10 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | fixes #13359
| License         | MIT


bigboss86:
@lchrusciel

vvasiloi:
Hey @bigboss86, thank you for your contribution.
Although Sylius uses integer IDs by default, this choice is not enforced on users and I know some choose to use UUIDs instead.

bigboss86:
Hi @vvasiloi. Thanks for your reply. Because Sylius default is an integer ID, wouldn' It be better to have it by default? And for the users who want to choose UUIDs, they should overwrite it.

vvasiloi:
As @Zales0123 noted in a Slack conversation, this would be a BC break, because it's a narrowing requirement.
What are the benefits of this requirement in you opinion?

vvasiloi:
Can you elaborate on why is it the correct improvement?
Does it fix a bug?
Or maybe adds a new feature?
Don't get me wrong, I'm not necessarily against it
I fact, I have an idea on how to make it work for integers, UUIDs and whatnot.
I just want to know what's the purpose and if it's worth the effort.

bigboss86:
We had an issue in a project when we wanted to create a new route for the admin. For example: /admin/products/import. When we enter to that route we see an error because Symfony matches the product show url. 
One solution can be put the new route at the top so it will match first. But we're thinking that if the original model only accepts integers for the product id, the route should be in the same line and accept only integers. 
In that way, if one developer wants to use UUID instead of integer in the id, they should:
1- Modify the model mapping.
2- Override the product routes.

What is your opinion?

vvasiloi:
I though so... I had a similar issues.
I potential solution is to parameterize the requirement, either though configuration or a directly though parameter.

PR #13358, opened 2021-12-06
[Maintenance] Update dependencies Symfony/password-hasher

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| License         | MIT

PR #13363, opened 2021-12-06
symfony/password-hasher compatibility issue

### LINKED ISSUE #13362: Incompatible symfony/password-hasher version
**Sylius version affected**: master

**Description**  
When you install master, update all depedencies, follow the installation instructions and try to login to the admin panel you get the error `No password hasher has been configured for account "[username]".`

**Possible Solution**  
Implement `PasswordAuthenticatedUserInterface` on `src/Sylius/Component/User/Model/User.php`

TheDevilOnLine:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes (Symfony 4.4 support breaks)
| Deprecations?   | no
| Related tickets | fixes #13362 
| License         | MIT

The latest version of symfony/password-hasher requires the User object to implement `PasswordAuthenticatedUserInterface`, or `PasswordHasherAwareInterface`, otherwise it'll convert the User object into a string (and throw the error `No password hasher has been configured for account "[username]".` when logging in to the admin panel).

`PasswordHasherAwareInterface` requires the hasher to be configured on the User class, which requires a lot of changes to be able to configure the hasher from the configuration files.

`PasswordAuthenticatedUserInterface` only requires a 'getPassword()' function to be present (which is already there) and uses the configuration from `security.yaml` to decide on the hasher. This therefore seems to be the prefered approach, which is therefore implemented in this PR.

### DIFF
--- a/src/Sylius/Component/User/Model/User.php
+++ b/src/Sylius/Component/User/Model/User.php
@@ -17,8 +17,9 @@ use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Sylius\Component\Resource\Model\TimestampableTrait;
 use Sylius\Component\Resource\Model\ToggleableTrait;
+use Symfony\Component\Security\Core\User\PasswordAuthenticatedUserInterface;
 
-class User implements UserInterface
+class User implements UserInterface, PasswordAuthenticatedUserInterface
 {
     use TimestampableTrait, ToggleableTrait;
 


TheDevilOnLine:
I noticed from the tests that it DOES break Symfony 4.4 compatiblity.

mbabker:
What version of the password hasher component is being installed?  https://github.com/Sylius/Sylius/pull/13358 most likely mitigates this and #13362 if 6.0 is being installed through the network of transient dependencies (since Sylius itself isn't aware of the new component yet and can't hard require it without dropping Symfony 4.4 support).

TheDevilOnLine:
According to composer.lock symfony/password-hasher 6.0.0 was installed, using Symfony 5.4.
The fix introduced in #13358 has a small bug, but should fix it as well and might indeed be the better solution

lchrusciel:
Still, thanks a lot for your contribution :)

PR #13368, opened 2021-12-08
[UI] change time format to 24H, autocomplete times on data change

SirDomin:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Time display format has been changed to Y-m-d H:m from Y-m-d h:m:s, where s was 00 by default and h was 12h format, form takes seconds to account

Before:
image

After:
image

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/views/CatalogPromotion/_form.html.twig
+++ b/src/Sylius/Bundle/AdminBundle/Resources/views/CatalogPromotion/_form.html.twig
@@ -29,4 +29,20 @@
     <div class="column">
         {{ translationForm(form.translations) }}
     </div>
+    <script>
+        document.querySelector('#sylius_catalog_promotion_startDate_date').addEventListener('change', event => {
+            setValueForTimeInput(event, '00:00');
+        })
+        document.querySelector('#sylius_catalog_promotion_endDate_date').addEventListener('change', event => {
+            setValueForTimeInput(event, '23:59');
+        })
+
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/AdminBundle/Resources/views/CatalogPromotion/_form.html.twig
```diff
@@ -29,4 +29,20 @@
     <div class="column">
         {{ translationForm(form.translations) }}
     </div>
+    <script>
```
AdamKasp, line 32: do we need it? it should be configured in Symfony form configuration (probably on `pre_submit` event, if a value is empty form should set proper time)

SirDomin: this way the form is empty by default, when user changes date it automatically setups times, so user is aware of what's happening. Leaving it empty by default and then submiting it may lead to misunderstanding of user, where some people may think it should be 23:59 or 00:00, this way we eliminate problem.

SirDomin: alright, that was unnecessary, removed that and added seconds to input

lchrusciel (approved):

GSadee (approved):

PR #13372, opened 2021-12-09
Remove ORM 2.10 conflict

mbabker:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | N/A
| License         | MIT

The conflict against `doctrine/orm:^2.10` has been unnecessary since the `gedmo/doctrine-extensions` 3.2 release.  In fact, the actual conflict was between the extensions package and the ORM, Sylius was never directly conflicted against it.  Now that newer versions of the extensions package either have the appropriate conflict or fixed the conflict and now require newer ORM versions, Sylius no longer needs to have this conflict.

### DIFF
--- a/CONFLICTS.md
+++ b/CONFLICTS.md
@@ -48,14 +48,6 @@ twig-bridge, var-dumper, var-exporter 6.0, which is not compatible with the curr
 
    Probably introduced in: https://github.com/symfony/symfony/pull/40811
 
- - `doctrine/orm:2.10.0`:
-
-   This version causes a problem with the creation of nested taxons by throwing the exception:
-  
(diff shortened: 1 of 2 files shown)


lchrusciel (approved):
Perhaps we should bump minimal requirement for `"gedmo/doctrine-extensions": "^2.4.12 || ^3.0",` to `"gedmo/doctrine-extensions": "^2.4.12 || ^3.2",`

lchrusciel:
Thanks, Michael! :1st_place_medal:

mbabker:
> Perhaps we should bump minimal requirement

2.4 isn't supported at all anymore and explicitly blocks PHP 8, so if you're going to bump this requirement then you should drop 2.4 support in full.

lchrusciel:
makes sense, as Sylius is ^8.0 already. Can you open a PR? Will be part of next release.

PR #13375, opened 2021-12-09
Catalog promotion channel pricing relation

Rafikooo:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no 
| License         | MIT


src/Sylius/Bundle/CoreBundle/Processor/CatalogPromotionClearer.php
```diff
@@ -18,40 +18,31 @@
...
         $channelPricings = $this->channelPricingRepository->findWithDiscountedPrice();
-        foreach ($channelPricings as $channelPricing) {
-            $appliedPromotionsCodes = array_unique(array_merge(
-                $appliedPromotionsCodes,
-                array_keys($channelPricing->getAppliedPromotions())
-            ));
+        $catalogPromotions = [];
 
+        foreach ($channelPricings as $channelPricing) {
+            $catalogPromotions = array_merge($catalogPromotions, $channelPricing->getAppliedPromotions()->toArray());
```
Rafikooo, line 42: I left `->toArray` on purpose although foreach below can handle array collection objects.
On line 42 I get appliedPromotions and store them in a variable but in the second step, I clear promotions stored in channelPricing object. After that, it occurs that removing catalog promotions inside channelPricing object also affect local variable $catalogPromotions and clear it also. Alternatively I could use instead of `->toArray` clone method which can be more explicit in the future, that we dealing with a pointer

tests/Api/Responses/Expected/shop/catalog_promotion/get_catalog_promotion_response.json
```diff
@@ -0,0 +1,4 @@
+{
+  "id": @integer@,
```
Rafikooo, line 2: There are no `@context` and related API Platform fields. The reason is that I wasn't able to include them in the shop context. In the admin context, everything works well

Rafikooo (commented):
Added two notes about the code

AdamKasp:
I think you should squash part of your commits (30 commits IMO is too many)

src/Sylius/Behat/Context/Api/Shop/ProductVariantContext.php
```diff
@@ -17,25 +17,27 @@
...
 use Sylius\Behat\Client\ResponseCheckerInterface;
 use Sylius\Behat\Service\SharedStorageInterface;
-use Sylius\Component\Core\Formatter\StringInflector;
-use Sylius\Component\Core\Model\ProductInterface;
 use Sylius\Component\Core\Model\ProductVariantInterface;
 use Webmozart\Assert\Assert;
 
 final class ProductVariantContext implements Context
 {
     private ApiClientInterface $client;
```
AdamKasp, line 25: it just my opinion but when we have more than a single client in context we should name it more explicit (for example `variantClient`

src/Sylius/Behat/Context/Api/Shop/ProductVariantContext.php
```diff
@@ -105,10 +108,11 @@ public function iShouldSeeVariantIsDiscountedFromToWithPromotions(
 
         Assert::same($content['price'], $price);
         Assert::same($content['originalPrice'], $originalPrice);
-        foreach ($promotionsNames as $promotionName) {
-            $catalogPromotionCode = StringInflector::nameToCode($promotionName);
-            Assert::same($content['appliedPromotions'][$catalogPromotionCode]['translations']['en_US']['name'], $promotionName);
-            Assert::same($content['appliedPromotions'][$catalogPromotionCode]['translations']['en_US']['description'], $promotionName . ' description');
+        foreach ($content['appliedPromotions'] as $promotionIri) {
+            $catalogPromotion = $this->responseChecker->getResponseContent(
```
AdamKasp, line 112:
```suggestion
            $catalogPromotionContent = $this->responseChecker->getResponseContent(
```

src/Sylius/Bundle/ApiBundle/Serializer/ProductVariantNormalizer.php
```diff
@@ -71,9 +72,15 @@ public function normalize($object, $format = null, array $context = [])
             unset($data['originalPrice']);
         }
 
+        /** @var ArrayCollection $appliedPromotions */
         $appliedPromotions = $object->getAppliedPromotionsForChannel($channel);
-        if (!empty($appliedPromotions)) {
-            $data['appliedPromotions'] = $appliedPromotions;
+        if (!$appliedPromotions->isEmpty()) {
+            $data['appliedPromotions'] = array_map(fn (CatalogPromotionInterface $catalogPromotion) => sprintf(
+                '/api/v2/shop/catalog-promotions/%s',
```
AdamKasp, line 79: we talked about it, can we make this path more dynamic? if a user will change prefixes in his project this serialization will cause an error.
if we are not able to make it dynamic, IMO we should move it to the parameter and inject it as an argument to the service and add note to our docs.

src/Sylius/Bundle/ApiBundle/spec/Serializer/ProductVariantNormalizerSpec.php
```diff
@@ -32,9 +35,10 @@ function let(
         ProductVariantPricesCalculatorInterface $pricesCalculator,
         ChannelContextInterface $channelContext,
         AvailabilityCheckerInterface $availabilityChecker,
-        SectionProviderInterface $sectionProvider
+        SectionProviderInterface $sectionProvider,
+        IriConverterInterface $iriConverter
     ): void {
-        $this->beConstructedWith($pricesCalculator, $channelContext, $availabilityChecker, $sectionProvider);
+        $this->beConstructedWith($pricesCalculator, $channelContext, $availabilityChecker, $sectionProvider, $iriConverter);
```
AdamKasp, line 41: if I see correctly we are not injecting it in `src/Sylius/Bundle/ApiBundle/Serializer/ProductVariantNormalizer.php`

src/Sylius/Bundle/CoreBundle/spec/Applicator/CatalogPromotionApplicatorSpec.php
```diff
@@ -40,101 +39,52 @@ function it_implements_catalog_promotion_applicator_interface(): void
...
-        $priceCalculator->calculate($secondChannelPricing, $catalogPromotionAction)->willReturn(980);
-
         $secondChannelPricing->setPrice(980)->shouldBeCalled();
-        $secondChannelPricing->addAppliedPromotion(['winter_sale' => ['name' => 'Winter sale']])->shouldBeCalled();
-        $catalogPromotion->isExclusive()->willReturn(false);
-
-        $this->applyOnVariant($variant, $catalogPromotion);
-    }
-
-    function it_applies_discount_on_product_variant_only_if_exclusive_promotion_is_not_already_applied(
```
AdamKasp, line 91: Why did you delete this scenario?

src/Sylius/Bundle/ShopBundle/Resources/private/js/sylius-variants-prices.js
```diff
@@ -9,17 +9,15 @@
...
   $('#appliedPromotions').html('');
 
   if (appliedPromotions !== '[]') {
     $.each(appliedPromotions, (index, promotion) => {
-      if (promotion['translations'][locale].description !== null) {
-        appliedPromotionsElement += `<div class="ui blue label promotion_label" style="margin: 1rem 0;"><div class="row ui small sylius_catalog_promotion">${promotion['translations'][locale].name} - ${promotion['translations'][locale].description}</div></div>`;
-      } else {
-        appliedPromotionsElement += `<div class="ui blue label promotion_label" style="margin: 1rem 0;"><div class="row ui small sylius_catalog_promotion">${promotion['translations'][locale].name}</div></div>`;
-      }
+      let promotionInfo = promotion.name;
```
AdamKasp, line 18: here we should display, translatable field - `label` instead of name  
```suggestion
      let promotionInfo = promotion.label;
```

src/Sylius/Bundle/ShopBundle/Resources/views/Product/Box/_content.html.twig
```diff
@@ -20,7 +20,7 @@
             {% set originalPrice = money.calculateOriginalPrice(product|sylius_resolve_variant) %}
 
             {% for promotion in appliedPromotions %}
-            <a class="ui blue label" {{ sylius_test_html_attribute('product-promotion-label') }}>{{ promotion['translations'][sylius.localeCode]['name'] }}</a>
+            <a class="ui blue label" {{ sylius_test_html_attribute('product-promotion-label') }}>{{ promotion.name}}</a>
```
AdamKasp, line 23: same here

src/Sylius/Bundle/ShopBundle/Resources/views/Product/Show/_price.html.twig
```diff
@@ -14,9 +14,8 @@
 <div id="appliedPromotions" data-applied-promotions-locale="{{ sylius.localeCode }}">
     {% for appliedPromotion in channelPricing.appliedPromotions %}
     <div class="ui blue label promotion_label" style="margin: 1rem 0;">
-        {% set promotion = appliedPromotion['translations'][sylius.localeCode] %}
         <div class="row ui small sylius_catalog_promotion">
-            {{ promotion.name }}{% if promotion.description %} - {{ promotion.description }}{% endif %}
+            {{ appliedPromotion.name }}{% if appliedPromotion.description %} - {{ appliedPromotion.description }}{% endif %}
```
AdamKasp, line 18: and here :)

tests/Api/Responses/Expected/shop/catalog_promotion/get_catalog_promotion_response.json
AdamKasp: But you should have the hydra date in response, I think that you have some problem with serialization.

AdamKasp (changes requested):

src/Sylius/Bundle/ApiBundle/Serializer/ProductVariantNormalizer.php
AdamKasp: OR if we do not be able to use IRI here, maybe a better option will be to serialize here necessary data, instead of using IRI.
@lchrusciel @Zales0123 @GSadee, what do you think? which option is best, or maybe you have a better idea?

GSadee: Probably once again, but, what was the origin problem with not using IriConverter here? I've just pulled this branch and checked using IriConverter here and it seems to be working correctly (I've got IRI with shop prefix in shop context)

Rafikooo: It seems that is actually works well

tests/Api/Responses/Expected/shop/catalog_promotion/get_catalog_promotion_response.json
Rafikooo: Adding json+ld header resolved the issue

src/Sylius/Bundle/CoreBundle/Processor/CatalogPromotionClearer.php
Rafikooo: I could write a spec to assure right behavior

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/ChannelPricingRepository.php
```diff
@@ -20,8 +20,11 @@ class ChannelPricingRepository extends EntityRepository implements ChannelPricin
 {
     public function findWithDiscountedPrice(): array
     {
-        return $this->createQueryBuilder('o')
-            ->andWhere('o.appliedPromotions != \'[]\'')
+        return $this->createQueryBuilder('channelPricing')
```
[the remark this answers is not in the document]
Rafikooo: I wasn't aware of that, pardon

Rafikooo (commented):
Checked if I applied all review changes

AdamKasp (approved):

src/Sylius/Bundle/CoreBundle/Processor/CatalogPromotionClearer.php
Rafikooo: I resolved mapping-related stuff. I'm not sure if we are on the same page with my comment above, because the problem happens in the application state rather than DB.

A moment ago I discovered that I was wrong. Indeed, there is that pointer/reference case, but it is in `src/Sylius/Bundle/CoreBundle/Processor/CatalogPromotionClearer.php:42`. After getting `ArrayCollection`, storing it in `$appliedPromotions`  and calling`clearChannelPricing`, `foreach` is skipped. `->toArray` produces deep copy so `foreach` is executed correctly

lchrusciel (approved):

src/Sylius/Component/Core/Model/ChannelPricing.php
```diff
@@ -43,8 +46,16 @@ class ChannelPricing implements ChannelPricingInterface
      */
     protected $minimumPrice = 0;
 
-    /** @var ?array */
-    protected $appliedPromotions = [];
+    /**
+     * @var ArrayCollection
+     * @psalm-var ArrayCollection<array-key, CatalogPromotionInterface>
+     */
+    protected $appliedPromotions;
```
GSadee, line 53:
```suggestion
     * @psalm-var ArrayCollection<array-key, CatalogPromotionInterface>
     */
    protected Collection $appliedPromotions;
```

GSadee (approved):

GSadee:
Thank you, Rafał! :1st_place_medal:

PR #13390, opened 2021-12-14
[Products] move channel pricing to separate tab

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         |  master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| License         | MIT

now when we display more info about prices it should be in a 
Screenshot 2021-12-14 at 15 01 28
separate tab (similar to variant page)


Zales0123:
I believe it was already merged in the meantime 🖖 Thank you, Adam! 🎉

PR #13325, opened 2021-11-24
[CatalogPromotion][API][UI] Fixed discount action

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

PR #13399, opened 2021-12-15
[RFC][CatalogPromotion] Refactor actions and scopes to dynamically set parameters with types

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | no
| New feature?    | no?
| BC breaks?      | no?
| Deprecations?   | no
| Related tickets | fixes https://github.com/Sylius/Sylius/pull/13325#discussion_r756750238
| License         | MIT

To make adding new catalog promotion actions and scopes easier, I refactored them a bit to store the type of action/scopes in calculator/provider class (which is an obligatory service of each action/scopes) and dynamically set parameters with types (which is needed for example for validation).

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Calculator/PercentageDiscountPriceCalculator.php
+++ b/src/Sylius/Bundle/CoreBundle/Calculator/PercentageDiscountPriceCalculator.php
@@ -18,9 +18,16 @@ use Sylius\Component\Promotion\Model\CatalogPromotionActionInterface;
 
 final class PercentageDiscountPriceCalculator implements ActionBasedPriceCalculatorInterface
 {
+    public const TYPE = 'percentage_discount';
+
+    public function getType(): string
+    {
+        return self::TYPE;
+    }
+
     public function supports(CatalogPromotionActionInterface $action): bool
     {
-        return $action->getType() === CatalogPromotionActionInterface::TYPE_PERCENTAGE_DISCOUNT;
+        return $action->getType() === self::TYPE;
     }
 
     public function calculate(ChannelPricingInterface $channelPricing, CatalogPromotionActionInterface $action): int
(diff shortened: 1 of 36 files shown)


AdamKasp (approved):

src/Sylius/Bundle/CoreBundle/Calculator/PercentageDiscountPriceCalculator.php
```diff
@@ -18,9 +18,16 @@
 
 final class PercentageDiscountPriceCalculator implements ActionBasedPriceCalculatorInterface
 {
+    public const TYPE = 'percentage_discount';
+
+    public function getType(): string
```
lchrusciel, line 23: What do you think about making this field static? 
```suggestion
    public static function getType(): string
```

src/Sylius/Bundle/CoreBundle/Resources/config/services/form.xml
```diff
@@ -52,6 +52,8 @@
             <parameter>sylius</parameter>
         </parameter>
         <parameter key="sylius.model.shop_billing_data.class">Sylius\Component\Core\Model\ShopBillingData</parameter>
+        <parameter key="sylius.catalog_promotion.action.fixed_discount" type="constant">Sylius\Bundle\CoreBundle\Calculator\FixedDiscountPriceCalculator::TYPE</parameter>
```
lchrusciel, line 55: I'm wondering if we could also remove these parameters, but I don't have any clear idea how for now

lchrusciel (approved):

src/Sylius/Bundle/CoreBundle/Resources/config/services/form.xml
GSadee: To be honest, I would be for not using these parameters, because there are used only once, for tag in form definition, IMO there could be explicitly defined the proper type.

lchrusciel:
Closing in favour of #13403

PR #13431, opened 2021-12-31
FIX: Avoid passing empty needle to strpos()

rimas-kudelis:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

The province validator returns early if it finds preexisting validation errors on the address. However, when checking that, it uses `strpos()` and passes current `$propertyPath` as a needle. This might cause problems when an address is validated as a top-level entity, because `$propertyPath` is an empty string in such case.

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Validator/Constraints/ProvinceAddressConstraintValidator.php
+++ b/src/Sylius/Bundle/AddressingBundle/Validator/Constraints/ProvinceAddressConstraintValidator.php
@@ -47,7 +47,7 @@ class ProvinceAddressConstraintValidator extends ConstraintValidator
         $propertyPath = $this->context->getPropertyPath();
 
         foreach (iterator_to_array($this->context->getViolations()) as $violation) {
-            if (0 === strpos($violation->getPropertyPath(), $propertyPath)) {
+            if ('' === $propertyPath || 0 === strpos($violation->getPropertyPath(), $propertyPath)) {
                 return;
             }
         }
--- a/src/Sylius/Bundle/AddressingBundle/spec/Validator/Constraints/ProvinceAddressConstraintValidatorSpec.php
+++ b/src/Sylius/Bundle/AddressingBundle/spec/Validator/Constraints/ProvinceAddressConstraintValidatorSpec.php
@@ -19,6 +19,7 @@ use Sylius\Bundle\AddressingBundle\Validator\Constraints\ProvinceAddressConstrai
 use Sylius\Bundle\AddressingBundle\Validator\Constraints\ProvinceAddressConstraintValidator;
 use Sylius\Component\Addressing\Model\AddressInterface;
 use Sylius\Component\Addressing\Model\Country;
+use Sylius\Component\Addressing\Model\CountryInterface;
 use Sylius\Component\Addressing\Model\Province;
 use Sylius\Component\Resource\Repository\RepositoryInterface;
 use Symfony\Component\Validator\Constraint;
@@ -46,10 +47,19 @@ final class ProvinceAddressConstraintValidatorSpec extends ObjectBehavior
     }
 
     function it_does_not_add_violation_because_a_violation_exists(
+        RepositoryInterface $countryRepository,
         AddressInterface $address,
+        Country $country,
         ProvinceAddressConstraint $constraint,
         ExecutionContextInterface $context
     ): void {
+        $country->getCode()->willReturn('PL');
+        $address->getCountryCode()->willReturn('PL');
+        $countryRepository->findOneBy(['code' => 'PL'])->willReturn($country);
+
+        $country->hasProvinces()->willReturn(true);
+        $address->getProvinceCode()->willReturn(null);
+
         $this->initialize($context);
 
         $context->getPropertyPath()->willReturn('property_path');
@@ -63,10 +73,19 @@ final class ProvinceAddressConstraintValidatorSpec extends ObjectBehavior
     }
 
     function it_does_not_add_violation_because_a_violation_exists_when_address_is_the_root_object(
+        RepositoryInterface $countryRepository,
         AddressInterface $address,
+        Country $country,
         ProvinceAddressConstraint $constraint,
         ExecutionContextInterface $context
     ): void {
+        $country->getCode()->willReturn('PL');
+        $address->getCountryCode()->willReturn('PL');
+        $countryRepository->findOneBy(['code' => 'PL'])->willReturn($country);
+
+        $country->hasProvinces()->willReturn(true);
+        $address->getProvinceCode()->willReturn(null);
+
         $this->initialize($context);
 
         $context->getPropertyPath()->willReturn('');
--- a/src/Sylius/Bundle/AddressingBundle/spec/Validator/Constraints/ProvinceAddressConstraintValidatorSpec.php
+++ b/src/Sylius/Bundle/AddressingBundle/spec/Validator/Constraints/ProvinceAddressConstraintValidatorSpec.php
@@ -62,6 +62,23 @@ final class ProvinceAddressConstraintValidatorSpec extends ObjectBehavior
         $this->validate($address, $constraint);
     }
 
+    function it_does_not_add_violation_because_a_violation_exists_when_address_is_the_root_object(
+        AddressInterface $address,
+        ProvinceAddressConstraint $constraint,
+        ExecutionContextInterface $context
+    ): void {
+        $this->initialize($context);
+
+        $context->getPropertyPath()->willReturn('');
+        $context->getViolations()->willReturn(new \ArrayIterator([
+            $this->createViolation('property_path'),
+        ]));
+
+        $context->addViolation(Argument::any())->shouldNotBeCalled();
+
+        $this->validate($address, $constraint);
+    }
+
     function it_adds_violation_because_address_has_no_province(
         RepositoryInterface $countryRepository,
         AddressInterface $address,
--- a/src/Sylius/Bundle/AddressingBundle/spec/Validator/Constraints/ProvinceAddressConstraintValidatorSpec.php
+++ b/src/Sylius/Bundle/AddressingBundle/spec/Validator/Constraints/ProvinceAddressConstraintValidatorSpec.php
@@ -53,6 +53,8 @@ final class ProvinceAddressConstraintValidatorSpec extends ObjectBehavior
         ProvinceAddressConstraint $constraint,
         ExecutionContextInterface $context
     ): void {
+        // DO NOT REMOVE THIS CODE: it ensures that there is a violation that can be added
+        // if the logic that is being tested (NOT adding a validation) does not work.
         $country->getCode()->willReturn('PL');
         $address->getCountryCode()->willReturn('PL');
         $countryRepository->findOneBy(['code' => 'PL'])->willReturn($country);
@@ -79,6 +81,8 @@ final class ProvinceAddressConstraintValidatorSpec extends ObjectBehavior
         ProvinceAddressConstraint $constraint,
         ExecutionContextInterface $context
     ): void {
+        // DO NOT REMOVE THIS CODE: it ensures that there is a violation that can be added
+        // if the logic that is being tested (NOT adding a validation) does not work.
         $country->getCode()->willReturn('PL');
         $address->getCountryCode()->willReturn('PL');
         $countryRepository->findOneBy(['code' => 'PL'])->willReturn($country);


src/Sylius/Bundle/AddressingBundle/Validator/Constraints/ProvinceAddressConstraintValidator.php
```diff
@@ -47,7 +47,7 @@ public function validate($value, Constraint $constraint): void
         $propertyPath = $this->context->getPropertyPath();
 
         foreach (iterator_to_array($this->context->getViolations()) as $violation) {
-            if (0 === strpos($violation->getPropertyPath(), $propertyPath)) {
+            if ('' === $propertyPath || 0 === strpos($violation->getPropertyPath(), $propertyPath)) {
```
vvasiloi, line 50: Wouldn't that skip the validation if you pass to the validator an instance of Address class and there's a violation before this constraint is validated?
The property path is an empty string when the root value is being validated.
The way you wrote the condition (`'' === $propertyPath ||`), will check if the property path is not an empty string first, which will always be true when the root object is being validated, so it won't even evaluate the left side of the `||`.
If there's is a constraint before this one that results in a violation, then it will enter the loop and will return immediately.

I suggest changing the condition as show below.
Also, a test is definitely necessary here, especially for the unfortunate scenario I described.
Can you, please, change the condition and also update the PHP Spec file, @rimas-kudelis?

```suggestion
            if ('' !== $propertyPath && 0 === strpos($violation->getPropertyPath(), $propertyPath)) {
```

vvasiloi (changes requested):

src/Sylius/Bundle/AddressingBundle/Validator/Constraints/ProvinceAddressConstraintValidator.php
rimas-kudelis: > Wouldn't that skip the validation if you pass to the validator an instance of Address class and there's a violation before this constraint is validated?

Like I mentioned in the initial comment, I suppose that's the whole point of this condition in the first place: if an address is deemed invalid already, this validator does not perform further checks. The only problem was the root object case, due to how `strpos()` works.

rimas-kudelis: > Also, a test is definitely necessary here

Good point! I was reluctant to add one, because I don't have a quick way to run them. But t I've added one now (which iis mostly a copy of one that already exists).

rimas-kudelis: @vvasiloi I'll leave it for you to mark this conversation as resolved if you agree with me.

vvasiloi: Thank you for the clarification!
Unfortunately, now I have more questions about the whole `foreach` block... 😅 
I wish there would be some unit tests for it, because the spec file doesn't answer those questions.
Speaking of spec, I have a feeling that the test you added will give false-positive results, just like the one you copied it from.
It expects that `addViolation` should not be called, but without proper setup on the address, it will never be called, because the call is wrapped in an `if` statement that does the actual validation.
You can probably remove the foreach and the test will still pass...
WDYT?

rimas-kudelis: @vvasiloi the way I understand it is:
- This validator checks an Address **Object**, because it needs to look at two of its fields: `country` and `province`
- Before doing the actual validation, it loops (hence the `foreach`) through the violations already in the context, and checks if there's at least one added to either **the Address object itself**, or **one of its properties** (hence the `strpos` instead of a simple comparison). If it finds at least one violation on the address/its property, it skips its own validation.

For example, if you're validating an order that is being submitted, and a shipping addess has an invalid country, or province (or actually any property that is invalid), that violation will have a property path like `shipping_address.country`, and this validator, which will be defined with a `shipping_address` property path will return early.

Similarly, if the address itself already has a violation (property path `shipping_address`), that will also match the `strpos()` test and cause an early return.

On the other hand, if all violations will have unrelated property paths (e.g. `billing_address`), this validator **will** perform its check, because it's property path `shipping_address` will not match the beginning of all these property paths with which the violations are added.

But the problem arises when our property path is empty, because `strpos('some_path', '')` does not simply return `0` or `false`, but triggers an error due to its second argument (`needle`) being an empty string (because, it's indeed weird to test whether an empty string is a substring of any other string).  That's what my patch fixes.

You're probably right about the test though: it seems rather superfluous and would probably not fail even without the `foreach` loop. Hmm...

src/Sylius/Bundle/AddressingBundle/spec/Validator/Constraints/ProvinceAddressConstraintValidatorSpec.php
```diff
@@ -62,6 +62,23 @@ function it_does_not_add_violation_because_a_violation_exists(
         $this->validate($address, $constraint);
     }
 
+    function it_does_not_add_violation_because_a_violation_exists_when_address_is_the_root_object(
+        AddressInterface $address,
+        ProvinceAddressConstraint $constraint,
+        ExecutionContextInterface $context
+    ): void {
```
vvasiloi, line 69:
```suggestion
        ExecutionContextInterface $context,
        Country $country,
        RepositoryInterface $countryRepository
    ): void {
        $country->getCode()->willReturn('PL');
        $address->getCountryCode()->willReturn('PL');
        $countryRepository->findOneBy(['code' => 'PL'])->willReturn($country);

        $country->hasProvinces()->willReturn(true);
        $address->getProvinceCode()->willReturn(null);
```

src/Sylius/Bundle/AddressingBundle/Validator/Constraints/ProvinceAddressConstraintValidator.php
vvasiloi: It's probably fine, no need to complicate matters.
I suggested a change for the test, take a look.

src/Sylius/Bundle/AddressingBundle/spec/Validator/Constraints/ProvinceAddressConstraintValidatorSpec.php
vvasiloi: Same should probably be done for `it_does_not_add_violation_because_a_violation_exists`...

rimas-kudelis: Done.
Kinda easy, when all I had to do was copy and paste. :D

vvasiloi: You wish it was like that every time, don't you? 😄

vvasiloi (approved):
Looks good to me. Thank you, @rimas-kudelis!

rimas-kudelis:
@lchrusciel ping

src/Sylius/Bundle/AddressingBundle/spec/Validator/Constraints/ProvinceAddressConstraintValidatorSpec.php
```diff
@@ -62,6 +72,32 @@ function it_does_not_add_violation_because_a_violation_exists(
...
+        Country $country,
+        ProvinceAddressConstraint $constraint,
+        ExecutionContextInterface $context
+    ): void {
+        $country->getCode()->willReturn('PL');
+        $address->getCountryCode()->willReturn('PL');
+        $countryRepository->findOneBy(['code' => 'PL'])->willReturn($country);
+
+        $country->hasProvinces()->willReturn(true);
+        $address->getProvinceCode()->willReturn(null);
```
Zales0123, line 87: I believe this and the similar mocks in the spec above are not needed? These functions are never called as the `isProvinceValid` function is not reached. I even removed them and specs were still green 💃

Zales0123 (commented):
👍

src/Sylius/Bundle/AddressingBundle/spec/Validator/Constraints/ProvinceAddressConstraintValidatorSpec.php
vvasiloi: I think I suggested to add those, but I can't remember why.
Were the tests false positive, @rimas-kudelis?

vvasiloi: @Zales0123 Yes, I think that was it... that's how we tested that we need the fix.
With the fix, the builds are green either way, but without the fix, it only fails with the proper setup (the mocks).
The mocks are required to prevent regression.

vvasiloi: @rimas-kudelis maybe leave a comment there, because in another month who knows if anyone can remember this. 😅

rimas-kudelis: I actually put that info in the commit message. But sure, I can put it in a comment as well.

vvasiloi: Right. I didn't see it because GitHub hides everything after the first line and you have to click a button to see it. Thanks!

rimas-kudelis: I've added a comment to both specs.

lchrusciel: I guess Mateusz was pointing out just the `$address->getProvinceCode()->willReturn(null);` line, as returning null is default phpspec behvaiour

lchrusciel (approved):

src/Sylius/Bundle/AddressingBundle/spec/Validator/Constraints/ProvinceAddressConstraintValidatorSpec.php
lchrusciel: Or after the second reading - shouldn't we rather take usage of `->shouldNotBeCalled()`?

vvasiloi: @lchrusciel You should read it again, I think you missed the point. It would be easier to understand if you would apply the spec changes without the fix and run phpspec.
Spoiler: it will fail, which is precisely the goal of the tests. Without those changes in the spec, it will pass.

rimas-kudelis: @vvasiloi still, `shouldNotBeCalled()` could be a valid way out too.
But I'm wondering if that wouldn't be an assertion of an implementation detail.

vvasiloi: Yes, that's an implementation detail, but it's not even what is tested here.
Usually,  methods like `shouldNotBeCalled`, which are asserts for behavior, are used for the test subject, which in this case is the validator class.
The address and it's related objects are the input in this case, which is how the test should be set up for failure **before** writing the implementation.
In this case, originally, it was probably written **after** the implementation, so it never had the proper input setup.
The test was not useful from the start, because it passes both with and without the implementation.
That's why TDD is about writing tests first, see that it fails, write the implementation to pass the test.

vvasiloi: @rimas-kudelis I checked it out again and I see that `it_does_not_add_violation_because_a_violation_exists_when_address_is_the_root_object` passes with and without your fix. 🤔 
How did this happen?

lchrusciel: @vvasiloi I get you point, but `$address->getProvinceCode()->willReturn(null);` should not change anything, as this is default behaviour. On the other hand, keeping it won't change anything so we can leave it as is.

Still it probably requires manual testing from our side

rimas-kudelis: @lchrusciel I added `->willReturn(null)` just to make it more explicit that stuff would not match. That line is certainly not necessary for the test, I won't deny that. :)

lchrusciel: Hey folks, I've tried to extract bloated changes to separate class, so it would be easier for me to comprehend what is happening here. You can find something here: https://github.com/Sylius/Sylius/pull/13599. My findings are as follows:
1. "passing empty needle to strpos()" triggers php warning, should be omitted. Therefore the bug is present
2. Since #13562 in branch 1.11 we are using string_starts_with(), so it is fix only for 1.10
3. There is no value in the proposed tests in my opinion. I'm willing to merge just one line fix in validator if we will remove proposed specs.
4. If we want to move forward, some refactor of this logic is a musthave. Amount of different paths here is just hilarious. I'm not insisting on my solution (especially that bc layer may be controversial), but for sure we should do something about it if we want to touch this class more. Also, the refactor should probably be merge to master branch.

rimas-kudelis: @lchrusciel are you saying the validator itself should be refactored, or were you talking about the tests?
The validator ir just 92 lines – that doesn't sound too bloated to me (although I didn't try to analyze its logic now).

rimas-kudelis: Also, regarding the tests, as mentioned above, we saw that current tests were passing regardless of whether or not the functionality in question worked, so we tried extending them to not pass that easy. And the special cases for `address_is_the_root_object` were added because this is a fix for a bug which only manifests itself when the address is the root object.

lchrusciel: Class is short, but have 10 if statements and it makes writing test quite painful. Also, I get the reason of your test - but they don't change anything. Tests are green no matter if this fix is added or not.

lchrusciel: image

rimas-kudelis: and if you comment out that `foreach` block?

lchrusciel: Ok, so I've tested it on PHP8.0, but the warning is triggered on PHP7.4 http://sandbox.onlinephpfunctions.com/code/ee553ff2e53d1ea3b454dea89f4dda61fdbfa3c3

rimas-kudelis: > Ok, so I've tested it on PHP8.0, but the warning is triggered on PHP7.4

OMG, and this change is not even documented by PHP!

Still, what happens if you comment out the `foreach`?

Zales0123: Let me come back to the discussion 😄 As far as I see, the warning about empty string indeed happens only on the PHP 7.4. I think it was one of the roots of the misunderstanding there 🚀 

Zrzut ekranu 2022-02-4 o 17 21 48

And **indeed** when the foreach is commented out, the specs are still green. However, as far as I see, in these tests that you're changing we can call `$address->getCountryCode()->shouldNotBeCalled();` instead of only `$context->addViolation(Argument::any())->shouldNotBeCalled();` so it ensures us the `isProvinceValid` is not used nor accessed. 

```php
    function it_does_not_add_violation_because_a_violation_exists_when_address_is_the_root_object(
        RepositoryInterface $countryRepository,
        AddressInterface $address,
        Country $country,
        ProvinceAddressConstraint $constraint,
        ExecutionContextInterface $context
    ): void {
        $this->initialize($context);

        $context->getPropertyPath()->willReturn('');
        $context->getViolations()->willReturn(new \ArrayIterator([
            $this->createViolation('property_path'),
        ]));

        $address->getCountryCode()->shouldNotBeCalled();
        $context->addViolation(Argument::any())->shouldNotBeCalled();

        $this->validate($address, $constraint);
    }
```

And even though it's not **strictly** connected with the proposed fix, it's fair enough to propose both of them together in one PR 🖖 What do you think about it @rimas-kudelis @lchrusciel?

vvasiloi: Those extra bits of setup that were added for both tests are for the whole foreach, not just for the condition that Rimas added, which can be reproduced with PHP 7.4.
If you revert everything in changed in this PR, or just start with a 1.10 branch and remove that foreach, then the tests will still pass. Somehow I forgot to mention this in my previous comments.

vvasiloi: Also, why are you not using https://3v4l.org/, @lchrusciel?

vvasiloi: You can see the result for all versions at once there: https://3v4l.org/5ma7S

rimas-kudelis: @Zales0123 like I said, I agree that asserting that `->getCountryCode()` must not be called would make the tests fail. But it feels to me like an assertion on **How the validator works** instead of **What the expected result is**. The way I wrote it allows the test subject to remain more of a blackbox, doesn't it?

vvasiloi: Sorry, I missed the Mateusz's latest message.
I second Rimas here and I also said something among those line earlier.
We're testing the validator here and the test is called `it_does_not_add_violation`, so I don't see why we would assert the behavior on the address object instead of the validator context.
There's a good synergy between: `it_does_not_add_violation` and `$context->addViolation(Argument::any())->shouldNotBeCalled();`

lchrusciel (approved):

lchrusciel:
Thank you, Rimas! :tada:

PR #13432, opened 2021-12-31
Update SalesDataProvider.php

### LINKED ISSUE #12049: Doctrine syntax in one place produces sql incompatible with postgres/standard sql
**Sylius version affected**: 1.8

**Description**  
https://github.com/Sylius/Sylius/blob/4b1ea10f1369cc49daa2ae074a4a3c5ca1dc0cad/src/Sylius/Component/Core/Dashboard/SalesDataProvider.php#L71

This entire switch statement is affected in multiple places.  

Produces SQL code incompatible with postgres and is probably non standard sql. I have been running Sylius on Postgres for almost two years in production. 

You cannot rely on aliases in a having clause as it produces 
Screen Shot 2020-11-17 at 12 30 20 AM

See https://www.postgresql.org/message-id/18195.1038530280@sss.pgh.pa.us

The following code executes correctly but is obviously ugly.
```
  $queryBuilder
                    ->addSelect('YEAR(o.checkoutCompletedAt) as year')
                    ->addSelect('MONTH(o.checkoutCompletedAt) as month')
                    ->groupBy('year')
                    ->addGroupBy('month')
                    ->andHaving($queryBuilder->expr()->orX(
                        'YEAR(o.checkoutCompletedAt) = :startYear AND YEAR(o.checkoutCompletedAt) = :endYear AND MONTH(o.checkoutCompletedAt) >= :startMonth AND MONTH(o.checkoutCompletedAt) <= :endMonth',
                        'YEAR(o.checkoutCompletedAt) = :startYear AND YEAR(o.checkoutCompletedAt) != :endYear AND MONTH(o.checkoutCompletedAt) >= :startMonth',
                        'YEAR(o.checkoutCompletedAt) = :endYear AND YEAR(o.checkoutCompletedAt) != :startYear AND MONTH(o.checkoutCompletedAt) <= :endMonth',
                        'YEAR(o.checkoutCompletedAt) > :startYear AND YEAR(o.checkoutCompletedAt) < :endYear'
                    ))

```

remoteclient:
This would fix https://github.com/Sylius/Sylius/issues/12049

| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | fixes #12049
| License         | MIT

The sales provider uses non standard sql code. On postgresql systems the admin area is not working. The changes makes the admin area work.

### DIFF
--- a/src/Sylius/Component/Core/Dashboard/SalesDataProvider.php
+++ b/src/Sylius/Component/Core/Dashboard/SalesDataProvider.php
@@ -48,7 +48,7 @@ final class SalesDataProvider implements SalesDataProviderInterface
                 $queryBuilder
                     ->addSelect('YEAR(o.checkoutCompletedAt) as year')
                     ->groupBy('year')
-                    ->andHaving('YEAR(o.checkoutCompletedAt) >= :startYear AND YEAR(o.checkoutCompletedAt) <= :endYear')
+                    ->andWhere('YEAR(o.checkoutCompletedAt) >= :startYear AND YEAR(o.checkoutCompletedAt) <= :endYear')
                     ->setParameter('startYear', $startDate->format('Y'))
                     ->setParameter('endYear', $endDate->format('Y'))
                 ;
@@ -66,7 +66,7 @@ final class SalesDataProvider implements SalesDataProviderInterface
                     ->addSelect('MONTH(o.checkoutCompletedAt) as month')
                     ->groupBy('year')
                     ->addGroupBy('month')
-                    ->andHaving($queryBuilder->expr()->orX(
+                    ->andWhere($queryBuilder->expr()->orX(
                         'YEAR(o.checkoutCompletedAt) = :startYear AND YEAR(o.checkoutCompletedAt) = :endYear AND MONTH(o.checkoutCompletedAt) >= :startMonth AND MONTH(o.checkoutCompletedAt) <= :endMonth',
                         'YEAR(o.checkoutCompletedAt) = :startYear AND YEAR(o.checkoutCompletedAt) != :endYear AND MONTH(o.checkoutCompletedAt) >= :startMonth',
                         'YEAR(o.checkoutCompletedAt) = :endYear AND YEAR(o.checkoutCompletedAt) != :startYear AND MONTH(o.checkoutCompletedAt) <= :endMonth',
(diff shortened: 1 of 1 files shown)


vvasiloi (changes requested):
@remoteclient thank you for your contribution.
While this solves the problem with Postgres, it creates another one for MySQL.
Please see the failing tests and try make it green again.

remoteclient:
Hile this commit seems to work with MySQL, it fails on Postgresql again with:
```
An exception has been thrown during the rendering of a template ("[Semantical Error] line 0, col 222 near 'YEAR(o.checkoutCompletedAt),': Error: Cannot group by undefined identification or result variable.").
```
Problem in the original file is, that it uses non-standard sql.

In standard sql HAVING is evaluated before SELECT, so the alias is not known to the HAVING clause.

vvasiloi:
How about using an alias in group by, but not in having?

remoteclient:
> How about using an alias in group by, but not in having?

As you can see here: [mysql having](https://www.mysqltutorial.org/mysql-having.aspx), in standard sql GROUP BY is evaluated before HAVING.

remoteclient:
Please check the results of total. I replaced the HAVING clause by WHERE

I don't believe it makes sense to use HAVING here because:

>WHERE clause introduces a condition on individual rows; HAVING clause introduces a condition on aggregations, i.e. results of selection where a single result, such as count, average, min, max, or sum, has been produced from multiple rows.

Your HAVING is not using aggregate functions, it uses MONTH, YEAR...

vvasiloi:
Indeed, it make more sense to first filter out the order that are outside the specified time range, then group them by the time unit.
I'm curious how it reflects on the performance.

flolivaud:
Hey, is it possible to move forward about this pull requests?

vvasiloi (approved):

vvasiloi:
I would love to see a performance comparison, but nevertheless this will fix the bug. Thank you, @remoteclient!

remoteclient:
@vvasiloi You are welcome! I am happy that I could contribute.

GSadee (approved):

GSadee:
Thank you, @remoteclient! :tada:

PR #13441, opened 2022-01-04
Use PATCH requests where PUT are used. Fixes #13416

### LINKED ISSUE #13416: [API] Use PATCH requests where PUT are used
It seems like Api Platform are planning to change the way PUT requests work, see api-platform/core#4344. The way they currently work is essentially that of PATCH, and the idea is to implement them propertly, while retaining the current functionality as PATCH proper.

Thus I think it might make sense for Sylius to move to PATCH before the new API gains wide adoption and changing it becomes painful.

rimas-kudelis:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | yes
| Deprecations?   | yes
| Related tickets | fixes #13416 
| License         | MIT

Some tests still might need fixing, but since so many fail, I'm just not sure which fail because of me and which don't. Comments welcome.


lchrusciel:
Latest master build seems to be green https://github.com/Sylius/Sylius/actions/runs/1653334588, therefore it seems to be related to your changes, at least PHPUnit changes. Behat with JS is not showing me anything, so it is hard to judge

lchrusciel:
Hey Rimas, 

taking look into spec test cases:
image

by default I would prefer to follow API Platform recommendation, unless we will find some better arguments for that. Thanks for your work and research 👍

rimas-kudelis:
Actually, this is a bit more complicated than I thought:
- Not returning `null` values affects all requests, not just PATCH ones, unless perhaps if I specify the config value for each PATCH request separately (haven't tried that yet).
- API Platform requires the `Content-Type` header to be different for PATCH requests (`application/merge-patch+json` instead of `application/ld+json`) by default.
- Quick googling brought me to json-ld/json-ld.org#599 – apparently, a patch format specification for JSON-LD is yet to be produced.

Seeing how APIP's implementation already supports IRIs in relations in PATCH requests, I'd say in a way they're already using "JSON Merge PATCH on steroids", because it combines the IRI feature from LD with the format of Merge Patch.

Also, I'm not sure what happened, but today migration tests fail, so the API tests that I updated aren't even being run.

tests/Api/JsonApiTestCase.php
```diff
@@ -17,7 +17,8 @@
 
 abstract class JsonApiTestCase extends BaseJsonApiTestCase
 {
-    public const CONTENT_TYPE_HEADER = ['CONTENT_TYPE' => 'application/ld+json', 'HTTP_ACCEPT' => 'application/ld+json'];
```
vvasiloi, line 20: This might break someone's tests... better add a note in the upgrade file.

lchrusciel:
Hey Rimas, I'm ok with targetting 1.11. The migration issue was solved yesterday. Can you also a note requested by Victor?

rimas-kudelis:
Interestingly, there are still issues, and I'm quite surprised they haven't manifested before, because I believe I can reproduce them on 1.11 alpha, when issuing a PUT request (not a PATCH one). The problem is in two tests, both of which try to update items in an indexed (is it the correct term?) collection, such as:

```http
PUT /api/v2/admin/product-variants/product1

{
    "channelPricings": {
        "default": {
            "@id": "/api/v2/admin/channel-pricings/1",
            "price": 3000,
            "originalPrice": 4000
        }
    }
}
```

Somehow, at least for me this fails to resolve channel pricing with an id 1 and instead tries to add a new pricing to the collection, without a channel name (`default`) set. Which then fails validation (when I issue the PUT request to my installation) or causes an exception during DB insert (in tests).

I'm not sure why this would be happening. Any ideas?

rimas-kudelis:
Oh, I just figured it out, it's due to the Content-Type I was sending (`application/json` instead of `application/ld+json`).

rimas-kudelis:
Soo, I wonder what course to take now.

If I reserve `application/merge-patch+json` as a patch mimetype for `jsonld`, that will probably prevent users from using it with simple JSON PATCH requests. Not sure if that would cause any negative consequences, but it might not be desired.

An alternative would be to keep using `application/ld+json` maybe, but that would kinda make us less in line with the JSON Merge PATCH spec. Although I actually prefer this approach and will try to implement it if nobody objects.

Another alternative would be to use a custom mimetype (e.g. `application/merge-patch+ld+json`), but I suppose we wouldn't want to go there.

PR #13464, opened 2022-01-11
Bugfix: display full cart promotion rule form when adding a new rule

### LINKED ISSUE #13463: [AdminBundle]:  BUG: "Add rule" in cart promotions no longer opens full rule form
After catalog promotions were merged and rules renamed to scopes, the "New rule" button in Cart promotion management window opens an incomplete rule form. This happens because when catalog promotion rules were renamed to scopes, the respective JS snippet (which I guess worked for both Catalog and Cart rules) was also modified without being duplicated first (see the changes to app.js in 821247f972e20ced74977a30d6a5399e9e00a5ef). After this change, the `change()` event is no longer triggered when a new rule is added, so the user has to select a different rule type first for the full rule form to be opened.

rimas-kudelis:
Fixes #13463

| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #13463
| License         | MIT

Additionally to the bugfix, I made the jquery selectors more explicit by using direct ancestors. In fact, that's how I discovered this bug, because we were patching app.js to make these selectors more explicit, which allowed us embed additional collection forms inside action/rule collections forms.

### DIFF
--- a/features/promotion/managing_catalog_promotions/creating_catalog_promotion.feature
+++ b/features/promotion/managing_catalog_promotions/creating_catalog_promotion.feature
@@ -131,3 +131,11 @@ Feature: Creating a catalog promotion
         Then there should be 1 new catalog promotion on the list
         And it should have "winter_sale" code and "Winter sale" name
         And it should have priority equal to 10
+
+    @ui @javascript
+    Scenario: Adding a new catalog promotion of default type with one action
+        When I want to create a new catalog promotion
+        And I add a new catalog promotion scope
+        And I add a new catalog promotion action
+        Then I should see the catalog promotion scope configuration form
+        And I should see the catalog promotion action configuration form
--- a/src/Sylius/Behat/Page/Admin/Promotion/CreatePage.php
+++ b/src/Sylius/Behat/Page/Admin/Promotion/CreatePage.php
@@ -26,7 +26,7 @@ class CreatePage extends BaseCreatePage implements CreatePageInterface
     use NamesIt;
     use SpecifiesItsCode;
 
-    public function addRule(string $ruleName): void
+    public function addRule(?string $ruleName): void
     {
         $count = count($this->getCollectionItems('rules'));
 
@@ -36,7 +36,9 @@ class CreatePage extends BaseCreatePage implements CreatePageInterface
             return $count + 1 === count($this->getCollectionItems('rules'));
         });
 
-        $this->selectRuleOption('Type', $ruleName);
+        if (null !== $ruleName) {
+            $this->selectRuleOption('Type', $ruleName);
+        }
     }
 
     public function selectRuleOption(string $option, string $value, bool $multiple = false): void
@@ -74,7 +76,7 @@ class CreatePage extends BaseCreatePage implements CreatePageInterface
         $lastAction->fillField($option, $value);
     }
 
-    public function addAction(string $actionName): void
+    public function addAction(?string $actionName): void
     {
         $count = count($this->getCollectionItems('actions'));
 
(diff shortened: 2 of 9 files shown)


src/Sylius/Behat/Page/Admin/Promotion/CreatePage.php
```diff
@@ -74,17 +76,19 @@ public function fillRuleOptionForChannel(string $channelName, string $option, st
...
 
-    public function addAction(string $actionName): void
+    public function addAction(?string $actionName): void
     {
         $count = count($this->getCollectionItems('actions'));
 
         $this->getDocument()->clickLink('Add action');
 
-        $this->getDocument()->waitFor(5, function () use ($count) {
+        $this->getDocument()->waitFor(500, function () use ($count) {
```
vvasiloi, line 85: That's too much timeout.

vvasiloi (changes requested):

vvasiloi (approved):

rimas-kudelis:
I really want to thank @vvasiloi for guiding me over the process of writing these behat scenarios, so – thank you!

Roshyo (approved):

lchrusciel (approved):

lchrusciel:
Thank you, Rimas! :1st_place_medal:

features/promotion/managing_catalog_promotions/creating_catalog_promotion.feature
```diff
@@ -131,3 +131,11 @@ Feature: Creating a catalog promotion
         Then there should be 1 new catalog promotion on the list
         And it should have "winter_sale" code and "Winter sale" name
         And it should have priority equal to 10
+
+    @ui @javascript
```
lchrusciel, line 135: we could add `@no-api` tag

rimas-kudelis:
@lchrusciel so should I add the tag in a new PR? :grinning:

features/promotion/managing_catalog_promotions/creating_catalog_promotion.feature
vvasiloi: I don't thinks so. API scenarios should have the `@api` tag and you either run Behat with `--tags="@api"` to include API scenarios or with `--tags="~@api"` to exclude them.

rimas-kudelis: `@no-api` is an already existing tag though. Perhaps it would make sense to remove it everywhere, if `~@api` is to be used instead.

vvasiloi: Then I'm curious why it exists...

vvasiloi: By the looks of it, Lukasz is the one who introduced the tag and it was added to about 80 scenarios, but I don't see it actually being used. Perhaps it's obsolete.  Let's see what Lukasz has to say in his defense.

GSadee: The `@no-api` tag was created to mark scenarios that we don't want to cover in API. The main reason was to be aware of how many features and scenarios still need to be covered

vvasiloi: Thanks for the clarification. It make a bit more sense now.
At this point, isn't it already easier to explicitly mark the scenarios that we want to cover in the API and assume that the rest are "@no-api"?
I'm just concerned that this is going to be confusing for users and contributors and probably even become redundant once all current features are covered in the new API.

PR #13465, opened 2022-01-11
Catalog promotion priority

Rafikooo:
| Q               | A
| --------------- | -----
| Branch?         | 1.11 
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/features/promotion/managing_catalog_promotions/prioritizing_catalog_promotions.feature
+++ b/features/promotion/managing_catalog_promotions/prioritizing_catalog_promotions.feature
@@ -1,0 +1,74 @@
+@managing_catalog_promotions
+Feature: Prioritizing a catalog promotion
+    In order to avoid the same catalog promotion priority
+    As an Administrator
+    I want to set distinct priority on multiple catalog promotions
+
+    Background:
+        Given the store operates on a single channel in "United States"
+        And there is a catalog promotion "Winter sale" with priority 20
+        And there is a catalog promotion "Year-end sale" with priority 30
+        And there is a catalog promotion "Spring sale" with priority 40
+        And I am logged in as an administrator
+
+    @api @ui
+    Scenario: Adding a catalog promotion with a priority higher than all existing catalog promotions does not change priority values
+        When I create a new catalog promotion with "collection_sale" code and "Collection sale" name and 50 priority
+        Then there should be 4 catalog promotions on the list
+        And the catalog promotion named "Winter sale" should have priority 20
+        And the catalog promotion named "Year-end sale" should have priority 30
+        And the catalog promotion named "Spring sale" should have priority 40
+        And the catalog promotion named "Collection sale" should have priority 50
+
+    @api @ui
+    Scenario: Adding a catalog promotion with a priority lower than all existing ones
+              increases the priority value of other catalog promotions by 1
+        When I create a new catalog promotion with "collection_sale" code and "Collection sale" name and 10 priority
+        Then there should be 4 catalog promotions on the list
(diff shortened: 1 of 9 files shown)


tests/Api/Responses/Expected/admin/catalog_promotion/put_catalog_promotion_response.json
```diff
@@ -7,7 +7,7 @@
     "code": "mugs_discount",
     "startDate": null,
     "endDate": null,
-    "priority": 1000,
+    "priority": 0,
```
Rafikooo, line 10: Having one row in the database when updating, gedmo sets the priority to 0, even if the value given is different

features/promotion/managing_catalog_promotions/seeing_catalog_promotion_details.feature
```diff
@@ -12,7 +12,6 @@ Feature: Seeing catalog promotion's details
         And it applies also on "T-Shirt" product
         And it reduces also price by fixed "$10.00" in the "Web-US" channel
         And the catalog promotion "Winter sale" operates between "2021-11-10" and "2022-01-08"
-        And its priority is 1200
```
Rafikooo, line 15: The default priority value when creating is 0, which is in this scenario. This line updates the database row, but gedmo cannot allow you to set different priority values when only one row existing thus I decided to remove that assertion.
Alternatively, I can set this value to `1200` during creating, and then gedmo allows it

features/promotion/managing_catalog_promotions/seeing_catalog_promotion_details.feature
```diff
@@ -24,4 +23,3 @@ Feature: Seeing catalog promotion's details
         And it should apply on "PHP T-Shirt" variant
         And it should apply on "T-Shirt" product
         And it should start at "2021-11-10" and end at "2022-01-08"
-        And its priority should be 1200
```
Rafikooo, line 27: With current behavior, I can assert that the priority value is `0`

docs/cookbook/promotions/catalog-promotion-priority.rst
```diff
@@ -0,0 +1,20 @@
+How to manage catalog promotion priority?
```
GSadee, line 1: I wonder if the cookbook is a proper place for this explanation 🤔 I would be for the catalog promotion section in the book: https://docs.sylius.com/en/latest/book/products/catalog_promotions.html

features/promotion/managing_catalog_promotions/prioritizing_catalog_promotions.feature
```diff
@@ -0,0 +1,74 @@
...
+        When I create a new catalog promotion with "collection_sale" code and "Collection sale" name and 50 priority
+        Then there should be 4 catalog promotions on the list
+        And the catalog promotion named "Winter sale" should have priority 20
+        And the catalog promotion named "Year-end sale" should have priority 30
+        And the catalog promotion named "Spring sale" should have priority 40
+        And the catalog promotion named "Collection sale" should have priority 50
+
+    @api @ui
+    Scenario: Adding a catalog promotion with a priority lower than all existing ones
+              increases the priority value of other catalog promotions by 1
```
AdamKasp, line 25: I am not sure if we need these details, in the scenario we are describing what we do, not an effect, IMO assertions are enough to understand priorities logic :)

AdamKasp (approved):

GSadee (approved):

GSadee:
Thank you, Rafał! :tada:

Issue #12893, opened 2021-08-14
`yarn install` fails with error: no template named 'remove_cv_t' in namespace 'std'; did you mean 'remove_cv'?


honigkuchenbaer:
**Sylius version affected**: 1.10

**Description**  

When I follow the installation instructions for installing Sylius 1.10 on a brand new Mac, a compilation error happens while `yarn install`.

**Steps to reproduce**  

```
composer create-project sylius/sylius-standard acme
cd acme
# prepare database connection URL into .env properly
php bin/console sylius:install
yarn install
```

gives:

```
In file included from ../src/binding.cpp:1:
In file included from ../../nan/nan.h:58:
In file included from ~/.node-gyp/16.6.1/include/node/node.h:63:
In file included from ~/.node-gyp/16.6.1/include/node/v8.h:30:
~/.node-gyp/16.6.1/include/node/v8-internal.h:488:38: error: no template named 'remove_cv_t' in namespace 'std'; did you mean 'remove_cv'?
            !std::is_same<Data, std::remove_cv_t<T>>::value>::Perform(data);
                                ~~~~~^~~~~~~~~~~
                                     remove_cv
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/type_traits:776:50: note: 'remove_cv' declared here
template <class _Tp> struct _LIBCPP_TEMPLATE_VIS remove_cv
                                                 ^
1 error generated.
make: *** [Release/obj.target/binding/src/binding.o] Error 1
gyp ERR! build error
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/private/tmp/acme/node_modules/node-gyp/lib/build.js:262:23)
gyp ERR! stack     at ChildProcess.emit (node:events:394:28)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (node:internal/child_process:290:12)
gyp ERR! System Darwin 20.5.0
gyp ERR! command "/opt/homebrew/Cellar/node/16.6.1/bin/node" "/private/tmp/acme/node_modules/node-gyp/bin/node-gyp.js" "rebuild" "--verbose" "--libsass_ext=" "--libsass_cflags=" "--libsass_ldflags=" "--libsass_library="
gyp ERR! cwd /private/tmp/acme/node_modules/node-sass
gyp ERR! node -v v16.6.1
[...]

PR #13488, opened 2022-01-14
feat: use sass instead of node-sass

Nek-:
| Q               | A
| --------------- | -----
| Branch?         | master (could be done on others, your call)
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | yes (it won't work with gulp-sass 4.0 anymore I think)
| Deprecations?   | no/yes 
| Related tickets | #12893 
| License         | MIT

node-sass is deprecated and its installation breaks on many many setups depending on:
- Node version (up to 12 it seems fine, it breaks for 13+)
- gpy installed or not
- Already installed tooling
- Platform (linux/mac/windows)

All of these problems make the `sass` package a benediction. This commit makes the sylius codebase compatible with it.

### DIFF
--- a/UPGRADE-1.11.md
+++ b/UPGRADE-1.11.md
@@ -130,3 +130,27 @@ If you don't already have configured the messenger transport, configure it accor
 ### API v2
 
 For changes according to the API v2, please visit [API v2 upgrade file](UPGRADE-API-1.11.md).
+
+### Asset management changes
+
+We updated gulp-sass plugin as well as the sass implementation we use to be compatible with most installation
+([node-sass](https://sass-lang.com/blog/libsass-is-deprecated) is deprecated and incompatible with many systems).
+Therefore you need to update your code to follow this change.
+
+1. Change the gulp-sass version you are using to `^5.1.0` (package.json file)
+   ```diff
+   - "gulp-sass": "^4.0.1",
+   + "gulp-sass": "^5.1.0",
+   ```
+2. Add sass to your package.json:
+   ```diff
+   + "sass": "^1.48.0",
+   ```
+3. Follow [this guide](https://github.com/dlmanning/gulp-sass/tree/master#migrating-to-version-5) to upgrade your
+   code when using gulp-sass this is an example:
+   ```diff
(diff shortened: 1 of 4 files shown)


lchrusciel (commented):
I assume, that it will require update of package.json on the end developers apps. Can you add a note to UPGRADE.md?

Nek-:
@lchrusciel done!

lchrusciel:
Hey Maxime, to no hold you without answer, I've requested a review from @kulczy, as he is our main frontend guy. Next week I hope we will be ready to merge it :)

joshuabyler:
having this would be perfect to help out with updating node versions

lchrusciel:
it is still quite high on our list, we didn't forget about you

Nek-:
Rebased!

kulczy (approved):

lchrusciel:
Thank you, Maxime! :1st_place_medal:

Nek-:
@lchrusciel why isn't this merged in 1.11 ? 🤔 Do you need another PR ?

lchrusciel:
Hey Maxime,

I'm reluctant to js deps in the patch releases. If it would land in 1.11, after the update all users would have to change theirs package.json. Patch releases are considered as safer than minor releases from my experience, so I would prefer to leave it to a more conscious update - which is how I preserve minor updates.

porl:
Hi Ichrusciel

Does this mean it won't be in any release until 1.12? Or will there be a 1.11.x release with it?

I am currently on 1.10 and looking at building a newer site. I will hold off until 1.12 if this is the case so I don't need to manually patch things for my newer server (running Node 17).

Nek-:
@porl it will come with 1.12, not before, sorry.

lchrusciel:
Hey folks, Sylius v1.12 is not yet scheduled, but we will see if it can be sooner than later. 

We are looking for someone who will help us provide an anti-corruption layer to preserve backward compatibility with this task, but that may be hard as we are serving js together with the main Sylius codebase. If you have any advice, I'm open hearing them :)

Issue #13470, opened 2022-01-11
Remove ACL layer need for MasterRequest to MainRequest renaming 


lchrusciel:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Sylius version   | 1.12

- [x] Bump Symfony requirement to 5.4 on master
- [x] Remove all if/else statements introduced in https://github.com/Sylius/Sylius/pull/13251/

Ref: https://github.com/Sylius/Sylius/pull/13251#discussion_r778107488

GSadee:
I'm closing the issue as it has been fixed by https://github.com/Sylius/Sylius/pull/14735

PR #13489, opened 2022-01-17
[Maintenance] Drop Sf 4.4 support

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

As we are supporting Sf 5.4 on Sylius 1.11, there is no need to support two LTS on master. Especially, as we would like to support Symfony 6.0 ASAP


Zales0123 (commented):
Definitely, for 1.12 we should have 5.4 and 6.0 👍 Don't we have any workaround or deprecations we can drop/changed when dropping this version?

lchrusciel:
@Zales0123 for sure #13470 should be reverted. Do you think it should be part of this PR?

Zales0123:
It... can be 😄 On the other hand, we can do it also separately and it would be no harm as well 🖖

Zales0123 (approved):

.github/workflows/application.yml
```diff
@@ -161,8 +161,8 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: [ "8.0" ]
-                symfony: [ "^4.4", "^5.2" ]
+                php: ["8.0"]
+                symfony: ["^5.2"]
```
loic425, line 165: ^5.4 ?

loic425 (changes requested):

loic425 (approved):

src/Sylius/Bundle/MoneyBundle/composer.json
```diff
@@ -27,8 +27,8 @@
     "require": {
         "php": "^8.0",
         "sylius/resource-bundle": "^1.7",
-        "symfony/framework-bundle": "^4.4 || ^5.4",
-        "symfony/intl": "^4.4 || ^5.4",
+        "symfony/framework-bundle": "^5.4",
+        "symfony/intl": "^5.4",
         "symfony/templating": "^4.4 || ^5.2",
```
vvasiloi, line 32: What about this one?

composer.json
```diff
@@ -68,46 +68,46 @@
...
+        "symfony/property-info": "^5.4",
+        "symfony/proxy-manager-bridge": "^5.4",
+        "symfony/routing": "^5.4",
+        "symfony/security-bundle": "^5.4",
+        "symfony/security-core": "^5.4",
+        "symfony/security-csrf": "^5.4",
+        "symfony/security-guard": "^5.4",
+        "symfony/security-http": "^5.4",
+        "symfony/serializer": "^5.4",
         "symfony/string": "^5.3",
```
vvasiloi, line 103: Since this PR also _secretly_ drops support for Symfony 5.2 and 5.3, maybe also bump this requirement.
```suggestion
        "symfony/string": "^5.3",
```

Roshyo:
Since we now support ^5.4, we can also go through the `MasterRequest` calls I believe. (Or right after this PR is merged perhaps ?)

Roshyo (approved):

src/Sylius/Bundle/MoneyBundle/composer.json
lchrusciel: wrong regexp 👍

composer.json
lchrusciel: wrong regexp 👍

vvasiloi (approved):

GSadee (approved):

PR #13491, opened 2022-01-17
[API][Minor] Archiving shipping methods filter rename to maintain naming convention

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

I've moved `filter` word to the end of the service name, as this is a convention we have for the rest of the filters. The question arises if we shouldn't move to the old Symfony service convention for filters to have the same structure here, but I don't have any strong opinion about it.


GSadee (approved):

GSadee:
Thank you, Łukasz! :tada:

PR #13492, opened 2022-01-17
Add administrator timezone

ernestWarwas:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets |
| License         | MIT

This is the first part of this task, the second will be display dates in chosen timezone.
Added a new field called timezone to preferences when editing or creating a new admin user.

Zrzut ekranu 2022-01-17 o 13 08 06


src/Sylius/Component/Core/Model/AdminUser.php
```diff
@@ -94,4 +96,14 @@ public function setAvatar(?ImageInterface $avatar): void
     {
         $this->setImage($avatar);
     }
+
+    public function getTimezone(): ?string
```
vvasiloi, line 100: This should be copied to the interface.

src/Sylius/Component/Core/Model/AdminUser.php
```diff
@@ -94,4 +96,14 @@ public function setAvatar(?ImageInterface $avatar): void
...
+    {
+        return $this->timezone;
+    }
+
+    public function setTimezone(?string $timezone): void
```
vvasiloi, line 105: This should be copied to the interface.

vvasiloi (commented):
Nice feature.
I wonder if this can be added without a BC break... 🤔

Roshyo:
My only question is: why ?

Why is that needed ?

vvasiloi:
> My only question is: why ?
> 
> Why is that needed ?

Elementary, my dear Stephane.
For international teams and companies, where not everyone lives in the same timezone.
But sometimes, even for national ones, in case of US or Russia (which spans 11 timezones).
By the looks of it, this PR only introduces the option to set a timezone for admin users.
To complete the feature, it also needs to set the timezone when rendering dates in admin panel.

Roshyo:
I agree, but then this feature should come with a behaviour of that sort. Because as is, this brings no value IMO

vvasiloi:
It's probably still work in progress, even though it wasn't marked as draft. Or maybe there will be a complementary PR.

src/Sylius/Behat/Context/Api/Admin/ManagingAdministratorsContext.php
```diff
@@ -182,6 +182,22 @@ public function iRemoveTheAvatarImage(): void
...
+        $this->client->addRequestData('timezone', $timezone);
+    }
+
+    /**
+     * @When I choose :timezone as my timezone
+     */
+    public function iChooseAsMyTimezone(string $timezone): void
+    {
+        $this->client->addRequestData('timezone', $timezone);
+    }
```
Zales0123, line 199: These two steps can be merged, as they have the same body:

```php
/**
 * @When I specify its timezone as :timezone
 * @When I choose :timezone as my timezone
 */
public function iSpecifyItsTimezoneAs(string $timezone): void
{
    $this->client->addRequestData('timezone', $timezone);}
```

src/Sylius/Behat/Page/Admin/Administrator/CreatePage.php
```diff
@@ -42,12 +42,18 @@ public function specifyLocale(string $localeCode): void
...
+        $this->getElement('timezone')->selectOption($timezone);
+    }
+
     protected function getDefinedElements(): array
     {
         return array_merge(parent::getDefinedElements(), [
             'email' => '#sylius_admin_user_email',
             'enabled' => '#sylius_admin_user_enabled',
             'locale_code' => '#sylius_admin_user_localeCode',
+            'timezone' => '#sylius_admin_user_timezone',
```
Zales0123, line 56: alphabetical order R.I.P [*] 😢

src/Sylius/Behat/Page/Admin/Administrator/UpdatePageInterface.php
```diff
@@ -30,4 +30,8 @@ public function changeLocale(string $localeCode): void;
     public function removeAvatar(): void;
 
     public function hasAvatar(string $avatarPath): bool;
+
+    public function changeTimezone(string $timezone): void;
```
Zales0123, line 34: I would put this method under `changeLocale`, to keep all `change*` methods together

src/Sylius/Bundle/CoreBundle/Resources/config/validation/AdminUser.xml
```diff
@@ -51,5 +51,8 @@
                 <option name="groups">sylius</option>
             </constraint>
         </property>
+        <property name="timezone">
+            <constraint name="Timezone"/>
```
Zales0123, line 55: I believe we also need an API Behat scenario to check this behaviour. I think it would not work right now 😄 as it does not have a proper `sylius` validation group defined

tests/Api/Admin/AdminUsersTest.php
```diff
@@ -41,4 +43,42 @@ public function it_allows_admin_users_to_log_in(): void
...
+                'timezone' => 'Europe/Paris'
+            ])
+        );
+
+        $response = $this->client->getResponse();
+
+        $this->assertResponse($response, 'admin/administrators/change_timezone_admin_response', Response::HTTP_OK);
+    }
+
+    private function loadFixturesAndGetAdminUser(): AdminUserInterface
```
Zales0123, line 69: I believe we can skip this private function and use `loadFixturesFromFile` method directly in the test function. 1 more line there - 6 less lines here :dancer:

src/Sylius/Behat/Page/Admin/Administrator/UpdatePageInterface.php
ernestWarwas: I've changed it too in UpdatePage 😃

src/Sylius/Bundle/CoreBundle/Migrations/Version20220114153236.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    public function getDescription(): string
+    {
+        return 'Add timezone to AdminUser entity';
+    }
+
+    public function up(Schema $schema): void
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE sylius_admin_user ADD timezone VARCHAR(255) DEFAULT NULL');
```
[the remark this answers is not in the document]
ernestWarwas: Added 👍

jakubtobiasz:
We decided to close this PR as in the nearest feature we don't plan to provide such functionality. Merging it now may lead to confusion and raising "bug" tickets. Thanks anyway 🖖🏻.

PR #13494, opened 2022-01-17
[API] Taxon filter moved to Doctrine\Filter namespace to make it coherent with other doctrine filters

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT


Roshyo:
Isn't this a BC break ?

lchrusciel:
It is, but API is still experimental, so I'm allowed to do it. Also, that's why I've left the note in upgrade file

Roshyo (approved):

PR #12329, opened 2021-02-10
[Api] Add test app for ApiBundle

Tomanhez:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Tests for adding new api resource for new entities: 
- Not sylius resource entity: `Foo`
- Sylius resource entity: `FooSyliusResource`
- Extended sylius resource entity: `Taxon`

Tests for overwriting Sylius ApiPlatform's resource entities : `Promotion`

To test locally, run commands:
`(cd src/Sylius/Bundle/ApiBundle && composer install)`
`(cd src/Sylius/Bundle/ApiBundle/test && bin/console doctrine:database:create -e test)`
`(cd src/Sylius/Bundle/ApiBundle/test && bin/console doctrine:schema:update --force -e test)`
`(cd src/Sylius/Bundle/ApiBundle/test && bin/console assets:install public)`
`(cd src/Sylius/Bundle/ApiBundle/test && APP_ENV=test symfony serve)`

To run tests:
`(cd src/Sylius/Bundle/ApiBundle && vendor/bin/phpunit)`

Swagger is working properly:
Screenshot 2021-02-10 at 16 19 26
Screenshot 2021-02-11 at 09 08 00

PR #13507, opened 2022-01-19
Update security config to use the new authenticator-based system

rimas-kudelis:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

This patch migrates security config to the new authenticator-based system introduced in Symfony 5.1.
Guard-based authentication has been deprecated in Symfony 5.3. Luckily, not much needs to change for us since Sylius didn't have any custom Guard authenticators to begin with.

The patch also drops `always_authenticate_before_granting: true` from config. This option [has been deprecated](https://symfony.com/doc/5.4/reference/configuration/security.html#always-authenticate-before-granting) since Symfony 5.4, and it's unclear why or whether at all it was needed in the first place, and it caused me personally some headache (see symfony/symfony#43375).

Similar PR for Sylius Standard: Sylius/Sylius-Standard#660.

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/test/config/security.yaml
+++ b/src/Sylius/Bundle/ApiBundle/test/config/security.yaml
@@ -1,12 +1,12 @@
 security:
-    always_authenticate_before_granting: true
+    enable_authenticator_manager: true
     providers:
         sylius_api_admin_user_provider:
             id: sylius.admin_user_provider.email_or_name_based
         sylius_api_shop_user_provider:
             id: sylius.shop_user_provider.email_or_name_based
 
-    encoders:
+    password_hashers:
         sha512: sha512
         Sylius\Component\User\Model\UserInterface: sha512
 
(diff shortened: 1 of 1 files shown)


loic425 (approved):

loic425:
@Zales0123 @lchrusciel 
I checked the documentations, and everything seems ok. cf docs links below

* [Public access](https://symfony.com/doc/current/security.html#allowing-unsecured-access-i-e-anonymous-users)
* [JWT configuration](https://github.com/lexik/LexikJWTAuthenticationBundle/blob/2.x/Resources/doc/index.md#configuration)

rimas-kudelis:
BTW i'm not sure about about password hashing config (the three lines with `sha512` in the security config). I didn't touch these lines, bet are they really necessary?

loic425:
> BTW i'm not sure about about password hashing config (the three lines with `sha512` in the security config). I didn't touch these lines, bet are they really necessary?

Hum... In a Symfony app, that should be auto. I think we can do this in another PR.

rimas-kudelis:
The lines were added in #12329, so I suppose it has something to do with tests maybe.
But then I guess these lines should go into `test/security.yaml` instead. 

And there's no need to `auto` explicitly, it's the default anyway, isn't it? [Symfony docs](https://symfony.com/doc/5.4/reference/configuration/security.html) don't even mention these options anymore for some reason.

loic425:
> And there's no need to `auto` explicitly, it's the default anyway, isn't it? [Symfony docs](https://symfony.com/doc/5.4/reference/configuration/security.html) don't even mention these options anymore for some reason.

For the auto configuration, it's configured via flex recipe => https://github.com/symfony/recipes/blob/master/symfony/security-bundle/5.3/config/packages/security.yaml

Roshyo (approved):

Roshyo:
I also agree to remove it in a separate PR

lchrusciel (approved):
Thanks a lot! However, this config is used only in test env. Perhaps, we should add some recommendation in UPGRADE.md

lchrusciel:
may become messed-up.**

execute `git fetch origin && git reset "chore/update-security-config"` to update your local branch.
 
rimas-kudelis:
@lchrusciel but this change is specific to Symfony 5.1+. Shouldn't it target master since that's the only place where 4.4 is no longer supported?

lchrusciel:
may become messed-up.**

execute `git fetch origin && git reset "chore/update-security-config"` to update your local branch.
 
lchrusciel:
Hey Rimas, my mistake. Thanks for mentioning it

lchrusciel:
Thanks, Rimas! :tada:

GSadee (commented):
@rimas-kudelis If I see correctly, you've changed only `security.yaml` file in the test application in ApiBundle, what about the main application `config/packages/security.yaml`?

rimas-kudelis:
@GSadee ouch, that one somehow escaped my attention.

PR #13522, opened 2022-01-21
OAuth | `UserResponseInterface::getUsername()` can return integer

stloyd:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT


stloyd:
In fact the signature is not correct, I'm struggling to fix it here or there cause ie using GitHub OAuth you will get integer ID not string one.

Roshyo:
Shouldn't we then fix this on HWI first ?

stloyd:
Counter option https://github.com/hwi/HWIOAuthBundle/pull/1883, but still due too forced type-hint in Sylius it requires some fix.

PR #13528, opened 2022-01-24
Remove Symfony 4 circular dependency exception listener fix

### LINKED ISSUE #13525: Remove symfony/http-kernel 4 symbols usage
Like said on #10664, the class `Sylius\Bundle\CoreBundle\EventListener\CircularDependencyBreakingExceptionListener` uses symbols from `symfony/http-kernel` version 4. Sylius does not support Symfony 4 anymore so this class must be fixed to not use those symbols.

Then, symbols added [here](https://github.com/Sylius/Sylius/commit/e09551c016752eff03339cc90ab678afabc3765d) in the ComposerRequireChecker configuration could be removed from the whitelist.

mmenozzi:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #13525
| License         | MIT

I removed two classes and its tests:

- `Sylius\Bundle\CoreBundle\DependencyInjection\Compiler\CircularDependencyBreakingExceptionListenerPass`
- `Sylius\Bundle\CoreBundle\EventListener\CircularDependencyBreakingExceptionListener`

The second one (the listener) was marked as `@internal` so it can be removed without breaking the BC promise. The first one was not marked as internal so its removal is a BC break.

Let me know what you prefer, its only job was to add the listener but we can also not remove it and leave it empty.


mmenozzi:
@Sylius/core-team let me know what you think.

lchrusciel (approved):

lchrusciel:
Thanks, Manuele! :tada:

lchrusciel:
I've probably merged too quickly, but please add note about it in UPGRADE.md file. I don't see too much value in keeping it in the codebase

PR #13529, opened 2022-01-24
Exclude PHP Arkitect rules files from autoloader

mmenozzi:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

@lchrusciel `src/Sylius/*/*/phparkitect.php` files must be excluded from the autoloader because they use PHP Arkitect symbols which is dev dependency and not a production one.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -258,7 +258,8 @@
             "src/Sylius/*/*/Tests/",
             "src/Sylius/Component/Core/Test/Tests/",
             "src/Sylius/*/*/spec/",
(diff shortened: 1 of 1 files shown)


GSadee (approved):

GSadee:
Thank you, Manuele! :1st_place_medal:

PR #13531, opened 2022-01-24
Remove Sylius < 1.3 old Kernel class (#13524)

### LINKED ISSUE #13524: Remove old Kernel class from core bundle
The old Kernel class `Sylius\Bundle\CoreBundle\Application\Kernel` is deprecated since  Sylius 1.3. This class uses symbols from old dependencies, some of them discontinued.
With #10664 we had to whitelist the [those symbols](https://github.com/Sylius/Sylius/commit/30c42ff5a45e593e894d13060c6593178f27888b) due to this old Kernel class.

It's true that removing this class would be a BC break but it's also true that this class is deprecated since Sylius 1.3 and no one with a "modern" version of Sylius should be using it.

I propose to remove it, highlight its removal in the changelog and remove from ComposerRequireChecker whitelist its used symbols.

mmenozzi:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #13524 
| License         | MIT


Roshyo:
I agree with the changes, but still it is BC break imo, as we can not guarantee upgrade from 1.3 to 1.11 with this change. Can we ?

mbabker:
You might be able to blank out the `registerBundles()` method so anyone extending from the Sylius kernel isn't getting a fatally broken application, but flat out removing it has a lot more side effects than just requiring someone to change the class they're extending in their application (i.e. the Sylius version is referenced from this kernel in a handful of places within Sylius itself, or the plugin skeleton using the version info to load version-specific configs for the test application it provides).

mmenozzi:
Hi @mbabker, yes you're right. Indeed the build currently fails due to what you say.

What do you think if this class becomes final and we leave only the public constants in it? Making it final means that if someone is extending it he will get a fatal error but in my opinion this is a good thing: that class is not meant to be extended, it must be replaced with a real Kernel class in projects using it.

And again I strongly believe that no one that will upgrade to Sylius 1.11 is still using this old kernel in his project.

mbabker:
> What do you think if this class becomes final and we leave only the public constants in it?

I wouldn't break B/C by adding the "final" keyword.  Blank the class (minus the public constants) since the core Symfony kernel functions well enough and the application kernel is expected to deal with the micro-kernel trait and its methods, find a new home for the version constants so those can be properly deprecated, and be done with it until a 2.0 release comes around.  Yes, the kernel as it exists now is catastrophically broken, but I don't think it's worth the hard B/C break to just dump it entirely when there are a few options available that can mitigate the break, make the class at least usable, and not cause a domino effect with more nuanced B/C breaks (i.e. the constants).

vvasiloi:
I agree with @mbabker concerns and proposal.
Removing the class right now will at least break a lot of plugins that are using those constants (inherited from the skeleton).
I'm not sure why those constants were not moved when the class was deprecated and were actively used afterwards, but this a good opportunity to fix that.

mmenozzi:
Yes guys I agree that the class shouldn't be removed.

I do not agree about trying to make it work somehow. For me is much better to make it final and have an hard break instead creating something that might work and then having unexpected behavior maybe in production.

mbabker:
Making it final and flat out removing it are both hard B/C breaks that are going to bite people still using this kernel in the backside whenever they try to upgrade.  The only real benefit to adding the final keyword over removal is you don't have to deal with the constants right away.

Removing all of the methods from the class gets rid of the real problems for applications still inheriting from this kernel (namely the `registerBundles()` method existing and overriding the Flex-managed `config/bundles.php` file and the `registerContainerConfiguration()` method not supporting the `.yaml` file extension) and avoids breaking B/C for the sake of breaking B/C.  I don't think anyone is saying that this kernel class needs to be functional on its own like it could have been in the earlier versions, but making it so that it doesn't cause fatal errors just by extending it is much better than the current status quo.

mmenozzi:
@mbabker @vvasiloi, what do you think now? I still have to move constants to `Sylius\Bundle\CoreBundle\SyliusCoreBundle`.

vvasiloi:
LGTM.

mmenozzi:
@vvasiloi I moved the constants. What do you think?

vvasiloi:
@mmenozzi looks good. I found one more place where that references the kernel constant
https://github.com/Sylius/Sylius/blob/66ef7dd15bbe9fc11d3e5200d1809bb6846a603f/src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml#L111

lchrusciel (approved):
Hey Manuele, thanks for your effort. 

Micheal and Victor thanks for your review and shared thoughts about it. I'm with you one it, lets go with lesser evil for now. Perhaps we can add commented final key word, just to show our intention that this call should not be extended

mmenozzi:
Hi @vvasiloi and @lchrusciel, sorry for the delay.
I now removed the last usage of the Sylius\Bundle\CoreBundle\Application\Kernel::VERSION in [src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml](https://github.com/Sylius/Sylius/blob/66ef7dd15bbe9fc11d3e5200d1809bb6846a603f/src/Sylius/Bundle/CoreBundle/Resources/config/app/config.yml#L111).
I also added the final keyword to the old Kernel class.

This should be ready for merge now.

Roshyo (approved):

GSadee (approved):

GSadee:
Thanks, Manuele! :tada:

PR #13533, opened 2022-01-25
[WIP] Calendar class extraction

TheMilek:
| Q               | A
| --------------- | -----
| Branch?         | 1.11 
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| License         | MIT

In order to unify the usage of the same method, we decided to create a separate library to avoid creating services that are based on the same functionality and use one to many instead


composer.json
```diff
@@ -19,6 +19,12 @@
             "homepage": "https://github.com/Sylius/Sylius/contributors"
         }
     ],
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/AdamKasp/Calendar"
+        }
+    ],
```
Zales0123, line 27: `sylius/calendar` repository is [already on Packagist](https://packagist.org/packages/sylius/calendar), so we should be able to use it without using a custom repository setting. What we definitely should do is release a 0.1 version of it to not use `dev-master` branch 🖖

composer.json
```diff
@@ -113,7 +119,8 @@
         "webmozart/assert": "^1.9",
         "willdurand/hateoas": "^3.0",
         "willdurand/hateoas-bundle": "^2.0",
-        "winzou/state-machine-bundle": "^0.5"
+        "winzou/state-machine-bundle": "^0.5",
+        "sylius/calendar": "dev-configure-behat-step-with-calendar"
```
Zales0123, line 123: See above. It should also be sorted alphabetically

src/Sylius/Behat/Service/Provider/Calendar.php
```diff
@@ -14,7 +14,7 @@
 namespace Sylius\Behat\Service\Provider;
 
 use Sylius\Bundle\ShippingBundle\Provider\DateTimeProvider;
-use Sylius\Component\Promotion\Provider\DateTimeProviderInterface;
+use Sylius\Calendar\Provider\DateTimeProviderInterface;
 
 final class Calendar implements DateTimeProvider, DateTimeProviderInterface
```
Zales0123, line 19: The idea of extracting this component was partially to get rid of this class 😄 so we should remove it as it's not necessary anymore, we can use [FakeCalendar](https://github.com/Sylius/Calendar/blob/master/tests/Behat/Provider/FakeCalendar.php) class from the new component

Roshyo:
This is a big BC break as some application may rely on this class.
Also since the service changed 

```diff
-        <service id="Sylius\Component\Promotion\Provider\DateTimeProviderInterface" class="Sylius\Component\Promotion\Provider\Calendar" />
+        <service id="Sylius\Calendar\Provider\DateTimeProviderInterface" class="Sylius\Behat\Service\Provider\Calendar" />
```

src/Sylius/Behat/Resources/config/services.xml
```diff
@@ -90,7 +90,7 @@
             <argument>%kernel.project_dir%</argument>
         </service>
         <service
-            id="Sylius\Component\Promotion\Provider\DateTimeProviderInterface"
+            id="Sylius\Calendar\Provider\DateTimeProviderInterface"
```
AdamKasp, line 93: you should remove these configs and use service from `sylius/calendar`

src/Sylius/Behat/Service/Provider/Calendar.php
AdamKasp, line 19: this service shoud be removed and replaced by `fakeDateProvider` from `sylius/calendar`

src/Sylius/Bundle/PromotionBundle/Resources/config/services.xml
```diff
@@ -80,6 +80,6 @@
             <argument type="service" id="sylius.repository.catalog_promotion" />
         </service>
 
-        <service id="Sylius\Component\Promotion\Provider\DateTimeProviderInterface" class="Sylius\Component\Promotion\Provider\Calendar" />
+        <service id="Sylius\Calendar\Provider\DateTimeProviderInterface" class="Sylius\Behat\Service\Provider\Calendar" />
```
AdamKasp, line 83: this service should be replaced by `Sylius\Calendar\Provider\CalendarInterface` from `sylius/calendar`

AdamKasp (changes requested):

AdamKasp:
@Roshyo If I see correctly this service was added after `1.10` was released so it isn't BC break :)

composer.json
```diff
@@ -19,6 +19,12 @@
             "homepage": "https://github.com/Sylius/Sylius/contributors"
         }
     ],
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "https://github.com/Sylius/Calendar"
+        }
+    ],
```
Zales0123, line 27: Should not be needed, as `sylius/calendar` is already on Packagist

Roshyo:
@AdamKasp This service exists in 1.10.0, so it means that 1.11.0 will break with it, or am I mistaken ?

Zales0123:
@Roshyo as far as I see and remember `Sylius\Component\Promotion\Provider\DateTimeProviderInterface` does **not** exist in `1.10` (looking at [this file](https://github.com/Sylius/Sylius/blob/1.10/src/Sylius/Bundle/PromotionBundle/Resources/config/services.xml)) 💃 so it can be safely renamed. Usage of this component in Shipment instead of [the Calendar from ShippingBundle](https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/ShippingBundle/Provider/Calendar.php) would need proper BC layer 🖖

composer.json
```diff
@@ -59,6 +59,7 @@
         "sylius-labs/polyfill-symfony-event-dispatcher": "^1.0.1",
         "sylius-labs/polyfill-symfony-framework-bundle": "^1.0",
         "sylius-labs/polyfill-symfony-security": "^1.0",
+        "sylius/calendar": "v0.1.0",
```
Zales0123, line 62: I believe it should be `^0.1` here and everywhere else

GSadee:
Closing in favour of #13557

PR #13534, opened 2022-01-25
[Fixtures] Init data fixtures

loic425:
| Q               | A
| --------------- | -----
| Branch?         | foundry-fixtures
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | 
| Related tickets | 
| License         | MIT


src/Sylius/Bundle/CoreBundle/DataFixtures/ShopConfigurationFixtures.php
```diff
@@ -0,0 +1,30 @@
...
+
+class ShopConfigurationFixtures extends Fixture implements FixtureGroupInterface
+{
+    public function load(ObjectManager $manager): void
+    {
+    }
+
+    public static function getGroups(): array
+    {
+        return ['shop_configuration'];
```
vvasiloi, line 28: Does it make sense to inject the groups through the constructor?

vvasiloi (approved):

src/Sylius/Bundle/CoreBundle/DataFixtures/ShopConfigurationFixtures.php
loic425: Hum... for now I don't see any reason to change these groups.

loic425: Either we run all the tests (with all groups) or we can run all groups we want. So I don't see any reason to change these group names. But maybe I am wrong and will se some reasons on customization part. But for now, I suggest to keep it as simple as this.

PR #13539, opened 2022-01-25
[Fixtures] Country factory

loic425:
| Q               | A
| --------------- | -----
| Branch?         | foundry-fixtures
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/DataFixtures/Factory/CountryFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/DataFixtures/Factory/CountryFactory.php
@@ -36,7 +36,7 @@ use Zenstruck\Foundry\Proxy;
  * @method static CountryInterface[]|Proxy[] randomRange(int $min, int $max, array $attributes = [])
  * @method CountryInterface|Proxy create(array|callable $attributes = [])
  */
-final class CountryFactory extends ModelFactory implements CountryFactoryInterface
+class CountryFactory extends ModelFactory implements CountryFactoryInterface
 {
     public function __construct(private FactoryInterface $countryFactory)
     {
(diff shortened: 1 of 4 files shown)


src/Sylius/Bundle/CoreBundle/DataFixtures/Factory/CountryFactory.php
```diff
@@ -0,0 +1,88 @@
...
+                $country->setEnabled($attributes['enabled']);
+
+                return $country;
+            })
+        ;
+    }
+
+    protected static function getClass(): string
+    {
+        return Country::class;
```
vvasiloi, line 86: I missed it in the previous PRs, but how does that work in the end-user application where all the resources classes are extended? Normally, parameters like `sylius.model.country.class` would be used.

tests/DataFixtures/Factory/CountryFactoryTest.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $country = CountryFactory::new()->create();
+
+        $this->assertInstanceOf(CountryInterface::class, $country->object());
+        $this->assertNotNull($country->getCode());
+        $this->assertTrue(Countries::exists($country->getCode()));
+    }
+
+    /** @test */
+    function it_creates_country_with_custom_code(): void
```
vvasiloi, line 39:
```suggestion
    function it_creates_country_with_given_code(): void
```
It's not really a custom code.

src/Sylius/Bundle/CoreBundle/DataFixtures/Factory/CountryFactory.php
loic425: Yes I know that, this `getClass` method is required. But the problem is, it is a static method, so we just can not use the "sylius.model.country.class" parameter. But with using the `instantiateWith` callable which uses the entity factory, I think this string does not change anything.

vvasiloi: Alright! We can also think about this later if it's an issue.

loic425: I added it on the main "issue".

Issue #9987, opened 2018-11-30
Multiple channels with php-cli is not possible


Prometee:
**Sylius versions affected**: > 1.2 I suppose

**Description**  
When having two channels and try to use a command witch use a `ChannelContext`, it throw a `ChannelNotFoundException`

**Steps to reproduce**  
Two channels defined and trying to use `ChannelContext` while executing commands.

**Possible Solution**  
Create a `PhpCliChannelContext` (`priority: -256`) that check if we are in a command line env and load a `Channel` witch can configured by service configuration (`calls: [['setDefaultChannelCode', ['MY_DEFAULT_CHANNEL_CODE']]]`) or by a global Sylius `Channel` configuration.
If the `defaultChannelCode` is not provided, we could fallback to the first channel we found.

mamazu:
I would suggest a different approach. Write a `ChannelAwareTrait` that defines a `channel` property and `setChannel` and `getChannel` then you can configure it right in the command when you use the trait.

mikemix:
Yeah this sucks. It's been one year already and it seems impossible to set custom channel context inside CLI command.

kayue:
Any recommended solution?

vvasiloi:
The solution @Prometee suggested should work.
I suggested something similar  to @mickaellosiewicz not long ago and he confirmed it works.

It has the following elements:
1. A `ChannelAwareCommandInterface` and an abstract `ChannelAwareCommand` (or a  `ChannelAwareCommandTrait`) implementing it, with a `--sylius-channel=channel_code` configured.
2. A new channel context with a way to set the current channel and with the highest priority that will return a channel if set.
3. A listener/subscriber that will listen to [ConsoleEvents::COMMAND](https://symfony.com/doc/current/components/console/events.html#the-consoleevents-command-event) event and if the command implements `ChannelAwareCommandInterface` and `sylius-channel` option is set, then use it's value to set  the current channel using the new context.
Not sure if it's necessary to listen to `ConsoleEvents::TERMINATE` event too and remove the channel from the context.
This will allow to easily make any command channel aware.
In addition, the context can be reused to forcefully set the channel in other circumstance.

lchrusciel:
Hey folks! 

I would like to add my 2 cents here. Contexts by design should be considered as stateful services. They reference to session, current request etc. As a result they should be used mostly in HTTP Controllers or as close as possible to them. 

[...]

PR #13545, opened 2022-01-25
[Core][Optimalization] Assign fallback locale for command based requests

SirDomin:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

After research we decided to assign fallback locale for messenger:consume command as it allows CatalogPromotions to be applied way faster than when checking locale for ~2000 product variants

There is also another way to check if we are running from command, with $_SERVER variable we can acccess:
image
but it gets abit tricky when using $_SERVER so decided to stay with this solution as the output can be the same

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services/checkers.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services/checkers.xml
@@ -23,6 +23,11 @@
         <service id="sylius.checker.order_payment_method_selection_requirement" class="Sylius\Component\Core\Checker\OrderPaymentMethodSelectionRequirementChecker" >
             <argument type="service" id="sylius.payment_methods_resolver"/>
         </service>
+
+        <service id="Sylius\Component\Core\Checker\CommandBasedContextCheckerInterface" class="Sylius\Component\Core\Checker\CommandBasedContextChecker">
+            <argument type="string">%env(resolve:APP_ENV)%</argument>
+        </service>
+
         <service id="Sylius\Component\Core\Checker\OrderPaymentMethodSelectionRequirementCheckerInterface" alias="sylius.checker.order_payment_method_selection_requirement" />
     </services>
 </container>
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services/checkers.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services/checkers.xml
@@ -24,7 +24,7 @@
             <argument type="service" id="sylius.payment_methods_resolver"/>
         </service>
 
-        <service id="Sylius\Component\Core\Checker\CommandBasedContextCheckerInterface" class="Sylius\Component\Core\Checker\CommandBasedContextChecker">
+        <service id="Sylius\Component\Core\Checker\CLIContextCheckerInterface" class="Sylius\Component\Core\Checker\CLIContextChecker">
             <argument type="string">%kernel.environment%</argument>
             <argument type="collection">
                 <argument type="string">test</argument>
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services/checkers.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services/checkers.xml
@@ -25,7 +25,11 @@
         </service>
 
         <service id="Sylius\Component\Core\Checker\CommandBasedContextCheckerInterface" class="Sylius\Component\Core\Checker\CommandBasedContextChecker">
-            <argument type="string">%env(resolve:APP_ENV)%</argument>
+            <argument type="string">%kernel.environment%</argument>
+            <argument type="collection">
+                <argument type="string">test</argument>
+                <argument type="string">test_cached</argument>
+            </argument>
         </service>
 
         <service id="Sylius\Component\Core\Checker\OrderPaymentMethodSelectionRequirementCheckerInterface" alias="sylius.checker.order_payment_method_selection_requirement" />
--- a/src/Sylius/Component/Core/Checker/CommandBasedContextChecker.php
+++ b/src/Sylius/Component/Core/Checker/CommandBasedContextChecker.php
@@ -18,11 +18,12 @@ use Sylius\Component\Core\Context\ProcessContextInterface;
 final class CommandBasedContextChecker implements CommandBasedContextCheckerInterface
 {
     public function __construct(
-      private string $runningEnvironment
+        private string $runningEnvironment,
+        private array $restrictedEnvironments
     ) { }
 
     public function isRunningFromCommand(): bool
     {
-        return !str_contains($this->runningEnvironment, 'test') && \php_sapi_name() === ProcessContextInterface::CLI;
+        return !in_array($this->runningEnvironment, $this->restrictedEnvironments) && \php_sapi_name() === ProcessContextInterface::CLI;
     }
 }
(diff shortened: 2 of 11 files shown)


vvasiloi:
@SirDomin WDYT about an implementation similar to what I have described here: https://github.com/Sylius/Sylius/issues/9987#issuecomment-1021723690

SirDomin:
@vvasiloi it could work too, but im not sure about different contexts when using cli from behat/specs. This way make it generic in case when we add more ideas of this type.

src/Sylius/Bundle/CoreBundle/Resources/config/services/checkers.xml
```diff
@@ -23,6 +23,11 @@
         <service id="sylius.checker.order_payment_method_selection_requirement" class="Sylius\Component\Core\Checker\OrderPaymentMethodSelectionRequirementChecker" >
             <argument type="service" id="sylius.payment_methods_resolver"/>
         </service>
+
+        <service id="Sylius\Component\Core\Checker\CommandBasedContextCheckerInterface" class="Sylius\Component\Core\Checker\CommandBasedContextChecker">
+            <argument type="string">%env(resolve:APP_ENV)%</argument>
```
vvasiloi, line 28:
```suggestion
            <argument type="string">%kernel.environment%</argument>
```

src/Sylius/Component/Core/Checker/CommandBasedContextChecker.php
```diff
@@ -0,0 +1,28 @@
...
+
+final class CommandBasedContextChecker implements CommandBasedContextCheckerInterface
+{
+    public function __construct(
+      private string $runningEnvironment
+    ) { }
+
+    public function isRunningFromCommand(): bool
+    {
+        return !str_contains($this->runningEnvironment, 'test') && \php_sapi_name() === ProcessContextInterface::CLI;
```
vvasiloi, line 26: Maybe pass an array of restricted environments as a constructor argument and do an `!in_array()` here.

src/Sylius/Component/Core/Translation/TranslatableEntityLocaleAssigner.php
```diff
@@ -21,14 +22,23 @@
...
 final class TranslatableEntityLocaleAssigner implements TranslatableEntityLocaleAssignerInterface
 {
-    public function __construct(private LocaleContextInterface $localeContext, private TranslationLocaleProviderInterface $translationLocaleProvider)
-    {
-    }
+    public function __construct(
+        private LocaleContextInterface $localeContext,
+        private TranslationLocaleProviderInterface $translationLocaleProvider,
+        private ?CommandBasedContextCheckerInterface $commandBasedChecker = null
+    ) { }
```
vvasiloi, line 29: Add deprecation note if `$commandBasedChecker` is not passed.

src/Sylius/Component/Core/Translation/TranslatableEntityLocaleAssigner.php
```diff
@@ -21,14 +22,23 @@
...
 
     public function assignLocale(TranslatableInterface $translatableEntity): void
     {
         $fallbackLocale = $this->translationLocaleProvider->getDefaultLocaleCode();
 
+        if ($this->commandBasedChecker !== null && $this->commandBasedChecker->isRunningFromCommand()) {
```
vvasiloi, line 35: Seeing this plus the extra constructor argument got me thinking, why not decorate this service instead of modifying it?

SirDomin: I created 3rd not required argument along with deprecation note, i think both cases are good.

src/Sylius/Component/Core/Translation/TranslatableEntityLocaleAssigner.php
```diff
@@ -21,14 +22,27 @@
...
 {
-    public function __construct(private LocaleContextInterface $localeContext, private TranslationLocaleProviderInterface $translationLocaleProvider)
-    {
+    public function __construct(
+        private LocaleContextInterface $localeContext,
+        private TranslationLocaleProviderInterface $translationLocaleProvider,
+        private ?CommandBasedContextCheckerInterface $commandBasedChecker = null
+    ) {
+        if ($this->commandBasedChecker === null) {
+            @trigger_error('Not passing 3rd argument is deprecated. Pass CommandBasedContextCheckedInterface', E_USER_DEPRECATED);
```
GSadee, line 31:
```suggestion
            @trigger_error('Not passing CommandBasedContextCheckedInterface explicitly as the third argument is deprecated since 1.11 and will be prohibited in 2.0.', E_USER_DEPRECATED);
```

src/Sylius/Component/Core/Checker/CommandBasedContextChecker.php
```diff
@@ -0,0 +1,29 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\Checker;
+
+use Sylius\Component\Core\Context\ProcessContextInterface;
+
+final class CommandBasedContextChecker implements CommandBasedContextCheckerInterface
```
lchrusciel, line 18:
```suggestion
final class CLIContextChecker implements CLIContextCheckerInterface
```
I would like to avoid pure `Command` based naming to not make confusion with commands from messenger

src/Sylius/Component/Core/Checker/CommandBasedContextChecker.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    public function __construct(
+        private string $runningEnvironment,
+        private array $restrictedEnvironments
+    ) { }
+
+    public function isRunningFromCommand(): bool
```
lchrusciel, line 25: What do you think about this naming? 
```suggestion
    public function isExecutedFromCLI(): bool
```

src/Sylius/Component/Core/Context/ProcessContextInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Sylius\Component\Core\Context;
+
+interface ProcessContextInterface
+{
+    public const CLI = 'cli';
```
lchrusciel, line 7: Is it worth it to have separate interface for this const? Can't we place it in `CommandBasedContextChecker`? Also, this const could be private

src/Sylius/Component/Core/Translation/TranslatableEntityLocaleAssigner.php
```diff
@@ -21,14 +22,30 @@
...
     {
         $fallbackLocale = $this->translationLocaleProvider->getDefaultLocaleCode();
 
+        if ($this->commandBasedChecker !== null && $this->commandBasedChecker->isRunningFromCommand()) {
+            $translatableEntity->setCurrentLocale($fallbackLocale);
+            $translatableEntity->setFallbackLocale($fallbackLocale);
+
+            return;
+        }
+
```
lchrusciel, line 48:
```suggestion
        $translatableEntity->setFallbackLocale($fallbackLocale);
            
        if ($this->commandBasedChecker !== null && $this->commandBasedChecker->isRunningFromCommand()) {
            $translatableEntity->setCurrentLocale($fallbackLocale);

            return;
        }
```
and remove the line 56

lchrusciel (approved):
I can confirm, that this change makes significant improvement of CP process execution: https://blackfire.io/profiles/compare/39e14b49-45ec-4c52-91e9-9a8e7fbaedcd/graph

GSadee (approved):

GSadee:
Thank you, @SirDomin! :tada:

PR #13547, opened 2022-01-26
[Promotion] Fix js errors associated with actions/rules/scopes

### LINKED ISSUE #13447: [Promotion] When added a new rule to promotion in console is an js error
**Sylius version affected**: v1.12.0-DEV

**Description**  
On promotions add a new rule there is an javascript error logged in the console. The rule Cart Quantity is not showing the part with Quantity etc.

In console I see:
Uncaught TypeError: e is null

**Steps to reproduce issue**

1. Go to Sylius demo https://master.demo.sylius.com/admin/
2. Select Next release feature (v1.12.0-DEV)
3. Go to Cart promotions - create or update existing one
4. Add rule
5. Check console log

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #13447
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
+++ b/src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
@@ -68,17 +68,7 @@ $(document).ready(() => {
   });
   $('#sylius_promotion_rules > a[data-form-collection="add"]').on('click', () => {
     setTimeout(() => {
-      $(`select[name^="sylius_promotion[rules]"][name$="[type]"]`).last().change();
-    }, 50);
-  });
-  $('#sylius_catalog_promotion_actions > a[data-form-collection="add"]').on('click', () => {
-    setTimeout(() => {
-      $('select[name^="sylius_catalog_promotion[actions]"][name$="[type]"]').last().change();
-    }, 50);
-  });
-  $('#sylius_catalog_promotion_scopes > a[data-form-collection="add"]').on('click', () => {
-    setTimeout(() => {
-      $(`select[name^="sylius_catalog_promotion[scopes]"][name$="[type]"]`).last().change();
+      $('select[name^="sylius_promotion[rules]"][name$="[type]"]').last().change();
     }, 50);
   });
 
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/AdminBundle/Resources/private/js/app.js
```diff
@@ -68,17 +68,7 @@ $(document).ready(() => {
...
-  });
-  $('#sylius_catalog_promotion_actions > a[data-form-collection="add"]').on('click', () => {
-    setTimeout(() => {
-      $('select[name^="sylius_catalog_promotion[actions]"][name$="[type]"]').last().change();
-    }, 50);
-  });
-  $('#sylius_catalog_promotion_scopes > a[data-form-collection="add"]').on('click', () => {
-    setTimeout(() => {
-      $(`select[name^="sylius_catalog_promotion[scopes]"][name$="[type]"]`).last().change();
+      $('select[name^="sylius_promotion[rules]"][name$="[type]"]').last().change();
```
rimas-kudelis, line 71: This change from three different cases to one certainly doesn't look right.
Also, please use the direct ancestor selector (`>`), because this will allow adding collections in child forms without having to hack JS.

GSadee: Here, I've only removed the lines that do nothing in my opinion, but of course, I might have missed sth. But the tests are green and I've check both cart and catalog promotions manually also

rimas-kudelis: Don't catalog promotions use same JS?
And yeah, disregard that second part of my previous comment. It's irrelevant.

vvasiloi: Those lines allow to use nested collections inside the promotion rules and actions.
Unfortunately there's no test for that, because there's no such use case in Sylius by default.

rimas-kudelis: @vvasiloi no, it's just the `>` selector that allows nested collections. It doesn't mean that these lines were actually used.

And remember, we added behat scenarios for both catalog promotions and cart promotions. If all of them still succeed even without these lines, that means that either the tests are bad, or the lines aren't really used after all.

vvasiloi: Yes, I was referring to the `>` in the selector.
We tested that the form is displayed when a new action or rule is added and if it works with a single selector then it's even better.

lchrusciel (approved):
If I understood correctly, we are fine to have it merge. Thanks for your activity @rimas-kudelis and @vvasiloi ?

lchrusciel:

PR #13555, opened 2022-01-27
[Core][Optimalization] Assign fallback locale for command based requests

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

I'm proposing alternative implementation to #13545, that takes into account request stack instead of env & SAPI. This way, we will have coherent behaviour among all envs which, in my opinion, will reduce possibility of bugs that exists on prod on dev, but where not detected by test.

Perhaps, we should consider similar behaviour just as a locale context - so fallbacking in the locale context to config value if CLI is executed

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/services/checkers.xml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/services/checkers.xml
@@ -25,11 +25,7 @@
         </service>
 
         <service id="Sylius\Component\Core\Checker\CLIContextCheckerInterface" class="Sylius\Component\Core\Checker\CLIContextChecker">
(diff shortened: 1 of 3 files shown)


lchrusciel:
The performance change may be neglected between these two implementations: https://blackfire.io/profiles/compare/8fe620d6-4cc9-456b-90a3-861ee485935d/graph

SirDomin (approved):

GSadee (approved):

GSadee:

PR #13557, opened 2022-01-28
[Promotion] Use sylius/calendar instead of the existing Calendar

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | replaces https://github.com/Sylius/Sylius/pull/13533
| License         | MIT

In order to unify the usage of two same services from Promotion and Shipping bundles, we've decided to extract the Calendar functionality to a separate [repository](https://github.com/Sylius/Calendar).

PR #13558, opened 2022-01-28
[Shipping][Behat] Use sylius/calendar instead of the existing Calendar

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | yes
| Related tickets | continuation of #13557 
| License         | MIT


src/Sylius/Behat/Resources/config/suites/api/order/managing_orders.yml
```diff
@@ -5,9 +5,9 @@ default:
     suites:
         api_managing_orders:
             contexts:
-                - sylius.behat.context.hook.calendar
```
lchrusciel, line 8: This could be mentioned in UPGRADE.md as well

src/Sylius/Bundle/ShippingBundle/Assigner/ShippingDateAssigner.php
```diff
@@ -14,19 +14,36 @@
...
 use Sylius\Bundle\ShippingBundle\Provider\DateTimeProvider;
+use Sylius\Calendar\Provider\DateTimeProviderInterface;
 use Sylius\Component\Shipping\Model\ShipmentInterface;
 
 final class ShippingDateAssigner implements ShippingDateAssignerInterface
 {
-    private DateTimeProvider $calendar;
-
-    public function __construct(DateTimeProvider $calendar)
+    public function __construct(private ?DateTimeProvider $calendar, private ?DateTimeProviderInterface $dateTimeProvider = null)
```
lchrusciel, line 22: As we want to replace this argument at the end, I would suggest removal of types to allow both of services be injected into constructor
```suggestion
    public function __construct(private $calendar)
```
or use PHP8.0 union type feature 🎉 
```suggestion
    public function __construct(private DateTimeProvider|DateTimeProviderInterface $calendar)
```

And trigger error if `DateTimeProvider` is provided. If so, it will be assigned to the same field, so the method selection logic needs to be moved to method call. Would be perfect, to do it in private method, as it is required in two places

lchrusciel (approved):

lchrusciel:

PR #13576, opened 2022-02-01
[Catalog Promotion] Select more data during processing to avoid inefficient db calls

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

With the proposed changes, we are omitting the n+1 problem. The idea is to load all data required for CP processing in one batch, so we can drastically reduce the number of DB queries(in the cost of computation memory, but it can be neglected, as still the biggest problem of it is that we do not clear UnitOfWork)

I've also adjusted alias of query builder to `o`, as this is Sylius default 

Ref. https://blackfire.io/profiles/compare/3d7cc083-575f-4e58-acba-e2f60f444cbc/graph

### DIFF
--- a/src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
+++ b/src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
@@ -101,6 +101,16 @@ class ProductVariantRepository extends EntityRepository implements ProductVarian
     public function findByCodes(array $codes): array
     {
         return $this->createQueryBuilder('o')
+            ->addSelect('product')
+            ->addSelect('channelPricings')
+            ->addSelect('appliedPromotions')
+            ->addSelect('productTaxon')
+            ->addSelect('taxon')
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
```diff
@@ -98,9 +98,19 @@ public function findByCodesAndProductCode(array $codes, string $productCode): ar
         ;
     }
 
-    public function findByCodes(array $codes): array
+    public function findByCodesWithProductCannelPricingsAndTaxons(array $codes): array
```
GSadee, line 101:
```suggestion
    public function findByCodesWithProductChannelPricingsAndTaxons(array $codes): array
```
or even leave the old name 🤔 :
```suggestion
    public function findByCodes(array $codes): array
```

GSadee (approved):

GSadee:

PR #13578, opened 2022-02-01
[Catalog Promotion] add update variant event to messenger routing

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         |1.11
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| License         | MIT

PR #13579, opened 2022-02-01
[Catalog Promotion] Move catalog promotion processing after the fixture execution

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Previous:
~With clearing entity manger we don't need to trace that many objects in UnitOfWork, therefore we are not calculating that much changes and we save ton of memory.~

~Ref. https://blackfire.io/profiles/compare/7c9cb9fd-c43f-4420-8e0a-0af2a7b4b000/graph~

Now:
Just a refactor of CP fixtures. The EntityManager will be cleared with https://github.com/symfony/doctrine-bridge/blob/5.4/Messenger/DoctrineClearEntityManagerWorkerSubscriber.php just by the fact, that it will be processed asynced. Which will be done in #13578. 

#13578 vs previous implementation -https://blackfire.io/profiles/compare/f560f723-f970-41b3-ac1a-7fe3a8858d77/graph

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Factory/CatalogPromotionExampleFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Factory/CatalogPromotionExampleFactory.php
@@ -40,8 +40,6 @@ class CatalogPromotionExampleFactory extends AbstractExampleFactory implements E
 
     private ExampleFactoryInterface $catalogPromotionActionExampleFactory;
 
-    private AllProductVariantsCatalogPromotionsProcessorInterface $allProductVariantsCatalogPromotionsProcessor;
-
     private Generator $faker;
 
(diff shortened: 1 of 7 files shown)


GSadee (approved):

stloyd:
This looks dangerous for me cause cleaning whole entity manager means that you remove **everything** that was put in it, either it was already saved or not into database, which means side effects in live applications can be unpredictable IMHO.

vvasiloi:
I agree with @stloyd. Usually it's flush then clear, doing just a clear is not a good idea.

GSadee (approved):

GSadee:
Thanks, Łukasz! :tada:

PR #13583, opened 2022-02-02
[Catalog Promotion] Add CP to database before processing in example factory

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Factory/CatalogPromotionExampleFactory.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Factory/CatalogPromotionExampleFactory.php
@@ -23,6 +23,7 @@ use Sylius\Component\Core\Model\CatalogPromotionInterface;
 use Sylius\Component\Locale\Model\LocaleInterface;
 use Sylius\Component\Promotion\Model\CatalogPromotionActionInterface;
 use Sylius\Component\Promotion\Model\CatalogPromotionScopeInterface;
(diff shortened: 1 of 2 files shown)


GSadee (commented):
This PR: https://github.com/Sylius/Sylius/pull/13579 could probably also resolve the same problem 🤔

src/Sylius/Bundle/CoreBundle/Fixture/Factory/CatalogPromotionExampleFactory.php
```diff
@@ -110,6 +94,7 @@ public function create(array $options = []): CatalogPromotionInterface
             }
         }
 
+        $this->catalogPromotionRepository->add($catalogPromotion);
```
Zales0123, line 97: We definitely should not persist the entity in the **factory** class which should only be responsible for creation. Especially, as fixture already persist it after creation :dancer: Perhaps we should move `$this->allCatalogPromotionsProcessor->process();` to listener implementing `AfterFixtureListenerInterface`

Zales0123: And, btw, it's already done [here](https://github.com/Sylius/Sylius/pull/13579/files#diff-6120934584af520c17ad611a2af6f0743108f3db475350768ce5b0f84ab03ae2R23) :smile:

PR #13584, opened 2022-02-02
Update catalog promotion instantly if date is set on future

SirDomin:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no/
| Related tickets | 
| License         | MIT

If we had CatalogPromotion already applied on channelPricings, after changing its startDate to next day, previous discounts would still be applied, with this event we force to recalculate catalog promotions.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Announcer/CatalogPromotionAnnouncer.php
+++ b/src/Sylius/Bundle/CoreBundle/Announcer/CatalogPromotionAnnouncer.php
@@ -42,6 +42,10 @@ final class CatalogPromotionAnnouncer implements CatalogPromotionAnnouncerInterf
 
     public function dispatchCatalogPromotionUpdatedEvent(CatalogPromotionInterface $catalogPromotion): void
     {
+        if ($catalogPromotion->getStartDate() > $this->dateTimeProvider->now()) {
+            $this->eventBus->dispatch(new CatalogPromotionUpdated($catalogPromotion->getCode()), []);
+        }
+
         $this->eventBus->dispatch(
             new CatalogPromotionUpdated($catalogPromotion->getCode()),
             $this->calculateStartDateStamp($catalogPromotion)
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/CoreBundle/Announcer/CatalogPromotionAnnouncer.php
```diff
@@ -42,6 +42,10 @@ public function dispatchCatalogPromotionCreatedEvent(CatalogPromotionInterface $
 
     public function dispatchCatalogPromotionUpdatedEvent(CatalogPromotionInterface $catalogPromotion): void
     {
+        if ($catalogPromotion->getStartDate() > $this->dateTimeProvider->now()) {
+            $this->eventBus->dispatch(new CatalogPromotionUpdated($catalogPromotion->getCode()), []);
```
Rafikooo, line 46:
```suggestion
            $this->eventBus->dispatch(new CatalogPromotionUpdated($catalogPromotion->getCode()));
```
Due to the fact that the dispatch's method second param has a default value we haven't to pass an empty array

SirDomin: I left it intentionally like this to keep it consistent with every other dispatching where we send extra argument even if it should be dispatched right away.

src/Sylius/Bundle/CoreBundle/spec/Announcer/CatalogPromotionAnnouncerSpec.php
```diff
@@ -127,4 +127,35 @@ function it_dispatches_catalog_promotion_updated_and_catalog_promotion_ended_eve
...
+        $startDelayStamp = new DelayStamp(200000);
+        $endDelayStamp = new DelayStamp(300000);
+
+        $delayStampCalculator->calculate(Argument::any(), $startDateTime)->willReturn($startDelayStamp);
+        $delayStampCalculator->calculate(Argument::any(), $endDateTime)->willReturn($endDelayStamp);
+
+        $messageUpdate = new CatalogPromotionUpdated('SALE');
+        $messageEnd = new CatalogPromotionEnded('SALE');
+
+        $eventBus->dispatch($messageUpdate, [])->willReturn(new Envelope($messageUpdate))->shouldBeCalled();
```
Rafikooo, line 155: Same as above

AdamKasp (approved):

lchrusciel (approved):

lchrusciel:
Thanks, @SirDomin! :tada:

PR #13588, opened 2022-02-02
Use semantic ui view instead of deprecated semantic_ui_translated

### LINKED ISSUE #12462: [UiBundle] Pagination macro uses deprecated PagerfantaBundle view
With the introduction of Twig support into Pagerfanta and PagerfantaBundle, I deprecated all of the PHP templates from the bundle that essentially were decorators around the base classes and used Symfony's translator to override the messages in favor of Twig templates.  The pagination macro in the UiBundle is still using the deprecated PHP view class and should be updated to remove the use of this deprecated API (note, this is a blocker to allowing Pagerfanta and PagerfantaBundle 3.0 in the main Sylius app).

From an end-user perspective, it becomes a little easier to customize pagination templates with Twig as they have full control of the various blocks that are rendered and it's easier to introduce new Twig templates, whereas with the PHP templates the options really only give control over some CSS classes and deeper markup adjustments would require new PHP classes and a compiler pass registering them to the view factory.

```twig
{{ pagerfanta(paginator, 'twig', options|default({'template': '@BabDevPagerfanta/semantic_ui.html.twig'})) }}
```

I *think* that should be the right change needed in the macro, but there are minor markup differences between the PHP template and the Twig template so it may be slightly more involved to make the change.  Personally, I'd also recommend updating the Sylius-Standard with the below config snippet and changing the macro call to `{{ pagerfanta(paginator, options) }}` as at that point you can control the default pagination view and Twig template through the Symfony configuration instead of needing to overload a macro which uses a hardcoded view name:

```yaml
babdev_pagerfanta:
    default_view: 'twig'
    default_twig_template: '@BabDevPagerfanta/semantic_ui.html.twig'
```

loic425:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #12462
| License         | MIT

### DIFF
--- a/UPGRADE-1.11.md
+++ b/UPGRADE-1.11.md
@@ -1,5 +1,23 @@
 # UPGRADE FROM `v1.10.X` TO `v1.11.0`
 
+### "pagerfanta/pagerfanta" semantic_ui_translated removed
+
+The `pagination.html.twig` has been changed to use the Twig view.
+
+There are differences in the markup between the PHP template and the Twig template.
+
+The wrapping container from 2.x branch of Pagerfanta in the PHP template was
+```html
+<div class="ui stackable fluid pagination menu">
+```
+
+while in the Twig template in 3.x branch it's 
+```html
+<div class="ui pagination menu">
+```
+
+The "stackable" class affects responsive display and "fluid" affects whether the pagination menu is full-width). 
+
 ### "polishsymfonycommunity/symfony-mocker-container" moved to dev-requirements
 
 This obvious dev dependency was part of Sylius requirements. In 1.11 we've moved it to proper place. However, it may lead to app break, as this container could be used in your Kernel, if you used Sylius-Standard as your template. In such a case, please update your `src/Kernel.php` class as follows:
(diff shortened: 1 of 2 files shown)


Zales0123:
cc @kulczy but seems good for me 😄

mbabker:
You need to use the “twig” view with the `@BabDevPagerfanta/semantic_ui.html.twig` template to keep translation support.  The “semantic_ui” view is PHP generated without translation support.

This’ll also close https://github.com/Sylius/Sylius/issues/12462 😉

loic425:
@mbabker Thx a lot, I will follow your issue details to update this PR.

loic425:
@mbabker I updated this PR, is that ok for you?

For the package configuration, I think it's needed into the sylius/sylius-standard repository.

mbabker (approved):

lchrusciel (approved):

lchrusciel:
Thank you, Loïc! :tada:

PR #13600, opened 2022-02-04
Avoid passing empty needle to strpos()

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | Extraction of https://github.com/Sylius/Sylius/pull/13431
| License         | MIT

It may happen that `propertyPath` is empty if the object is root object which may lead to php warning being triggered. It is better to omit such situation


rimas-kudelis:
@lchrusciel I guess this PR is now irrelevant since you merged in #13431.

PR #13607, opened 2022-02-07
[Catalog Promotion][UI] Fixed error when trying to save catalog promotion with empty percentage discount value

ernestWarwas:
| Q               | A
| --------------- | -----
| Branch?         | 1.11 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no

When trying to edit the percentage discount value and save it with an empty amount there was a 500 error. This PR provides only a workaround for this problem.

image

image

### DIFF
--- a/src/Sylius/Bundle/PromotionBundle/Form/Type/CatalogPromotionActionType.php
+++ b/src/Sylius/Bundle/PromotionBundle/Form/Type/CatalogPromotionActionType.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Sylius\Bundle\PromotionBundle\Form\Type;
 
 use Sylius\Bundle\CoreBundle\CatalogPromotion\Calculator\FixedDiscountPriceCalculator;
+use Sylius\Bundle\CoreBundle\CatalogPromotion\Calculator\PercentageDiscountPriceCalculator;
 use Sylius\Bundle\ResourceBundle\Form\Type\AbstractResourceType;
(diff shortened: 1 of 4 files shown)


src/Sylius/Bundle/PromotionBundle/Form/Type/CatalogPromotionActionType.php
```diff
@@ -72,6 +73,12 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                         }
                     }
 
+                    if ($data['type'] === PercentageDiscountPriceCalculator::TYPE) {
+                        if ($data['configuration']['amount'] === '') {
+                            return;
+                        }
+                    }
+
```
AdamKasp, line 81: do we really want to hardcode these values (line 76 and 68), it looks like a workaround :(
what will happen with a new custom calculator?
I don't know any idea how to improve it :D but maybe @GSadee @Zales0123 or @lchrusciel  have some hint :)

lchrusciel: It should be cover with the regular validation and proper validation message

Zales0123 (approved):
This is definitely a workaround/hotfix and we need a deeper look into the problem... but it at least solves an issue and still can be fixed after the stable release 🖖

Zales0123:
Thanks, @ernestWarwas! :tada:

PR #13611, opened 2022-02-07
Optimize variant autocomplete selector

rafalswierczek:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT


src/Sylius/Bundle/CoreBundle/Doctrine/ORM/QueryBuilder/QueryBuilderPagination.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\Doctrine\ORM\QueryBuilder;
+
+use Doctrine\ORM\QueryBuilder;
+
+class QueryBuilderPagination
```
Zales0123, line 9: I do not think this class extraction is needed right now - it's used only in one place, so it seems to do more harm than good 🖖

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
```diff
@@ -152,19 +171,18 @@ public function findByPhraseAndProductCode(string $phrase, string $locale, strin
 
     public function findByPhrase(string $phrase, string $locale): array
```
Zales0123, line 172: What if we just add optional arguments here rather than creating a separate repository method? Especially as now the `findByPhrase` is not used at all

src/Sylius/Component/Product/Model/ProductVariant.php
```diff
@@ -36,6 +36,8 @@ class ProductVariant implements ProductVariantInterface
      */
     protected $code;
 
+    protected int $productId;
```
Zales0123, line 39: I believe this additional field is not needed - we can join the product table in the query build if needed

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
```diff
@@ -44,6 +45,24 @@ public function createQueryBuilderByProductCode(string $locale, string $productC
...
+        return $this->createQueryBuilder('o')
+            ->innerJoin('o.translations', 'translation', 'WITH', 'translation.locale = :locale')
+            ->andWhere($expr->orX(
+                'translation.name LIKE :phrase',
+                'o.code LIKE :phrase'
+            ))
+            ->setParameter('phrase', '%' . $phrase . '%')
+            ->setParameter('locale', $locale)
+            ->orderBy('o.productId', 'ASC')
+            ->addOrderBy('o.position', 'ASC')
```
Zales0123, line 61: This is the change in the behaviour - is it needed? Moreover, if we want to sort the results it could be done by `position` and `id` rather than `productId`

tests/Controller/AdminProductVariantAjaxTest.php
```diff
@@ -0,0 +1,88 @@
...
+     */
+    public function it_returns_only_specified_part_of_all_product_variants()
+    {
+        $this->authenticateAdminUser();
+
+        $this->client->request('GET', '/admin/ajax/product-variants/search-all');
+
+        $response = $this->client->getResponse();
+
+        $expectedData = '[
```
Zales0123, line 35: We should use an ApiTestCase feature here and extract expected data to file like `ajax/product_variant/index_response`

tests/Controller/AdminProductVariantAjaxTest.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * @test
+     */
+    public function it_returns_only_specified_part_of_all_product_variants()
+    {
+        $this->authenticateAdminUser();
+
+        $this->client->request('GET', '/admin/ajax/product-variants/search-all');
```
Zales0123, line 31: What about other behaviours? Those with added functionalities (max result and page parameters?)

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
```diff
@@ -33,10 +33,12 @@ sylius_admin_ajax_all_product_variants_by_phrase:
             serialization_groups: [Autocomplete]
             permission: true
             repository:
-                method: findByPhrase
+                method: findByPhrasePartial
                 arguments:
                     phrase: $phrase
                     locale: expr:service('sylius.context.locale').getLocaleCode()
+                    page: 1
```
AdamKasp, line 40: Can we simply add optional argument `limit`? and make it in a similar way like in `Sylius\Bundle\ProductBundle\Doctrine\ORM\ProductRepository::findByNamePart`

you are adding pagination, but it isn't mentioned in task at all.

GSadee (commented):
This PR should be opened against 1.11 branch

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
rafalswierczek: No. `productId` is needed to control data order regardless of it's position is table. I checked it to be sure.

src/Sylius/Component/Product/Model/ProductVariant.php
rafalswierczek: We really don't want any `JOIN` in this functionality. It's query overhead. Sometimes it's needed to have only id rather than a reference. I believe that accessing product id by relation will create inner/left join by Doctrine.

Zales0123: However we don't do a database-driven development, we're designing the application and code to be nice and shiny and let ORM do its job 😄  This `productId` here is used **only** to simplify the query which cannot even overload the database too much as we're limiting the number of data taken from the database by default. If it would be a problem for anyone - one would just override the query and customize the entity in their application :dancer:

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
Zales0123: Agreed 👍  we can just make a simple limit without pagination and it would already be an improvement 🚀

src/Sylius/Component/Product/Model/ProductVariant.php
vvasiloi: Product variant is the owner of this relationship, so there's already a `product_id` field the database table `sylius_product_variant.`. You can do `$productVariant->getProduct()->getId()` to get the value and this won't perform any additional queries.

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
vvasiloi: Is anyone going beyond the first page of Google Search? 🤷‍♂️ 
If it's on on the first one, then your query is wrong and you change it. 😆

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
vvasiloi: It makes sense to me. You'll see variants of the same product together ordered the same way as everywhere in the admin or shop - by position.
However, this new column is redundant, since there's already a `product_id` column in that table.
You can replace it with `->orderBy('o.product', 'ASC')` and Doctrine will translate it to `ORDER BY o.product_id ASC`.

src/Sylius/Component/Product/Model/ProductVariant.php
rafalswierczek: Yes but Doctrine will generate JOIN as soon as it detects calling getter method in relation object.
`$productVariant->getProduct()->getId()` is ok to satisfy "we're designing the application and code to be nice and shiny".
Anyway, it's not a performance problem but rather a good practice. Calling getters in relation objects will always lead to very bad designed application in case of performance and there is almost no escape from it. I guess it's the problem of people using Sylius - not ours.

vvasiloi: You say it's not a performance problem, but then the very next sentence you say the opposite, which is confusing.
In this particular case, we're not calling any getters except `getId()`,a getter for the identifier, which is treated differently in Doctrine proxies. It will not generate any joins or other extra queries.

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
```diff
@@ -150,7 +150,7 @@ public function findByPhraseAndProductCode(string $phrase, string $locale, strin
         ;
     }
 
-    public function findByPhrase(string $phrase, string $locale): array
+    public function findByPhrase(string $phrase, string $locale, $limit): array
```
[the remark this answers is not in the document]
Zales0123: Agreed - just because we don't **have to** keep BC in repositories doesn't mean we **cannot** 🖖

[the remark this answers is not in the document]
Zales0123: Also, the type of `$limit` is missing

tests/Controller/AdminProductVariantAjaxTest.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+namespace Sylius\Tests\Controller;
+
+use ApiTestCase\JsonApiTestCase;
+use Symfony\Component\BrowserKit\Cookie;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
+
+class AdminProductVariantAjaxTest extends JsonApiTestCase
```
Zales0123, line 10: Missing `final`

tests/Controller/AdminProductVariantAjaxTest.php
```diff
@@ -0,0 +1,66 @@
...
+{
+    /**
+     * @test
+     */
```
Zales0123, line 14: Could be one-liner `/** @test */`

tests/Controller/AdminProductVariantAjaxTest.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+namespace Sylius\Tests\Controller;
```
Zales0123, line 3: Missing `declare(strict_types=1);` and a license block

tests/Controller/AdminProductVariantAjaxTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * @test
+     */
+    public function it_returns_only_specified_part_of_all_product_variants()
+    {
+        $this->loadFixturesFromFile('authentication/administrator.yml');
+        $this->loadFixturesFromFiles([
+            'resources/product_variants.yml',
+
```
Zales0123, line 32: Extra blank line

src/Sylius/Component/Product/Model/ProductVariant.php
rafalswierczek: Good then I didn't know it.

tests/Controller/AdminProductVariantAjaxTest.php
Zales0123: *And* a license block 😄

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
```diff
@@ -150,21 +150,28 @@ public function findByPhraseAndProductCode(string $phrase, string $locale, strin
...
+            ->orderBy('o.product', 'ASC')
+            ->addOrderBy('o.position', 'ASC')
+            ->setMaxResults($limit)
         ;
+
+        if (null !== $limit) {
+            $queryBuilder->setMaxResults($limit);
+        }
+
+        return $queryBuilder->getQuery()->getResult();
```
AdamKasp, line 174:
```suggestion
       ->setMaxResults($limit)
       ->getQuery()
       ->getResoult()
   ;
```

AdamKasp: should be enough

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
```diff
@@ -37,6 +37,7 @@ sylius_admin_ajax_all_product_variants_by_phrase:
                 arguments:
                     phrase: $phrase
                     locale: expr:service('sylius.context.locale').getLocaleCode()
+                    limit: 5
```
AdamKasp, line 40: Why 5? we didn't decide how big this limit should be but in other places we have a limit of 25. The shop can offer many variants so IMO 5 is a bit too small value :)

vvasiloi: It's indeed small, 5 often won't even cover the variants of a single product.
Besides that, maybe it's better define a parameter for it, so it's easier to customize.

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
vvasiloi: Careful, there's a typo in the suggestion: `getResoult` -> `getResult`

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
rafalswierczek: I wanted to make it dynamic but for now only 1 variable is supported in this configuration with js included. IIRC you can make it work by manually passing parameter to URL but doing so programatically is a task beyond me.
I'll make it 25 then.

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
rafalswierczek: I'll check how it works with null ¯\_(ツ)_/¯
It's always a matter of experience with these untyped Doctrine functions...

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
vvasiloi: I meant to use a parameter for the default parameter, like this `limit: '%sylius_admin.some_parameter%'`.
But what you suggest is even better and it's actually not that complicated to do:
Try this:
```suggestion
                    limit: $limit
        limit: 25
```
And to test it's overridable:
```twig
{{ path('sylius_admin_ajax_all_product_variants_by_codes', {'limit': 10}) }}
```

rafalswierczek: Moreover, I made this function's `public function findByPhrase(string $phrase, string $locale, ?int $limit): array` `$limit` parameter untyped when I created pagination because whatever comes to these parameters are strings :/
Now if I make `limit: $limit` then passing it to URL will cause a type error.

vvasiloi: No worries, there's a feature for that in Sylius: `limit: '!!int $limit'`

vvasiloi: It's not a documented feature though... 🤷‍♂️ 
/cc @loic425 @Roshyo maybe you would be interested in this one.

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
vvasiloi: It's safe. The method accepts `null` which will not set a limit.

tests/Controller/AdminProductVariantAjaxTest.php
```diff
@@ -0,0 +1,70 @@
...
+use ApiTestCase\JsonApiTestCase;
+use Symfony\Component\BrowserKit\Cookie;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
+
+final class AdminProductVariantAjaxTest extends JsonApiTestCase
+{
+    /** @test */
+    public function it_denies_access_to_product_variants_for_not_authenticated_user()
```
AdamKasp, line 24:
```suggestion
    public function it_denies_access_to_product_variants_for_not_authenticated_user(): void
```
in the following too

AdamKasp (approved):

AdamKasp (approved):

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
vvasiloi: So no dynamic parameter after all? 😞

rafalswierczek: No because it would be wiser to keep it hardcoded as everywhere so far and if it comes the time we could make all limits accessible in twig. Dynamic limit seems more likely to be used in pagination but I understand the need of having just limit dynamic. Since my pagination went to trash I decided that dynamic limit would be hardcoded because it would be "used only in one place, so it seems to do more harm than good".

tests/Controller/AdminProductVariantAjaxTest.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /** @test */
+    public function it_returns_only_specified_part_of_all_product_variants(): void
+    {
+        $this->loadFixturesFromFile('authentication/administrator.yml');
+        $this->loadFixturesFromFiles(['resources/product_variants.yml']);
+
+        $this->authenticateAdminUser();
+
+        $this->client->request('GET', '/admin/ajax/product-variants/search-all?phrase=');
```
vvasiloi, line 41: You didn't provide a phrase, is that intentional?

rafalswierczek: Yes because real scenario is where phrase is parsed to empty string - I don't know why.

vvasiloi: I don't follow. What do you mean by "real scenario"?
I think there should be a test for both cases - with and without a phrase.

rafalswierczek: In this case when you enter the edit page of specific catalog product and add a scope with type `for variants` and click on empty input - this is the moment when phrase become empty in SQL query even if its not empty as first.
I'll try to investigate that.
Tests are surely needed for it (maybe there are some already).

vvasiloi: Now I understand. In this case we're only missing the case when there's a search phrase.
I know that the original feature that did a search by phrase in a product context has a [Behat test](https://github.com/Sylius/Sylius/blob/master/features/product/managing_product_variants/autocomplete_product_variant_descriptor.feature). And there's a similar one for taxons.
Is an end-to-end with test with Behat necessary here or is it enough to have another integration test with PHPUnit?
@GSadee @Zales0123 @AdamKasp WDYT?

SirDomin (approved):

tests/Controller/AdminProductVariantAjaxTest.php
AdamKasp: IMHO another case in PHPUnit is a minimum, we have a case without specified `phrase` so the case with `phrase` also should be tested. And like You already mentioned we are testing autocompleting in different places so in Catalog Promotion also it should be tested. What is more, we even can make behat which will testing `limit` feature :)

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
AdamKasp: @rafalswierczek our mission is making Sylius as easy to customize as possible (especially if it needs only a few lines of code 👯‍♀️  , so I suggest bringing up this topic to our daily and make a small additional discussion on audio :)

tests/Controller/AdminProductVariantAjaxTest.php
vvasiloi: Limit is hardcoded and can be tested with PHPUnit too, but we would need more than 25 variants in fixtures.
To keep things simple I would go for another test with PHPUnit for now.
The other 2 tests I mentioned are API only, so there would no difference to the extent both tests cover the feature.
The main difference is in the format and I prefer PHPUnit.

I'm also wondering wherever AJAX endpoints like this one should be ported to the new api.

Zales0123: 👍 for another test in PHPUnit, Behat is a pain for auto-complete as they need to be `@javascript` scenarios 😄 Covering it nicely with PHPUnit should be enough 👍

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
rafalswierczek: It will be dynamic

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
```diff
@@ -37,6 +37,7 @@ sylius_admin_ajax_all_product_variants_by_phrase:
                 arguments:
                     phrase: $phrase
                     locale: expr:service('sylius.context.locale').getLocaleCode()
+                    limit: '!!int $limit'
```
vvasiloi, line 40: It should be optional and have a default value, no need to require the user to pass it even if they're happy with the default value.
```suggestion
                    limit: '!!int $limit'
        limit: 25
```

src/Sylius/Bundle/AdminBundle/Resources/views/CatalogPromotion/Scope/for_variants.html.twig
```diff
@@ -1,5 +1,5 @@
 {% form_theme field '@SyliusAdmin/Form/theme.html.twig' %}
 
-{{ form_row(field.variants, {'remote_url': path('sylius_admin_ajax_all_product_variants_by_phrase'), 'remote_criteria_type': 'contains', 'remote_criteria_name': 'phrase', 'load_edit_url': path('sylius_admin_ajax_all_product_variants_by_codes')}) }}
+{{ form_row(field.variants, {'remote_url': path('sylius_admin_ajax_all_product_variants_by_phrase', {limit: 25}), 'remote_criteria_type': 'contains', 'remote_criteria_name': 'phrase', 'load_edit_url': path('sylius_admin_ajax_all_product_variants_by_codes')}) }}
```
vvasiloi, line 3: Define a default limit value in the route configuration and remove it from here.
```suggestion
{{ form_row(field.variants, {'remote_url': path('sylius_admin_ajax_all_product_variants_by_phrase'), 'remote_criteria_type': 'contains', 'remote_criteria_name': 'phrase', 'load_edit_url': path('sylius_admin_ajax_all_product_variants_by_codes')}) }}
```

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
```diff
@@ -162,6 +162,9 @@ public function findByPhrase(string $phrase, string $locale): array
             ))
             ->setParameter('phrase', '%' . $phrase . '%')
             ->setParameter('locale', $locale)
+            ->orderBy('o.product', 'ASC')
+            ->addOrderBy('o.position', 'ASC')
+            ->setMaxResults($limit === 0 ? null : $limit)
```
vvasiloi, line 167: Is this a side effect of the type casting `'!!int $limit'`?

tests/Controller/AdminProductVariantAjaxTest.php
```diff
@@ -0,0 +1,117 @@
...
+
+    /** @test */
+    public function it_throws_type_error_when_phrase_is_not_specified(): void
+    {
+        $this->loadFixturesFromFile('authentication/administrator.yml');
+        $this->loadFixturesFromFiles(['resources/product_variants.yml']);
+
+        $this->authenticateAdminUser();
+
+        $this->expectException(\TypeError::class);
```
vvasiloi, line 56: I think this is better handled on the route configuration level, so it throws an exception instead of a TypeError.
The TypeError can come from somewhere else.
Something like this:
```yaml
    requirements:
        phrase: '.+'
```

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
rafalswierczek: Did you test it?
Inserting `limit: 25` below `defaults` denies possibility to pass `limit=123` in URL parameter (limit is always 25).

vvasiloi: 🤔 what...

rafalswierczek: Removing `, {limit: 25}` from `for_variants.html.twig` for this routing configuration:
```
defaults:
        _controller: sylius.controller.product_variant:indexAction
        _format: json
        _sylius:
            serialization_groups: [Autocomplete]
            permission: true
            repository:
                method: findByPhrase
                arguments:
                    phrase: $phrase
                    locale: expr:service('sylius.context.locale').getLocaleCode()
                    limit: '!!int $limit'
        limit: 2
```
and entering `/admin/ajax/product-variants/search-all?phrase=&limit=4` will return 2 variants in response.

rafalswierczek: You don't need to pass `limit` at all. That's why I added `0` check in repository method because empty limit becomes `0` instead of `null` which is imho not good but it's beyond this pr.

src/Sylius/Bundle/AdminBundle/Resources/views/CatalogPromotion/Scope/for_variants.html.twig
rafalswierczek: I realised you cannot add default value in route config. It's static or dynamic but cannot be default there.
Correct me if I'm wrong.

src/Sylius/Bundle/ProductBundle/Doctrine/ORM/ProductVariantRepository.php
rafalswierczek: Yes. `0` is unwanted behavior. I would like nullable int instead of forced/casted int.

tests/Controller/AdminProductVariantAjaxTest.php
vvasiloi: Sadly with doesn't work with query parameters...

AdamKasp (approved):

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
vvasiloi: So, I had some time and I investigate the issue.
The problem is in the way resource bundle resolve those request parameters.
It uses [Request::get()](https://github.com/Sylius/SyliusResourceBundle/blob/8326f4c4b1daf6a430e242063d9f86988531520e/src/Bundle/Controller/ParametersParser.php#L64), which first looks in the request attributes bag.
Coincidentally, that's where the default route parameters values are stored.
This case proves that that logic gives a bad DX and should be changed, just like Symfony recommends it.
So, here's my solution around it (that I also tested):
```
                    limit: "expr:service('request_stack').getCurrentRequest().query.getInt('limit') > 0 ? service('request_stack').getCurrentRequest().query.getInt('limit') : null"
        limit: 25
```
This will also solve the issue with null being converted to 0 and you will have either a null or integer sent to the repository method, so you can probably remove the condition there.

/cc @loic425 since we talked about that recently

rafalswierczek: What is that second limit key for?

tests/Controller/AdminProductVariantAjaxTest.php
vvasiloi: I would rather add a default value of an empty string for phrase instead of expecting a TypeError.
High level logic like this should not rely on the internal PHP errors.
Moreover, in production environments this will result in a 500 Internal Server Error, which is not good.

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product_variant.yml
vvasiloi: In the expression or the default `limit: 25`?

vvasiloi: Hopefully we can come up with a better solution in the resource bundle, so things like that are easier to do.

tests/Controller/AdminProductVariantAjaxTest.php
rafalswierczek: AFAIK any unhandled throwable will result 500 in production because there is no fallback throwable listener in Sylius.
I tried adding `requirements` but it didn't work for me. Just like in case of your `limit` in `defaults` (I tested it as well and it didn't work) it's only work for slug.
I have no idea how to control the exception throwing instead of type error.

vvasiloi: Yes, requirements don't work for query parameters.
And the default won't work for the same reason as the limit.
How about we use a similar solution then:
```
phrase: expr:service('request_stack').getCurrentRequest().query.get('limit', '')
```

vvasiloi: This way there won't be an error at all, it will fallback to an empty string.

Zales0123 (approved):
Good for now, even better solution than in other autocomplete routes 👍

Zales0123:
Thanks, Rafał! :1st_place_medal:

PR #13622, opened 2022-02-09
Update ResourceBundle and GridBundle

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | no
| New feature?    | yes (well, new features in bundles 😄)
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

~This PR is to see does the current Sylius 1.11 works with upcoming SyliusResourceBundle and SyliusGridBundle versions 🖖 Also would tell us what we need to do more (either in bundles or Sylius itself). Would be great to have this support in 1.11 🚀~

UPDATE: We didn't manage to put it in 1.11 :dancer: But 1.12 is still great. I thought about not bumping the required versions, but that would need us to create another level of build matrix and would eventually result in a workflow nightmare. I believe it's a fair assumption we would like to go with new, modern grid and resource versions 🖖


Roshyo (approved):

lchrusciel (approved):

lchrusciel:
Thanks, Mateusz! :tada:

PR #13625, opened 2022-02-09
Fix: handle SQLite as the operation is not supported otherwise

till:
https://github.com/Sylius/Sylius-Standard/issues/671

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Installer/Provider/DatabaseSetupCommandsProvider.php
+++ b/src/Sylius/Bundle/CoreBundle/Installer/Provider/DatabaseSetupCommandsProvider.php
@@ -54,6 +54,10 @@ final class DatabaseSetupCommandsProvider implements DatabaseSetupCommandsProvid
         $databaseName = $this->getDatabaseName();
 
         try {
+            if ($this->isSqlite()) {
+                return file_exists($databaseName);
+            }
+
             $schemaManager = $this->getSchemaManager();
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/CoreBundle/Installer/Provider/DatabaseSetupCommandsProvider.php
```diff
@@ -122,4 +126,13 @@ private function getEntityManager(): EntityManagerInterface
...
         return $this->doctrineRegistry->getManager();
     }
+
+    private function isSqlite(): bool
+    {
+        $platform = $this->getEntityManager()->getConnection()->getDatabasePlatform();
+        if ($platform instanceof \Doctrine\DBAL\Platforms\SqlitePlatform) {
+            return true;
+        }
+        return false;
```
Roshyo, line 136: `return $platform instanceof \Doctrine\DBAL\Platforms\SqlitePlatform` ?

Also we could use that namespace instead

Roshyo:
It would be great to have more information about the context and things like this. Either in the PR description, or in a separate issue. Can you do that please ?

Roshyo (changes requested):

till:
> It would be great to have more information about the context and things like this. Either in the PR description, or in a separate issue. Can you do that please ?

I referenced my ticket. Let me know what you think.

PR #13642, opened 2022-02-11
[1.11][User] Reduce usage of deprecated `\Serializable` interface

stloyd:
Reduce usage of deprecated `\Serializable` interface which is reported in PHP 8.1.
```
Deprecated: Sylius\Component\User\Model\User implements the Serializable interface, which is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary) in /srv/acme/vendor/sylius/sylius/src/Sylius/Component/User/Model/User.php on line 21
```

Added new `__serialize/__unserialize` methods to cover serialization & made `serialize/unserialize` internal to prevent usage of them (we can remove in next version).

| Q               | A
| --------------- | -----
| Branch?         | 1.11 
| Bug fix?        | yes for PHP 8.1
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | fixed
| License         | MIT

### DIFF
--- a/src/Sylius/Component/User/Model/User.php
+++ b/src/Sylius/Component/User/Model/User.php
@@ -391,12 +391,9 @@ class User implements UserInterface, \Stringable
         $this->encoderName = $encoderName;
     }
 
-    /**
-     * The serialized data have to contain the fields used by the equals method and the username.
-     */
-    public function serialize(): string
+    public function __serialize(): array
     {
-        return serialize([
+        return [
             $this->password,
             $this->salt,
             $this->usernameCanonical,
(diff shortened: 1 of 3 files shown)


Nek-:
@stloyd you have a static check failing :) . This PR is nice for PHP 8.1 full support. +1 for it to be merged!

Roshyo (approved):

src/Sylius/Component/User/Model/User.php
```diff
@@ -405,29 +402,42 @@ public function serialize(): string
...
+        $this->encoderName = $data['encoderName'] ?? null;
+    }
+
+    /**
+     * @internal
      */
     public function unserialize($serialized): void
     {
-        $data = unserialize($serialized);
+        $data = (array) unserialize((string) $serialized);
```
sad270, line 435:
```suggestion
        $data = unserialize($data);
        
        if (!is_array($data)) {
            throw new UnexpectedValueException(sprintf('Cannot deserialize %s instance', self::class));
        }
```
Should fix the Static checks.

src/Sylius/Component/User/Model/User.php
```diff
@@ -405,29 +402,42 @@ public function serialize(): string
...
+    public function __unserialize(array $data): void
+    {
+        $this->password = $data['password'] ?? null;
+        $this->salt = $data['salt'] ?? null;
+        $this->usernameCanonical = $data['usernameCanonical'] ?? null;
+        $this->username = $data['username'] ?? null;
+        $this->locked = $data['locked'] ?? null;
+        $this->enabled = $data['enabled'] ?? null;
+        $this->id = $data['id'] ?? null;
+        $this->encoderName = $data['encoderName'] ?? null;
```
sad270, line 427: This will not work, when we serialize we don't return a associative array !

PR #13656, opened 2022-02-17
Moving the template dependency to the ShopBundle

### LINKED ISSUE #13143: [RFC] Removing the sylius/theme-bundle from the admin package
**The proposal**
Removing the theming functionality for headless Sylius shops. The current move in Sylius is more and more towards a headless approach and making everything exposed via an API. This means that in the core of Sylius we don't need theming support. This is only needed for shops that have a front-end which is based on Sylius.

**Additional context**
The changes that needed to be made are quite minimal as there is little coupling between the theming system and the admin panel in the first place:
* [BC break] The channel doesn't have a `themeName` property if the shop package is not installed (or just leave it at null)
* This also means that the form and the template have to only have the `themeName` fields rendered when the shop is installed.

This also would speed up the application a little (since the theming bundle also has event listeners on kernel.request).

**Is it a good idea?** 
What do you think.

mamazu:
| Q               | A
| --------------- | -----
| Branch?         | master (probably 2.0) 
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | yes
| Deprecations?   | no
| Related tickets | fixes #13143
| License         | MIT

## Implementation details
Currently the only thing I have done is moved the templating stuff to the core bundle. This is enough to also move the dependency officially. I am not sure if this is the best idea to add the if statement in the admin template. I don't know if there is a way to override the AdminBundle template from the ShopBundle.

However this will then leave an unused column in the database as the `themeName` property is still on the channel only not configurable anymore.

## About the BC break
The BC break only occurs for users that use the AdminBundle alone and used the theming functionality.
Depending on which version this is going to be merged in here are is a possible Changelog entry:
```
If you are using the AdminBundle without the Sylius frontend, but you want to keep using the theming functionality sylius provides you now have to require that dependency explictly with

composer require sylius/theme-bundle
```


composer.json
```diff
@@ -239,7 +239,12 @@
         "preferred-install": {
             "*": "dist"
         },
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
+            "dealerdirect/phpcodesniffer-composer-installer": true,
+            "symfony/flex": true,
+            "symfony/thanks": true
+        }
```
AdamKasp, line 247: it should be in separate PR.

src/Sylius/Bundle/ShopBundle/Form/Extension/ChannelTypeExtension.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Sylius\Bundle\ShopBundle\Form\Extension;
+
+use Sylius\Bundle\ChannelBundle\Form\Type\ChannelType;
+use Symfony\Component\Form\AbstractTypeExtension;
+use Symfony\Component\Form\FormBuilderInterface;
+use Sylius\Bundle\ThemeBundle\Form\Type\ThemeNameChoiceType;
+
+class ChannelTypeExtension extends AbstractTypeExtension
```
AdamKasp, line 10:
```suggestion
final class ChannelTypeExtension extends AbstractTypeExtension
```

src/Sylius/Bundle/ShopBundle/Form/Extension/ChannelTypeExtension.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
```
AdamKasp, line 2: missing license block

src/Sylius/Bundle/ShopBundle/Theme/ChannelBasedThemeContext.php
```diff
@@ -11,7 +11,7 @@
 
 declare(strict_types=1);
 
-namespace Sylius\Bundle\CoreBundle\Theme;
```
AdamKasp, line 14: This change is a BC break. this class should stay in `core` we should deprecate it and remove it in 2.0, also this class should be injected into the container conditionally based on `theme bundle` existence.

AdamKasp: It should be tested.

AdamKasp (changes requested):

src/Sylius/Bundle/ShopBundle/Form/Extension/ChannelTypeExtension.php
mamazu: Yes.

composer.json
mamazu: True, this has been removed.

src/Sylius/Bundle/ShopBundle/Theme/ChannelBasedThemeContext.php
mamazu: And about the tests. There are tests for this class. I moved them to the shop bundle as well. Or what exactly should be tested?

src/Sylius/Bundle/ShopBundle/Form/Extension/ChannelTypeExtension.php
```diff
@@ -0,0 +1,39 @@
...
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
```
diimpp:
```suggestion
 * (c) Sylius Sp. z o.o.
```

src/Sylius/Bundle/ShopBundle/Resources/config/services/context.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+
+ This file is part of the Sylius package.
+
+ (c) Paweł Jędrzejewski
```
diimpp:
```suggestion
 (c) Sylius Sp. z o.o.
```

src/Sylius/Bundle/ShopBundle/Resources/config/services/form.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+
+ This file is part of the Sylius package.
+
+ (c) Paweł Jędrzejewski
```
diimpp:
```suggestion
 (c) Sylius Sp. z o.o.
```

src/Sylius/Bundle/ShopBundle/Resources/config/services/form.xml
```diff
@@ -0,0 +1,20 @@
...
+
+ For the full copyright and license information, please view the LICENSE
+ file that was distributed with this source code.
+
+-->
+
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <service id="Sylius\Bundle\ShopBundle\Form\Extension\ChannelTypeExtension">
+            <tag name="form.type_extension" extended-type="Sylius\Bundle\ChannelBundle\Form\Type\ChannelType" />
```
diimpp: I think `extended-type` is not necessary anymore, but it's in line with other form extension configs in project

src/Sylius/Bundle/ShopBundle/Theme/ChannelBasedThemeContext.php
diimpp: Given that service definition `sylius.theme.context.channel_based` didn't change, I would consider this a soft BC.

Class cannot be re-used directly, interface is in theme bundle, decoration will work without an issue, if service definition is overridden, then it will most likely use different class.

Though with 2.0 branch already existing, adding deprecation notice first will be a nice touch.

src/Sylius/Bundle/ShopBundle/Resources/config/services.xml
```diff
@@ -14,7 +14,9 @@
 <container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
     <imports>
         <import resource="services/controller.xml" />
+        <import resource="services/context.xml" />
```
diimpp, line 16: By alphabetical order, `context` should be defined before `controller`

src/Sylius/Bundle/CoreBundle/Form/Extension/ChannelTypeExtension.php
```diff
@@ -67,12 +66,6 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             ->add('taxCalculationStrategy', TaxCalculationStrategyChoiceType::class, [
                 'label' => 'sylius.form.channel.tax_calculation_strategy',
             ])
-            ->add('themeName', ThemeNameChoiceType::class, [
```
diimpp, line 72: I think in should be mentioned in UPGRADE guide as somebody already in headless context could have removed this field, which will crash now.

src/Sylius/Bundle/ShopBundle/Theme/ChannelBasedThemeContext.php
mamazu: The service definition didn't change but it got moved out of the admin bundle into the shop bundle. I don't know how to add a deprecation notice for this. Otherwise I could also make this MR with 2.0 as base then the BC break would be okay.

mamazu:
Yeah I think this would be a thing that makes more sense to be merged against 2.0 since it removes the ability to use the `sylius.theme.context.channel_based` in the admin bundle as it has been moved to the shop bundle. But I can also make another MR adding a deprecation in 1.13.

src/Sylius/Bundle/ShopBundle/Resources/config/services.xml
mamazu: Fixed.

mamazu:
I've rebased this to 2.0 as I said in one of my comments I don't know how to communicate the BC break best in the code of 1.13. Any suggestions would be appreciated.

mamazu:
I think I understand the problem, the integration of the Sylius core with the theme bundle should not depend on the shop bundle but on the theme bundle itself. The question is, does that even make sense to have an admin bundle with theming?

mamazu:
This also has a positive performance impact for headless projects as using the ThemeLocaleAwareListener (part of the kernel.terminate) can be expensive.

src/Sylius/Bundle/CoreBundle/Form/Extension/ChannelTypeExtension.php
```diff
@@ -66,13 +66,18 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             ])
             ->add('taxCalculationStrategy', TaxCalculationStrategyChoiceType::class, [
                 'label' => 'sylius.form.channel.tax_calculation_strategy',
-            ])
-            ->add('themeName', ThemeNameChoiceType::class, [
+            ]);
+
+        if (class_exists('\Sylius\Bundle\ThemeBundle\Form\Type\ThemeNameChoiceType')) {
```
lchrusciel: we could just add another extension in ShopBundle, which will be automatically executed if it is enabled

src/Sylius/Bundle/ShopBundle/Theme/ChannelBasedThemeContext.php
lchrusciel: if we place it before 2.0, then we should leave the class, deprecate it and make a duplication in ShopBundle. If going with 2.0 we can just move it. From the test perspective, we can just try to run tests/build container

src/Sylius/Bundle/ShopBundle/Form/Extension/ChannelTypeExtension.php
```diff
@@ -0,0 +1,30 @@
...
+use Sylius\Bundle\ThemeBundle\Form\Type\ThemeNameChoiceType;
+use Symfony\Component\Form\AbstractTypeExtension;
+use Sylius\Bundle\ChannelBundle\Form\Type\ChannelType;
+use Symfony\Component\Form\FormBuilderInterface;
+
+final class ChannelTypeExtension extends AbstractTypeExtension
+{
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        if (class_exists('\Sylius\Bundle\ThemeBundle\Form\Type\ThemeNameChoiceType')) {
```
diimpp, line 16: Is it possible to move fqcn to `use` and use it here as `ThemeNameChoiceType::class`?

Or would it be better to check for ThemeBundle presence instead and let it fail if for some reason `ThemeNameChoiceType` would be removed? Because otherwise you can have ThemeBundle and not this form type and code will continue to work correctly, which is not desired I think.

src/Sylius/Bundle/ShopBundle/Form/Extension/ChannelTypeExtension.php
```diff
@@ -0,0 +1,30 @@
...
+            $builder->add('themeName', ThemeNameChoiceType::class, [
+                'label' => 'sylius.form.channel.theme',
```
diimpp, line 18: Should translation `sylius.form.channel.theme` be moved to UiBundle to ShopBundle?

UPGRADE-2.0.md
```diff
@@ -181,3 +181,8 @@
 * `use_webpack` option was removed from the `sylius_ui` configuration, and the Webpack has become the only module
   bundler provided by Sylius.
 * `use_webpack` twig global variable was removed. Webpack is always used now, and there is no need to check for it.
+
+## Theming
+If you are running your shop in the headless mode the `sylius/theme-bundle` is no longer installed as it has been moved to the shop bundle.
+* The `themeName` form field has been removed from the Channel, if the shop bundle is not installed.
```
diimpp: Rephrasing is needed, my take (which is not ideal as well)

```suggestion
* Channel's `themeName` form field existence is made optional and depends `ShopBundle` presence.
```

UPGRADE-2.0.md
```diff
@@ -181,3 +181,8 @@
 * `use_webpack` option was removed from the `sylius_ui` configuration, and the Webpack has become the only module
   bundler provided by Sylius.
 * `use_webpack` twig global variable was removed. Webpack is always used now, and there is no need to check for it.
+
+## Theming
+If you are running your shop in the headless mode the `sylius/theme-bundle` is no longer installed as it has been moved to the shop bundle.
```
diimpp:
```suggestion
Dependency on `sylius/theme-bundle` is moved from CoreBundle to ShopBundle and it will no longer be installed if you're running your shop in headless mode.
```

src/Sylius/Bundle/ShopBundle/Form/Extension/ChannelTypeExtension.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ShopBundle\Form\Extension;
+
+use Sylius\Bundle\ThemeBundle\Form\Type\ThemeNameChoiceType;
+use Symfony\Component\Form\AbstractTypeExtension;
+use Sylius\Bundle\ChannelBundle\Form\Type\ChannelType;
+use Symfony\Component\Form\FormBuilderInterface;
```
diimpp: Please apply CS fixes.

```suggestion
use Sylius\Bundle\ChannelBundle\Form\Type\ChannelType;
use Sylius\Bundle\ThemeBundle\Form\Type\ThemeNameChoiceType;
use Symfony\Component\Form\AbstractTypeExtension;
use Symfony\Component\Form\FormBuilderInterface;
```

src/Sylius/Bundle/ShopBundle/Resources/config/services/context.xml
```diff
@@ -0,0 +1,23 @@
...
+ (c) Sylius Sp. z o.o.
+
+ For the full copyright and license information, please view the LICENSE
+ file that was distributed with this source code.
+
+-->
+
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <defaults public="true" />
```
diimpp: Does it needs to be public true for the whole file?

```suggestion
```

src/Sylius/Bundle/ShopBundle/Resources/config/services/context.xml
```diff
@@ -0,0 +1,23 @@
...
+
+        <service id="sylius.theme.context.channel_based" class="Sylius\Bundle\ShopBundle\Theme\ChannelBasedThemeContext" public="false">
```
diimpp:
```suggestion
        <service id="sylius.theme.context.channel_based" class="Sylius\Bundle\ShopBundle\Theme\ChannelBasedThemeContext">
```

src/Sylius/Bundle/ShopBundle/Resources/config/services/form.xml
```diff
@@ -0,0 +1,20 @@
...
+
+ For the full copyright and license information, please view the LICENSE
+ file that was distributed with this source code.
+
+-->
+
+<container xmlns="http://symfony.com/schema/dic/services" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+    <services>
+        <service id="Sylius\Bundle\ShopBundle\Form\Extension\ChannelTypeExtension">
+            <tag name="form.type_extension" extended-type="Sylius\Bundle\ChannelBundle\Form\Type\ChannelType" priority="100" />
```
diimpp: `extended-type` is older syntax, which I think is deprecated (I didn't recheck)

```suggestion
            <tag name="form.type_extension" priority="100" />
```

src/Sylius/Bundle/ShopBundle/Form/Extension/ChannelTypeExtension.php
mamazu: It would probably make sense to check for the presence of the ThemeBundle, this way the code would notice if the `ThemeNameChoiceType` doesn't exist in the bundle anymore.

src/Sylius/Bundle/ShopBundle/Form/Extension/ChannelTypeExtension.php
mamazu: Then we also need to duplicate the translation. But what key would you suggest.

src/Sylius/Bundle/ShopBundle/Form/Extension/ChannelTypeExtension.php
diimpp: License block is missing. (I guess something happened at rebase)

src/Sylius/Bundle/ShopBundle/Form/Extension/ChannelTypeExtension.php
diimpp: I think we can just keep existing key and duplication is not necessary (Since this PR targets 2.0), so to move existing `sylius.form.channel.theme` translation key to ShopBundle.

mamazu: Fair enough. Is there an easy way to do this or do I need to touch all the 46 translation files of the admin bundle? :D

GSadee:
I'm closing in favour of #17106

PR #13666, opened 2022-02-18
Use the index of the last child of a collection to construct the element's form

delyriand:
| Q               | A
| --------------- | -----
| Branch?         | 1.8
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

When we have a CollectionType with children that have a numerical index, the "add" button can have a strange behaviour when deleting elements and adding new items.
For example, we have an array of items like this:
- 1 => [...]
- 3 => [...]
- 4 => [...]
- 5 => [...]

Previously, the JS code count the number of items, so 4, and add a new form with inputs name `my_beautiful_collection[items][4][XXXX]`. And when we submit the form, my "4" item is replace by the new "4"…

Like explained in Symfony doc, I use the last collection item to calculate the index of the new element: https://symfony.com/doc/5.4/form/form_collections.html#allowing-new-tags-with-the-prototype

This patch should be applied on all still maintained branches: 1.9, 1.10, 1.11 and master

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-form-collection.js
+++ b/src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-form-collection.js
@@ -24,7 +24,7 @@ class CollectionForm {
 
     this.$element = $(element);
     this.$list = this.$element.find('[data-form-collection="list"]:first');
-    this.count = this.$list.children().length;
+    this.nextIndexElement = this.$element.data('index');
     this.lastChoice = null;
     this.$element.on('click', '[data-form-collection="add"]:last', this.addItem);
     this.$element.on('click', '[data-form-collection="delete"]', this.deleteItem);
@@ -46,10 +46,10 @@ class CollectionForm {
     let prototype = this.$element.data('prototype');
     let prototypeName = new RegExp(this.$element.data('prototype-name'), 'g');
 
-    prototype = prototype.replace(prototypeName, this.count);
+    prototype = prototype.replace(prototypeName, this.nextIndexElement);
 
     this.$list.append(prototype);
-    this.count = this.count + 1;
+    this.nextIndexElement = this.nextIndexElement + 1;
 
     $(document).trigger('collection-form-add', [this.$list.children().last()]);
   }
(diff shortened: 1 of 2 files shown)


AdamKasp:
hello, @delyriand I am not able to reproduce it, can you provide more detailed descriptions (maybe some screenshots?)
any tests also will be appreciated :)

maximehuran:
It seems the post data are not correct : 

image

And the second time when it's ok : 

image

I search a little and found it's a configuration to add in my form :
```php

            ->add('openings', CollectionType::class, [
                'entry_type' => OpeningType::class,
                'prototype_name' => '__opening__',
                'label' => false,
```

I added `prototype_name` in all my collection forms to have `__name__` by default this will cause issue while replace.
So there is no issue with the embed collection in Sylius' javascript.

Put this pull request still important to avoid problems with the elements numerotation.

jacquesbh:
We've got an issue with this PR when we edit an attribute: `An exception has been thrown during the rendering of a template ("Warning: A non-numeric value encountered").` in `vendor/sylius/sylius/src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig:90`

The issue is because `form.children|last.vars.name` has a non numeric valud (`45fee6f4-1e37-11ed-8248-827b9ec727bb`).

A change could be:

```diff
+               {% if form.children|last.vars.name matches '/^\\d+$/' %}
                 data-index='{{ form.children|length > 0 ? form.children|last.vars.name + 1 : 0 }}'
+               {% endif %}
```

but I don't know if it keep fixing your own issue :/

maximehuran:
Hi @AdamKasp 

We have this issue in form collections in AJAX, like in [Rich Editor Plugin](https://github.com/monsieurbiz/SyliusRichEditorPlugin/).
For now, we apply the patch on our projects but some people have the same issue and they would like to have the fix in Sylius directly.

- https://github.com/monsieurbiz/SyliusRichEditorPlugin/pull/192
- https://github.com/monsieurbiz/SyliusRichEditorPlugin/pull/220

maximehuran:
Hi @GSadee can you have a look to this PR ?

maximehuran:
Hi @GSadee can you have a look to this PR ?

welcoMattic:
Hi @GSadee can you have a look to this PR, I've met this bug several times in the past weeks.

PR #13603, opened 2022-02-04
[UI] Prevent anonymous user to claim logged user cart

SirDomin:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

PR #13676, opened 2022-02-22
[Cart] Fix retrieving/overriding cart of logged in user by guest

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | replaces https://github.com/Sylius/Sylius/pull/13603 and https://github.com/Sylius/Sylius/pull/13634
| License         | MIT


src/Sylius/Bundle/CoreBundle/Migrations/Version20220210135918.php
```diff
@@ -0,0 +1,30 @@
...
+final class Version20220210135918 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add by_guest field to mark orders made by guests';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE sylius_order ADD by_guest TINYINT(1) DEFAULT \'1\' NOT NULL');
```
AdamKasp, line 22:
```suggestion
        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');

        $this->addSql('ALTER TABLE sylius_order ADD by_guest TINYINT(1) DEFAULT \'1\' NOT NULL');
```

AdamKasp: this field is not-nullable, what will happen with already placed orders or carts? should we prepare migration for those orders?

GSadee: Added 👌🏻

UPGRADE-1.10.md
```diff
@@ -1,3 +1,10 @@
+# UPGRADE FROM `v1.10.8` TO `v1.10.10`
+
+1. Field `byGuest` has been added to `Sylius\Component\Core\Model\Order`, this change will allow us to distinguish carts 
+between guests and logged in customers.
+
+2. Not passing `tokenStorage` through constructor is deprecated in Sylius 1.10.9 and it will be removed in Sylius 2.0.
```
Zales0123, line 6: What constructor? 😄 

```suggestion
2. Not passing `tokenStorage` through `Sylius\Component\Core\Cart\Context\ShopBasedCartContext` constructor is deprecated in Sylius 1.10.9 and it will be removed in Sylius 2.0.
```

adr/2022_02_21_using_cart_by_guest_and_logged_in_customer.md
```diff
@@ -0,0 +1,53 @@
...
+## Decision Outcome
+
+Chosen option: **"Introducing flag on order entity to mark order created by guest"**
+
+For now, it is the most straightforward solution, that resolves the initial problem, does not introduce any BC break
+and it is the solution that could be used to additional features in the future.
+
+## References
+
+* [Approach with chnaging priorities](https://github.com/Sylius/Sylius/pull/13603)
```
Zales0123, line 52:
```suggestion
* [Approach with changing priorities](https://github.com/Sylius/Sylius/pull/13603)
```

src/Sylius/Behat/Context/Ui/Shop/AuthorizationContext.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * @When I sign in again with email :email and password :password in the previous session
+     */
+    public function iSignInAgainWithEmailAndPasswordInThePreviousSession(string $email, string $password): void
+    {
+        $this->loginPage->open();
+        $this->loginPage->specifyUsername($email);
+        $this->loginPage->specifyPassword($password);
+        $this->loginPage->logIn();
```
Zales0123, line 63: It does not differ from the previous function at all. Shouldn't it? Or this "previous session" part is done somewhere else?

src/Sylius/Behat/Service/SessionManager.php
```diff
@@ -0,0 +1,79 @@
...
+
+final class SessionManager implements SessionManagerInterface
+{
+    private Mink $mink;
+
+    private SharedStorageInterface $sharedStorage;
+
+    private SecurityServiceInterface $securityService;
+
+    private const SESSION_CHROME_HEADLESS_SECOND = 'chrome_headless_second_session';
```
Zales0123, line 27: Const above properties :dancer:

src/Sylius/Behat/Service/SessionManager.php
```diff
@@ -0,0 +1,79 @@
...
+        $this->securityService = $securityService;
+    }
+
+    public function changeSession(): void
+    {
+        $sessionName = self::SESSION_CHROME_HEADLESS_SECOND;
+
+        $this->saveAndRestartSession($sessionName);
+
+        if ($this->sharedStorage->has(sprintf('behat_previous_session_token_%s', $sessionName))) {
```
Zales0123, line 42: `sprintf('behat_previous_session_token_%s', $sessionName)` could be extracted to variable... or private function to avoid duplications and possible mistakes

src/Sylius/Behat/spec/Service/SessionManagerSpec.php
```diff
@@ -0,0 +1,131 @@
...
+    {
+        $this->beConstructedWith($mink, $sharedStorage, $securityService);
+    }
+
+    function it_implements_session_service_interface(): void
+    {
+        $this->shouldImplement(SessionManagerInterface::class);
+    }
+
+    function it_changes_session_and_doesnt_restore_session_token_if_session_was_not_called_before(
```
Zales0123, line 36: `doesnt` :( `does not`?

src/Sylius/Component/Core/Cart/Context/ShopBasedCartContext.php
```diff
@@ -84,6 +99,21 @@ private function setCustomerAndAddressOnCart(OrderInterface $cart, CustomerInter
         }
     }
 
+    private function resolveByGuestFlag(): bool
```
Zales0123, line 102: I would definitely put this whole logic into a separate service, as this way it's not customisable at all. btw, private function is placed above public 🐎

adr/2022_02_21_using_cart_by_guest_and_logged_in_customer.md
```diff
@@ -0,0 +1,53 @@
...
+* Bad, because it does not solve the case when a logged in customer does not have a cart
+* Bad, because it changes the current expected behaviour with keeping cart after logging in
+* Bad, because it breaks a backward compatibility from the business perspective
+
+### Introducing flag on order entity to mark order created by guest
+
+* Good, because it solves the initial problem
+* Good, because it does not break a backward compatibility
+* Good, because it can be used to additional features in the future
+* Bad, because it requires to set the flag in Sylius\Component\Core\Cart\Context\ShopBasedCartContext
```
Zales0123, line 40: Why cannot we set this flag on the Order during login/logout? It would be the most natural solution, as it seems. If it's not possible for any reason, I would put this information here 🖖

GSadee: Added some information to ADR, I hope it will be enough 👌🏻

AdamKasp (approved):

src/Sylius/Component/Core/Cart/Resolver/ByGuestFlagResolver.php
```diff
@@ -0,0 +1,41 @@
...
+        if ($token === null) {
+            return true;
+        }
+
+        $user = $token->getUser();
+        if ($user !== null) {
+            return false;
+        }
+
+        return true;
```
stloyd, line 39:
```suggestion
        return null === $token->getUser();
```

adr/2022_02_21_using_cart_by_guest_and_logged_in_customer.md
```diff
@@ -0,0 +1,60 @@
+# Using cart by guest and logged in customer
+
+* Status: accepted
+* Date: 2022-02-22
+
+## Context and Problem Statement
+
+Cart and its processing is one of the key aspects of Sylius. It turned out that it has a vulnerability and there is
+possible for an anonymous user to retrieve or override the cart of logged in customer by using only its email.
```
lchrusciel, line 9: Can we really retrieve their cart? We can override it and then we may have access to it
```suggestion
possible for an anonymous user to override the cart of the logged-in customer by using only its email.
```

src/Sylius/Component/Core/Model/Order.php
```diff
@@ -110,6 +110,8 @@ class Order extends BaseOrder implements OrderInterface
      */
     protected $customerIp;
 
+    protected bool $byGuest = true;
```
lchrusciel, line 113: I'm not a fan of `byGuest`, but it is hard to propose something much better instead indeed. What would you say for one of these suggestions: 

This one seems to be the most readable:
```suggestion
    protected bool $guestOrder = true;
```

```suggestion
    protected bool $guest = true;
```

```suggestion
    protected bool $guested = true;
```

```suggestion
    protected bool $authrorized = true;
```

```suggestion
    protected bool $authrorizedOrder = true;
```

GSadee: To be honest, I'm also not fully convinced to `byGuest` but for me it seems to be the most readable version and none of your suggestions convinces me more 😢 maybe there should be sth like `createdByGuest`?
But for now, I will leave the current version

lchrusciel: `createdByGuest` is the winner

SirDomin (approved):

AdamKasp:

PR #13694, opened 2022-02-24
[Cache][Varnish][WIP] Invalidate shop cache if relevant admin resources are changed

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

When using Varnish with Api Platform, when the e.g. product variant is updated in the admin API, its cache is refreshed for Admin **but not in the Shop**. With this quick fix, we force refreshing the cache for all resources updated in the admin panel that is also relevant in the shop. It's definitely not a perfect solution - but it should work.

BEWARE! 

It's not working for now :dancer: Because I cannot force varnish to work on Github actions :( I tried, but my knowledge and experience are limited in this area. So if anyone has some idea how to make Varnish work - please, check it out 🖖 cc @Sylius/core-team

### DIFF
--- a/.github/workflows/application.yml
+++ b/.github/workflows/application.yml
@@ -583,6 +583,67 @@ jobs:
                     path: etc/build/
                     if-no-files-found: ignore
 
+    test-application-caching-mechanism:
+        name: PHP 8.0, Symfony 6, Varnish 6
+        runs-on: ubuntu-20.04
+        strategy:
+            fail-fast: false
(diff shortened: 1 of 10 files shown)


Zales0123:
Closing as it's not a huge issue at the moment 🖖

PR #13695, opened 2022-02-24
[Order] Change getters and setters to use a proper flag name

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/Sylius/Sylius/pull/13676
| License         | MIT

PR #13700, opened 2022-02-24
[Order] Fix getters and setters to use a proper flag name after upmerge issue

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | after https://github.com/Sylius/Sylius/pull/13695
| License         | MIT

### DIFF
--- a/src/Sylius/Component/Core/Model/Order.php
+++ b/src/Sylius/Component/Core/Model/Order.php
@@ -477,12 +477,12 @@ class Order extends BaseOrder implements OrderInterface
         return $total;
     }
 
-    public function getByGuest(): bool
+    public function getCreatedByGuest(): bool
     {
         return $this->createdByGuest;
     }
 
-    public function setByGuest(bool $createdByGuest): void
+    public function setCreatedByGuest(bool $createdByGuest): void
     {
         $this->createdByGuest = $createdByGuest;
     }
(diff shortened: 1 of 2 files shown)


src/Sylius/Component/Core/Model/Order.php
```diff
@@ -477,12 +477,12 @@ public function getNonDiscountedItemsTotal(): int
         return $total;
     }
 
-    public function getByGuest(): bool
+    public function getCreatedByGuest(): bool
```
Roshyo, line 480: Shouldn't it be `isCreatedByGuest` instead ?

AdamKasp (approved):

AdamKasp:

src/Sylius/Component/Core/Model/Order.php
Zales0123: It probably should... but this is a PR just to fix conflicts after upmerge, we can fix it from 1.10 ⬆️ later on :)

GSadee: It could be, it looks probably more natural, but it should be changed on 1.10 branch, as the new field has been introduced there, in this PR I'm only fixing the problem after upmerge branches

PR #13708, opened 2022-02-28
[Locale] Add header based locale context

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/LocaleBundle/Context/RequestHeaderBasedLocaleContext.php
+++ b/src/Sylius/Bundle/LocaleBundle/Context/RequestHeaderBasedLocaleContext.php
@@ -1,0 +1,57 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\LocaleBundle\Context;
+
+use Sylius\Component\Locale\Context\LocaleContextInterface;
+use Sylius\Component\Locale\Context\LocaleNotFoundException;
+use Sylius\Component\Locale\Provider\LocaleProviderInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+final class RequestHeaderBasedLocaleContext implements LocaleContextInterface
+{
+    public function __construct(private RequestStack $requestStack, private LocaleProviderInterface $localeProvider)
+    {
+    }
+
+    public function getLocaleCode(): string
+    {
+        $request = $this->getMainRequest();
+
+        if (null === $request) {
+            throw new LocaleNotFoundException('No main request available.');
+        }
+
+        $localeCode = $request->headers->get('Accept-Language');
+        if (null === $localeCode) {
+            throw new LocaleNotFoundException('No locale is set on the master request\'s headers.');
+        }
+
+        $availableLocalesCodes = $this->localeProvider->getAvailableLocalesCodes();
+        if (!in_array($localeCode, $availableLocalesCodes, true)) {
+            throw LocaleNotFoundException::notAvailable($localeCode, $availableLocalesCodes);
+        }
+
+        return $localeCode;
+    }
+
+    private function getMainRequest(): ?Request
+    {
+        if (\method_exists($this->requestStack, 'getMainRequest')) {
+            return $this->requestStack->getMainRequest();
+        }
+
+       return $this->requestStack->getMasterRequest();
+    }
+}
--- a/src/Sylius/Bundle/LocaleBundle/spec/Context/RequestHeaderBasedLocaleContextSpec.php
+++ b/src/Sylius/Bundle/LocaleBundle/spec/Context/RequestHeaderBasedLocaleContextSpec.php
@@ -1,0 +1,97 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace spec\Sylius\Bundle\LocaleBundle\Context;
+
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Locale\Context\LocaleContextInterface;
+use Sylius\Component\Locale\Context\LocaleNotFoundException;
+use Sylius\Component\Locale\Provider\LocaleProviderInterface;
+use Symfony\Component\HttpFoundation\HeaderBag;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+
+final class RequestHeaderBasedLocaleContextSpec extends ObjectBehavior
+{
+    function let(RequestStack $requestStack, LocaleProviderInterface $localeProvider): void
+    {
+        $this->beConstructedWith($requestStack, $localeProvider);
+    }
+
+    function it_is_a_locale_context(): void
+    {
+        $this->shouldImplement(LocaleContextInterface::class);
+    }
+
+    function it_throws_locale_not_found_exception_if_main_request_is_not_found(RequestStack $requestStack): void
+    {
+        if (\method_exists(RequestStack::class, 'getMainRequest')) {
+            $requestStack->getMainRequest()->willReturn(null);
+        } else {
+            $requestStack->getMasterRequest()->willReturn(null);
+        }
+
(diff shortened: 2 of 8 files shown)


src/Sylius/Bundle/LocaleBundle/Context/RequestHeaderBasedLocaleContext.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+    }
+
+    public function getLocaleCode(): string
+    {
+        if (\method_exists($this->requestStack, 'getMainRequest')) {
+            $request = $this->requestStack->getMainRequest();
+        } else {
+            $request = $this->requestStack->getMasterRequest();
+        }
```
GSadee, line 33: Extract this code to a private method to avoid using `else`

tests/Api/Responses/Expected/shop/product/get_product_with_de_DE_locale_translation.json
```diff
@@ -0,0 +1,26 @@
...
+    "\/api\/v2\/shop\/product-options\/COLOR"
+  ],
+  "createdAt": @date@,
+  "updatedAt": @date@,
+  "description": "das ist becher",
+  "shortDescription": "Kurzbeschreibung der Becher",
+  "slug": "Becher",
+  "name": "Becher",
+  "defaultVariant": "\/api\/v2\/shop\/product-variants\/MUG_BLUE"
+}
```
GSadee, line 26: I would be for using 4 spaces for indentation

GSadee (commented):
Is there a possibility to use it in Swagger docs?

AdamKasp:
@GSadee at this moment, we can't use it in swagger, but we overestimated this task so I will check how to add it to swagger

Roshyo (approved):

tests/Api/Shop/ProductsTest.php
```diff
@@ -20,19 +20,40 @@
...
-    {
-        $this->loadFixturesFromFile('product/product_variant_with_original_price.yaml');
-
-        $this->client->request('GET', '/api/v2/shop/products-by-slug/mug?paramName=paramValue', [], [], self::CONTENT_TYPE_HEADER);
-        $response = $this->client->getResponse();
-
-        $this->assertEquals('/api/v2/shop/products/MUG?paramName=paramValue', $response->headers->get(('Location')));
-        $this->assertResponseCode($response, Response::HTTP_MOVED_PERMANENTLY);
-    }
+//    public function it_preserves_query_param_when_redirecting_from_product_slug_to_product_code(): void
```
GSadee, line 23: Probably this code should not be commented out 🕵🏻

src/Sylius/Bundle/LocaleBundle/spec/Context/RequestHeaderBasedLocaleContextSpec.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        $this->beConstructedWith($requestStack, $localeProvider);
+    }
+
+    function it_is_a_locale_context(): void
+    {
+        $this->shouldImplement(LocaleContextInterface::class);
+    }
+
+    function it_throws_locale_not_found_exception_if_master_request_is_not_found(RequestStack $requestStack): void
```
lchrusciel, line 36:
```suggestion
    function it_throws_locale_not_found_exception_if_main_request_is_not_found(RequestStack $requestStack): void
```

according to exception message

src/Sylius/Bundle/LocaleBundle/spec/Context/RequestHeaderBasedLocaleContextSpec.php
```diff
@@ -0,0 +1,97 @@
...
+        if (\method_exists(RequestStack::class, 'getMainRequest')) {
+            $requestStack->getMainRequest()->willReturn(null);
+        } else {
+            $requestStack->getMasterRequest()->willReturn(null);
+        }
+
+        $this->shouldThrow(LocaleNotFoundException::class)->during('getLocaleCode');
+    }
+
+    function it_throws_locale_not_found_exception_if_master_request_does_not_have_accept_language_in_header(
```
lchrusciel, line 47:
```suggestion
    function it_throws_locale_not_found_exception_if_main_request_does_not_have_accept_language_in_header(
```

src/Sylius/Bundle/LocaleBundle/spec/Context/RequestHeaderBasedLocaleContextSpec.php
```diff
@@ -0,0 +1,97 @@
...
+        } else {
+            $requestStack->getMasterRequest()->willReturn($request);
+        }
+
+        $request->headers = new HeaderBag();
+
+        $this->shouldThrow(LocaleNotFoundException::class)->during('getLocaleCode');
+    }
+
+    function it_throws_locale_not_found_exception_if_master_request_locale_code_is_not_among_available_ones(
```
lchrusciel, line 62:
```suggestion
    function it_throws_locale_not_found_exception_if_main_request_locale_code_is_not_among_available_ones(
```

src/Sylius/Bundle/LocaleBundle/spec/Context/RequestHeaderBasedLocaleContextSpec.php
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        $request->headers = new HeaderBag(['ACCEPT_LANGUAGE' => 'en_US']);
+
+        $localeProvider->getAvailableLocalesCodes()->willReturn(['pl_PL', 'de_DE']);
+
+        $this->shouldThrow(LocaleNotFoundException::class)->during('getLocaleCode');
+    }
+
+    function it_returns_master_request_locale_code(
```
lchrusciel, line 80:
```suggestion
    function it_returns_main_request_locale_code(
```

GSadee (approved):

GSadee:
Thank you, Adam! :tada:

src/Sylius/Bundle/LocaleBundle/spec/Context/RequestHeaderBasedLocaleContextSpec.php
```diff
@@ -0,0 +1,97 @@
...
+        LocaleProviderInterface $localeProvider,
+        Request $request
+    ): void {
+        if (\method_exists(RequestStack::class, 'getMainRequest')) {
+            $requestStack->getMainRequest()->willReturn($request);
+        } else {
+            $requestStack->getMasterRequest()->willReturn($request);
+        }
+
+        $request->headers = new HeaderBag(['ACCEPT_LANGUAGE' => 'en_US']);
```
lchrusciel, line 73: Why is this header written differently than the other one? Perhaps we could extract it to private const

gseric:
Sorry for commenting too late (after merge), there are some issues with this implementation.

This language negotiation will work for simple locales (en, fr...), but not for those with region included (en_GB, fr_FR...). The reason is that Sylius uses posix locales and Accept-Language uses [language tags](https://datatracker.ietf.org/doc/html/rfc5646). Example of language tag: en-GB. Those two are compared as-is, even without case normalization despite the fact they are case insensitive.

Also it doesn't handle generalization, e.g. en_GB -> en. Matching locales is not simple task, check [how Symfony is doing it](https://github.com/symfony/http-foundation/blob/b460fb15905eef449c4c43a4f0c113eccee103b9/Request.php#L1539). Though, they don't support more advanced scenarios either.

Another issue worth mentioning is that this language negotiation has bigger priority (256) than one based on routing "_locale" attribute (64). Explicit locale set via URL should be more important.

LucaGallinari:
Hi guys, why this has been released as a bugfix (on 1.11.1) and not as a new feature (on 1.12)? At least this should be a new minor version.
For some, this will probably mess up their custom LocaleContext. Until now the highest priority for services tagged with "sylius.context.locale" was 128, so many could have created LocaleContext's with priority >= 128 and <256, and these will be ignored.

lchrusciel:
Hey guys! 

Let me explain it all together. It was open to 1.11, as we removed all locales for products etc. in the shop part and without this logic, you can't see other translations of products. Therefore, we've concluded it will be bugfix. 

About priority - I'm sorry if we broke your app, but adding something above is not breaking change according to our policy. Perhaps we should add it to UPGRADE.md. 

the generalization is a generic issue of Sylius and I would love to solve it, but it wasn't the scope of this task. 

The case of: "en-GB" should be probably fixed in our current implementation, but it does not invalidate that with the current context Sylius works better. Are you open to open a PR with fix?

gseric:
@lchrusciel np, I'll look into it this weekend

lchrusciel:
Hey David, 

I'm sorry for your inconvenience. What you described, is for sure a proper workaround for you. In the meantime, we will work on the proper fix in our codebase

PR #13709, opened 2022-02-28
Fix fatal error when OPCache preloading is used

mmenozzi:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

While in Symfony 5 the `Symfony\Bundle\FrameworkBundle\Templating\EngineInterface` was removed, in Sylius there were still references to that interface in docblock comments.

With Sylius 1.11 ([with this commit](https://github.com/Sylius/Sylius/commit/ced91b9bd6c3f3d944e57bb54a71fc309c0cb7b0)) the type hint has been promoted to a PHP type hint. This normally does not cause any problem because apparently the class loading is lazy and does not try to load the non-existent interface.

But if PHP OPCache preload is used, like suggested [here](https://symfony.com/doc/current/performance.html#use-the-opcache-class-preloading), it appears that PHP switches to eager class loading and you get the following fatal error:

```
NOTICE: PHP message: PHP Fatal error:  Failed to load class Symfony\Bundle\FrameworkBundle\Templating\EngineInterface used by typed property Sylius\Bundle\ShopBundle\Controller\SecurityWidgetController::$templatingEngine during preloading in Unknown on line 0
```

With this PR we propose to remove all references to the old interface as Symfony 4 is not supported anymore by Sylius.


mmenozzi:
The build is red but I don't think it's related to these changes.

Roshyo (approved):

vvasiloi:
Hey @mmenozzi! I finally tried to reproduce this issue and it occurs with PHP 8.0, but not with PHP 8.1.
It's an issue faced by Symfony too and they simply [bumped the minimum required PHP version to 8.1 in Symfony 6.1](https://symfony.com/blog/symfony-6-1-will-require-php-8-1).
Perhaps Sylius should do the same in the next minor release.

@Sylius/core-team I would like to see more opinions on this.

mmenozzi:
Thank you @Ferror this is exactly what I think.

vvasiloi:
Removing the reference to that interface will mitigate the problem at hand.
However, if we choose to keep support for PHP 8.0, then we will need a test for preloading to avoid another issue.

vvasiloi (approved):

mmenozzi:
Yes I agree it could be useful to have a test for that but I think it should be done in a separate PR. Note also that there's already a kind of test: it's PHPStan/psalm.

vvasiloi:
Sure, I was also thinking of a separate PR. Psalm/PHPStorm will probably be enough for Sylius code, but another issue might be coming from it's 3rd party dependencies.

GSadee (approved):

lchrusciel (approved):

GSadee:
Thank you, Manuele! :1st_place_medal:

AdamKasp (approved):

mmenozzi:
> @mmenozzi are you willing to do the PR with testing for it?

@lchrusciel do you have any idea on how to test it? AFAIK the preload is done only in the `prod` environment but we use the `test_cached` in CI...

PR #13721, opened 2022-03-03
Fix disappearing products in Admin Panel when translation for current locale is missing

### LINKED ISSUE #13296: [AdminBundle] Product isn't displaying if the translation for the current locale is missing

jakubtobiasz:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #13296
| License         | MIT

When accepted I can open similar PRs for `1.9` and `1.10`.

**Description**
Hi 🙋🏻‍♂️!

Let's say our products has translations for `Polish (Poland)` and `English (United States)` locales. Currently if we switch to `Polish` or `English` or any other "non-supported" locale we can't see any product in Admin Panel.

This PR fixes it by displaying all the products and in case of missing name's translation displays `missing translation` info.
CleanShot 2022-03-03 at 7 53 16

Cases covered by tests:
- Products with missing translations sorted by name ascending should be displayed as first
- Products with missing translations sorted by name descending should be displayed as last

These two cases also covers implicitly:
- If product has missing translation display `missing translation` as a product name

In case any change needed just let me know ✌🏻.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -240,7 +240,13 @@
         "preferred-install": {
             "*": "dist"
         },
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
+            "composer/package-versions-deprecated": true,
+            "dealerdirect/phpcodesniffer-composer-installer": true,
+            "symfony/flex": true,
+            "symfony/thanks": true
+        }
     },
     "extra": {
         "branch-alias": {
--- a/features/product/managing_products/sorting_products.feature
+++ b/features/product/managing_products/sorting_products.feature
@@ -13,6 +13,7 @@ Feature: Sorting listed products
         And this product is named "Mops Miłości" in the "Polish (Poland)" locale
         And the store also has a product "Xtreme Pug" with code "X_PUG"
         And this product is named "Ekstremalny Mops" in the "Polish (Poland)" locale
+        And this product is named "Ekstremalny Mops" in the "Polish" locale
         And I am logged in as an administrator
 
     @ui @api
@@ -63,3 +64,22 @@ Feature: Sorting listed products
         And I sort the products descending by name
         Then I should see 3 products in the list
         And the first product on the list should have name "Szałowy Mops"
+
+    @ui
+    Scenario: Missing translations are sorted as first when sorting by name ascending
+        When I change my locale to "Polish"
+        And I browse products
+        And the products are already sorted ascending by name
+        Then I should see 3 products in the list
+        And the first product on the list should have name "brakujące tłumaczenie"
+        And the last product on the list should have name "Ekstremalny Mops"
+
+    @ui
+    Scenario: Missing translation are sorted as last when sorting by name descending
+        When I change my locale to "Polish"
+        And I browse products
+        And the products are already sorted ascending by name
+        And I sort the products descending by name
+        Then I should see 3 products in the list
+        And the first product on the list should have name "Ekstremalny Mops"
+        And the last product on the list should have name "brakujące tłumaczenie"
--- a/src/Sylius/Bundle/UiBundle/Resources/translations/messages.pl.yml
+++ b/src/Sylius/Bundle/UiBundle/Resources/translations/messages.pl.yml
@@ -491,6 +491,7 @@ sylius:
         meta_keywords: 'Słowa kluczowe w sekcji meta'
         metadata_index: 'Informacje o dokumencie'
         method: 'metoda'
+        missing_translation: 'brakujące tłumaczenie'
         modified: 'modyfikacja'
         money: 'Pieniądze'
         most_expensive_first: 'Najpierw najdroższe'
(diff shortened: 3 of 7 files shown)


lchrusciel (approved):
However, I would prefer to have the "allow-plugins" fix separated to another PR. Nonetheless, I'm in favour of both changes

lchrusciel:
As this is bugfx, it should be open to the oldest maintained branch - 1.10

src/Sylius/Bundle/UiBundle/Resources/translations/messages.pl.yml
```diff
@@ -491,6 +491,7 @@ sylius:
         meta_keywords: 'Słowa kluczowe w sekcji meta'
         metadata_index: 'Informacje o dokumencie'
         method: 'metoda'
+        missing_translation: 'brakujące tłumaczenie'
```
Zales0123, line 494: We should not provide any non-English translation - Crowdin would take care of that 🖖

features/product/managing_products/sorting_products.feature
```diff
@@ -63,3 +64,22 @@ Feature: Sorting listed products
...
+
+    @ui
+    Scenario: Missing translation are sorted as last when sorting by name descending
+        When I change my locale to "Polish"
+        And I browse products
+        And the products are already sorted ascending by name
+        And I sort the products descending by name
+        Then I should see 3 products in the list
+        And the first product on the list should have name "Ekstremalny Mops"
+        And the last product on the list should have name "brakujące tłumaczenie"
```
Zales0123, line 85: We should probably not check the exact text, but rather check is there a test data attribute `data-test-missing-translation` so it would not fail if the translation is changed in Crowdin :)

composer.json
```diff
@@ -240,7 +240,13 @@
...
             "*": "dist"
         },
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
+            "composer/package-versions-deprecated": true,
+            "dealerdirect/phpcodesniffer-composer-installer": true,
+            "symfony/flex": true,
+            "symfony/thanks": true
+        }
```
Zales0123, line 249: Should be done in a separate PR

src/Sylius/Bundle/UiBundle/Resources/translations/messages.pl.yml
AdamKasp, line 494: translation different that English should be made crowdin

Zales0123 (commented):
Can you open it already to 1.10? It would be upmerged to newer branches 🚀 Thanks for your contribution!

features/product/managing_products/sorting_products.feature
```diff
@@ -63,3 +64,22 @@ Feature: Sorting listed products
         And I sort the products descending by name
         Then I should see 3 products in the list
         And the first product on the list should have name "Szałowy Mops"
+
+    @ui
+    Scenario: Missing translations are sorted as first when sorting by name ascending
+        When I change my locale to "Polish"
+        And I browse products
+        And the products are already sorted ascending by name
```
AdamKasp, line 72:
```suggestion
        And  I sort products  ascending by name
```

composer.json
loic425: But I agree to fix it in a separate PR.

jakubtobiasz:
As this PR was based on `master` branch I've created a new PR based on `1.10`. Also I've applied all suggestions from CR. Thanks for all feedback @Zales0123 @lchrusciel @AdamKasp @loic425 

New PR is #13724. I'm closing this PR as it's not needed anymore ✌🏻.

PR #13724, opened 2022-03-03
Fix disappearing products in Admin Panel when translation for current locale is missing

### LINKED ISSUE #13296: [AdminBundle] Product isn't displaying if the translation for the current locale is missing

jakubtobiasz:
> Previous PR (#13721) was based on a master branch so I've created a new branch based on the `1.10` one, cherry-picked changes and applied suggestions from CR :).

| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #13296
| License         | MIT

**Description**
Hi 🙋🏻‍♂️!

Let's say our products has translations for `Polish (Poland)` and `English (United States)` locales. Currently if we switch to `Polish` or `English` or any other "non-supported" locale we can't see any product in Admin Panel.

This PR fixes it by displaying all the products and in case of missing name's translation displays `missing translation` info.
CleanShot 2022-03-03 at 7 53 16

Cases covered by tests:
- Products with missing translations sorted by name ascending should be displayed as first
- Products with missing translations sorted by name descending should be displayed as last

These two cases also covers implicitly:
- If product has missing translation display `missing translation` as a product name

In case any change needed just let me know ✌🏻.

### DIFF
--- a/features/product/managing_products/sorting_products.feature
+++ b/features/product/managing_products/sorting_products.feature
@@ -71,7 +71,7 @@ Feature: Sorting listed products
         And I browse products
         And I sort the products ascending by name
         Then I should see 3 products in the list
-        And the first product on the list should have "data-test-missing-translation-paragraph" data attribute
+        And the first product on the list shouldn't have a name
         And the last product on the list should have name "Ekstremalny Mops"
 
     @ui
@@ -82,4 +82,4 @@ Feature: Sorting listed products
         And I sort the products descending by name
         Then I should see 3 products in the list
         And the first product on the list should have name "Ekstremalny Mops"
-        And the last product on the list should have "data-test-missing-translation-paragraph" data attribute
+        And the last product on the list shouldn't have a name
--- a/features/product/managing_products/sorting_products.feature
+++ b/features/product/managing_products/sorting_products.feature
@@ -13,6 +13,7 @@ Feature: Sorting listed products
         And this product is named "Mops Miłości" in the "Polish (Poland)" locale
         And the store also has a product "Xtreme Pug" with code "X_PUG"
         And this product is named "Ekstremalny Mops" in the "Polish (Poland)" locale
+        And this product is named "Ekstremalny Mops" in the "Polish" locale
         And I am logged in as an administrator
 
     @ui @api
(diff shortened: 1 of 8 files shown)


features/product/managing_products/sorting_products.feature
```diff
@@ -63,3 +64,22 @@ Feature: Sorting listed products
...
         Then I should see 3 products in the list
         And the first product on the list should have name "Szałowy Mops"
+
+    @ui
+    Scenario: Missing translations are sorted as first when sorting by name ascending
+        When I change my locale to "Polish"
+        And I browse products
+        And I sort the products ascending by name
+        Then I should see 3 products in the list
+        And the first product on the list should have "data-test-missing-translation-paragraph" data attribute
```
lchrusciel, line 74: One of the rules of rightful BDD is to omit implementation details in the steps. I wouldn't expect anyone non-tech to say a sentence like this in real life. Also, such a step force us to interpret it only in an HTML context. 
```suggestion
        And the first product on the list shouldn't have a name
```

features/product/managing_products/sorting_products.feature
```diff
@@ -63,3 +64,22 @@ Feature: Sorting listed products
...
+
+    @ui
+    Scenario: Missing translation are sorted as last when sorting by name descending
+        When I change my locale to "Polish"
+        And I browse products
+        And the products are already sorted ascending by name
+        And I sort the products descending by name
+        Then I should see 3 products in the list
+        And the first product on the list should have name "Ekstremalny Mops"
+        And the last product on the list should have "data-test-missing-translation-paragraph" data attribute
```
lchrusciel, line 85:
```suggestion
        And the last product on the list shouldn't have a name
```

features/product/managing_products/sorting_products.feature
jakubtobiasz: I've implemented you suggestion but due to some reason 3 actions are failing and they seem to not be related to my changes at all. Can you take a look? Maybe something like this happened earlier? The error messages doesn't tell me too much 🤷🏻‍♀️.

lchrusciel (approved):

Roshyo (approved):

lchrusciel:
Thanks, Jakub! :tada:

loevgaard:
This is awesome 🎉

Issue #13134, opened 2021-09-22
Short locale notation altering search to zero-results.


oleg-andreyev:
**Sylius version affected**: 1.10.2

**Description**  
- Go to "Admin Panel"
- Go to "My Account"
- Change Language Preferences and select "English" and not "English (United States)" or something else.
- Save
- Got to "Products" in the admin panel
- Try to search by "name"
- Empty result

This is because all transactions are stored using full local notation en_US, lv_LV and etc.

```sql
SELECT
  DISTINCT id_4
FROM
  (
    SELECT
      DISTINCT id_4,
      code_0
    FROM
      (
        SELECT
          s0_.code AS code_0,
          s0_.created_at AS created_at_1,
          s0_.updated_at AS updated_at_2,
          s0_.enabled AS enabled_3,
          s0_.id AS id_4,
          s0_.variant_selection_method AS variant_selection_method_5,
          s0_.average_rating AS average_rating_6,
          s1_.name AS name_7,
          s1_.slug AS slug_8,
          s1_.description AS description_9,
          s1_.meta_keywords AS meta_keywords_10,
          s1_.meta_description AS meta_description_11,
          s1_.id AS id_12,
          s1_.short_description AS short_description_13,
          s1_.locale AS locale_14
        FROM
          sylius_product s0_
          INNER JOIN sylius_product_translation s1_ ON s0_.id = s1_.translatable_id
          AND (s1_.locale = ?)
        WHERE
          s0_.code LIKE '%white%'
          OR s1_.name LIKE '%white%'
      ) dctrn_result_inner
    ORDER BY
      code_0 ASC
  ) dctrn_result
LIMIT
  10
```

here `locale` argument is "en" not "en_US"

**Possible Solution**  
- Make sure that customer cannot select "short" (en, lv and etc.) locale in "Language Preferences"

vvasiloi:
How about that: an admin should see all resources in all locales, regardless of the account language, which should be mostly used for translating the interface.

diimpp:
This problem appeared through the app multiple times (ajax requests for autocomplete, taxon tree) and was solved by introducing search without specifying locale code.

I guess it's time for grid search filter fix now :)

Though my information may be out of date.
https://github.com/Sylius/Sylius/blob/1cb2981a4ccd89af9928822d8cc86547ed91f03c/src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/product.yml#L21

molabs:
[...]

PR #13737, opened 2022-03-08
[Admin] Fix undefined labels in taxon autocomplete

ernestWarwas:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Before:
image

After:
image


src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/taxon.yml
```diff
@@ -50,7 +50,7 @@ sylius_admin_ajax_taxon_by_name_phrase:
                 method: findByNamePart
                 arguments:
                     phrase: $phrase
-                    locale: null
+                    locale: expr:service('sylius.context.locale').getLocaleCode()
```
ernestWarwas, line 53: I'm not sure about this change. In my opinion, this causes a problem with autocompletes, but I don't know why it was set to null?

vvasiloi: When it's null, it searches though the translations in all locales, when a locale is passed, it will only search though the translations in that locale.
https://github.com/Sylius/Sylius/blob/f42d92602683a41dabe4fd5514f6882aa4e48908/src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php#L93-L102
https://github.com/Sylius/Sylius/blob/f42d92602683a41dabe4fd5514f6882aa4e48908/src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php#L109-L124

vvasiloi: In other words, this will limit the search to the admin's locale, which will create another issue.
The way I would approach it, is by using the translation which matched the query to display the name / fullname.

src/Sylius/Bundle/CoreBundle/Form/Type/CatalogPromotionScope/ForTaxonsScopeConfigurationType.php
```diff
@@ -31,7 +31,7 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             'label' => 'sylius.ui.taxons',
             'multiple' => true,
             'required' => false,
-            'choice_name' => 'name',
+            'choice_name' => 'fullname',
```
vvasiloi, line 34: This can be removed, because that's the default value in `Sylius\Bundle\TaxonomyBundle\Form\Type\TaxonAutocompleteChoiceType`

src/Sylius/Bundle/AdminBundle/Resources/config/routing/ajax/taxon.yml
ernestWarwas: Hmm, I did not specify my question, sorry 😞 I should ask why limiting the search to the admin's locale will create another issue? Now we are searching for all translations for eg. `LIKE "%t%"` and if found it hydrates it to your locale eg. `en_US` and this causes a problem because we found two additional categories which are not in your locale and it is not translating to `en_US` because we don't have this translation (that's why the name was undefined) and I think limiting to admin’s locale is ok. But if I’m wrong please correct me 😃

vvasiloi: The question is: why limit it to the admin's locale when you can search though all translations?
It's an unnecessary limitation and I'm sure users are going to complain about it just like they did regarding other parts of the admin panel: https://github.com/Sylius/Sylius/issues/13296, https://github.com/Sylius/Sylius/issues/13134, https://github.com/Sylius/Sylius/issues/12145
The admin locale should be used mostly for translating the UI, not to restrict the content.

vvasiloi: To emphasize it even more, consider that the admin can choose **ANY** locale and if it's not one of the enabled locale in the store, the the admin won't see no results.

vvasiloi: To help you, here are two ideas of how this could be solved:
1. Add `Autocomplete` group, to Taxon's `translations` property and to the relevant properties in TaxonTranslation, in their serialization config. Enabled the `sylius-auto-complete.js` to access nested attributes, so you could pass a path like `translations[0].fullname` to `choice_name`  option of `TaxonAutocompleteChoiceType`.
2. Loop though the result set and set the fallback locale for each taxon to the locale of it's first translation. This can be done either directly in the repository method, or using the index event.

ernestWarwas: Ok, I think I understand this, thanks 😃

Rafikooo (approved):

AdamKasp:
it looks fine, but I have no idea if it works :( we need some tests for it :)

vvasiloi:
Maybe do that in the Core Bundle instead of TaxonomyBundle? 🤔

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
```diff
@@ -92,13 +92,20 @@ public function findRootNodes(): array
...
+        $results = $this->createTranslationBasedQueryBuilder($locale)
             ->andWhere('translation.name LIKE :name')
             ->setParameter('name', '%' . $phrase . '%')
             ->setMaxResults($limit)
             ->getQuery()
             ->getResult()
         ;
+
+        foreach ($results as $result) {
+            $result->setFallbackLocale(array_key_first($result->getTranslations()->toArray()));
```
Ferror, line 105: And since this is a collection, there is no option to get the first value right away? the function `first ()` should get the hang of the subject. The second point is casting the collection to an array and extracting one element is disputable. It would add a kinda big overhead to the search especially when it is foreached.

Ferror (changes requested):

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
vvasiloi: The `key` method can be used: `$result->getTranslations()->key()`, but this will modify the array pointer.
However, I don't understand where does it cause a _big overhead_, can you elaborate?
The `->toArray()` call just returns the underlying array structure and `array_key_first` is the most efficient way to do the job without modifying the pointer, according to the [RFC ](https://wiki.php.net/rfc/array_key_first_last_index) and the [docs](https://www.php.net/manual/en/function.array-key-first.php).

ernestWarwas: I also don't see any big overhead here 🤔

Ferror: I just created a super answer and removed it ehh... Long story short.

The `$result->getTranslations()->toArray()` is costly when it comes to memory, because under `$result->getTranslations()` is Doctrine Collection and as far as I remember the `toArray()` function will create copy of this array and don't get me wrong there is no fault in `array_key_first` function, yet `$result->getTranslations()->first()` will be simpler and faster approach. Yet I was wrong, because then I looked at the https://github.com/doctrine/collections/blob/2.0.x/lib/Doctrine/Common/Collections/ArrayCollection.php and see that first and key actually are modifying the array.

Ferror (approved):

src/Sylius/Bundle/TaxonomyBundle/Doctrine/ORM/TaxonRepository.php
vvasiloi: Since you looked at the implementation, take a look at `toArray()` method too, because it just returns the underlying array.
Yes, that will perform a shallow copy of the **one element array**, which is far from being a **big overhead**.

https://github.com/doctrine/collections/blob/c1ec9823f9250fefb274cccf62970b5c167e6bc7/lib/Doctrine/Common/Collections/ArrayCollection.php#L69-L72

tests/Controller/AdminTaxonAjaxTest.php
```diff
@@ -0,0 +1,103 @@
...
+        $this->authenticateAdminUser();
+
+        $this->client->request('GET', '/admin/ajax/taxons/search?phrase=men');
+
+        $response = $this->client->getResponse();
+
+        $taxons = json_decode($response->getContent());
+
+        $this->assertEquals('Men T-Shirts', $taxons[0]->name);
+        $this->assertEquals('Women T-Shirts', $taxons[1]->name);
```
GSadee, line 78: Why didn't you assert here an entire response similar to other tests?

vvasiloi (approved):

GSadee (approved):

GSadee:
Thanks, Ernest! :1st_place_medal:

PR #13738, opened 2022-03-08
[Docs] Synchronous messenger transport

Rafikooo:
| Q               | A
| --------------- | -----
| Branch?         | 1.11 
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| License         | MIT

Synchronous messenger transport explained.

### DIFF
--- a/docs/book/products/catalog_promotions.rst
+++ b/docs/book/products/catalog_promotions.rst
@@ -313,6 +313,28 @@ After changes in CatalogPromotion, we dispatch proper message with delay calcula
     To enable asynchronous Catalog Promotion, remember about running messenger consumer in a separate process, use the command: ``php bin/console messenger:consume main``
     For more information check official `Symfony docs <https://symfony.com/doc/current/messenger.html#consuming-messages-running-the-worker>`_
 
+Catalog promotion synchronicity
+-------------------------------
+
+TL;DR: There is no reason to use synchronous processing unless you have only a few products and manually activate and deactivate catalog promotions.
+
+Since the `main` transport is asynchronous by default, in order to use synchronous processing, you need to override configuration by creating the `config/packages/messenger.yaml` file and pasting the following configuration:
+
+.. code-block:: yaml
+
+    framework:
+        messenger:
+            transports:
+                main: 'sync://'
+
+
+Synchronous processing means that, after submitting a catalog promotion form, one must wait for a server response until processing is complete.
+This may cause a worse user experience if there are other catalog promotions as it will trigger their recalculation as well.
+
+It is not possible to schedule the start of a catalog promotion on a given start date, it has to be started manually after that start date.
+It can be done by creating or editing any promotion.
+The same happens with the end date, so if the catalog promotion should have expired, it will not become inactive automatically, inactivation has to be triggered manually.
+
 How the Catalog Promotions are applied?
 ---------------------------------------
 


Ferror (approved):
I like it!

GSadee (approved):

docs/book/products/catalog_promotions.rst
```diff
@@ -313,6 +313,27 @@ After changes in CatalogPromotion, we dispatch proper message with delay calcula
...
+    MESSENGER_TRANSPORT_DSN=sync://
+
+Details:
+''''''''
+
+- synchronous processing in opposite to asynchronous means that after confirmation the administrator will wait for a server response until processing is complete
+- this may cause a bad user experience if there are other promotions as it causes them to be recalculated as well
+- it is not possible to create a scheduled catalog promotion that starts automatically when the start date is reached, it has to be manually started
+- if there is a catalog promotion that is within the timeframe but has not been applied for the reason described above, it will be triggered when any action related to creating or editing any promotion is performed
+- if the validity period of the promotion has expired, the promotion will not be automatically inactive
```
CoderMaggie, line 334:
```suggestion
- The same happens with the end date, so if the catalog promotion should have expired, it will not become inactive automatically, inactivation has to be triggered manually.
```

docs/book/products/catalog_promotions.rst
```diff
@@ -313,6 +313,27 @@ After changes in CatalogPromotion, we dispatch proper message with delay calcula
...
+- if there is an active catalog promotion that does not fit within the timeframe for the reason described above, it will be disabled after editing another catalog promotion, but its state remains active. To update the state as well, this particular catalog promotion has to be edited
```
CoderMaggie, line 335:
```suggestion
```

docs/book/products/catalog_promotions.rst
```diff
@@ -313,6 +313,24 @@ After changes in CatalogPromotion, we dispatch proper message with delay calcula
...
+-------------------------------
+
+TL;DR: There is no reason to use synchronous processing unless you have only a few products and manually activate and deactivate catalog promotions.
+
+Since the transport is asynchronous by default, in order to use synchronous processing, you need to update your ``.env`` configuration file:
+
+.. code-block:: bash
+
+    MESSENGER_TRANSPORT_DSN=sync://
+
```
Roshyo, line 326: :thinking: I'm not sure we should recommend such change since it is a global config. Not just for this behaviour. This could clearly change some other processes.

If we want to propose a way to disable async for catalog promotion imo we should propose a way to do it just for this, not for the global app. Or am I mistaken ?

Rafikooo: I think you're right. I'm not well experienced with Symfony messenger but I did some digging and probably a better solution would be to reconfigure `messenger.yaml`.
```yaml
framework:
    messenger:
        # Uncomment this (and the failed transport below) to send failed messages to this transport for later handling.

        transports:
        # https://symfony.com/doc/current/messenger.html#transport-configuration
#            main: '%env(MESSENGER_TRANSPORT_DSN)%'
            main: 'sync://'
```
At this moment `main` transport is used only for catalog promotion related routings, therefore, it would not collide if we would introduce another async (sync) feature.

What do you think?

docs/book/products/catalog_promotions.rst
```diff
@@ -313,6 +313,28 @@ After changes in CatalogPromotion, we dispatch proper message with delay calcula
...
+Since the `main` transport is asynchronous by default, in order to use synchronous processing, you need to override configuration by creating the `config/packages/messenger.yaml` file and pasting the following configuration:
+
+.. code-block:: yaml
+
+    framework:
+        messenger:
+            transports:
+                main: 'sync://'
+
+
```
GSadee, line 330:
```suggestion
```

GSadee (approved):

GSadee:
Thank you, Rafał! :tada:

docs/book/products/catalog_promotions.rst
Rafikooo: https://github.com/Sylius/Sylius/pull/13751

PR #13743, opened 2022-03-09
[WIP][SymfonyTranslator] Decorate Theme Bundle translator to avoid memory overloading

SirDomin:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

builds are failing and I'm trying to do anything, the problem is with memory leak when using js beats (problem is with Sylius theme bundle, the proposed solution doesn't work, the place must be on a different service)

# Fix Attempts

### Removing Theme Bundles related tests or just Theme Bundle from test execution
Due to dependency on Theme Bundle, other tests were failing.

### Possible SymfonyTranslatorResourceProvider memory leak

The function does `initializeIfNeeded` initializes variables every time one of the functions is executed. The function appends resources and locales each time it is executed, meaning that if I have 100 translation files and execute it 10 times I would have 10 000 translations and locales in session. It's a BUG no matter this case.

### Possible BC at chrome driver

The GitHub actions were updated a few days ago where one of the dependencies - chrome driver was updated from 98 to 99 version https://github.com/actions/virtual-environments

* tried https://github.com/browser-actions/setup-chrome - same output
* tried ubuntu-18.04 image
* health check on chrome

### placehold.it images does not load

The main reason the pages were unable to load is the `placehold.it` service that was badly maintained and each image request for ex 50x50 png was taking about 30-45 seconds - therefore the Client Timeout exception appeared. Yet the error message was not available from the beginning of our work. We had to fork and contribute to the: https://gitlab.com/Ferror/chrome-mink-driver/-/tree/bugfix-debug to remove.

```
PHP Warning:  Trying to access array offset on the value of type null in 
/home/runner/work/Sylius/Sylius/vendor/dmore/chrome-mink-driver/src/DevToolsConnection.php on line 108
```

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Provider/SymfonyTranslatorResourceProvider.php
+++ b/src/Sylius/Bundle/CoreBundle/Provider/SymfonyTranslatorResourceProvider.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\Provider;
+
+use Sylius\Bundle\ThemeBundle\Translation\Provider\Resource\TranslatorResourceProviderInterface;
+use Sylius\Bundle\ThemeBundle\Translation\Resource\TranslationResource;
+use Sylius\Bundle\ThemeBundle\Translation\Resource\TranslationResourceInterface;
+
+/** @inner */
+class SymfonyTranslatorResourceProvider implements TranslatorResourceProviderInterface
+{
+    /** @var TranslationResourceInterface[] */
+    private array $resources = [];
+
+    private array $resourcesLocales = [];
+
+    private array $filepaths;
+
+    public function __construct(array $filepaths = [])
+    {
+        $this->filepaths = $filepaths;
+    }
+
+    public function getResources(): array
+    {
+        $this->initialize();
+
+        return $this->resources;
+    }
+
+    public function getResourcesLocales(): array
+    {
+        $this->initialize();
+
+        return $this->resourcesLocales;
+    }
+
+    /**
+     * 1. The function does not initialize if needed. It initializes everytime it is executed.
+     * 2. The function appends resources and locales each time it is executed, meaning that if I have 100 translation
+     * files and execute it
+     */
+    private function initialize(): void
+    {
+        $this->resources = [];
+        $this->resourcesLocales = [];
+
+        foreach ($this->filepaths as $key => $filepath) {
+            $resource = new TranslationResource($filepath);
+
+            $this->resources[] = $resource;
+            $this->resourcesLocales[] = $resource->getLocale();
+        }
+
+        $this->resourcesLocales = array_unique($this->resourcesLocales);
+        $this->filepaths = [];
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/CoreBundle/Provider/SymfonyTranslatorResourceProvider.php
```diff
@@ -0,0 +1,58 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\Provider;
+
+use Sylius\Bundle\ThemeBundle\Translation\Provider\Resource\TranslatorResourceProviderInterface;
+use Sylius\Bundle\ThemeBundle\Translation\Resource\TranslationResource;
+use Sylius\Bundle\ThemeBundle\Translation\Resource\TranslationResourceInterface;
+
+/** @inner */
```
GSadee, line 11:
```suggestion
/** @internal */
```
according to https://docs.sylius.com/en/1.9/book/organization/backward-compatibility-promise.html#code-covered

src/Sylius/Bundle/CoreBundle/Provider/SymfonyTranslatorResourceProvider.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $this->decorated = $decorated;
+    }
+
+    public function getResources(): array
+    {
+        $this->initialize();
+
+        return $this->resources;
+
```
AdamKasp, line 33:
```suggestion
```

src/Sylius/Bundle/CoreBundle/Provider/SymfonyTranslatorResourceProvider.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
```
AdamKasp, line 2: missing doc block

src/Sylius/Bundle/CoreBundle/Provider/SymfonyTranslatorResourceProvider.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        $this->initialize();
+
+        return $this->resources;
+
+    }
+
+    public function getResourcesLocales(): array
+    {
+
```
AdamKasp, line 38:
```suggestion
```

src/Sylius/Bundle/CoreBundle/Provider/SymfonyTranslatorResourceProvider.php
```diff
@@ -0,0 +1,61 @@
...
+        return $this->resourcesLocales;
+    }
+
+    /**
+     * 1. The function does not initialize if needed. It initializes everytime it is executed.
+     * 2. The function appends resources and locales each time it is executed, meaning that if I have 100 translation
+     * files and execute it
+     */
+    private function initialize(): void
```
Ferror, line 49: @SirDomin the phpdoc would be the thing you would like to put in the Description. Second, there is no fix ;p The SymfonyTranslatorResourceProvider is 1to1 with the ThemeBundle one.

Ferror (changes requested):

src/Sylius/Bundle/CoreBundle/Provider/SymfonyTranslatorResourceProvider.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $this->resources = [];
```
stloyd, line 47: What about adding if+return when both variable has some content? Cause still looping over & over is not very efficient

jakubtobiasz:
Our CI is working, so I guess this PR is obsolete.

PR #13746, opened 2022-03-09
[Maintenance]Remove auto-scripts from composer.json file

Ferror:
In my opinion, using auto-scripts in the composer.json file is bad practice. Why?

1. **I lose my ability to define where and when commands are executed.** I would like to have all power of when, where, and what is executed. The auto-magic is the approach that may break installation or execution.

2. The Docker build that I'm working on cannot handle the auto-scripts approach, because of the **memory limitations**.

3. Default `cache:clear` command is very **costly when it comes to resources** and there are not many situations when you actually trigger that on dev or test environment. And you definitely don't want to trigger that default command on production without proper flags.

Maybe I will find more arguments, but for no that's it. Tell me why should we keep the composer auto-scripts.

| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -311,17 +311,5 @@
         "classmap": [
             "src/Kernel.php"
         ]
-    },
-    "scripts": {
-        "post-install-cmd": [
-            "@auto-scripts"
-        ],
-        "post-update-cmd": [
-            "@auto-scripts"
-        ],
-        "auto-scripts": {
-            "cache:clear": "symfony-cmd",
-            "assets:install %PUBLIC_DIR%": "symfony-cmd"
-        }
     }
 }


vvasiloi:
You can just run composer commands with `--no-scripts` if it's a problem.

lchrusciel:
I don't have a strong opinion about it either way. Symfony still uses it: https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/5.4/manifest.json#L10-L13 and therefore I would consider it as safe default. When doing prod deployment we're doing `--no-scripts` anyway and Docker may be the same.

An old article still "recommends" using of it": http://fabien.potencier.org/symfony4-workflow-automation.html

If you are just testing it locally would you need to call this script anyway? 

I'm fine with merging, I'm fine with leaving it as it is. The build is red anyway for now. Does anyone from @Sylius/core-team have a specific opinion?

vvasiloi:
I'm against it. This is the default Symfony workflow automation since Symfony 2 and most developers are familiar with it and expect it to work this way. I don't see a benefit in removing it.

PR #13747, opened 2022-03-09
[WIP] Theme removal

lchrusciel:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

### DIFF
--- a/config/bundles.php
+++ b/config/bundles.php
@@ -43,7 +43,7 @@ return [
     Doctrine\Bundle\MigrationsBundle\DoctrineMigrationsBundle::class => ['all' => true],
     Sylius\Bundle\FixturesBundle\SyliusFixturesBundle::class => ['all' => true],
     Sylius\Bundle\PayumBundle\SyliusPayumBundle::class => ['all' => true],
-    Sylius\Bundle\ThemeBundle\SyliusThemeBundle::class => ['all' => true],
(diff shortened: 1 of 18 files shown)


config/bundles.php
```diff
@@ -43,7 +43,7 @@
     Doctrine\Bundle\MigrationsBundle\DoctrineMigrationsBundle::class => ['all' => true],
     Sylius\Bundle\FixturesBundle\SyliusFixturesBundle::class => ['all' => true],
     Sylius\Bundle\PayumBundle\SyliusPayumBundle::class => ['all' => true],
-    Sylius\Bundle\ThemeBundle\SyliusThemeBundle::class => ['all' => true],
+//    Sylius\Bundle\ThemeBundle\SyliusThemeBundle::class => ['all' => true],
```
Ferror, line 46: If I understand it correctly we should leave the SyliusBundle active, but not on the `test` environment.

```suggestion
    Sylius\Bundle\ThemeBundle\SyliusThemeBundle::class => ['dev' => true, 'prod' => true, 'test' => false],
```

Ferror: default is false, yet i would like to highlight this decison

PR #13750, opened 2022-03-10
[Admin][Shop] placehold.it replaced to local placeholders

ernestWarwas:
| Q               | A
| --------------- | -----
| Branch?         | 1.9
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

According to problems with uptime with placeholder.com we decided to replace it with local placeholders.

https://www.saashub.com/placeholder-com-status

PR #13753, opened 2022-03-10
[UI] Replacing the placeholder avatar with an icon

mamazu:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | no (but it could improve performance)
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | -
| License         | MIT

# Replacing the default user picture from placeholder with an icon
**Why?** First of all if the placeholder api is slow-ish. This slows down the Sylius site. Yes it's a very weird bug believe me, and try to explain that to a user of the software. :D

Secondly, since Sylius is very popular we are putting on more stress on the placeholder guys than there actually has to be and the icon doesn't look bad at all.

## Before
image

## After
image


vvasiloi (approved):
LGTM. Thanks @mamazu!

mamazu:
Thanks. Is it okay to be a MR to the master branch or should I also target older versions as well?

vvasiloi:
I see no reason to not have it in 1.10 and 1.11 too.

mamazu:
I have rebased it now.

mamazu:
Appearently someone was quicker than me. #13750

But I still think that we should replace the placeholder image in Sylius with a more good looking icon.

GSadee (commented):
The file `src/Sylius/Bundle/UiBundle/Resources/private/img/50x50.png` shouldn't be removed as it is also used in other template

mamazu:
I have re-added the image. But the build is still failing for some reason.

src/Sylius/Behat/Element/Admin/TopBarElement.php
```diff
@@ -24,7 +24,9 @@ public function hasAvatarInMainBar(string $avatarPath): bool
 
     public function hasDefaultAvatarInMainBar(): bool
     {
-        return strpos($this->getAvatarImagePath(), '/assets/admin/img/50x50.png') !== false;
+        $avatarElement = $this->getDocument()->find('css', 'i.ui.avatar.user.icon');
```
lruozzi9, line 27: Shouldn't this selector look more like:
```suggestion
        $avatarElement = $this->getDocument()->find('css', 'img.ui.avatar.image');
```
https://www.w3schools.com/cssref/sel_element_class.asp

mamazu:
I have fixed it. I accidentally removed the i tag that renders the icon while rebasing.

mamazu:
The failure exists in the frontend as far as I know which I didn't touch. Does someone have an idea how to fix that. I can't really get any helpful information from the Github Action Logs.

lchrusciel:
I've rerun all failed jobs, perhaps it will help

mamazu:
Yep it's green.

src/Sylius/Behat/Element/Admin/TopBarElement.php
lchrusciel: avatar is empty so we are checking `<i class="ui avatar icon user"></i>`, isn't that correct?

lchrusciel (approved):

lchrusciel:
Thank you, @mamazu! :tada:

src/Sylius/Behat/Element/Admin/TopBarElement.php
lruozzi9: Of course, when I wrote this comment the build were failing due to this line. The element was an image instead of an icon. I can't resolve this comment 😅

PR #13756, opened 2022-03-10
[GitHub Actions] Change PHP ini values + clear cache

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

This is another attempt to fix our builds, as they seem to be fixed on 1.11 and master branches, 1.10 is still failing. I change here the ini values of PHP and add clearing cache before some steps in our application workflow to be consistent with 1.11 where it had been changed some time ago.

### DIFF
--- a/.github/workflows/application.yml
+++ b/.github/workflows/application.yml
@@ -43,8 +43,8 @@ jobs:
                 uses: shivammathur/setup-php@v2
                 with:
                     php-version: "${{ matrix.php }}"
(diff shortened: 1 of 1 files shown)


Ferror (commented):
But why are we disabling opcache for CLI?

GSadee:
> But why are we disabling opcache for CLI?

It is a good question 😃, to be honest, I'm not sure if is needed, but after this PR, configs for 1.10 and 1.11 branches will be the same, this is what I wanted to achieve. 
I will open another PR to revert the change of opcache for CLI to see the status of build 🕵🏻

PR #13333, opened 2021-11-25
Replace data providers with doctrine extensions

paullla:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

PR #12766, opened 2021-07-01
[Locale]Available locales in channel by api

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         |  master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| License         | MIT

PR #13772, opened 2022-03-15
[ADR][API] Update ADR for providing locales after changing the approach

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | https://github.com/Sylius/Sylius/pull/12766 and https://github.com/Sylius/Sylius/pull/13333
| License         | MIT

### DIFF
--- a/adr/2021_07_05_api_providing_locales_available_in_active_channel.md
+++ b/adr/2021_07_05_api_providing_locales_available_in_active_channel.md
@@ -1,7 +1,8 @@
-# Using data provider for getting available locales in active channel for shop user
+# Using Doctrine Collection extension for getting available locales in active channel for shop user
 
-* Status: proposed
-* Date: 2021-07-05
+* Status: accepted
+* Propose date: 2021-07-05
+* Update date: 2022-03-15
 
 ## Context and Problem Statement
 Customer should have access only to locales available in their channel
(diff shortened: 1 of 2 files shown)


Rafikooo (approved):

Ferror:
I don't think we should update the ADRs when we change our decision, especially after a year. In that case, I think we should create another ADR with relation to the first one that explains why the previous approach appeared to be wrong.

The decision was made, people died, this cannot be undone :)

GSadee:
> I don't think we should update the ADRs when we change our decision, especially after a year. In that case, I think we should create another ADR with relation to the first one that explains why the previous approach appeared to be wrong.
> 
> The decision was made, people died, this cannot be undone :)

I wonder about doing it in this way, but I wanted to simplify our ADRs, especially that the changed ADR was in `proposed` state. However, I agree with your arguments and I've added the new ADR with reference to the old one.

lchrusciel:
as a contr-argument for changing ADRs with time - we have git in place. Therefore, we could replace them and base our current decisions only on existing ADRs at the moment. Of course, there is one more case - when we are transitioning from one approach to another, while two ADRs may exist: legacy and new one

AdamKasp (approved):

AdamKasp:

PR #13787, opened 2022-03-18
[ADR] Use raw data in commands and events

ernestWarwas:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/adr/2022_03_18_raw_data_in_command_and_events.md
+++ b/adr/2022_03_18_raw_data_in_command_and_events.md
@@ -1,0 +1,13 @@
+# Use raw data in commands and events
+
+* Status: accepted
+* Date: 2022-03-18
+
+## Context and Problem Statement
+
+While defining new commands or events we may define them with entities or any objects. Objects passed to commands or 
+events may desynchronize with the main object which can be causing problems with handlers that may operate on wrong data.
+
+## Decision Drivers <!-- optional -->
+
+* always operate on the actual state of the object
--- a/adr/2022_03_18_raw_data_in_command_and_events.md
+++ b/adr/2022_03_18_raw_data_in_command_and_events.md
@@ -5,9 +5,37 @@
 
 ## Context and Problem Statement
 
-While defining new commands or events we may define them with entities or any objects. Objects passed to commands or 
+While defining new commands or events we may define them with entities or any objects. Objects passed to commands or
 events may desynchronize with the main object which can be causing problems with handlers that may operate on wrong data.
+Passing objects to commands can ease us to implement new features and make it faster but this solution gives us plenty of
+problems with integrations with external systems or eg. operating with the actual state of an object especially when
+we are doing it asynchronously.
 
-## Decision Drivers <!-- optional -->
+## Decision Drivers
 
 * always operate on the actual state of the object
+* ease to integrate with external systems
+* to be more consistent with CQRS pattern
+
+## Considered Options
+
+### Passing raw data or objects to commands
+
+* good, because it can be ease for us to implement new features
+* bad, because it can be problematicaly when we are using async communication
+* bad, because it can be hard to integrate commands written like that with extenal systems
+
+### Passing only raw data or identifiers of the object(s) we want to change etc.
+
+* good, because there is no problem with desynchronizing data while async communication
+* good, because we can easy integrate with extenal systems
+* good, because it's more consistent with CQRS pattern
+* bad, because we need to always make queries to DB to get objects we want to modify (but we are caching queries from 
+DB when we are using doctrine as our ORM)
+
+## Decision outcome
+
+Chosen option: **"Passing only raw data or identifiers of the object(s) we want to change etc."**
+
+This option gives us plenty of advantages with only a small problem with always making queries when we want to work 
+with DB objects.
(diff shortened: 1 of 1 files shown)


adr/2022_03_18_raw_data_in_command_and_events.md
```diff
@@ -5,9 +5,37 @@
...
+
+* good, because there is no problem with desynchronizing data while async communication
+* good, because we can easy integrate with extenal systems
+* good, because it's more consistent with CQRS pattern
+* bad, because we need to always make queries to DB to get objects we want to modify (but we are caching queries from 
+DB when we are using doctrine as our ORM)
+
+## Decision outcome
+
+Chosen option: **"Passing only raw data or ids of object we want to change etc."**
```
stloyd, line 38:
```suggestion
Chosen option: **"Passing only raw data or identifiers of the object(s) we want to change etc."**
```

adr/2022_03_18_raw_data_in_command_and_events.md
```diff
@@ -0,0 +1,41 @@
...
+events may desynchronize with the main object which can be causing problems with handlers that may operate on wrong data.
+Passing objects to commands can ease us to implement new features and make it faster but this solution gives us plenty of
+problems with integrations with external systems or eg. operating with the actual state of an object especially when
+we are doing it asynchronously.
+
+## Decision Drivers
+
+* always operate on the actual state of the object
+* ease to integrate with external systems
+* to be more consistent with CQRS pattern
```
lchrusciel, line 18:
```suggestion
* be more consistent with CQRS pattern
```

adr/2022_03_18_raw_data_in_command_and_events.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Considered Options
+
+### Passing raw data or objects to commands
+
+* good, because it can be ease for us to implement new features
```
lchrusciel, line 24:
```suggestion
* good, because it can be easier for us to implement new features
```

adr/2022_03_18_raw_data_in_command_and_events.md
```diff
@@ -0,0 +1,41 @@
+# Use raw data in commands and events
+
+* Status: accepted
+* Date: 2022-03-18
+
+## Context and Problem Statement
+
+While defining new commands or events we may define them with entities or any objects. Objects passed to commands or
```
lchrusciel, line 8: This is more of my semantical taste, but this section seems to have mixed context and description of the proposed solution. I'm fine with it as it is, but what could we consider for this (or future) ADRs is to focus in this section on the problem itself, like:
```
While working with commands and handlers, we are always working with objects; therefore, these objects need to be present in handlers. We may either fetch them from the repository based on identifiers or pass them as a part of commands. The selected solution needs to take into account that our commands may be dispatched to external systems or possessed asynchronously
```

and then split the current content that is written here into a description of each possible choice

adr/2022_03_18_raw_data_in_command_and_events.md
```diff
@@ -0,0 +1,41 @@
...
+* always operate on the actual state of the object
+* ease to integrate with external systems
+* to be more consistent with CQRS pattern
+
+## Considered Options
+
+### Passing raw data or objects to commands
+
+* good, because it can be ease for us to implement new features
+* bad, because it can be problematicaly when we are using async communication
```
GSadee, line 25:
```suggestion
* bad, because it can be problematic when we are using async communication
```

adr/2022_03_18_raw_data_in_command_and_events.md
```diff
@@ -0,0 +1,41 @@
...
+* bad, because it can be hard to integrate commands written like that with extenal systems
```
GSadee, line 26:
```suggestion
* bad, because it can be hard to integrate commands written like that with external systems
```

adr/2022_03_18_raw_data_in_command_and_events.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Passing only raw data or identifiers of the object(s) we want to change etc.
+
+* good, because there is no problem with desynchronizing data while async communication
+* good, because we can easy integrate with extenal systems
```
GSadee, line 31:
```suggestion
* good, because we can easily integrate with external systems
```

GSadee (approved):

adr/2022_03_18_raw_data_in_command_and_events.md
```diff
@@ -0,0 +1,49 @@
...
+
+* Status: accepted
+* Date: 2022-03-18
+
+## Context and Problem Statement
+
+While working with commands and handlers, we are always working with objects; therefore, these objects need to be 
+present in handlers. We may either fetch them from the repository based on identifiers or pass them as a part of 
+commands. The selected solution needs to take into account that our commands may be dispatched to external systems or 
+possessed asynchronously
```
lchrusciel, line 11: 👻 
```suggestion
processed asynchronously
```

lchrusciel (approved):

GSadee (approved):

PR #13793, opened 2022-03-24
[WIP][API]Enable completing cart with free and not shippable item without addressing, shipping and payment

Ferror:
| Q               | A
| --------------- | -----
| Branch?         |  master
| Bugfix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | no
| License         | MIT

## Introduction
There are differences in the API and UI checkout processes. Let's say that we would like to save some time and API calls by handling multiple scenarios. One of them is the possibility to complete the cart with the free, not shippable item, without doing addressing, shipping selection, and payment selection. The solution is relatively simple because the only thing that was missing was the State Machine logic that was changed.

## The approach
Let's introduce the Sylius new concept whose working name would be Sylius Async Checkout - the checkout that transitions are always available and at the end the only thing we care if all Order requirements are matched. To make it happen we will use the Winzou Guards and forget about skipped shipping or payment steps as you don't need to tell API that you would like to skip some step... just SKIP IT!

## Additionally
* Created Specification abstraction with only `and` use-case

## Questions
1. What to do with the `@no-api` tags as they are scenarios that work only in UI


features/checkout/completing_cart_in_any_order.feature
```diff
@@ -0,0 +1,82 @@
...
+        And the store has a product "PHP Mascot" priced at "$15.00"
+        And the store has a product "Sylius T-Shirt" priced at "$0.00"
+        And the store has a product "Toyota Kata eBook" priced at "$50.00"
+        And this product does not require shipping
+        And the store has a product "Sylius eBook" priced at "$0.00"
+        And this product does not require shipping
+        And the store has "SHL" shipping method with "$5.00" fee
+        And the store allows paying "offline"
+        And I am a logged in customer
+        And I use "async" checkout type
```
AdamKasp, line 15: IMHO this is a technical detail, all checkouts made by API should work in any order of steps

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -812,10 +833,31 @@ public function theyShouldBeNotifiedThatTheyCannotAddressAnEmptyCart(): void
...
-        Assert::same($this->getCheckoutState(), OrderCheckoutStates::STATE_COMPLETED);
+        if ($this->sharedStorage->has('checkout_type') && $this->sharedStorage->get('checkout_type') === 'async') {
+            Assert::same($this->getCheckoutState(), AsynchronousOrderCheckoutStates::STATE_COMPLETED);
+        } else {
+            Assert::same($this->getCheckoutState(), OrderCheckoutStates::STATE_COMPLETED);
+        }
+    }
+
+    /**
+     * @Then /^the cart should NOT be placed$/
```
AdamKasp, line 848:
```suggestion
     * @Then the cart should NOT be placed
```

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -812,10 +833,31 @@ public function theyShouldBeNotifiedThatTheyCannotAddressAnEmptyCart(): void
...
+     */
+    public function theCartShouldNotBePlaced()
```
AdamKasp, line 850:
```suggestion
    public function theCartShouldNotBePlaced(): void
```

features/checkout/completing_cart_in_any_order.feature
```diff
@@ -0,0 +1,82 @@
...
+
+    @api @no-ui
+    Scenario: Completing the cart with free, not shippable item
+    (without doing addressing, shipping selection, and payment selection)
```
Zales0123, line 19: `(without doing...` is quite obvious as we... don't do them in the scenario 😄 these brackets could be skipped in this and other scenarios

features/checkout/completing_cart_in_any_order.feature
```diff
@@ -0,0 +1,82 @@
...
+        Given I have product "Sylius eBook" in the cart
+        And I specify the billing address as "Ankh Morpork", "Frost Alley", "90210", "United States" for "Jon Snow"
+        And I complete the addressing step
+        When I confirm my order
+        Then the cart should be placed
```
Zales0123, line 24: `the cart` -> `the order`? When it is placed, it _transforms_ from "cart" to "order" 💃

features/checkout/completing_cart_in_any_order.feature
```diff
@@ -0,0 +1,82 @@
...
+
+    @api @no-ui
+    Scenario: Completing the cart with free, shippable item
+    (without doing addressing and payment selection)
+        Given I have product "Sylius T-Shirt" in the cart
+        And I specify the billing address as "Ankh Morpork", "Frost Alley", "90210", "United States" for "Jon Snow"
```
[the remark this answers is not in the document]
Zales0123: Same for other scenarios

features/checkout/completing_cart_in_any_order.feature
```diff
@@ -0,0 +1,82 @@
...
+        And I specify the shipping address as "Ankh Morpork", "Frost Alley", "90210", "United States" for "Jon Snow"
+        And I complete the addressing step
+        And I completed the shipping step with "SHL" shipping method
```
Zales0123, line 60: `I completed` -> `I complete`

features/checkout/completing_cart_in_any_order.feature
```diff
@@ -0,0 +1,82 @@
...
+    (without doing addressing and shipping selection)
+        Given I have product "Toyota Kata eBook" in the cart
+        And I specify the billing address as "Ankh Morpork", "Frost Alley", "90210", "United States" for "Jon Snow"
+        And I complete the addressing step
+        And I choose "offline" payment method
+        When I confirm my order
+        Then the cart should be placed
+
+    @api @no-ui
+    Scenario: Completing cart with address, selecting payment and selecting shipment (in this order)
```
Zales0123, line 75: `(in this order)` let's remove it and not be a captain obvious 😄

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -541,10 +546,14 @@ public function addressShouldBeFilledAsShippingAddress(AddressInterface $address
...
-            [OrderCheckoutStates::STATE_PAYMENT_SKIPPED, OrderCheckoutStates::STATE_PAYMENT_SELECTED]
-        );
+        if ($this->sharedStorage->has('checkout_type') && $this->sharedStorage->get('checkout_type') === 'async') {
+            Assert::same($this->getCheckoutState(), AsynchronousOrderCheckoutStates::STATE_PAYMENT_SELECTED);
+        } else {
+            Assert::inArray(
+                $this->getCheckoutState(),
+                [OrderCheckoutStates::STATE_PAYMENT_SKIPPED, OrderCheckoutStates::STATE_PAYMENT_SELECTED]
+            );
+        }
```
Zales0123, line 556: According to [this comment](https://github.com/Sylius/Sylius/pull/13793/files#r835206224) we should only use the async checkout in the API, so this if statement is not needed 🖖

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
Zales0123: Is this step used anywhere?

src/Sylius/Component/Core/StateGuard/CompleteStepGuard.php
```diff
@@ -0,0 +1,71 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\StateGuard;
+
+use Sylius\Component\Core\Model\OrderInterface;
+
+class CompleteStepGuard
```
Zales0123, line 18: Could be final and made with an interface for easier decoration/overriding 🔥

src/Sylius/Component/Core/OrderCheckout/AsynchronousOrderCheckoutStates.php
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\OrderCheckout;
```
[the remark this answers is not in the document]
Zales0123: I would live the separated file with these constants, as even if steps are named the same, the concept is different (that's why we have a separated state machine). But I would not make it an interface but a final class with private constructor 💃

src/Sylius/Component/Core/StateGuard/CompleteStepGuard.php
Zales0123: Btw, this is named **Complete**StepGuard but contains only guard for the other transitions. I feel all of them deserves a tiny but specific separate service 🖖

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -654,18 +663,26 @@ public function iShouldHavePaymentMethodAvailableAsTheChoice(string $paymentMeth
...
      * @Then I should be on the checkout payment step
      */
     public function iShouldBeOnTheCheckoutPaymentStep(): void
     {
-        Assert::inArray(
-            $this->getCheckoutState(),
-            [OrderCheckoutStates::STATE_SHIPPING_SELECTED, OrderCheckoutStates::STATE_SHIPPING_SKIPPED]
-        );
+        if ($this->sharedStorage->has('checkout_type') && $this->sharedStorage->get('checkout_type') === 'async') {
+            Assert::same($this->getCheckoutState(), AsynchronousOrderCheckoutStates::STATE_SHIPPING_SELECTED);
```
SirDomin, line 679: If you want to stick to few conditions, we could make it with early return after Assertion to avoid "else" in code

src/Sylius/Component/Core/StateGuard/CompleteStepGuard.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        if ($order->isEmpty()) {
+            return false;
+        }
+
+        if ($order->isShippingRequired() === true) {
+            if ($order->getShippingAddress() === null) {
+                return false;
+            }
+        }
```
SirDomin, line 57: we could move that to private methods to know better what is going on, it looks like spaghetti made of ifs xD
f.eg if ($this->isShippingReady() && $this->isPaymentReady()...). I think it would improve DX

src/Sylius/Bundle/ApiBundle/spec/Modifier/OrderAddressModifierSpec.php
```diff
@@ -45,11 +45,11 @@ function it_handles_addressing_an_order_without_provided_shipping_address(
...
         $order->setShippingAddress(Argument::type(AddressInterface::class))->shouldBeCalled();
 
-        $stateMachineFactory->get($order, OrderCheckoutTransitions::GRAPH)->willReturn($stateMachine);
-        $stateMachine->can(OrderCheckoutTransitions::TRANSITION_ADDRESS)->willReturn(true);
-        $stateMachine->apply(OrderCheckoutTransitions::TRANSITION_ADDRESS)->shouldBeCalled();
+        $stateMachineFactory->get($order, AsynchronousOrderCheckoutTransitions::GRAPH)->willReturn($stateMachine);
+        $stateMachine->can(AsynchronousOrderCheckoutTransitions::TRANSITION_ADDRESS)->willReturn(true);
+        $stateMachine->apply(AsynchronousOrderCheckoutTransitions::TRANSITION_ADDRESS)->shouldBeCalled();
 
-        $this->modify($order, $billingAddress->getWrappedObject(), null, 'r2d2@droid.com');
```
SirDomin, line 52: shouldnt we keep both functionalities? Why was this email removed if the test was working on default checkout?

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -812,10 +833,31 @@ public function theyShouldBeNotifiedThatTheyCannotAddressAnEmptyCart(): void
...
+    /**
+     * @Then /^the cart should NOT be placed$/
+     */
+    public function theCartShouldNotBePlaced()
+    {
+        if ($this->sharedStorage->has('checkout_type') && $this->sharedStorage->get('checkout_type') === 'async') {
+            Assert::notSame(
+                $this->getCheckoutState(),
+                AsynchronousOrderCheckoutStates::STATE_COMPLETED,
+                'Expected not COMPLETED, got: ' . $this->getCheckoutState()
```
SirDomin, line 856:
```suggestion
                'Expected state different than COMPLETED, got: ' . $this->getCheckoutState()
```

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
```diff
@@ -1176,6 +1218,18 @@ public function iShouldSeeWithUnitPrice(ProductInterface $product, int $unitPric
         Assert::true($this->hasProductWithUnitPrice($product->getName(), $unitPrice));
     }
 
+    /**
+     * @Given I use :type checkout type
+     */
+    public function iUseSpecificCheckoutType(string $type): void
+    {
+        if ($type === 'async') {
```
SirDomin, line 1226: type is not secured in any way, f.eg Given I use XD checkout type would lead to normal checkout.

Ferror:
I kinda didn't specify what I need from the review - My bad XD

As you may notice the Async Checkout Transition and States are separated and Kinda duplicated from the perspective of CODE, yet it is not a bad practice in my perspective, because I avoid influence on UI.

There are a lot of tests API/UI for checkout and as you see I created some sort of Toggle for Async / Regular Checkout in tests, but changing all of them in one PR - would be a challenge. **Then Should I do it in this PR or maybe create another one?**  - Where implemented here Async Checkout would be the default for API Context. Please see that some of these tests will be useless for API as it has much simpler logic than UI. Therefore some of `@no-api` tags will take place.

src/Sylius/Component/Core/OrderCheckout/AsynchronousOrderCheckoutStates.php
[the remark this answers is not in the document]
AdamKasp: okay, I buy that they are different concepts (UI and API checkout) but in this case, these states and transitions shouldn't be in `api`  bundle?

src/Sylius/Bundle/ApiBundle/spec/Modifier/OrderAddressModifierSpec.php
Ferror: The email wes removed because the interface uses only 3 parameters. Look at the  definition XD

```php
interface OrderAddressModifierInterface
{
    public function modify(OrderInterface $order, AddressInterface $billingAddress, ?AddressInterface $shippingAddress = null): OrderInterface;
}
```

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
Ferror: Nope, removed

features/checkout/completing_cart_in_any_order.feature
Ferror: No can do. There is `I complete the shipping step with first shipping method` step that makes this change ambiguous, but I will use any existing step ex `I proceed with "SHL" shipping method`

src/Sylius/Component/Core/OrderCheckout/AsynchronousOrderCheckoutStates.php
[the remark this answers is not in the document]
Ferror: @Zales0123 
> But I would not make it an interface but a final class with a private constructor

Isn't that kinda make this class interface, but with cancer? Why would you like to create such things? We need ENUMS asap XD

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
Ferror: I am strongly against empty returns.

src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine/sylius_async_order_checkout.yml
```diff
@@ -0,0 +1,69 @@
...
+                to: shipping_selected
+            select_payment:
+                from: [cart, addressed, shipping_selected, payment_selected]
+                to: payment_selected
+            complete:
+                from: [cart, addressed, shipping_selected, payment_selected]
+                to: completed
+        callbacks:
+            guard:
+                sylius_complete_guard:
```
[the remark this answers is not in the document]
Ferror: That's a neat part, there is no order, but just for sake of perfection I'm gonna do this ;p

src/Sylius/Component/Core/StateGuard/CompleteStepGuard.php
```diff
@@ -0,0 +1,71 @@
...
+
+class CompleteStepGuard
+{
+    /**
+     * When address is required? It seems like always
+     * But should we in any instance block the state change? No
+     */
+    public function address(OrderInterface $order): bool
+    {
+        return false === $order->isEmpty();
```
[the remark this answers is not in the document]
Ferror: We can talk about it a little bit because I find the negation as not that good visible when it comes to this type of checks

src/Sylius/Component/Core/OrderCheckout/AsynchronousOrderCheckoutStates.php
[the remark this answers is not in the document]
AdamKasp: I think it is a good topic for some discussion :) at this moment I believe that all think that we are using only in API are placed in API Bundle. 

classes with a private contractor is a substitute of enums :D

src/Sylius/Component/Core/StateGuard/CompleteStepGuard.php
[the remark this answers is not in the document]
AdamKasp: I think it is a private preference, but mateusz's suggestion looks cleaner and is more consistent with standards in Sylius :)

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
AdamKasp: I am afraid that we are strongly against `else` statement :D but we really like early returns (even empty :P )

src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
```diff
@@ -263,5 +263,10 @@
         <service id="Sylius\Component\Core\Cart\Resolver\CreatedByGuestFlagResolverInterface" class="Sylius\Component\Core\Cart\Resolver\CreatedByGuestFlagResolver">
             <argument type="service" id="security.token_storage" />
         </service>
+
+        <service id="sylius.complete_state_guard" class="Sylius\Component\Core\StateGuard\CompleteStepGuard" />
```
AdamKasp, line 267:
```suggestion
        <service id="sylius.guard.complete_state" class="Sylius\Component\Core\StateGuard\CompleteStepGuard" />
```
imo it is more readable name : `sylius.type_of_service.detail`

src/Sylius/Component/Core/Order/Requirements/RequiredBillingAddressSpecification.php
```diff
@@ -0,0 +1,29 @@
...
+namespace Sylius\Component\Core\Order\Requirements;
+
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\Specification\CompositeSpecification;
+
+class RequiredBillingAddressSpecification extends CompositeSpecification
+{
+    public function isSatisfiedBy(object $candidate): bool
+    {
+        if (false === $candidate instanceof OrderInterface) {
```
AdamKasp, line 23:
```suggestion
        if ($candidate instanceof OrderInterface) {
```
isn't enough ?

src/Sylius/Component/Core/Specification/Specification.php
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\Specification;
+
+interface Specification
+{
+    public function isSatisfiedBy(object $candidate): bool;
+    public function and(Specification $specification): Specification;
```
AdamKasp, line 19:
```suggestion
    
    public function and(Specification $specification): Specification;
```

AdamKasp: for more readability and consistency with sylius

src/Sylius/Component/Core/StateGuard/CompleteStepGuard.php
```diff
@@ -0,0 +1,35 @@
...
+
+class CompleteStepGuard implements OrderGuardInterface
+{
+    public function isSatisfiedBy(OrderInterface $order): bool
+    {
+        $requirements = new RequiredNonEmptyCartSpecification();
+        $requirements
+            ->and(new RequiredBillingAddressSpecification())
+            ->and(new RequiredShippingSpecification())
+            ->and(new RequiredPaymentSpecification())
```
AdamKasp, line 30: Do we really need so many levels of abstractions?
- imo logic from this `specifications` can by directly in guard (they are 4 lines services)
- this guard should use the rest of the guard, it makes this code much easier to read.
- we should; avoid creating classes by `new` why don't you inject it like normal services?
- and definitely, we are missing scenarios, each service returns true or false so we have 8 situations but only one scenario (the similar situation for the rest of the services)
- it is checkout, so we should (MUST) test it as good as it is possible because if sth does not work it always is code without tests :)

but, i really like rost on my person so WDYT @lchrusciel @GSadee @Zales0123

src/Sylius/Component/Core/spec/Order/Requirements/RequiredBillingAddressSpecificationSpec.php
```diff
@@ -0,0 +1,33 @@
...
+class RequiredBillingAddressSpecificationSpec extends ObjectBehavior
+{
+    function it_is_not_satisfied_by_anonymous_object()
+    {
+        $this->isSatisfiedBy(new class(){})->shouldBe(false);
+    }
+
+    function it_is_satisfied_by_order(OrderInterface $order)
+    {
+        $order->getBillingAddress()->willReturn(new Address());
```
AdamKasp, line 29: can we make mock instead creating it by `new`/

src/Sylius/Component/Core/spec/Order/Requirements/RequiredBillingAddressSpecificationSpec.php
```diff
@@ -0,0 +1,33 @@
...
+use Sylius\Component\Core\Model\OrderInterface;
+
+class RequiredBillingAddressSpecificationSpec extends ObjectBehavior
+{
+    function it_is_not_satisfied_by_anonymous_object()
+    {
+        $this->isSatisfiedBy(new class(){})->shouldBe(false);
+    }
+
+    function it_is_satisfied_by_order(OrderInterface $order)
```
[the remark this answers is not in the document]
AdamKasp:
```suggestion
    function it_is_satisfied_by_addressed_order(OrderInterface $order)
```
and another scenario for order without address
IMO this scenario should be more descriptive 
and I believe all specs need a similar to refactor :)

src/Sylius/Component/Core/Order/Requirements/RequiredBillingAddressSpecification.php
Ferror: No. That changes the `instanceof` logic.

src/Sylius/Component/Core/spec/Order/Requirements/RequiredBillingAddressSpecificationSpec.php
Ferror: Yea, but what will it change?

src/Sylius/Component/Core/StateGuard/CompleteStepGuard.php
Ferror: Yup, we can discuss that. The Specification provides reusable business logic (Domain). Guard classes are basically the Framework or Infrastructure layer of the checkout system. I could create even more abstraction and provide Strategy over coupled Specification or not use Guards at all. The less neat part is that they are all in the Core namespace, but if I will move it somewhere else they will be coupling between components or bundles. Define best practices as I'm not aware of Sylius's approach :)


PS. We don't want to provide specifications via DI - it's gonna look terrible.

src/Sylius/Component/Core/Order/Requirements/RequiredBillingAddressSpecification.php
AdamKasp: true :D so
```suggestion
        if (!$candidate instanceof OrderInterface) {
```

src/Sylius/Component/Core/spec/Order/Requirements/RequiredBillingAddressSpecificationSpec.php
AdamKasp: We do not use the `new` statement in specs (except for a few exceptions). If spec can do sth for us Imo he should :D
in this case, for me, mock will provide information about what an `address` is

Ferror: Alright, if we mock then let's mock

src/Sylius/Component/Core/Order/Requirements/RequiredBillingAddressSpecification.php
Ferror: I personally don't like negation on instanceof - this is one of the PHP language-specific flaw.. but yea changed it for sake of codestyle

src/Sylius/Bundle/CoreBundle/Resources/config/services.xml
Ferror: sylius.state_guard.address/complete etc work for you?

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
Ferror: That's a good presentation/discussion topic `is else so bad that we use empty return?`

src/Sylius/Component/Core/OrderCheckout/AsynchronousOrderCheckoutStates.php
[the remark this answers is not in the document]
Ferror: That's a thing. Async Checkout is no API Specific concept. Is it used by API, but if you want you could use it also in UI. The fact is the tests will fail as it is a new concept and UI is strongly attached to the State Machine. We could refactor it later and say.. State Machine is very close to the presentation layer, but not core/domain as we thought.

src/Sylius/Component/Core/Order/Requirements/RequiredBillingAddressSpecification.php
vvasiloi: Even though `instanceof` has precedence over `!`, some devs prefer to add parentheses "just in case". I'm pretty sure even I do it sometimes. 🤷‍♂️

src/Sylius/Component/Core/spec/Order/Requirements/RequiredBillingAddressSpecificationSpec.php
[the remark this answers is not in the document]
Ferror: I changed it a little bit, and tests the matrix of logic

src/Sylius/Component/Core/OrderCheckout/AsynchronousOrderCheckoutStates.php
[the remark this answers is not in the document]
GSadee: My first thought was the same, that the new concept of checkout is specific for `API`, but now, I'm not so sure 🤔 After all, it's a bit of duplication, this is another concept for the same feature that we've already have in `Core`, so adding these services and configs to `Core`, could be confusing. For now, it is the concept that we are providing due to `API`, so I would vote for moving it there.

src/Sylius/Component/Core/StateGuard/CompleteStepGuard.php
GSadee: I wonder about the way of customizing that by the end application, if someone would like to change the specification, there is a need to define a new one and overwrite the proper guard. Probably it is simple enough 🤔 
I know that I'm not helpful enough, the concept looks nice, easy to adjust and probably quite easy to customize, but it is also a new one in Sylius, there will be a need for proper documentation of it.

GSadee (commented):
My additional thought is that after providing different state machine, we make it impossible to share the order between `UI` and `API` during checkout. It is also not possible now, so the problem (if it is a problem of course 😃) exists, but after this PR it will be much more difficult to achieve.

GSadee:
And @Ferror, squash some commits 🙏🏻, there is no need for so many of them, especially that there is a lot of commits with meaningless titles 😄

Ferror:
> And @Ferror, squash some commits 🙏🏻, there is no need for so many of them, especially that there is a lot of commits with meaningless titles 😄 

Sure. Its even required before merge as we dont use squash and merge approach. But i dont think Its bad that there are many commits as you may see my flow of thinking. Which is not possible when creating one or two commits especially for this kind of task kek

src/Sylius/Component/Core/OrderCheckout/AsynchronousOrderCheckoutStates.php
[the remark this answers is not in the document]
AdamKasp: `"SimpelOrderCheckout"`+1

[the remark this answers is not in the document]
lchrusciel: "SimpleOrderCheckout" of course

src/Sylius/Behat/Context/Api/Shop/CheckoutContext.php
lchrusciel: it is just a consequence of trying to keep functions flat, simple and reduce if statements. Also, it is a little bit easier to track flow paths (but that's subjective). I may have at least idea for a presentation :D

src/Sylius/Component/Core/Specification/Specification.php
```diff
@@ -0,0 +1,21 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Component\Core\Specification;
+
+interface Specification
+{
+    public function isSatisfiedBy(object $candidate): bool;
+
+    public function and(Specification $specification): Specification;
```
vvasiloi, line 20: The Specification should only require `isSatisfiedBy` and the composition should be handled by AndSpecification, OrSpecification, etc.
That's why you had to add the CompositeSpecification with an implementation of `and`.

Ferror:
@lchrusciel & @vvasiloi & @GSadee & @Zales0123 & @AdamKasp I created special thread to discuss the Specification pattern and how should we use and create it: https://github.com/Sylius/Sylius/discussions/13807

PR #13795, opened 2022-03-24
[ADR] Remove flush() from handlers.

ernestWarwas:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/adr/2022_03_24_remove_flush_from_handlers.md
+++ b/adr/2022_03_24_remove_flush_from_handlers.md
@@ -22,4 +22,4 @@ it gives us ability to avoid an inconsistent data and if something fails we have
 
 ## References <!-- optional -->
 
-* [Abstracting the Doctrine ORM flush](https://www.thinktocode.com/2019/03/28/abstracting-the-doctrine-orm-flush/)
+* [Transactions and Concurrency](https://www.doctrine-project.org/projects/doctrine-orm/en/latest/reference/transactions-and-concurrency.html)
--- a/adr/2022_03_24_remove_flush_from_handlers.md
+++ b/adr/2022_03_24_remove_flush_from_handlers.md
@@ -1,0 +1,25 @@
+# Remove flush from handlers
+
+* Status: accepted
+* Date: 2022-03-24
+
+## Context and Problem Statement
+
+Using flush() in command handlers can cause problems with inconsistent DB state. Doctrine always uses transactions but on
+flush() Doctrine will execute all the changes that have been tracked so far. Implicitly flushing will commit all the 
+changes queued up so far. Not just the entity you just persisted. This is also one of the reasons why it is dangerous to
+just call flush.
+
+## Decision Drivers
+
+* avoid inconsistent data in DB
+* flexibility to rollback changes
+
+## Decision Outcome
+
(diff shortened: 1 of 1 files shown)


adr/2022_03_24_remove_flush_from_handlers.md
```diff
@@ -0,0 +1,25 @@
...
+flush() Doctrine will execute all the changes that have been tracked so far. Implicitly flushing will commit all the 
+changes queued up so far. Not just the entity you just persisted. This is also one of the reasons why it is dangerous to
+just call flush.
+
+## Decision Drivers
+
+* avoid inconsistent data in DB
+* flexibility to rollback changes
+
+## Decision Outcome
```
lchrusciel, line 18:
```suggestion

## Considered Options

### Flushing in handlers 

Using flush() in command handlers can cause problems with inconsistent DB states if the end-user decides to adjust models' definitions. Implicitly flushing will commit all the changes queued up so far, which may lead to DB error or committing not consistent data. What is more, we would need an additional feature that would allow adjusting the entity's state before the transaction will close.

### Flushing outside of handlers with the manually triggered transaction

Manual setting up of transactions around handlers clearly defines their boundaries. Even tho this may still lead to DB error or committing not consistent data, setting boundaries outside of our code improves DX greatly, as an adjustment on objects may be made with decorator pattern outside of predefined handler.

## Decision Outcome
```

Probably you should also extract some good and bad points from my descriptions.

lchrusciel (approved):

adr/2022_03_24_remove_flush_from_handlers.md
```diff
@@ -0,0 +1,50 @@
...
+* Bad, because implicitly flushing will commit all the changes queued up so far
+* Bad, because we would need an additional feature
+
+### Flushing outside of handlers with the manually triggered transaction
+
+Manual setting up of transactions around handlers clearly defines their boundaries. Even tho this may still lead to DB 
+error or committing not consistent data, setting boundaries outside of our code improves DX greatly, as an adjustment 
+on objects may be made with decorator pattern outside of predefined handler.
+
+* Good, because it improve DX
```
GSadee, line 36:
```suggestion
* Good, because it improves DX
```

adr/2022_03_24_remove_flush_from_handlers.md
```diff
@@ -0,0 +1,50 @@
...
+* Good, because transactions around handlers clearly defines their boundaries
```
GSadee, line 37:
```suggestion
* Good, because transactions around handlers clearly define their boundaries
```

adr/2022_03_24_remove_flush_from_handlers.md
```diff
@@ -0,0 +1,50 @@
...
+* Good, because we have flexibility to rollback changes
+* Bad, because it may still lead to errors or commiting not consistent data
```
GSadee, line 39:
```suggestion
* Bad, because it may still lead to errors or committing not consistent data
```

GSadee (approved):

PR #13825, opened 2022-04-05
[API][Shop] Channel endpoint that returns current channel created

TheMilek:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

In order to improve DX we decided to create endpoint that will return customer's current channel with default locale and currency.

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/DataProvider/ChannelsCollectionDataProvider.php
+++ b/src/Sylius/Bundle/ApiBundle/DataProvider/ChannelsCollectionDataProvider.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\DataProvider;
+
+use ApiPlatform\Core\DataProvider\ContextAwareCollectionDataProviderInterface;
+use ApiPlatform\Core\DataProvider\RestrictedDataProviderInterface;
+use Sylius\Component\Channel\Context\ChannelContextInterface;
+use Sylius\Component\Core\Model\ChannelInterface;
+
+/** @experimental */
+final class ChannelsCollectionDataProvider implements ContextAwareCollectionDataProviderInterface, RestrictedDataProviderInterface
+{
+    public function __construct(private ChannelContextInterface $channelContext)
+    {}
+
+    public function getCollection(string $resourceClass, string $operationName = null, array $context = []): array
+    {
+        return [$this->channelContext->getChannel()];
+    }
+
+    public function supports(string $resourceClass, string $operationName = null, array $context = []): bool
+    {
+        return is_a($resourceClass, ChannelInterface::class, true) && $this->isShopGetCollectionOperation($context);
+    }
+
+    private function isShopGetCollectionOperation(array $context): bool
+    {
+        return isset($context['collection_operation_name']) && \str_starts_with($context['collection_operation_name'], 'shop');
+    }
+}
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Channel.xml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Channel.xml
@@ -25,6 +25,14 @@
                 </attribute>
             </collectionOperation>
 
+            <collectionOperation name="shop_get">
+                <attribute name="method">GET</attribute>
+                <attribute name="path">/shop/channels</attribute>
+                <attribute name="normalization_context">
+                    <attribute name="groups">shop:channel:read</attribute>
+                </attribute>
+            </collectionOperation>
+
             <collectionOperation name="admin_post">
                 <attribute name="method">POST</attribute>
                 <attribute name="path">/admin/channels</attribute>
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/services/data_providers.xml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/services/data_providers.xml
@@ -75,6 +75,11 @@
             <tag name="api_platform.collection_data_provider" priority="10" />
         </service>
 
+        <service id="Sylius\Bundle\ApiBundle\DataProvider\ChannelsCollectionDataProvider">
+            <argument type="service" id="sylius.context.channel" />
+            <tag name="api_platform.collection_data_provider" priority="10" />
+        </service>
+
         <service id="Sylius\Bundle\ApiBundle\DataProvider\OrderAdjustmentsSubresourceDataProvider">
             <argument type="service" id="sylius.repository.order" />
             <tag name="api_platform.subresource_data_provider" priority="10" />
--- a/src/Sylius/Behat/Context/Api/Shop/ChannelContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/ChannelContext.php
@@ -63,8 +63,8 @@ final class ChannelContext implements Context
             $this->responseChecker->getValue(
             $this->client->getLastResponse(),
             'baseCurrency'
-        )['code'],
-            $currencyCode
+            ),
+            sprintf('/api/v2/shop/currencies/%s', $currencyCode)
         );
     }
 }
--- a/src/Sylius/Bundle/ApiBundle/Resources/config/serialization/Channel.xml
+++ b/src/Sylius/Bundle/ApiBundle/Resources/config/serialization/Channel.xml
@@ -46,6 +46,7 @@
         <attribute name="defaultLocale">
             <group>admin:channel:read</group>
             <group>admin:channel:create</group>
+            <group>shop:channel:read</group>
         </attribute>
         <attribute name="defaultTaxZone">
             <group>admin:channel:read</group>
@@ -58,10 +59,12 @@
         <attribute name="currencies">
             <group>admin:channel:read</group>
             <group>admin:channel:create</group>
+            <group>shop:channel:read</group>
         </attribute>
         <attribute name="locales">
             <group>admin:channel:read</group>
             <group>admin:channel:create</group>
+            <group>shop:channel:read</group>
         </attribute>
         <attribute name="countries">
             <group>admin:channel:read</group>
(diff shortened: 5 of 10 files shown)


src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Channel.xml
```diff
@@ -25,6 +25,14 @@
                 </attribute>
             </collectionOperation>
 
+            <collectionOperation name="shop_get">
+                <attribute name="method">GET</attribute>
+                <attribute name="path">/shop/channels</attribute>
+                <attribute name="normalization_context">
+                    <attribute name="groups">shop:channels:read</attribute>
```
coldic3, line 32:
```suggestion
                    <attribute name="groups">shop:channel:read</attribute>
```

src/Sylius/Bundle/ApiBundle/spec/DataProvider/ChannelCollectionDataProviderSpec.php
```diff
@@ -0,0 +1,71 @@
...
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Channel\Context\ChannelContextInterface;
+use Sylius\Component\Core\Model\ChannelInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+final class ChannelCollectionDataProviderSpec extends ObjectBehavior
+{
+    function let(
+        ChannelContextInterface $channelContext
+    ): void
```
AdamKasp, line 25:
```suggestion
    function let(ChannelContextInterface $channelContext): void
```

src/Sylius/Bundle/ApiBundle/spec/DataProvider/ChannelCollectionDataProviderSpec.php
```diff
@@ -0,0 +1,71 @@
...
+                [
+                    'collection_operation_name' => 'shop_get',
+                ],
+            )
+            ->shouldReturn(true)
+        ;
+
+        $this
+            ->supports(
+                ChannelInterface::class,
```
AdamKasp, line 45: you can also check different resource

src/Sylius/Bundle/ApiBundle/spec/DataProvider/ChannelCollectionDataProviderSpec.php
```diff
@@ -0,0 +1,71 @@
...
+            )
+            ->shouldReturn(false)
+        ;
+    }
+
+    function it_returns_channel_collection(
+        ChannelContextInterface $channelContext,
+        ChannelInterface $channel
+    ): void
+    {
```
AdamKasp, line 59:
```suggestion
    ): void {
```

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Channel.xml
TheMilek: if I use the same name I will then serialise another 2 fields to different endpoint

coldic3: Oh, I see now ;d
If 2 extra fields in a different endpoint is a problem (maybe it's not a problem), I think it will be better to set `shop:channel:read` in the normalization context and add a second group only for these 2 fields.

AdamKasp: are these extra fields a problem? probably not (but I don't know which responses will be changed), ALL serialization groups are consistent, so we shouldn't make an exception for it

tests/Api/Shop/ChannelTest.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Sylius\Tests\Api\Shop;
+
+use Sylius\Tests\Api\JsonApiTestCase;
+use Symfony\Component\HttpFoundation\Response;
+
+final class ChannelTest extends JsonApiTestCase
+{
+    /** @test */
+    public function it_gets_collection_of_channel(): void
```
AdamKasp, line 22:
```suggestion
    public function it_gets_collection_of_channels(): void
```
but i am wondering if sth more precise will be better:
```suggestion
    public function it_gets_collection_with_single_default_channel(): void
```
or sth similar

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Channel.xml
lchrusciel: Ok, If I understand the issue correctly it is related to a lack of distinguishing between item and collection serialization. Idea was to split read serialization into two groups:
```xml
                    <attribute name="groups">shop:channel:read:list</attribute>
```
and
```xml
                    <attribute name="groups">shop:channel:read:item</attribute>
```

but for now, I would say that it is not a big deal to expose two additional fields (on the other hand reduction of them would be bc break). But split of serialization groups will be separate effort

TheMilek: To make it clear it's gonna be `defaultLocale`, `locales` and `currencies` additional fields. I decided to leave the same group name and serialise them anyway.

src/Sylius/Bundle/ApiBundle/DataProvider/ChannelCollectionDataProvider.php
```diff
@@ -0,0 +1,43 @@
...
+    {
+        return [$this->channelContext->getChannel()];
+    }
+
+    public function supports(string $resourceClass, string $operationName = null, array $context = []): bool
+    {
+        return is_a($resourceClass, ChannelInterface::class, true) && $this->isShopGetCollectionOperation($context);
+    }
+
+    private function isShopGetCollectionOperation(array $context): bool
```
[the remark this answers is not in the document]
TheMilek: In my opinion 3 places isn't that much to move it separate service, especially when it comes only to shop get collection operations, but maybe i'm wrong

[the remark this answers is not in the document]
coldic3: IMO even 2 places it's too much but it can be refactored in separate PR

UPGRADE-API-1.12.md
```diff
@@ -71,3 +71,31 @@ Here is how the response looks like:
    Wrong parameters otherwise cause empty array `[]` in response and correct parameters return `paymentMethods` available for your `payment`.
 
 1. The 2nd parameter `MetadataInterface` has been removed from `src/Sylius/Bundle/ApiBundle/CommandHandler/Account/ResetPasswordHandler` and replaced by `Sylius\Component\User\Security\PasswordUpdaterInterface` (previously 3rd parameter). From now on a token TTL value must be used instead as the 3rd parameter.
+
+1. The `GET` `api/v2/shop/channels/{code}` endpoint response has been changed and now contains also `defaultLocale`, `locales` and `currencies` fields.
```
AdamKasp, line 75: IMO it is redundant information, in `upgrade file` we should inform about potentially "dangerous" changes like BC breaks or bigger changes in logic.

src/Sylius/Bundle/ApiBundle/DataProvider/ChannelsCollectionDataProvider.php
```diff
@@ -0,0 +1,43 @@
...
+use Sylius\Component\Channel\Context\ChannelContextInterface;
+use Sylius\Component\Core\Model\ChannelInterface;
+
+/** @experimental */
+final class ChannelsCollectionDataProvider implements ContextAwareCollectionDataProviderInterface, RestrictedDataProviderInterface
+{
+    public function __construct(
+        private ChannelContextInterface $channelContext
+    ) {
+    }
```
AdamKasp, line 27:
```suggestion
    public function __construct(private ChannelContextInterface $channelContext)
     {}
```

AdamKasp (approved):

tests/Api/Responses/Expected/shop/get_channel_by_code.json
```diff
@@ -0,0 +1,18 @@
+{
+    "@context": "\/api\/v2\/contexts\/Channel",
+    "@id": "\/api\/v2\/shop\/channels\/WEB",
+    "@type": "Channel",
+    "baseCurrency": {
+        "@id": "\/api\/v2\/shop\/currencies\/USD",
+        "@type": "Currency",
+        "code": "USD"
+    },
```
lchrusciel, line 9: It looks wrong in comparison to the "defaultLocale". Why do we have IRI in one place and object in another? 
```suggestion
    "baseCurrency": "\/api\/v2\/shop\/currencies\/USD",
```

Also, I've requested IRI explicitly here

tests/Api/Responses/Expected/shop/get_channel_by_code.json
```diff
@@ -0,0 +1,14 @@
+{
+    "@context": "\/api\/v2\/contexts\/Channel",
```
AdamKasp, line 2: all responses should be suffixed with `response`, check different files in this catalog

tests/Api/Responses/Expected/shop/get_channel_collection.json
```diff
@@ -0,0 +1,21 @@
+{
+    "@context": "\/api\/v2\/contexts\/Channel",
```
AdamKasp, line 2: name of this file `get_channels_response` will be more consistent

tests/Api/Shop/ChannelTest.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Tests\Api\Shop;
+
+use Sylius\Tests\Api\JsonApiTestCase;
+use Symfony\Component\HttpFoundation\Response;
+
+final class ChannelTest extends JsonApiTestCase
```
AdamKasp, line 19:
```suggestion
final class ChannelsTest extends JsonApiTestCase
```
to be more consistent

src/Sylius/Bundle/ApiBundle/Resources/config/services/data_providers.xml
```diff
@@ -75,6 +75,11 @@
             <tag name="api_platform.collection_data_provider" priority="10" />
         </service>
 
+        <service id="Sylius\Bundle\ApiBundle\DataProvider\ChannelsCollectionDataProvider">
+            <argument type="service" id="Sylius\Component\Channel\Context\ChannelContextInterface" />
```
AdamKasp, line 79:
```suggestion
            <argument type="service" id="sylius.context.channel" />
```
if I see correctly we are using this name of service  (but i know that it is an alias for `"Sylius\Component\Channel\Context\ChannelContextInterface"`

GSadee (approved):

AdamKasp:

PR #13827, opened 2022-04-05
Throwing a not found when order is still a cart

### LINKED ISSUE #13826: [Admin] 500 when trying to access a cart in the order list
**Sylius version affected**
`>1.10.9` and probably more

**Description**  
When trying to access a cart in the order listing in the admin panel, Sylius throws a 500. There is no check in the code that checks if the order is actually completed in the `sylius_order_show` route.

**Steps to reproduce**  
* Create a cart in the shop and don't do the checkout
* Look into the database and take the id of the cart
* Go to /admin/orders/<id> with the id and see it's a 500 instead of the expected 404

**Possible Solution**  
Add an event listener to the route that checks if the order is completed and if not then throw a `NotFoundException`. I'll do a merge request for that.

mamazu:
If the order is in state cart the order show should not throw a 500.

| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #13826
| License         | MIT

When showing an order it should be checked before rendering the template if the order is actually completed or still in state cart.

### DIFF
--- a/docs/cookbook/entities/custom-model-accessible-for-channel-admin.rst
+++ b/docs/cookbook/entities/custom-model-accessible-for-channel-admin.rst
@@ -118,55 +118,28 @@ Remember to register ``App\Form\SupplierType`` for resource:
                     model: App\Entity\Supplier
        +            form: App\Form\Type\SupplierType
 
-1. Restrict access to the entity for the respective channel administrator:
+2. Restrict access to the entity for the respective channel administrator:
 --------------------------------------------------------------------------
 
 .. note::
 
     More information about using administrator roles (ACL/RBAC) can be found :doc:`here </book/customers/admin_user>`.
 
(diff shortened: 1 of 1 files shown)


Roshyo (approved):
I like this ! Thanks!

src/Sylius/Bundle/AdminBundle/EventListener/OrderShowListener.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\AdminBundle\EventListener;
+
+use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
```
stloyd, line 16: It's not used.

mamazu: Thanks for catching that. Fixed it.

mamazu:
Reopened as #13831. Because github decided to close it.

PR #13829, opened 2022-04-05
[WIP][Maintenance] Build fix 

TheMilek:
| Q               | A
| --------------- | -----
| Branch?         | master 
| License         | MIT

After Symfony have changed the minimal version on symfony/flex, we are encountering problems with builds, that we are trying to fix here.
We will probably have to also drop support for Symfony 5.2, because of security issues.
- https://symfony.com/blog/upgrade-flex-on-your-symfony-projects


composer.json
```diff
@@ -199,7 +199,7 @@
         "behat/behat": "^3.6.1",
         "behat/mink-selenium2-driver": "^1.4",
         "dmore/behat-chrome-extension": "^1.3",
-        "dmore/chrome-mink-driver": "^2.7",
```
Ferror, line 202: What happended in `2.8.1-beta1` version that we have to update it?

TheMilek: There was a `Child process error` that kept coming, as you can see now and I discovered that the solution for this problem is here: https://gitlab.com/DMore/chrome-mink-driver/-/merge_requests/111
That was just a try if it  fixes that and it does, but it isn't probably the best way.

lchrusciel (commented):
Can we test it already with flex 2.0? 

ref. https://packagist.org/packages/symfony/flex

Ferror:
> Can we test it already with flex 2.0?
> 
> ref. https://packagist.org/packages/symfony/flex

In my opinion, we should fix pipelines ASAP and then try to test it with the latest flex version :)

TheMilek:
Closed, as problem is fixed in another [PR](https://github.com/Sylius/Sylius/pull/13837)

PR #13831, opened 2022-04-06
Order overview throwings a NotFoundEception when order is still a cart

### LINKED ISSUE #13826: [Admin] 500 when trying to access a cart in the order list

mamazu:
If the order is in state cart the order show should not throw a 500. (It's like #13827 but github decided to close the other MR so I created a new one on a different branch.)

| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes #13826
| License         | MIT

When showing an order it should be checked before rendering the template if the order is actually completed or still in state cart.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
@@ -39,7 +39,7 @@ class OrderRepository extends BaseOrderRepository implements OrderRepositoryInte
         $this->associationHydrator = new AssociationHydrator($entityManager, $class);
     }
 
-    public function getOrderById(string $id): ?OrderInterface
+    public function findOrderById(string $id): ?OrderInterface
     {
         return $this->createQueryBuilder('o')
             ->andWhere('o.id = :id')
--- a/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
+++ b/src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
@@ -39,6 +39,18 @@ class OrderRepository extends BaseOrderRepository implements OrderRepositoryInte
         $this->associationHydrator = new AssociationHydrator($entityManager, $class);
     }
 
+    public function getOrderById(string $id): ?OrderInterface
+    {
+        return $this->createQueryBuilder('o')
+            ->andWhere('o.id = :id')
+            ->andWhere('o.state != :state')
+            ->setParameter('id', $id)
+            ->setParameter('state', OrderInterface::STATE_CART)
+            ->getQuery()
+            ->getOneOrNullResult()
+        ;
+    }
+
     public function createListQueryBuilder(): QueryBuilder
     {
         return $this->createQueryBuilder('o')
--- a/src/Sylius/Bundle/AdminBundle/Resources/config/routing/order.yml
+++ b/src/Sylius/Bundle/AdminBundle/Resources/config/routing/order.yml
@@ -21,7 +21,7 @@ sylius_admin_order_show:
         _sylius:
             section: admin
             repository:
-              method: getOrderById
+              method: findOrderById
               arguments:
                 - '$id'
             permission: true
@@ -35,7 +35,7 @@ sylius_admin_order_history:
         _sylius:
             section: admin
             repository:
-              method: getOrderById
+              method: findOrderById
               arguments:
                 - '$id'
             permission: true
@@ -49,7 +49,7 @@ sylius_admin_order_update:
         _sylius:
             section: admin
             repository:
-              method: getOrderById
+              method: findOrderById
               arguments:
                 - '$id'
             permission: true
@@ -66,7 +66,7 @@ sylius_admin_order_cancel:
         _controller: sylius.controller.order:applyStateMachineTransitionAction
         _sylius:
             repository:
-              method: getOrderById
+              method: findOrderById
               arguments:
                 - '$id'
             permission: true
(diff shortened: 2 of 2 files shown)


src/Sylius/Bundle/AdminBundle/EventListener/OrderShowListener.php
```diff
@@ -0,0 +1,34 @@
...
+use Sylius\Component\Core\Model\OrderInterface;
+use Sylius\Component\Core\OrderCheckoutStates;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+final class OrderShowListener
+{
+    public function showOrder(ResourceControllerEvent $event): void
+    {
+        /** @var OrderInterface $order */
+        $order = $event->getSubject();
```
Roshyo, line 26: What about using an `Assert` instead here ?

mamazu: Yes, this makes much more sense because otherwise it will crash on the next line anyways, when it is not an order.

mamazu:
Suggestion was fixed but the pipeline still fails for some reason.

Roshyo:
It fails because of this: https://github.com/Sylius/Sylius/runs/5866107216?check_suite_focus=true

mamazu:
Thanks for pointing that out. Should be fixed now.

ernestWarwas (changes requested):
Hi, @mamazu thanks for your PR however we need to test the use case and we should adjust the routing configuration here: `src/Sylius/Bundle/AdminBundle/Resources/config/routing/order.yml` rather than create a new listener 💃

mamazu:
@ernestWarwas You mean that we should use a dedicated repository method that filters for that? That sounds like a good idea. I don't know if there is a better solution since this would also apply to all of the other routes like Order history. Technically it could exist for a cart but it doesn't make sense.

mamazu:
I have moved it to the configuration to a method in the repository. Can you check it again @ernestWarwas

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
```diff
@@ -39,6 +39,18 @@ public function __construct(EntityManager $entityManager, ClassMetadata $class)
         $this->associationHydrator = new AssociationHydrator($entityManager, $class);
     }
 
+    public function getOrderById(string $id): ?OrderInterface
```
lchrusciel, line 42: as we are allowing for the null result we should adjust the naming. What is more, it could go to Order namespace, is suppose 
```suggestion
    public function findOrderById(string $id): ?OrderInterface
```

mamazu: I've renamed it.

lchrusciel (approved):

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
```diff
@@ -39,6 +39,18 @@ public function __construct(EntityManager $entityManager, ClassMetadata $class)
         $this->associationHydrator = new AssociationHydrator($entityManager, $class);
     }
 
+    public function findOrderById(string $id): ?OrderInterface
```
HeahDude, line 42:
```suggestion
    public function findCartOrderById(string $id): ?OrderInterface
```
?

lchrusciel: I'm in favour of the current name, but lets see what others have to say

HeahDude: No hard feelings :).

Although I would argue that with the current naming and what the code actually does I would expect:
```php
public function findOrderById(string $id, string $state = OrderInterface::STATE_CART): ?OrderInterface
```

otherwise it might be confusing/inconsistent.

mamazu: I mean in my opinion the best solution would be to have two repositories. One for orders and one for carts but that is too late now.

But the last suggestion of @HeahDude with the second parameter also sounds good.

lchrusciel: oh, now I get your suggestion - for me naming was ok because for me order is in a different state than "cart" - I'm open to refactoring PRs

mamazu:
Replaced by #14174 

Thanks Lukasz for making it your problem now. :D

src/Sylius/Bundle/CoreBundle/Doctrine/ORM/OrderRepository.php
diimpp: > Although I would argue that with the current naming and what the code actually does I would expect:
> 
> ```
> public function findOrderById(string $id, string $state = OrderInterface::STATE_CART): ?OrderInterface
> ```
> 
> otherwise it might be confusing/inconsistent.

It's indeed confusing, even if I'm familiar with `findCartById` construction, but what you're offering is more of `findInStateById` type of method.

As someone who on boards non-sylius dev to sylius projects from time to time, I can guarantee `findByOrder` will not be understood and might lead to methods like `findByTaxon` in taxon repository as made-by-example. :D

One more option: `findPlacedOrderBy***`/`findCartBy***`

vvasiloi: I would stick to Doctrine's magic method convention and name it `findOneByIdAndState`.
Also, it should be added to the interface no matter its name.

PR #13832, opened 2022-04-06
[Behat] Refactor API clients

AdamKasp:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| BC breaks?      | 
| License         | MIT

With many API clients, we have problems with getting the last response, now we have only 2 clients: shop and admin.


src/Sylius/Behat/Context/Api/Admin/ManagingPaymentsContext.php
```diff
@@ -26,6 +26,8 @@
 
 final class ManagingPaymentsContext implements Context
 {
+    private const RESOURCE_PAYMENTS = 'payments';
```
lchrusciel, line 29: sometimes you are using more detailed names of resources, sometimes you use just resource: 
image

Would be nice to be coherent among all contexts with it. Perhaps it would make sense to just extract these consts to separate object? How often it happens, that we are declaring the same value as const?

src/Sylius/Behat/Context/Api/Shop/OrderContext.php
```diff
@@ -366,7 +368,7 @@ public function theCustomerServiceShouldKnowAboutThisAdditionalNotes(
 
         Assert::same(
             $notes,
-            $this->responseChecker->getValue($this->adminOrderClient->show($order->getTokenValue()), 'notes')
+            $this->responseChecker->getValue($this->adminClient->show(self::RESOURCE, $order->getTokenValue()), 'notes')
```
lchrusciel, line 371: Nice to be able to spot it, but perhaps we should extract separate context for it? May be done in separate PR

src/Sylius/Behat/Context/Api/Admin/ManagingPaymentsContext.php
AdamKasp: I was coherent :D but in a different way :). if context use a single resource it was `resource` if it use more resources I declared the resource with a name :)

but I like the idea of separate object with all resource names (especially when for `shop` and `admin` we have same names`

lchrusciel: I was expecting such an answer - but either I'm missing something or you are wrong.

"if context use a single resource it was resource if it use more resources I declared the resource with a name :)"

The exact const that I commented disproves your sentence:
image

There is only single resource, yet it is with a name.

It is also like this in ProductVariant: 
image

AdamKasp: I've created the rule but also I've doesn't follow it :(
you're right this inconsistency was made by mistake.
but I've rethought it and decided to remove these consts at all :)

src/Sylius/Behat/Context/Api/Admin/ManagingZonesContext.php
```diff
@@ -26,6 +26,8 @@
 
 final class ManagingZonesContext implements Context
 {
+    private const RESOURCE = 'zones';
+
```
GSadee, line 30:
```suggestion
```

src/Sylius/Behat/Context/Api/Admin/ManagingExchangeRatesContext.php
```diff
@@ -57,15 +57,15 @@ public function iWantToEditThisExchangeRate(ExchangeRateInterface $exchangeRate)
      */
     public function iBrowseExchangeRatesOfTheStore(): void
     {
-        $this->client->index();
+        $this->client->index('exchange-rates',);
```
Zales0123, line 60: Extra comma, here and below :dancer:

src/Sylius/Behat/Context/Api/Admin/ManagingProductsContext.php
```diff
@@ -438,7 +437,9 @@ public function thisProductShouldHaveOption(ProductInterface $product, ProductOp
      */
     public function theFirstProductOnTheListShouldHave(string $field, string $value): void
     {
-        $products = $this->responseChecker->getCollection($this->client->getLastResponse());
+        $response = $this->sharedStorage->has('response') ? $this->sharedStorage->get('response') : $this->client->getLastResponse();
```
Zales0123, line 440: It's 3 times in the class, could be extracted to the private function

src/Sylius/Behat/Resources/config/services/api.xml
```diff
@@ -244,11 +42,5 @@
             <argument>shop</argument>
             <argument type="service" id="sylius.behat.shared_storage" />
         </service>
-
-        <service id="sylius.behat.api_platform_client.iri" class="Sylius\Behat\Client\ApiPlatformIriClient">
-            <argument type="service" id="test.client" />
-            <argument type="service" id="sylius.behat.shared_storage" />
-            <argument>%sylius.api.authorization_header%</argument>
-        </service>
```
Zales0123, line 252: Changes in this file are just beautiful 🎉

Zales0123 (commented):
👍

Zales0123 (approved):

PR #13839, opened 2022-04-07
Limit the number of packages build run on 1.10

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

It's not required on 1.11, as we test only 2 Symfony versions there, but on 1.10 there are 3 (`^4.4`, `5.3.*`, `5.4.*`), so the number of combinations exceeds the limit of 256 builds :dancer:

### DIFF
--- a/.github/workflows/packages.yml
+++ b/.github/workflows/packages.yml
@@ -50,7 +50,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: ["7.4", "8.0", "8.1"]
+                php: ["7.4", "8.0"]
                 symfony: ["^4.4", "5.3.*", "5.4.*"]
                 package: "${{ fromJson(needs.list.outputs.packages) }}"
 


.github/workflows/packages.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         strategy:
             fail-fast: false
             matrix:
-                php: ["7.4", "8.0", "8.1"]
+                php: ["7.4", "8.0"]
```
GSadee, line 53: I wonder if we shouldn't leave here PHP 8.1 as the newest one 🤔

GSadee (approved):

.github/workflows/packages.yml
Zales0123: I was wondering about it too.. on the other hand, it's more probable that if something is working on 8.0 it will also work on 8.1 than the other way (or it does not make any difference but I need some excuse for this decision xD)

GSadee: Or drop support of Symfony 5.3 as it is not supported anymore and we did sth similar with 5.2

Zales0123: We generally need to make the build configuration consistent, as in some places we still use Symfony 5.2.* in some not 😄 Let's merge it to make the build green and then iterate over the topic on a separate PR, what say you? 🖖

GSadee:

PR #13837, opened 2022-04-07
[Maintenance] Upgrade Symfony Flex, PHPStan and drop support of Symfony 5.2 to fix the build

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

I've removed builds for Symfony 5.2 as it not maintained version of Symfony and it has security vulnerabilities.
I will add a note to the UPGRADE file in a separate PR.

PR #13841, opened 2022-04-07
[Maintenance] Add note to UPGRADE file about dropping Symfony 5.2 support

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | after #13837 
| License         | MIT

### DIFF
--- a/UPGRADE-1.10.md
+++ b/UPGRADE-1.10.md
@@ -1,4 +1,10 @@
+# UPGRADE FROM `v1.10.12` TO `v1.10.13`
+
+1. The support for Symfony 5.2 has been dropped, because it is not maintained version that has some security vulnerabilities. 
+   The recommended Symfony version to use with Sylius is 5.4 as it is the current long-term support version.
+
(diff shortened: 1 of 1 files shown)


UPGRADE-1.10.md
```diff
@@ -1,4 +1,10 @@
+# UPGRADE FROM `v1.10.12` TO `v1.10.13`
+
+1. The support for Symfony 5.2 has been dropped, because it is not maintained version anymore. The recommended Symfony version
```
Ferror, line 3: I would also add information about security issues - it is primary reason why we drop it :)
Screenshot 2022-04-07 at 12 38 13

Zales0123: I would say the combination of these reasons (not maintained + not secured) is definitely a good reason to not test something 👌

Ferror (approved):

Zales0123:
Thank you, @GSadee! 🏅

PR #13849, opened 2022-04-07
[API] Fix sorting products by price of their first variant

coldic3:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT


src/Sylius/Behat/Service/Resolver/PriceResolver.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Behat\Service\Resolver;
+
+final class PriceResolver implements PriceResolverInterface
```
SirDomin, line 16: missing spec for this class 💃

src/Sylius/Bundle/ApiBundle/spec/Doctrine/QueryCollectionExtension/FilterEagerLoadingExtensionSpec.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->beConstructedWith(
+            $decoratedExtension,
+            ['shop_get' => ['resources' => [Product::class => ['enabled' => true]]]]
+        );
+    }
+
+    function it_does_nothing_if_current_resource_and_operation_is_restricted(
+        QueryBuilder $queryBuilder,
+        QueryNameGeneratorInterface $queryNameGenerator,
+        ContextAwareQueryCollectionExtensionInterface $decoratedExtension
```
Rafikooo, line 36: Should be the first parameter since it is used in the `let` function

src/Sylius/Bundle/ApiBundle/Resources/config/api_resources/Product.xml
```diff
@@ -51,6 +51,7 @@
                 <attribute name="normalization_context">
                     <attribute name="groups">shop:product:read</attribute>
                 </attribute>
+                <attribute name="force_eager">false</attribute>
```
lchrusciel, line 54: Is it still needed?

src/Sylius/Bundle/ApiBundle/Resources/config/app/config.yaml
```diff
@@ -16,6 +16,14 @@ parameters:
     sylius.security.new_api_shop_regex: "^%sylius.security.new_api_shop_route%"
     sylius.security.new_api_user_account_route: "%sylius.security.new_api_shop_route%/account"
     sylius.security.new_api_user_account_regex: "^%sylius.security.new_api_user_account_route%"
+      
+sylius_api:
+    filter_eager_loading_extension:
+        restricted_operations:
+            shop_get:
+                resources:
+                    '%sylius.model.product.class%': ~
```
lchrusciel, line 25: can't it be simplified?

src/Sylius/Bundle/ApiBundle/spec/Doctrine/QueryCollectionExtension/FilterEagerLoadingExtensionSpec.php
```diff
@@ -0,0 +1,65 @@
...
+namespace spec\Sylius\Bundle\ApiBundle\Doctrine\QueryCollectionExtension;
+
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\ContextAwareQueryCollectionExtensionInterface;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
+use Doctrine\ORM\QueryBuilder;
+use PhpSpec\ObjectBehavior;
+use Sylius\Component\Core\Model\Order;
+use Sylius\Component\Core\Model\Product;
+
+class FilterEagerLoadingExtensionSpec extends ObjectBehavior
```
lchrusciel, line 23:
```suggestion
final class FilterEagerLoadingExtensionSpec extends ObjectBehavior
```

src/Sylius/Behat/Service/Resolver/PriceResolver.php
GSadee, line 16: Instead of introducing a new class, you should use transformer `LexicalContext` in your steps

src/Sylius/Behat/Context/Api/Shop/ProductContext.php
```diff
@@ -329,6 +313,22 @@ public function theFirstProductOnTheListShouldHaveName(string $name): void
...
+    {
+        $product = $this->responseChecker->getCollection($this->client->getLastResponse())[0];
+
+        $defaultVariantPrice = $this->responseChecker->getValue(
+            $this->productVariantClient->showByIri($product['defaultVariant']),
+            'price'
+        );
+
+        Assert::same($product['name'], $name);
+        Assert::same($defaultVariantPrice, $this->priceResolver->getPriceFromString($price));
```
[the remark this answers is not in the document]
coldic3: There are 5 more occurrences with the `getPriceFromString` private method, so I've decided to do a resolver, and I'm going to replace all of these occurrences in the separate PR.
image

I was thinking about the transformer, but not all of these occurrences can be replaced with the transformer, e.g. on the UI I compare price strings, while on the API I compare integers.

src/Sylius/Behat/Service/Resolver/PriceResolver.php
coldic3: I've answered on that to Łukasz. Can you look at my answer https://github.com/Sylius/Sylius/pull/13849#discussion_r849441973?

src/Sylius/Bundle/ApiBundle/Resources/config/app/config.yaml
coldic3: I remember your suggestion to keep it simple by using `request_uri` from the context. I'm fine with that, but I decided to change it a little bit because I noticed one problem – `request_uri` depends on the endpoint prefix and if someone follows https://docs.sylius.com/en/1.11/customization/api/modifying_endpoints.html#how-to-modify-the-endpoints-prefixes, our workaround will stop working.

src/Sylius/Behat/Context/Api/Shop/ProductContext.php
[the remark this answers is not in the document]
coldic3: Never mind! :smile: I forgot I could use different transformers on the API and UI. So I've removed the price resolver.

src/Sylius/Bundle/ApiBundle/spec/Doctrine/QueryCollectionExtension/FilterEagerLoadingExtensionSpec.php
```diff
@@ -0,0 +1,65 @@
...
+use Sylius\Component\Core\Model\Order;
+use Sylius\Component\Core\Model\Product;
+
+final class FilterEagerLoadingExtensionSpec extends ObjectBehavior
+{
+    function let(ContextAwareQueryCollectionExtensionInterface $decoratedExtension): void
+    {
+        $this->beConstructedWith(
+            $decoratedExtension,
+            ['shop_get' => ['resources' => [Product::class => ['enabled' => true]]]]
```
lchrusciel, line 29: Minor thing, but we don't have a test for the "false" value under the enabled key

src/Sylius/Bundle/ApiBundle/Tests/DependencyInjection/SyliusApiExtensionTest.php
```diff
@@ -51,6 +51,43 @@ public function it_does_not_load_swagger_integration_if_it_does_not_exists()
...
+    /** @test */
+    public function it_loads_filter_eager_loading_extension_restricted_operations_configuration_properly(): void
+    {
+        $this->load([
+            'filter_eager_loading_extension' => [
+                'restricted_operations' => [
+                    'shop_get' => [
+                        'resources' => [
+                            'FirstResourceClass' => [],
+                            'SecondResourceClass' => false,
```
lchrusciel, line 63: I get why you decided on this structure instead of the path - and I totally agree, that it is a better option. However, I don't understand why did you decide to reverse the class <> operation name order. Won't it be more natural for end users to use the same structure as they are using for path definitions? 
```suggestion
                    'FirstResourceClass' => [
                        'resources' => [
                            'shop_get' => [],
                         ],
                     ],
                    'SecondResourceClass' => [
                        'resources' => [
                            'shop_get' => [],
```

src/Sylius/Bundle/ApiBundle/DependencyInjection/Configuration.php
```diff
@@ -38,6 +38,29 @@ public function getConfigTreeBuilder(): TreeBuilder
...
+                        ->arrayNode('restricted_operations')
+                            ->useAttributeAsKey('name')
+                            ->arrayPrototype()
+                                ->children()
+                                    ->arrayNode('resources')
+                                        ->useAttributeAsKey('name')
+                                        ->arrayPrototype()
+                                            ->canBeDisabled()
+                                            ->children()
+                                                ->booleanNode('enabled')->defaultTrue()->end()
```
lchrusciel, line 54: If I understands this tree correctly we have following configs possible:

```yml
sylius_api:
    filter_eager_loading_extension:
        restricted_operations:
            shop_get:
                resources:
                    '%sylius.model.product.class%': ~
```
and equivalent 
```yml
sylius_api:
    filter_eager_loading_extension:
        restricted_operations:
            shop_get:
                resources:
                    '%sylius.model.product.class%': true
```
and equivalent 
```yml
sylius_api:
    filter_eager_loading_extension:
        restricted_operations:
            shop_get:
                resources:
                    '%sylius.model.product.class%': 
                        enabled: true
```

and on another spectrum:
```yml
sylius_api:
    filter_eager_loading_extension:
        restricted_operations:
            shop_get:
                resources:
                    '%sylius.model.product.class%': false
```
and equivalent 
```yml
sylius_api:
    filter_eager_loading_extension:
        restricted_operations:
            shop_get:
                resources:
                    '%sylius.model.product.class%': 
                        enabled: false
```

but we have an extension test only for one of them. Can you add a configuration test for other possibilities? Please omit the ones, that I misunderstood

src/Sylius/Bundle/ApiBundle/Doctrine/QueryCollectionExtension/FilterEagerLoadingExtension.php
```diff
@@ -0,0 +1,44 @@
...
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Extension\ContextAwareQueryCollectionExtensionInterface;
+use ApiPlatform\Core\Bridge\Doctrine\Orm\Util\QueryNameGeneratorInterface;
+use Doctrine\ORM\QueryBuilder;
+
+/**
+ * @experimental
+ * This class decorates api_platform.doctrine.orm.query_extension.filter_eager_loading.
+ * It is a workaround for https://github.com/api-platform/core/issues/2253.
+ */
+final class FilterEagerLoadingExtension implements ContextAwareQueryCollectionExtensionInterface
```
lchrusciel, line 25:
```suggestion
final class RestrictingFilterEagerLoadingExtension implements ContextAwareQueryCollectionExtensionInterface
```

lchrusciel:

src/Sylius/Bundle/ApiBundle/Tests/DependencyInjection/SyliusApiExtensionTest.php
coldic3: I did it that way because we can have multiple resources with the same operation name, so instead of specifying many times 'shop_get' in the configuration, we can do it only once. Then configuration is just shorter, but I agree it's less natural for end users, so I'm open to the refactor, but I think it should be rather:
```php
            'filter_eager_loading_extension' => [
                'restricted_resources' => [                   // restricted_operations -> restricted_resources
                    'FirstResourceClass' => [
                        'operations' => [                     // operations -> resources
                            'shop_get' => [],
                        ],
                    ],
                    'SecondResourceClass' => [
                        'operations' => [                     // operations -> resources
                            'shop_get' => [],
                        ],
                    ],
                ],
            ],
```
WDYT?

lchrusciel: Ok, I get it. But it works because of our soft rule to name things in a given way - and this is not a solid fundament for the decision IMHO. What is more, reversed structure would still benefit if someone would like to restrict several paths from one resource (but I can agree, that it is less likely than what we have now. And the biggest argument for me is the consistency with the resource structures of AP.

In conclusion, I would prefer to have it reversed.

PR #13850, opened 2022-04-08
[Maintenance][Docker]Create simple environment

Ferror:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | no
| License         | MIT


.docker/dev/php.ini
```diff
@@ -0,0 +1,12 @@
+[PHP]
+memory_limit=512M
+
+[date]
+date.timezone=${PHP_DATE_TIMEZONE}
+
+[xdebug]
```
coldic3, line 7: Is xdebug installed in this configuration? I don't see xdebug installation in https://github.com/Sylius/Sylius-Docker-Standard/blob/master/traditional/Dockerfile

Ferror: @coldic3 I added an explanation in the PR description. I considered removing xdebug configs, but It does not bother me XD What do you think should I remove it until next PR with xdebug?
> I decided to not install xdebug yet, but it will be provided in the nearest future

coldic3: It doesn't bother me as well but it may be a little bit misleading if someone like me didn't read PR description XD
Anyway thanks for clarifying that :P

lchrusciel: We tend to keep source code free of unused configs. Would be nice to remove it until next PR

docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+    #        image: postgres:13
+    #        environment:
+    #            - POSTGRES_USER=root
+    #            - POSTGRES_PASSWORD=postgres
+    #        ports:
+    #            - 5432:5432
+    #        networks:
+    #            - sylius
+
+    mailhog:
```
lchrusciel, line 37: Is this service connected to the container? I don't see any MAILER_URL config

Makefile
```diff
@@ -0,0 +1,30 @@
...
+	node_modules/gulp/bin/gulp.js
+
+ci:
+	composer install --no-interaction --no-scripts
+	bin/console doctrine:database:create --if-not-exists
+	bin/console sylius:install --no-interaction
+	bin/console sylius:fixtures:load default --no-interaction
+	bin/console cache:warmup
+	yarn install --pure-lockfile
+	node_modules/gulp/bin/gulp.js
```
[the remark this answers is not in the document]
Ferror: I mean for sake of readability I can change that. But the reason for that.. they are the same commands. When running gulp without any command parameter he will run the `export default` from `gulpfile.js` which in our case is`build`.

Long story short in our case `node_modules/gulp/bin/gulp.js` and `yarn build` are the same.

Makefile
```diff
@@ -0,0 +1,28 @@
...
+ci:
+	composer install --no-interaction --no-scripts
+	bin/console sylius:install --no-interaction
+	bin/console sylius:fixtures:load default --no-interaction
+	bin/console cache:warmup
+	yarn install --pure-lockfile
+	yarn build
+	vendor/bin/phpunit
+	vendor/bin/phpspec run --ansi --no-interaction -f dot
+	vendor/bin/behat --colors --strict --no-interaction -vvv -f progress --tags="~@javascript&&@cli&&~@todo"  # CLI Behat
```
lchrusciel, line 17: Thought for later - won't it be better to reference this file instead of duplication of logic?

lchrusciel (approved):
I'm not sure, I've mentioned it, but in the next iteration we should provide some background worker for catalogue promotions in perhaps separate container or something like it

PR #13852, opened 2022-04-08
[Maintenance] Run the build on tags rather than on releases

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | 
| License         | MIT

The tag push is an actual moment when the new version of Sylius is done, the release is just the icing on the cake 🍰 So it makes more sense to run the workflow when the real thing is happening, not when it's nicely served 🖖

### DIFF
--- a/.github/workflows/application.yml
+++ b/.github/workflows/application.yml
@@ -2,6 +2,8 @@ name: Application
 
 on:
     push:
+        tags:
+            - v1.*
         branches-ignore:
             - 'dependabot/**'
         paths-ignore:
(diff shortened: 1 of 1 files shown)


.github/workflows/application.yml
```diff
@@ -2,6 +2,8 @@ name: Application
 
 on:
     push:
+        tags:
+            - v1.*
```
lchrusciel, line 6: Why did you decide to run them only on 1.x?

lchrusciel (approved):
We should do it on all our packages

GSadee (approved):

.github/workflows/application.yml
GSadee, line 6:
```suggestion
            - v*
```
or
```suggestion
            - *
```

PR #13856, opened 2022-04-10
[UI][Swagger] Add info to swagger paths about iri reference

SirDomin:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Its sometimes hard to tell if you should send iri or string reference in different endpoints, with this normalizer it will be easy to tell if you need to send string or actual iri.

Before:
image

After:
image


loic425:
@SirDomin why just "iri-reference", I think it could be better to add a real IRI example "/api/v2/shop/currencies/{code}"

SirDomin:
@loic425 indeed, thats just simple solution, im working on real iris representation, got working wip on different project with real iri examples, here it doesnt work yet for collections, f.eg currencies : ['iri']. I will work on that case this week and ill let you know once it will be working 🚀

PR #13862, opened 2022-04-12
[API] Sorting products by date

Rafikooo:
| Q               | A
| --------------- | -----
| Branch?         | master 
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| License         | MIT

1) Sorting shop products via API - test covered
2) Sorting products by `taxonPosition` if no other sorting parameter is specified

### DIFF
--- a/features/product/viewing_products/viewing_sorted_products.feature
+++ b/features/product/viewing_products/viewing_sorted_products.feature
@@ -20,16 +20,18 @@ Feature: Sorting listed products
         And this product belongs to "Fluffy Pets"
         And this product's price is "$13.27"
 
-    @ui
+    @api @ui
     Scenario: Sorting products by their dates with descending order
-        When I view newest products from taxon "Fluffy Pets"
+        When I browse products from taxon "Fluffy Pets"
+        And I sort products by the newest date first
         Then I should see 4 products in the list
         And I should see a product with name "Xtreme Pug"
         But the first product on the list should have name "Pug of Love"
 
-    @ui
+    @api @ui
     Scenario: Sorting products by their dates with ascending order
-        When I view oldest products from taxon "Fluffy Pets"
+        When I browse products from taxon "Fluffy Pets"
+        And I sort products by the oldest date first
         Then I should see 4 products in the list
         And I should see a product with name "Berserk Pug"
         But the first product on the list should have name "Xtreme Pug"
--- a/src/Sylius/Behat/Context/Api/Shop/ProductContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/ProductContext.php
@@ -112,6 +112,16 @@ final class ProductContext implements Context
     }
 
     /**
+     * @When /^I sort products by the (oldest|newest) date first$/
+     */
+    public function iSortProductsByTheDateFirst(string $sortDirection): void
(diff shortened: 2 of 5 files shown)


coldic3 (approved):

AdamKasp (approved):

src/Sylius/Behat/Context/Api/Shop/ProductContext.php
```diff
@@ -111,6 +111,14 @@ public function iBrowseProductsFromTaxon(?TaxonInterface $taxon = null): void
         }
     }
 
+    /**
+     * @When /^I sort products by the (oldest|newest) date first$/
+     */
+    public function iSortProductsByTheDateFirst(string $sortDirection): void
+    {
+        $this->client->sort(['createdAt' => 'oldest' === $sortDirection ? 'asc' : 'desc']);
```
AdamKasp, line 119: maybe it is only my private preferences but making it as a two separate steps will be more readable for me :)

TheMilek (approved):

features/product/viewing_products/viewing_sorted_products.feature
```diff
@@ -20,16 +20,18 @@ Feature: Sorting listed products
         And this product belongs to "Fluffy Pets"
         And this product's price is "$13.27"
 
-    @ui
+    @api @ui
     Scenario: Sorting products by their dates with descending order
-        When I view newest products from taxon "Fluffy Pets"
+        When I browse products from taxon "Fluffy Pets"
+        And I sort products by the newest date first
```
GSadee, line 26: It could have stayed as one step

src/Sylius/Bundle/ApiBundle/spec/Filter/Doctrine/TaxonFilterSpec.php
```diff
@@ -55,4 +55,35 @@ function it_adds_taxon_filter_if_property_is_taxon(
             'resourceClass'
         );
     }
+
+    function it_does_not_adds_the_default_order_by_position_if_a_different_order_parameter_is_specified(
```
GSadee, line 59:
```suggestion
    function it_does_not_add_the_default_order_by_taxon_position_if_a_different_order_parameter_is_specified(
```

GSadee (approved):

features/product/viewing_products/viewing_sorted_products.feature
Rafikooo: I wanted to be more consistent with the rest of the tests. The second case is that, in fact, from the UI perspective we cannot open a specific taxon and have already sorted the products by anything other than position

GSadee: Yes, you are right and I am OK with that, but from the business perspective nothing has changed and the user's intention is to view sorted products regardless of the number of steps. So still, it could have stayed as one step, but it is OK as it is right now 😃

Rafikooo: Okay, now I see what you meant :)

GSadee:
Thank you, Rafał! :1st_place_medal:

PR #13602, opened 2022-02-04
relax conflict on liip imagine bundle

Nek-:
| Q               | A
| --------------- | -----
| Branch?         | 1.10, 1.11 and master 
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no 
| Related tickets | https://github.com/Sylius/Sylius/pull/13288
| License         | MIT

With the hope the build is successful on the first try.

Context: In my company, we upgraded to liip-imagine when sylius was still compatible with it, and we are now blocked with sylius upgrades.

PR #13866, opened 2022-04-13
[Maintenance] Relax conflict on liip imagine bundle - rebased version

lchrusciel:
This commit removes support for liip imagine bundle 3.0 because we
extends a class that will not be in the bundle in its version 3.0.

| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | Rebased version of #13602
| License         | MIT


lchrusciel:
Thanks, Łukasz! :tada:

PR #13870, opened 2022-04-14
[Behat][ShippingMethod] Adjust scenarios for shipping rules to be based on items total instead of order total

GSadee:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no?
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Basing on order total instead of items total during choosing an available shipping method can lead to unpredictable behaviours as the shipping fee is a part of the order total. To fix that behaviour, we would like to change the shipping rule to be based on items total.


coldic3 (approved):

lchrusciel (approved):

AdamKasp:
Thank you, Grzegorz! :tada:

PR #13875, opened 2022-04-14
[Swagger] Enum with available locale codes in Accept-Language header

ernestWarwas:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

image

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/Swagger/AcceptLanguageHeaderDocumentationNormalizer.php
+++ b/src/Sylius/Bundle/ApiBundle/Swagger/AcceptLanguageHeaderDocumentationNormalizer.php
@@ -39,6 +39,7 @@ final class AcceptLanguageHeaderDocumentationNormalizer implements NormalizerInt
             'name' => 'Accept-Language',
             'in' => 'header',
             'required' => false,
+            'description' => 'Locales in this enum are all locales defined in the shop and only enabled ones will work in the given channel in the shop.',
             'schema' => [
                 'type' => 'string',
                 'enum' => array_map(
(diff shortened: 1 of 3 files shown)


src/Sylius/Bundle/ApiBundle/Swagger/AcceptLanguageHeaderDocumentationNormalizer.php
```diff
@@ -36,8 +40,12 @@ public function normalize($object, $format = null, array $context = [])
             'in' => 'header',
             'required' => false,
             'schema' => [
-                'type' => 'string'
-            ]
+                'type' => 'string',
+                'enum' => array_map(
+                    fn (LocaleInterface $locale): string => $locale->getCode(),
+                    $this->localeRepository->findAll()
```
AdamKasp, line 46: is it a bit misleading? because users can put whatever locale want but it works only for locales enabled in the channel. what do you think about getting a channel and showing available options? or as a minimum improvement, you can add information that these locales are ALL locales declared in Shop

lchrusciel: the comment with the information that these are all the locales available, but only enabled ones would work in the given channel in the shop could be a nice improvement.

Still it's much better than before. Thanks Ernest!

ernestWarwas: Ok guys I'll improve this a little 😃

ernestWarwas: @lchrusciel, @AdamKasp done 🚀

lchrusciel (approved):

PR #13878, opened 2022-04-14
[Template] Variant name instead of product name

ernestWarwas:
| Q               | A
| --------------- | -----
| Branch?         | 1.10
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

Show variant name instead of the product name.

### DIFF
--- a/features/product/viewing_product_in_admin_panel/viewing_details_of_product_with_variants.feature
+++ b/features/product/viewing_product_in_admin_panel/viewing_details_of_product_with_variants.feature
@@ -41,8 +41,8 @@ Feature: Viewing details of a product with variants
     Scenario: Viewing variants block
         When I access "Iron Shield" product page
         Then I should see 2 variants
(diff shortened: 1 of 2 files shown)


Ferror (approved):
Nice catch!

Ferror (commented):
Btw. We have tons of tests and we don't have a scenario to check if the variant name is actually the variant name?

Maybe we could provide a scenario like this in the next iteration? Even in sprint :)

ernestWarwas:
> Btw. We have tons of tests and we don't have a scenario to check if the variant name is actually the variant name?

We have tests for that 😄 I didn't check it. Fixed.

lchrusciel (approved):

GSadee (approved):

lchrusciel:
Thank you, Ernest! :1st_place_medal:

PR #13879, opened 2022-04-15
[HotFix][Behat][API] Fix suddenly failing tests

coldic3:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

Sometimes there is a need to resend a request if the last response has been overwritten.

### DIFF
--- a/src/Sylius/Behat/Client/ApiClientInterface.php
+++ b/src/Sylius/Behat/Client/ApiClientInterface.php
@@ -44,6 +44,8 @@ interface ApiClientInterface
 
     public function upload(): Response;
 
+    public function resend(): Response;
+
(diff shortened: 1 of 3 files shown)


GSadee (approved):

AdamKasp (approved):

lchrusciel (commented):
It appears to me to be quite a smell in the design. If we are "resending" something, that is stateful within the object, because the returned value was overridden. To fix this issue, I would recommend sending a proper request before making checks and in another round removal of the statefulness of clients at all

lchrusciel:
Thank you, Kevin! :tada:

PR #13896, opened 2022-04-26
[behat]Extract Request creation to separate service

Ferror:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

It is a part of improving Developer Experience in Behat tests. I removed factory methods out of the Request class, as the object has really grown up. Moved these methods to separate Factory required a lot of changes, that's why I decided to postpone here and merge current changes.

Created Request Builder and Factory will provide an additional layer for future developers to build their own Requests

### DIFF
--- a/src/Sylius/Behat/Client/ContentTypeGuide.php
+++ b/src/Sylius/Behat/Client/ContentTypeGuide.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Behat\Client;
+
+use Symfony\Component\HttpFoundation\Request as HttpRequest;
+
+class ContentTypeGuide implements ContentTypeGuideInterface
+{
+    private const JSON_CONTENT_TYPE = 'application/json';
+    private const PATCH_CONTENT_TYPE = 'application/merge-patch+json';
+    private const LINKED_DATA_JSON_CONTENT_TYPE = 'application/ld+json';
+
+    public function guide(string $method): string
+    {
+        if ($method === HttpRequest::METHOD_PATCH) {
+            return self::PATCH_CONTENT_TYPE;
+        }
+
+        if ($method === HttpRequest::METHOD_PUT) {
+            return self::LINKED_DATA_JSON_CONTENT_TYPE;
+        }
+
+        return self::JSON_CONTENT_TYPE;
+    }
+}
--- a/src/Sylius/Behat/Client/RequestBuilder.php
+++ b/src/Sylius/Behat/Client/RequestBuilder.php
@@ -1,0 +1,77 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Behat\Client;
+
+use Symfony\Component\HttpFoundation\File\UploadedFile;
+
+final class RequestBuilder
+{
+    private string $uri;
+    private string $method;
+    private array $parameters = [];
+    private array $headers = [];
+    private array $content = [];
+    private array $files = [];
+
+    private function __construct(string $uri, string $method)
+    {
+        $this->uri = $uri;
+        $this->method = $method;
+    }
+
+    public static function create(string $uri, string $method): self
+    {
+        return new self($uri, $method);
+    }
+
+    public function withHeader(string $name, string $value): self
+    {
+        $this->headers[$name] = $value;
+
+        return $this;
+    }
+
+    public function withFile(string $key, UploadedFile $file): self
+    {
+        $this->files[$key] = $file;
+
+        return $this;
+    }
+
+    public function withContent(array $content): self
+    {
+        $this->content = $content;
+
+        return $this;
+    }
+
+    public function withParameter(string $key, string $value): self
+    {
+        $this->parameters[$key] = $value;
+
+        return $this;
+    }
+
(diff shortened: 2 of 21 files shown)


src/Sylius/Behat/Client/RequestFactory.php
```diff
@@ -0,0 +1,179 @@
...
+        string $section,
+        string $resource,
+        string $authorizationHeader,
+        ?string $token = null
+    ): RequestInterface {
+        $builder = RequestBuilder::create(
+            sprintf('/api/v2/%s/%s', $section, $resource),
+            HttpRequest::METHOD_POST,
+        );
+        $builder->withHeader('CONTENT_TYPE', 'application/ld+json');
```
lchrusciel, line 71: headers could go to const. probably protected if we want to leave this object not finalized

src/Sylius/Behat/Client/RequestFactory.php
```diff
@@ -0,0 +1,179 @@
...
+
+        if ($token) {
+            $builder->withHeader('HTTP_Authorization', 'Bearer ' . $token);
+        }
+
+        foreach ($additionalHeaders as $name => $value) {
+            $builder->withHeader($name, $value);
+        }
+
+        return $builder->build();
```
lchrusciel, line 172: I'm thinking if we shouldn't return the builder and allow for a build call in the contexts. This would allow for the decoration of this object as well. Perhaps two services would be a good choice? Then we could finalize this service and the default approach to extending it would be with decoration

src/Sylius/Behat/Client/RequestFactory.php
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private static function resolveHttpMethod(string $method): string
```
lchrusciel, line 175: If you decide to keep this class open for extending, we should make this method protected or extract it to separate service as it is used in 5 places 
```suggestion
    protected static function resolveHttpMethod(string $method): string
```

src/Sylius/Behat/Client/RequestBuilder.php
```diff
@@ -0,0 +1,82 @@
...
+
+        return $this;
+    }
+
+    public function withoutPartialContent(string $key, array $content): self
+    {
+        $this->content[$key] = $content;
+
+        return $this;
+    }
```
lchrusciel, line 62: What is the difference between these two functions?

Ferror: Yea, I deleted it as they are not used yet.

src/Sylius/Behat/Client/RequestFactory.php
Ferror: I created ContentTypeGuide - a small little service that we can forget about ;p

src/Sylius/Behat/Client/RequestFactory.php
Ferror: I like this idea, but I would rather stick with the current solution for now, and in another PR return Builder. To achieve that I will need to refactor a lot ApiPlatform Client

src/Sylius/Behat/Client/ContentTypeGuide.php
```diff
@@ -0,0 +1,36 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Behat\Client;
+
+use Symfony\Component\HttpFoundation\Request as HttpRequest;
+
+class ContentTypeGuide implements ContentTypeGuideInterface
```
lchrusciel, line 18:
```suggestion
final class ContentTypeGuide implements ContentTypeGuideInterface
```

While having only one public method there is no benefit of having class not finalized

src/Sylius/Behat/Client/RequestBuilder.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        return new self($uri, $method);
+    }
+
+    public function withHeader(string $name, string $value): self
+    {
+        $this->headers[$name] = $value;
+
+        return $this;
+    }
```
lchrusciel, line 43: I'm wondering if there would be any benefit of making these calls immutable
```suggestion
    public function withHeader(string $name, string $value): self
    {
        $builder = self::create($this->uri, $this->method);
        $builder->headers[$name] = $value;

        return $builder;
    }
```

src/Sylius/Behat/Client/RequestFactory.php
lchrusciel: We could introduce RequestBuilderFactory which will have almost the same API but will return builder, which will be called in factories. But perhaps I'm making it too complex.

lchrusciel (approved):

lchrusciel:
Thanks, Zbigniew! :1st_place_medal:

AdamKasp (approved):

src/Sylius/Behat/Client/RequestBuilder.php
Ferror: I'm thinking that the builder is kind of mutable pattern (but I might be wrong). Your example implementation will fail with the use case:
```php
$builder
  ->withHeader('header1', 'value1')
  ->withHeader('header2', 'value2')
```

We could change builder implementation like this: (postponed most of the methods)

```php
<?php

final class RequestBuilder
{
    private function __construct(
        private string $uri,
        private string $method,
        private array $parameters = [],
        private array $headers = [],
        private array $content = [],
        private array $files = [],
    )
    {
    }

    public static function create(string $uri, string $method): self
    {
        return new self($uri, $method);
    }

    public function withHeader(string $name, string $value): self
    {
        $builder = self::create($this->uri, $this->method, $this->parameters, $this->headers /* ... */);
        $builder->headers[$name] = $value;
        
        return $builder;
    }

    public function build(): RequestInterface
    {
        return new Request(
            $this->uri,
            $this->method,
            $this->parameters,
            $this->headers,
            $this->content,
            $this->files,
        );
    }
}
```

But i don't know if this isn't a little bit of overenginnering

PR #13898, opened 2022-04-26
[maintenance][mailer]Assert that email is not null

Ferror:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | none
| License         | MIT

The latest Sylius Mailer update changed the email address type as not nullable. Therefore our psalm and phpstan pipelines recorded errors. This PR checks if the email is null. Similarly, as we did it in Sylius/Plus. There is no need to send a mail when the email address is null.


lchrusciel (approved):

GSadee (commented):
I guess that the same problem will be in 1.10 and 1.11, could you change the base branch and adjust PR to lower versions?

GSadee (approved):

GSadee:
Thank you, Zbigniew! :1st_place_medal:

PR #13904, opened 2022-04-27
[behat]Refactor Managing Countries Behat Context to RequestBuilder solution

Ferror:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bugfix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | iterates over #13896
| License         | MIT

# Summarise 
1. Remove `addParameter` method as it was not used. Btw `addParameter` and `addFilter` were the same functions
2. Make `getToken` method private and use SharedStorage `token` parameter
3. Refactor `addFile` method to the new builder solution as it was used only once
4. Make `ContentTypeGuide` final
5. Create `RequestBuilderFactoryInterface` as more customizable version of `RequestFactoryInterface`
6. Refactor `src/Sylius/Behat/Context/Api/Admin/ManagingCountriesContext.php` behat context to use RequestBuilder in battle


src/Sylius/Behat/Context/Api/Admin/ManagingCountriesContext.php
```diff
@@ -16,58 +16,63 @@
...
     public function iWantToAddANewCountry(): void
     {
-        $this->client->buildCreateRequest(Resources::COUNTRIES);
+        $requestBuilder = $this->requestBuilderFactory->create(
+            PathPrefixes::ADMIN_PREFIX,
+            Resources::COUNTRIES,
+            'Authorization',
+            $this->sharedStorage->get('token')
+        );
+        $this->sharedStorage->set('request-builder', $requestBuilder);
```
GSadee, line 54:
```suggestion
        $this->sharedStorage->set('request_builder', $requestBuilder);
```
and everywhere else, as we are using snake_case for keys in shared storage

src/Sylius/Behat/Context/Api/Admin/ManagingCountriesContext.php
```diff
@@ -76,23 +81,47 @@ public function iAddIt(): void
...
     {
-        $this->client->buildUpdateRequest(Resources::COUNTRIES, $country->getCode());
+        $showRequestBuilder = $this->requestBuilderFactory->show(
+            PathPrefixes::ADMIN_PREFIX,
+            Resources::COUNTRIES,
+            $country->getCode(),
+            'Authorization',
+            $this->sharedStorage->get('token')
+        );
+        $showResponse = $this->client->executeCustomRequest($showRequestBuilder->build());
```
GSadee, line 91: I wonder if we shouldn't stay with:
```suggestion
        $showResponse = $this->client->show(Resources::COUNTRIES, $country->getCode());
```
as the creation of request is there, so the problem with statefulness of this service is solved, but this will reduce the amount of unnecessary code in contexts.

Or
```suggestion
        $showRequest = $this->requestFactory->show(
            PathPrefixes::ADMIN_PREFIX,
            Resources::COUNTRIES,
            $country->getCode(),
            'Authorization',
            $this->sharedStorage->get('token')
        );
        $showResponse = $this->client->executeCustomRequest($showRequest);
```

src/Sylius/Behat/Context/Api/Admin/ManagingCountriesContext.php
```diff
@@ -76,23 +81,47 @@ public function iAddIt(): void
...
     }
 
     /**
      * @When I enable it
      */
     public function iEnableIt(): void
     {
-        $this->client->addRequestData('enabled', true);
+        /** @var RequestBuilder $request */
+        $request = $this->sharedStorage->get('request-builder');
```
GSadee, line 111:
```suggestion
        /** @var RequestBuilder $requestBuilder */
        $requestBuilder = $this->sharedStorage->get('request_builder');
```

src/Sylius/Behat/Context/Api/Admin/ManagingCountriesContext.php
```diff
@@ -177,9 +227,26 @@ public function iDoNotSpecifyTheProvince(): void
...
+            PathPrefixes::ADMIN_PREFIX,
+            Resources::PROVINCES,
+            $province->getCode(),
+            'Authorization',
+            $this->sharedStorage->get('token')
+        );
+        $updateRequestBuilder->withContent(json_decode($showResponse->getContent(), true));
+        $updateRequestBuilder->withPartialContent('name', '');
+
+        $this->client->executeCustomRequest($updateRequestBuilder->build());
```
lchrusciel, line 249: I'm convinced that the stuff you are proposing is a nice improvement and a good direction for it, however, I'm concerned about the chosen path. 

 This is quite a large refactor at once if we want to switch all our implementations to the new structure. What would you say, about keeping quite a similar API for that we had on client, but moving it to RequestFactory. E.g.
```php
        $this->requestFactory->update(Resources::PROVINCES, $province->getCode()); // Would request for show of give province and initiate content for update with it, then will store it in shared storage or builder is returned
        $this->requestFactory->addRequestData('name', '');  // mutates data hold in shared storage
        $request = $this->requestFactory->create(); // builds requests and returns it back

        $this->client->executeCustomRequest($request);
```

The advantage of it would be, that it simplifies the migration process, as we would already have client and request creation decoupled and both services would be stateless. The state of request will be held and maintained in shared storage. In addition if one would like to change something with builder, it would be available from shared storage

Ferror: I agree that it is quite a bit of refactor, and we probably should look for a little more approachable solution.

Yet, I can't agree with the move of `Shared Storage` into the `Request Factory`, as it will just move the pain/problem we are trying to heal. It is true that we need some sort of helper class (or are we?), but I would stick to the single responsibility of the services/classes. Isn't `addRequestData` gonna create state within `Request Factory`?

lchrusciel: I can agree, that we are only moving the problem away. But it is still beneficial in my opinion - we would have the client decoupled already. So instead of a stateful object with broken boundaries, we would have a stateful object and an object with proper boundaries. The next step of refactoring would be to place the state in a separate service(`SharedStorage`) to push it to a side of our processing logic. Later, the question remains, how would we provide an already created request object. Are we leaving `SharedStorage` in place or not? We can always provide it as an argument and keep the reference to it in context (a little bit similar to what SymfonyForm type is doing)

src/Sylius/Behat/Context/Api/Admin/ManagingCountriesContext.php
```diff
@@ -265,13 +226,11 @@ public function iShouldBeNotifiedThatItHasBeenSuccessfullyCreated(): void
...
     {
-        $request = $this->requestFactory->index(
+        $requestBuilder = $this->requestBuilderFactory->get(
             PathPrefixes::ADMIN_PREFIX,
             Resources::COUNTRIES,
-            'Authorization',
-            $this->sharedStorage->get('token')
         );
-        $response = $this->client->executeCustomRequest($request);
+        $response = $this->client->executeCustomRequest($requestBuilder->build());
```
lchrusciel, line 231: Shouldn't this request be authorized?

src/Sylius/Behat/Client/Helper.php
```diff
@@ -0,0 +1,61 @@
...
+
+    public function authorize(RequestBuilder $requestBuilder): RequestBuilder
+    {
+        return $requestBuilder->withHeader(
+            sprintf('HTTP_%s', $this->authorizationHeader),
+            sprintf('Bearer %s',$this->sharedStorage->get('token'))
+        );
+    }
+
+    public function preparePut(string $section, string $resource, string $id): RequestBuilder
```
lchrusciel, line 47:
```suggestion
    public function preparePut(string ...$resources): RequestBuilder
```
Shouldn't we keep the contract of get and put operations? Just to provide the same functionality

src/Sylius/Behat/Client/Helper.php
```diff
@@ -0,0 +1,61 @@
...
+    {
+        /** @var RequestBuilder $requestBuilder */
+        $requestBuilder = $this->sharedStorage->get('request_builder');
+
+        $requestBuilder = $closure($requestBuilder);
+
+        $this->sharedStorage->set('request_builder', $requestBuilder);
+    }
+
+    public function authorize(RequestBuilder $requestBuilder): RequestBuilder
```
lchrusciel, line 39: This method should be done automatically if we are authorized(we called the "I am logged in as..." step in behat). Perhaps it should be done as a wrapper on request factory

src/Sylius/Behat/Client/Helper.php
```diff
@@ -0,0 +1,61 @@
...
+        private RequestBuilderFactoryInterface $requestBuilderFactory,
+        private ApiClientInterface $apiClient,
+        private string $authorizationHeader,
+    ) {
+    }
+
+    public function update(callable $closure): void
+    {
+        /** @var RequestBuilder $requestBuilder */
+        $requestBuilder = $this->sharedStorage->get('request_builder');
```
lchrusciel, line 32: the problem with this implementation is that we need to remember to put request builder in shared storage anyway. Thats quite an assumption, and it is implicit, which may lead to errors or just people will not use it

src/Sylius/Behat/Client/Helper.php
Ferror: The SharedStorage sets Bearer token. The token is null when the account is not logged in so the header will be empty

PR #13912, opened 2022-04-29
Support Twig 3

Zales0123:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | 
| License         | MIT

I was wondering, is it worth running builds on both Twig 2 and 3 🤔 But we already have a lot of them, so I decided to skip it for now. Maybe we should have one additional build with Twig 2 (as with the current `composer update` we will always use Twig 3)? 🤷‍♂️ Nevertheless, it's nice we can finally support Twig 3 🖖

### DIFF
--- a/.github/workflows/application.yml
+++ b/.github/workflows/application.yml
@@ -108,7 +108,7 @@ jobs:
                 name: Run PHPStan
                 run: vendor/bin/phpstan analyse
                 if: always() && steps.end-of-setup.outcome == 'success'
-            
+
             -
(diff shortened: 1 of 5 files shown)


lchrusciel (approved):
But I would vote for one additional build with twig 2 anyway

GSadee (approved):

GSadee:
Thanks, Mateusz! :tada:

PR #13914, opened 2022-04-29
Revert "[docker]Run Catalog Promotion via SYNC messenger transport"

lchrusciel:
Reverts Sylius/Sylius#13897, due to too quick merge. The build itself was red and it should not land in the codebase in the first place


lchrusciel:
Thanks, Łukasz! :tada:

PR #13958, opened 2022-05-11
[docker][maintenance]Extend Makefile possibilities and compose tests

Ferror:
1. Removed inline-docker pipeline - it is exactly like docker-compose but uglier
2. Separated tests for static analysis and behat (integration tests)
3. Created special docker profiles (https://docs.docker.com/compose/profiles/)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,53 +4,34 @@ on:
     push:
 
 jobs:
-    docker-compose:
-        name: Execute Tests on Docker Compose
+    static-analyze:
+        name: Execute Static Analyze on Docker Compose
         runs-on: ubuntu-latest
         env:
             COMPOSE_FILE: docker-compose.test.yml
 
         steps:
-            -   name: Checkout code
+            -
+                name: Checkout Code
                 uses: actions/checkout@v2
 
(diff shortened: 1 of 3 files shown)


.github/workflows/docker.yml
```diff
@@ -4,52 +4,34 @@ on:
...
+            -
+                name: Checkout Code
                 uses: actions/checkout@v2
 
-            -   name: Run PHPUnit, PHPSpec and Behat tests
-                run: docker-compose run ci
+            -
+                name: Run PHPUnit, PHPSpec and Psalm
+                # docker-compose -f docker-compose.test.yml --profile static-analyze run static
+                run: docker-compose --profile static-analyze run static
```
coldic3, line 21: Shouldn't it be like the one below?
```suggestion
                run: docker-compose -f docker-compose.test.yml --profile static-analyze run static
```

Ferror: I use the `COMPOSE_FILE: docker-compose.test.yml` env to "sugar" the code :) The `COMPOSE_FILE` basically points to which file is "default" to the docker compose

lchrusciel (approved):

coldic3 (approved):

PR #13962, opened 2022-05-12
[API] Checkout as guest with registered account

lchrusciel:
| Branch?         | master |
| Deprecations?   | no |
| Related tickets |  |

### DIFF
--- a/features/cart/shopping_cart/clearing_cart_after_logging_out.feature
+++ b/features/cart/shopping_cart/clearing_cart_after_logging_out.feature
@@ -7,16 +7,27 @@ Feature: Clearing cart after logging out
     Background:
         Given the store operates on a single channel in "United States"
         And the store has a product "Stark T-shirt" priced at "$12.00"
-        And I am a logged in customer
-        And I have product "Stark T-Shirt" in the cart
 
     @ui
     Scenario: Clearing cart after logging out
+        Given I am a logged in customer
+        And I have product "Stark T-Shirt" in the cart
         When I log out
         And I see the summary of my cart
         Then my cart should be empty
 
     @api
     Scenario: Clearing cart after logging out
+        Given I am a logged in customer
+        And I have product "Stark T-Shirt" in the cart
         When I log out
-        Then I don't have access to see the summary of my previous cart
+        Then I should not have access to the summary of my previous cart
+
+    @api
+    Scenario: Blocking access to cart if logged user did any action over it (what can be treated as signing it)
+        Given there is a user "john@snow.com"
+        When I add this product to the cart
+        And I log in as "john@snow.com" with "sylius" password
+        And I add this product to the cart
+        And I log out
+        Then I should not have access to the summary of my previous cart
--- a/src/Sylius/Bundle/ApiBundle/CommandHandler/Cart/BlameCartHandler.php
+++ b/src/Sylius/Bundle/ApiBundle/CommandHandler/Cart/BlameCartHandler.php
@@ -51,6 +51,7 @@ final class BlameCartHandler implements MessageHandlerInterface
         }
 
         $cart->setCustomer($user->getCustomer());
+        $cart->setCreatedByGuest(false);
 
         $this->orderProcessor->process($cart);
     }
--- a/src/Sylius/Bundle/ApiBundle/CommandHandler/Cart/PickupCartHandler.php
+++ b/src/Sylius/Bundle/ApiBundle/CommandHandler/Cart/PickupCartHandler.php
@@ -71,9 +71,11 @@ final class PickupCartHandler implements MessageHandlerInterface
         $cart->setLocaleCode($this->getLocaleCode($pickupCart->localeCode, $channel));
         $cart->setCurrencyCode($currency->getCode());
         $cart->setTokenValue($pickupCart->tokenValue ?? $this->generator->generateUriSafeString(10));
+
         if ($customer !== null) {
             $cart->setCustomer($customer);
             $cart->setBillingAddress($this->getDefaultAddress($customer));
+            $cart->setCreatedByGuest(false);
         }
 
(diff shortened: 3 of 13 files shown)


features/cart/shopping_cart/clearing_cart_after_logging_out.feature
```diff
@@ -7,16 +7,27 @@ Feature: Clearing cart after logging out
...
         Then I don't have access to see the summary of my previous cart
+
+    @api
+    Scenario: Blocking access to cart if logged user did any action over it (what can be treated as signing it)
+        Given there is a user "john@snow.com"
+        When I add this product to the cart
+        And I log in as "john@snow.com" with "sylius" password
+        And I add this product to the cart
+        And I log out
+        Then I don't have access to see the summary of my previous cart
```
Zales0123, line 33:
```suggestion
        Then I should not have access to the summary of my previous cart
```
?

src/Sylius/Bundle/ApiBundle/CommandHandler/Cart/BlameCartHandler.php
```diff
@@ -51,6 +51,7 @@ public function __invoke(BlameCart $blameCart): void
         }
 
         $cart->setCustomer($user->getCustomer());
+        $cart->setCreatedByGuest(false);
```
Zales0123, line 54: Maybe we should set this flag together with the customer in some `assignLoggedInCustomer` method inside the `Order`? 🤔 That would make the object responsible for its behaviour. But it's definitely an improvement we can introduce later 🖖

src/Sylius/Bundle/ApiBundle/CommandHandler/Cart/PickupCartHandler.php
```diff
@@ -111,6 +108,20 @@ private function getLocaleCode(?string $localeCode, ChannelInterface $channel):
...
+    {
+        if ($customer !== null) {
+            $cart->setCustomer($customer);
+            $cart->setBillingAddress($this->getDefaultAddress($customer));
+            $cart->setCreatedByGuest(false);
+
+            return;
+        }
+
+
```
Zales0123, line 121: Doubled blank line 😱

src/Sylius/Bundle/ApiBundle/CommandHandler/Cart/PickupCartHandler.php
```diff
@@ -111,6 +108,20 @@ private function getLocaleCode(?string $localeCode, ChannelInterface $channel):
...
+        $cart->setCreatedByGuest(true);
```
Zales0123, line 122:
```suggestion
        if ($customer === null) {
            $cart->setCreatedByGuest(true);

            return;
        }

        $cart->setCustomer($customer);
        $cart->setBillingAddress($this->getDefaultAddress($customer));
        $cart->setCreatedByGuest(false);
```

? 😄

src/Sylius/Bundle/ApiBundle/Doctrine/QueryItemExtension/OrderMethodsItemExtension.php
```diff
@@ -79,7 +79,14 @@ private function applyForVisitor(QueryBuilder $queryBuilder, QueryNameGeneratorI
...
             ->leftJoin('customer.user', 'user')
-            ->andWhere($queryBuilder->expr()->orX('user IS NULL', sprintf('%s.customer IS NULL', $rootAlias)))
+            ->andWhere($queryBuilder->expr()->orX(
+                'user IS NULL',
+                sprintf('%s.customer IS NULL', $rootAlias),
+                $queryBuilder->expr()->andX(
+                    sprintf('%s.customer IS NOT NULL', $rootAlias),
+                    sprintf('%s.createdByGuest = :createdByGuest', $rootAlias)
+                )
+            ))->setParameter('createdByGuest', true)
```
Zales0123, line 89: This is the same as in the previous repository, maybe we should introduce some custom doctrine filter/function for that? Again, an improvement to be considered 🖖

src/Sylius/Bundle/ApiBundle/EventListener/ApiCartBlamerListener.php
```diff
@@ -45,7 +45,7 @@ public function onInteractiveLogin(InteractiveLoginEvent $interactiveLoginEvent)
         }
 
         $cart = $this->getCart();
-        if (null === $cart || null !== $cart->getCustomer()) {
+        if (null === $cart || !$cart->getCreatedByGuest()) {
```
Zales0123, line 48: That change assumes that the flag would **always** be set when the customer is set as well. So it only bumps the need of some custom method on the `Order` entity (as mentioned in one of the above comment 🚀)

lchrusciel: In this case, it makes total sense, because of the order wasn't already claimed, now it will be. But for sure custom method is viable option for setting it.

src/Sylius/Bundle/ApiBundle/Doctrine/QueryItemExtension/OrderMethodsItemExtension.php
lchrusciel: I would like to open improvement PR with doctrine criteria for this later

src/Sylius/Bundle/ApiBundle/CommandHandler/Cart/BlameCartHandler.php
lchrusciel: I will open a PR with proposal based on this one

lchrusciel: https://github.com/Sylius/Sylius/pull/13968

PR #13986, opened 2022-05-16
[Bugfix] [Core] Calculate order item subtotal as the sum of units discounted price

### LINKED ISSUE #13532: [Bug] Inaccurate order item subtotal with uneven unit level promotion adjustments
**Sylius version affected**: 1.10+

**Description**  
Inaccurate order item subtotal displays in cart summary when it has uneven unit level promotion adjustments.
It's the same issue described in https://github.com/Sylius/Sylius/issues/11112, but it's was only fixed in admin panel.

**Steps to reproduce**  
The issue occurred in a custom promotion action.
The scenario is as follows:
Given the store has a product "PHP T-Shirt" priced at "$100.00"
And there's a promotion "50% off second item" that gives 50% discount on every the 2nd item.
When I add 3 products "PHP T-Shirt" to the cart
Then total price of "PHP T-Shirt" item should be "$250.00"

This will fail because the total is $249.99 instead of $250.00.

The $50 discount amount is distributed across the 3 order item units.
The units will have the following adjustments
Unit 1: -$16.67
Unit 2: -$16.67
Unit 3: -$16.66

The subtotal is displayed in https://github.com/Sylius/Sylius/blob/bc565db9eb39372b22a94963ac302445a0efbac8/src/Sylius/Bundle/ShopBundle/Resources/views/Cart/Summary/_item.html.twig#L29
where it calls
https://github.com/Sylius/Sylius/blob/bc565db9eb39372b22a94963ac302445a0efbac8/src/Sylius/Component/Core/Model/OrderItem.php#L151-L154
which multiplies the quantity with
https://github.com/Sylius/Sylius/blob/bc565db9eb39372b22a94963ac302445a0efbac8/src/Sylius/Component/Core/Model/OrderItem.php#L116-L131
which takes the unit price and add the adjustments from the 1st unit.
In this case it will be `($100.00 + (-$16.67)) * 3 = $249.99`.

**Possible Solution**  
In `Sylius\Component\Core\Model\OrderItem::getSubtotal`, instead of using `Sylius\Component\Core\Model\OrderItem::getDiscountedUnitPrice`, iterate over the units and sum their total.

vvasiloi:
| Branch?         | 1.10          |
| Deprecations?   | no |
| Related tickets | fixes #13532 and #11585                   |

### DIFF
--- a/src/Sylius/Component/Core/spec/Model/OrderItemSpec.php
+++ b/src/Sylius/Component/Core/spec/Model/OrderItemSpec.php
@@ -14,7 +14,9 @@ declare(strict_types=1);
 namespace spec\Sylius\Component\Core\Model;
 
 use PhpSpec\ObjectBehavior;
+use Sylius\Component\Core\Model\Adjustment;
 use Sylius\Component\Core\Model\AdjustmentInterface;
+use Sylius\Component\Core\Model\OrderItemUnit;
 use Sylius\Component\Core\Model\OrderItemUnitInterface;
 use Sylius\Component\Resource\Model\VersionedInterface;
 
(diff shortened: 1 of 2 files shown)


GSadee:
may become messed-up.**

execute `git fetch origin && git reset "bugfix/13532"` to update your local branch.
 
src/Sylius/Component/Core/spec/Model/OrderItemSpec.php
```diff
@@ -97,24 +99,29 @@ function it_returns_unit_price_as_discounted_unit_price_if_there_are_no_units():
...
+
+        $secondUnit = new OrderItemUnit($this->getWrappedObject());
+        $adjustment2 = new Adjustment();
+        $adjustment2->setType(AdjustmentInterface::ORDER_UNIT_PROMOTION_ADJUSTMENT);
+        $adjustment2->setAmount(-1667);
+        $secondUnit->addAdjustment($adjustment2);
+
+        $secondUnit = new OrderItemUnit($this->getWrappedObject());
+        $adjustment3 = new Adjustment();
+        $adjustment3->setType(AdjustmentInterface::ORDER_UNIT_PROMOTION_ADJUSTMENT);
```
lchrusciel, line 120: We should have at least one adjustment of different type in this test

lchrusciel (commented):
Before merging, it would be good to add at least one behat scenario (with exact data that you provided in a neat bug report #13532)

src/Sylius/Component/Core/spec/Model/OrderItemSpec.php
vvasiloi: Why?

lchrusciel: If I will remove the type from here: 
image
test will still pass. We may not catch it with higher level tests

vvasiloi: ok

vvasiloi:
Replaced by https://github.com/Sylius/Sylius/pull/14090

PR #14022, opened 2022-05-27
Commnad to show order details

dantleech:
| Branch?         | master                                                       |
| Bug fix?        | no                                                           |
| New feature?    | yes                                                          |
| BC breaks?      | no                                                           |
| Deprecations?   | no                                                           |
| Related tickets |                                                              |

This PR adds a command to show details of an order.

image

I've previously found it useful for local development and on production to quickly check the state of an order on the CLI without looking into the admin interface. Not sure how useful this would be for others.

Shows:

- Order attributes
- Order Items
- Payments
- Shipments

TODO?

- Add an option to retrieve an oder by ID or token instead of number ?

---

tests seem to be currently failing in master

### DIFF
--- a/src/Sylius/Behat/Context/Cli/ShowOrderContext.php
+++ b/src/Sylius/Behat/Context/Cli/ShowOrderContext.php
@@ -52,12 +52,15 @@ final class ShowOrderContext implements Context
     public function runShowOrderCommand(string $number): void
     {
         $this->application = new Application($this->kernel);
-        $this->application->add(new ShowOrderCommand());
+        $this->application->add(new ShowOrderCommand($this->orderRepository));
 
         $this->command = $this->application->find('sylius:order:show');
         $this->tester = new CommandTester($this->command);
 
-        $this->tester->execute(['command' => 'sylius:order:show']);
+        $this->tester->execute([
+            'command' => 'sylius:order:show',
+            'number' => ltrim($number, '#'),
+        ]);
     }
 
     /**
--- a/src/Sylius/Behat/Context/Cli/ShowOrderContext.php
+++ b/src/Sylius/Behat/Context/Cli/ShowOrderContext.php
@@ -44,7 +44,7 @@ final class ShowOrderContext implements Context
     }
 
     /**
-     * @When I run show order command for order :arg1
+     * @When I run show order command for order :number
      */
     public function runShowOrderCommand(string $number): void
     {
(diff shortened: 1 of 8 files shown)


src/Sylius/Behat/Context/Cli/ShowOrderContext.php
```diff
@@ -0,0 +1,73 @@
...
+    public function __construct(
+        KernelInterface $kernel,
+        OrderRepository $orderRepository,
+    ) {
+        $this->kernel = $kernel;
+        $this->orderRepository = $orderRepository;
+    }
+
+    /**
+     * @When I run show order command for order :arg1
```
lchrusciel, line 47:
```suggestion
     * @When I run show order command for order :number
```

src/Sylius/Bundle/OrderBundle/Command/ShowOrderCommand.php
```diff
@@ -0,0 +1,293 @@
...
+use Symfony\Component\Console\Helper\TableCell;
+use Symfony\Component\Console\Helper\TableCellStyle;
+use Symfony\Component\Console\Helper\TableSeparator;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use Traversable;
+
+class ShowOrderCommand extends Command
```
lchrusciel, line 34:
```suggestion
final class ShowOrderCommand extends Command
```

lchrusciel (approved):
Hey Dan! Welcome back!

Nice feature 👍 I see some benefit for others here. But we should at least mention about it in some docs /cc @CoderMaggie. Also, I'm wondering if we should provide some extensibility of it. On the other hand, I'm not sure it is worth complicating. Anyway, we are waiting for green master

dantleech:
> Also, I'm wondering if we should provide some extensibility of it.

yeah -- i wondered that also, and this feature is a bit open ended. e.g. do we also want to add support for showing products, cusotmers, etc, in which case having a "framework" to render selected entites on the CLI migt be beneficial and it should probably be a separate bundle.

either way should probably be named in a way that facilitates that: `sylius:show:<thing>` rather than `sylius:<thing>:show`

lchrusciel:
After thinking about it, such CLI tooling is superb for Sylius usage as a separate service without any UI for Admin and Shop. Would be great to have it in the core, so I'm looking forward to hear from you about how useful it was, and did you have any recommendations about the naming convention and feature extensibility

jakubtobiasz:
Hi @dantleech!
Thanks for your contribution! We decided to close this PR as we don't see big value in such feature directly in Sylius' core. If we change our mind, we'll back here :). Once again, thanks.

PR #14025, opened 2022-05-28
Maintain | Deprecate `$salesDataProvider` in Admin `DashboardController`

stloyd:
| Branch?         | master
| Deprecations?   | yes  |


src/Sylius/Bundle/AdminBundle/Controller/DashboardController.php
```diff
@@ -33,6 +33,9 @@ public function __construct(
         private ?SalesDataProviderInterface $salesDataProvider = null,
         private ?StatisticsDataProviderInterface $statisticsDataProvider = null
```
lchrusciel, line 34: Probably we should place both services into one argument and provide proper null handling

lchrusciel (approved):

PR #14041, opened 2022-06-01
[Addressing] Prevent deleting zones and provinces that are zone members

coldic3:
| Branch?         | 1.11
| Deprecations?   | no
| Related tickets |                      |

It is possible to delete the resource (zone or province) that is already in use as a zone member. When we delete such a resource, we cannot access the zone edit screen (and probably even more) because we try to load an unexisting resource 💥

Sample error:
```
Unable to transform data for property path "code": Object "Sylius\Component\Addressing\Model\Zone" with identifier "code"="LETS_DELETE_ME" does not exist.
```

### DIFF
--- a/features/addressing/managing_countries/managing_provinces_of_country.feature
+++ b/features/addressing/managing_countries/managing_provinces_of_country.feature
@@ -26,6 +26,24 @@ Feature: Managing provinces of a country
         And this country should not have the "Northern Ireland" province
 
     @ui @javascript @api
+    Scenario: Removing a province that is a zone member should not be possible
+        Given this country has the "Northern Ireland" province with "GB-NIR" code
+        And this country also has the "Scotland" province with "GB-SCT" code
+        And this country also has the "England" province with "GB-ENG" code
+        And the store has a zone "Great Britain" with code "GB"
+        And it has the "Northern Ireland" province member
+        And it also has the "Scotland" province member
+        And it also has the "England" province member
+        When I want to edit this country
+        And I delete the "Northern Ireland" province of this country
+        And I also delete the "Scotland" province of this country
+        And I save my changes
+        Then I should be notified that provinces that are in use cannot be deleted
+        And this country should still have the "Northern Ireland" province
+        And this country should still have the "Scotland" province
+        And this country should still have the "England" province
+
+    @ui @javascript @api
     Scenario: Removing and adding a new province to an existing country
         Given this country has the "Northern Ireland" province with "GB-NIR" code
         When I want to edit this country
--- a/features/addressing/managing_zones/deleting_zone.feature
+++ b/features/addressing/managing_zones/deleting_zone.feature
@@ -7,18 +7,27 @@ Feature: Deleting a zone
     Background:
         Given the store has a zone "North America" with code "NA"
         And the store also has a zone "South America" with code "SA"
+        And the store also has a zone "Central and Eastern Europe" with code "CEE"
+        And the store also has a zone "Europe" with code "EU"
+        And it has the zone named "Central and Eastern Europe"
         And the store has a tax category "Sports gear"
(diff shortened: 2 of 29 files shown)


src/Sylius/Behat/Context/Api/Admin/ManagingCountriesContext.php
```diff
@@ -348,6 +348,17 @@ public function iShouldBeNotifiedThatNameOfTheProvinceIsRequired(): void
...
     }
 
+    /**
+     * @Then I should be notified that provinces that are in use cannot be deleted
+     */
+    public function iShouldBeNotifiedThatProvincesThatAreInUseCannotBeDeleted(): void
+    {
+        Assert::contains(
+            $this->responseChecker->getError($this->client->getLastResponse()),
+            'Cannot delete, the province is in use.'
```
Ferror, line 358: This error message and `Error Cannot delete, the province is in use.` src/Sylius/Behat/Context/Ui/Admin/ManagingCountriesContext.php:284 shouldn't be more likely the same?

coldic3: "Error" before the error message is something that exists only in flash messages as a message header.
image

I think it's a good idea to compare just a message in the UI context but I would be for doing it in separate PR as it requires changes in many behat contexts.

features/addressing/managing_countries/managing_provinces_of_country.feature
```diff
@@ -25,6 +25,17 @@ Feature: Managing provinces of a country
...
         And this country should not have the "Northern Ireland" province
 
+    @ui @javascript @api
+    Scenario: Removing a province that is a zone member should not be possible
+        Given this country has the "Northern Ireland" province with "GB-NIR" code
+        And the store has a zone "All in GB" with code "GB"
+        And it has the "Northern Ireland" province member
+        When I want to edit this country
+        And I delete the "Northern Ireland" province of this country
+        And I save my changes
```
AdamKasp, line 35:
```suggestion
        And I try to save my changes
```

src/Sylius/Bundle/AddressingBundle/EventListener/ZoneMemberIntegrityListener.php
```diff
@@ -0,0 +1,68 @@
...
+
+            $event->stopPropagation();
+        }
+    }
+
+    public function protectFromRemovingProvinceWithinCountry(GenericEvent $event): void
+    {
+        /** @var CountryInterface $country */
+        $country = $event->getSubject();
+        Assert::isInstanceOf($country, CountryInterface::class);
```
GSadee, line 55: Shouldn't we check to remove the subject that is `ProvinceInterface`?

features/addressing/managing_zones/deleting_zone.feature
```diff
@@ -22,3 +25,9 @@ Feature: Deleting a zone
         When I delete zone named "South America"
         Then I should be notified that this zone cannot be deleted
         And I should still see the zone named "South America" in the list
+
+    @ui @api
+    Scenario: Deleting zone that is a zone member should not be possible
+        When I delete zone named "Central and Eastern Europe"
```
AdamKasp, line 31:
```suggestion
        When I try to delete zone named "Central and Eastern Europe"
```

src/Sylius/Bundle/AddressingBundle/EventListener/ZoneMemberIntegrityListener.php
coldic3: The problem is we don't delete the Province directly but `orphanRemoval` configured in relation within a Country does it for us. That makes me unable to listen for an event `sylius.province.pre_delete` and I had to use `sylius.country.pre_update`.

GSadee (approved):

GSadee:

PR #14055, opened 2022-06-04
Fix channel color in fixture

iKsSs:
| Branch?         | 1.10        |
| Deprecations?   | no  |

If there is not color in HEX on the Channel, the color will be displayed incorrectly when editing the Channel.
The default (black) color is prefilled in the form and the old value is overwritten after saving.


jakubtobiasz (approved):

jakubtobiasz:
Thank you, @iKsSs!

PR #14058, opened 2022-06-06
[Admin][Zones] Choosing disabled country as a zone member made possible for admin user

TheMilek:
| Branch?         | 1.11           |
| Deprecations?   | no  |

Right now we’re experiencing a bug-o-feature, where when you disable a country that is a zone member, the members list on the Zone shows empty dropdown and field is unable to be controlled.

Changes made:

- adding disabled countries as zone members made possible for admin users
- search dropdown added instead of select dropdown field

Before:
502e0587-5f2a-46e6-bf0a-978a51e0462c
After:
Zrzut ekranu 2022-06-7 o 15 24 15

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
+++ b/src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
@@ -41,13 +41,11 @@ final class CountryChoiceType extends AbstractType
             ->setDefaults([
                 'choice_filter' => null,
                 'choices' => function (Options $options): iterable {
-                    if (null === $options['enabled']) {
-                        $countries = $this->countryRepository->findAll();
-                    } else {
-                        $countries = $this->countryRepository->findBy(['enabled' => $options['enabled']]);
+                    if ($options['enabled'] === true) {
+                        return  $this->countryRepository->findBy(['enabled' => $options['enabled']]);
                     }
 
-                    return $countries;
+                    return $this->countryRepository->findAll();
                 },
                 'choice_value' => 'code',
                 'choice_label' => 'name',
--- a/src/Sylius/Bundle/AddressingBundle/Tests/Form/Type/CountryChoiceTypeTest.php
+++ b/src/Sylius/Bundle/AddressingBundle/Tests/Form/Type/CountryChoiceTypeTest.php
@@ -33,6 +33,9 @@ final class CountryChoiceTypeTest extends TypeTestCase
     /** @var ProphecyInterface|CountryInterface */
     private $poland;
 
+    /** @var ProphecyInterface|CountryInterface */
+    private $austria;
+
     protected function setUp(): void
     {
         $this->countryRepository = $this->prophesize(RepositoryInterface::class);
@@ -49,6 +52,13 @@ final class CountryChoiceTypeTest extends TypeTestCase
         $poland->getName()->willReturn('Poland');
         $this->poland = $poland;
 
+        /** @var ProphecyInterface|CountryInterface $austria */
+        $austria = $this->prophesize(CountryInterface::class);
+        $austria->getCode()->willReturn('AT');
+        $austria->getName()->willReturn('Austria');
+        $austria->isEnabled()->willReturn(false);
+        $this->austria = $austria;
+
         parent::setUp();
     }
 
@@ -61,9 +71,7 @@ final class CountryChoiceTypeTest extends TypeTestCase
         ];
     }
 
-    /**
-     * @test
-     */
+    /** @test */
     public function it_returns_only_enabled_countries_by_default(): void
     {
         $this->countryRepository->findBy(['enabled' => true])->willReturn([
@@ -74,23 +82,20 @@ final class CountryChoiceTypeTest extends TypeTestCase
         $this->assertChoicesLabels(['France', 'Poland']);
     }
 
-    /**
-     * @test
-     */
-    public function it_returns_all_countries(): void
+    /** @test */
+    public function it_returns_all_countries_when_option_enabled_is_false(): void
     {
         $this->countryRepository->findAll()->willReturn([
             $this->france->reveal(),
             $this->poland->reveal(),
+            $this->austria->reveal(),
         ]);
 
-        $this->assertChoicesLabels(['France', 'Poland'], ['enabled' => null]);
+        $this->assertChoicesLabels(['Austria','France', 'Poland'], ['enabled' => false]);
     }
 
-    /**
-     * @test
-     */
-    public function it_returns_countries_in_an_alphabetical_order(): void
+    /** @test */
+    public function it_returns_enabled_countries_in_an_alphabetical_order(): void
     {
         $this->countryRepository->findBy(['enabled' => true])->willReturn([
             $this->poland->reveal(),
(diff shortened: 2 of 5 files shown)


src/Sylius/Bundle/AddressingBundle/Tests/Form/Type/CountryChoiceTypeTest.php
```diff
@@ -87,32 +92,56 @@ public function it_returns_all_countries(): void
...
+        $this->countryRepository->findBy(['enabled' => true])->willReturn([
+            $this->france->reveal(),
+            $this->poland->reveal(),
+        ]);
+
+        $this
+            ->assertChoicesLabels(['Poland'], ['choice_filter' => static fn (?CountryInterface $country): bool => $country !== null && $country->getName() === 'Poland',
+            'enabled' => true,
+            'multiple' => true
+        ]);
```
coldic3, line 142:
```suggestion
        $this
            ->assertChoicesLabels(['Poland'], [
                'choice_filter' => static fn (?CountryInterface $country): bool => $country !== null && $country->getName() === 'Poland',
                'enabled' => true,
                'multiple' => true
            ])
        ;
```

src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
```diff
@@ -41,14 +41,24 @@ public function configureOptions(OptionsResolver $resolver): void
...
-                    if (null === $options['enabled']) {
-                        $countries = $this->countryRepository->findAll();
-                    } else {
+                    if (null !== $options['enabled'] && $options['multiple']) {
                         $countries = $this->countryRepository->findBy(['enabled' => $options['enabled']]);
+                    } else {
+                        $countries = $this->countryRepository->findAll();
                     }
 
                     return $countries;
```
GSadee, line 50:
```suggestion
                    if (null !== $options['enabled'] && $options['multiple']) {
                        $return $this->countryRepository->findBy(['enabled' => $options['enabled']]);
                    }

                    return $this->countryRepository->findAll();
```

src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
```diff
@@ -41,14 +41,24 @@ public function configureOptions(OptionsResolver $resolver): void
             ->setDefaults([
                 'choice_filter' => null,
                 'choices' => function (Options $options): iterable {
-                    if (null === $options['enabled']) {
-                        $countries = $this->countryRepository->findAll();
-                    } else {
+                    if (null !== $options['enabled'] && $options['multiple']) {
```
GSadee, line 44: Instead of changing the logic here by adding a condition with `multiple` here, you can probably set the proper value of `enabled` in the appropriate form type

src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
```diff
@@ -41,14 +41,24 @@ public function configureOptions(OptionsResolver $resolver): void
...
                 },
+                'attr' => static function (Options $options): array {
+                    if (!$options['multiple']) {
+
+                        return ['class' => 'sylius_zone_member_country_select ui fluid search selection dropdown'];
+                    }
+                    else {
+
+                        return [];
+                    }
```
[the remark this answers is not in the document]
Zales0123: 👍 however I have mixed feelings about it generally. We're providing strictly UI- and SemanticUI-oriented data (names of classes) in the form type which theoretically could be used in any other template :dancer: We need to think about how to safely move from this nonsense (without BC breaks). Still, nothing to change now 🖖

src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
```diff
@@ -41,18 +41,16 @@ public function configureOptions(OptionsResolver $resolver): void
             ->setDefaults([
                 'choice_filter' => null,
                 'choices' => function (Options $options): iterable {
-                    if (null === $options['enabled']) {
-                        $countries = $this->countryRepository->findAll();
-                    } else {
-                        $countries = $this->countryRepository->findBy(['enabled' => $options['enabled']]);
+                    if ($options['enabled'] === true) {
```
GSadee, line 44:
```suggestion
                    if ($options['enabled']) {
```

src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
```diff
@@ -41,18 +41,16 @@ public function configureOptions(OptionsResolver $resolver): void
...
                     }
 
-                    return $countries;
+                    return $this->countryRepository->findAll();
                 },
                 'choice_value' => 'code',
                 'choice_label' => 'name',
                 'choice_translation_domain' => false,
-                'enabled' => true,
+                'enabled' => false,
```
GSadee, line 53: Revert this change and leave it as it was before not to change the default behaviour, because this `CountryChoiceType` can be used in many places. Instead of changing it here, set the proper value of this field in the appropriate form: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/AddressingBundle/Form/Type/ZoneType.php#L104

src/Sylius/Bundle/AddressingBundle/Tests/Form/Type/CountryChoiceTypeTest.php
```diff
@@ -49,6 +52,12 @@ protected function setUp(): void
         $poland->getName()->willReturn('Poland');
         $this->poland = $poland;
 
+        /** @var ProphecyInterface|CountryInterface $austria */
+        $austria = $this->prophesize(CountryInterface::class);
+        $austria->getCode()->willReturn('AT');
+        $austria->getName()->willReturn('Austria');
+        $this->austria = $austria;
```
GSadee, line 59: Shouldn't this country be disabled for the tests?

src/Sylius/Bundle/CoreBundle/Form/Extension/ChannelTypeExtension.php
```diff
@@ -57,6 +57,7 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                 'label' => 'sylius.form.channel.countries',
                 'required' => false,
                 'multiple' => true,
+                'enabled' => true,
```
GSadee, line 60: Shouldn't be needed after my above suggestions

src/Sylius/Bundle/AddressingBundle/Form/Type/CountryChoiceType.php
TheMilek: Actually I would stay with checking equality, because without it, it is also checking existance of the key, so for example
`'enabled' => false` will be in this case true

src/Sylius/Bundle/AddressingBundle/Form/Type/CountryCodeChoiceType.php
```diff
@@ -30,6 +31,16 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
         $builder->addModelTransformer(new ReversedTransformer(new ResourceToIdentifierTransformer($this->countryRepository, 'code')));
     }
 
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver
+            ->setDefaults([
+                'attr' => ['class' => 'country_search_dropdown ui fluid search selection dropdown'],
+                'enabled' => 'false',
```
GSadee, line 39: I can add here almost the same comment as in the previous iteration: https://github.com/Sylius/Sylius/pull/14058#discussion_r891936298

`CountryCodeChoiceType` is used in several different places if I'm correct, but you want to change the behaviour only for selecting countries for zones, so this parameter should be disabled only in `ZoneType`, at least for now

GSadee: I'm almost sure that if you will change this line: https://github.com/Sylius/Sylius/blob/master/src/Sylius/Bundle/AddressingBundle/Form/Type/ZoneType.php#L104 by adding there an `enabled` parameter, it will work correctly:
```
ZoneInterface::TYPE_COUNTRY => ['label' => 'sylius.form.zone.types.country', 'enabled' => false],
```
And you will not need to add a custom form type. Did you check this approach?

GSadee (approved):

GSadee:
Thank you, Kamil! :tada:

Issue #13274, opened 2021-11-05
Symfony 6 support


loic425:
### Internal Packages

- [x] Addressing
- [x] Attribute
- [x] Channel
- [x] Core
    - [x] Sylius/Sylius#14078 
- [x] Currency
- [x] Customer
- [x] Inventory
- [x] Locale
- [x] Order
- [x] Payment
- [x] Product
- [x] Promotion
- [x] Review
- [x] Shipping
- [x] Taxation
- [x] Taxonomy
- [x] User
    - [x] Sylius/Sylius#14068  

### Internal Bundle

- [x] AddressingBundle https://github.com/Sylius/Sylius/runs/6809340436?check_suite_focus=true
    - [x] Sylius/Sylius#13975
    - [x] Sylius/Sylius#14034 
    - [x] Sylius/Sylius#14065
- [x] AdminBundle (🦀 needs sylius/core-bundle) https://github.com/Sylius/Sylius/runs/6776689949?check_suite_focus=true#step:8:19
    - [x] Sylius/Sylius#14191 
    - [x] Sylius/Sylius#14192
    - [x] Sylius/Sylius#14287 
- [x] ApiBundle
    - [x] Sylius/Sylius#14042
    - [x] Sylius/Sylius#14284
- [x] AttributeBundle
    - [x] Sylius/Sylius#13976
- [x] ChannelBundle
    - [x] Sylius/Sylius#13977 
- [x] CoreBundle (🦀 needs sylius/order-bundle, and sylius/theme-bundle)
    - [x] Sylius/Sylius#14184 
    - [x] Sylius/Sylius#14136
- [x] CurrencyBundle
    - [x] Sylius/Sylius#13978 
- [x] CustomerBundle
    - [x] Sylius/Sylius#13979 
- [x] InventoryBundle
    - [x] Sylius/Sylius#13980 
- [x] LocaleBundle
    - [x] Sylius/Sylius#13981 
- [x] MoneyBundle
    - [x] Sylius/Sylius#13982 
- [x] OrderBundle (:heavy_check_mark: Fixed with following PRs)
    - [x] Sylius/Sylius#14066
    - [x] Sylius/Sylius#14113
- [x] PaymentBundle
    - [x] Sylius/Sylius#13983
- [x] PayumBundle  (:heavy_check_mark: Fixed with this following PR)
    - [x] Sylius/Sylius#14182
- [x] ProductBundle
    - [x] Sylius/Sylius#13989
[...]

PR #13969, opened 2022-05-13
Symfony 6 support

loic425:
| Q               | A                                                            |
|-----------------|--------------------------------------------------------------|
| Branch?         | master          |
| Bug fix?        | no                                                       |
| New feature?    | yes                                                       |
| BC breaks?      | hope no                                                       |
| Deprecations?   | no |
| Related tickets | fixes #13274                     |
| License         | MIT                                                          |

PR #14074, opened 2022-06-10
[Symfony 6] Fix user bundle

loic425:
| Branch?         | symfony 6         |
| Bug fix?        | yes (for Symfony 6)                                                      |
| BC breaks?      | hope no                                                       |
| Deprecations?   | no |
| Related tickets | partially #13274                       |

This is fixed here https://github.com/Sylius/Sylius/runs/6986299100?check_suite_focus=true

### DIFF
--- a/src/Sylius/Bundle/UserBundle/Provider/AbstractUserProvider.php
+++ b/src/Sylius/Bundle/UserBundle/Provider/AbstractUserProvider.php
@@ -17,7 +17,7 @@ use Sylius\Component\User\Canonicalizer\CanonicalizerInterface;
 use Sylius\Component\User\Model\UserInterface as SyliusUserInterface;
 use Sylius\Component\User\Repository\UserRepositoryInterface;
 use Symfony\Component\Security\Core\Exception\UnsupportedUserException;
-use Symfony\Component\Security\Core\Exception\UsernameNotFoundException;
+use Symfony\Component\Security\Core\Exception\UserNotFoundException;
 use Symfony\Component\Security\Core\User\UserInterface;
 
 abstract class AbstractUserProvider implements UserProviderInterface
@@ -38,7 +38,7 @@ abstract class AbstractUserProvider implements UserProviderInterface
         $user = $this->findUser($username);
 
         if (null === $user) {
-            throw new UsernameNotFoundException(
+            throw new UserNotFoundException(
                 sprintf('Username "%s" does not exist.', $username)
             );
         }
@@ -68,7 +68,7 @@ abstract class AbstractUserProvider implements UserProviderInterface
         /** @var UserInterface|null $reloadedUser */
         $reloadedUser = $this->userRepository->find($user->getId());
         if (null === $reloadedUser) {
-            throw new UsernameNotFoundException(
+            throw new UserNotFoundException(
                 sprintf('User with ID "%d" could not be refreshed.', $user->getId())
             );
         }
--- a/src/Sylius/Bundle/UserBundle/Provider/AbstractUserProvider.php
+++ b/src/Sylius/Bundle/UserBundle/Provider/AbstractUserProvider.php
@@ -16,8 +16,8 @@ namespace Sylius\Bundle\UserBundle\Provider;
 use Sylius\Component\User\Canonicalizer\CanonicalizerInterface;
 use Sylius\Component\User\Model\UserInterface as SyliusUserInterface;
 use Sylius\Component\User\Repository\UserRepositoryInterface;
+use SyliusLabs\Polyfill\Symfony\Security\Core\Exception\UserNotFoundException;
 use Symfony\Component\Security\Core\Exception\UnsupportedUserException;
-use Symfony\Component\Security\Core\Exception\UserNotFoundException;
 use Symfony\Component\Security\Core\User\UserInterface;
 
 abstract class AbstractUserProvider implements UserProviderInterface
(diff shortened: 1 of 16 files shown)


src/Sylius/Bundle/UserBundle/Provider/AbstractUserProvider.php
```diff
@@ -38,7 +38,7 @@ public function loadUserByUsername($username): UserInterface
         $user = $this->findUser($username);
 
         if (null === $user) {
-            throw new UsernameNotFoundException(
+            throw new UserNotFoundException(
```
loic425, line 41: @Zales0123 This may be a bc-break, but we can't create a polyfill cause it would not work on try catch.

src/Sylius/Bundle/UserBundle/Tests/DependencyInjection/Compiler/RemoveUserPasswordEncoderPassTest.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $this->registerService('sylius.security.password_encoder', UserPasswordEncoder::class);
+
+        $this->compile();
+
+        $this->assertContainerBuilderNotHasService('sylius.security.password_encoder');
+    }
+
+    /** @test */
+    public function it_does_removes_user_password_encoder_definition_on_symfony_lower_than_6(): void
```
loic425, line 39:
```suggestion
    public function it_does_not_removes_user_password_encoder_definition_on_symfony_lower_than_6(): void
```

src/Sylius/Bundle/UserBundle/Provider/AbstractUserProvider.php
HeahDude: It shouldn't be considered a BC break IMO, given that any concerned `catch` block would emit a proper deprecation already (https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Security/Core/Exception/UsernameNotFoundException.php#L14).

src/Sylius/Bundle/UserBundle/Security/UserLogin.php
```diff
@@ -39,7 +39,7 @@ public function login(UserInterface $user, ?string $firewallName = null): void
         $this->userChecker->checkPostAuth($user);
 
         $token = $this->createToken($user, $firewallName);
-        if (!$token->isAuthenticated()) {
+        if (null === $token->getUser() || [] === $token->getUser()->getRoles()) {
```
HeahDude, line 42: Why keeping `[] === $token->getUser()->getRoles()` here?

Authorization could still happen and be legit with any voter but the default Symfony `RoleVoter`, given there is a proper user.
(https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Security/Core/Authentication/Token/TokenInterface.php#L74).

loic425: I've added that because of this current spec => https://github.com/Sylius/Sylius/blob/fa4c1d30dd9f94f4c8bb4e80b8fb3208fdf38c62/src/Sylius/Bundle/UserBundle/spec/Security/UserLoginSpec.php#L73

But maybe we could remove it.

src/Sylius/Bundle/UserBundle/Provider/AbstractUserProvider.php
HeahDude: However, given the current approach (https://github.com/Sylius/Sylius/pull/13969#issuecomment-1159379884), a way to let ppl opt-in into the new behavior could be to add a constructor argument in this class to be defined via bundle config in order to throw the expected exception.
This deprecation layer (arg + config) could then be deprecated in the next major of Sylius when 5.4 would be dropped.

src/Sylius/Bundle/UserBundle/Security/UserLogin.php
lchrusciel: as we have this test, lets keep it

src/Sylius/Bundle/UserBundle/Provider/AbstractUserProvider.php
lchrusciel: we can also add our custom exception, which will implement `UsernameNotFoundException` if exists and `UsernameNotFoundException` otherwise. This way, the default behaviour will stay the same and Sf6 will already cover requirement of having this class.

loic425: Indeed, I remember having this idea, but I wanted yours 👍

HeahDude: >which will implement `UsernameNotFoundException` if exists and `UsernameNotFoundException` otherwise

Both classes exist in 5.4, so we cannot rely on a class existence check, since we cannot detect which one is used in the `catch` in user-land code, this needs to be controlled and opt-in by the end-user.

lchrusciel: but we are maintaining BC - if you had the code that was working up to this moment, nothing changes. We do not trigger new exception tho. With the current implementation, you will be always able to migrate to polifiled class (however it should be discouraged). Once moved to Sf6, the previous exception won't be existing, so you had to changed your code already. Am I missing something?

Also, don't get me wrong, I don't want to insist too much. I'm just looking for flaws in what we have now :)

lchrusciel (approved):

PR #14079, opened 2022-06-13
[Symfony 6] Fix Core bundle

loic425:
| Branch?         | symfony-6          |
| Bug fix?        | yes (for Symfony 6)                                                       |
| Deprecations?   | no |
| Related tickets | partially #13274                     |


src/Sylius/Bundle/CoreBundle/EventListener/TaxonDeletionListener.php
```diff
@@ -27,11 +28,15 @@ final class TaxonDeletionListener
...
     public function __construct(
-        private SessionInterface $session,
+        private SessionInterface|RequestStack $requestStackOrSession,
         private ChannelRepositoryInterface $channelRepository,
         TaxonAwareRuleUpdaterInterface ...$ruleUpdaters
     ) {
         $this->ruleUpdaters = $ruleUpdaters;
+
+        if ($requestStackOrSession instanceof SessionInterface) {
+            @trigger_error(sprintf('Passing an instance of %s as constructor argument for %s is deprecated as of Sylius 1.12 and will be removed in 2.0. Pass an instance of %s instead.', SessionInterface::class, self::class, RequestStack::class), \E_USER_DEPRECATED);
```
HeahDude, line 38: What about using https://github.com/symfony/symfony/blob/6.2/src/Symfony/Contracts/Deprecation/function.php#L23?

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
```diff
@@ -69,7 +69,7 @@ public function loadUserByOAuthUserResponse(UserResponseInterface $response): Us
             return $this->createUserByOAuthUserResponse($response);
         }
 
-        throw new UsernameNotFoundException('Email is null or not provided');
+        throw new UserNotFoundException('Email is null or not provided');
```
HeahDude, line 72: Sorry if I repeat myself, but we may have a BC break here.

This polyfill is not enough as there is no way to know which exception is expected in a user-land `catch`.

Instead we could have a constructor arg (like a boolean flag) defined in a compiler pass that would check whether or not the new authentication system is wired (with `enable_authenticator_manager` set to `true` in the project configuration), it could be done by checking the existence of a definition of a service from the new layer.

src/Sylius/Bundle/CoreBundle/spec/OAuth/UserProviderSpec.php
```diff
@@ -223,7 +224,7 @@ function it_should_throw_exception_when_no_email_was_provided(
         $response->getEmail()->willReturn(null);
         $userRepository->findOneByEmail(Argument::any())->shouldNotBeCalled();
 
-        $this->shouldThrow(UsernameNotFoundException::class)->during('loadUserByOAuthUserResponse', [$response]);
+        $this->shouldThrow(UserNotFoundException::class)->during('loadUserByOAuthUserResponse', [$response]);
```
HeahDude, line 227: So we're missing a test here to be able to opt-in. This is a BC break.

src/Sylius/Bundle/CoreBundle/EventListener/TaxonDeletionListener.php
loic425: Thanks, I did it in this PR => https://github.com/Sylius/Sylius/pull/14136

src/Sylius/Bundle/CoreBundle/OAuth/UserProvider.php
loic425: IMHO, there is nothing with new authenticator system, that was not thrown by Symfony here.
Before, we threw UsernameNotFoundAction which is removed so, we try to add a bc-layer to keep the catch working for an end user if his project is on Symfony 5.4. But if he upgrades on SF6, he must change his catch.
@lchrusciel do you remember our discussions about this deprecation?

lchrusciel: This is what we have been doing for some time already. Until the contract is valid we preserve it unless it is not possible to preserve it. For me, the easiest way to solve it is to write an UPGRADE file which will be recommending the usage of `try catch` with both exceptions, as you may have no existent classes in a catching part, with the union pipe.

However, as @HeahDude suggests a boolean flag is a viable option. Probably even better from the DX point of view, but a noticeable debt from the maintainer POV

src/Sylius/Bundle/CoreBundle/spec/OAuth/UserProviderSpec.php
lchrusciel: The longer I think about it, I'm more convinced it is not. Adding class in the middle is the backward compatible change - https://github.com/SyliusLabs/PolyfillSymfonySecurity/blob/master/src/Core/Exception/UserNotFoundException.php. Therefore, we are free to do it. As long as you are on Sf5.4 your old code will work with an update here. Once you will bump to Sf6.0, the class will change but the previous won't exist anyway, so you had to adjust class anyway.

The valid point, you brought is that people may be using a new exception in Sf5.4 - but that's not exactly the case for us in my opinion. First of all, usage of the new exception will not work (as we are triggering the old one anyway). If you want to prepare your codebase for the next Symfony version you still have at least two options:
```php
try {

} catch (Symfony\Component\Security\Core\Exception\UsernameNotFoundException|Symfony\Component\Security\Core\Exception\UserNotFoundException) {}
```
or
```php
try {

} catch (SyliusLabs\Polyfill\Symfony\Security\Core\Exception\UserNotFoundException) {}
```

which will behave quite similar.

My doubts about the new argument are that we would need to maintain this parameter for the next major version. In addition, it will be problematic to remove it in backward compatible way

loic425:
Closed in favour of #14184

PR #14080, opened 2022-06-13
[Symfony 6] Use Password hasher

loic425:
| Branch?         | master          |
| BC breaks?      | hope no                                                       |
| Deprecations?   | hope no |
| Related tickets | partially #13274                       |

The goal is to remove usage of deprecated (and removed on Symfony 6) UserPasswordEncoder.


src/Sylius/Component/User/Security/PasswordUpdater.php
```diff
@@ -17,14 +17,14 @@
 
 final class PasswordUpdater implements PasswordUpdaterInterface
 {
-    public function __construct(private UserPasswordEncoderInterface $userPasswordEncoder)
+    public function __construct(private UserPasswordHasherInterface $userPasswordHasher)
```
loic425, line 20: @lchrusciel Is this a bc-break?

loic425: I've tried to bring a bc-layer.

src/Sylius/Component/User/Security/UserPasswordHasherInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Component\User\Security;
+
+use Sylius\Component\User\Model\CredentialsHolderInterface;
+
+interface UserPasswordHasherInterface
```
lchrusciel, line 18: Instead of creating the second interface, can't we just use our previous interface as ACL and just replace the implementation?

src/Sylius/Bundle/UserBundle/Security/UserPasswordHasher.php
```diff
@@ -0,0 +1,39 @@
...
+    public function __construct(private PasswordHasherFactoryInterface $passwordHasherFactory)
+    {
+    }
+
+    public function encode(CredentialsHolderInterface $user): string
+    {
+        return $this->hash($user);
+    }
+
+    public function hash(CredentialsHolderInterface $user): string
```
lchrusciel, line 32: This method is not part of any contract, let's make it private (or introduce a second interface here, implement both of them in this class and eventually we will change the contract with Sylius 2.0, but I would vote for making this function just private)
```suggestion
    private function hash(CredentialsHolderInterface $user): string
```

lchrusciel (approved):

PR #14085, opened 2022-06-14
[Migrations] Minimum price migration fixed

TheMilek:
| Branch?         | 1.11           |
| Deprecations?   | no  |

Now we encounter bug that when we already have some data in the db and we use the prepared migrations they fail with 500 wherever we call the minimum price. It is caused by migration that sets default value to null, we managed to add another migration that changes default value to 0, but still does not update existing values and only new ones are correct

changes:

- new migration added that changes existing `minimum_price` null values to 0

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Migrations/Version20220614124639.php
+++ b/src/Sylius/Bundle/CoreBundle/Migrations/Version20220614124639.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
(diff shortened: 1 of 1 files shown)


src/Sylius/Bundle/CoreBundle/Migrations/Version20220614124639.php
```diff
@@ -0,0 +1,39 @@
...
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20220614124639 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'null values in minimum price field changed to default';
```
GSadee, line 26:
```suggestion
        return 'Null values in minimum price field changed to default';
```

src/Sylius/Bundle/CoreBundle/Migrations/Version20220614124639.php
```diff
@@ -0,0 +1,39 @@
...
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
```
GSadee, line 21: Can be removed:
```suggestion
```

GSadee (approved):

GSadee:
Thank you, Kamil! :tada:

PR #14086, opened 2022-06-14
[OrderProcessing] Parametrize hardcoded states in order processors

Zales0123:
| Branch?         | 1.11   |
| Bug fix?        | yes (DX bug 💃)                                                       |
| BC breaks?      | no |
| Deprecations?   | no |
| Related tickets |  |

With the current implementation, it's impossible to use order processors outside the checkout process, even if it would be a valid scenario (e.g. edit a placed order from the Admin panel). I believe we should avoid any hardcode in if statements (both with _magic_ strings and constants) by parametrizing the checked value or checking the result of service 🖖

### DIFF
--- a/src/Sylius/Component/Core/OrderProcessing/OrderAdjustmentsClearer.php
+++ b/src/Sylius/Component/Core/OrderProcessing/OrderAdjustmentsClearer.php
@@ -21,8 +21,10 @@ final class OrderAdjustmentsClearer implements OrderProcessorInterface
 {
     private array $adjustmentsToRemove;
 
-    public function __construct(array $adjustmentsToRemove = [])
-    {
+    public function __construct(
+        array $adjustmentsToRemove = [],
+        private string $validState = OrderInterface::STATE_CART
+    ) {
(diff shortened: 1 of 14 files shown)


src/Sylius/Component/Core/OrderProcessing/OrderAdjustmentsClearer.php
```diff
@@ -21,8 +21,10 @@ final class OrderAdjustmentsClearer implements OrderProcessorInterface
 {
     private array $adjustmentsToRemove;
 
-    public function __construct(array $adjustmentsToRemove = [])
-    {
+    public function __construct(
+        array $adjustmentsToRemove = [],
+        private string $validState = OrderInterface::STATE_CART
```
GSadee, line 26: Maybe there should be an array as there is a possibility that more than one state will be the valid one? The same applies to other processors, wdyt?

GSadee:
Closing in favour of https://github.com/Sylius/Sylius/pull/14935

PR #14095, opened 2022-06-20
[Fixtures] Add a fixture listener for removing images

GSadee:
| Branch?         | 1.11|
| Bug fix?        | yes?                                                       |
| Deprecations?   | no|
| Related tickets | |

Currently without removing pictures when loading fixtures, the images directory can grow unnecessarily very quickly.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Fixture/Listener/ImagesPurgerListener.php
+++ b/src/Sylius/Bundle/CoreBundle/Fixture/Listener/ImagesPurgerListener.php
@@ -26,7 +26,8 @@ final class ImagesPurgerListener extends AbstractListener implements BeforeSuite
 
     public function beforeSuite(SuiteEvent $suiteEvent, array $options): void
     {
-        $this->filesystem->remove($this->imagesDirectoryPath);
+        $this->filesystem->remove($this->imagesDirectoryPath . '/*');
+        $this->filesystem->touch($this->imagesDirectoryPath . '/.gitkeep');
     }
 
     public function getName(): string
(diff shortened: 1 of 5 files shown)


Zales0123 (commented):
👍 but the problem is we already have a [.gitkeep file](https://github.com/Sylius/Sylius/blob/master/public/media/image/.gitkeep) placed in `public/media/image/` directory, which will also be removed after running the fixtures and therefore would result in changed in the repository :dancer: Maybe it would be better to remove all the things (files and catalogues) from the inside of images directory omitting the `.gitkeep` file? 🤔

GSadee:
> 👍 but the problem is we already have a [.gitkeep file](https://github.com/Sylius/Sylius/blob/master/public/media/image/.gitkeep) placed in `public/media/image/` directory, which will also be removed after running the fixtures and therefore would result in changed in the repository 💃 Maybe it would be better to remove all the things (files and catalogues) from the inside of images directory omitting the `.gitkeep` file? 🤔

I've refactored this code by removing all files from the given directory and creating the `.gitkeep` file after that, as it seems to be the simplest solution to resolve the issue you mentioned, wdyt?

Zales0123 (approved):
👍 I wonder now couldn't it be named `FilesPurgerListener` and land in the SyliusFixturesBundle itself 🚀 It would then be possible to use it e.g. for invoices removal with SyliusInvoicingPlugin

Rafikooo (approved):

AdamKasp (approved):

AdamKasp:
Thanks, Grzegorz! :tada:

PR #13860, opened 2022-04-11
Updating Gaufrette to avoid warning on php 8.1

David-Crty:
| Q               | A
| --------------- | -----
| Branch?         | 1.11
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | fixes https://github.com/KnpLabs/KnpGaufretteBundle/issues/258
| License         | MIT

When using php 8.1 their is a warning about Gaufrette, updating the dep should fix the warning.

PR #14098, opened 2022-06-21
Make compatible with knplabs/gaufrette up to 0.10

esserj:
there are no breaking changes introduced in either version , but due
to how sub v1 version management works, only minors
are installed when you require with ^

| Branch?         | 1.11
| Deprecations?   | no

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "friendsofsymfony/rest-bundle": "^3.0",
         "gedmo/doctrine-extensions": "^3.2",
         "jms/serializer-bundle": "^3.5",
-        "knplabs/gaufrette": "^0.8",
+        "knplabs/gaufrette": "^0.8|^0.9|^0.10",
         "knplabs/knp-gaufrette-bundle": "^0.7",
         "knplabs/knp-menu": "^3.1",
         "knplabs/knp-menu-bundle": "^3.0",
--- a/src/Sylius/Bundle/CoreBundle/composer.json
+++ b/src/Sylius/Bundle/CoreBundle/composer.json
@@ -31,7 +31,7 @@
         "egulias/email-validator": "^3.0",
         "fakerphp/faker": "^1.10",
         "jms/serializer-bundle": "^3.5",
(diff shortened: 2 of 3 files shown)


mbabker:
> there are no breaking changes introduced in either version

Not 100% true.  The API is mostly compatible (one public method did get removed along the way), but 0.10 removed a number of adapters.  This should probably land in 1.12 with a note in the UPGRADE guide about it so folks know they'll either need to pin their apps to `knplabs/gaufrette:0.9.*` or upgrade their app to use another adapter (i.e. switch AmazonS3 to AwsS3 or AsyncAws).

esserj:
Ah yes, fair enough @mbabker, nothing breaks in sylius by default, but if you are using one of these deprecated adapters to store your sylius images than it would break your application should you choose to update all

On the other hand, 1.11 should then probably not allow installing on 8.1 as gaufrette cannot deal with that
as mentioned in https://github.com/Sylius/Sylius/pull/13860 (did not verify myself though)

What do you think @Zales0123 ?

1. Move the update to 1.12 with an upgrade note and update 1.11 to `"php": "8.0*"`
2. Move the update to 1.12 with an upgrade note and without update of 1.11 to `"php": "8.0*"`
3. Update 1.11 with an upgrade note `1.11.5 to 1.11.6` to lock your gaufrette version (or conflict with `>=0.10`

For the removed method we can pass the advice from gaufrette in the notes
https://github.com/KnpLabs/Gaufrette/commit/6f113b677cf054f222ab66f0a80906ee04b28728#diff-7082035033078c5da48aacb96ae4574188ff4227e213afe9ba276d880ea89d68L71

I would personally prefer updating 1.11, to all gaufrette version 0.8 -> 0.10 and optionally remove 0.8 en 0.9 from 1.12 (although for sylius core there is no downside in allowing them)

Zales0123:
Hello Jan! Sorry for the long response time, it was introduced into 1.11 in the meantime. Thank you very much for your contribution! 🖖 🚀 🏅

PR #14099, opened 2022-06-21
fix MoneyIntToLocalizedStringTransformer ignores comma separated numbers in Catalog Promotions Fixed Discount type field #138844

### LINKED ISSUE #13884: MoneyIntToLocalizedStringTransformer ignores comma separated numbers in Catalog Promotions Fixed Discount type field
**Sylius version affected**: 1.1.11

**Description**  
[Here](https://github.com/Sylius/Sylius/commit/cb69b221c0a82f40c96cc8108cd412b705e47d97#diff-70d1ea3d00703668498eb0b1c89c84bc4dd19bc53e73cb5a2ed53ce7c628aea8) you added transformer to the `amount` field. In Polish locales all numbers are formatted with comma as decimal separator. When user enters any amount with decimal part for example `10,10` then `is_numeric` function in [reverseTransform](https://github.com/Sylius/Sylius/blob/cb69b221c0a82f40c96cc8108cd412b705e47d97/src/Sylius/Bundle/PromotionBundle/Form/DataTransformer/MoneyIntToLocalizedStringTransformer.php#L33) method breaks the transformer as `10,10` value isn't considered as numeric.

obraz

The entered value `10,10` is transformed to `null` and the configuration is incorrectly written to the database.

obraz

In your demo if you enter a comma separated value there you will get 500 internal server error.

Jonathan-Dedi:
| Branch?         | 1.11 |
| Deprecations?   | no |
| Related tickets | fixes #13884                     |

### DIFF
--- a/src/Sylius/Bundle/PromotionBundle/Form/DataTransformer/MoneyIntToLocalizedStringTransformer.php
+++ b/src/Sylius/Bundle/PromotionBundle/Form/DataTransformer/MoneyIntToLocalizedStringTransformer.php
@@ -19,11 +19,9 @@ final class MoneyIntToLocalizedStringTransformer extends MoneyToLocalizedStringT
 {
     public function reverseTransform($value)
     {
-        if (!is_numeric($value)) {
-            return;
-        }
+        $result = parent::reverseTransform($value);
 
-        return (int) parent::reverseTransform($value);
+        return null !== $result ? (int) $result : null;
     }
 
     public function transform($value)


Zales0123 (commented):
Hello @Jonathan-Dedi and thank you for this proposal 🖖 Would you be able to provide a test that would cover this bug? I see there is no spec for this class 😱 but we should definitely add it when we see something is not working well to avoid the regression. Thanks! 🚀

Zales0123:
Hello @Jonathan-Dedi! 🖖 We tried to reproduce it but it seems to work already 😱 If you feel it's still a bug, feel free to reopen it. Thanks! 🚀

PR #14106, opened 2022-06-22
[Maintenance] Composer allow plugins finally sorted out + removal of symfony thanks

lchrusciel:
| Branch?         | 1.11 |
| Deprecations?   | no |
| Related tickets |  |

I'm proposing adding flex to our allowed plugins, as we are using it and it is in our req/dev. In addition, I've also disallowed `dealerdirect/phpcodesniffer-composer-installer`, as we are not using it directly. 

What is more, I'm proposing the removal of `symfony/thanks` lib, as it does not provide valuable features and is not present in any Symfony package anymore. We are the main package that still installs it https://packagist.org/packages/symfony/thanks/dependents?order_by=downloads&requires=all. Let's give it rest.

These plugins are executed by default right now, but this behaviour will change since the beginning of July. If we don't do anything we can expect the following error: 
image

Once merged, we should iterate over our other packages.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -105,7 +105,6 @@
         "symfony/string": "^5.3",
         "symfony/swiftmailer-bundle": "^3.4",
         "symfony/templating": "^4.4 || ^5.4",
-        "symfony/thanks": "^1.2",
         "symfony/translation": "^4.4 || ^5.4",
         "symfony/twig-bundle": "^4.4 || ^5.4",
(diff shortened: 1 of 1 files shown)


vvasiloi (approved):

GSadee (approved):

GSadee:
Thanks, Łukasz! :tada:

PR #14133, opened 2022-07-07
Disassociate OrderItemUnit from all shipments before removal

kayue:
| Branch?         | 1.10 |
| Deprecations?   | no |
| Related tickets | |

The `$order->removeShipments()` would trigger the following Doctrine error because OrderItemUnit is still associated with shipment.

> A new entity was found through the relationship 'OrderItemUnit#shipment' that was not configured to cascade persist operations for entity: . To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={\"persist\"}).

One `removeShipments()` usage is in `OrderShipmentProcessor` when there is no shipping method available to the order.

### DIFF
--- a/src/Sylius/Component/Core/Model/Order.php
+++ b/src/Sylius/Component/Core/Model/Order.php
@@ -291,6 +291,13 @@ class Order extends BaseOrder implements OrderInterface
 
     public function removeShipments(): void
     {
+        // Disassociate OrderItemUnit from all shipments before removal
+        foreach ($this->shipments as $shipment) {
+            foreach ($shipment->getUnits() as $unit) {
+                $shipment->removeUnit($unit);
+            }
+        }
+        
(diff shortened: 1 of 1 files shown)


lchrusciel:
may become messed-up.**

execute `git fetch origin && git reset "patch-23"` to update your local branch.
 
lchrusciel (commented):
Hey @kayue, thanks for contribution 👌 

I've changed the branch to 1.11, as we do not support bug fixes on 1.10 anymore. Can you also provide some tests for this case?

jakubtobiasz:
Hi @kayue!
Thank you for your contribution. We've finished this PR and merged it in #14651. We used one of your commits, so you're a contributor as well :).

PR #14134, opened 2022-07-07
Fix Percent float to localized string transformer

loic425:
| Q               | A                                                          
|-----------------|--------------------------------------------------------------
| Branch?         | symfony-6
| Bug fix?        | yes                                                       
| New feature?    | no                                                       
| BC breaks?      | no
| Deprecations?   | no
| Related tickets | partially #13274 
| License         | MIT                                                          

It throws an error there https://github.com/Sylius/Sylius/runs/7229517957?check_suite_focus=true#step:15:13 on Sylius/Sylius#14079

It's green here https://github.com/Sylius/Sylius/runs/7229800892?check_suite_focus=true
and there https://github.com/Sylius/Sylius/runs/7229796910?check_suite_focus=true

PR #14135, opened 2022-07-07
Fix data transformers for SF6

loic425:
| Q               | A                                                          
| Branch?         |master        
| Bug fix?        | yes  (for Symfony 6) 
| New feature?    | no                                                       
| BC breaks?      | no                                                       
| Deprecations?   | no
| Related tickets | partially #13274               
| License         | MIT                                                        

I was wrong with the previous PR #14134, it was not enough.
https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Form/Extension/Core/DataTransformer/MoneyToLocalizedStringTransformer.php
https://github.com/symfony/symfony/blob/6.1/src/Symfony/Component/Form/Extension/Core/DataTransformer/PercentToLocalizedStringTransformer.php

### DIFF
--- a/src/Sylius/Bundle/PromotionBundle/Form/DataTransformer/MoneyIntToLocalizedStringTransformer.php
+++ b/src/Sylius/Bundle/PromotionBundle/Form/DataTransformer/MoneyIntToLocalizedStringTransformer.php
@@ -17,19 +17,19 @@ use Symfony\Component\Form\Extension\Core\DataTransformer\MoneyToLocalizedString
 
 final class MoneyIntToLocalizedStringTransformer extends MoneyToLocalizedStringTransformer
 {
-    public function reverseTransform($value)
+    public function reverseTransform(mixed $value): int|float|null
     {
         if (!is_numeric($value)) {
-            return;
+            return null;
         }
 
         return (int) parent::reverseTransform($value);
     }
 
-    public function transform($value)
+    public function transform(mixed $value): string
     {
         if (!is_numeric($value)) {
-            return;
+            return '';
         }
 
         return parent::transform($value);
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/PromotionBundle/Form/DataTransformer/MoneyIntToLocalizedStringTransformer.php
```diff
@@ -17,19 +17,19 @@
...
 
         return (int) parent::reverseTransform($value);
     }
 
-    public function transform($value)
+    public function transform(mixed $value): string
     {
         if (!is_numeric($value)) {
-            return;
+            return '';
```
HeahDude, line 32: Could be a BC break, a property could be nullable in the model data, what about returning `null` and type hinting `: ?string` instead?

src/Sylius/Bundle/PromotionBundle/Form/DataTransformer/PercentFloatToLocalizedStringTransformer.php
```diff
@@ -40,10 +40,10 @@ public function reverseTransform($value)
     /**
      * @param float|string $value
      */
-    public function transform($value)
+    public function transform(mixed $value): string
     {
         if (!is_numeric($value)) {
-            return;
+            return '';
```
HeahDude, line 46: Same here.

src/Sylius/Bundle/PromotionBundle/Form/DataTransformer/MoneyIntToLocalizedStringTransformer.php
lchrusciel: that would violate Symfony 6 contract :/ but! what about fall backing to the Symfony logic instead? In Sf 5.4 we would return the current value and they will handle the return value switch for us? nah... that will break the same case

HeahDude: Thinking again about it, I may be wrong.
There would be no BC break as a nullable property would be type hinted `?string` thus accepting a strict string return from here (hopefully a `NotBlank` constraint may help to behave the same).

lchrusciel: as this class is final, adding returning typehint was not a BC Break - it was an acceptable solution anyway. Returning an empty string instead of a null is changing the class logic, which does not fall directly under BC Policy. Thanks for rising your tensions, tho! It was good to give it a second thoughts

lchrusciel (approved):

PR #14145, opened 2022-07-11
[Maintenance] Add support for doctrine/dbal:^3.0

jakubtobiasz:
| Branch?         | 1.12          |
| Deprecations?   | no  |

This PR adds support for `doctrine/dbal:^3.0`. Some tests are failing and I have to fix them but it's ready to review the approach.

Removing `NULL COMMENT \'(DC2Type:json_array)\'` was necessary because setting up a fresh database using migrations was failing.


jakubtobiasz:
It seems support for dbal3 works (and dbal2, too!). Two areas required changes:
1. Migrations (as they stopped working after upgrade to dbal3)
2. Deprecations

**Migrations**
I had to remove every occurrence of `COMMENT \'(DC2Type:json_array)\'` as it was trying to "cast" column type to `json_array` (removed in dbal3). Due to it, we couldn't perform even a fresh install of Sylius.

**Deprecations**
Psalm was complaining about deprecations, but I decided to fix them as it might be easier to upgrade to a newer version in the future. I decided to create `AbstractMySqlMigration` class and add a logic behind checking if current platform is `MySql` and/or `MariaDB`. Also I had to add some `psalm-suppress` and `phpstan-ignore` annotations because in dbal2 we had `MySqlPlatform` and now in dbal3 we have `MySQLPlatform`.

app/migrations/Version20170116215417.php
```diff
@@ -20,7 +20,7 @@ class Version20170116215417 extends AbstractMigration
 {
     public function up(Schema $schema): void
     {
-        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');
+        $this->abortIf(!is_a($this->connection->getDatabasePlatform(), \Doctrine\DBAL\Platforms\MySqlPlatform::class, true), 'Migration can only be executed safely on \'mysql\'.');
```
GSadee: Why don't you change abstract class as in migrations from CoreBundle?

And maybe you know and you will remind me, because I don't remember, why we need these migrations here too? 🤔

GSadee (approved):

app/migrations/Version20170116215417.php
jakubtobiasz: Good question as I missed the existence of this directory :D. I guess it's a legacy from the times when Symfony 3 was alive :D.

jakubtobiasz: The last migration in this directory is from `16.09.2020` and I'm 99% they're not used. The reason they're changed is... I've used `Find & Replace` throughout the whole project the first time, the second time I had to replace `AbstractMigration` manually so they were left unchanged.

GSadee: I know that migrations existed in this directory before, but when we switched to Doctrine/Migrations 3.0, they were moved to CoreBundle and I'm just trying to remind myself the reason why they stayed here and if we still need them duplicated 🤔 I guess no, but I wanted to see the opinion of others 😃

lchrusciel: I believe, these migrations are here due to BC-Policy, as someone could point out to them from their app

jakubtobiasz: @lchrusciel a few migrations are missing in `app/migrations` so anybody using the old path had to switch to new path anyway 🤔 So should they still be there?

src/Sylius/Bundle/CoreBundle/Doctrine/Migrations/AbstractMySqlMigration.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        $this->abortIf(!$this->isMySql(), 'Migration can only be executed safely on \'mysql\'.');
+    }
+
+    public function preDown(Schema $schema): void
+    {
+        $this->abortIf(!$this->isMySql(), 'Migration can only be executed safely on \'mysql\'.');
+    }
+
+    protected function isMariaDb(): bool
```
lchrusciel: I have some doubts about the naming of this class - we have `AbstractMySqlMigration`, which has methods `isMySql` and `isMySql` - we did something wrong with the naming here

src/Sylius/Bundle/CoreBundle/Doctrine/Migrations/AbstractMySqlMigration.php
```diff
@@ -0,0 +1,70 @@
...
+namespace Sylius\Bundle\CoreBundle\Doctrine\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+abstract class AbstractMySqlMigration extends AbstractMigration
+{
+    public function preUp(Schema $schema): void
+    {
+        $this->abortIf(!$this->isMySql(), 'Migration can only be executed safely on \'mysql\'.');
```
lchrusciel: are we sure, it will work with MariaDB?

app/migrations/Version20170116215417.php
lchrusciel: it is not about missing - people can use them and generate their own. There was something about coping with these migrations to the end app of users IIRC. But then, we moved to the new structure, so if someone wants to use ours, needs to move on as well. 

long story short, we can remove them in Sylius 2.0 in my opinion

app/migrations/Version20161202011555.php
```diff
@@ -62,7 +62,7 @@ public function up(Schema $schema): void
...
         $this->addSql('CREATE TABLE sylius_order_item_unit (id INT AUTO_INCREMENT NOT NULL, order_item_id INT NOT NULL, shipment_id INT DEFAULT NULL, adjustments_total INT NOT NULL, created_at DATETIME NOT NULL, updated_at DATETIME DEFAULT NULL, INDEX IDX_82BF226EE415FB15 (order_item_id), INDEX IDX_82BF226E7BE036FC (shipment_id), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB');
-        $this->addSql('CREATE TABLE sylius_payment (id INT AUTO_INCREMENT NOT NULL, method_id INT DEFAULT NULL, credit_card_id INT DEFAULT NULL, order_id INT NOT NULL, currency_code VARCHAR(3) NOT NULL, amount INT NOT NULL, state VARCHAR(255) NOT NULL, details LONGTEXT NOT NULL COMMENT \'(DC2Type:json_array)\', created_at DATETIME NOT NULL, updated_at DATETIME DEFAULT NULL, INDEX IDX_D9191BD419883967 (method_id), INDEX IDX_D9191BD47048FD0F (credit_card_id), INDEX IDX_D9191BD48D9F6D38 (order_id), PRIMARY KEY(id)) DEFAULT CHARACTER SET utf8 COLLATE utf8_unicode_ci ENGINE = InnoDB');
```
stloyd, line 65: If you remove this comment cause it fails to work, there should be new migration on existing tables, otherwise, those will be broken for Doctrine, no?

app/migrations/Version20170116215417.php
stloyd: `is_a()` is good method, but for this check it's overkill, use simpler:
```suggestion
        $this->abortIf($this->connection->getDatabasePlatform() instanceof \Doctrine\DBAL\Platforms\MySqlPlatform::class, 'Migration can only be executed safely on \'mysql\'.');
```
Also running cs-fixer to import in `use` that class would be a good idea.

src/Sylius/Behat/Context/Domain/ManagingProductsContext.php
```diff
@@ -46,7 +45,7 @@ public function iTryToDeleteTheVariantOfProduct(ProductVariantInterface $product
     {
         try {
             $this->productVariantRepository->remove($productVariant);
-        } catch (DBALException $exception) {
+        } catch (\Exception $exception) {
```
stloyd: This is not the same, old exception still exists, it was renamed to `Doctrine\DBAL\Exception`

src/Sylius/Bundle/CoreBundle/Doctrine/DQL/Day.php
```diff
@@ -36,12 +38,16 @@ public function parse(Parser $parser): void
 
     public function getSql(SqlWalker $sqlWalker): string
     {
-        $platformName = $sqlWalker->getConnection()->getDatabasePlatform()->getName();
+        $platform = $sqlWalker->getConnection()->getDatabasePlatform();
 
-        return match ($platformName) {
```
stloyd, line 41: Better:
```
        return match (get_class($platformName)) {
            MySQLPlatform::class => sprintf('DAY(%s)', $sqlWalker->walkArithmeticPrimary($this->date)),
            PostgreSQLPlatform::class => sprintf('EXTRACT(DAY FROM %s)', $sqlWalker->walkArithmeticPrimary($this->date)),
            default => throw new \RuntimeException(sprintf('Platform "%s" is not supported!', $platformName)),
        };
```
but even that is broken cause `MySQLPlatform::class` was renamed from `MySqlPlatform::class`.

src/Sylius/Bundle/CoreBundle/Doctrine/Migrations/AbstractMySqlMigration.php
jakubtobiasz: Yep. I do a `is_a()` check. In every version `MariaDBPlatform` (or `MariaDb1027Platform` in dbal2) extends `MySQLPlatform` (or `MySqlPlatform` in dbal2) so `$this->isMySql()` will return true for both `MySql` and `MariaDB`. Otherwise our MariaDB workflow would fail :P.

app/migrations/Version20170116215417.php
stloyd: In fact, this is still broken on the newer version cause this class exists `<3.0`, in `3.0+` its `MySQLPlatform`

src/Sylius/Behat/Context/Domain/ManagingProductsContext.php
stloyd: Probably should be:
```suggestion
        } catch (DBALException|Exception $exception) {
```
With old & new class, or two catches...

src/Sylius/Bundle/CoreBundle/Installer/Provider/DatabaseSetupCommandsProvider.php
```diff
@@ -115,7 +115,7 @@ private function getDatabaseName(): string
 
     private function getSchemaManager(): AbstractSchemaManager
     {
-        return $this->getEntityManager()->getConnection()->getSchemaManager();
+        return $this->getEntityManager()->getConnection()->createSchemaManager();
```
stloyd: Should contain `method_exists()` and fallback to the old method.

src/Sylius/Bundle/CoreBundle/Doctrine/Migrations/AbstractMySqlMigration.php
jakubtobiasz: My first thought was
> (with a little bit simplification) we can tell MariaDB is a MySQL, but MySQL is not a MariaDB.

Due to that fact I put these methods in a single class. Now when you noticed it I think we can just call it `AbstractMigration` and extend the base one.

src/Sylius/Bundle/CoreBundle/Doctrine/Migrations/AbstractMySqlMigration.php
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\Doctrine\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+abstract class AbstractMySqlMigration extends AbstractMigration
```
stloyd: Be consistent with Doctrine:
```suggestion
abstract class AbstractMySQLMigration extends AbstractMigration
```

app/migrations/Version20170116215417.php
jakubtobiasz: So I guess it would be nice to change them, too (extend our custom `AbstractMigration` and remove if clauses), am I correct?

stloyd (changes requested):

app/migrations/Version20170116215417.php
jakubtobiasz: While providing the first version I've changed migrations in both places `app/migrations` and `CoreBundle`. While working on "v2" I've omitted accidentally `app/migrations`. So it'll be changed anyway :P.

jakubtobiasz: Migrations in `app/migrations` are updated the same way I've updated those from `CoreBundle`.

src/Sylius/Bundle/CoreBundle/Doctrine/Migrations/AbstractMySqlMigration.php
jakubtobiasz: I've renamed `AbstractMySqlMigration` to just `AbstractMigration` so these two methods won't be confusing anymore.

src/Sylius/Bundle/CoreBundle/Doctrine/DQL/Day.php
jakubtobiasz: I used the `match` expression for the first time but renamed the `MySqlPlatform` class forced me to change the approach to the regular `if`.

src/Sylius/Behat/Context/Domain/ManagingProductsContext.php
jakubtobiasz: Due to the fact they've renamed `\Doctrine\DBAL\DBALException` to `\Doctrine\DBAL\Exception` and we consider it in the context of Behat, I decided to use generic `\Exception` as any exception thrown by `remove()` means something failed. I'm against using such generic types but in this case, I have no value added in handling a specific type (as it'll require two if's checking which one is existing). What do you think @lchrusciel?

app/migrations/Version20161202011555.php
jakubtobiasz: @stloyd 
We had a conversation with Łukasz and we've found if those comments are present and you have `doctrine/dbal: ^3.0` you won't be able to install Sylius. There are some references to repositories in other migrations, then Doctrine creates metadata for the tables in the database and has a problem with figuring out what type is the field.

When we remove this comment it treats those fields as a regular "string" so doesn't throw exceptions about an unknown type. Thanks to it we can process the next migrations where we set the expected type. In the case of already existing projects, those migrations are already marked as executed so they won't fire. Additionally, projects updated to Sylius 1.11 probably have the correct type thanks to migrations from late 2021. 

This solution is battle-field tested (I've added testing dbal2 setup to GitHub Workflow). If you have any questions or you see something what I missed feel free to unresolved this comment :P.

src/Sylius/Behat/Context/Domain/ManagingProductsContext.php
lchrusciel: Luckily, we do not mute this exception entirely but store them in a shared context, so the difference is not that huge. But as we can be closer to the previous catch, don't be too generic and do it without any problem I would vote for these two catches

.github/workflows/application.yml
```diff
@@ -455,6 +455,271 @@ jobs:
...
+                with:
+                    path: ${{ steps.composer-cache.outputs.dir }}
+                    key: ${{ runner.os }}-php-${{ matrix.php }}-symfony-${{ matrix.symfony }}-composer-${{ hashFiles('**/composer.json') }}
+                    restore-keys: |
+                        ${{ runner.os }}-php-${{ matrix.php }}-symfony-${{ matrix.symfony }}-composer-
+
+            -
+                name: Restrict doctrine/dbal version
+                run: |
+                    composer require "doctrine/dbal:^2.0" --no-update --no-scripts --no-interaction
```
lchrusciel: What do you think about adding these two new builds as `include` to previous ones and adding if to this restriction?

src/Sylius/Behat/Context/Domain/ManagingProductsContext.php
jakubtobiasz: @lchrusciel as in the other cases to support both Exceptions we need to add two ifs checking if a concrete exception class exists. In my opinion, it's not worth it.

.github/workflows/application.yml
jakubtobiasz: Great idea, done :). It seems it works but let's wait until all tests finish :D.

GSadee (approved):

src/Sylius/Bundle/CoreBundle/Migrations/Version20220407131547.php
```diff
@@ -57,7 +53,7 @@ public function down(Schema $schema): void
 
     private function getExistingIndexesNames(string $tableName): array
     {
-        $indexes = $this->connection->getSchemaManager()->listTableIndexes($tableName);
+        $indexes = $this->connection->createSchemaManager()->listTableIndexes($tableName);
```
stloyd: Missing fallback.

stloyd (changes requested):

src/Sylius/Bundle/CoreBundle/Migrations/Version20220407131547.php
Zales0123: Should be probably something like [this](https://github.com/Sylius/Sylius/pull/14145/files#diff-0c8fee2942783d5009a5de71149a9eac0f95435e1341e0b18478a150c5284931R116-R130) 🖖

GSadee:
Thanks, Jakub! :1st_place_medal:

PR #14157, opened 2022-07-15
[DX] Load test services in all envs starting with test keyword

lchrusciel:
| Branch?         | 1.12 |
| Deprecations?   | no  |
| Related tickets |  |

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/DependencyInjection/SyliusCoreExtension.php
+++ b/src/Sylius/Bundle/CoreBundle/DependencyInjection/SyliusCoreExtension.php
@@ -57,8 +57,9 @@ final class SyliusCoreExtension extends AbstractResourceExtension implements Pre
         $container->setParameter('sylius_core.taxation.shipping_address_based_taxation', $config['shipping_address_based_taxation']);
         $container->setParameter('sylius_core.catalog_promotions.batch_size', $config['catalog_promotions']['batch_size']);
 
+        /** @var string $env */
         $env = $container->getParameter('kernel.environment');
-        if ('test' === $env || 'test_cached' === $env) {
+        if (str_starts_with($env, 'test')) {
             $loader->load('test_services.xml');
         }
 


composer.json
```diff
@@ -34,7 +34,7 @@
         "behat/transliterator": "^1.3",
         "doctrine/collections": "^1.6",
         "doctrine/common": "^3.2",
-        "doctrine/dbal": "^2.7",
+        "doctrine/dbal": "^3.0",
```
stloyd: That's quite a bold change for PR saying about testing.

src/Sylius/Bundle/CoreBundle/DependencyInjection/SyliusCoreExtension.php
```diff
@@ -58,7 +58,7 @@ public function load(array $configs, ContainerBuilder $container): void
         $container->setParameter('sylius_core.catalog_promotions.batch_size', $config['catalog_promotions']['batch_size']);
 
         $env = $container->getParameter('kernel.environment');
-        if ('test' === $env || 'test_cached' === $env) {
+        if (str_starts_with($env, 'test')) {
```
stloyd, line 62: This change doesn't mean the same as it was before, it's way more generic.

src/Sylius/Bundle/CoreBundle/Migrations/Version20220711145410.php
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Sylius\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20220711145410 extends AbstractMigration
```
stloyd: Hmmm? It's has nothing to do with the PR title :)

stloyd (changes requested):

composer.json
lchrusciel: my mistake - noise from testing of something else

src/Sylius/Bundle/CoreBundle/Migrations/Version20220711145410.php
lchrusciel: my mistake - noise from testing of something else

src/Sylius/Bundle/CoreBundle/DependencyInjection/SyliusCoreExtension.php
lchrusciel: that's the whole point - our use case is as follows:
we are modularizing Sylius Plus. In order to ensure its interoperability among other libs, we have to test them sometimes with other plugins. This influences configuration in some cases. So we can either have a script that will adjust the application to these changes or use a separate test env. Separate test env seems to be a better way. To be exact:
We have MultiStorePlugin and it has `test` env for testing purposes of it is own. It has additional `test_invoicing` env for testing MultiStorePlugin with InvoicingPlugin (because there are small adjustments between them two). This change will allow me to use Behat services without additional include :)

stloyd: IMHO it should be marked as BC change, even if it's a small one.

lchrusciel: why do you consider it a Backward Compatibility change? How can I break someone's app? The only thing, that will change is to add these services: https://github.com/Sylius/Sylius/blob/1.12/src/Sylius/Bundle/CoreBundle/Resources/config/test_services.xml

Perhaps, I'm too optimistic, but I didn't see the usage of custom envs too often. I'm still expecting to have `test` keyword at the beginning of env.

Also, I don't see any rule broken according to our BC-Policy: https://docs.sylius.com/en/latest/book/organization/backward-compatibility-promise.html

loic425 (approved):

GSadee (approved):

src/Sylius/Bundle/CoreBundle/DependencyInjection/SyliusCoreExtension.php
Zales0123: I agree there is no _hard_ BC here (according to ours and Symfony's BC policy), but it's still worth mentioning this change in the UPGRADE file, to make it evident for everyone upgrading to 1.12. History has already shown us, people do a lot of things we cannot even imagine with their software 😄

Zales0123:
Thank you, Łukasz! 🚀 but please, take a look at my latest comment and respond to it ;)

PR #14160, opened 2022-07-19
[Maintenance] Fixed name of the client when using it in HomepageContext

ernestWarwas:
| Branch?         | 1.12           |
| Deprecations?   | no  |

Fix for client name in two lines. Probably after rebase or upmerge.

### DIFF
--- a/src/Sylius/Behat/Context/Api/Shop/HomepageContext.php
+++ b/src/Sylius/Behat/Context/Api/Shop/HomepageContext.php
@@ -46,7 +46,7 @@ final class HomepageContext implements Context
     {
         Assert::true(
             $this->responseChecker->hasItemWithValue(
-                $this->productsClient->getLastResponse(),
+                $this->client->getLastResponse(),
                 'name',
                 $productName
             )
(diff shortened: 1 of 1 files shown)


NoResponseMate (approved):

Rafikooo (approved):

AdamKasp (approved):

src/Sylius/Behat/Context/Api/Shop/HomepageContext.php
```diff
@@ -22,7 +22,7 @@
 final class HomepageContext implements Context
 {
     public function __construct(
-        private ApiClientInterface $client,
+        private ApiClientInterface $productsClient,
```
GSadee: IMO it should be named as it was:
```suggestion
        private ApiClientInterface $client,
```
because API clients in Behats have been refactored and now, on this branch, we have only two, for shop and admin instead of having client per API resource as it was before

GSadee (changes requested):

GSadee (approved):

lchrusciel:
merged in #14136

PR #14164, opened 2022-07-19
Catch ChannelNotFoundException in profiler

kayue:
| Branch?         | master |
| Deprecations?   | no |

Project with Sylius installed should work even when Channel is not found.

Sylius attached some event listener to `onKernelResponse` and the web profiler, and these listener shouldn't break the application when channel is not found (which is likely match by hostname).

Use case is I have a Sylius application with multiple hostname, and one of the hostname is not registered as Sylius channel because it is not a ecommerce site.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Collector/CartCollector.php
+++ b/src/Sylius/Bundle/CoreBundle/Collector/CartCollector.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sylius\Bundle\CoreBundle\Collector;
 
+use Sylius\Component\Channel\Context\ChannelNotFoundException;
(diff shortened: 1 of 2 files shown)


coldic3:
Hey @kayue, I might be wrong but this PR seems to be a workaround for your problem rather than a real fix. If you have some non-e-commerce sites, Sylius should not impact them at all. If I'm wrong, are you able to provide some tests on that?

Rafikooo:
Hello @kayue,

We have discussed this further and decided that this issue is not valid, since your use case involves using Sylius without such a basic concept as a channel. Therefore, we will close this ticket.
Don't hesitate to open another PR if you think something else in Sylius should be improved 💃 

Best regards,
Rafał

PR #14165, opened 2022-07-19
Prevent ChannelNotFoundException in kernel event and profiler

kayue:
| Branch?         | 1.10 |
| Deprecations?   | no |

Project with Sylius installed should work even when Channel is not found.

Sylius attached some event listener to `onKernelResponse` and the web profiler, and these listener shouldn't break the application when channel is not found (which is likely match by hostname).

Use case is I have a Sylius application with multiple hostname, and one of the hostname is not registered as Sylius channel because it is not a ecommerce site.

This is same as https://github.com/Sylius/Sylius/pull/14164 but based on different branch.


coldic3:
Hi @kayue, I'm closing this one as it's a duplicate of https://github.com/Sylius/Sylius/pull/14164.

PR #14195, opened 2022-07-25
Protect from removing attribute

### LINKED ISSUE #14185: Remove attribute
Hello,

Recently, I noticed the possibility of removing attributes even if a product has that attribute. With a small amount of products, it may not be noticeable, but what with databases that have 10,000 or 100,000+ products, it can kill the database. Maybe it would be worth using the solution that is with taxons, if some object uses this attribute, block the possibility of deletion.

MrPogos:
| Q               | A                                                            
| Branch?         | 1.11
| Bug fix?        | yes                                                   
| New feature?    | no                                                       
| BC breaks?      | no                                                       
| Deprecations?   | no
| Related tickets | fixes #14185 
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AttributeBundle/EventListener/RemoveAttributeListener.php
+++ b/src/Sylius/Bundle/AttributeBundle/EventListener/RemoveAttributeListener.php
@@ -1,0 +1,57 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\AttributeBundle\EventListener;
+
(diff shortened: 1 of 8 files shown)


senghe (changes requested):
All seems good, but you've to write tests for new logic

jakubtobiasz:
Hi 🙋🏻‍♂️!
Thank you for your contribution 🎉! As code quality is important for us I'd like to ask you to:
- Change the base branch this PR is open against. As it resolves the bug we are obligated to open the PR against the latest supported branch (currently it is `1.11`).
- Add Behats and PHPSpecs.
- It would be great if you rename the first commit to make it a little bit for fitting to the standard.

Overall, feedback from me is ok, and the only thing I'm not sure about is a separate service to check if the given attribute can be deleted. Wrapping `find` and simple logical check in a separate class seems to complicate simple things, but I may be wrong 😄.

lchrusciel:
Is it still needed after #14202?

jakubtobiasz:
Closing as #14202 was a successor and fixed the issue.

PR #14200, opened 2022-07-29
[Checkout] Add possibility to choose the required address in the checkout

### LINKED ISSUE #13197: [Checkout] Allow user to choose the default address (billing or delivery)
With 1.7 update came this [feature](https://github.com/Sylius/Sylius/commit/457fc91c155fa9562d3bc7126ab6d2f68ab11d25).
I came to understand, it was added because in certain countries, you would ask for billing address first and if the delivery address is different, it would be asked later.

Unfortunately, in France we are used to ask for the delivery address first and the billing one later if it's different.

Instead of changing this back and forth for cultural reasons, it would be nice if it was available as a setting to the user. I am fairly new to Sylius and I don't have a precise idea on how to do this exactly. I will look into it but I was thinking someone else might also have the same issue.

Greetings,

Sebastien

GSadee:
| Branch?         | 1.12|
| Deprecations?   | no|
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/13197|

Channel form in Admin panel:
image

Checkout:
screencapture-127-0-0-1-8009-en-US-checkout-address-2022-08-01-13_04_37


features/checkout/addressing_order/filling_only_shipping_address_when_it_is_required_for_channel.feature
```diff
@@ -0,0 +1,21 @@
+@checkout
+Feature: Addressing an order with shipping address when it is the required one for the channel
```
Zales0123: Should be aligned with the name of the file (or the filename should be aligned with the feature name :dancer:)

src/Sylius/Behat/Context/Ui/Admin/ManagingChannelsContext.php
```diff
@@ -519,6 +535,20 @@ public function thisChannelMenuTaxonShouldBe(ChannelInterface $channel, string $
...
+     */
+    public function theRequiredAddressInTheCheckoutForThisChannelShouldBe(string $type): void
+    {
+        if ($type === 'shipping') {
+            Assert::true($this->shippingAddressInCheckoutRequiredElement->isShippingAddressInCheckoutRequired());
+
+            return;
+        }
+
+        Assert::false($this->shippingAddressInCheckoutRequiredElement->isShippingAddressInCheckoutRequired());
```
Zales0123: Maybe we should have `ShippingAddressInCheckoutRequiredElementInterface::getShippingAddressInCheckoutRequired(): string` to simplify the assertion here? It looks quite strange :dancer: Similar with setting the address in the `iChooseAddressAsARequiredAddressInTheCheckout` step.

src/Sylius/Behat/Page/Shop/Checkout/AddressPage.php
```diff
@@ -64,6 +64,24 @@ public function chooseDifferentShippingAddress(): void
...
+            $this->getElement('different_billing_address_label')->click();
+
+            return;
+        }
+
+        $billingAddressSwitch = $this->getElement('different_billing_address');
+        Assert::false(
+            $billingAddressSwitch->isChecked(),
+            'Previous state of different billing address switch was true expected to be false',
+        );
```
Zales0123: Do we need to check it? This situation would be an indication of the wrong Behat scenario construction before. Also, if we want to check it, don't we need it for js scenarios as well?

src/Sylius/Behat/Page/Shop/Checkout/AddressPage.php
```diff
@@ -64,6 +64,24 @@ public function chooseDifferentShippingAddress(): void
...
+
+        $billingAddressSwitch->check();
```
Zales0123: This whole content is quite similar to the previous function, could be extracted with some `string $addressType` parameter

src/Sylius/Bundle/ApiBundle/Modifier/OrderAddressModifier.php
```diff
@@ -28,36 +28,69 @@ public function __construct(
...
 
+        $channel = $order->getChannel();
+        Assert::notNull($channel);
+
+        if ($channel->isShippingAddressInCheckoutRequired()) {
+            Assert::notNull($shippingAddress);
+            $billingAddress = $billingAddress ?? clone $shippingAddress;
+        }
+
+        if (!$channel->isShippingAddressInCheckoutRequired()) {
```
Zales0123: I know we don't like that... but it seems as a place for `else` keyword 😄

src/Sylius/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php
```diff
@@ -117,17 +119,32 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+
                 $differentBillingAddress = $orderData['differentBillingAddress'] ?? false;
                 $differentShippingAddress = $orderData['differentShippingAddress'] ?? false;
 
-                if (isset($orderData['billingAddress']) && !$differentBillingAddress && !$differentShippingAddress) {
+                if (
+                    isset($orderData['billingAddress']) &&
+                    !$differentBillingAddress &&
+                    !$differentShippingAddress &&
+                    !$channel->isShippingAddressInCheckoutRequired()
```
Zales0123: **A lot** of conditions :D maybe it would be worth to make a checker private function with them for better readability (similar below)

src/Sylius/Component/Core/Model/Channel.php
```diff
@@ -74,6 +74,8 @@ class Channel extends BaseChannel implements ChannelInterface
     /** @var bool */
     protected $accountVerificationRequired = true;
 
+    protected bool $shippingAddressInCheckoutRequired = false;
```
Zales0123, line 77: Maybe it would be more natural to make it a `string` argument (enum would be great :D but not yet :dancer:)? `string $defaultCheckoutAddress = AddressType::BILLING`? Would be also probably more extendable (if more than 2 addresses are used in the store)

GSadee: I thought about it, but I decided that it would be too exaggerated. The usage of the flag meets the current business requirements, its implementation is still much simpler and I don't think we need this kind of potential customization 🤔

AdamKasp (approved):

src/Sylius/Component/Core/Model/Channel.php
Zales0123: **We** do not need it, but someone using Sylius could need 😄 Moreover, the name of the flag is quite misleading, because, as far as I see, it's not about the requirement of shipping address in checkout or not, but about which of the addresses is used as a _base_ address. Also, making it an enum-like string would not add too much complexity (it's just a change of the property type with some potential simple validation in the `setBaseAddressType(string $type)` method, but would make Behat implementation event simpler (we could change the toggle to the select box with 2 predefined types 🚀). I generally feel that the business requirement is about the **type** of the default checkout address, but we use a **flag** to reflect it in the code.

Still, we can continue this discussion in another PR, we have some time to potentially iterate over it before 1.12.0 release 🖖

GSadee: Yup, I would be for leaving this PR with the current approach and potentially continue the refactor in another PR 👌🏻

src/Sylius/Behat/Page/Shop/Checkout/AddressPage.php
```diff
@@ -393,4 +385,16 @@ private function assertAddressType(string $type): void
 
         Assert::oneOf($type, $availableTypes, sprintf('There are only two available types %s, %s. %s given', self::TYPE_BILLING, self::TYPE_SHIPPING, $type));
     }
+
+    private function chooseDifferentAddress(string $type): void
+    {
+        $driver = $this->getDriver();
+        if ($driver instanceof Selenium2Driver || $driver instanceof ChromeDriver) {
```
Zales0123: https://github.com/Sylius/Sylius/pull/14211

`DriverHelper::isJavascript($this->getDriver())`? :dancer:

Zales0123 (approved):

Zales0123:
Thank you, Grzegorz! :tada:

PR #14201, opened 2022-07-29
Validate non exist coupon code

kayue:
| Branch?         | 1.10 |
| Deprecations?   | no |

If user provide an non exist coupon code validation should fail


lchrusciel (commented):
Hey Kayue! 

thanks for submitting your fix, it makes sense. But we need to at least cover it with the spec (or preferably even with Behat). If you would like us to take it over, please open a support ticket for it.

lchrusciel:
One thing tho - I've just tested it on 1.11 and it seems to work: 
image

jakubtobiasz:
Hi @kayue!
Thank you for your time! I've checked and in the currently supported versions we've fixed it :).

PR #14211, opened 2022-08-01
[Behat] Extract javascript driver check

Zales0123:
| Branch?         | 1.11 |
| Bug fix?        | no |
| New feature?    | no  |
| BC breaks?      | no  |
| Deprecations?   |  |
| Related tickets |   |
| License         | MIT |

It was killing me for ages :dancer:

### DIFF
--- a/src/Sylius/Behat/Service/DriverHelper.php
+++ b/src/Sylius/Behat/Service/DriverHelper.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sylius\Behat\Service;
+
+use Behat\Mink\Driver\DriverInterface;
+use Behat\Mink\Driver\Selenium2Driver;
+use DMore\ChromeDriver\ChromeDriver;
+
+abstract class DriverHelper
+{
+    public static function isJavascript(DriverInterface $driver): bool
+    {
+        return $driver instanceof Selenium2Driver || $driver instanceof ChromeDriver;
+    }
+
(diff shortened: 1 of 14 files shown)


NoResponseMate (approved):

src/Sylius/Behat/Service/DriverHelper.php
```diff
@@ -0,0 +1,22 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Behat\Service;
+
+use Behat\Mink\Driver\DriverInterface;
+use Behat\Mink\Driver\Selenium2Driver;
+use DMore\ChromeDriver\ChromeDriver;
+
+abstract class DriverHelper
```
lchrusciel, line 11: what about
```suggestion
final class DriverHelper
```
with a private constructor, so the class won't be initializable?

lchrusciel (approved):
I'm wondering if an injection of this service won't be a better way to solve it, so the 3rd party contributors would be able to use different drivers within. On the other hand, the external DX stays the same (you are still not able to change these drivers easily), but it boosting our DX and reduce possible maintenance issues with changing drivers everywhere manually, so all in all, big 👍

src/Sylius/Behat/Service/DriverHelper.php
Zales0123: I've used the same construction as in `JQueryHelper` but I agree, we could change them both 👍

loic425:
@Zales0123 great Idea, maybe it's Time to finish panther support I started months ago

PR #12693, opened 2021-06-07
[API] allow extending and overwrite api in yaml + test

SirDomin:
| Q               | A
| --------------- | -----
| Branch?         | master
| Bug fix?        | no
| New feature?    | yes
| BC breaks?      | no
| Deprecations?   | no
| Related tickets |
| License         | MIT

PR #14236, opened 2022-08-16
[API] Remove problematic mapping path

coldic3:
| Branch?         | 1.11                 |
| Deprecations?   | no |
| Related tickets | fixes https://github.com/Sylius/Sylius/issues/14268                     |

This PR removes the `%kernel.project_dir%/config/api_platform` mapping path that was introduced in https://github.com/Sylius/Sylius/pull/12693. In API Platform 2.6 users' mapping configuration is overwritten with bundles' mapping. Because of it, developers cannot override resources in `config/api_platform` as Sylius overrides users' resources placed in `config/api_platform`. It's already fixed in APIP 2.7 but for now, this configuration doesn't work as intended.
We haven't noticed in unit tests that it breaks resource overriding because we have a distinction between `config/api_platform` and `config/api_resource`, where `api_resource` can override resources while `api_platform` can't. This distinction is something reasonable but for testing purposes, we should keep everything in one directory because it makes possible to test this bug.


GSadee (approved):

NoResponseMate (approved):

GSadee:

PR #14262, opened 2022-08-24
Corrected import for money Twig macro.

kkevindev:
| Branch?         | 1.12  |
| Deprecations?   | no |
| Related tickets | |

### DIFF
--- a/docs/book/configuration/currencies.rst
+++ b/docs/book/configuration/currencies.rst
@@ -84,12 +84,12 @@ Displaying Currencies in the templates
 --------------------------------------
 
 There are some useful helpers for rendering money values in the front end.
-Simply import the money macros of the ``AdminBundle`` in your twig template and use the functions to display the value:
+Simply import the money macros of the ``CoreBundle`` in your twig template and use the functions to display the value:
 
 .. code-block:: twig
 
     ..
-    {% import "@SyliusAdmin/Common/Macro/money.html.twig" as money %}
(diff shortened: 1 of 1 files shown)


Roshyo (commented):
Both are correct, there is even one in the CoreBundle. It all depends on what/where you want to display, and what bundles you have enabled.

kkevindev:
@Roshyo 

> Both are correct, there is even one in the CoreBundle. It all depends on what/where you want to display, and what bundles you have enabled.

Than I would rather see the one from the CoreBundle instead of the one in the ShopBundle or AdminBundle. So the documentation is always right, wether you have the ShopBundle or AdminBundle installed or not.

kkevindev:
Do you agree? @Roshyo

CoderMaggie (approved):

kkevindev:
@CoderMaggie @Roshyo PR still open?

*< Surprised Pikachu.png >*

kkevindev:
@jakubtobiasz 👀

jakubtobiasz:
We have this file (`money.html.twig`) in `CoreBundle`, `AdminBundle` and `ShopBundle`. If you look inside each of them you can find there is a different set of macros. The presence of this file in `CoreBundle` is a solve to some use-case. In Admin Panel you should use the one from `AdminBundle`, and in the shop you should use the one from `ShopBundle`.

If you want you can change your PR to include this information, otherwise we will close this PR soon. Thanks for your contribution!

PR #14264, opened 2022-08-25
Fix ProvinceAddressConstraintValidator

liyo:
Currently, provinces validation can fail for provinces sharing the same code, like for example "WI", used by both USA Wisconsin and Austria's Wien.

| Branch?         | 1.12
| Bug fix?        | yes
| New feature?    | no
| BC breaks?      | no
| Deprecations?   | no
| License         | MIT

### DIFF
--- a/src/Sylius/Bundle/AddressingBundle/Validator/Constraints/ProvinceAddressConstraintValidator.php
+++ b/src/Sylius/Bundle/AddressingBundle/Validator/Constraints/ProvinceAddressConstraintValidator.php
@@ -81,7 +81,7 @@ class ProvinceAddressConstraintValidator extends ConstraintValidator
         }
 
         /** @var ProvinceInterface|null $province */
-        $province = $this->provinceRepository->findOneBy(['code' => $address->getProvinceCode()]);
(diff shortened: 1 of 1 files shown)


coldic3:
Hey @liyo,

thanks for the PR! We cannot merge it as it's not tested. Could you write behat scenarios that prove it works and fixes the issue you've described? For any assistance with behat scenarios feel free to ask directly here or on our [Slack](https://sylius.com/slack) 💃

jakubtobiasz:
Hello @liyo 👋🏼
I've checked this today, and I guess this bug is not a valid thing (at least in supported versions of Sylius).

`WI` is not a shared **code** between the USA and Austria. They share the **Abbreviation**, which is 100% possible. In Sylius 1.12 when you try to set `WI` as a province code, you get the following information:
CleanShot 2023-08-16 at 06 54 22@2x

GSadee:
Furthermore, I might add that the code is unique for all provinces, so it is not possible to create two provinces with the same code, even in different countries. I've just tested that the validation will catch it properly. 

I'm closing the issue as it seems not to be valid, but of course, feel free to add a new comment or create a new issue if we have omitted or misunderstood something 😃

PR #14270, opened 2022-08-31
[CatalogPromotions][API] Removal of catalog promotions

Rafikooo:
| Bug fix?        | no                                                 |
| Deprecations?   | no  |

This PR introduced a way to remove catalog promotions. As this operation may have an impact on the prices of the products, it is divided into two parts. First, ( if catalog promotion is eligible ) disabling considered promotion, and second, actually remove it. We can take advantage of Symfony messenger and the prepared configuration of three transports ( failed included ) or perform it without the messenger - synchronously.

image

Configuration of transports:
image

### DIFF
--- a/features/promotion/removing_catalog_promotions/removing_catalog_promotions.feature
+++ b/features/promotion/removing_catalog_promotions/removing_catalog_promotions.feature
@@ -13,37 +13,36 @@ Feature: Removing a catalog promotions
         And there is a catalog promotion "Winter Sale" that reduces price by "30%" and applies on "Clothes" taxon
         And I am logged in as an administrator
 
-    @api @ui @todo
+    @api
     Scenario: Removing an expired catalog promotion
         Given it is "2022-08-22" now
         And the catalog promotion "Winter sale" operates between "2021-12-20" and "2021-12-30"
-        When I remove a "Winter Sale" catalog promotion
-        Then I should be notified that it has been successfully deleted
+        When I request the removal of "Winter Sale" catalog promotion
+        Then I should be notified that the removal operation has started successfully
         And there should be an empty list of catalog promotions
         And "PHP T-Shirt" variant should not be discounted
 
-    @api @ui @todo
+    @api
     Scenario: Removing an active catalog promotion without any time limits
-        When I remove a "Winter Sale" catalog promotion
-        Then I should be notified that it has been successfully deleted
+        When I request the removal of "Winter Sale" catalog promotion
+        Then I should be notified that the removal operation has started successfully
         And there should be an empty list of catalog promotions
         And "PHP T-Shirt" variant should not be discounted
 
-    @api @ui @todo
+    @api
     Scenario: Removing an active catalog promotion in the time range
         Given it is "2022-12-15" now
         And the catalog promotion "Winter sale" operates between "2022-12-01" and "2022-12-30"
-        When I remove a "Winter Sale" catalog promotion
-        Then I should be notified that it has been successfully deleted
+        When I request the removal of "Winter Sale" catalog promotion
+        Then I should be notified that the removal operation has started successfully
         And there should be an empty list of catalog promotions
         And "PHP T-Shirt" variant should not be discounted
 
-    @api @ui @todo
+    @api
     Scenario: Removing a scheduled catalog promotion
         Given it is "2022-08-22" now
         And the catalog promotion "Winter sale" operates between "2022-12-01" and "2023-02-28"
-        When I remove a "Winter Sale" catalog promotion
-        Then I should be notified that it has been successfully deleted
+        When I request the removal of "Winter Sale" catalog promotion
+        Then I should be notified that the removal operation has started successfully
         And there should be an empty list of catalog promotions
         And "PHP T-Shirt" variant should not be discounted
-
--- a/src/Sylius/Behat/Client/ResponseChecker.php
+++ b/src/Sylius/Behat/Client/ResponseChecker.php
@@ -62,6 +62,11 @@ final class ResponseChecker implements ResponseCheckerInterface
         return $this->getResponseContentValue($response, 'hydra:description');
     }
 
+    public function isAccepted(Response $response): bool
+    {
+        return $response->getStatusCode() === Response::HTTP_ACCEPTED;
+    }
+
     public function isCreationSuccessful(Response $response): bool
     {
         return $response->getStatusCode() === Response::HTTP_CREATED;
--- a/src/Sylius/Bundle/ApiBundle/Controller/RemoveCatalogPromotionAction.php
+++ b/src/Sylius/Bundle/ApiBundle/Controller/RemoveCatalogPromotionAction.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\ApiBundle\Controller;
+
+use Sylius\Bundle\CoreBundle\CatalogPromotion\Processor\CatalogPromotionRemovalProcessorInterface;
+use Sylius\Component\Promotion\Exception\CatalogPromotionNotFoundException;
+use Sylius\Component\Promotion\Exception\InvalidCatalogPromotionStateException;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+/** @experimental */
+final class RemoveCatalogPromotionAction
+{
+    public function __construct(
+        private CatalogPromotionRemovalProcessorInterface $catalogPromotionRemovalProcessor
+    ) {
+    }
+
+    public function __invoke(Request $request): Response
+    {
+        try {
+            $this->catalogPromotionRemovalProcessor->removeCatalogPromotion($request->attributes->get('code'));
+
+            return new Response(status: Response::HTTP_ACCEPTED);
+        } catch (CatalogPromotionNotFoundException) {
+            return new JsonResponse(status: Response::HTTP_NOT_FOUND);
+        } catch (InvalidCatalogPromotionStateException $exception) {
+            return new JsonResponse(
+                ['code' => Response::HTTP_BAD_REQUEST, 'message' => $exception->getMessage()],
+                Response::HTTP_BAD_REQUEST,
+            );
+        }
+    }
+}
--- a/src/Sylius/Bundle/CoreBundle/CatalogPromotion/Command/RemoveInactiveCatalogPromotion.php
+++ b/src/Sylius/Bundle/CoreBundle/CatalogPromotion/Command/RemoveInactiveCatalogPromotion.php
@@ -1,0 +1,21 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\CatalogPromotion\Command;
+
+final class RemoveInactiveCatalogPromotion
+{
+    public function __construct(public string $code)
+    {
+    }
+}
--- a/src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/RemoveInactiveCatalogPromotionHandler.php
+++ b/src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/RemoveInactiveCatalogPromotionHandler.php
@@ -1,0 +1,51 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\CatalogPromotion\CommandHandler;
+
+use Doctrine\ORM\EntityManager;
+use Sylius\Bundle\CoreBundle\CatalogPromotion\Command\RemoveInactiveCatalogPromotion;
+use Sylius\Component\Core\Model\CatalogPromotionInterface;
+use Sylius\Component\Promotion\Exception\InvalidCatalogPromotionStateException;
+use Sylius\Component\Promotion\Model\CatalogPromotionStates;
+use Sylius\Component\Promotion\Repository\CatalogPromotionRepositoryInterface;
+
+final class RemoveInactiveCatalogPromotionHandler
+{
+    public function __construct(
+        private CatalogPromotionRepositoryInterface $catalogPromotionRepository,
+        private EntityManager $entityManager,
+    ) {
+    }
+
+    public function __invoke(RemoveInactiveCatalogPromotion $command): void
+    {
+        /** @var CatalogPromotionInterface|null $catalogPromotion */
+        $catalogPromotion = $this->catalogPromotionRepository->findOneBy(['code' => $command->code]);
+
+        if (null === $catalogPromotion) {
+            return;
+        }
+
+        if ($catalogPromotion->getState() !== CatalogPromotionStates::STATE_INACTIVE) {
+            throw new InvalidCatalogPromotionStateException(
+                sprintf(
+                    'Catalog promotion with code "%s" cannot be removed as it is not in an inactive state.',
+                    $catalogPromotion->getCode()
+                )
+            );
+        }
+
+        $this->entityManager->remove($catalogPromotion);
+    }
+}
--- a/src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php
+++ b/src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php
@@ -1,0 +1,91 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\CatalogPromotion\Processor;
+
+use Sylius\Bundle\CoreBundle\CatalogPromotion\Command\RemoveInactiveCatalogPromotion;
+use Sylius\Component\Core\Model\CatalogPromotionInterface;
+use Sylius\Component\Promotion\Event\CatalogPromotionEnded;
+use Sylius\Component\Promotion\Exception\CatalogPromotionNotFoundException;
+use Sylius\Component\Promotion\Exception\InvalidCatalogPromotionStateException;
+use Sylius\Component\Promotion\Model\CatalogPromotionStates;
+use Sylius\Component\Promotion\Repository\CatalogPromotionRepositoryInterface;
+use Symfony\Component\Messenger\MessageBusInterface;
+
+final class CatalogPromotionRemovalProcessor implements CatalogPromotionRemovalProcessorInterface
+{
+    public function __construct(
+        private CatalogPromotionRepositoryInterface $catalogPromotionRepository,
+        private MessageBusInterface $commandBus,
+        private MessageBusInterface $eventBus,
+    ) {
+    }
+
+    public function removeCatalogPromotion(string $catalogPromotionCode): void
+    {
+        /** @var CatalogPromotionInterface|null $catalogPromotion */
+        $catalogPromotion = $this->getCatalogPromotion($catalogPromotionCode);
+
+        if ($catalogPromotion->getState() === CatalogPromotionStates::STATE_INACTIVE) {
+            $this->announceInactiveCatalogPromotionRemoval($catalogPromotionCode);
+
+            return;
+        }
+
+        if ($catalogPromotion->getState() === CatalogPromotionStates::STATE_ACTIVE) {
+            $this->disableCatalogPromotion($catalogPromotion);
+            $this->announceCatalogPromotionEnd($catalogPromotionCode);
+            $this->announceInactiveCatalogPromotionRemoval($catalogPromotionCode);
+
+            return;
+        }
+
+        if ($catalogPromotion->getState() === CatalogPromotionStates::STATE_PROCESSING) {
+            throw new InvalidCatalogPromotionStateException(
+                sprintf(
+                    'Catalog promotion with code "%s" cannot be removed as it is now being processed.',
+                    $catalogPromotionCode
+                )
+            );
+        }
+
+        throw new \DomainException('Invalid catalog promotion state.');
+    }
+
+    private function announceCatalogPromotionEnd(string $catalogPromotionCode): void
+    {
+        $this->eventBus->dispatch(new CatalogPromotionEnded($catalogPromotionCode));
+    }
+
+    private function announceInactiveCatalogPromotionRemoval(string $catalogPromotionCode): void
+    {
+        $this->commandBus->dispatch(new RemoveInactiveCatalogPromotion($catalogPromotionCode));
+    }
+
+    private function disableCatalogPromotion(CatalogPromotionInterface $catalogPromotion): void
+    {
+        $catalogPromotion->setEnabled(false);
+    }
+
+    private function getCatalogPromotion(string $catalogPromotionCode): CatalogPromotionInterface
+    {
+        /** @var CatalogPromotionInterface|null $catalogPromotion */
+        $catalogPromotion = $this->catalogPromotionRepository->findOneBy(['code' => $catalogPromotionCode]);
+
+        if (null === $catalogPromotion) {
+            throw new CatalogPromotionNotFoundException('Catalog promotion with given code does not exist.');
+        }
+
+        return $catalogPromotion;
+    }
+}
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/messenger.yaml
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/messenger.yaml
@@ -1,18 +1,30 @@
 framework:
     messenger:
-        # Uncomment this (and the failed transport below) to send failed messages to this transport for later handling.
-
         transports:
-        # https://symfony.com/doc/current/messenger.html#transport-configuration
-            main: '%env(MESSENGER_TRANSPORT_DSN)%'
-
+            main: 
+                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
+                failure_transport: 'main_failed'
+            main_failed: 
+                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
+                options:
+                    queue_name: 'main_failed'
+            catalog_promotion_removal: 
+                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
+                failure_transport: 'catalog_promotion_removal_failed'
+                options:
+                    queue_name: 'catalog_promotion_removal'
+            catalog_promotion_removal_failed:
+                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
+                options:
+                    queue_name: 'catalog_promotion_removal_failed'
+        reset_on_message: true
         routing:
             'Sylius\Bundle\CoreBundle\CatalogPromotion\Command\UpdateCatalogPromotionState': main
             'Sylius\Bundle\CoreBundle\CatalogPromotion\Command\ApplyCatalogPromotionsOnVariants': main
+            'Sylius\Bundle\CoreBundle\CatalogPromotion\Command\RemoveInactiveCatalogPromotion': catalog_promotion_removal
             'Sylius\Component\Promotion\Event\CatalogPromotionCreated': main
             'Sylius\Component\Promotion\Event\CatalogPromotionEnded': main
             'Sylius\Component\Promotion\Event\CatalogPromotionUpdated': main
-
         default_bus: sylius.command_bus
         buses:
             sylius.command_bus: &command_bus
--- a/src/Sylius/Bundle/CoreBundle/spec/CatalogPromotion/CommandHandler/RemoveInactiveCatalogPromotionHandlerSpec.php
+++ b/src/Sylius/Bundle/CoreBundle/spec/CatalogPromotion/CommandHandler/RemoveInactiveCatalogPromotionHandlerSpec.php
@@ -1,0 +1,75 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace spec\Sylius\Bundle\CoreBundle\CatalogPromotion\CommandHandler;
+
+use Doctrine\ORM\EntityManager;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Sylius\Bundle\CoreBundle\CatalogPromotion\Command\RemoveInactiveCatalogPromotion;
+use Sylius\Component\Core\Model\CatalogPromotionInterface;
+use Sylius\Component\Promotion\Exception\InvalidCatalogPromotionStateException;
+use Sylius\Component\Promotion\Repository\CatalogPromotionRepositoryInterface;
+
+final class RemoveInactiveCatalogPromotionHandlerSpec extends ObjectBehavior
+{
+    public function let(
+        CatalogPromotionRepositoryInterface $catalogPromotionRepository,
+        EntityManager $entityManager,
+    ): void {
+        $this->beConstructedWith($catalogPromotionRepository, $entityManager);
+    }
+
+    public function it_removes_an_inactive_catalog_promotion(
+        CatalogPromotionRepositoryInterface $catalogPromotionRepository,
+        EntityManager $entityManager,
+        CatalogPromotionInterface $catalogPromotion,
+    ): void {
+        $catalogPromotionRepository->findOneBy(['code' => 'CATALOG_PROMOTION_CODE'])->willReturn($catalogPromotion);
+
+        $catalogPromotion->getState()->willReturn('inactive');
+
+        $this(new RemoveInactiveCatalogPromotion('CATALOG_PROMOTION_CODE'));
+
+        $entityManager->remove($catalogPromotion)->shouldBeCalled();
+    }
+
+    public function it_throws_an_exception_if_catalog_promotion_is_not_in_an_inactive_state(
+        CatalogPromotionRepositoryInterface $catalogPromotionRepository,
+        EntityManager $entityManager,
+        CatalogPromotionInterface $catalogPromotion,
+    ): void {
+        $catalogPromotionRepository->findOneBy(['code' => 'CATALOG_PROMOTION_CODE'])->willReturn($catalogPromotion);
+
+        $catalogPromotion->getState()->willReturn('active');
+        $catalogPromotion->getCode()->willReturn('CATALOG_PROMOTION_CODE');
+
+        $entityManager->remove(Argument::any())->shouldNotBeCalled();
+
+        $this
+            ->shouldThrow(InvalidCatalogPromotionStateException::class)
+            ->during('__invoke', [new RemoveInactiveCatalogPromotion('CATALOG_PROMOTION_CODE')])
+        ;
+    }
+
+    public function it_returns_if_there_is_no_catalog_promotion_with_given_code(
+        CatalogPromotionRepositoryInterface $catalogPromotionRepository,
+        EntityManager $entityManager,
+    ): void {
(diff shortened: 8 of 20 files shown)


src/Sylius/Bundle/ApiBundle/Controller/TerminateCatalogPromotionAction.php
```diff
@@ -0,0 +1,46 @@
...
+        /** @var CatalogPromotionInterface|null $catalogPromotion */
+        $catalogPromotion = $this->catalogPromotionRepository->findOneBy(['code' => $code]);
+
+        if (null === $catalogPromotion) {
+            return new Response(status: Response::HTTP_NOT_FOUND);
+        }
+
+        $this->commandBus->dispatch(new TerminateCatalogPromotion($code));
+
+        return new Response(status: Response::HTTP_NO_CONTENT);
```
Rafikooo, line 44: I left a 204 status code due to the behat response checker which's looking for 204 indeed. I wonder if we should return 204 or 202 depending on messenger sync / async config, wdyt?

.github/workflows/application.yml
```diff
@@ -500,19 +500,6 @@ jobs:
                 mysql: ["5.7", "8.0"]
                 twig: ["3.x"]
                 dbal: ["3"]
-                include:
```
Rafikooo, line 503: I temporary "disabled" the following suites to unlock first iteration of CR
https://github.com/Sylius/Sylius/runs/8183036430?check_suite_focus=true#step:24:217

This build below was successful when I ran behat without `-f progress` option
https://github.com/Sylius/Sylius/actions/runs/2978035295

composer.json
```diff
@@ -229,7 +229,7 @@
         "symfony/browser-kit": "^5.4",
         "symfony/debug-bundle": "^5.4",
         "symfony/dotenv": "^5.4",
-        "symfony/flex": "^1.7",
+        "symfony/flex": "^2.2",
```
NoResponseMate: Is this change necessary?

features/promotion/removing_catalog_promotions/removing_catalog_promotions.feature
```diff
@@ -13,37 +13,36 @@ Feature: Removing a catalog promotions
...
         And I am logged in as an administrator
 
-    @api @ui @todo
+    @api
     Scenario: Removing an expired catalog promotion
         Given it is "2022-08-22" now
         And the catalog promotion "Winter sale" operates between "2021-12-20" and "2021-12-30"
-        When I remove a "Winter Sale" catalog promotion
-        Then I should be notified that it has been successfully deleted
+        When I order the removal of "Winter Sale" catalog promotion
```
NoResponseMate: IMO `request` instead of `order` fits better

src/Sylius/Bundle/CoreBundle/Resources/config/app/messenger.yaml
```diff
@@ -1,14 +1,25 @@
...
+            catalog_promotion_removal: 
+                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
+                options:
+                    queue_name: 'catalog_promotion_removal'
+            failed: 
+                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
+                options:
+                    queue_name: 'failed'
+                    
+        failure_transport: failed
```
NoResponseMate: Just thinking aloud, but wouldn't it be better to have a more localized failure transport since we're adding a new one for removing?
Instead of a global `failed` have something like:
```yaml
catalog_promotion_removal: 
    dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
    failure_transport: 'catalog_promotion_removal_failed'
    options:
        queue_name: 'catalog_promotion_removal'
catalog_promotion_removal_failed: 
    dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
    options:
        queue_name: 'catalog_promotion_removal_failed'
```
with no `failure_transport` node at the top level.
WDYT?

src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/TerminateCatalogPromotionHandler.php
```diff
@@ -0,0 +1,49 @@
...
+    {
+        /** @var CatalogPromotion|null $catalogPromotion */
+        $catalogPromotion = $this->catalogPromotionRepository->findOneBy(['code' => $command->code]);
+
+        if (null === $catalogPromotion) {
+            return;
+        }
+
+        $catalogPromotion->setEnabled(false);
+        $catalogPromotion->setState(CatalogPromotionStates::STATE_PROCESSING);
```
NoResponseMate: Shouldn't this go through the state machine?

src/Sylius/Bundle/ApiBundle/Controller/TerminateCatalogPromotionAction.php
```diff
@@ -0,0 +1,46 @@
...
+namespace Sylius\Bundle\ApiBundle\Controller;
+
+use Sylius\Bundle\CoreBundle\CatalogPromotion\Command\TerminateCatalogPromotion;
+use Sylius\Component\Core\Model\CatalogPromotionInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Messenger\MessageBusInterface;
+
+final class TerminateCatalogPromotionAction
```
GSadee: Add an `@experimental` tag as this service is a part of ApiBundle

src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/TerminateCatalogPromotionHandler.php
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    public function __invoke(TerminateCatalogPromotion $command): void
+    {
+        /** @var CatalogPromotion|null $catalogPromotion */
+        $catalogPromotion = $this->catalogPromotionRepository->findOneBy(['code' => $command->code]);
+
+        if (null === $catalogPromotion) {
+            return;
+        }
```
GSadee: IMO we should throw an exception here, the same in the `RemoveCatalogPromotionHandler`

src/Sylius/Bundle/CoreBundle/spec/CatalogPromotion/CommandHandler/TerminateCatalogPromotionHandlerSpec.php
```diff
@@ -0,0 +1,73 @@
...
+        $event = new CatalogPromotionUpdated('CATALOG_PROMOTION_CODE');
+        $command = new RemoveCatalogPromotion('CATALOG_PROMOTION_CODE');
+
+        $eventBus->dispatch($event, [new DispatchAfterCurrentBusStamp()])->willReturn(new Envelope($event));
+        $commandBus->dispatch($command, [new DispatchAfterCurrentBusStamp()])->willReturn(new Envelope($command));
+
+        $this(new TerminateCatalogPromotion('CATALOG_PROMOTION_CODE'));
+    }
+
+    public function it_does_not_dispatch_commands_if_catalog_promotion_from_command_does_not_exists(
```
GSadee:
```suggestion
    public function it_does_not_dispatch_commands_if_catalog_promotion_from_command_does_not_exist(
```

src/Sylius/Bundle/CoreBundle/spec/CatalogPromotion/CommandHandler/RemoveCatalogPromotionHandlerSpec.php
```diff
@@ -0,0 +1,52 @@
...
+        CatalogPromotionRepositoryInterface $catalogPromotionRepository,
+        EntityManager $entityManager,
+    ): void {
+        $catalogPromotionRepository->findOneBy(['code' => 'CATALOG_PROMOTION_CODE'])->willReturn($catalogPromotion);
+        $entityManager->remove($catalogPromotion)->shouldBeCalled();
+
+        $this(new RemoveCatalogPromotion('CATALOG_PROMOTION_CODE'));
+    }
+
+    public function it_does_not_remove_catalog_promotion_if_catalog_promotion_from_command_does_not_exists(
```
GSadee:
```suggestion
    public function it_does_not_remove_catalog_promotion_if_catalog_promotion_from_command_does_not_exist(
```

src/Sylius/Bundle/CoreBundle/Resources/config/app/messenger.yaml
Rafikooo: Good point! I missed that we can set up a dedicated failure transport

src/Sylius/Bundle/CoreBundle/Resources/config/app/state_machine/sylius_catalog_promotion.yaml
```diff
@@ -5,12 +5,16 @@ winzou_state_machine:
...
+            inactive: ~
+            processing: ~
         transitions:
             activate:
                 from: [inactive]
                 to: active
             deactivate:
                 from: [active]
                 to: inactive
+            process:
```
GSadee: There should be the ability to activate and deactivate from the `process` state

src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/RemoveCatalogPromotionHandler.php
```diff
@@ -0,0 +1,41 @@
...
+    public function __construct(
+        private CatalogPromotionRepositoryInterface $catalogPromotionRepository,
+        private EntityManager $entityManager,
+    ) {
+    }
+
+    public function __invoke(RemoveCatalogPromotion $command): void
+    {
+        /** @var CatalogPromotion|null $catalogPromotion */
+        $catalogPromotion = $this->catalogPromotionRepository->findOneBy(['code' => $command->code]);
```
GSadee: Should we also check the state of the catalog promotion here, so as not to remove the one that is active?

src/Sylius/Behat/Client/ResponseChecker.php
```diff
@@ -62,6 +62,11 @@ public function getError(Response $response): string
         return $this->getResponseContentValue($response, 'hydra:description');
     }
 
+    public function isAcceptSuccessful(Response $response): bool
```
Zales0123: To be 100% grammarly proper, it probably should be `isAcceptanceSuccessful`

src/Sylius/Bundle/ApiBundle/Controller/TerminateCatalogPromotionAction.php
```diff
@@ -0,0 +1,47 @@
...
+
+use Sylius\Bundle\CoreBundle\CatalogPromotion\Command\TerminateCatalogPromotion;
+use Sylius\Component\Core\Model\CatalogPromotionInterface;
+use Sylius\Component\Resource\Repository\RepositoryInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Messenger\MessageBusInterface;
+
+/** @experimental */
+final class TerminateCatalogPromotionAction
```
Zales0123: We use "removal" in scenarios, but "terminate" here. Cannot it be `RemoveCatalogPromotionAction`?

src/Sylius/Bundle/ApiBundle/Controller/TerminateCatalogPromotionAction.php
```diff
@@ -0,0 +1,47 @@
...
+        $code = $request->attributes->get('code');
+
+        /** @var CatalogPromotionInterface|null $catalogPromotion */
+        $catalogPromotion = $this->catalogPromotionRepository->findOneBy(['code' => $code]);
+
+        if (null === $catalogPromotion) {
+            return new Response(status: Response::HTTP_NOT_FOUND);
+        }
+
+        $this->commandBus->dispatch(new TerminateCatalogPromotion($code));
```
Zales0123: Same here

src/Sylius/Bundle/ApiBundle/Controller/TerminateCatalogPromotionAction.php
```diff
@@ -0,0 +1,47 @@
...
+    public function __invoke(Request $request): Response
+    {
+        $code = $request->attributes->get('code');
+
+        /** @var CatalogPromotionInterface|null $catalogPromotion */
+        $catalogPromotion = $this->catalogPromotionRepository->findOneBy(['code' => $code]);
+
+        if (null === $catalogPromotion) {
+            return new Response(status: Response::HTTP_NOT_FOUND);
+        }
```
Zales0123: [Here](https://github.com/Sylius/Sylius/pull/14270/files#diff-6b6691add61aa0c437440437a343c2469edb4cb206231ac173bbf47149d19606R40-R44) we have the same check. Maybe we should leave the one in the command handler and catch the exception here? 🤔 It just seems as the unneeded duplication

src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/TerminateCatalogPromotionHandler.php
```diff
@@ -0,0 +1,54 @@
...
+            throw new InvalidArgumentException('Catalog promotion with given code does not exist.');
+        }
+
+        $catalogPromotion->setEnabled(false);
+
+        $stateMachine = $this->stateMachineFactory->get($catalogPromotion, CatalogPromotionTransitions::GRAPH);
+        $stateMachine->apply(CatalogPromotionTransitions::TRANSITION_PROCESS);
+
+        $this->eventBus->dispatch(new CatalogPromotionUpdated($command->code), [new DispatchAfterCurrentBusStamp()]);
+        $this->commandBus->dispatch(new RemoveCatalogPromotion($command->code), [new DispatchAfterCurrentBusStamp()]);
```
Zales0123: I see now 😄 It's quite shady for me, that we introduce a new term `termination` to reflect, that the **removal** of catalog promotions consists of two steps (disabling and an actual removal). Maybe just use the word `terminate` in Behat scenarios to make it consistent? Or do everything in one handler, reflecting **one** intention of an administrator - in the end, the administrator **wants to remove** the catalog promotion, the two-step process is just an implementation detail. 

Moreover, maybe we should make some saga/process manager here? It feels inconsistent, that we have 2 steps of the process, but the first command consists of two of them (disablement logic and dispatching the command of catalog promotion removal) - which feels like we have a command (intention) with 2 steps inside and another with 1 step instead of two commands with 1 step each :dancer: But it can be a subject of refactor later 🖖

src/Sylius/Bundle/CoreBundle/Resources/config/app/messenger.yaml
```diff
@@ -1,14 +1,30 @@
...
+        reset_on_message: true
 
         routing:
             'Sylius\Bundle\CoreBundle\CatalogPromotion\Command\UpdateCatalogPromotionState': main
             'Sylius\Bundle\CoreBundle\CatalogPromotion\Command\ApplyCatalogPromotionsOnVariants': main
+            'Sylius\Bundle\CoreBundle\CatalogPromotion\Command\TerminateCatalogPromotion': main
+            'Sylius\Bundle\CoreBundle\CatalogPromotion\Command\RemoveCatalogPromotion': catalog_promotion_removal
             'Sylius\Component\Promotion\Event\CatalogPromotionCreated': main
             'Sylius\Component\Promotion\Event\CatalogPromotionEnded': main
             'Sylius\Component\Promotion\Event\CatalogPromotionUpdated': main
```
coldic3: I know we already have transport called `main` but there is a risk someone can have the same transport name in their app. IMO we should prefix transport names with `sylius_` – similar to what we have in DB table names.

```suggestion
            main: 
                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
            sylius_catalog_promotion_removal: 
                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
                failure_transport: 'sylius_catalog_promotion_removal_failed'
                options:
                    queue_name: 'sylius_catalog_promotion_removal'
            sylius_catalog_promotion_removal_failed:
                dsn: '%env(MESSENGER_TRANSPORT_DSN)%'
                options:
                    queue_name: 'sylius_catalog_promotion_removal_failed'
                    
        reset_on_message: true

        routing:
            'Sylius\Bundle\CoreBundle\CatalogPromotion\Command\UpdateCatalogPromotionState': main
            'Sylius\Bundle\CoreBundle\CatalogPromotion\Command\ApplyCatalogPromotionsOnVariants': main
            'Sylius\Bundle\CoreBundle\CatalogPromotion\Command\TerminateCatalogPromotion': main
            'Sylius\Bundle\CoreBundle\CatalogPromotion\Command\RemoveCatalogPromotion': sylius_catalog_promotion_removal
            'Sylius\Component\Promotion\Event\CatalogPromotionCreated': main
            'Sylius\Component\Promotion\Event\CatalogPromotionEnded': main
            'Sylius\Component\Promotion\Event\CatalogPromotionUpdated': main
```

I would also be for deprecating `main` transport and using `sylius_main` instead in a separate PR. WDYT?
And is `failed` transport needed as we have `sylius_catalog_promotion_removal_failed`?

src/Sylius/Behat/Client/ResponseChecker.php
NoResponseMate: IMO `isAccepted` would fit best, both `isAcceptSuccessful` and `isAcceptanceSuccessful` don't really make much sense, as we're not requesting the acceptance, its just the result

src/Sylius/Bundle/CoreBundle/Resources/config/app/messenger.yaml
Rafikooo: It makes sense to me, any other opinions? :) @GSadee @Zales0123 @NoResponseMate?

Rafikooo: On the other hand, if processed messages have the same priority for both Sylius and end app custom tasks, is there a  need to introduce another transport?

Rafikooo: @coldic3 , "And is failed transport needed as we have sylius_catalog_promotion_removal_failed"
Failed transport was introduced as a fallback for the `main` transport, Agree that this is not the case of removing catalog promotions but does give the opportunity of manually resolving failed ( by default 3x at row ) `main` transport messages by the administrator instead of abandoning the that unsuccessful jobs

coldic3: > On the other hand, if processed messages have the same priority for both Sylius and end app custom tasks, is there a need to introduce another transport?

IMO In theory it's not needed to introduce another transport in such a case but we don't know the transport name in the end app. It might be `main` but it doesn't have to be.


> @coldic3 , "And is failed transport needed as we have sylius_catalog_promotion_removal_failed" Failed transport was introduced as a fallback for the `main` transport, Agree that this is not the case of removing catalog promotions but does give the opportunity of manually resolving failed ( by default 3x at row ) `main` transport messages by the administrator instead of abandoning the that unsuccessful jobs

I'm in favor of having `failed` transport for `main` but changing `failure_transport` globally (for all transports) we interfere not only in Sylius behavior itself but also in the end app behavior. Therefore maybe specifying this failed transport only for `main` transport will be better? 🤔

src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/TerminateCatalogPromotionHandler.php
Rafikooo: Refactoring done 🚀

src/Sylius/Bundle/CoreBundle/spec/CatalogPromotion/CommandHandler/RemoveInactiveCatalogPromotionHandlerSpec.php
```diff
@@ -0,0 +1,89 @@
...
+        StateMachine $stateMachine,
+    ): void {
+        $catalogPromotionRepository->findOneBy(['code' => 'CATALOG_PROMOTION_CODE'])->willReturn($catalogPromotion);
+
+        $stateMachineFactory->get($catalogPromotion, 'sylius_catalog_promotion')->willReturn($stateMachine);
+        $stateMachine->getState()->willReturn('inactive');
+        $this(new RemoveInactiveCatalogPromotion('CATALOG_PROMOTION_CODE'));
+
+        $entityManager->remove($catalogPromotion)->shouldBeCalled();
+
```
AdamKasp:
```suggestion
```

src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/RemoveInactiveCatalogPromotionHandler.php
```diff
@@ -0,0 +1,58 @@
...
+    public function __invoke(RemoveInactiveCatalogPromotion $command): void
+    {
+        /** @var CatalogPromotionInterface|null $catalogPromotion */
+        $catalogPromotion = $this->catalogPromotionRepository->findOneBy(['code' => $command->code]);
+
+        if (null === $catalogPromotion) {
+            return;
+        }
+
+        /** @var StateMachine $stateMachine */
```
GSadee: Is it needed? The factory can return only an object implementing `StateMachineInterface` if I see correctly
```suggestion
```
or if it is necessary:
```suggestion
        /** @var StateMachineInterface $stateMachine */
```

src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/RemoveInactiveCatalogPromotionHandler.php
```diff
@@ -0,0 +1,58 @@
...
+        $stateMachine = $this->stateMachineFactory->get($catalogPromotion, CatalogPromotionTransitions::GRAPH);
+
+        if ($stateMachine->getState() !== CatalogPromotionStates::STATE_INACTIVE) {
```
GSadee: Why don't you check the state on the catalog promotion?

src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/RemoveInactiveCatalogPromotionHandler.php
```diff
@@ -0,0 +1,58 @@
...
+        private CatalogPromotionRepositoryInterface $catalogPromotionRepository,
+        private EntityManager $entityManager,
+        private FactoryInterface $stateMachineFactory,
+    ) {
+    }
+
+    public function __invoke(RemoveInactiveCatalogPromotion $command): void
+    {
+        /** @var CatalogPromotionInterface|null $catalogPromotion */
+        $catalogPromotion = $this->catalogPromotionRepository->findOneBy(['code' => $command->code]);
```
GSadee, line 34: Maybe you can add state to the criteria not to check that later?
```suggestion
        $catalogPromotion = $this->catalogPromotionRepository->findOneBy(['code' => $command->code, 'state' => CatalogPromotionStates::STATE_INACTIVE]);
```

src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php
```diff
@@ -0,0 +1,104 @@
...
+use Sylius\Component\Core\Model\CatalogPromotionInterface;
+use Sylius\Component\Promotion\Event\CatalogPromotionEnded;
+use Sylius\Component\Promotion\Model\CatalogPromotionStates;
+use Sylius\Component\Promotion\Model\CatalogPromotionTransitions;
+use Sylius\Component\Promotion\Repository\CatalogPromotionRepositoryInterface;
+use Symfony\Component\Messenger\MessageBusInterface;
+
+final class CatalogPromotionRemovalProcessor implements CatalogPromotionRemovalProcessorInterface
+{
+    private StateMachineInterface $stateMachine;
```
GSadee: I do not see a need to set a variable with state machine here, you can get state machine in the method and use it there

src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    public function removeCatalogPromotion(string $catalogPromotionCode): void
+    {
+        /** @var CatalogPromotionInterface|null $catalogPromotion */
+        $catalogPromotion = $this->getCatalogPromotion($catalogPromotionCode);
+
+        $this->initializeStateMachine($catalogPromotion);
+
+        if ($this->stateMachine->getState() === CatalogPromotionStates::STATE_INACTIVE) {
```
GSadee: The state set on catalog promotion is not always up-to-date that you are using the state machine?

src/Sylius/Bundle/CoreBundle/CatalogPromotion/Command/RemoveInactiveCatalogPromotion.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\CatalogPromotion\Command;
+
+final class RemoveInactiveCatalogPromotion
```
AdamKasp, line 16: I don't have strong opinion but `RemoveCatalogPromotion` should be enough, information about `inactive` is an implementation detail, and IMO can be ommit.

src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/RemoveInactiveCatalogPromotionHandler.php
Rafikooo: Thought that we should rely on the state machine

src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/RemoveInactiveCatalogPromotionHandler.php
Rafikooo: First I did it this way, but there is a possibility ( in `async` processing with multiple workers ) that
 ```
$this->catalogPromotionRepository->findOneBy(['code' => $command->code, 'state' => CatalogPromotionStates::STATE_INACTIVE]);
```
returns `null` but in fact, this promotion is still being processed and we end up with an unremoved promotion.
Throwing an exception gives us another three ( by default ) attempts to process that message when other processes have finished price recalculating

File: src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php (hunk 23 above)
Comment (Rafikooo) [reply src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php:30]: Same as here
https://github.com/Sylius/Sylius/pull/14270#discussion_r970469589
Thought that we should rely on the state machine :)

File: src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php (hunk 24 above)
Comment (Rafikooo) [reply src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php]: Same as here
https://github.com/Sylius/Sylius/pull/14270#discussion_r970469589
Thought that we should rely on the state machine :)

File: src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php [hunk 26]
```diff
@@ -0,0 +1,91 @@
...
+    {
+        $catalogPromotion->setEnabled(false);
+    }
+
+    private function dispatchCatalogPromotionEndedEvent(string $catalogPromotionCode): void
+    {
+        $this->eventBus->dispatch(new CatalogPromotionEnded($catalogPromotionCode));
+    }
+
+    private function dispatchRemoveInactiveCatalogPromotionCommand(string $catalogPromotionCode): void
```
Zales0123: I would be from removing `dispatch` prefix and `Command` suffix (so `removeInactiveCatalogPromotionCommand` instead of `dispatchRemoveInactiveCatalogPromotion`). Again, these are implementation details, names of the function should _generally_ reflects intentions. Similarly, you do not name the function from the line 65 `setCatalogPromotionEnabledFlag` ;)

src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php
```diff
@@ -0,0 +1,91 @@
...
+        $this->commandBus->dispatch(new RemoveInactiveCatalogPromotion($catalogPromotionCode));
+    }
+
+    private function getCatalogPromotion(string $catalogPromotionCode): CatalogPromotionInterface
+    {
+        /** @var CatalogPromotionInterface|null $catalogPromotion */
+        $catalogPromotion = $this->catalogPromotionRepository->findOneBy(['code' => $catalogPromotionCode]);
+
+        if (null === $catalogPromotion) {
+            throw new InvalidArgumentException('Catalog promotion with given code does not exist.');
```
Zales0123: Maybe something more specific than `InvalidArgumentException`? `CatalogPromotionNotFoundException`? Could be improved in a separate PR though 🖖

src/Sylius/Bundle/ApiBundle/Controller/RemoveCatalogPromotionAction.php
```diff
@@ -0,0 +1,45 @@
...
+            $this->catalogPromotionRemovalProcessor->removeCatalogPromotion($request->attributes->get('code'));
+
+            return new Response(status: Response::HTTP_ACCEPTED);
+        } catch (InvalidArgumentException) {
+            return new Response(status: Response::HTTP_NOT_FOUND);
+        } catch (InvalidCatalogPromotionStateException $exception) {
+            return new Response(
+                json_encode(["message" => $exception->getMessage(), 'code' => Response::HTTP_BAD_REQUEST]),
+                Response::HTTP_BAD_REQUEST,
+            );
```
Zales0123: Is this response content consistent with other API endpoints? We should pay attention to that 🚀

src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/RemoveInactiveCatalogPromotionHandler.php
```diff
@@ -0,0 +1,51 @@
...
+
+        if (null === $catalogPromotion) {
+            return;
+        }
+
+        if ($catalogPromotion->getState() !== CatalogPromotionStates::STATE_INACTIVE) {
```
Zales0123, line 40: Following [Adam's comment ](https://github.com/Sylius/Sylius/pull/14270/files#r970407617), I believe **need** to keep the `Inactive` in the name of the class, as it's an unchangeable functionality of it. UNLESS (😄) we extract this condition into a separate service (`RemovedCatalogPromotionCriteriaChecker`?) and make it customisable 💥 Which I believe should be done to improve extendability, but also could be done as an improvement after the merge 🖖

src/Sylius/Bundle/CoreBundle/CatalogPromotion/Exception/InvalidCatalogPromotionStateException.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\CoreBundle\CatalogPromotion\Exception;
+
+final class InvalidCatalogPromotionStateException extends \RuntimeException
```
Zales0123: How about putting it in the `Promotion` component? 🤔

src/Sylius/Bundle/CoreBundle/CatalogPromotion/Command/RemoveInactiveCatalogPromotion.php
Rafikooo: For now, I have left `RemoveInactiveCatalogPromotion`. Referring to [Mateusz's comment](https://github.com/Sylius/Sylius/pull/14270/files#r970658689)
if we introduce the aforementioned criteria checker, I'm in favor of changing it  🚀

src/Sylius/Bundle/CoreBundle/CatalogPromotion/CommandHandler/RemoveInactiveCatalogPromotionHandler.php
Rafikooo: I left the handler name `RemoveInactiveCatalogPromotion` because I'm not sure how to implement it, more specifically, what the flow would be. Only one criteria checker per action or many. I thought of a tagged iterator ( for multi-check ) or additional custom stamps ( for single check). I need to discuss this more :)

src/Sylius/Bundle/CoreBundle/Resources/config/app/messenger.yaml
Rafikooo: I've changed the configuration according to your proposal ✋

src/Sylius/Bundle/CoreBundle/CatalogPromotion/Exception/InvalidCatalogPromotionStateException.php
Rafikooo: Done ✋

src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php
Rafikooo: Done as well

src/Sylius/Bundle/ApiBundle/Controller/RemoveCatalogPromotionAction.php
Rafikooo: I improved it a little bit by using `JsonReponse`. I wanted to be consistent with, for instance, `404` response format
image
image

I'm not sure about the blank `Response` vs `JsonResponse` since the first returns no content and the second returns an empty JSON object
image

src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php
Rafikooo: I chose a slightly different naming. `removeInactiveCatalogPromotion` would work but in line with [this](https://github.com/Sylius/Sylius/pull/14270#discussion_r970658689) discussion I assumed it would be `removeCatalogPromotion` in the future, which is the same name as the public method name of this processor

src/Sylius/Bundle/ApiBundle/Controller/RemoveCatalogPromotionAction.php
```diff
@@ -0,0 +1,46 @@
...
+        private CatalogPromotionRemovalProcessorInterface $catalogPromotionRemovalProcessor
+    ) {
+    }
+
+    public function __invoke(Request $request): Response
+    {
+        try {
+            $this->catalogPromotionRemovalProcessor->removeCatalogPromotion($request->attributes->get('code'));
+
+            return new Response(status: Response::HTTP_ACCEPTED);
```
GSadee, line 36:
```suggestion
            return new JsonResponse(status: Response::HTTP_ACCEPTED);
```
?

src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php
```diff
@@ -0,0 +1,91 @@
...
+    ) {
+    }
+
+    public function removeCatalogPromotion(string $catalogPromotionCode): void
+    {
+        /** @var CatalogPromotionInterface|null $catalogPromotion */
+        $catalogPromotion = $this->getCatalogPromotion($catalogPromotionCode);
+
+        if ($catalogPromotion->getState() === CatalogPromotionStates::STATE_INACTIVE) {
+            $this->announceInactiveCatalogPromotionRemoval($catalogPromotionCode);
```
coldic3, line 40: If in this case Catalog Promotion goes into the `processing` state and then is removed?

src/Sylius/Bundle/CoreBundle/CatalogPromotion/Processor/CatalogPromotionRemovalProcessor.php
```diff
@@ -0,0 +1,91 @@
...
+
+        throw new \DomainException('Invalid catalog promotion state.');
+    }
+
+    private function announceCatalogPromotionEnd(string $catalogPromotionCode): void
+    {
+        $this->eventBus->dispatch(new CatalogPromotionEnded($catalogPromotionCode));
+    }
+
+    private function announceInactiveCatalogPromotionRemoval(string $catalogPromotionCode): void
```
GSadee, line 70: You are dispatching here a command, not an event, so `announce` is not the best word in my opinion, I would be for simplifying it to:
```suggestion
    private function removeInactiveCatalogPromotion(string $catalogPromotionCode): void
```

GSadee (approved):

GSadee:
Thanks, Rafał! :1st_place_medal:

PR #14162, opened 2022-07-19
[Symfony 6] Enable authenticator manager

loic425:
| Q               | A                                                            
|-----------------|--------------------------------------------------------------
| Branch?         | 1.12
| Bug fix?        | yes (For Symfony 6)                                                       
| New feature?    | no                                                       
| BC breaks?      | no                                                       
| Deprecations?   | no
| Related tickets | partially #13274             
| License         | MIT

PR #14283, opened 2022-09-07
[Symfony 6] Enable authenticator manager

ernestWarwas:
| Q               | A                                                            
| Branch?         | 1.12
| Bug fix?        | yes (For Symfony 6)                                                       
| New feature?    | no                                                       
| BC breaks?      | no                                                       
| Deprecations?   | no
| Related tickets | relates to #14162         
| License         | MIT

### DIFF
--- a/UPGRADE-1.12.md
+++ b/UPGRADE-1.12.md
@@ -38,6 +38,42 @@ with test or remove these services with complier pass.
 
 7. The `Sylius\Component\Promotion\Event\CatalogPromotionFailed` has been removed as it is not used anymore.
 
+8. Due to updating to Symfony 6 security file was changed to use the updated security system so you need to adjust your `config/packages/security.yaml` file:
+    
+    ```diff
+   security:
+        - always_authenticate_before_granting: true
+        + enable_authenticator_manager: true
+   ```
+
+   and you need to adjust all of your firewalls like that:
+
+   ```diff
+        admin:
+            # ...
+            form_login:
+                # ...
+                - csrf_token_generator: security.csrf.token_manager
+                + enable_csrf: true
+                # ...
+        new_api_admin_user:
+            # ...
+            - anonymous: true
+            + entry_point: jwt
+            # ...
+            - guard:
+                 # ...
+            + jwt: true
+   ```
+   
+    and also you need to adjust all of your access_control like that:
+    
+    ```diff
+        - - { path: "%sylius.security.admin_regex%/forgotten-password", role: IS_AUTHENTICATED_ANONYMOUSLY }
+   
+        + - { path: "%sylius.security.admin_regex%/forgotten-password", role: PUBLIC_ACCESS }
+    ```
+
 ### Asset management changes
 
 We updated gulp-sass plugin as well as the sass implementation we use to be compatible with most installation
--- a/src/Sylius/Bundle/ApiBundle/Context/TokenBasedUserContext.php
+++ b/src/Sylius/Bundle/ApiBundle/Context/TokenBasedUserContext.php
@@ -25,18 +25,9 @@ final class TokenBasedUserContext implements UserContextInterface
 
     public function getUser(): ?UserInterface
     {
-        $token = $this->tokenStorage->getToken();
-        if ($token === null) {
-            return null;
-        }
+        /** @var UserInterface|null $user */
+        $user = $this->tokenStorage->getToken()?->getUser();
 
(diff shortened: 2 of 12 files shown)


src/Sylius/Bundle/ApiBundle/Context/TokenBasedUserContext.php
```diff
@@ -30,9 +30,10 @@ public function getUser(): ?UserInterface
             return null;
         }
 
-        /** @var UserInterface|string $user */
+        /** @var UserInterface|string|null $user */
         $user = $token->getUser();
-        if (is_string($user)) {
+
+        if (is_string($user) || null === $user) {
```
coldic3:
```suggestion
        if (!$user instanceof UserInterface) {
```

src/Sylius/Bundle/CoreBundle/Validator/Constraints/UniqueReviewerEmailValidator.php
```diff
@@ -56,8 +56,12 @@ public function validate($value, Constraint $constraint): void
         }
     }
 
-    private function getAuthenticatedUserEmail(TokenInterface $token): ?string
+    private function getAuthenticatedUserEmail(?TokenInterface $token): ?string
     {
+        if (null === $token) {
+            return null;
+        }
```
coldic3, line 64: IMO It could be moved after the line `$token = $this->tokenStorage->getToken();` and if token is `null` we can instantly return nothing in the `validate` method.

coldic3 (approved):

GSadee (commented):
You need to add changes from `security.yaml` to the UPGRADE file

src/Sylius/Bundle/CoreBundle/Validator/Constraints/UniqueReviewerEmailValidator.php
NoResponseMate: Alternatively, move `$token = $this->tokenStorage->getToken();` inside `getAuthenticatedUserEmail` itself since the token is not used for anything else.
Either way, I see a bunch of problems with this validator, refactoring it would be a good idea, but that's outside the scope of this PR.

src/Sylius/Bundle/ApiBundle/EventListener/ApiCartBlamerListener.php
```diff
@@ -33,13 +35,13 @@ public function __construct(
     ) {
     }
 
-    public function onInteractiveLogin(InteractiveLoginEvent $interactiveLoginEvent): void
+    public function onSuccessLogin(LoginSuccessEvent $loginSuccessEvent): void
```
NoResponseMate:
```suggestion
    public function onSuccessfulLogin(LoginSuccessEvent $loginSuccessEvent): void
```
or
```suggestion
    public function onLoginSuccess(LoginSuccessEvent $loginSuccessEvent): void
```

src/Sylius/Bundle/ApiBundle/Context/TokenBasedUserContext.php
```diff
@@ -30,9 +30,10 @@ public function getUser(): ?UserInterface
...
             return null;
```
NoResponseMate: Most of this method could be condensed into
```php
$user = $this->tokenStorage()->getToken()?->getUser();

return $user instanceof UserInterface ? $user : null;
```
I'm guessing we'll be only supporting PHP 8 and up when the support for symfony 6 is finished.

NoResponseMate (approved):

UPGRADE-1.12.md
```diff
@@ -38,6 +38,42 @@ with test or remove these services with complier pass.
...
+            + jwt: true
+   ```
+   
+    and also you need to adjust all of your access_control like that:
+    
+    ```diff
+        - - { path: "%sylius.security.admin_regex%/forgotten-password", role: IS_AUTHENTICATED_ANONYMOUSLY }
+   
+        + - { path: "%sylius.security.admin_regex%/forgotten-password", role: PUBLIC_ACCESS }
+    ```
```
coldic3, line 75: In that way, pluses and minuses in `diff` blocks may not be properly highlighted in the markdown file. It's not a big deal but just wanted to mention it ;)

image

coldic3 (approved):

GSadee (approved):

GSadee:
Thanks, Ernest! :tada:

PR #14289, opened 2022-09-08
[Api] Change int to integer in ProductVariant and ShippingMethod documentation normalizers

EsamAl-Dabagh:
In the `ProductVariantDocumentationNormalizer`, the `price` and `originalPrice` properties have a manual configuration pushed to the array. In this config, `type` is defined as `int`. The issue here is that `int` is not a valid OpenApi data type. 

The same issue occurs with the `price` property in `ShippingMethodDocumentationNormalizer`.  

This fix simply changes `int` to `integer` for the above three properties. 

| Branch?         | 1.11             |
| BC breaks?      | no                                                     |
| Deprecations?   | no |
| Related tickets |                   |


GSadee (approved):

GSadee:
Thanks, Esam! :1st_place_medal:

PR #14290, opened 2022-09-08
[Catalog Promotions] Introduce the "processing" state and drop the unused "failed" step

coldic3:
| Deprecations?   | no |

image
image


UPGRADE-1.12.md
```diff
@@ -32,6 +32,12 @@ and should be used only this way.
 are now included in every env starting with `test` keyword. If you wish to not have them, either you need to rename your env to not start 
 with test or remove these services with complier pass.
 
+6. The following listeners have been removed as they are not used anymore:
+- `Sylius\Bundle\CoreBundle\CatalogPromotion\Listener\CatalogPromotionFailedListener`
+- `Sylius\Bundle\CoreBundle\CatalogPromotion\Listener\CatalogPromotionUpdateFailedMessageListener`
```
Zales0123, line 37: I'm wondering if we should remove it 🤔 In theory, listeners are not included in our [BC policy](https://docs.sylius.com/en/1.12/book/organization/backward-compatibility-promise.html?highlight=BC%20policy#event-listeners), but their result is (`Behaviour they’re providing (the end result) is still included in BC promise.`). Therefore, it's possible one could already implement functionality in their application that uses these listeners and the removed event as well. It's obviously a dead code in Sylius, but does not necessarily have to be such in the project that is using Sylius.
On the other hand, it would be sad if we have to keep these unused classes until Sylius 2.0 in our codebase, so it's tempting to remove them :D I need a second opinion cc @lchrusciel @GSadee

tests/Functional/CatalogPromotionAnnouncerTest.php
```diff
@@ -0,0 +1,63 @@
...
+
+namespace Sylius\Tests\Functional;
+
+use Fidry\AliceDataFixtures\LoaderInterface;
+use Fidry\AliceDataFixtures\Persistence\PurgeMode;
+use Sylius\Bundle\CoreBundle\CatalogPromotion\Announcer\CatalogPromotionAnnouncer;
+use Sylius\Component\Core\Model\CatalogPromotion;
+use Sylius\Component\Core\Model\CatalogPromotionInterface;
+
+class CatalogPromotionAnnouncerTest extends AbstractWebTestCase
```
Zales0123: Could be final

GSadee (approved):

UPGRADE-1.12.md
GSadee: I would be for removing them as listeners are excluded from the BC policy (as you wrote) and these ones do not provide any behaviour for the Sylius application without customization as they were a dead code. However, I'm not sure if it is a proper approach 🤔

NoResponseMate (approved):

UPGRADE-1.12.md
lchrusciel: Just thinking about it, I would vote for removal, as well. Plus - if someone needs it, duplication is super straightforward, we can even add it to UPGRADE file

GSadee (approved):

GSadee:
Thanks, Kevin! :1st_place_medal:

PR #14302, opened 2022-09-14
[maintenance]Do not continue-on-error pipeline

Ferror:
When the pipeline fails then it should FAIL. Let's not continue our pipelines on fail... it's making more harm for the whole ecosystem 😢 

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error


loic425 (approved):

loic425:
Maybe it should be done on 1.11

Rafikooo (approved):

PR #14319, opened 2022-09-16
[Maintenance] Update JS dependencies

jakubtobiasz:
| Deprecations?   | no  |
| Related tickets | -                      |


bin/fix-semantic-css
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
coldic3: https://www.npmjs.com/package/@semantic-ui-react/css-patch WDYT?

coldic3 (approved):

bin/fix-semantic-css
coldic3: In addition, this bash script doesn't cover all cases like BSD OS 😄

jakubtobiasz: Thanks for the package, I've committed replacement of my custom script with this one. It seems to work locally, let's wait for CI.

Zales0123 (approved):
Is there any info we should put into the UPGRADE file to make an upgrade easier? Nevertheless, seems good 👍

jakubtobiasz:
@Zales0123 In terms of this PR no actions are required on the end user (developer) side, so there's nothing to add to the UPGRADE file.

Zales0123:
Thank you, @jakubtobiasz! 🎉

PR #14328, opened 2022-09-19
[Template][Events] Introduce Context Providers for Template Blocks

coldic3:
| Deprecations?   | no |
| Related tickets | https://github.com/Sylius/Sylius/issues/10997#issuecomment-993627014                      |

Since `Sylius\Bundle\UiBundle\Block\BlockEventListener` is deprecated, there is no easy way to add additional data to the template context before rendering. This PR introduces `ContextProviderInterface` that allows you to implement your own Context Provider for template blocks.


src/Sylius/Bundle/UiBundle/Renderer/TwigTemplateBlockRenderer.php
```diff
@@ -13,23 +13,29 @@
...
 {
-    public function __construct(private Environment $twig)
+    public function __construct(private Environment $twig, private ContainerInterface $container)
     {
     }
 
     public function render(TemplateBlock $templateBlock, array $context = []): string
     {
+        $contextProvider = $this->container->get($templateBlock->getContextProviderClass());
+        Assert::isInstanceOf($contextProvider, ContextProviderInterface::class);
```
Zales0123: I understand what is the concept here, but I believe we definitely should not go this way :dancer: I mean - way of injecting the container here. It's an obvious consequence of the design of the proposed solution, so maybe we should rearrange it a little bit. 
Let's start simple and have one `DefaultContextProvider` that can behave [as already proposed](https://github.com/Sylius/Sylius/pull/14328/files#diff-c25841738287b8332660025dab8f506a4d051c5a0d75f990eb9c477640dc6e43R20). We can maybe pass a `TemplateBlock::$name` to the `provide` function as `string $blockName` parameter, so it would be easier to provide block-specific context providing functionality. This way we avoid container injection and simplify the configuration while still keeping the possibility to extend the feature later (with, e.g. a strategy pattern for context providers). WDYT? Or am I missing something? 🖖

coldic3: Totally agree. Thank you for the hints. I will try to refactor it with a strategy pattern 👍️

coldic3: Done :slightly_smiling_face:  Could you take a look now?

src/Sylius/Bundle/UiBundle/ContextProvider/ContextProviderInterface.php
```diff
@@ -0,0 +1,26 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\UiBundle\ContextProvider;
+
+use Sylius\Bundle\UiBundle\Registry\TemplateBlock;
+
+/**
+ * @experimental
+ */
```
GSadee, line 20:
```suggestion
/** @experimental */
```

src/Sylius/Bundle/UiBundle/ContextProvider/LegacySonataEventContextProvider.php
```diff
@@ -0,0 +1,38 @@
...
+
+declare(strict_types=1);
+
+namespace Sylius\Bundle\UiBundle\ContextProvider;
+
+use Sylius\Bundle\UiBundle\Registry\TemplateBlock;
+
+/**
+ * @internal
+ */
```
GSadee:
```suggestion
/** @internal */
```

src/Sylius/Bundle/AdminBundle/Resources/config/app/events.yaml
```diff
@@ -192,8 +164,6 @@ sylius_ui:
                 legacy:
                     template: "@SyliusUi/Block/_legacySonataEvent.html.twig"
                     priority: 0
-                    context:
-                        event: sylius.admin.layout.stylesheets
```
GSadee: Why are you removing all events? Are they not needed for legacy sonata to generate proper event names?

coldic3: I did `LegacySonataEventContextProvider` that handles it.

GSadee: Yes, I see that, but it takes the event name from template block, but sometimes (not in above example) the defined event for sonata is different than the template block 🤔 Maybe I misunderstand something

coldic3: Right, sometimes they are postfixed so I use `sonata_postfix` in the context in such cases.

coldic3: Hmm but I see there are also 3 with `Intentionally using "create" event for backwards compatibility` comment and they are different than the event name – good point. I should take them into account somehow.

coldic3: @GSadee As you noticed there are much more sonata events with a different name than the template block names. Therefore I've deleted `LegacySonataEventContextProvider` completely.

src/Sylius/Bundle/UiBundle/ContextProvider/ContextProviderInterface.php
coldic3: I forgot to refactor it but in the next PR, I will completely remove `@experimental` tags so I would be for leaving it as it is for now.

coldic3: And all of them, I guess, are written with 3 lines so it will be easier to do search&replace :grinning:

GSadee (approved):

GSadee:
Thank you, Kevin! :tada:

PR #14344, opened 2022-09-22
[Behat][PostgreSQL] Fix removing items from cart

GSadee:
| Branch?         | 1.12|
| Bug fix?        | yes?                                                       |
| Deprecations?   | no|
| Related tickets | fixes #14336 |

It is a little bit strange to me, because the build was green (yesterday) in PR with enabling Behat tests for Postgres, but it seems that this scenario is failing (randomly?) because of the different order of items compared to MySQL. As I've reproduced that locally, I was able to fix that 😃 

Failing build: https://github.com/Sylius/Sylius/actions/runs/3099676219/jobs/5026644424#step:26:517


Zales0123 (approved):

Zales0123:
Thanks, Grzegorz! :tada:

PR #14376, opened 2022-09-28
Fix no-js Behat scenarios for Symfony 6

Zales0123:
| Branch?         | symfony-6  |
| New feature?    | no |
| BC breaks?      | no |
| Deprecations?   | no |
| Related tickets | contributes to https://github.com/Sylius/Sylius/issues/13274 |

~I've disabled most of the build and left only one Behat build for Symfony 6 to speed up the feedback loop :dancer:~ EDIT: all of them returned 🚀 

~As far as I see, the most important problem right now is the `PathPrefixProvider` service, which uses the `UserContext`... which cannot resolve the AdminUser in admin scenarios :| Hopefully, I will manage to fix it soon, and then I will restore the rest of the builds 🖖~ EDIT: see [this comment](https://github.com/Sylius/Sylius/pull/14376#discussion_r984262319)


src/Sylius/Behat/Service/Converter/IriConverter.php
```diff
@@ -0,0 +1,218 @@
...
+use ApiPlatform\Core\Util\ResourceClassInfoTrait;
+use Symfony\Component\PropertyAccess\PropertyAccess;
+use Symfony\Component\PropertyAccess\PropertyAccessorInterface;
+use Symfony\Component\Routing\Exception\ExceptionInterface as RoutingExceptionInterface;
+use Symfony\Component\Routing\RouterInterface;
+
+/**
+ * Copied from src/Bridge/Symfony/Routing/IriConverter.php to add getIriFromItemInSection method
+ */
+final class IriConverter implements IriConverterInterface
```
Zales0123, line 45: This is obviously a hot-fix 🖖 The problem was, after recent changes in Symfony (between versions 5 and 6), there were some services that were not reset before and are now, and one of them is the token storage. That means, that our `PathPrefixProvider` which was *stateful* before (as it uses the `UserContext` to resolve the prefix) does not work [on some occasions](https://github.com/Sylius/Sylius/pull/14376/files#diff-649a0eb931d4880a02e494ed534d1b956a49cf22f191382522150db425c7d9f0R1196) (e.g. does not provide an admin user). The hot-fix which I propose **for now,** is to make it stateless by passing the desired section (admin/shop) in which we're resolving the Behat scenario. The truth is **we always now** in which section we are in Behat steps, so this resolver could be stateless for Behat from the very beginning :dancer: 
I had to copy the whole `IriConverter` class to add [a new method](https://github.com/Sylius/Sylius/pull/14376/files#diff-23ca6bd44e152f8a107cfd8e51d1cb5c25c59438a1de3271b77cabce9065a6c9R148), I will work on a better solution (or at least a cleaner one), but it can count as refactor in Behat which can be done even after 1.12-beta 🖖

src/Sylius/Bundle/CoreBundle/DependencyInjection/Compiler/BackwardsCompatibility/Symfony6PrivateServicesPass.php
```diff
@@ -0,0 +1,40 @@
...
+        // Before we fix it let's make them public to keep backward compatibility
+
+        $this->makeServicePublic('form.factory', $container);
+        $this->makeServicePublic('security.authorization_checker', $container);
+        $this->makeServicePublic('security.authorization_checker', $container);
+        $this->makeServicePublic('security.csrf.token_manager', $container);
+        $this->makeServicePublic('security.csrf.token_manager', $container);
+        $this->makeServicePublic('security.token_storage', $container);
+        $this->makeServicePublic('security.token_storage', $container);
+        $this->makeServicePublic('validator', $container);
```
Zales0123, line 25: Another hot-fix, but the easiest solution for now. We still use in some of our services a container injection :/ The problem is, some of these happen in Sylius itself, and some in extracted bundles (like Resource bundle, for example). So I make these services public temporarily, until we remove this abomination from the code (without breaking the backward compatibility, obviously 🎉)

src/Sylius/Bundle/CoreBundle/DependencyInjection/Compiler/BackwardsCompatibility/Symfony6PrivateServicesPass.php
```diff
@@ -0,0 +1,40 @@
...
+
+        if (str_starts_with((string) $container->getParameter('kernel.environment'), 'test')) {
+            $this->makeServicePublic('filesystem', $container);
+            $this->makeServicePublic('session.factory', $container);
+            $this->makeServicePublic('sylius.command_bus', $container);
+        }
```
Zales0123, line 31: A slightly different topic - these are services used in our PHPUnit tests and, as far as I know, they _have_ to be taken directly from the container :/ That's why I make them public in a test (or test_cached) environment 🚀

src/Sylius/Bundle/CoreBundle/DependencyInjection/Compiler/BackwardsCompatibility/Symfony6PrivateServicesPass.php
```diff
@@ -0,0 +1,37 @@
...
+    public function process(ContainerBuilder $container): void
+    {
+        // Some services are still fetched directly from the container in both Sylius and its extracted bundles (e.g. ResourceBundle)
+        // Before we fix it let's make them public to keep backward compatibility
+
+        $this->makeServicePublic('form.factory', $container);
+        $this->makeServicePublic('security.authorization_checker', $container);
+        $this->makeServicePublic('security.csrf.token_manager', $container);
+        $this->makeServicePublic('security.token_storage', $container);
+        $this->makeServicePublic('validator', $container);
```
Zales0123, line 22: Another hot-fix, but the easiest solution for now. We still use in some of our services a container injection :/ The problem is, some of these happen in Sylius itself, and some in extracted bundles (like Resource bundle, for example). So I make these services public temporarily until we remove this abomination from the code (without breaking the backward compatibility, obviously 🎉)

src/Sylius/Bundle/CoreBundle/DependencyInjection/Compiler/BackwardsCompatibility/Symfony6PrivateServicesPass.php
```diff
@@ -0,0 +1,37 @@
...
+
+        if (str_starts_with((string) $container->getParameter('kernel.environment'), 'test')) {
+            $this->makeServicePublic('filesystem', $container);
+            $this->makeServicePublic('session.factory', $container);
+            $this->makeServicePublic('sylius.command_bus', $container);
+        }
```
Zales0123, line 28: Another hot-fix, but the easiest solution for now. We still use in some of our services a container injection :/ The problem is, some of these happen in Sylius itself, and some in extracted bundles (like Resource bundle, for example). So I make these services public temporarily, until we remove this abomination from the code (without breaking the backward compatibility, obviously 🎉)

Zales0123, line 28: A slightly different topic - these services are used in our PHPUnit tests and, as far as I know, they _have to_ be taken directly from the container. Therefore I make them public in the test (and test_cached) environments 🚀

loic425: Are you sure, on PHPUnit a service can be retrieved from the container even if it's private, it's not the same container on PHPUnit tests

loic425: On KernelTestCase:

```
try {
    return self::$kernelContainer->get('test.service_container');
} catch (ServiceNotFoundException $e) {
    throw new \LogicException('Could not find service "test.service_container". Try updating the "framework.test" config to "true".', 0, $e);
}
```

jakubtobiasz: > The only remaining drawback of "private services by default" is that testing was harder than before. Some developers even defined some config in the test environment to make all services public in tests. **In Symfony 4.1, we did the same and now tests allow fetching private services by default.**

Afaik there's no need to mark services as public for testing purposes 🤔.

ref: https://symfony.com/blog/new-in-symfony-4-1-simpler-service-testing

jakubtobiasz (approved):

Rafikooo (approved):

GSadee (approved):

GSadee:

src/Sylius/Bundle/CoreBundle/DependencyInjection/Compiler/BackwardsCompatibility/Symfony6PrivateServicesPass.php
Zales0123: I was 100% sure it was failing and I was also surprised 😱 https://github.com/Sylius/Sylius/pull/14407 fix proposed

PR #14398, opened 2022-10-03
Allow ramsey/uuid ^4.0 to avoid PHP 8.1 warning

vasilvestre:
| Q               | A                                                            |
|-----------------|--------------------------------------------------------------|
| Branch?         | 1.12                   |
| Bug fix?        | no                                                       |
| New feature?    | no                                                       |
| BC breaks?      | no                                                       |
| Deprecations?   | no  |
| Related tickets | fixes #X, partially #Y, mentioned in #Z                      |
| License         | MIT                                                          |

There's a type error in 8.1 due to return type. 
`Error: During inheritance of JsonSerializable: Uncaught Behat\\Testwork\\Call\\Exception\\CallErrorException: 8192: Return type of Ramsey\\Uuid\\Uuid::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[\\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in \/var\/www\/html\/vendor\/ramsey\/uuid\/src\/Uuid.php line 216 `

PR #14409, opened 2022-10-04
Allow wider range of ramsey/uuid

vasilvestre:
Fixes change introduced in #14398

We talked with zales and dropping 3.x seems fine. Version 5 will come soon and it seems ok to just support version 4.x right now.


vasilvestre:
@Zales0123 I think you would appreciate to see this PR fast

PR #14414, opened 2022-10-05
[Github Actions] Make timeout for non-JS build longer

Zales0123:
| Branch?         | symfony-6  |
| New feature?    | no |
| BC breaks?      | no |
| Deprecations?   | no |
| Related tickets |  |

On the recent builds, we've experienced a lot of cancelled jobs due to the longer build time. I raise the timeout to 35 minutes, although we should think about fixing the problem rather than overpassing it 🖖


GSadee (approved):

GSadee:
Thank you, Mateusz! :tada:

PR #14422, opened 2022-10-08
[DOCS] Update Sylius components - Addressing

arti0090:

### DIFF
--- a/docs/components_and_bundles/components/Addressing/basic_usage.rst
+++ b/docs/components_and_bundles/components/Addressing/basic_usage.rst
@@ -1,12 +1,5 @@
-.. rst-class:: outdated
-
 Basic Usage
 ===========
-
-.. danger::
-
-   We're sorry but **this documentation section is outdated**. Please have that in mind when trying to use it.
-   You can help us making documentation up to date via Sylius Github. Thank you!
 
 .. _component_addressing_matcher_zone-matcher:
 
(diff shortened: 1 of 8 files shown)


arti0090:
@lchrusciel 
Missing components that I found in composer packagist are:
Calendar / Core / Grid / Resource / Registry
Should I add all of them also?

@CoderMaggie in https://docs.sylius.com/en/latest/components_and_bundles/components/Addressing/basic_usage.html
there is Caution block about "Throws InvalidArgumentException" - I have never meet anything like this in the docs codebase (or maybe missed it), but should we leave it or remove it ?

More of components will come soon 😉

lchrusciel:
Calendar - could be added with one/two sentences about why we created it: "Simple component, that abstracts date object creation, which the main purpose is to ease testing of date depended features" + simple installation and usage guide? 
Core - this is the whole Sylius, so I would skip it. Unless, we would like to make it easier to use Sylus as eCommerce engine ;)  Something for the future
Grid - its docs were consciously moved to https://github.com/Sylius/SyliusGridBundle/tree/1.12/docs. I wouldn't add anything there other then just a link to other documentation
Resource - same as above
Registry - same as above

vvasiloi:
Is this still relevant?

jakubtobiasz:
Let's close it for now. We have more important things to do, and it'll be (or even it is already) covered in the new docs.

PR #14432, opened 2022-10-12
Fix package versions for Symfony 6

loic425:
| Deprecations?   | no |
| Related tickets |                   |


src/Sylius/Bundle/UserBundle/composer.json
```diff
@@ -44,7 +44,7 @@
         "sylius-labs/polyfill-symfony-framework-bundle": "^1.0 || ^1.1",
         "sylius/mailer-bundle": "^1.8 || ^2.0@beta",
         "sylius/resource-bundle": "^1.9 || ^1.10@alpha",
-        "sylius/user": "^1.6 || dev-symfony-6",
+        "sylius/user": "^1.12@beta",
```
lchrusciel, line 47: why do you force v1.12 only?

loic425: cause we use an interface from user component that is new on 1.12

loic425: Capture d’écran du 2022-10-12 10-59-10

Zales0123: I wonder now, shouldn't we have these packages as minimal version at all? 🤔 I mean - the 1.12 version of the package should require `^1.12` of other packages, the 1.11 version should require `^1.11` version of other packages etc... What is the point on allowing `1.8` version of the User Component in the `1.12` version of the UserBundle? 🤷

lchrusciel: Should we update right now all the dependencies? Of course, in separate PR

Zales0123: I guess so... The only thing I'm worried about is potential problems with the upgrade process, as it was _technically_ possible to use e.g. SyliusLocaleBundle 1.6 with [SyliusProductBundle 1.12](https://github.com/Sylius/Sylius/blob/1.12/src/Sylius/Bundle/ProductBundle/composer.json#L32) :dancer: I do not believe anyone does it 😄 so maybe mentioning this change in the UPGRADE file would be enough 🖖

Zales0123:
Merged in https://github.com/Sylius/Sylius/pull/14464. Thanks, Loic! 🖖

PR #14434, opened 2022-10-12
Fix package alias on API Bundle

loic425:
| Deprecations?   | no |
| Related tickets |                      |

I've fixed that running the following command
```bash
$ vendor/bin/monorepo-builder package-alias
```

### DIFF
--- a/src/Sylius/Bundle/ApiBundle/composer.json
+++ b/src/Sylius/Bundle/ApiBundle/composer.json
@@ -2,10 +2,7 @@
     "name": "sylius/api-bundle",
     "type": "symfony-bundle",
     "description": "APIs management for Symfony projects.",
-    "keywords": [
-        "api",
-        "rest"
-    ],
+    "keywords": ["api", "rest"],
     "homepage": "https://sylius.com",
     "license": "MIT",
     "authors": [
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ApiBundle/composer.json
```diff
@@ -58,7 +55,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.7-dev"
+            "dev-master": "1.9-dev"
```
Zales0123, line 58: I believe it should be `1.12-dev` for all packages. We definitely do not update packages' composer.json files regularly :/

loic425: Yes, indeed and that's why we need to run package alias command on refactor

loic425: But I don't know why this puts 1.9....

Zales0123: Maybe let's change it manually for all the packages and then try to automatize it with some tool in the refactor workflow?

Zales0123: Or maybe we should just delete it, as we're not using `master` branch anymore? 🤔

Zales0123: Or not, we still use it on packages themselves 💃 so maybe it should be removed in the main `composer.json` and fixed in those in packages

Zales0123:
Merged in https://github.com/Sylius/Sylius/pull/14458 🖖

PR #14437, opened 2022-10-12
Lint Twig templates on packages

loic425:
| Branch?         | 1.12                 |
| Deprecations?   | no |

Trying to fix https://github.com/Sylius/Sylius/actions/runs/3568390155/jobs/5997217963

Capture d’écran du 2022-11-28 21-10-33


everwhatever:
Hi @loic425 
Do we still need this feature?

loic425:
@everwhatever Yes that's be cool, cause that prevents the main layout to use external dependencies.
This idea is to have a layout that could be used on Symfony projects (such as Monofony).

To prevent this https://github.com/loic425/sylius-ui/blob/main/tests/Application/templates/bundles/SyliusUiBundle/Form/theme.html.twig

This is a tiny part to provide an UI with the resource & grid bundle.

src/Sylius/Bundle/UiBundle/Resources/views/Form/theme.html.twig
```diff
@@ -1,43 +1,4 @@
-{% extends 'form_div_layout.html.twig' %}
```
lchrusciel, line 1: Can you onboard me with these changes? Linter seems to be complaining about the lack of existence of some functions, but I need help finding them here. You've also moved some parts of the forms to another file. Isn't it a BC-Break?

loic425: I don't think so, the current layout extend a new one which can be use on a Symfony/Monofony project without overriding it.

No CI is green now, the goal for now is to check no dependency is needed for this new layout only.

PR #14449, opened 2022-10-16
[WIP] Chore: convert grid yaml definitions to php

Florian-Merle:
| Deprecations?   | no  |


Florian-Merle:
It's impossible to set the label of a filter to false while configuring a grid using PHP. This causes an issue with the search field for products as it displays it with a label if we don't do anything.

image

I could either change the string filter template to something like this, but I'm not a huge fan of it.

```twig
{% form_theme form '@SyliusUi/Form/theme.html.twig' %}

<div class="sylius-filters__group">
    {% if form.type is defined %}
        {{ form_row(form.type, {'label': filter.label}) }}
    {% endif %}

    {% set label = form.type is defined ? form.value.vars.label : filter.label %}
    {% if filter.options.disableLabel == true %} {# This means we should adapt the grid to add an option to explicitly disable the label #}
        {% set label = false %}
    {% endif %}
    {{ form_row(form.value, {attr: {placeholder: form.value.vars.label}, 'label': label}) }}
</div>
```

See https://github.com/Sylius/Sylius/blob/1.12/src/Sylius/Bundle/UiBundle/Resources/views/Grid/Filter/string.html.twig

Another solution is to change the signature of the `FilterInterface` so it can accepts either a false, null or string value. See https://github.com/Sylius/SyliusGridBundle/blob/1.12/src/Bundle/Builder/Filter/FilterInterface.php#L24

What do you think ?

Ping @loic425

loic425:
That's a very good initiative, we'll be the refactor team of Sylius :) As we said, that'd be cool to use a converter that users could use to convert their own grids. A non-official one exists and maybe we could just try it and improve it and maybe Moving it to the grid package.

loic425:
> It's impossible to set the label of a filter to false while configuring a grid using PHP. This causes an issue with the search field for products as it displays it with a label if we don't do anything.
> 
> image
> 
> I could either change the string filter template to something like this, but I'm not a huge fan of it.
> 
> ```
> {% form_theme form '@SyliusUi/Form/theme.html.twig' %}
> 
> <div class="sylius-filters__group">
>     {% if form.type is defined %}
>         {{ form_row(form.type, {'label': filter.label}) }}
>     {% endif %}
> 
>     {% set label = form.type is defined ? form.value.vars.label : filter.label %}
>     {% if filter.options.disableLabel == true %} {# This means we should adapt the grid to add an option to explicitly disable the label #}
>         {% set label = false %}
>     {% endif %}
>     {{ form_row(form.value, {attr: {placeholder: form.value.vars.label}, 'label': label}) }}
> </div>
> ```
> 
> See https://github.com/Sylius/Sylius/blob/1.12/src/Sylius/Bundle/UiBundle/Resources/views/Grid/Filter/string.html.twig
> 
> Another solution is to change the signature of the `FilterInterface` so it can accepts either a false, null or string value. See https://github.com/Sylius/SyliusGridBundle/blob/1.12/src/Bundle/Builder/Filter/FilterInterface.php#L24
> 
> What do you think ?
> 
> Ping @loic425

I think we can use 
```php
StringFilter::create('search', ['translation.name'])
    ->addFormOption('label' => false);
```

Florian-Merle:
> I think we can use
> 
> ```
> StringFilter::create('search', ['translation.name'])
>     ->addFormOption('label' => false);
> ```

I've already tried it, and it won't work, because we don't have a `form.type` so it uses the `filter.label`. See https://github.com/Sylius/Sylius/blob/1.12/src/Sylius/Bundle/UiBundle/Resources/views/Grid/Filter/string.html.twig#L7

mamazu:
Just wanted to leave a note here that this pull request contains a lot of duplication with the `setOptions` call. While it could be removed, leaving it there would make it easier to extend in the long run. 

For more information see [sylius/SyliusGridBundle!267](https://github.com/Sylius/SyliusGridBundle/issues/267)

src/Sylius/Bundle/AdminBundle/Resources/config/app/config.yml
```diff
@@ -3,38 +3,39 @@ imports:
...
-    - { resource: "@SyliusAdminBundle/Resources/config/grids/taxon.yml" }
-    - { resource: "@SyliusAdminBundle/Resources/config/grids/zone.yml" }
+    # - { resource: "@SyliusAdminBundle/Resources/config/grids/promotion.yml" }
+    # - { resource: "@SyliusAdminBundle/Resources/config/grids/promotion_coupon.yml" }
+    # - { resource: "@SyliusAdminBundle/Resources/config/grids/shipment.yml" }
+    # - { resource: "@SyliusAdminBundle/Resources/config/grids/shipping_category.yml" }
+    # - { resource: "@SyliusAdminBundle/Resources/config/grids/shipping_method.yml" }
+    # - { resource: "@SyliusAdminBundle/Resources/config/grids/tax_category.yml" }
+    # - { resource: "@SyliusAdminBundle/Resources/config/grids/tax_rate.yml" }
+    # todo make sure the following grid is not used anymore before deleting it completly
```
Florian-Merle, line 36: todo

src/Sylius/Bundle/AdminBundle/Resources/config/app/config.yml
```diff
@@ -3,38 +3,6 @@ imports:
...
-    - { resource: "@SyliusAdminBundle/Resources/config/grids/product_variant.yml" }
-    - { resource: "@SyliusAdminBundle/Resources/config/grids/product_variant_with_catalog_promotion.yml" }
-    - { resource: "@SyliusAdminBundle/Resources/config/grids/promotion.yml" }
-    - { resource: "@SyliusAdminBundle/Resources/config/grids/promotion_coupon.yml" }
-    - { resource: "@SyliusAdminBundle/Resources/config/grids/shipment.yml" }
-    - { resource: "@SyliusAdminBundle/Resources/config/grids/shipping_category.yml" }
-    - { resource: "@SyliusAdminBundle/Resources/config/grids/shipping_method.yml" }
-    - { resource: "@SyliusAdminBundle/Resources/config/grids/tax_category.yml" }
-    - { resource: "@SyliusAdminBundle/Resources/config/grids/tax_rate.yml" }
-    - { resource: "@SyliusAdminBundle/Resources/config/grids/taxon.yml" }
```
Florian-Merle, line 36: This grid has been removed and not replaced because it has not been in used  since before 1.0.0. See  https://github.com/Sylius/Sylius/commit/95b1cbc000eab9b9e06b495a34afff82e942efbc#diff-8b388c80c5f84373ad2d744ae1ead4cbf0187a798596191d65cece1b48315e35R7

mamazu:
Is this still planned? Otherwise I might be able to pick this up.

loic425:
> Is this still planned? Otherwise I might be able to pick this up.

You're right, I was talking to Florian about this during api con. I think we should finish this for 2.0

mamazu:
Replaced by the effort in #18136

PR #14452, opened 2022-10-17
:us: Add missing translation (promo coupon error)

Nek-:
| Branch?         | 1.10                  |
| Deprecations?   | no |

Not sure where it should be added, but I can only add it here, sorry if it was supposed to be added on crowdin.


Zales0123:
We're managing en translations in the main repo so it's 👍... but as far as I see it's already there, a few lines above https://github.com/Sylius/Sylius/blob/1.10/src/Sylius/Bundle/CoreBundle/Resources/translations/validators.en.yml#L76-L78 😅 Or am I missing any difference?

Nek-:
Hum, it's actually not present in french and also not possible to translate on crowdin...

Nek-:
My bad I use wrongly crowdin, I needed to look in 1.10 version translations.

PR #14469, opened 2022-10-21
PoC Symfony workflow

loic425:
| Q               | A                                                            |
|-----------------|--------------------------------------------------------------|
| Branch?         | 1.13                  |
| Bug fix?        | no                                                       |
| New feature?    | yes                                                       |
| BC breaks?      | no                                                       |
| Deprecations?   | no |
| Related tickets |      |
| License         | MIT                                                          |

### Catalog promotion

- [x] State machine configuration

### Order

- [x] State machine configuration

**Before Placed Order**
- [x] sylius_assign_number
- [x] sylius_assign_token

**After Placed Order**
- [x] sylius_request_shipping
- [x] sylius_request_payment
- [x] sylius_create_payment
- [x] sylius_create_shipment
- [x] sylius_hold_inventory
- [x] sylius_increment_promotions_usages
- [x] sylius_save_addresses_on_customer
- [x] sylius_set_order_immutable_names

**After Canceled Order**
- [x] sylius_cancel_payment
- [x] sylius_cancel_shipment
- [x] sylius_cancel_order_payment
- [x] sylius_cancel_order_shipment
- [x] sylius_cancel_order_inventory
- [x] sylius_decrement_promotions_usages

### Order checkout

- [x] State machine configuration

**After Completed Checkout**
- [x] sylius_create_order
- [x] sylius_save_checkout_completion_date
- [x] sylius_control_payment_state
- [x] sylius_control_shipping_state

**After Selected Shipping**
- [x] sylius_process_cart
- [x] sylius_skip_payment

**After Addressed**
- [x] sylius_process_cart

**After Selected Payment**
- [x] sylius_process_cart

**After Skipped Shipping**
- [x] sylius_process_cart

#### After Skipped Payment
- [x] sylius_process_cart

### Order payment

- [x] State machine configuration

**After Paid**
- [x] sylius_order_paid
- [x] sylius_resolve_state

### Order shipping

- [x] State machine configuration

**After Shipped**
- [x] sylius_resolve_state

### Payment

- [x] State machine configuration

**After Failed Payment**
- [x] sylius_process_order

**After Canceled Payment**
- [x] sylius_process_order

**After Completed Payment**
- [x] sylius_resolve_state

**After Processed Payment**
- [x] sylius_resolve_state

**After Refunded Payment**
- [x] sylius_resolve_state

**After Authorized Payment**
- [x] sylius_resolve_state

### Product Review

- [x] State machine configuration

**After Accepted Product Review**
- [x] sylius_update_rating

[...]

PR #14494, opened 2022-10-31
[SF-Workflow] Add Catalog promotion state machine

loic425:
| Branch?         | sf-workflow                  |
| Deprecations?   | no |
| Related tickets | partially #14469                      |

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -126,6 +126,7 @@
         "symfony/twig-bundle": "^5.4 || ^6.0",
         "symfony/validator": "^5.4 || ^6.0",
         "symfony/webpack-encore-bundle": "^1.15",
+        "symfony/workflow": "^5.4 || ^6.0",
         "symfony/yaml": "^5.4 || ^6.0",
         "twig/intl-extra": "^2.12 || ^3.4",
         "twig/twig": "^2.12 || ^3.3",
--- a/config/packages/workflow.yaml
+++ b/config/packages/workflow.yaml
@@ -1,0 +1,2 @@
+framework:
+    workflows: null
--- a/config/packages/workflow.yaml
+++ b/config/packages/workflow.yaml
@@ -1,2 +1,0 @@
-framework:
-    workflows: null
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+use Sylius\Component\Core\Model\CatalogPromotion;
+use Symfony\Config\FrameworkConfig;
+
+return static function (FrameworkConfig $framework): void {
+    $catalogPromotion = $framework->workflows()->workflows('sylius_catalog_promotion');
+    $catalogPromotion
+        ->type('state_machine')
+        ->supports([CatalogPromotion::class])
+        ->initialMarking(['active']);
+
+    $catalogPromotion->markingStore()
+        ->type('method')
+        ->property('state');
+
+    $catalogPromotion->place()->name('active');
+    $catalogPromotion->place()->name('inactive');
+    $catalogPromotion->place()->name('processing');
+
+    $catalogPromotion->transition()
+        ->name('activate')
+        ->from(['processing'])
+        ->to(['active']);
+
+    $catalogPromotion->transition()
+        ->name('deactivate')
+        ->from(['processing'])
+        ->to(['inactive']);
+
+    $catalogPromotion->transition()
+        ->name('process')
+        ->from(['inactive', 'active'])
+        ->to(['processing']);
+};
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
@@ -33,16 +33,19 @@ return static function (FrameworkConfig $framework): void {
 
     $catalogPromotion->transition()
         ->name(CatalogPromotionTransitions::TRANSITION_ACTIVATE)
-        ->from([CatalogPromotionStates:: STATE_PROCESSING])
-        ->to([CatalogPromotionStates:: STATE_ACTIVE]);
+        ->from([CatalogPromotionStates::STATE_PROCESSING])
+        ->to([CatalogPromotionStates::STATE_ACTIVE])
+    ;
 
     $catalogPromotion->transition()
-        ->name('deactivate')
-        ->from(['processing'])
-        ->to(['inactive']);
+        ->name(CatalogPromotionTransitions::TRANSITION_DEACTIVATE)
+        ->from([CatalogPromotionStates::STATE_PROCESSING])
+        ->to([CatalogPromotionStates::STATE_INACTIVE])
+    ;
 
     $catalogPromotion->transition()
-        ->name('process')
-        ->from(['inactive', 'active'])
-        ->to(['processing']);
+        ->name(CatalogPromotionTransitions::TRANSITION_PROCESS)
+        ->from([CatalogPromotionStates::STATE_INACTIVE, CatalogPromotionStates::STATE_ACTIVE])
+        ->to([CatalogPromotionStates::STATE_PROCESSING])
+    ;
 };
--- a/src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
+++ b/src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
@@ -29,7 +29,7 @@ return static function (FrameworkConfig $framework): void {
 
     $catalogPromotion->place()->name(CatalogPromotionStates::STATE_ACTIVE);
     $catalogPromotion->place()->name(CatalogPromotionStates::STATE_INACTIVE);
-    $catalogPromotion->place()->name(CatalogPromotionStates:: STATE_PROCESSING);
+    $catalogPromotion->place()->name(CatalogPromotionStates::STATE_PROCESSING);
 
     $catalogPromotion->transition()
         ->name(CatalogPromotionTransitions::TRANSITION_ACTIVATE)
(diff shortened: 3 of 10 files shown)


src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
```diff
@@ -0,0 +1,46 @@
...
+declare(strict_types=1);
+
+use Sylius\Component\Core\Model\CatalogPromotion;
+use Symfony\Config\FrameworkConfig;
+
+return static function (FrameworkConfig $framework): void {
+    $catalogPromotion = $framework->workflows()->workflows('sylius_catalog_promotion');
+    $catalogPromotion
+        ->type('state_machine')
+        ->supports([CatalogPromotion::class])
```
loic425, line 23: Parameter class does not work for now (unknown parameter). Let it as is for now, and let it add a task to resolve these resource classes.

composer.json
```diff
@@ -126,6 +126,7 @@
         "symfony/twig-bundle": "^5.4 || ^6.0",
         "symfony/validator": "^5.4 || ^6.0",
         "symfony/webpack-encore-bundle": "^1.15",
+        "symfony/workflow": "^5.4 || ^6.0",
```
lchrusciel, line 129: I'm wondering if it should be hard dependency

config/packages/workflow.yaml
```diff
@@ -0,0 +1,2 @@
+framework:
+    workflows: null
```
lchrusciel: Is it required to be placed in all packages that will be using this feature? If we will keep the workflow as a hard dependency, should we keep this config in part of a package that will be loaded?

src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
```diff
@@ -0,0 +1,46 @@
...
+        ->initialMarking(['active']);
```
lchrusciel: The initial state should be inactive according to this fragment https://github.com/Sylius/Sylius/blob/1.13/src/Sylius/Component/Promotion/Model/CatalogPromotion.php#L43

What is more, we could use const for that :) 
```suggestion
        ->initialMarking([Sylius\Component\Promotion\Model\CatalogPromotionStates::STATE_INACTIVE]);
```

src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
```diff
@@ -0,0 +1,46 @@
...
+
+    $catalogPromotion->markingStore()
+        ->type('method')
+        ->property('state');
+
+    $catalogPromotion->place()->name('active');
+    $catalogPromotion->place()->name('inactive');
+    $catalogPromotion->place()->name('processing');
```
lchrusciel:
```suggestion
    $catalogPromotion->place()->name(Sylius\Component\Promotion\Model\CatalogPromotionStates::STATE_ACTIVE);
    $catalogPromotion->place()->name(Sylius\Component\Promotion\Model\CatalogPromotionStates::STATE_INACTIVE);
    $catalogPromotion->place()->name(Sylius\Component\Promotion\Model\CatalogPromotionStates:: STATE_PROCESSING);
```

src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
```diff
@@ -0,0 +1,46 @@
...
+
+    $catalogPromotion->transition()
+        ->name('activate')
+        ->from(['processing'])
+        ->to(['active']);
```
lchrusciel: And the same for the other transitions
```suggestion
    $catalogPromotion->transition()
        ->name(Sylius\Component\Promotion\Model\CatalogPromotionTransitions::TRANSITION_ACTIVATE)
        ->from([Sylius\Component\Promotion\Model\CatalogPromotionStates:: STATE_PROCESSING])
        ->to([Sylius\Component\Promotion\Model\CatalogPromotionStates:: STATE_ACTIVE]);
```

src/Sylius/Bundle/CoreBundle/Tests/Workflow/CatalogPromotionWorkflowTest.php
```diff
@@ -0,0 +1,127 @@
...
+
+        $this->assertEquals([
+            'active' => 'active',
+            'inactive' => 'inactive',
+            'processing' => 'processing',
+        ], $stateMachine->getDefinition()->getPlaces());
+    }
+
+    /** @test */
+    public function it_has_activate_transition(): void
```
lchrusciel: I'm not sure about the values of these tests. Perhaps we could just dump the whole config and assert it? Or even better dump whole config and assert it with winzou state machine (or at least compare parts of config)

lchrusciel (commented):
Do I understand it correctly, that for this to work we will have to use this switch in config: https://github.com/Sylius/SyliusResourceBundle/blob/c5974f25ac4a93b1b8ccfef90d386d9ac6243128/docs/state_machine.md#configuring-symfony-workflow-as-state-machine? 

Perhaps we should provide a possibility to switch to workflow just for a given resource? That would help people to migrate to the new structure. 

What is more, we should trigger high-level, end-to-end tests as soon as possible to ensure similar behaviour. Therefore, I would like to set up new GHA job with workflow and just trigger new tests as we go with new implementation

config/packages/workflow.yaml
loic425: You're right, we should enable it on Core bundle.

composer.json
loic425: My idea was to keep the two dependencies as required packages, deprecate the usage of winzou and remove it on Sylius 2.0. Moving winzou on dev requirements should be annoying for migrations to 1.13. I think it's a lot easier here to add it on prod requirements if it'll be the future package.

src/Sylius/Bundle/CoreBundle/Tests/Workflow/CatalogPromotionWorkflowTest.php
loic425: I improved a lot this part with the both configurations. I hope you'll like it.

config/packages/workflow.yaml
loic425: I added this configuration on Core bundle too. Have I to remove it here?

composer.json
lchrusciel: Let's make it hard dependency then 👍

config/packages/workflow.yaml
lchrusciel: yes please, to see if it work without any change in config in end apps

src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
```diff
@@ -0,0 +1,48 @@
...
+        ->supports([CatalogPromotion::class])
+        ->initialMarking([CatalogPromotionStates::STATE_INACTIVE]);
+
+    $catalogPromotion->markingStore()
+        ->type('method')
+        ->property('state');
+
+    $catalogPromotion->place()->name(CatalogPromotionStates::STATE_ACTIVE);
+    $catalogPromotion->place()->name(CatalogPromotionStates::STATE_INACTIVE);
+    $catalogPromotion->place()->name(CatalogPromotionStates:: STATE_PROCESSING);
```
lchrusciel: some linter should be called 
```suggestion
    $catalogPromotion->place()->name(CatalogPromotionStates::STATE_PROCESSING);
```
below as well

src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
lchrusciel: What about the interface? Is there anything else injected into this callable? Do you think that there may be a problem with the order of loading of bundles?

src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
```diff
@@ -0,0 +1,48 @@
...
+
+    $catalogPromotion->transition()
+        ->name('deactivate')
+        ->from(['processing'])
+        ->to(['inactive']);
+
+    $catalogPromotion->transition()
+        ->name('process')
+        ->from(['inactive', 'active'])
+        ->to(['processing']);
```
lchrusciel: can we use consts instead of strings in all these definitions?

src/Sylius/Bundle/CoreBundle/Tests/Workflow/CatalogPromotionWorkflowTest.php
lchrusciel: Tests like `it_can_pass_from_processing_to_active` look good, but the ones in the shape of `it_has_activate_transition` are too fragile, in my opinion. Let's keep only ones in which we are setting object and checking the behaviour of the state machine instead of its configuration

composer.json
Nek-: Is there a discussion somewhere about removing winzou/sm package(s)? I'm interested in motivations of this change

loic425:
@lchrusciel I will fix this one soon. I've just seen your comments. My notifications were not well configured.

loic425:
> Do I understand it correctly, that for this to work we will have to use this switch in config: https://github.com/Sylius/SyliusResourceBundle/blob/c5974f25ac4a93b1b8ccfef90d386d9ac6243128/docs/state_machine.md#configuring-symfony-workflow-as-state-machine?
> 
> Perhaps we should provide a possibility to switch to workflow just for a given resource? That would help people to migrate to the new structure.
> 
> What is more, we should trigger high-level, end-to-end tests as soon as possible to ensure similar behaviour. Therefore, I would like to set up new GHA job with workflow and just trigger new tests as we go with new implementation

That's very interesting. Indead that'd be great to provide this configuration resource by resource :100: 

GH configuration will be configured on next PRs, for now that's not interesting.

src/Sylius/Bundle/CoreBundle/Resources/config/app/workflow/sylius_catalog_promotion.php
loic425: I don't know why it does not work, we have to investigate. I'm sure I've tried to change the order of bundles.

loic425:
> Perhaps we should provide a possibility to switch to workflow just for a given resource? That would help people to migrate to the new structure.

@lchrusciel And there it is https://github.com/Sylius/SyliusResourceBundle/pull/613 :)

jakubtobiasz:
Closing, as we covered this while switching all state machines to Symfony Workflow :). Anyway thanks for your work and inspiration!

PR #14499, opened 2022-10-31
Separate pipeline for non-JS tests on unstable Symfony version

Zales0123:
| Bug fix?        | no |
| Deprecations?   | no |
| Related tickets |  |

To be sure the upcoming Symfony 6.2 version does not result in any **major** problems, let's test it already :dancer: I've decided to add a separate job, to be able to simplify it a little bit. Also, I've added `continue-on-error: true` flag, as it does not have to pass fully (some minor tweaks could be needed after the 6.2 release - and it's **ok**), we just need to see if there are any very serious problems we need to address right now. 

For some reason 🤷‍♂️ If run all together Behat scenarios were spectacularly failing after ~4k of them 💥 As a quick fix for this bug (that needs to be investigated for sure), I've split running Behat scenarios into multiple commands.

I believe we could keep this pipeline after the Symfony 6.2 release and test with it Symfony 6.3 already. Maybe it would make supporting newer Symfony versions easier 🖖

### DIFF
--- a/.github/workflows/application.yml
+++ b/.github/workflows/application.yml
@@ -962,6 +962,160 @@ jobs:
                     path: etc/build/
                     if-no-files-found: ignore
 
+    test-application-without-frontend-unstable-symfony:
+        runs-on: ubuntu-latest
+
+        name: "Test non-JS application with the unstable Symfony version"
+
+        timeout-minutes: 35
+
+        strategy:
+            fail-fast: false
+            matrix:
+                php: [ "8.1" ]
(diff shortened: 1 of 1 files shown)


loic425 (approved):

jakubtobiasz (approved):

TheMilek (approved):

Rafikooo (approved):

ernestWarwas:
Thanks, Mateusz! :tada:

PR #14511, opened 2022-11-02
[maintenance] - Use symfony runtime component

Ferror:
| Deprecations?   | no  |

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -114,7 +114,6 @@
         "symfony/property-info": "^5.4 || ^6.0",
         "symfony/proxy-manager-bridge": "^5.4 || ^6.0",
         "symfony/routing": "^5.4 || ^6.0",
-        "symfony/runtime": "^5.4 || ^6.0",
         "symfony/security-bundle": "^5.4 || ^6.0",
         "symfony/security-core": "^5.4 || ^6.0",
         "symfony/security-csrf": "^5.4 || ^6.0",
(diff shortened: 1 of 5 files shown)


loic425 (approved):

src/Kernel.php
```diff
@@ -33,6 +33,6 @@ protected function getContainerBaseClass(): string
 
     private function isTestEnvironment(): bool
     {
-        return 0 === strpos($this->getEnvironment(), 'test');
+        return str_contains($this->getEnvironment(), 'test');
```
jakubtobiasz: We should go with `str_start_with` instead of looking for `test` in the whole env name.

jakubtobiasz (approved):

composer.json
```diff
@@ -114,6 +114,7 @@
         "symfony/property-info": "^5.4 || ^6.0",
         "symfony/proxy-manager-bridge": "^5.4 || ^6.0",
         "symfony/routing": "^5.4 || ^6.0",
+        "symfony/runtime": "^5.4 || ^6.0",
```
lchrusciel: Can we move it to the dev requirement? So we won't preinstall it for others?

jakubtobiasz: @lchrusciel why? It is pointless as the runtime is also for production. Creating and handling in `index.php` whether we are on dev env or not doesn't seem to bring any benefit.

jakubtobiasz: Primarily, we aim to provide it for Sylius-Standard to allow running Sylius on RoadRunner or FrankenPHP.

Ferror: @lchrusciel that's true. We should move it to require-dev as we don't depend on it in src. We use it just for testing/development

TheMilek (approved):

coldic3 (approved):

coldic3:
Thanks, Zbigniew! :tada:

Issue #14308, opened 2022-09-15
Product taxon position don't affect parent taxon


ehibes:
**Sylius version affected**: 1.11

**Description**  

On the demo store, when I change product position for women t-shirts in `t-shirts/women` taxon, the frontend is reorganized according to the chosen positions. If I make the same on `t-shirts` taxon, the frontend is not affected.  
**Steps to reproduce**  

<img width="1222" alt="Capture d’écran 2022-09-15 à 10 21 27" src="https://user-images.githubusercontent.com/487124/190362773-2b32add9-a28e-44d5-8339-6a0e175e96c4.png">
<img width="1146" alt="Capture d’écran 2022-09-15 à 10 03 24" src="https://user-images.githubusercontent.com/487124/190362829-92b65059-37c2-490e-93b4-54beee5b770e.png">

loic425:
@ehibes Sorry, I've removed your screenshots. Could you add them again please?

Rafikooo:
Hello @ehibes,
thank you for reporting the issue. We confirm that the bug exists. Repair is not in our current priorities, so if you could, don't hesitate to contribute to Sylius. A behat scenario covering this case or even a PR with a solution would be welcome 🚀

GSadee:
Hi @ehibes, I've just wanted to let you know that we will take care of it 😃

ehibes:
I just searched a bit on my side, and it seems that the query returns duplicate products with different positions.
It must come from writing joins, but I'm not expert enough 😄 
![image](https://user-images.githubusercontent.com/487124/197743953-a28e9853-b336-4287-9519-c09e7613cf8e.png)

ehibes:
@Rafikooo @GSadee I made a failing behat scenario to explain the case...
https://github.com/ehibes/Sylius/tree/parent_taxon_product_position

PR #14512, opened 2022-11-03
remove param method request

everwhatever:
| Branch?         | 1.13  |
| Bug fix?        | yes  |
| New feature?    | no |
| BC breaks?      | no |
| Deprecations?   |no |
| Related tickets | [jira ](https://sylius.atlassian.net/browse/SYL-2349) |
| License         | MIT |

### DIFF
--- a/src/Sylius/Bundle/ShopBundle/Resources/config/grids/product.yml
+++ b/src/Sylius/Bundle/ShopBundle/Resources/config/grids/product.yml
@@ -15,6 +15,7 @@ sylius_grid:
                             taxon: "expr:notFoundOnNull(service('sylius.repository.taxon').findOneBySlug($slug, service('sylius.context.locale').getLocaleCode()))"
                             locale: "expr:service('sylius.context.locale').getLocaleCode()"
                             sorting: "expr:service('request_stack').getCurrentRequest().get('sorting', [])"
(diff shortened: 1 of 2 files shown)


src/Sylius/Bundle/ShopBundle/Resources/config/grids/product.yml
```diff
@@ -15,7 +15,6 @@ sylius_grid:
                             taxon: "expr:notFoundOnNull(service('sylius.repository.taxon').findOneBySlug($slug, service('sylius.context.locale').getLocaleCode()))"
                             locale: "expr:service('sylius.context.locale').getLocaleCode()"
                             sorting: "expr:service('request_stack').getCurrentRequest().get('sorting', [])"
-                            includeAllDescendants: "expr:parameter('sylius_shop.product_grid.include_all_descendants')"
```
GSadee: At first glance, it should not be removed, we want to have this parameter set here.

GSadee (changes requested):
It seems like sth that does not resolve the initial problem from the issue: https://github.com/Sylius/Sylius/issues/14308.  You should definitely start from the scenario that reproduces the problem and then fix it properly.

GSadee:
And as it is a bug fix, it should be opened to 1.11 branch 😃

everwhatever:
changes are already on 1.12 branch

PR #14524, opened 2022-11-08
Allows to add one by one the attributes in product edit page by calculating the right length for the input name

### LINKED ISSUE #14521: Only one product attribute is saved when we add them one by one in the product edit page
**Sylius version affected**: 1.10, 1.11

**Description**  

In the admin, when we add attributes to a product one by one and then save it, only the first attribute added is registered.

**Steps to reproduce**  

- Edit or create product 
- Go to attributes tabs
- Choose an attribute in the select and add it 
- Choose another attribute and add it
- Add attribute values and save the product

The first values of the attribute are saved, not the values of the other of attributes added after!

**Possible Solution**  

I think it's the search for the length of the added attributes, in the `src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-product-attributes.js` file, that is the problem. If i replace the JS:

```js
addAttributesNumber($.grep(attributeFormElements, a => $(a).hasClass('attribute')).length);
```

by 

```js
addAttributesNumber(attributeFormElements.find('.attribute').length);
```

The `data-count` value is updated and the saved works.

delyriand:
| Branch?         | 1.10, 1.11, 1.12              |
| Deprecations?   | no |
| Related tickets | fixes #14521                      |

### DIFF
--- a/src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-product-attributes.js
+++ b/src/Sylius/Bundle/UiBundle/Resources/private/js/sylius-product-attributes.js
@@ -128,7 +128,7 @@ const setAttributeChoiceListener = function setAttributeChoiceListener() {
 
         $('#sylius_product_attribute_choice').val('');
 
-        addAttributesNumber($.grep(attributeFormElements, a => $(a).hasClass('attribute')).length);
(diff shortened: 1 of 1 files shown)


delyriand:
Done in #15006

PR #14525, opened 2022-11-08
Adding skipIf for migrations to allow future runs when they should apply

stefantalen:
| Branch?         | 1.11                   |
| Deprecations?   | no |

While upgrading to Sylius 1.11 I came across this issue where these migrations were executed but I hadn't set the `MESSENGER_TRANSPORT_DSN` environment variable yet.

This resulted in a diff on my database after I added the variable.

By using the `skipIf()` for these migrations will not be executed, it will just give a notice:

```
[notice] Migration Sylius\Bundle\CoreBundle\Migrations\Version20220203115813 skipped during Execution. Reason: "MESSENGER_TRANSPORT_DSN was not found or is not using Doctrine"
[notice] Migration Sylius\Bundle\CoreBundle\Migrations\Version20220407131547 skipped during Execution. Reason: "messenger_messages table was not found"
```
They keep getting skipped on future runs, however when you add the environment variable they will be executed and only then.

### DIFF
--- a/src/Sylius/Bundle/CoreBundle/Migrations/Version20220203115813.php
+++ b/src/Sylius/Bundle/CoreBundle/Migrations/Version20220203115813.php
@@ -30,7 +30,9 @@ final class Version20220203115813 extends AbstractMigration
     {
         $this->abortIf($this->connection->getDatabasePlatform()->getName() != 'mysql', 'Migration can only be executed safely on \'mysql\'.');
 
-        if ($this->isUsingDoctrineTransport() && !$schema->hasTable('messenger_messages')) {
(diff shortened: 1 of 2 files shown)


stefantalen:
Thinking about this, should it even be the responsibility of Sylius to create these tables through a migration?

>The ability to automatically generate a migration for the messenger_messages table was introduced in Symfony 5.1 and DoctrineBundle 2.1.

https://symfony.com/doc/5.4/messenger.html#doctrine-transport

Zales0123 (approved):

GSadee (approved):

GSadee:
Thank you, Stefan! :tada:

PR #14482, opened 2022-10-26
Adding batching to the expired carts remover v2

lchrusciel:
| Q               | A                                                            |
|-----------------|--------------------------------------------------------------|
| Branch?         | 1.11 |
| Bug fix?        | yes                                                       |
| New feature?    | no                                                       |
| BC breaks?      | no                                                       |
| Deprecations?   | no |
| Related tickets | Backporting and fixing #14481 |
| License         | MIT                                                          |

PR #14481, opened 2022-10-26
Adding batching to the expired carts remover

mamazu:
| Q               | A                                                            |
|-----------------|--------------------------------------------------------------|
| Branch?         | 1.11                   |
| Bug fix?        | no                                                       |
| New feature?    | no                                                       |
| BC breaks?      | no                                                       |
| Deprecations?   | no |
| Related tickets | -                      |
| License         | MIT                                                          |

When running the clear cart remover it breaks the database or the memory usage, whatever dies first. So in this we flush regularly to also deal with big datasets and lots of carts to remove.

PR #14532, opened 2022-11-11
Fixing the expired cart command

mamazu:
| Branch?         | 1.13                   |
| Deprecations?   | no  |
| Related tickets | related to #14481, #14482                      |

## The problem

Currently when trying to remove a huge amount of expired carts (like 1 million) the process will crash before it even begins. The issue is that in the current implementation `findCartsNotModifiedSince` returns a list of all carts which kills the memory when trying to execute it.

## The solution

Run the find command in a batch and then remove the batch. (The problem is I don't really understand phpspec so someone which more experience with that could maybe help out here.)

I am also using the fact that php has a leaky scope, it's not pretty. But this is the only way to not dispatch more than one `SyliusExpiredCartsEvents::POST_REMOVE`.

## Other considerations

It could even be replaced with a direct delete statement?

### DIFF
--- a/src/Sylius/Bundle/OrderBundle/Doctrine/ORM/OrderRepository.php
+++ b/src/Sylius/Bundle/OrderBundle/Doctrine/ORM/OrderRepository.php
@@ -115,13 +115,14 @@ class OrderRepository extends EntityRepository implements OrderRepositoryInterfa
         ;
     }
 
-    public function findCartsNotModifiedSince(\DateTimeInterface $terminalDate): array
+    public function findCartsNotModifiedSince(\DateTimeInterface $terminalDate, int $limit): array
     {
         return $this->createQueryBuilder('o')
             ->andWhere('o.state = :state')
             ->andWhere('o.updatedAt < :terminalDate')
             ->setParameter('state', OrderInterface::STATE_CART)
             ->setParameter('terminalDate', $terminalDate)
+            ->setMaxResults($limit)
             ->getQuery()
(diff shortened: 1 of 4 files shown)


jakubtobiasz:
I was working on it for a while and despite it isn't crashing on the very beginning, the memory seems to leak. Did you try to process ~10k/20k of carts?

mamazu:
Yeah, we haven't run the command in a while and gathered over 1 million carts. The proper solution would be to return a result iterator from doctrine and do it like that.

The problem is that this isn't a solution for the Sylius default because this would mean changing the return type of the function. So this duct tape solution must suffice for now.

vvasiloi:
Just make a new command. 😄 
(and deprecate the existing)

jakubtobiasz:
Iterable isn't working, too. I spent about 1 man day and nothing, after all memory still has been leaking.

vvasiloi:
Try clearing the entity manager after flushes.

jakubtobiasz:
@vvasiloi I really checked everything, that too 😂.

mamazu:
Well that worked for us. `flush` and `clear`.

mamazu:
> Just make a new command. smile (and deprecate the existing)

Or we just take the old command and wire up a different cart remover. But in my opinion we should refactor the command anyways. So that it has a process indicator. Because there is nothing more annoying than having a command with no process indication and only hope that it does something. :D

mamazu:
Has been replaced by #14569

PR #14537, opened 2022-11-15
[Behat] Add services' aliases for admin pages

loic425:
| Q               | A                                                            
| Branch?         | 1.12
| Bug fix?        | kind of when adding custom contexts, but this is not really a new feature                                                       
| New feature?    | no                                                       
| BC breaks?      | no                                                       
| Deprecations?   | no
| Related tickets |
| License         | MIT

### DIFF
--- a/src/Sylius/Behat/Page/Admin/Administrator/IndexPageInterface.php
+++ b/src/Sylius/Behat/Page/Admin/Administrator/IndexPageInterface.php
@@ -1,0 +1,20 @@
+<?php
+
+/*
+ * This file is part of the Sylius package.
+ *
+ * (c) Paweł Jędrzejewski
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Behat\Page\Admin\Administrator;
+
+use Sylius\Behat\Page\Admin\Crud\IndexPageInterface as BaseIndexPageInterface;
+
+interface IndexPageInterface extends BaseIndexPageInterface
+{
+}
--- a/src/Sylius/Behat/Resources/config/services/pages.xml
+++ b/src/Sylius/Behat/Resources/config/services/pages.xml
@@ -23,20 +23,25 @@
     </parameters>
 
     <services>
-        <defaults public="true" />
+        <defaults public="false" />
 
-        <service id="sylius.behat.page" class="FriendsOfBehat\PageObjectExtension\Page\Page" abstract="true" public="false">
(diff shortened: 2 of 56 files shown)


src/Sylius/Behat/Resources/config/services/pages.xml
```diff
@@ -23,20 +23,25 @@
     </parameters>
 
     <services>
-        <defaults public="true" />
+        <defaults public="false" />
```
jakubtobiasz, line 26: Services are private by default.

src/Sylius/Behat/Resources/config/services/pages/admin/catalog_promotion.xml
```diff
@@ -17,7 +17,7 @@
 >
     <parameters>
         <parameter key="sylius.behat.page.admin.catalog_promotion.create.class">Sylius\Behat\Page\Admin\CatalogPromotion\CreatePage</parameter>
-        <parameter key="sylius.behat.page.admin.catalog_promotion.index.class">%sylius.behat.page.admin.crud.index.class%</parameter>
+        <parameter key="sylius.behat.page.admin.catalog_promotion.index.class">Sylius\Behat\Page\Admin\CatalogPromotion\IndexPage</parameter>
```
jakubtobiasz, line 20: Is it necessary?

src/Sylius/Behat/Page/Admin/Administrator/IndexPageInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace Sylius\Behat\Page\Admin\Administrator;
+
+use Sylius\Behat\Page\Admin\Crud\IndexPageInterface as BaseIndexPageInterface;
+
+interface IndexPageInterface extends BaseIndexPageInterface
```
lchrusciel, line 18: These interfaces are not needed for sure. If we decide to leave the index page classes for DI convenience, they should still implement CRUD interface, as you did already and therefore, this interface may be removed

lchrusciel (commented):
Hey Loic, thanks a lot for your proposal. Some better DX with behat injection is much appreciated, but I'm not convinced that this direction is correct. Even tho you've put a lot of effort into doing it, I'm pretty sure that people will tend to forget about it during day-to-day development.

At first, I was thinking about doing argument binding in this case. But due to the way we are requesting these contexts in behat configuration, I'm not sure it is a good way. Is the creation of a class even required? What about some new naming conventions? Like `Sylius\Behat\Page\Admin\Crud\IndexPage\ResourceName` for example?  

won't merge
name binding instead
public false as default +1

